PR #3, opened 2015-05-04
require field descriptions

rmosolgo:
not crazy about the client API 

```ruby
class MyNode < GraphQL::Node 
  field.string(:title, "The title")
  field.custom_node("Custom node value")
end
```


devknoll:
Also probably need to consider descriptions for the node itself (and calls)... not sure what the idiomatic Ruby solution would be.

One crazy idea might be to parse something like TomDoc at schema load time:
```ruby
# Class Description
class MyNode < GraphQL::Node
  # The title
  field.string(:title)
end
```

PR #8, opened 2015-07-15
Remove activesupport dependency

bkrausz:
Make the dependency loading explicit (we're not lazy anyway since we
instantiate many types at load time). Getting the order right here is a
bit messy, but having it be explicit seems better than masking the chain
of dependencies.

r? @rmosolgo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.ruby-version
 doc/
 .yardoc/
 pkg/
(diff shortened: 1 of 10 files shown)


rmosolgo:
:+1: totally agree, better to have it explicit & fixable than implicit and mysterious :) 

sorry i botched the merge, if you want to rebase or merge master it's all :ok_hand: by me!

bkrausz:
no prob, rebased and force-pushed

rmosolgo:
:money_with_wings:  thanks!

PR #9, opened 2015-07-22
Add ID type as per the GraphQL spec

pushrax:
Ref: http://facebook.github.io/graphql/#sec-ID

### DIFF
--- a/lib/graph_ql/types/id_type.rb
+++ b/lib/graph_ql/types/id_type.rb
@@ -1,0 +1,6 @@
+GraphQL::ID_TYPE = GraphQL::ScalarType.new do |t|
+  t.name "ID"
+  def t.coerce(value)
+    value.to_s
+  end
+end
--- a/spec/graph_ql/query_spec.rb
+++ b/spec/graph_ql/query_spec.rb
@@ -50,6 +50,26 @@ describe GraphQL::Query do
       assert_equal(GraphQL::Nodes::FragmentDefinition, query.fragments['cheeseFields'].class)
     end
 
+    describe 'scalar coercion' do
+      describe 'ID coercion for int inputs' do
+        let(:query_string) { %|query getMilk { cow: milk(id: 1) { id } }| }
+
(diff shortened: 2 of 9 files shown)


lib/graph_ql/types/id_type.rb
```diff
@@ -0,0 +1,6 @@
+GraphQL::ID_TYPE = GraphQL::ScalarType.new do |t|
+  t.name "ID"
+  def t.coerce(value)
+    value.to_s
```
pushrax: > While it is often numeric, it should always serialize as a String.

spec/support/dummy_app.rb
```diff
@@ -93,7 +93,7 @@ def description
   end
 
   def resolve(target, arguments, context)
-    @data[arguments["id"]]
+    @data[arguments["id"].to_i]
```
pushrax: > GraphQL servers should coerce as appropriate given the ID formats they expect

spec/graph_ql/query_spec.rb
```diff
@@ -50,6 +50,26 @@
       assert_equal(GraphQL::Nodes::FragmentDefinition, query.fragments['cheeseFields'].class)
     end
 
+    describe 'scalar coercion' do
```
rmosolgo: Could you create  `spec/graph_ql/types/id_type_spec.rb` to test this behavior? Fine with me to just copy-paste the other `let`s etc. 

Even though `Query` is one of the hotspots, I'm hoping to keep the spec pretty slim and test other behaviors in specific places!

rmosolgo:
:+1: looks awesome, thanks for getting us up to spec!

spec/graph_ql/query_spec.rb
pushrax: Moved to `scalar_type_spec.rb`, which I think makes sense (the scalars are of type `ScalarType`) and will remain small.

rmosolgo:
:confetti_ball:  :tada: :beers:

PR #10, opened 2015-07-23
Add explicit reference to GraphQL Document.

balloob:


rmosolgo:
image

:cry: 


I'm guessing this instantiated some other object named `Document`? Should have seen that coming! 

Instead of fixing these one-by-one, maybe we should fix them _all_, either: 

- Make all the references explicit (it's just laziness anyways)
- `include GraphQL::Nodes`, would that add the constants to this namespace?

balloob:
Sorry for not providing a description. It indeed tried instantiating a different Document object in our project. Making it explicit fixed it.

Adding `include GraphQL::Nodes` inside the module definition will also fix the problem. I would personally prefer the `include` over writing explicit namespaces.

rmosolgo:
Cool, I added `include...` here: https://github.com/rmosolgo/graphql-ruby/commit/3e8e5d0ffa0c8077c9ed85e8aa1d63a62ca61f5a

That way, this gets fixed for _all_ those different classes! :confetti_ball:  Thanks for pointing this out!

PR #15, opened 2015-08-04
Project values ahead of time

rmosolgo:
One possibility is to "look ahead" at fields and expose those values to parent fields. That way you could pre-load values from the DB. 

I'm inspired by Sangria's approach. 

__Pros:__ 
- Simple API, simple implementation 
- Very flexible 

__Cons:__ 
- Runs through fields twice 
- How to handle Union & Interface?


rmosolgo:
:sleeping:

PR #18, opened 2015-08-12
new definition API

rmosolgo:
This was inspired by seanchas's implementation: https://github.com/seanchas/graphql

He said I could use it :)

You can see an example of it here: https://github.com/rmosolgo/graphql-ruby/blob/define-by-config/spec/support/dairy_app.rb#L31-L60

What do you think? Is it an improvement? Does it introduce new problems?

I think it makes for much neater code. I also like that there's a `DefinitionConfig` object between the definition declaration and the object itself, which provides flexibility for the future. 

If nobody objects, I'll mark the old API as deprecated, then remove it in a couple versions later.


keithpitt:
It's certainly better to look at. I like that you've dropped having to use the `t.` prefix when defining stuff - it makes it a lot easier to read and to the point (the previous syntax made it hard to see the actual intent).

I also like the fact that you don't need to supply a hash to `fields` anymore. Makes it a bit easier to programmatically create fields:

```
field :foo
field :bar

array.each do |name|
  field name
end
```

I like it! Note I've only started using GraphQL today, so from a feature standpoint I'm pretty ¯\_(ツ)_/¯, but I've used a lot of API's, so :+1:!

rmosolgo:
Thanks for saying so, I agree about those advantages! @seanchas found such a sweet balance between exposing a good API and keeping objects simple

sgwilym:
:+1: Looks a lot more legible to me.

rmosolgo:
thanks guys!

PR #21, opened 2015-08-14
Define a field type in the Overview example

devknoll:


rmosolgo:
:money_with_wings:  thanks!

PR #26, opened 2015-08-23
Remove deprecated params: keyword

xuorig:
Code clean up: remove deprecated `params:` keyword in `Query#initialize`


rmosolgo:
Nice, thank you! :confetti_ball: :tada:

PR #27, opened 2015-08-23
Skip directive has precedence over include

xuorig:
Tried to look at other implementations and how they did this but their implementation doesn't usually this type of chain and will just return only if there's a `@skip` not taking in consideration future `directives`.

This checks if there is a `skip` directive and if so removes all `include` directives from `ast_directives`

Not sure if that's the best way to go about it. Let me know if you see a better way :dancers:

### DIFF
--- a/spec/graphql/directive_spec.rb
+++ b/spec/graphql/directive_spec.rb
@@ -8,6 +8,8 @@ describe GraphQL::Directive do
         # plain fields:
         skipFlavor: flavor @skip(if: true)
         dontSkipFlavor: flavor @skip(if: false)
+        dontSkipDontIncludeFlavor: flavor @skip(if: false), @include(if: false)
+        skipAndInclude: flavor @skip(if: true), @include(if: true)
         includeFlavor: flavor @include(if: $t)
         dontIncludeFlavor: flavor @include(if: $f)
         # fields in fragments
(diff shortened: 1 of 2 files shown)


spec/graphql/directive_spec.rb
```diff
@@ -8,6 +8,7 @@
         # plain fields:
         skipFlavor: flavor @skip(if: true)
         dontSkipFlavor: flavor @skip(if: false)
+        dontSkipDontIncludeFlavor: flavor @skip(if: false), @include(if: false)
```
rmosolgo: Does this mean _don't_ skip and _don't_ include? 

Would `@skip(if: true) @include(if: true)` be a better test for this feature? Maybe I've got my wires crossed.

rmosolgo:
Yeah, I think the JS implementation has directives couple to execution quite closely. I added this pretty early on and I wasn't sure where they were really going with directives, so I wanted an implementation that would be easy to swap out :) 

I like the way you've added this feature though -- seems like a good fit here, and when the time comes (if it comes?) to overhaul directives, we'll have a regression test.

spec/graphql/directive_spec.rb
xuorig: Yes it does mean *don't* skip and *don't* include.

The thing is that having `@skip(if: true) @include(if: true)` would result in skipping (since skip has precedence) and I'd rather test the `presence` of the field instead of testing the absence of the field, which could be anything.

No strong opinion on this though, let me know what you think.

In that test, if skip

rmosolgo: I see your point on testing presence, but could you also add a test for absence? To be sure that skip still "does its thing" even if `@include` is present

xuorig: Good idea, added `skipAndIncludeFlavor` which should be skipped if skip has precedence.

xuorig:
@rmosolgo addressed your comment

rmosolgo:
:moneybag:  awesome, thanks!

PR #28, opened 2015-08-24
Add ParallelExecution strategy 

rmosolgo:
Uses Celluloid to provide concurrent field resolution


devknoll:
I was going to jump in and mention how it might be cleaner to just do a regular fork & join rather than a counting semaphore, only to discover that celluloid doesn't actually really support that, apparently...

Either way, awesome :smile: Do you know how well this would work together with Rails/MRI?

rmosolgo:
I think the main win is this: https://github.com/celluloid/celluloid/wiki/Frequently-Asked-Questions#q-can-i-do-blocking-io-inside-an-actor-or-do-i-have-to-use-celluloidio

blocking operations (DB, HTTP) only block the current actor, not the whole query

devknoll:
Oh, looks like you were able to simplify it after all! Nice :+1:

rmosolgo:
Weird stuff with the pool -- If I set  a constant `size:` it passes on Travis, otherwise :boom:. It always works locally :S

rmosolgo:
Got some help here: https://github.com/celluloid/celluloid/issues/668

Using a fresh pool for every run seems to work, but that might negate some advantages of using a pool, specifically you can't restrict the growth

rmosolgo:
Performance here is _really_ bad. Here's the `INTROSPECTION_QUERY` test:

__SerialExecution__

image


__ParallelExecution__

image


I mean, what in the world? Is it overhead from joining all these different futures?

I pushed a profile of SerialExecution here: 

https://github.com/rmosolgo/graphql-ruby/compare/prof-serial

digitalextremist:
@rmosolgo I suspect you may want to avoid pools... In the description of your goal, I can't see needing a `Pool`, though that is a popular approach to parallel tasks. I will look through the code to confirm my suspicion. But for tests, we do reset the pool every test ourselves.

Also, it seems you are coming to the point of wanting Multiplexers ( https://github.com/celluloid/celluloid/issues/632 )

digitalextremist:
It seems you could just use `.future` on a single supervised actor, or use conditions with `.async`

...but again, I would need to verify.

rmosolgo:
Thanks for those suggestions, I'll try them out! I'm new to  the Celluloid toolkit so it was a bit hard to pick which things would be best suited to this purpose.

rmosolgo:
Ok, I think i'm happy:

- Don't default to async. The overhead of collecting values is just too high.
- Provide opt-in async behavior via `context.async(&block)` to allow async calls during reads 
- For mutation, `context.async` isn't actually async 
- Allow people to choose which execution strategies to use at a schema level

rmosolgo:
It breaks down if you have an async field -> async field (since it doesn't continue until `finish_all_futures`) but I'll cross that bridge when I get to it. Could be async + conditions as mentioned above

digitalextremist:
This last comment you made is 99% solved by celluloid/celluloid#632, fyi.

rmosolgo:
I'll keep an eye on it, thanks!

digitalextremist:
Do you have an idea of when that functionality would become critical?

If not until the mid/end Fall, I probably come demonstrate `0.18.0` for your meetup and then you can support it more easily after that. Apparently you know a few of the same people I know, so some time in the next few months I should probably pop in for a session anyway.

rmosolgo:
The whole GraphQL idea is still quite young, and even in this implementation there will be "bigger fish to fry" for at least the rest of the year :)

digitalextremist:
Right on.

rmosolgo:
I extracted this behavior to https://github.com/rmosolgo/graphql-parallel

Issue #29, opened 2015-08-27
fragments shallowly merging causing dropped fields


charlieschwabacher:
I noticed an issue when using two fragments containing the same key, w/ others nested below.  It should be deep merging, but instead the last fragment replaces any earlier ones.  It ends up being an issue because Relay generates a lot of repetitive fragments.

this query:

```
{
  viewer {
    ...fragmentWithTwoFields
    ...fragmentWithOneField
  }
}

fragment fragmentWithTwoFields on User {
  todos {
    totalCount,
    completedCount,
  }
}

fragment fragmentWithOneField on User {
  todos {
    totalCount,
  }
}
```

produces:

```
{
  "data": {
    "viewer": {
      "todos": {
        "totalCount": 2
      }
    }
  }
}
```

when it should also include `completedCount`:

```
{
  "data": {
    "viewer": {
      "todos": {
        "totalCount": 2,
        "completedCount": 1
      }
    }
  }
}
```

Going to take a stab at fixing this tonight.

rmosolgo:
wow! what a bummer. 

Maybe you found it already, but I think this is the faulty merge:

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/selection_resolution.rb#L24

It `field_value` there is actually a hash. In the case of a field, it's `name => value`, but in the case of fragments (inline or fragment spread) it has many names and values. 

So, in the case of fragments with the same key, the first fragment is merged in, but then the second on overwrites that key! 

Seems like there's two approaches:
- Make that merge better -- make it handle this case 
- Flatten fields on the way in. Not really sure how this would work but it would also address the issue of selecting the _same_ field in 2 fragments, eg 

  ```
  { myField { ...frag1, ...frag2 } }
   fragment frag1 on Type { name }
   fragment frag2 on Type { name } 
  ```

  right now, graphql-ruby evaluates the name field twice :weary: !

charlieschwabacher:
thanks! I was still kind of wrapping my head around parslet, so this was really helpful.  I think the second approach you mention is the right one, because deep merge falls apart when the values are arrays, i.e:

```
{
  viewer {
    ...fragmentWithTwoFields
    ...fragmentWithOneField
  }
}

fragment fragmentWithTwoFields on User {
  todos {
    edges {
      node {
        text,
        id
      }
    }
  }
}

fragment fragmentWithOneField on User {
  todos {
    edges {
      node {
        id
      }
    }
  }
[...]

PR #30, opened 2015-08-27
deep merge fragments in order to avoid dropping keys

charlieschwabacher:
resolves https://github.com/rmosolgo/graphql-ruby/issues/29

### DIFF
--- a/lib/graphql/query/serial_execution/selection_resolution.rb
+++ b/lib/graphql/query/serial_execution/selection_resolution.rb
@@ -29,14 +29,11 @@ module GraphQL
 
         def deep_merge(h1, h2)
           h1.merge(h2) do |key, oldval, newval|
-            oldval = oldval.to_hash if oldval.respond_to?(:to_hash)
-            newval = newval.to_hash if newval.respond_to?(:to_hash)
-
-            if oldval.class.to_s == 'Array' && newval.class.to_s == 'Array'
+            if oldval.is_a?(Array) && newval.is_a?(Array)
               oldval.each_index.map do |i|
                 deep_merge oldval[i], newval[i]
               end
-            elsif oldval.class.to_s == 'Hash' && newval.class.to_s == 'Hash'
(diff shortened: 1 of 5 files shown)


lib/graphql/query/serial_execution/selection_resolution.rb
```diff
@@ -21,12 +21,29 @@ def initialize(target, type, selections, query, execution_strategy)
...
+            deep_merge memo, field_value
           end
         end
 
         private
 
+        def deep_merge(h1, h2)
+          h1.merge(h2) do |key, oldval, newval|
+            oldval = oldval.to_hash if oldval.respond_to?(:to_hash)
+            newval = newval.to_hash if newval.respond_to?(:to_hash)
```
rmosolgo: Why is this check / coercion necessary?

lib/graphql/query/serial_execution/selection_resolution.rb
```diff
@@ -21,12 +21,29 @@ def initialize(target, type, selections, query, execution_strategy)
...
+
+            if oldval.class.to_s == 'Array' && newval.class.to_s == 'Array'
+              oldval.each_index.map do |i|
+                deep_merge oldval[i], newval[i]
+              end
+            elsif oldval.class.to_s == 'Hash' && newval.class.to_s == 'Hash'
```
rmosolgo: Is there an advantage to `class.to_s` over `.is_a?` ?

rmosolgo:
Cool, the direction of this implementation seems good, how about adding a test? 

You could put it in query_spec.rb with a new `describe "merging fragments with different keys" do ...` block. You could set up a `let(:query_string)` with a query with 2 fragments, then make sure the result has fields from both fragments.

charlieschwabacher:
I had just pulled that deep merge method from ActiveSupport, you are right that the coercion isn't needed / is_a? is faster.

rmosolgo:
:money_with_wings: awesome, thanks!

rmosolgo:
just pushed 0.7.1 with this fix!

PR #32, opened 2015-09-04
Accept a string of JSON for variables in query initialisation

sgwilym:
While working with some mutations, I realised that when you have a mutation that sends a file, the `Content-type` of the request sent by Relay will no longer be `application/json` (https://github.com/facebook/relay/blob/master/src/network-layer/default/RelayDefaultNetworkLayer.js#L113) What that means is that any query variables will be interpreted as a String rather than a Hash (at least this is the case in Rails).

The final effect is that your query variables are invisibly swallowed and your query variables are no longer available to your resolve methods. This PR fixes that.


rmosolgo:
Hmmm to me this just feels like a controller issue, not a GraphQL implementation issue. I think it would make a nice private method in a controller: 

```ruby
  def create 
    # ...
    variables = ensure_hash(params[:variables])
    GraphQL::Query.new(Schema, query, variables: variables)
  end

  private 
  
  # If the request wasn't `Content-Type: application/json`, parse the variables:
  def ensure_hash(variables_param)
    variables_param.kind_of?(Hash) ? variables_param : JSON.parse(variables_param)
  end
```

I prefer that because it keeps the GraphQL API simple and it keeps HTTP request details in the HTTP handler, not in `GraphQL::Query`. What do you think?

sgwilym:
I had the same thought in the back of my mind this afternoon — you're absolutely right!

PR #33, opened 2015-09-11
Better raise for parse errors

bkrausz:
1. Passing `error.cause.pos` into `line_and_column is necessary to avoid getting 1,1. It still doesn't help a ton, since the error is almost always the first character of your query (i.e. it only skips whitespace), but at least now it matches the pos in the error message.
1. Wrap the response in a `GraphQL::ParseError` so that it's easier to catch/handle.

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -3,6 +3,16 @@ require "parslet"
 require "singleton"
 
 module GraphQL
+  class ParseError < StandardError
+    attr_accessor :line, :col, :query
+    def initialize(message, line, col, query)
+      super(message)
+      @line = line
+      @col = col
+      @query = query
+    end
+  end
+
   # Turn a query string into an AST
   # @param string [String] a GraphQL query string
   # @param as [Symbol] If you want to use this to parse some _piece_ of a document, pass the rule name (from {GraphQL::Parser})
@@ -12,8 +22,8 @@ module GraphQL
     tree = parser.parse(string)
     GraphQL::TRANSFORM.apply(tree)
   rescue Parslet::ParseFailed => error
-    line, col = error.cause.source.line_and_column
-    raise [error.message, line, col, string].join(", ")
+    line, col = error.cause.source.line_and_column(error.cause.pos)
+    raise GraphQL::ParseError.new(error.message, line, col, string)
   end
 end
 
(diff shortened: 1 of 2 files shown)


lib/graphql.rb
```diff
@@ -3,6 +3,16 @@
 require "singleton"
 
 module GraphQL
+  class ParseError < StandardError
+    attr_accessor :line, :col, :query
```
rmosolgo: could this be `attr_reader`? No need to expose setters, right?

bkrausz: :+1: done

rmosolgo:
Woah, good call on getting the actual line & col from parslet. 

How about adding a test for this feature?  I think you could just add one to `query_spec`, since that's the main place this would come up. You could just feed an invalid query string to `GraphQL.parse` and make sure: 

- it raises a `GraphQL::ParseError` (with `assert_raises`)
- the raised error has non-`1` line & column values (`assert_raises` returns the error, so you could check it for those values)

bkrausz:
Can do. I was going to do that, then I realized that all I really changed was getting parslet to skip over whitespace and pick the first char as the error location. Not much better, but can't hurt to have more tests.

Related food for though: I'd love to do some kind of easy exposure for exception handling beyond validation. It looks like your only option now is to turn `debug` off and raise any `StandardError`, which will get caught and turned into a single error message. Being able to raise a `GraphQL::ExecutionError` or similar that will get caught and have the line/col picked up would be excellent. For now I'm doing it myself, which is fine.

bkrausz:
(done)

rmosolgo:
Hmmm looks like it might have been a `git rm` instead of a `git add` ?

image

:grimacing:

bkrausz:
Wow I am super bad at things today. Fixed.

rmosolgo:
:moneybag:  awesome, thanks! 

BTW I think your suggestion about error handling sounds good! You could probably rescue in this method, which is where field resolve methods are actually called: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/field_resolution.rb#L33

And attach `ast_node` at that point. Then ... re-raise? I'm not sure how to get it into the `errors` key!

PR #34, opened 2015-09-12
merge fragments as ast nodes before resolving

charlieschwabacher:
I finally had a chance to really read through the code and work something out for this... this will prevent double resolution of fields when they are repeated in multiple fragments.  It ends up being important for relay - before these changes querying the todo example looks like:

```
  User Load (0.1ms)  SELECT  "users".* FROM "users"  ORDER BY "users"."id" ASC LIMIT 1
  Todo Load (0.2ms)  SELECT "todos".* FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  Todo Load (0.3ms)  SELECT  "todos".* FROM "todos" WHERE "todos"."user_id" = ?  ORDER BY "todos"."id" ASC LIMIT 9007199254740991  [["user_id", 1]]
   (0.1ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
   (0.2ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
```

and after:

```
  User Load (0.1ms)  SELECT  "users".* FROM "users"  ORDER BY "users"."id" ASC LIMIT 1
  Todo Load (0.2ms)  SELECT "todos".* FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  Todo Load (0.3ms)  SELECT  "todos".* FROM "todos" WHERE "todos"."user_id" = ?  ORDER BY "todos"."id" ASC LIMIT 9007199254740991  [["user_id", 1]]
   (0.1ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
   (0.1ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ? AND "todos"."complete" = 't'  [["user_id", 1]]
  CACHE (0.0ms)  SELECT COUNT(*) FROM "todos" WHERE "todos"."user_id" = ?  [["user_id", 1]]
```


rmosolgo:
Wow, this is so awesome! Reading the code now...

lib/graphql/query/serial_execution.rb
```diff
@@ -6,7 +6,5 @@ class SerialExecution < GraphQL::Query::BaseExecution
 end
 
 require 'graphql/query/serial_execution/field_resolution'
-require 'graphql/query/serial_execution/fragment_spread_resolution'
-require 'graphql/query/serial_execution/inline_fragment_resolution'
```
rmosolgo: nice to just _not_ need these anymore

lib/graphql/query/serial_execution/selection_resolution.rb
```diff
@@ -19,34 +13,94 @@ def initialize(target, type, selections, query, execution_strategy)
...
+          else
+            field2
+          end
+        end
+
+        def fields_can_merge?(field1, field2)
+          type1 = query.schema.get_field(type, field1.name).type
+          type2 = query.schema.get_field(type, field2.name).type
+
+          while type1.is_a?(GraphQL::ListType) || type1.is_a?(GraphQL::NonNullType)
```
rmosolgo: You can use `#unwrap` to get the "inner" type. It's a no-op for ObjectTypes & its recursive for ListTypes and NonNullTypes ([src](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L17-L27))

So I think you could:

```ruby
type1 = query.schema.get_field(type, field1.name).type.unwrap
type2 = query.schema.get_field(type, field2.name).type.unwrap
```

That would give you those "inner" types

spec/graphql/query/executor_spec.rb
```diff
@@ -68,6 +70,44 @@
...
+          ...repetativeFragment
+        }
+      }
+      fragment repetativeFragment on Dairy {
+        id
+      }
+    |}
+
+    it 'resolves each field only one time, even when present in multiple fragments' do
+      result
```
rmosolgo: I take it this would raise if it called `mock.id` more than 1 time?

lib/graphql/query/serial_execution/selection_resolution.rb
```diff
@@ -19,34 +13,94 @@ def initialize(target, type, selections, query, execution_strategy)
...
+            type1 = type1.of_type
+          end
+
+          while type2.is_a?(GraphQL::ListType) || type2.is_a?(GraphQL::NonNullType)
+            type2 = type2.of_type
+          end
+
+          type1.is_a?(GraphQL::ObjectType) && type2.is_a?(GraphQL::ObjectType)
```
rmosolgo: How come we only have to check that they're both ObjectTypes? Shouldn't we also check that they're the _same_ object type? 

Oh, I guess you know they're basically the same already, since you're getting fields with the same name from the same type. 

Ok, I think I get it :)

rmosolgo:
:+1: :+1: :+1: :clap: :clap: :clap: :smile_cat: :

lib/graphql/query/serial_execution/selection_resolution.rb
charlieschwabacher: awesome! way better

spec/graphql/query/executor_spec.rb
charlieschwabacher: Actaully, I don't think this spec does work as I intended - going to update it

lib/graphql/query/serial_execution/selection_resolution.rb
charlieschwabacher: Yeah - good point, this would not be great for something like
``` 
query Something {
  field: fieldOfOneType
  field: fieldOfAnotherType
}
```
I think in the long run we are meant to catch stuff like that w/ static validation https://facebook.github.io/graphql/#sec-Field-Selection-Merging, but in the meantime do you think it would make sense to raise an error here if `type1 != type2`?

rmosolgo:
Looking good! I just saw your comment about static validation, and woah, you're quite right. It looks like we can already assume that `fields_can_merge?`: 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/static_validation/rules/fields_will_merge_spec.rb#L38-L43

Since it's handled by static validation, I think you can just assume they'll merge!

charlieschwabacher:
Oh, nice!  I didn't catch that this was already happening. I just pushed an update.

spec/graphql/query/executor_spec.rb
rmosolgo: Sorry to drag this on -- could you add an assertion that the `result` _is_ what you're expecting? 

I'm worried there might be an error in there, but it's being rescued, then put into `result["errors"]`. Just an assertion that 

```ruby
expected = {"data" => {
  "dairy" => { "id" => "??" }
}}
assert_equal(expected, result)
```

charlieschwabacher: No worries, that is a good point - just pushed that up

rmosolgo:
:confetti_ball: thanks!!!

PR #35, opened 2015-09-15
Add InputObject fields to type reducer

bkrausz:
r? @rmosolgo 

Fortunately GraphiQL does introspection validation, I discovered this while trying to get it running on my schema.

### DIFF
--- a/lib/graphql/schema/type_reducer.rb
+++ b/lib/graphql/schema/type_reducer.rb
@@ -48,6 +48,12 @@ class GraphQL::Schema::TypeReducer
         reduce_type(possible_type, type_hash)
       end
     end
+    if type.kind.input_object?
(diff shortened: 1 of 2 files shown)


rmosolgo:
Looks good, but could you add a test for this behavior? 

You could riff off [this test in `type_reducer_spec.rb`](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/schema/type_reducer_spec.rb#L26-L38):

- define a type that will demonstrate the correct behavior 
- in the spec, run `TypeReducer` on that type 
- make sure the result has the expected types in it

bkrausz:
Done

rmosolgo:
:moneybag:  thanks!

PR #39, opened 2015-09-18
Remove SelectedObjectResolution base class

dylanahsmith:
SelectedObjectResolution is only being used as a base class for FieldResolution now, since https://github.com/rmosolgo/graphql-ruby/pull/34 removed the other derived classes.  It isn't used anywhere else, including in graphql-parallel, so there isn't a need for this base class anymore.  I moved the attribute initialization to FieldResolution.


rmosolgo:
:ok_hand: good catch, less is more!

PR #40, opened 2015-09-18
Use self.class::Name rather than get_class(:Name)

dylanahsmith:
This removed the get_class method, since it was always called with a constant class name.

E.g. `self.class::FieldResolution` does the same thing as `self.class.const_get(:FieldResolution)`


rmosolgo:
:moneybag: definitely an improvement, thanks!

PR #41, opened 2015-09-20
Add include method to has possible types module

xuorig:
https://github.com/rmosolgo/graphql-ruby/blob/60d27b1351cc7212925c06f92ef36a75d02349eb/lib/graphql/query/type_resolver.rb#L11

This line here raises undefined method `#include?` for `UnionType` when using a union type field.

I've added an `#include?` method to the `HasPossibleTypes` module. Please correct me if this is not a bug and just me incorrectly using the type :)

### DIFF
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -53,6 +53,10 @@ module GraphQL
       def resolve_type=(new_proc)
         @resolve_type_proc = new_proc || DEFAULT_RESOLVE_TYPE
       end
+
+      def include?(type)
+        possible_types.include?(type)
+      end
     end
 
     # Print the human-readable name of this type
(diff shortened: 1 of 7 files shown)


rmosolgo:
Ick, certainly looks like a bug! 

What's the failing case there? Could you add a test for the behavior that _was_ broken, so we can be sure it's fixed?

xuorig:
Take this query for example where I have a Post object type which has a Media field (Union Type).
This query will fail in `TypeResolver.new` because

`target` will be lets `Video` ( A rails model ), 
`child_type` will be `Media:GraphQL::UnionType`
`parent_type` `Video:GraphQL::ObjectType`

`child_type.include?(parent_type) => undefined`

```
query PostQueries{post(id:'1'){id,..._33q6dtl}} 

fragment _57pt41s on Video{url,id} 
fragment _6h1pr5z on Audio{url,id}
fragment _794ls0i on Text{content,id}
fragment _41rhdgp on Media{..._57pt41s,..._6h1pr5z,..._794ls0i}
fragment _33q6dtl on Post{title,description,media{..._41rhdgp},id}
```

I'll add a test for type resolver

rmosolgo:
Sorry, I didn't see that you pushed! 

Thanks, I'll get this out on Rubygems tomorrow! :confetti_ball:

rmosolgo:
shipped in `0.9.4`!

PR #42, opened 2015-09-21
Link to already written ruby bindings for libgraphqlparser.

dylanahsmith:
The "To Do" section of the readme said "Write Ruby bindings for libgraphqlparser and use that instead of Parslet", so I linked to existing ruby bindings for libgraphqlparser.


rmosolgo:
:tada: seems like a good first step :)

PR #43, opened 2015-09-23
Middleware

rmosolgo:
- Add middleware to query execution 
  - This could also be used for metrics or permissions 
- Add RescueMiddleware for handling exceptions 

The resulting API: 

```ruby
MySchema.rescue_from(ActiveRecord::NotFound) { |err| "Some items weren't found" }
```

When a field raises that error, the field returns `nil` and the error message is put in the response

### DIFF
--- a/lib/graphql/schema/rescue_middleware.rb
+++ b/lib/graphql/schema/rescue_middleware.rb
@@ -1,0 +1,53 @@
+module GraphQL
+  class Schema
+    # - Store a table of errors & handlers
+    # - Rescue errors in a middleware chain, then check for a handler
+    # - If a handler is found, use it & return a {GraphQL::ExecutionError}
+    # - If no handler is found, re-raise the error
+    class RescueMiddleware
+      # @return [Hash] `{class => proc}` pairs for handling errors
+      attr_reader :rescue_table
+      def initialize
+        @rescue_table = {}
+      end
+
+      # @example Rescue from not-found by telling the user
+      #   MySchema.rescue_from(ActiveRecord::NotFound) { "An item could not be found" }
+      #
+      # @param [Class] a class of error to rescue from
+      # @yield [err] A handler to return a message for this error instance
+      # @yieldparam [Exception] an error that was rescued
+      # @yieldreturn [String] message to put in GraphQL response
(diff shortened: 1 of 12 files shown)


lib/graphql/schema/rescue_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+  class Schema
+    # - Store a table of errors & handlers
+    # - Rescue errors in a middleware chain, then check for a handler
+    # - If a handler is found, use it & return a {GraphQL::ExecutionError}
+    # - If no handler is found, re-raise the error
+    class RescueMiddleware
+      # @return [Hash] `{class => proc}` pairs for handling errors
+      attr_reader :rescue_table
+      def initialize
+        @rescue_table = {}
```
dylanahsmith: Using a table means that you can't rescue derived classes.  E.g. `MySchema.rescue_from(StandardError) { "internal error" }` would only catch StandardError, not derived classes like RuntimeError.

lib/graphql/schema/rescue_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+      # @param [Class] the error class whose handler should be removed
+      def remove_handler(error_class)
+        rescue_table.delete(error_class)
+      end
+
+      # Implement the requirement for {GraphQL::Schema::MiddlewareChain}
+      def call(*args, next_middleware)
+        begin
+          next_middleware.call
+        rescue StandardError => err
```
dylanahsmith: Perhaps you could rescue ExecutionError and just return the error.  That way a helper function that is called in the resolve proc can raise errors that are meant to be displayed to the client.

rmosolgo: Oh, that's good, one handler for free

lib/graphql/schema/rescue_middleware.rb
rmosolgo: Yeah, this is very much the simplest-thing-that-could-possibly-work. The implementation is private enough that I think it would be straightforward to swap out the Hash + `[]` for an Array + `.is_a?`

PR #45, opened 2015-09-30
Implement a schema printer

dylanahsmith:
This implements the schema printer, similar to the one in [graphl-js](https://github.com/graphql/graphql-js/blob/v0.4.4/src/utilities/schemaPrinter.js), which outputs the schema as a string in graphql format.

This can be useful for inspecting the schema, which actually turned up some differences between the inspection type system in graphql-ruby and the [graphql specification](http://facebook.github.io/graphql/#sec-Schema-Introspection), which I also fixed in this pull request.

Being able to save the schema is also useful for sharing with a client, which may be written in another language.

### DIFF
--- a/lib/graphql/schema/printer.rb
+++ b/lib/graphql/schema/printer.rb
@@ -42,90 +42,104 @@ module GraphQL
     end
 
     def print_type(type)
-      case type
-      when ScalarType
-        print_scalar(type)
-      when ObjectType
-        print_object(type)
-      when InterfaceType
-        print_interface(type)
-      when UnionType
-        print_union(type)
-      when EnumType
-        print_enum(type)
-      when InputObjectType
-        print_input_object(type)
-      else
-        raise NotImplementedError, "Unexpected type #{type.class}"
+      TypeKindPrinters::STRATEGIES.fetch(type.kind).print(type)
+    end
+
+    module TypeKindPrinters
+      module FieldPrinter
+        def print_fields(type)
+          type.fields.values.map{ |field| "  #{field.name}#{print_args(field)}: #{field.type}" }.join("\n")
+        end
+
+        def print_args(field)
+          return if field.arguments.empty?
+          "(#{field.arguments.values.map{ |arg| print_input_value(arg) }.join(", ")})"
+        end
+
+        def print_input_value(arg)
+          default_string = " = #{print_value(arg.default_value, arg.type)}" unless arg.default_value.nil?
+          "#{arg.name}: #{arg.type.to_s}#{default_string}"
+        end
+
+        def print_value(value, type)
+          case type
+          when FLOAT_TYPE
+            value.to_f.inspect
+          when INT_TYPE
+            value.to_i.inspect
+          when BOOLEAN_TYPE
+            (!!value).inspect
+          when ScalarType, ID_TYPE, STRING_TYPE
+            value.to_s.inspect
+          when EnumType
+            value.to_s
+          when InputObjectType
+            fields = value.to_h.map{ |field_name, field_value|
+              field_type = type.input_fields.fetch(field_name.to_s).type
+              "#{field_name}: #{print_value(field_value, field_type)}"
+            }.join(", ")
+            "{ #{fields} }"
+          when NonNullType
+            print_value(value, type.of_type)
+          when ListType
+            "[#{value.to_a.map{ |v| print_value(v, type.of_type) }.join(", ")}]"
+          else
+            raise NotImplementedError, "Unexpected value type #{type.inspect}"
+          end
+        end
       end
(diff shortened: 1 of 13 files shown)


lib/graphql/schema/printer.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    private
+
+    def print_filtered_schema(schema, type_filter)
+      types = schema.types.values.select{ |type| type_filter.call(type) }.sort_by(&:name)
+      types.map{ |type| print_type(type) }.join("\n\n")
+    end
+
+    BUILTIN_SCALARS = Set.new(["String", "Boolean", "Int", "Float", "ID"])
+    private_constant :BUILTIN_SCALARS
```
rmosolgo: woah, cool, TIL

rmosolgo:
Wow, this is a great feature! I remember when this was added to the JS implementation but I wasn't sure if we'd really need it. Can't hurt to implement it though, for sure.

lib/graphql/schema/printer.rb
```diff
@@ -0,0 +1,131 @@
...
+    def is_introspection_type(type)
+      type.name.start_with?("__")
+    end
+
+    def is_defined_type(type)
+      !is_introspection_type(type) && !BUILTIN_SCALARS.include?(type.name)
+    end
+
+    def print_type(type)
+      case type
```
rmosolgo: I'm not a big fan of giant `case` statements, I prefer a strategy class-based approach. Are there advantages to this style that I haven't considered? What do you think of applying that pattern here?

Here's one application of the strategy pattern:

- Setup a mapping of objects -> strategies: [BaseExecution#TYPE_KIND_STRATEGIES ](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/base_execution/value_resolution.rb#L78-L85)
- When you receive an object, find the matching strategy and use it: [BaseResolution#result](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/field_resolution.rb#L36-L38)

In short, you spread API surface area over a bunch of classes instead of adding them to one class. I like it because:

- Each class's API is smaller. In a class with a big API, it's harder to tell which methods get called where. In a class with a small API, it's easier to find all the possibilities.
- Any state is private. Each class can have its own instance variables without fear of messing with the others. (Doesn't apply here, but, in general!)
- A strategy-as-a-class is very expressive. Since the strategy is just a class, it can do anything a class can do (include, extend, define methods, etc), in the privacy of its own scope.

What do you think of applying that pattern here?

lib/graphql/schema/printer.rb
```diff
@@ -0,0 +1,131 @@
+module GraphQL
+  # Used to convert your {GraphQL::Schema} to a GraphQL schema string
+  #
+  # @example print your schema to standard output
+  #   Schema = GraphQL::Schema.new(query: QueryType)
+  #   puts GraphQL::Schema::Printer.print_schema(Schema)
+  #
+  module Schema::Printer
+    extend self
```
rmosolgo: Is this to support the two class methods, `print_schema` and `print_introspection_schema`? 

What if those were implemented as:

```ruby
def self.print_schema 
  printer =  self.new(schema, :is_defined_type)
  printer.to_s 
end

  printer =  self.new(schema, :is_introspection_type)
  printer.to_s 
end

# and, to go with that:
def to_s
  print_filtered_schema(@schema, @type_filter)
end
```

I think the advantage there is that most methods are instance-only, rather than instance methods _and_ class methods.

rmosolgo:
Thanks, this looks great! I had a couple question about the implementation. Thanks also for fixing bugs in introspection, it's good to have this as a way of comparing to the JS version!

lib/graphql/schema/printer.rb
dylanahsmith: I pushed commit 0b5f05fea51b7da3d41434c2883a2d6f7b58dbd3 which takes the approach you suggested.  The only disadvantage would be that it is more verbose, but I have nothing against it if it makes the code more readable.

lib/graphql/schema/printer.rb
dylanahsmith: This was just because there was no need to store state.  The private methods took a subset of the state, so using class methods makes it clearer than they don't use any implicit `self` state.  What you are suggesting is to instantiate an object just to pass arguments to a single private instance method.

rmosolgo: Cool, that sounds good to me. Yeah, it's nice that those methods are all stateless!

rmosolgo:
:confetti_ball: thanks, that's awesome!

rmosolgo:
shipped in `0.9.5`!

Issue #47, opened 2015-10-06
Cannot specify input coercion separately from result coercion


dylanahsmith:
The [GraphQL specification](http://facebook.github.io/graphql/#sec-Scalars) distinguishes between input and result coercion, but this library doesn't.

Being able to define these separately would be very useful for custom scalar types, since the input coercion can deserialize the value into an appropriate object (e.g. [Time](http://ruby-doc.org/core/Time.html)) for use in the resolver, then result coercion can serialize the value so it can be represented in JSON.  Otherwise, the resolver would always have to explicitly parse the input, so custom scalars would only be useful for validation.

rmosolgo:
:+1: I didn't see the value in both helpers at first, but I get it for custom scalars!

rmosolgo:
Fixed by #48

Issue #46, opened 2015-10-06
Variables aren't coerced.


dylanahsmith:
Here is an example script that shows the problem:

```ruby
require 'graphql'
require 'json'

QueryRoot = GraphQL::ObjectType.define do
  name "Query"

  field :foo, types.String do
    argument :input, types.Int
    resolve -> (obj, args, ctx) {
      args['input'].inspect
    }
  end
end

Schema = GraphQL::Schema.new(query: QueryRoot)

variables = { "input" => "not a int" }
result = Schema.execute(<<GRAPHQL, variables: variables)
query Q($input: Int) {
  foo(input: $input)
}
GRAPHQL

puts JSON.pretty_generate(result)
```

outputs

```json
{
  "data": {
    "foo": "\"not a int\""
  }
}
```

despite the fact that a string variable value was accepted in an Int typed argument.

There is a check to make sure the declared variables type matches the argument type, but no coercion of the variable values themselves.

dylanahsmith:
Oh, it actually looks like values are also not coerced.  E.g.

```ruby
require 'graphql'
require 'json'

QueryRoot = GraphQL::ObjectType.define do
  name "Query"

  field :add, types.String do
    argument :a, types.Int
    argument :b, types.Int
    resolve ->(obj, args, ctx) {
      "#{args['a'].inspect} + #{args['b'].inspect}"
    }
  end
end

Schema = GraphQL::Schema.new(query: QueryRoot)
result = Schema.execute('{ add(a: 1.2, b: 2.3) }')
puts JSON.pretty_generate(result)
```

outputs

```
{
  "data": {
    "add": "1.2 + 2.3"
  }
}
```

because the static validator coerces the inputs, but the result isn't used, so the original float values get passed in as integer arguments.

rmosolgo:
Here we go: https://github.com/rmosolgo/graphql-ruby/commit/fa7a2dba9dae5db516c0d276c80d7d4af4d1c9a5

You can see where it was just reading the value and not coercing it at all, now it's coercing it!
 
Also it will raise if it fails to coerce

PR #48, opened 2015-10-06
Seperate input and result coercion for scalars.

dylanahsmith:
This is the start of a fix for issue #47, which splits input and result coercion.

The most noticeable change is that enum's now get properly coerced to their value and their value gets coerced back to the enum name in result coercion.

However, as noted in https://github.com/rmosolgo/graphql-ruby/issues/46#issuecomment-145977865, we aren't actually actually coercing most values, so the follow-up to this will add input coercion for non-scalars, so we can coerce all values in `GraphQL::Query::Arguments`.

### DIFF
--- a/lib/graphql/enum_type.rb
+++ b/lib/graphql/enum_type.rb
@@ -14,8 +14,17 @@ class GraphQL::EnumType < GraphQL::BaseType
   attr_accessor :name, :description, :values
   defined_by_config :name, :description, :values
 
+  def values=(values)
+    @values_by_name = {}
+    @values_by_value = {}
+    values.each do |enum_value|
+      @values_by_name[enum_value.name] = enum_value
+      @values_by_value[enum_value.value] = enum_value
+    end
+  end
+
   def values
-    @values ||= {}
+    @values_by_name
   end
 
   # Define a value within this enum
@@ -40,8 +49,12 @@ class GraphQL::EnumType < GraphQL::BaseType
   #
   # @param value_name [String] the string representation of this enum value
   # @return [Object] the underlying value for this enum value
-  def coerce(value_name)
-    values[value_name].value
+  def coerce_input(value_name)
+    @values_by_name.fetch(value_name).value
+  end
+
+  def coerce_result(value)
+    @values_by_value.fetch(value).name
   end
 
   def to_s
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -1,6 +1,15 @@
 module GraphQL
   # The parent type for scalars, eg {GraphQL::STRING_TYPE}, {GraphQL::INT_TYPE}
   #
+  # @example defining a type for Time
+  #   TimeType = GraphQL::ObjectType.define do
+  #     name "Time"
+  #     description "Time since epoch in seconds"
+  #
+  #     coerce_input ->(value) { Time.at(Float(value)) }
+  #     coerce_result ->(value) { value.to_f }
+  #   end
+  #
   class ScalarType < GraphQL::BaseType
     defined_by_config :name, :coerce, :coerce_input, :coerce_result, :description
     attr_accessor :name, :description
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -2,15 +2,28 @@ module GraphQL
   # The parent type for scalars, eg {GraphQL::STRING_TYPE}, {GraphQL::INT_TYPE}
   #
   class ScalarType < GraphQL::BaseType
-    defined_by_config :name, :coerce, :description
+    defined_by_config :name, :coerce, :coerce_input, :coerce_result, :description
     attr_accessor :name, :description
 
-    def coerce(value)
-      @coerce_proc.call(value)
+    def coerce=(proc)
+      self.coerce_input = proc
+      self.coerce_result = proc
     end
 
-    def coerce=(proc)
-      @coerce_proc = proc
+    def coerce_input(value)
+      @coerce_input_proc.call(value)
+    end
+
+    def coerce_input=(proc)
+      @coerce_input_proc = proc unless proc.nil?
+    end
+
+    def coerce_result(value)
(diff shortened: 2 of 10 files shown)


spec/graphql/query/value_resolution_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  describe "enum resolution" do
+    let(:query_string) { %|
+      {
+        tomorrow(today: FRIDAY)
+      }
+    |}
+
+    describe "when an operation is named" do
+      it "runs the named one" do
```
rmosolgo: oops, i bet this was copy-pasted! could you put a more specific description for this example?

lib/graphql/scalar_type.rb
```diff
@@ -2,15 +2,28 @@ module GraphQL
   # The parent type for scalars, eg {GraphQL::STRING_TYPE}, {GraphQL::INT_TYPE}
   #
   class ScalarType < GraphQL::BaseType
-    defined_by_config :name, :coerce, :description
+    defined_by_config :name, :coerce, :coerce_input, :coerce_result, :description
     attr_accessor :name, :description
 
-    def coerce(value)
-      @coerce_proc.call(value)
+    def coerce=(proc)
```
rmosolgo: :+1:

lib/graphql/enum_type.rb
```diff
@@ -40,8 +49,12 @@ def kind
...
   # @param value_name [String] the string representation of this enum value
   # @return [Object] the underlying value for this enum value
-  def coerce(value_name)
-    values[value_name].value
+  def coerce_input(value_name)
+    @values_by_name.fetch(value_name).value
+  end
+
+  def coerce_result(value)
+    @values_by_value.fetch(value).name
```
rmosolgo: :+1: for `fetch`, I'll have to remember that

rmosolgo:
Love it. Do you have an example of a custom scalar? If you have one handy, it would be nice to add it to the API docs (maybe for the `ScalarType` class). That would make `coerce_input` / `coerce_result` more discoverable for a newcomer, especially since we don't use it for the built-in scalars.

spec/graphql/query/value_resolution_spec.rb
dylanahsmith: Oops.  Changed it to `it "coerces enum input to the value and result to the name" do`

dylanahsmith:
I added an example for a custom Time type.

rmosolgo:
Awesome, thanks!

nuclearspike:
Is there a way to have arguments passed to a custom scalar?

In my graphql js code, I can do:
```
const dateStringType = (description, selector) => {
    return {
        type: graphql.GraphQLString,
        description,
        args: {
            format: {
                type: graphql.GraphQLString,
                description: "Optional: Any DateJS format options. ex: MMM yyyy, MM-dd-yyyy. Only use when you want the time to be localized to the server's time. Otherwise, you should parse the default date in the browser/device and localize there to have it relevant."
            }
        },
        resolve: function (_, args, a, fieldSchema) {
            const fieldVal =  selector ? selector(_) : _[fieldSchema.fieldName]
            if (!fieldVal) return null
            if (args.format) {
                let d = typeof fieldVal === 'string' ? new Date(fieldVal) : fieldVal
                try {
                    return d.toString(args.format)
                } catch (e) {
                    return e.message
                }
            } else {
                return typeof fieldVal === 'string' ? fieldVal : fieldVal.toISOString()
            }
        }
    }
}
```

Then use it as:

```
const scheduledPaymentsType = new graphql.GraphQLObjectType({
    name: "ScheduledPayment",
    fields: {
        "due_date": dateStringType("Date the repayment is due"),
        "amount": { type: graphql.GraphQLFloat },
    }
})
```

This allows the queries to have stuff like:
```
due_date(format: "mm/dd/yyyy")
```

What's the analog with this ruby library?

rmosolgo:
> custom scalar

The example above uses the builtin `String` type, not a custom scalar, right? I _think_ it's a function that returns a field definition, and that field's type is `graphql.GraphQLString`. 

> What's the analog? 

You can define a method which returns a field definition. For example:

```ruby 
module StringDate 
  def self.field(desc, selector) 
    GraphQL::Field.define do 
      type(types.String) 
      description(desc)
      argument :format, types.String, "Optional: Any DateJS ..."
      resolve ->(obj, args, ctx) { 
        # ...
      }
    end 
  end 
end 

# And use it: 
field :due_date, StringDate.field(due_date_desc, due_date_selector)
```

This way, you can parameterize field definitions. 

Would that work in your case?

nuclearspike:
Ah.. Ok. Yeah, thanks for that!

>

nuclearspike:
For anyone else wanting to do something like this, the resolve function on a Field isn't super straight forward since it's passing in the parent object, not the field. So, this is what I ended up with:

```
module StringDate
  def self.field(desc)
    GraphQL::Field.define do
      type types.String
      description(desc)
      argument :format, types.String, 'See Ruby docs for strftime for options'
      resolve ->(obj, args, ctx) {
        value = obj.send ctx.field.name
        if args['format'].present?
          value.strftime(args['format'])
        else
          value.to_s
        end
      }
    end
  end
end
```

If anyone wants it just as a normal field:
```
DateTimeField = GraphQL::Field.define do
  type types.String
  description 'Date time with formatting'
  argument :format, types.String, 'Ruby date format string options'
  resolve ->(obj, args, ctx) {
    value = obj.send ctx.field.name
    if args['format'].present?
      value.strftime(args['format'])
    else
      value.to_s
    end
  }
end
```

I'm guessing I would also have to read the 'property' field.

Used as:
  field :updated_at, StringDate.field('When was the review updated')
  field :created_at, DateTimeField

The problem with the second is I'm not sure how to pass in the description. If you follow the normal convention and just include it after the field, it won't override what was defined.

It'd be simpler if there were just a Formatter type you could define, where you only define a resolve and it passes you in the value, the args, you return a new value but where you don't have to handle all of the other stuff and description overrides work, etc.

Issue #51, opened 2015-10-13
Default values for variables are ignored


dylanahsmith:
Example:

```ruby
require 'graphql'
require 'json'

QueryRoot = GraphQL::ObjectType.define do
  name "Query"

  field :double, types.String do
    argument :x, types.Int
    resolve ->(obj, args, ctx) {
      "#{args['x'].inspect} ** 2"
    }
  end
end

Schema = GraphQL::Schema.new(query: QueryRoot)
query_string = <<GRAPHQL
query Q($x: Int = 3) {
  double(x: $x)
}
GRAPHQL
variables = {}
result = Schema.execute(query_string, variables: variables, debug: true)
puts JSON.pretty_generate(result)
```

outputs

```
/Users/dylansmith/src/graphql-ruby/lib/graphql/base_type.rb:77:in `coerce_input!': Couldn't coerce nil to Int (GraphQL::ExecutionError)
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:28:in `reduce_value'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:8:in `block in initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:6:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:6:in `reduce'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:6:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:14:in `new'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:14:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:92:in `new'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:92:in `block in resolve_field'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/directive_chain.rb:30:in `call'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/directive_chain.rb:30:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:91:in `new'
[...]

PR #53, opened 2015-10-14
Fix variable coercion

### LINKED ISSUE #50: Regression: InputObjectType can't be specified with variables
commit fa7a2dba9dae5db516c0d276c80d7d4af4d1c9a5 introduced a regression where coerce gets called on ObjectInputType which doesn't implement the method

Example:

```ruby
require 'graphql'
require 'json'

AddInput = GraphQL::InputObjectType.define do
  name "AddInput"
  input_field :a, types.Int
  input_field :b, types.Int
end

QueryRoot = GraphQL::ObjectType.define do
  name "Query"

  field :add, types.String do
    argument :input, AddInput
    resolve ->(obj, args, ctx) {
      input = args['input']
      "#{input['a'].inspect} + #{input['b'].inspect}"
    }
  end
end

Schema = GraphQL::Schema.new(query: QueryRoot)
query_string = <<GRAPHQL
query Q($input: AddInput) {
  add(input: $input)
}
GRAPHQL
variables = { "input" => { "a" => 1.2, "b" => 2.3 } }
result = Schema.execute(query_string, variables: variables, debug: true)
puts JSON.pretty_generate(result)
```

outputs

```
/Users/dylansmith/src/graphql-ruby/lib/graphql/base_type.rb:74:in `coerce_input!': undefined method `coerce_input' for AddInput:GraphQL::InputObjectType (NoMethodError)
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:28:in `reduce_value'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:8:in `block in initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:6:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:6:in `reduce'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/arguments.rb:6:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:14:in `new'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:14:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:92:in `new'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:92:in `block in resolve_field'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/directive_chain.rb:30:in `call'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/directive_chain.rb:30:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:91:in `new'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:91:in `resolve_field'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:34:in `block in result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:33:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:33:in `reduce'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:33:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/operation_resolution.rb:17:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/base_execution.rb:16:in `execute'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/executor.rb:50:in `execute'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/executor.rb:27:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:43:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/schema.rb:41:in `execute'
	from tmp/example.rb:29:in `<main>'
```

previously it would output

```
{
  "data": {
    "add": "1.2 + 2.3"
  }
}
```

dylanahsmith:
~~Fixes #50 & #51~~

## Problem

~~Not all variable types could be coerced and the default value was being ignored.~~

## Solution

~~I moved the variable validation and coercion to the executor, so it is done before executing the operation.  This follows the same approach as graphql-js, which does the same thing when setting up the execution context.~~

I separated out the validation so that the error can reference the variable definition, although the API for scalars is the same, they can still return `nil` in their coerce_input proc to indicate that the value is invalid.

~~I also added coercion to methods to the input types that were missing them (e.g. input object type and list type) so they can handle variable coercion.~~

~~To handle default values, I had to decouple converting literals to values from GraphQL::Query::Arguments, so the code could easily be re-used.~~


rmosolgo:
Sorry, I just pushed some commits for this same purpose! https://github.com/rmosolgo/graphql-ruby/compare/4ac3142...4c1c25 (I also grabbed your specs to make sure it got the job done.)

I'm on the road this week with spotty internet, so sorry I wasn't communicative about these issues :( 

I really like a lot of the moves you made here, especially:

- moving that big, nasty `Arguments#reduce_value` into `Literal`. In my approach, I heaped up more debt there.
- `coerce_variables` is nice too, I shoved that into `reduce_value`. I also felt the need to separate the coercion of Ruby -> GraphQL values vs Query string -> GraphQL values, but I ended up just putting them in the same place :S 
- I like the `#valid_input?` API & adding `coerce_input` support for other types. It's such a win to make the types behave similarly!

I'd love to have those improvements in the codebase, do you want to copy them over? Otherwise I'll give it a shot!

rmosolgo:
I did some of that work here: https://github.com/rmosolgo/graphql-ruby/commit/8355b50b8d178ef36f5dca2731c3183a2bcdb402

dylanahsmith:
I've rebased this branch on top of the latest master and added additional regression tests that this branch was also fixing that weren't fixed in master (e.g. required input object fields could get given a `null` value through a variable).

spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
```diff
@@ -18,7 +18,7 @@
   let(:errors) { validator.validate(document) }
 
   it 'finds undefined arguments to fields and directives' do
-    assert_equal(3, errors.length)
+    assert_equal(4, errors.length)
```
rmosolgo: :+1: I see that I muddled input & result coercion from http://facebook.github.io/graphql/#sec-Boolean

lib/graphql/list_type.rb
```diff
@@ -17,8 +17,13 @@ def to_s
     "[#{of_type.to_s}]"
   end
 
-  def coerce_input(value)
-    inner_type = of_type
-    value.map { |item| inner_type.coerce_input!(item) }
+  def valid_non_null_input?(value)
+    value = [value] unless value.is_a?(Array)
```
rmosolgo: What's the case where a list type is asked to validate a single item? Should it always be given an array?

lib/graphql/query/executor.rb
```diff
@@ -13,7 +13,9 @@ def initialize(query)
       # @return [Hash] A GraphQL response, with either a "data" key or an "errors" key
       def result
         execute
-      rescue GraphQL::Query::OperationNameMissingError, GraphQL::Query::VariableMissingError => err
+      rescue GraphQL::ExecutionError => err
```
rmosolgo: :clap: :clap:  good call to add a proper exception hierarchy. Maybe `OperationNameMissingError` could be refactored down the road, too

lib/graphql/query/literal_input.rb
```diff
@@ -59,31 +28,61 @@ def self.from_arguments(ast_arguments, argument_defns, variables)
...
+            LiteralInput.coerce(type.of_type, value, variables)
+          end
+        end
 
-      def coerce(*args)
-        self.class.coerce(*args)
-      end
+        module ListLiteral
+          def self.coerce(value, type, variables)
+            if value.is_a?(Array)
```
rmosolgo: Same question I think, when is this _not_ already an array?

lib/graphql/static_validation/literal_validator.rb
```diff
@@ -6,10 +6,10 @@ def validate(ast_value, type)
     elsif type.kind.list? && ast_value.is_a?(Array)
       item_type = type.of_type
       ast_value.all? { |val| validate(val, item_type) }
-    elsif type.kind.scalar?
-      !type.coerce_input(ast_value).nil?
+    elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::AbstractNode) && !ast_value.is_a?(Array)
```
rmosolgo: What is the case here, where the type is scalar but the value is not a string, number or boolean?

rmosolgo:
Thank you! It'll be great to merge in these improvements and fixes.

lib/graphql/list_type.rb
dylanahsmith: http://facebook.github.io/graphql/#sec-Lists

> If the value passed as an input to a list type is not as list, it should be coerced as though the input was a list of size one, where the value passed is the only item in the list. This is to allow inputs that accept a “var args” to declare their input type as a list; if only one argument is passed (a common case), the client can just pass that value rather than constructing the list.

lib/graphql/static_validation/literal_validator.rb
dylanahsmith: Yes, scalar values should be a primitive value, so it should be safe to do this validation doing static validation rather than let scalars accidentally accept non-primitives.  E.g. an object or array could be passed into a string typed input field and the coercion right now will accept it and just call `.to_s` on it during coercion.

rmosolgo: yeah, seems like this should be the business of `GraphQL::StaticValidation::ArgumentLiteralsAreCompatible`. I think we should move that check here:

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/literal_validator.rb#L9

Feel free to move that check & add a test for that behavior _or_ just remove the check for now. We could easily fix that behavior separately.

lib/graphql/list_type.rb
rmosolgo: Wow, I wonder how I missed that! Kind of a weird behavior IMO.

I suspect that a real query would fail because of this check: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/literal_validator.rb#L6

If you want to add support for that case, could you add an "end-to-end" test for it to make sure it's fully supported? Otherwise, you could remove it from this PR and make it a to-do.

lib/graphql/static_validation/literal_validator.rb
dylanahsmith: > yeah, seems like this should be the business of GraphQL::StaticValidation::ArgumentLiteralsAreCompatible.

ArgumentLiteralsAreCompatible mostly just uses this method to do the actual validation, so this is a part of its business.

> I think we should move that check here:
> 
> https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/literal_validator.rb#L9

That is this line, which are commenting on.

lib/graphql/list_type.rb
dylanahsmith: Sure, I'll split it out into a separate PR, since this only fixes it for variables. Literals would still need to be fixed because of the check you linked to.

lib/graphql/static_validation/literal_validator.rb
rmosolgo: Lol, sorry! :tada:

rmosolgo: lineception

rmosolgo:
:+1: :+1: thanks!

PR #54, opened 2015-10-15
Fix parslet transformation for unnamed operations.

dylanahsmith:
Previously the operation wouldn't be transformed if the operation name is missing

```
irb> GraphQL.parse('query { name }')
=> {:document_parts=>[{:operation_type=>"query"@0, :variables=>[], :directives=>[], :selections=>[#<GraphQL::Language::Nodes::Field:0x007fac6157a278 @name="name", @alias=nil, @arguments=[], @directives=[], @selections=[], @line=1, @col=9>]}]}
```

returning a hash from GraphQL.parse rather than a GraphQL::Language::Nodes::Document object, resulting in queries to fail with an exception like `NoMethodError - undefined method `parts' for #<Hash:0x007f5fc63dc3e0>`

I fixed this by having it capture the optional name rather than making the capture optional, since the transform needs that `:name` key to exist in the hash.

### DIFF
--- a/lib/graphql/language/parser.rb
+++ b/lib/graphql/language/parser.rb
@@ -38,7 +38,7 @@ module GraphQL
       rule(:unnamed_selections) { selections.as(:unnamed_selections)}
       rule(:typed_operation_definition) {
         operation_type.as(:operation_type) >> space? >>
-        name.as(:name).maybe >> space? >>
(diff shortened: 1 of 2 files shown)


rmosolgo:
:moneybag: thanks!

PR #56, opened 2015-10-22
Fix delegators on GraphQL::Query::Arguments.

dylanahsmith:
## Problem

Calling `.each`, `.keys` or `.values` on an GraphQL::Query::Arguments object will result in a NoMethodError (e.g. undefined method `each' for nil:NilClass)

## Solution

def_delegators was just specifying the wrong instance variable name, so I corrected that.


rmosolgo:
:moneybag: thanks!

PR #57, opened 2015-11-05
Link to the individual guides right in the README

jjbohn:
Having them listed out there right at the top of the README I think
improves discoverability. Previously, it just link directly to the
introduction, which if you didn't read all the way to the bottom, you
might not realize that more guides exist.


rmosolgo:
Cool, thanks!

PR #61, opened 2015-11-09
Raise when a field's type is not a valid GraphQL::BaseType object

xuorig:
Currently when defining a field with a non-valid type, such an active record model class instead of the GraphQL type, this is raised:

`NoMethodError (undefined method unwrap for #<Class:0x007faf41021848>)`

This PR adds a check in `reduce_type` to see if the type is a subclass of `GraphQL::BaseType`. This removes the need for the nil check in initialize and will help debugging the schema when hitting that problem.

Let me know what you think!


rmosolgo:
:moneybag:  awesome, it will be great to have a much better message for this!

Issue #74, opened 2015-12-16
Parse error for empty input objects


kliuless:
```
mutation {
  createItem(input: 
    {
      clientMutationId: "112233",
      nested: {},  # <---- Empty input object can't be parsed
      title: "test mutation"
    }
  ) {
    returnField
  }
}
```
Error is: `#<GraphQL::ParseError: Extra input after last repetition at line 1 char 1.`
Please also verify that there can be spaces inside the braces.

rmosolgo:
:grimacing:  Sorry! 

Fixed by d887f5d. I want to make a refactoring on `InterfaceType` today then push out a new version!

kliuless:
Thanks for the quick reply. I finally grokked how the parse/transform code works, so I fixed the related transform code in PR #75 .

PR #75, opened 2015-12-17
fix(Transform) handle empty input objects

kliuless:
Follow-up to #74


rmosolgo:
Awesome! Sorry I missed this on the first pass!

PR #76, opened 2015-12-20
Query cache

rmosolgo:
This enables you to store queries on the server, then invoke them by name from the client. 

I want to try this out on Monday, then merge it in.


davidnorth:
I would like to have the performance boost from this but it seems to go against the idea of GraphQL to define queries on the server. Would it be realistic to cache the query some other way, maybe with a hash of the query string as the key? Then the client still has the control over those queries. You don't want to have to keep updating them on the server.

rmosolgo:
It is a bit odd, but I heard that this is what FB does for non-Relay uses of GraphQL. I heard that part of the build steps for their native apps is registering the required queries with the server. 

As I'm using GraphQL in my own app (Rails, no Relay), I can see that a similar workflow would work for me. Once I'm done development, I don't _need_ a dynamic query (the UI isn't going to change until the next deploy, neither is the query). 

So I'm interested to see if I can hook into Sprockets. Add `.graphql` files to the asset pipeline which, when you `//= require` them, cache themselves on the server and make themselves into a client-side helper. Then, from the React component, I'd just invoke that query by name with some variables and use the response.

charlieschwabacher:
I ended up storing parsed documents in Rails.cache towards a similar end.  This ends up providing a very noticeable speed up after the initial parse.

```
class CachedQuery < GraphQL::Query

  # We override initialize on GraphQL::Query in order to cache the parsed
  # document in the Rails cache
  def initialize(schema, query_string, context: nil, variables: {}, debug: false, validate: true, operation_name: nil)
    @schema = schema
    @debug = debug
    @context = Context.new(query: self, values: context)
    @validate = validate
    @operation_name = operation_name
    @fragments = {}
    @operations = {}
    @provided_variables = variables

    # replaces @document = GraphQL.parse(query_string)
    @document = Rails.cache.fetch(query_string) do
      GraphQL.parse(query_string)
    end

    @document.parts.each do |part|
      if part.is_a?(GraphQL::Language::Nodes::FragmentDefinition)
        @fragments[part.name] = part
      elsif part.is_a?(GraphQL::Language::Nodes::OperationDefinition)
        @operations[part.name] = part
      end
    end
  end

end
```

Rails cache in this example takes a digest of query_strings it received, and then stores the parsed document in memcached.

I'm using relay on the front end, so while the queries that get generated by the client are consistent - I don't have an easy way of generating them on the server.

rmosolgo:
Woahh that's very cool! You get some advantages of caching even though the queries are dynamic. 

I wonder how you could expose that kind of option in a way that doesn't depend on Rails specifically. 

Ha, another approach would be to make that a "private concern" of the parser -- the rest of the system doesn't have to know how it goes from query string to a document object. For example, you could just monkey-patch `GraphQL.parse`: 

```ruby 
module GraphQL 
  class << self 
    alias :old_parse :parse 

    def parse(query_string) 
      Rails.cache.fetch(query_string) do 
        old_parse(query_string) 
      end
    end 
  end 
end 
```

One of these days (er, weeks) I'd like to get going with libgraphqlparser so that parsing isn't so expensive in the first place!

rmosolgo:
Handled much better by #152

PR #77, opened 2015-12-21
Add rubinius support

sagmor:
[Rubinius](http://rubinius.com/) currently doesn't support refinements so made a little change to use a constant Proc instead of a Monkey Patch/Refinement


rmosolgo:
Awesome, thanks! :tada: 

Nice that it didn't take too much work to get it Rubinius-ready, thanks for adding it to CI, too.

PR #79, opened 2015-12-28
fix transform failure on nested lists in field arguments

charlieschwabacher:
The Parslet transform fails on nested lists in field arguments, leading to line 27 of lib/graphql.rb raising:

```
raise("Parse failed! Sorry, somehow we failed to turn this string into a document. Please report this bug!")
```

I used this spec while figuring out the fix, it's a good example of the issue: https://gist.github.com/charlieschwabacher/d97d6631523aad272b84, the tests at lines 43 and 48 will fail in master.  

I didn't include it in the pull - I think the addition to transform_spec.rb should be enough to cover this, but let me know if there is anywhere you think it would make sense to add a spec similar to that gist.

### DIFF
--- a/lib/graphql/language/transform.rb
+++ b/lib/graphql/language/transform.rb
@@ -71,7 +71,7 @@ module GraphQL
       rule(alias_name: simple(:a)) { a }
       optional_sequence(:optional_field_arguments)
       rule(field_argument_name: simple(:n), field_argument_value: simple(:v)) { CREATE_NODE[:Argument, name: n.to_s, value: v, position_source: n]}
-      rule(field_argument_name: simple(:n), field_argument_value: sequence(:v)) { CREATE_NODE[:Argument, name: n.to_s, value: v, position_source: n]}
+      rule(field_argument_name: simple(:n), field_argument_value: subtree(:v)) { CREATE_NODE[:Argument, name: n.to_s, value: v, position_source: n]}
       optional_sequence(:optional_selections)
       optional_sequence(:optional_directives)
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
Yikes, sorry about that! Is there a way to fix this without using `subtree`? I want to heed those warnings: 

> [So you don’t want to listen and really want that big gun with the foot aiming addon. You’ll be needing `subtree(symbol)`.](http://kschiess.github.io/parslet/transform.html)

> [Caution with the latter.](http://kschiess.github.io/parslet/get-started.html)

I'm not exactly sure _why_ he said that, I guess it's just that you get some unspecified behavior (so some other bugs might sneak through?

charlieschwabacher:
Hey - that line from the docs put me off at first too.  I could be overlooking something, but I don't think there is a way around using `subtree` here - if I understand correctly, `sequence` can never match a tree like `{array: [{array: [1]}]}` - it will start at the leaf, and transform to `{array: [[1]]}`, and then fail to match again.  With `subtree`, it will start at the leaf and make the same initial match, but then match again producing `[[1]]`.

One issue w/ this fix is that it allows something like `query Q { a(a: [[3], 1]) }` to pass the transform, when (i think) that is not valid graphql.  Do you think it would make sense to try to catch this w/ one of the static validation rules?

Surprisingly - it looks like graphql-js has the same issue:
```
const GraphQL = require('graphql');

const schema = new GraphQL.GraphQLSchema({
  query: new GraphQL.GraphQLObjectType({
    name: 'Query',
    fields: () => ({
      a: {
        type: GraphQL.GraphQLInt,
        args: {
          a: {
            type: (
              new GraphQL.GraphQLNonNull(
                new GraphQL.GraphQLList(
                  new GraphQL.GraphQLList(
                    GraphQL.GraphQLInt
                  )
                )
              )
            )
          },
        },
        resolve: (source, args, info) => args.a[0][0]
      }
    })
  })
});

const query = 'query Q { a(a: [[2], 1]) }';

GraphQL.graphql(schema, query).then((result) => console.log(result));
// logs: { data: { a: 2 } }
```

rmosolgo:
Sure 'nuf, there's the anti-example in the docs:

image

Thanks for getting to the bottom of this! :confetti_ball:

PR #80, opened 2015-12-29
Allow Automatic camelCase -> snake_case field properties option

lparry:
Adds an option to automatically try and use an underscored version of camelcase field names, saving a bit of boilerplate property configuration.

turned on by setting `GraphQL::Config. try_underscored_property_names = true`, which then allows fields like `appearsIn` to be defined without having to manually specify the `property: :appears_in` part. Without this option set, behaviour is unchanged.

I'd like to write some tests, but from a cursory dig I couldn't find tests for any of the field resolution stuff, and I'm not familiar enough with how the codebase ties together to know where to start writing a unit test for this. Happy to take on any pointers :D 

I've swapped this code into the `vendor/bundle` directory of https://github.com/rmosolgo/graphql-ruby-demo, added an intializer to turn on the option, and deleted a bunch of config on the types definitions and it seems to work as expected.

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -35,6 +35,9 @@ end
 
 # Order matters for these:
 
+require 'graphql/config'
+require 'graphql/util'
+
 require 'graphql/definition_helpers'
 require 'graphql/base_type'
 require 'graphql/object_type'
--- a/lib/graphql/config.rb
+++ b/lib/graphql/config.rb
@@ -1,0 +1,6 @@
+class GraphQL::Config
+  @try_underscored_property_names = false
+  class << self
+    attr_accessor :try_underscored_property_names
+  end
+end
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -64,7 +64,14 @@ module GraphQL
 
             if value == GraphQL::Query::DEFAULT_RESOLVE
               begin
-                value = target.public_send(ast_node.name)
+                if target.respond_to?(ast_node.name) || !GraphQL::Config.try_underscored_property_names
+                  value = target.public_send(ast_node.name)
+                elsif target.respond_to?(underscored_name = GraphQL::Util.underscore_string(ast_node.name))
+                  value = target.public_send(underscored_name)
+                else
+                  # Fallback just in case #respond_to? was lying
+                  value = target.public_send(ast_node.name)
+                end
               rescue NoMethodError => err
                 raise("Couldn't resolve field '#{ast_node.name}' to #{parent_object.class} '#{parent_object}' (resulted in #{err})")
               end
--- a/lib/graphql/util.rb
+++ b/lib/graphql/util.rb
@@ -1,0 +1,10 @@
+module GraphQL::Util
+  module_function
+
+  # Based on ActiveSupport's String#underscore but less aggressive and monkey-patchy
+  def underscore_string(string)
+    string.gsub(/([A-Z]+)([A-Z][a-z])/,'\1_\2').
+      gsub(/([a-z\d])([A-Z])/,'\1_\2').
+      downcase
+  end
+end


rmosolgo:
Nice, I like the idea of saving all that boilerplate! Going between Ruby & JS, it's hard to know when to start camelizing. 

It seems a bit odd to make all those runtime `respond_to?` checks. The object is unlikely to gain and lose methods while running, so it would be nice to determine the property name _ahead_ of time, then simply `send` it to resolve the field. 

I think you might be able to do that by adding to the field definition logic. Here's where the `property:` option is handled: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/definition_helpers/defined_by_config.rb#L62

Maybe you could add a line there to check if the field should be underscored, for example:

```ruby 
# If the name should be underscored, underscore it as the property:
if property.nil? && GraphQL::Config.underscore_field_names && camelcase_string?(name.to_s)
  property = GraphQL::Util.underscore_string(name.to_s)
end 
# ... 
```

(In that example, we'd also have do define the `camelcase_string?` helper to check whether the name should be underscored). 

That way, the field will get a custom resolve proc for the underscored and we don't have to add any code for query execution. Also, if you want that default behavior but you want to override it, you can always pass a `property:` value which will be prioritized.

What do you think of that implementation? Is it missing any features of the current one? 


## Testing 

I think you could add a test to `field_spec.rb`. If it was me, I would set up a test with the config , then unset the config during teardown. 

Inside the test, I would make a type, then test that its fields behave as expected. Here's an example:

```ruby 
# inside `describe GraphQL::Field do ...`
describe "underscored field names" do 
  # The config is true for this test:
  before do 
    GraphQL::Config.underscored_property_names = true 
  end 
  # But false for the rest of the test suite:
  after do 
    GraphQL::Config.underscored_property_names = false 
  end 

  let(:type) { 
    GraphQL::ObjectType.define do 
       field :camelCase, types.String
       field :already_underscore, types.String
       field: :plain, types.String 
     end
   }

   let(:target) {
     # This is our test double, make sure we get these strings by resolving the fields:
     OpenStruct.new(
      camel_case: "This field was transformed",
      already_underscored: "This field was not transformed",
      plain: "This field was left alone" 
     )
   }

   describe "camel-cased field names" do 
     it "uses underscore property names"
     # This field should return the string above ^^
   end 
   describe "non-camel-cased field names" do 
     it "uses the field name as a property name"
     # these fields should return GraphQL::Query::DEFAULT_RESOLVE
   end   
end
```

I didn't implement the specs themselves. But check out this part of the testing guide, it will show you how to fetch a specific field out of the schema, then test its resolve proc: http://www.rubydoc.info/github/rmosolgo/graphql-ruby/file/guides/testing.md#Testing_schema_elements_in_isolation

lparry:
Thanks for the detailed response! I'm just back from holidays now and have a bit of catching up to do with general life stuff, but I'll definitely come back to this and give it another shot. the camel vs snake thing

aviscasillas:
Hi guys! I'm actually looking for a solution for this camel vs snake issue. How is this going?

rmosolgo:
Just merged a change to how `.define { ... }` works to master (not on RubyGems yet) that might help for a case like this: https://github.com/rmosolgo/graphql-ruby/pull/116

You could make a definition for camelized field which would modify the input, then delegate to the "normal" field definition. For example, I _think_ this would work: 

```ruby 
module CamelizedField 
  def self.call(target, field_name, *args, &block)
    # modify the incoming name:
    camelized_field_name  = camelize(field_name)

    # use the original name as the `property:` keyword argument: 
    if args.last.is_a?(Hash)
      keyword_args = args.last 
    else 
      keyword_args = {} 
      args << keyword_args 
    end 
    keyword_args[:property] = field_name 

    # use the new name to create a GraphQL::Field:
    GraphQL::Define::AssignObjectField.call(target, camelized_field_name, *args, &block)
  end 
end

# Then hook up that callable to `.define`: 
GraphQL::ObjectType.accepts_definitions(camelized_field: CamelizedField)
GraphQL::InterfaceType.accepts_definitions(camelized_field: CamelizedField)

# Then use it: 
  camelized_field :my_field # creates field "myField" which resolves by sending :my_field
end 
```

lparry:
Sorry, I got bogged down in real life work and never got around to coming back to this :(

aviscasillas:
Thx @rmosolgo I will take a look.

PR #83, opened 2016-01-14
Fix parsing and transforming of string escapes.

dylanahsmith:
cc @eapache

## Problem

We noticed that newline characters in graphql strings were getting replaced with the character `n`, e.g.

```ruby
require 'graphql'

QueryRoot = GraphQL::ObjectType.define do
  name "Query"

  field :hello, types.String do
    argument :name, types.String
    resolve -> (obj, args, ctx) {
      puts args[:name].inspect
    }
  end
end

Schema = GraphQL::Schema.new(query: QueryRoot)
Schema.execute("{ hello(name: \"new\\nline\") }")
```

outputs `"newnline"` rather than `"new\nline"`

## Solution

The transform on the escaped string was being done with a regex which matched the escaped sequence `\n` but then just replaced it with it with the second character `n` rather than unescaping it.  I instead used a hash as a second argument to `gsub!` to replace `\n` with a newline, along with handling the other control character escapes properly.

During testing I also came across backslash not even getting matched by the parser, since `match('["\/bfnrt]')`was interpreting the backslash as a regexp escape for the following character `/`.  Instead, I needed to double escape the `\`, once because single quoted strings treat `\\` as an escape for `\`, and again to escape it for the regular expression.

### DIFF
--- a/lib/graphql/language/parser.rb
+++ b/lib/graphql/language/parser.rb
@@ -99,7 +99,7 @@ module GraphQL
       rule(:value_int) { (value_sign? >> match('\d').repeat(1)).as(:int) }
       rule(:value_string) { str('"') >> value_string_char.repeat.maybe.as(:optional_string_content).as(:string) >> str('"')}
       rule(:value_string_char) { value_string_escaped_char | value_string_escaped_unicode | value_string_source_char}
-      rule(:value_string_escaped_char) { str("\\") >> match('["\/bfnrt]') }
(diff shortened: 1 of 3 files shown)


rmosolgo:
:moneybag: shucks, looking back at it, the tests on that were pretty weak! :S 

Thanks!

rmosolgo:
Shipped in 0.10.8!

PR #84, opened 2016-01-14
Correctly handle `interfaces` being re-assigned

eapache:
cc @dylanahsmith 

Remove stale `possible_types` entries and add to only the ones that are new.

We ran into this due to the way we were structuring some of our code, our schema ended up with duplicate possible types.

### DIFF
--- a/lib/graphql/object_type.rb
+++ b/lib/graphql/object_type.rb
@@ -31,10 +31,11 @@ class GraphQL::ObjectType < GraphQL::BaseType
 
   #   Shovel this type into each interface's `possible_types` array.
   #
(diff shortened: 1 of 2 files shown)


dylanahsmith:
LGTM

rmosolgo:
Nice! In my dreams, this would somehow be tracked at Schema-level, since technically you could have two schemas with the same interfaces but different object types, then you'd want the interface's possible_types to also be a function of the schema 

BUT this is way better than it was, thanks!

rmosolgo:
Shipped in 0.10.9!

eapache:
Changelog says 0.10.8 again?

rmosolgo:
oops, fixed!

PR #88, opened 2016-01-22
feat(Field) expose property and respond to changes

### LINKED ISSUE #87: Making field property accessible after definition
When defining a field, you can specify a property name, which gets called on the backing object during resolve:
```ruby
field :createdBy, types.String, property: :created_by
```
However, it seems like the property isn't accessible on the `GraphQL::Field` object; it's wrapped inside the resolver lambda in `GraphQL::DefinitionHelpers::DefinedByConfig::DefinitionConfig`:
```ruby
property && field.resolve = -> (t,a,c) { t.public_send(property)}
```
Is there interest in storing this to an instance variable and exposing a getter?

My use case is to access the mappings in a mutation which needs to do a reverse mapping, to set attributes on an ActiveRecord model. E.g. the mutation receives a `createdBy` argument and has to call `record.created_by = input_value`. If there's a better way to do this, let me know.

kliuless:
Resolves #87

### DIFF
--- a/spec/graphql/field_spec.rb
+++ b/spec/graphql/field_spec.rb
@@ -7,4 +7,48 @@ describe GraphQL::Field do
     end
     assert_equal(DairyProductUnion, field.type)
   end
+
+
+  describe '.property ' do
+    let(:field) do
+      GraphQL::Field.define do
+        # satisfies 'can define by config' below
+        property :internal_prop
+      end
+    end
+
+    it 'can define by config' do
+      assert_equal(field.property, :internal_prop)
+    end
+
+    it 'has nil property if not defined' do
+      no_prop_field = GraphQL::Field.define { }
+      assert_equal(no_prop_field.property, nil)
+    end
+
+    describe 'default resolver' do
+      def acts_like_default_resolver(field, old_prop, new_prop)
+        object = OpenStruct.new(old_prop => 'old value', new_prop => 'new value')
+
+        old_result = field.resolve(object, nil, nil)
+        field.property = new_prop
+        new_result = field.resolve(object, nil, nil)
+        field.property = nil
+        unset_result = field.resolve(object, nil, nil)
+
+        assert_equal(old_result, 'old value')
+        assert_equal(new_result, 'new value')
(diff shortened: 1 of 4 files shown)


spec/graphql/object_type_spec.rb
```diff
@@ -24,5 +24,12 @@
       assert_equal(GraphQL::TypeKinds::NON_NULL, field.type.kind)
       assert_equal(GraphQL::TypeKinds::SCALAR, field.type.of_type.kind)
     end
+
+    it 'exposes defined field property' do
```
rmosolgo: I think we don't need this, this behavior is also tested in `field_spec`, right?

spec/graphql/field_spec.rb
```diff
@@ -7,4 +7,48 @@
...
+    it 'can define by config' do
+      assert_equal(field.property, :internal_prop)
+    end
+
+    it 'has nil property if not defined' do
+      no_prop_field = GraphQL::Field.define { }
+      assert_equal(no_prop_field.property, nil)
+    end
+
+    describe 'default resolver' do
```
rmosolgo: This test is making sure that if you change `property` after initialization, `resolve` still works?

rmosolgo:
:+1: big win!

spec/graphql/object_type_spec.rb
kliuless: I added that test to ensure that `DefinitionConfig#field` is assigning the property. `field_spec` doesn't test that flow.
I was going back & forth in my mind on whether to add this test, and then I had the thought that `DefinitionConfig#field` could either be moved into `Field` (possibly as a class method), or it can call a (not-yet-existing) method on `Field` to initialize the `Field` attributes. That way the `Field`-specific logic can be all in one place (same goes for tests).

spec/graphql/field_spec.rb
kliuless: Yes, that's the idea.

spec/graphql/object_type_spec.rb
rmosolgo: Ah, that sounds good!

rmosolgo:
Thanks! :tada:

PR #93, opened 2016-02-09
feat(Query): Rescue GraphQL::ExecutionError in field resolution.

dylanahsmith:
## Problem

If an GraphQL::ExecutionError object is returned from a field resolve proc, then it is reported as an error on that field.  However, it would be more natural to raise an exception of a specific class to report an error, which would require the rescue middleware.

Unfortunately, the rescue middleware currently doesn't catch all the errors during field resolution, since it misses type resolution and value coercion.

## Solution

Allow a GraphQL::ExecutionError to be intentionally raised from a resolve, coercion or type resolution proc and have that error affect only that field rather than causing the whole query to fail.

### DIFF
--- a/guides/defining_your_schema.md
+++ b/guides/defining_your_schema.md
@@ -92,6 +92,7 @@ result = MySchema.execute(query_string)
 # }
 ```
 
+When explicitly raising the exception, you can raise a GraphQL::ExecutionError with a message to add to the response without specifying an error handler.
 
 ## Middleware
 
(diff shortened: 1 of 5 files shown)


rmosolgo:
:+1: Nice to get the whole thing wrapped by error handling!

How about adding a note about the new behavior here: https://github.com/rmosolgo/graphql-ruby/blob/master/guides/defining_your_schema.md#handling-errors ?

I think you could describe this behavior first, then say "or, you can rescue from specific errors by ... "

dylanahsmith:
Added a note in the guide:

> Added the following note

> When explicitly raising the exception, you can raise a GraphQL::ExecutionError with a message to add to the response without specifying an error handler.

rmosolgo:
:moneybag:  Thanks!

PR #94, opened 2016-02-12
Error when non-null fields resolve to nil

eapache:
@rmosolgo @dylanahsmith I spent the last day and a bit down a rabbit-hole of nulls and type coercion. This fixes the immediate issue but raises several more interesting questions:

- We do not currently provide a way for types to specify a coercion for `nil`
  values. According to the GraphQL spec on non-null types [1], you *are*
  supposed to run coercion for `nil`s and only fail if the *coercion* produces
  `nil` (so e.g. the boolean scalar type is allowed to coerce `nil` to false if
  it wants). I took a quick try to do this but ran aground quickly.

- We do not distinguish at the scalar coercion level between nullable and
  non-null-wrapped types. It may make sense, for example, for a regular boolean
  to let `nil` stay as `nil` but to coerce `nil` to `false` when wrapped in a
  non-null; this requires knowledge in either the `coerce_result` method or the
  `ScalarResolution` strategy about whether the wrapping type is non-null.

- It is not clear to me from the GraphQL spec whether fields that raise errors
  (like the one added in this PR) should be present in the result with a null
  value or should be absent from the result entirely. Both feel odd, especially
  when the field itself is non-null. (edit: https://facebook.github.io/graphql/#sec-Nullability is clear, I just missed it the first time through).

- It is not clear to me from the GraphQL spec how nil interfaces should be
  coerced, since at least theoretically there could be multiple implementing
  object types, each with a distinct method of coercing `nil` to something
  non-null, but the executor has no way to determine which coercion to run.

[1] https://facebook.github.io/graphql/#sec-Non-Null

### DIFF
--- a/lib/graphql/null_non_null_error.rb
+++ b/lib/graphql/null_non_null_error.rb
@@ -1,0 +1,22 @@
+module GraphQL
+  # Raised automatically when a field's resolve function returns `nil`
+  # for a non-null field.
+  class NullNonNullError < RuntimeError
+    def initialize(field_name, value)
+      @field_name = field_name
+      @value = value
+    end
+
+    # @return [Hash] An entry for the response's "errors" key
+    def to_h
+      {
+        "message" => "Cannot return null for non-nullable field #{@field_name}"
+      }
+    end
+
+    # @return [Boolean] Whether the null in question was caused by another error
+    def parent_error?
+      @value.is_a?(GraphQL::ExecutionError)
+    end
+  end
+end
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -16,14 +16,14 @@ module GraphQL
 
         def result
           result_name = ast_node.alias || ast_node.name
-          result_value = begin
-            get_finished_value(get_raw_value)
+          raw_value = begin
+            get_raw_value
           rescue GraphQL::ExecutionError => err
             err.ast_node = ast_node
             query.context.errors << err
-            nil
+            err
           end
-          { result_name => result_value  }
+          { result_name => get_finished_value(raw_value) }
         end
 
         private
@@ -31,8 +31,6 @@ module GraphQL
         # After getting the value from the field's resolve method,
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
-          raise raw_value if raw_value.instance_of?(GraphQL::ExecutionError)
-
           strategy_class = GraphQL::Query::SerialExecution::ValueResolution.get_strategy_for_kind(field.type.kind)
           result_strategy = strategy_class.new(raw_value, field.type, target, parent_type, ast_node, query, execution_strategy)
           result_strategy.result
@@ -48,7 +46,9 @@ module GraphQL
             steps: steps,
             arguments: [parent_type, target, field, arguments, query.context]
           )
-          chain.call
+          value = chain.call
+          raise value if value.instance_of?(GraphQL::ExecutionError)
+          value
         end
 
 
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -32,14 +32,9 @@ module GraphQL
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
           raise raw_value if raw_value.instance_of?(GraphQL::ExecutionError)
-          if raw_value.nil?
-            raise GraphQL::ExecutionError, "Type mismatch resolving '#{field.name}'" if field.type.kind.non_null?
-            return nil
-          end
 
-          resolved_type = field.type.resolve_type(raw_value)
-          strategy_class = GraphQL::Query::SerialExecution::ValueResolution.get_strategy_for_kind(resolved_type.kind)
-          result_strategy = strategy_class.new(raw_value, resolved_type, target, parent_type, ast_node, query, execution_strategy)
+          strategy_class = GraphQL::Query::SerialExecution::ValueResolution.get_strategy_for_kind(field.type.kind)
+          result_strategy = strategy_class.new(raw_value, field.type, target, parent_type, ast_node, query, execution_strategy)
           result_strategy.result
         end
 
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -32,7 +32,10 @@ module GraphQL
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
           raise raw_value if raw_value.instance_of?(GraphQL::ExecutionError)
-          return nil if raw_value.nil?
+          if raw_value.nil?
+            raise GraphQL::ExecutionError, "Type mismatch resolving '#{field.name}'" if field.type.kind.non_null?
+            return nil
+          end
 
           resolved_type = field.type.resolve_type(raw_value)
           strategy_class = GraphQL::Query::SerialExecution::ValueResolution.get_strategy_for_kind(resolved_type.kind)
--- a/spec/graphql/query/executor_spec.rb
+++ b/spec/graphql/query/executor_spec.rb
@@ -132,6 +132,22 @@ describe GraphQL::Query::Executor do
       end
     end
 
+    describe 'if nil is given for a non-null field' do
+      let(:query_string) {%| query noMilk { cantBeNullButIs }|}
+      it 'turns into error messages' do
+        expected = {
+          "data" => { "cantBeNullButIs" => nil },
+          "errors" => [
+            {
+              "message" => "Type mismatch resolving 'cantBeNullButIs'",
+              "locations" => [ { "line" => 1, "column" => 17 } ]
+            }
+          ]
+        }
+        assert_equal(expected, result)
+      end
+    end
+
     describe "if the schema has a rescue handler" do
       before do
         schema.rescue_from(RuntimeError) { "Error was handled!" }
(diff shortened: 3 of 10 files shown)


lib/graphql/query/serial_execution/field_resolution.rb
```diff
@@ -32,7 +32,10 @@ def result
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
           raise raw_value if raw_value.instance_of?(GraphQL::ExecutionError)
-          return nil if raw_value.nil?
+          if raw_value.nil?
+            raise GraphQL::ExecutionError, "Type mismatch resolving '#{field.name}'" if field.type.kind.non_null?
```
dylanahsmith: I think this should be done in GraphQL::NonNullType#coerce_result if `of_type.coerce_result(value)` returns `nil`.  Doing it here means we are still missing the check on elements of a list.

dylanahsmith:
It would be nice if `coerce_result` was passed a boolean to indicate whether it is a nullable or non-null type.  However, doing that in a backwards compatible way would require an arity check on the method, which is kind of gross.

An alternative would be to allow a `coerce_non_null_result` proc to be specified for scalar types, which would default to using the `coerce_result` proc if it isn't specified.

lib/graphql/query/serial_execution/field_resolution.rb
eapache: AFAICT `NonNullType#coerce_result` is dead code?

dylanahsmith: Right, it looks like GraphQL::Query::BaseExecution::ValueResolution::NonNullResolution doesn't call it at all.  So we could either do this check there, or have that class delegate to NonNullType#coerce_result

eapache: Right, I knew there was a reason I did it this way to begin with:
- Complex types (objects, lists, etc) don't define `coerce_result` at all, so non-null's implementation is dead *and* badly broken. I'll remove it.
- If we do the check in `NonNullResolution`, then we have to be able to call `field.type.resolve_type(nil)` for all types in order to *get* the resolver class, but this method isn't well-defined for interfaces and it's not obvious that it should be (see the fourth point in my original PR post).

dylanahsmith: We should probably add a couple type kind strategies in ValueResolution for interface and union types, and let them resolve the type, since I think that is unnecessary in the other cases.

I think we should move the `nil` checking into BaseResolution#result, which should just return `nil` if the value is `nil`, otherwise, it should call `coerce_result`.  The classes that derive from BaseResolution should rename their `result` method to `coerce_result`.  We can then override `result` in NonNullResolution to raise if the value is `nil`.

I'm not exactly sure how a scalar type should specify what should be done for a `nil` on non-null type, but NonNullResolution#result could call that `if value.nil? && field_type.of_type == GraphQL::TypeKinds::SCALAR`.

eapache: OK yes, I'd already kind of wandered down the "write an `InterfaceResolution` class" path, and splitting `BaseResolution#result` in two was the extra piece I needed.

I think I'm OK with this PR as-is now; permitting coercion of nil for scalars is worthy of its own PR.

lib/graphql/query/base_execution/value_resolution.rb
```diff
@@ -20,6 +20,11 @@ def initialize(value, field_type, target, parent_type, ast_field, query, executi
           end
 
           def result
+            return nil if value.nil?
```
dylanahsmith: you can just use the ternary operator instead of the early return

lib/graphql/query/base_execution/value_resolution.rb
```diff
@@ -38,39 +43,40 @@ def result
...
-          # Get the string name for this enum value
-          def result
-            field_type.coerce_result(value)
-          end
-        end
-
         class NonNullResolution < BaseResolution
           # Get the "wrapped" type and resolve the value according to that type
           def result
+            raise GraphQL::ExecutionError, "Type mismatch resolving '#{ast_field.name}'" if value.nil?
```
dylanahsmith: `GraphQL::ExecutionError` has already been used extensively as a way of reporting exceptions to the client, but this is an internal server error which we would want to report to server developers.

lib/graphql/query/base_execution/value_resolution.rb
eapache: I avoid ternaries when the condition ends in `?` because I find the double-`?` hard to read

lib/graphql/query/base_execution/value_resolution.rb
eapache: At the moment `ExecutionError`s are the only ones capable of putting useful information (e.g. a `location` key) into the error data, so I think the exception hierarchy needs a larger cleanup. I'll address that in a separate PR unless @rmosolgo has a different opinion

dylanahsmith: Providing useful information to the client isn't as useful as providing useful information to the server application though.  The location key is much more useful when it actually helps the client debug an issue, but this type of issue they would at most only be able to workaround.

The GraphQL spec refers to these errors as a field error, so perhaps a GraphQL::FieldError class could be used as the base class to handle putting location information and ExecutionError could be could derive from it.

spec/graphql/query/executor_spec.rb
```diff
@@ -132,6 +132,22 @@
       end
     end
 
+    describe 'if nil is given for a non-null field' do
+      let(:query_string) {%| query noMilk { cantBeNullButIs }|}
+      it 'turns into error messages' do
+        expected = {
+          "data" => { "cantBeNullButIs" => nil },
```
dylanahsmith: This field shouldn't be present in the response according to the graphql spec: http://facebook.github.io/graphql/#sec-Nullability

spec/graphql/query/executor_spec.rb
```diff
@@ -132,6 +132,22 @@
...
+            {
+              "message" => "Type mismatch resolving 'cantBeNullButIs'",
+              "locations" => [ { "line" => 1, "column" => 17 } ]
+            }
+          ]
+        }
+        assert_equal(expected, result)
+      end
+    end
+
```
dylanahsmith: Could you add another test for a field nested within an object to make sure only one error appears in the errors list?

lib/graphql/query/base_execution/value_resolution.rb
dylanahsmith: Looks like graphql-js doesn't report location information for this error.  The response I got for an error on a field named `name` on an object type `User` was `{"data":null,"errors":[{"message":"Cannot return null for non-nullable field User.name."}]}`

spec/graphql/query/executor_spec.rb
rmosolgo: Huh, 

> ...then the result of evaluating the entire field set that contains this field is now null

So, all of its siblings should be be ignored, it would be `"data" => nil` ?

lib/graphql/query/base_execution/value_resolution.rb
rmosolgo: +1 for `Cannot return null for non-nullable field`

eapache:
OK, I believe I've addressed all feedback. Re-review please?

lib/graphql/query/serial_execution/field_resolution.rb
```diff
@@ -16,14 +16,14 @@ def initialize(ast_node, parent_type, target, query, execution_strategy)
...
+          raw_value = begin
+            get_raw_value
           rescue GraphQL::ExecutionError => err
             err.ast_node = ast_node
             query.context.errors << err
-            nil
+            err
           end
-          { result_name => result_value  }
+          { result_name => get_finished_value(raw_value) }
```
rmosolgo: Do we want `get_finished_value` to be within the `begin ... rescue` block? 

It was added intentionally here: https://github.com/rmosolgo/graphql-ruby/pull/93/files#diff-a66f6e348162887250fe0e5939b3a11bR20

eapache: I thought about that, but there's nothing in `get_finished_value` that will raise an `ExecutionError`, and no user-specified code either, so I don't think we need to wrap it.

dylanahsmith: type resolution and type coercion can be specified for user defined scalars and interfaces.

eapache: Good point.

Thinking through this more, I'm not even sure what we'd do if we rescued an error from one of those sources. We can't return `nil` because then we're breaking spec if the field is non-null, unless we tried to recursively call `get_finished_value` which would make it tricky to avoid infinite loops?

I'm also not sure when one of those methods would want to intentionally raise an `ExecutionError`... most such errors I expect would be of the wholly-unintentional type raised by ruby directly.

lib/graphql/null_non_null_error.rb
```diff
@@ -0,0 +1,22 @@
+module GraphQL
+  # Raised automatically when a field's resolve function returns `nil`
+  # for a non-null field.
+  class NullNonNullError < RuntimeError
```
dylanahsmith: I don't find NullNonNullError to be a very readable error message.  Perhaps `InvalidNullError` would be less confusing.

lib/graphql/query/serial_execution/field_resolution.rb
dylanahsmith: Yes, I guess `GraphQL::ExecutionError` wouldn't be appropriate for errors in type resolution or result coercion, since they would be more like the `nil` on a non-null field error that this PR addresses.

dylanahsmith:
LGTM

rmosolgo:
:+1: Thanks for taking care of this! One more benefit from a typed API.

Note to self: all types respond to `resolve_type`, but they might not have to anymore (https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L39-L43)

kliuless:
:+1:  This is excellent. Uncovered a couple bugs for me.

PR #95, opened 2016-02-14
Extract flatten and merge into own method

xuorig:
Extracted the flattening and merging of selections into it's own method, used by both `result` and `flatten_fragment` methods. Also changed a few style things, let me know if you disagree with them :)

### DIFF
--- a/lib/graphql/query/serial_execution/selection_resolution.rb
+++ b/lib/graphql/query/serial_execution/selection_resolution.rb
@@ -13,21 +13,10 @@ module GraphQL
         end
 
         def result
-          # In a first pass, we flatten the selection by merging in fields from
-          # any fragments - this prevents us from resolving the same fields
-          # more than one time in cases where fragments repeat fields.
-          # Then, In a second pass, we resolve the flattened set of fields
-          selections
-            .reduce({}){|memo, ast_node|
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: Nice find! I didn't realize those were the same. Thanks!

lib/graphql/query/serial_execution/selection_resolution.rb
```diff
@@ -87,26 +66,29 @@ def fragment_type_can_apply?(ast_fragment)
...
-              name: field2.name,
-              alias: field2.alias,
-              arguments: field2.arguments,
-              directives: field2.directives,
-              selections: field1.selections + field2.selections
-            )
-          else
-            field2
-          end
+          return field2 unless field_type.kind.fields?
```
rmosolgo: Style-wise, I shy away from early return when it separates the structure of code from the meaning. Like, the meaning of this method is, "if there are fields to merge, we should merge them." So prefer code that matches that phrasing. 

(Of course, maybe someone else thinks of this method as, "unless there are fields to merge, return one of the fields".)

But they're both valid and IMO, as long as it's well-tested & split into meaningful units of code, I'd rather not delay progress to bicker about style!

PR #96, opened 2016-02-15
Remove base execution, use serial_execution instead

xuorig:
As discussed on Twitter, let's just use `SerialExecution` instead of having `BaseExecution` for now.


rmosolgo:
:moneybag:  I wanted to make sure we didn't break `graphql-batch` and it looks like we're in the clear: https://github.com/Shopify/graphql-batch/blob/master/lib/graphql/batch/execution_strategy.rb#L2

Thanks!

Issue #91, opened 2016-02-01
Make a nice API for composition


rmosolgo:
What if you have a bunch of types with the same fields? I've been passing the objects into functions and adding fields to them, but it would be good to have a proper API for that. One idea I have is a "mixin", like: 

```ruby
HasBirthday = GraphQL::Mixin.define do 
  field(:birthday, DateType)
  field(:age_in_years, types.Int)
  field(:can_vote, types.Boolean) 
end

Person = GraphQL::ObjectType.define do 
  mixin(HasBirthday)
end

Cat = GraphQL::ObjectType.define do 
  mixin(HasBirthday)
end
```

peterhorne:
Perhaps this is because a simple example has been used for pedagogical purposes but it seems to me that types currently support this scenario just fine. `Person` and `Cat` both have a `birthday` field which resolves to a `Birthday` type with fields `date`, `age_in_years`, and `can_vote`?

That said I have run into scenarios where it would be useful to be able to define resolve methods on an interface's fields (perhaps that can be optionally overridden by the implementing type). Would that cover the use cases supported by mixins while also keeping the number of concepts small?

rmosolgo:
That's an interesting point! Use "subtypes" to capture that shared behavior. I share your desire to keep the number of concepts small. Maybe a suitable solution to this issue would be to add a guide for "Code Reuse Patterns"

That example is a bit contrived, here's a real-life example: 

#### With "mixin"

```ruby
HasCheckInCounts = GraphQL::Mixin.define do 
  field(:total_count, types.Int) 
  field(:guest_count, types.Int) 
  field(:regular_count, types.Int) 
  field(:volunteer_count, types.Int) 
end

EventType = GraphQL::ObjectType.define do 
  mixin(HasCheckInCounts)
end 

LocationType = GraphQL::ObjectType.define do 
  mixin(HasCheckInCounts)
end 
```

#### With a module 

This is how I've been solving problems like this lately. There's nothing pretty about it but I like it because it uses plain-Ruby concepts and the API gives plenty of room for customizing or changing the implementation later on. 

```ruby
module CheckInCountsFieldsFactory
  COUNT_FIELDS = [
    :total_count,
    :guest_count,
    :regular_count,
    :volunteer_count,
  ]

  def self.create(object_type)
    COUNT_FIELDS.each do |field_name|
      object_type.field(field_name, object_type.types.Int)
    end
  end
end

EventType = GraphQL::ObjectType.define do 
[...]

PR #97, opened 2016-02-16
Add field mixin to type configurations ?

xuorig:
Small PR just to spark discussion of adding a `mixin` method in `DefinedByConfig` like proposed by @rmosolgo in #91

`Mixin` class is simply a `DefinedByConfig` class containing fields only for now.


rmosolgo:
For my part, I was convinced by: https://github.com/rmosolgo/graphql-ruby/issues/91#issuecomment-184615939

It will make field lookups a bit tougher, but it will make interfaces _actually_ useful!

PR #100, opened 2016-02-19
refactor: Move code to add the error to the context in field resolution.

dylanahsmith:
@rmosolgo & @eapache for review

This refactoring will allow graphql-batch to remove the adding of the error to the context in its override of get_finished_value (https://github.com/Shopify/graphql-batch/blob/7467505e80d2fd21333f5ca028f689692f6a5393/lib/graphql/batch/execution_strategy.rb#L54-L56) and replace those lines with `super(error)`

The adding of the error to the context was actually done in get_finished_value in the latest release (https://github.com/rmosolgo/graphql-ruby/blob/v0.10.9/lib/graphql/query/serial_execution/field_resolution.rb#L30) so this just reverts it back to working that way, while still catching execution errors raised from get_raw_value

### DIFF
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -19,8 +19,6 @@ module GraphQL
           raw_value = begin
             get_raw_value
           rescue GraphQL::ExecutionError => err
-            err.ast_node = ast_node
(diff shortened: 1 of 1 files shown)


eapache:
LGTM

rmosolgo:
nice catch!

Issue #98, opened 2016-02-17
Any style guide for contributions?


aviscasillas:
Hi there!

I was wondering if you were following some style guide for coding, or kind of best practices.
What do you think about boost styling based on the [bbatsov/ruby-style-guide](https://github.com/bbatsov/ruby-style-guide)?

Thanks!

rmosolgo:
Sorry, not yet! I was thinking I should do this. I'll to add a `.rubocop.yml` soon. In the mean time, as long as the test coverage is good, we're free to change code pretty freely :)

aviscasillas:
Well, I've just started playing with the code to see how is it made, and I think that by the way I could take advantage of that to make some styling refactors. IMHO even you are not forcing it, If you have the idea in mind, I think that would not be bad to make some refactors to have as more code in the style as better  :blush: 

Just as an example, we could move thinks like this:
```ruby
module GraphQL::DefinitionHelpers::DefinedByConfig
  # some code here
end
```
to this:
```ruby
module GraphQL
  module DefinitionHelpers
    module DefinedByConfig
      # some code here
    end
  end
end
```

This way we could also remove empty/redundant definitions like that: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/definition_helpers.rb#L1-L4

What do you think?

rmosolgo:
Yes, that particular fix would be great! When I started this project, I was using `autoload` :grimacing: Those are left-over from the old way.

aviscasillas:
:+1: I will do it then! :wink:

PR #103, opened 2016-02-23
Refactor: One line class/module definitions to nested style

aviscasillas:
As commented here https://github.com/rmosolgo/graphql-ruby/issues/98, this PR is only a refactor to move inner class/module definitions to nested style.
It have a big diff (cause the indentation for all refactored files changed), but there are no functional changes so no specs update is required.

### DIFF
--- a/lib/graphql/definition_helpers.rb
+++ b/lib/graphql/definition_helpers.rb
@@ -1,8 +1,3 @@
-module GraphQL
-  module DefinitionHelpers
-  end
-end
-
 require 'graphql/definition_helpers/non_null_with_bang'
 require 'graphql/definition_helpers/defined_by_config'
 require 'graphql/definition_helpers/string_named_hash'
--- a/lib/graphql/definition_helpers/defined_by_config.rb
+++ b/lib/graphql/definition_helpers/defined_by_config.rb
@@ -1,119 +1,123 @@
-# Provide a two-step definition process.
-#
-# 1. Use a config object to gather definitions
-# 2. Transfer definitions to an actual instance of an object
-#
-module GraphQL::DefinitionHelpers::DefinedByConfig
-  def self.included(base)
-    base.extend(ClassMethods)
-  end
+module GraphQL
+  module DefinitionHelpers
+    # Provide a two-step definition process.
+    #
+    # 1. Use a config object to gather definitions
+    # 2. Transfer definitions to an actual instance of an object
+    #
+    module DefinedByConfig
+      def self.included(base)
+        base.extend(ClassMethods)
+      end
 
-  # This object is `instance_eval`'d  when defining _any_ object in the schema.
-  # Then, the applicable properties of this object are transfered to the given instance.
-  class DefinitionConfig
-    def self.attr_definable(*names)
-      attr_accessor(*names)
-      names.each do |name|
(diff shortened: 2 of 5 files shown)


lib/graphql/static_validation/all_rules.rb
```diff
@@ -1,23 +0,0 @@
-# Default rules for {GraphQL::StaticValidation::Validator}
-#
-# Order is important here. Some validators return {GraphQL::Language::Visitor::SKIP}
-# which stops the visit on that node. That way it doesn't try to find fields on types that
-# don't exist, etc.
-GraphQL::StaticValidation::ALL_RULES = [
```
rmosolgo: why the move? Why not 

```ruby 
module GraphQL 
  module StaticValidation 
    ALL_RULES = [ .... ] 
   end 
end
```

?

lib/graphql/static_validation.rb
```diff
@@ -1,15 +1,38 @@
...
-end
-
 require 'graphql/static_validation/message'
 require 'graphql/static_validation/arguments_validator'
 require 'graphql/static_validation/type_stack'
 require 'graphql/static_validation/validator'
 require 'graphql/static_validation/literal_validator'
 
-rules_glob = File.expand_path("../static_validation/rules/*.rb", __FILE__)
+rules_glob = File.expand_path('../static_validation/rules/*.rb', __FILE__)
```
rmosolgo: lol, i take it you're a single-quotes type

lib/graphql/static_validation/all_rules.rb
aviscasillas: Sorry, maybe this is a bit out of the scope fo the refactor. 
The change is because it's kind of general rule to have files by class/module definitions. Also is adopted as a rule [here](https://github.com/bbatsov/ruby-style-guide#one-class-per-file). 
So, as it's a constant in `StaticValidation` module I thought that it was more consistent having it defined there.

I can revert this change if you prefer...

lib/graphql/static_validation.rb
aviscasillas: Yup! I prefer to use single quotes when double ones are not needed. But in this case I did it for consistency with the ones above :blush:.

rmosolgo:
Earlier, you mentioned changing the class definitions from `::` syntax to nested definitions, eg, from:

```ruby
class A::B
end
```

to 

```ruby
module A 
  class B
  end 
end
```

That was ok for me because it actually improves the codebase -- you have fewer order-dependent requires, you could theoretically give people options about what they want to load. 

But I don't want to merge this PR as-is because it includes a lot of changes to make things fit your personal style without providing any fixes or improved behavior to outside users. My concern is that style-only changes add noise to the git history, making it a bit less useful. I really prefer `git blame` to be a useful tool for discovering _why_ some code exists. I think whitespace-only commits make `git blame` harder to use effectively!

So, do you mind removing your other style changes?

aviscasillas:
Ok I get your point, so I will rebase the commit removing style changes :+1:

aviscasillas:
@rmosolgo I rebased the commit to have just changes for definition_helpers. There are some spaces diff there. It's applied automatically by my editor, but I can fix them if you want.

rmosolgo:
Thanks!

PR #104, opened 2016-02-26
More details on input validation errors

theorygeek:
This PR adds more details to the error messages that you get when you provide invalid input variables to a GraphQL query.

Currently, you get error messages like this:
```js
errors: [
  {
    message: "Variable input of type [DairyProductInput] was provided invalid value [{\"foo\":\"bar\"}]",
    locations: [{line: 1, column: 11}]
  }
]
```
The message doesn't include much information about _what_ is wrong with the value. So the goal here is to provide more information, so that you can debug faster:
```js
errors: [
  {
    message: "Variable input of type [DairyProductInput] was provided invalid value",
    locations: [{ line: 1, column: 11 }],
    value: [{ foo: "bar" }],
    problems: [
      { path: [0, "foo"], explanation: "Field is not defined on DairyProductInput" },
      { path: [0, "source"], explanation: "Expected value to not be null" }
    ]
  }
]
```

Any feedback is appreciated!

### DIFF
--- a/lib/graphql/input_object_type.rb
+++ b/lib/graphql/input_object_type.rb
@@ -27,6 +27,32 @@ class GraphQL::InputObjectType < GraphQL::BaseType
       input_fields.all? { |name, field| field.type.valid_input?(input[name]) }
   end
 
+  def validate_non_null_input(input)
+    # Inputs are only required to provide an `all?` method and a `[]` method. That's not enough
+    # to get detailed errors, so if that's all we got, we can only provide generalized error.
+    # (see spec/support/mimimum_input_object.rb)
+    unless input.is_a?(Enumerable)
+      result = GraphQL::Query::InputValidationResult.new
+      result.add_problem("Invalid input was provided") unless valid_non_null_input?(input)
+      return result
+    end
+
+    result = GraphQL::Query::InputValidationResult.new
+
+    # Items in the input that are unexpected
+    input.reject { |name, value| input_fields[name] }.each do |name, value|
+      result.add_problem("Field is not defined on #{self.name}", [name])
+    end
+
+    # Items in the input that are expected, but have invalid values
+    invalid_fields = input_fields.map do |name, field|
+      field_result = field.type.validate_input(input[name])
+      result.merge_result!(name, field_result) unless field_result.is_valid?
+    end
+
+    result
+  end
+
   def coerce_non_null_input(value)
     input_values = {}
     input_fields.each do |input_key, input_field_defn|
--- a/lib/graphql/input_object_type.rb
+++ b/lib/graphql/input_object_type.rb
@@ -20,22 +20,8 @@ class GraphQL::InputObjectType < GraphQL::BaseType
     GraphQL::TypeKinds::INPUT_OBJECT
   end
 
-  # assert that all present fields are defined
-  # _and_ all defined fields have valid values
-  def valid_non_null_input?(input)
-    input.all? { |name, value| input_fields[name] } &&
-      input_fields.all? { |name, field| field.type.valid_input?(input[name]) }
-  end
-
   def validate_non_null_input(input)
-    # Inputs are only required to provide an `all?` method and a `[]` method. That's not enough
-    # to get detailed errors, so if that's all we got, we can only provide generalized error.
-    # (see spec/support/mimimum_input_object.rb)
-    unless input.is_a?(Enumerable)
-      result = GraphQL::Query::InputValidationResult.new
-      result.add_problem("Invalid input was provided") unless valid_non_null_input?(input)
-      return result
-    end
+    fail ArgumentError.new("input must be enumerable") unless input.is_a?(Enumerable)
 
     result = GraphQL::Query::InputValidationResult.new
 
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -6,46 +6,6 @@ class GraphQL::Query
     end
   end
 
-  class InputValidationResult
-    attr_accessor :problems
-
-    def initialize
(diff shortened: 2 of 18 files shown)


lib/graphql/input_object_type.rb
```diff
@@ -27,6 +27,32 @@ def valid_non_null_input?(input)
       input_fields.all? { |name, field| field.type.valid_input?(input[name]) }
   end
 
+  def validate_non_null_input(input)
+    # Inputs are only required to provide an `all?` method and a `[]` method. That's not enough
```
rmosolgo: It's ok to extend the requirements here -- the real reason I added that was to make sure there wasn't any hard type-checking. 

Previously, it required that inputs were a `Hash`, but that didn't play nice when inputs came as `ActionController::Params`, which I definitely want to support! 

So, if you want to add a few common methods to that "minimum requirements", please do!

lib/graphql/input_object_type.rb
```diff
@@ -27,6 +27,32 @@ def valid_non_null_input?(input)
...
+    unless input.is_a?(Enumerable)
+      result = GraphQL::Query::InputValidationResult.new
+      result.add_problem("Invalid input was provided") unless valid_non_null_input?(input)
+      return result
+    end
+
+    result = GraphQL::Query::InputValidationResult.new
+
+    # Items in the input that are unexpected
+    input.reject { |name, value| input_fields[name] }.each do |name, value|
```
rmosolgo: This _should_ be handled by static validation... oh, but I guess if you're making an input object out of query variables, it would be nice to get some feedback?

lib/graphql/query.rb
```diff
@@ -6,17 +6,43 @@ def initialize(names)
     end
   end
 
+  class InputValidationResult
```
rmosolgo: how about putting this in `lib/graphql/query/input_validation_result.rb`, just to keep `query.rb` a bit smaller?

rmosolgo:
Wow, this approach seems really solid! The recursive nature of it keeps each little part very simple, which is awesome. 

I wonder, could we integrate this behavior with the `valid_non_null_input?` methods? Like, you know how in Rails, if you call `valid?`, it validates the object and then returns whether it has errors or not? Could we do something similar here? 

(I ask because if I understand right, calling `valid_non_null_input?` recurses the input, then calling `validate_non_null_input` recurses it again, I wonder if we could get them both with one pass.)

lib/graphql/input_object_type.rb
theorygeek: Ok, no prob! Do you think that it'd be OK to require that it be `Enumerable` and support `[]`?

My line of thinking is, the minimum requirement would be an `each` method. But at that point, it's a small step to also ask that it be `Enumerable`, since that's all you need for that anyway. So may as well just ask for it to be `Enumerable`.

lib/graphql/input_object_type.rb
theorygeek: Hmm, so I was trying to follow the `valid_non_null_input?` method's logic. This was how I was interpreting the conditional in that method:
```ruby
input.all? { |name, value| input_fields[name] }
```
So I assumed that it was needed here.

rmosolgo: Yeah, I think you're right. If you pass in the input object from variables, then it's not handled by static validation!

theorygeek:
@rmosolgo Thanks, man. I'll work on combining those two methods.

theorygeek:
@rmosolgo Made some changes in response to your comments:
* Moved both `InputValidationResult` and `VariableValidationError` into separate files
* Combined `validate_non_null_input` and `valid_non_null_input?` methods
* Increased minimum variable requirements to be `Enumerable` and respond to `[]`

Ready for re-review.

lib/graphql/scalar_type.rb
```diff
@@ -19,8 +19,14 @@ def coerce=(proc)
       self.coerce_result = proc
     end
 
-    def valid_non_null_input?(value)
-      !coerce_non_null_input(value).nil?
+    def validate_non_null_input(value)
+      result = Query::InputValidationResult.new
+
+      unless !coerce_non_null_input(value).nil?
```
rmosolgo: `unless !` ... known to us mere mortals as `if` :P

lib/graphql/query/input_validation_result.rb
```diff
@@ -0,0 +1,23 @@
...
+      @problems.nil?
+    end
+
+    def add_problem(explanation, path = nil)
+      @problems ||= []
+      @problems.push({ 'path' => path || [], 'explanation' => explanation })
+    end
+
+    def merge_result!(path, inner_result)
+      return if inner_result.problems.nil?
```
rmosolgo: could use the public API here, `return if inner_result.is_valid?`

lib/graphql/query/input_validation_result.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      inner_result.problems.each do |p|
+        item_path = [path, *p['path']]
+        add_problem(p['explanation'], item_path)
+      end
+    end
+  end
+end
```
rmosolgo: Overall I think this object could be _simpler_ by initializing `@problems = []`, so it's never nil, but you could argue that in many cases that's a waste of an object!

theorygeek: It seemed like this was a semi-performance critical part of the code, since it's used to validate all of the input variables. So I kinda thought that only initializing `@problems` when needed could cut down on GC churn.

rmosolgo:
:moneybag: Thanks, this is so awesome!

PR #105, opened 2016-02-27
ExecutionContext object to wrap query and strategy during execution

xuorig:
As we add more and more features to the Executor, our `initialize` adds more and more arguments.

This PR introduces am `ExecutionContext` object that currently wraps `query` and `execution_strategy` as well as providing method to access or modify these fields.

This object wraps values that are carried all the way from the base execution up to the value resolution, and will be helpful to add any other values related to an entire query (from start to end)

I've added tests for it, but it might not be necessary as other parts are already tested.

Let me know what you think.


rmosolgo:
Nice, thanks, I think this will pay off down the road! 

Thanks for adding a few unit tests too, though I agree, it's pretty well-tested by the fact that all the _other_ tests pass!

PR #106, opened 2016-02-28
Maximum query depth at execution

xuorig:
Introduces a `max_depth` argument on `GraphQL::Query`

Raises an `ExecutionError` at runtime if the maximum depth is exceeded.

Defaults to `nil`, and is not enforced in that case.

If a query like this one is executed with `max_depth: 3`:

```
      query maxDepth {
        cheese(id: 1) {
          source
          similarCheese(source: COW) {
            flavor
            similarCheese(source: COW) {
              flavor
            }
          }
        }
      }
```

it will raise a `GraphQL::ExecutionError` because it has a depth of 4

Question is: shoud it return a global error like it does right now, or should the rest of the query be returned, without the selections that are too deep ?

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -24,7 +24,7 @@ class GraphQL::Query
   # The executor will send the field's name to the target object
   # and use the result.
   DEFAULT_RESOLVE = :__default_resolve
-  attr_reader :schema, :document, :context, :fragments, :operations, :debug
+  attr_reader :schema, :document, :context, :fragments, :operations, :debug, :max_depth
 
   # Prepare query `query_string` on `schema`
   # @param schema [GraphQL::Schema]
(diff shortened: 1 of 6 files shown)


rmosolgo:
Hmm, this can be known during static validation, right? I think we should enforce it there instead. That way, we don't have to answer the question about the "rest of the query". (Besides, in the case of a mutation, getting _halfway through_ then stopping could be really bad!)

The only problem is that right now, static validation is not configurable query-by-query. I think for depth validation, it would be best, in the long run, to have: 

- default for a schema 
- override for a query 

So, I think we would need to change this method: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/validator.rb#L21

Either it could accept some options, which are passed into validation (eg `validate(document, max_depth: 10)`) _or_ it could accept a list of additional validators to apply (eg `validate(document, rules: [DepthValidator.new(max: 10)])`). Personally I prefer the second approach because it keeps the door open for growth. (And I definitely want to add a customizable complexity validation, like Sangria has!) 

What do you think of those ideas?

xuorig:
@rmosolgo yeah I've thought about implementing it as static validation too, but went the execution way because of how `sangria` was doing it.

I can't think of any reason we can't do it as a static validation, you're right. I'll try to come up with something else this week, closing this for now!

PR #108, opened 2016-02-29
feat(ObjectType) lookup field implementations from interface, too

rmosolgo:
As described in [our earlier conversation](https://github.com/rmosolgo/graphql-ruby/issues/91#issuecomment-184615939), this allows interfaces to be used as mixins. 

You can get field implementations from the interface _or_ you can override them in the object type. 

One outstanding issue is, can we cache the result of looking up a field? At what level? Some things I want to support, but don't support well yet, are: 

- Rails development reloading 
- Multiple schemas in one application

### DIFF
--- a/lib/graphql/interface_type.rb
+++ b/lib/graphql/interface_type.rb
@@ -23,4 +23,14 @@ class GraphQL::InterfaceType < GraphQL::BaseType
   def possible_types
     @possible_types ||= []
   end
+
+  # @return [GraphQL::Field] The defined field for `field_name`
+  def get_field(field_name)
+    self.fields[field_name]
+  end
+
+  # @return [Array<GraphQL::Field>] All fields on this type
+  def all_fields
+    self.fields.values
+  end
 end
--- a/lib/graphql/interface_type.rb
+++ b/lib/graphql/interface_type.rb
@@ -26,11 +26,11 @@ class GraphQL::InterfaceType < GraphQL::BaseType
 
   # @return [GraphQL::Field] The defined field for `field_name`
   def get_field(field_name)
-    self.fields[field_name]
+    fields[field_name]
   end
 
   # @return [Array<GraphQL::Field>] All fields on this type
   def all_fields
-    self.fields.values
+    fields.values
   end
 end
--- a/lib/graphql/object_type.rb
+++ b/lib/graphql/object_type.rb
@@ -50,7 +50,7 @@ class GraphQL::ObjectType < GraphQL::BaseType
 
   # @return [Array<GraphQL::Field>] All fields, including ones inherited from interfaces
   def all_fields
-    interface_fields = interfaces.inject({}) do |memo, iface|
+    interface_fields = interfaces.reduce({}) do |memo, iface|
       memo.merge(iface.fields)
     end
     interface_fields.merge(self.fields).values
@@ -61,7 +61,7 @@ class GraphQL::ObjectType < GraphQL::BaseType
   # Find the _last_ definition for `field_name`
   def get_interface_field(field_name)
     interfaces.reduce(nil) do |memo, iface|
-      iface.fields[field_name] || memo
+      iface.get_field(field_name) || memo
     end
   end
 end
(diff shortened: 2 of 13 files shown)


lib/graphql/object_type.rb
```diff
@@ -42,4 +42,26 @@ def interfaces=(new_interfaces)
...
   end
+
+  # @return [GraphQL::Field] The field definition for `field_name` (may be inherited from interfaces)
+  def get_field(field_name)
+    fields[field_name] || get_interface_field(field_name)
+  end
+
+  # @return [Array<GraphQL::Field>] All fields, including ones inherited from interfaces
+  def all_fields
+    interface_fields = interfaces.inject({}) do |memo, iface|
```
xuorig: How about `interfaces.reduce({}, :merge)` ? :dancer:

lib/graphql/object_type.rb
```diff
@@ -42,4 +42,26 @@ def interfaces=(new_interfaces)
...
+    end
+    interface_fields.merge(self.fields).values
+  end
+
+  private
+
+  # Find the _last_ definition for `field_name`
+  def get_interface_field(field_name)
+    interfaces.reduce(nil) do |memo, iface|
+      iface.fields[field_name] || memo
```
xuorig: `iface.get_field(field_name)` ?

lib/graphql/object_type.rb
```diff
@@ -42,4 +42,26 @@ def interfaces=(new_interfaces)
...
+      memo.merge(iface.fields)
+    end
+    interface_fields.merge(self.fields).values
+  end
+
+  private
+
+  # Find the _last_ definition for `field_name`
+  def get_interface_field(field_name)
+    interfaces.reduce(nil) do |memo, iface|
```
xuorig: why do you need to reduce(nil) here ? `interface.get_field` should return `nil` if it's not present right ?

lib/graphql/interface_type.rb
```diff
@@ -23,4 +23,14 @@ def kind
   def possible_types
     @possible_types ||= []
   end
+
+  # @return [GraphQL::Field] The defined field for `field_name`
+  def get_field(field_name)
+    self.fields[field_name]
```
xuorig: nit: lose the `self` here and in `all_fields` to be consistent with other places ?

xuorig:
I like this! it's going to be super useful :tada:

lib/graphql/object_type.rb
rmosolgo: good point, `.fields` is almost a private API  now!

lib/graphql/object_type.rb
rmosolgo: But, if I don't provide an initial value, Ruby will take the first item from the list as the initial value. And `memo` should always be a `GraphQL::Field`-or-`nil`, not ever an interface!

xuorig: Hmm right! I guess what I'm saying is why not have `interface_fields` as a method itself, in which you can use `interfaces.reduce({}, :merge)` then you can replace that part in both `all_fields` and `get_interface_field` which becomes `interface_fields[field_name]`

Just a nit though, don't want to slow you down!

lib/graphql/interface_type.rb
rmosolgo: Seems like kind of a waste, `all_fields` flattens the hash to an array, then `get_fields` picks a member out of the array by name. We can save a few steps by looking up directly from the hash!

xuorig: My bad, Did not notice, :shipit:

lib/graphql/object_type.rb
rmosolgo: Oh, now I understand, yes, that's much simpler! 

I think I'll still use the block for merging though. I could add `InterfaceType#merge` for this purpose but I don't think it's worth expanding the API

xuorig: Perfect!

peterhorne:
Yay! Looking forward to giving this a spin :grin: :tada:

PR #109, opened 2016-03-04
Fixing bug with enum literals when using libgraphqlparser

theorygeek:
This fixes a bug when you're using the [`graphql-libgraphqlparser`](https://github.com/rmosolgo/graphql-libgraphqlparser-ruby) gem. The problem is when you have a query that contains enum literals:
```
query {
  someField(enumArg: SOME_ENUM_VALUE) {
    ...
  }
}
```
This results in an error message like this:
```
Argument 'enumArg' on Field 'someField' has an invalid value
```
The problem is that instead of getting a `GraphQL::Language::Nodes::Enum` for the argument, you just get a `String`. I think that the "correct" fix would be to modify the `graphql-libgraphqlparser` gem, but I wasn't entirely sure where the problem was, and at any rate, this little band-aid seems to resolve the problem.

### DIFF
--- a/lib/graphql/query/literal_input.rb
+++ b/lib/graphql/query/literal_input.rb
@@ -64,7 +64,7 @@ module GraphQL
 
         module EnumLiteral
           def self.coerce(value, type, variables)
-            type.coerce_input(value.name)
+            type.coerce_input(value.is_a?(String) ? value : value.name)
           end
         end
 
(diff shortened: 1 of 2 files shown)


theorygeek:
@rmosolgo Does this cause non-conformance with the spec? The [specific section I'm thinking about](https://facebook.github.io/graphql/#sec-Enums) says:
>GraphQL has a constant literal to represent enum input values. GraphQL string literals must not be accepted as an enum input and instead raise a query error.

I think this code might cause queries that are wrapping the enum literals in `"` quotes to succeed, rather than fail.

rmosolgo:
I think you're right, this is really a bug for `graphql-libgraphqlparser`. The behavior for parsing enums wasn't spec'd (so I added that 8c34ccb). And I fixed `graphql-libgraphqlparser` to match (cdc47b3) and released `0.1.3`. Could you try that new version and see if it works for you?

theorygeek:
Awesome! That fixed the problem. :)

PR #111, opened 2016-03-04
Meaningful error messages for InvalidTypeError

MatthiasWinkelmann:
This adds a meaningful message for InvalidTypeErrors when the type is defined in a proc. Example:

Before:

`Type "Unnamed type" is invalid`

After:

`Type is invalid: Proc defined in /Users/project/app/graph/types/viewer_type.rb:21     argument :timestep, -> {GraphQL::STRING_TYPE}
. Must be a GraphQL::BaseType):`

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -69,7 +69,13 @@ class GraphQL::Schema
 
   class InvalidTypeError < StandardError
     def initialize(type, errors)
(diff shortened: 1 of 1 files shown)


MatthiasWinkelmann:
...I have now started to doubt that a Proc should ever get to that point. Carry on...

PR #113, opened 2016-03-07
Contextual errors when invalid types are provided for fields

theorygeek:
Provides better error messages when you have an invalid field definition. Specifically adds context to error messages when you use an invalid type:
* As the return type for a field
* As the argument type for a field argument
* As an input type on an input object

It also tells you the class of the object that you provided as a type.

**Example Use Case**
It's probably not uncommon that a GraphQL type shares a name with an Active Record model. Sometimes, it's easy to accidentally specify the model as a field type, instead of the GraphQL type that corresponds to the model:

```ruby
CreateCommentMutation = GraphQL::Relay::Mutation.define do
  name "CreateComment"
  ...
  return_field :comment, !Comment # should be CommentType, not Comment
end
```

If you make this mistake presently, you get a fairly unhelpful error message:
```
Type Unnamed type is invalid: Must be a GraphQL::BaseType
```
This PR adds more context to the error, to make debugging easier:
```
Type Unnamed type (NilClass) is invalid: Must be a GraphQL::BaseType (check the comment field on CreateCommentPayload)
```

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -70,10 +70,8 @@ class GraphQL::Schema
   end
 
   class InvalidTypeError < GraphQL::Error
-    def initialize(type, errors, context_message = nil)
-      message = "Type #{type.respond_to?(:name) ? type.name :  "Unnamed type"} (#{type.class.inspect}) is invalid: #{errors.join(", ")}"
-      message << " (#{context_message})" if context_message
-      super(message)
+    def initialize(type, name)
+      super("#{name} has an invalid type: must be an instance of GraphQL::BaseType, not #{type.class.inspect} (#{type.inspect})")
     end
   end
 end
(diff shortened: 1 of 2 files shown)


rmosolgo:
Cool! 

This part still seems a bit weird to me: 

> Type Unnamed type (NilClass) ... 

Could it be

> "CreateCommentPayload.comment" has an invalid type: must be instance of GraphQL::BaseType, not NilClass (nil)

?

(Since we have that specific info about _where_ to go, I think we should give it first!)

theorygeek:
@rmosolgo Ready for re-review. I put in slightly more info than you had suggested, to distinguish between fields, arguments, and input fields. New error messages look like:
>Field CreateCommentPayload.comment has an invalid type: must be an instance of GraphQL::BaseType, not NilClass (nil)

For invalid argument types:
> Argument someArg on SomeType.someField has an invalid type: must be an instance of GraphQL::BaseType, not NilClass (nil)

I also added some more info to the errors on interfaces and possible types, but in my own tests, if you provide an invalid type for one of those, the code breaks long before it gets to the type reducer, so those error messages will probably never get used.

rmosolgo:
That's awesome, thanks! :moneybag: :rocket:

PR #117, opened 2016-03-16
Handle the default resolving in GraphQL::Field.

dylanahsmith:
@rmosolgo is there any reason why the default resolve proc doesn't just resolve the value?

## Problem

We are using a custom field that allows preload to be specified independently from the actually resolve proc, so we can call methods on a model and test that it doesn't result in N+1 queries.  We are doing this by enumerating fields and using the Field#resolve_proc directly after calling a preload proc.  However, we are overriding build_default_resolver to avoid having to detect and handle when the resolve proc returns GraphQL::Query::DEFAULT_RESOLVE.

## Solution

Call `obj.public_send(name)` in Field#build_default_resolver rather than returning `GraphQL::Query::DEFAULT_RESOLVE`

### DIFF
--- a/lib/graphql/field.rb
+++ b/lib/graphql/field.rb
@@ -97,7 +97,11 @@ class GraphQL::Field
       if property
         obj.public_send(property)
       else
-        GraphQL::Query::DEFAULT_RESOLVE
+        begin
+          obj.public_send(name)
+        rescue NoMethodError => err
+          raise("Couldn't resolve field '#{name}' to #{obj.class} '#{obj}' (resulted in #{err})")
+        end
       end
     end
   end
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -69,15 +69,6 @@ module GraphQL
             context.ast_node = ast_node
             value = field_definition.resolve(parent_object, field_args, context)
             context.ast_node = nil
-
-            if value == GraphQL::Query::DEFAULT_RESOLVE
-              begin
-                value = target.public_send(ast_node.name)
-              rescue NoMethodError => err
-                raise("Couldn't resolve field '#{ast_node.name}' to #{parent_object.class} '#{parent_object}' (resulted in #{err})")
-              end
-            end
-
             value
           }
         end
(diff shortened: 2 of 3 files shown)


rmosolgo:
There's this half-baked API right now where you can pass an _existing_ `GraphQL::Field` to the field helper with the `field:` argument. But it gets tricky when you assign the same `GraphQL::Field` to two object types with two _different_ names. For example: 

```ruby
NameField = GraphQL::Field.define do 
  resolve -> (obj, args, ctx) { 
    # something complicated 
  }
end

Type1 = GraphQL::ObjectType.define do 
  field :name, field: NameField
end

Type2 = GraphQL::ObjectType.define do 
  field :alias, field: NameField 
end
```

The bug is: 

```ruby
NameField.name                      # "alias" 
Type1.get_field("name").name # "alias"
Type2.get_field("alias").name # "alias"
```

So the workaround was to say "a field doesn't _know_ the name it's been given by its object type" (just like a value in a hash doesn't know its corresponding key). Then, you `public_send(ast_field.name)` so you can be sure it's the right thing. (In this case, what's the _point_ of `GraphQL::Field#name`? I don't remember :disappointed: )

Obviously, it's a mess! 

In my opinion, the goals are: 

- Don't mutate the object passed to `field:` because mutation is awful (case in point above)
- Support & communicate some means for reusing field configurations (maybe not passing around `GraphQL::Field` instances, but something)
- Make default resolution as fast as possible

dylanahsmith:
Right, I was actually using `name` because I didn't trust `ast_node.name` when using graphql-batch.  In FieldResolution that isn't a problem since it has an instance of ast_node which doesn't change when `context.ast_node` changes, but it becomes a problem when relying on it in Field through `context.ast_node`.

We don't re-use field instances between fields, so it hasn't been a problem for us.

Although we also take a different approach to defining the schema, since the way `GraphQL::Field.define` works made it complicated to extend compared to just having class methods where a derived class or mixin can easily override field and provide additional options (e.g. preload for preloading associations)

rmosolgo:
Yes, maybe better to prevent people from passing the same `GraphQL::Field` to multiple `field` calls. I'll follow up with that change!

rmosolgo:
Wow, that made `Query::DEFAULT_RESOLVE` dead code, too -- what a win! :tada:

Rodeoclash:
@dylanahsmith Would love to see your code on how you're handling the N+1 queries

PR #119, opened 2016-03-23
feat(Language) use Racc for parsing

rmosolgo:
Still some todos, but all-in-all, looks very promising: 

```
Parse INTROSPECTION_QUERY 100 times:
                           user     system      total        real
Parslet                5.820000   0.040000   5.860000 (  5.874497)
Racc + Ragel           0.330000   0.010000   0.340000 (  0.328350)
Libgraphqlparser       0.040000   0.000000   0.040000 (  0.052346)
```

__TODO__:

- [x] get position information 
- [x] raise custom errors
- [ ] somehow package these tests so that they can be shared by `graphql-libgraphqlparser`


rmosolgo:
Gonna look into Ragel for tokenizing, that might have built-in support for position tracking (and it's what the `parser` gem uses)

rmosolgo:
Using Ragel is significantly slower (but still way faster than Parslet). And it provides tools for position tracking which is a must-have in my opinion

New Results: 

```
~/projects/graphql $ ruby -Ilib parse_benchmark.rb
Parse INTROSPECTION_QUERY 100 times:
                           user     system      total        real
Parslet                5.820000   0.040000   5.860000 (  5.874497)
Racc + Ragel           0.330000   0.010000   0.340000 (  0.328350)
Libgraphqlparser       0.040000   0.000000   0.040000 (  0.052346)
```


Old Results: 

```
Parse INTROSPECTION_QUERY 100 times:
                           user     system      total        real
Parslet                6.310000   0.070000   6.380000 (  6.426334)
Racc + Rexical         0.100000   0.000000   0.100000 (  0.103924)
Libgraphqlparser       0.050000   0.000000   0.050000 (  0.054631)
```

fikriauliya:
I am interested to start converting my RESTful API to GraphQL, do you have benchmark comparing both? I am just afraid GraphQL Query/Type matching would be slower than simple params matching di RESTful.

Any insight to share?

rmosolgo:
I also think a benchmark would be interesting! But it would be hard to make an apples-to-apples comparison. 

For example, we might consider testing these two against each other: 

```
GET /api/things/1.json 
```

```
POST /graphql 
{
  thing(id: 1) {
    # request every field on `thing` here 
  }
}
```

In this case, I expect  `GET ... 1.json` to be faster, because like you said, you don't have to pay the "overhead" for GraphQL's flexibility. 

However, I don't think it's a good comparison, because it has skipped some advantages of GraphQL: 

- We asked for _all_ fields on `thing(id: 1)`, but which fields does the UI actually need? Perhaps we only need to request a few fields. 
- We only compared one `GET` request to one GraphQL query. But, will the client actually need to make _multiple_ `GET` requests to fetch all their data? If so, they can be grouped into one GraphQL query, so we should test _several_ `GET` requests against one GraphQL query. 

Given those two considerations, I think the best benchmark would be app-specific, even UI-specific. If you have a UI which is currently fetching JSON from a resource-based API, you could try switching it to GraphQL and timing the _whole_ load operation on the client. 

I did a benchmark like this back in December 2015 (before rewriting the parser). Each colored block is one API request:

image

I wrote it all into one giant GraphQL query and it was __slightly slower__ than doing each request separately. (It's hard to tell from the screenshot, but one calculation in particular requires recursive SQL queries :sob:) 

I only saved a screenshot of the flamegraph, but no labels 😖 . If I remember correctly, the dark blue at the very beginning was _all_ parsing ... which is gone now! I don't have a new flamegraph though.

image

I hope that's helpful. If you decide to do a comparison, I'd love to hear the results!

fikriauliya:
Thx for sharing the result :)

I migrated RESTful API + JBuilder into GraphQL. By no mean as a proper benchmark, but from repetitive call on the API endpoint, GraphQL is twice faster. (~500 ms vs ~1,000 ms, no cache)

PR #120, opened 2016-03-24
Avoid providing internal server error messages to clients.

dylanahsmith:
## Problem

This gem would send standard errors to clients, which could include sensitive information.  For example, when using the mysql2 gem, the error message could include the database query.

Ironically, we have been using the `debug: true` option to both hide the internal error messages and get access to the full error information for logging and for error reporting.  I would like to get rid of that hack.

## Solution

Use `"Internal error"` as the error message for standard errors.  ExecutionError should be used to report client errors.

Also, add the error to `query.context.errors` for logging and error reporting.

I also used an ExecutionError for OperationNameMissingError since I didn't want it to be treated as an internal error.  I also changed ExecutionError.to_h to omit the locations key so OperationNameMissingError continues to get rendered properly, and because there doesn't seem to be any reason to include the locations key when there is no location information for the error.

### DIFF
--- a/lib/graphql/execution_error.rb
+++ b/lib/graphql/execution_error.rb
@@ -11,9 +11,7 @@ module GraphQL
       hash = {
         "message" => message,
       }
-      if ast_node.nil?
-        hash["locations"] = []
-      else
+      if ast_node
         hash["locations"] = [
           {
             "line" => ast_node.line,
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -1,5 +1,5 @@
 class GraphQL::Query
-  class OperationNameMissingError < GraphQL::Error
+  class OperationNameMissingError < GraphQL::ExecutionError
     def initialize(names)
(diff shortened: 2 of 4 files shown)


lib/graphql/execution_error.rb
```diff
@@ -11,9 +11,7 @@ def to_h
       hash = {
         "message" => message,
       }
-      if ast_node.nil?
-        hash["locations"] = []
-      else
+      if ast_node
```
rmosolgo: :+1: from the spec, empahsis mine: 

> _If an error can be associated to a particular point_ in the requested GraphQL document, it should contain an entry with the key locations

rmosolgo:
Thanks! 


I feel the same way about `debug: true`. It's not really "debug" at all -- it's more like "please, just let me handle my errors". 

I wonder if it would make more sense to remove `debug:` and add `rescue_all_errors:` (default `true`, if we want to maintain backwards compatibility).

PR #121, opened 2016-03-24
feat(Directive) update to new spec for Directive locations

rmosolgo:
A little bit ago, the Directive spec changed: https://github.com/facebook/graphql/pull/152

So, update!

### DIFF
--- a/lib/graphql/introspection/directive_type.rb
+++ b/lib/graphql/introspection/directive_type.rb
@@ -4,7 +4,5 @@ GraphQL::Introspection::DirectiveType = GraphQL::ObjectType.define do
   field :name, !types.String, "The name of this directive"
   field :description, types.String, "The description for this type"
   field :args, field: GraphQL::Introspection::ArgumentsField
-  field :onOperation, !types.Boolean, "Does this directive apply to operations?", property: :on_operation?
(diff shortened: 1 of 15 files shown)


lib/graphql/introspection/directive_type.rb
```diff
@@ -4,7 +4,5 @@
   field :name, !types.String, "The name of this directive"
   field :description, types.String, "The description for this type"
   field :args, field: GraphQL::Introspection::ArgumentsField
-  field :onOperation, !types.Boolean, "Does this directive apply to operations?", property: :on_operation?
```
calebmer: Why did you remove this and not deprecate it like [GraphQL JS](https://github.com/graphql/graphql-js/blob/0aa78f61a2dc150b5ea9ee4f50b68a736796f068/src/type/introspection.js#L92-L114)? This breaks ecosystem tools like [GraphiQL](https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js#L29-L31) that have not yet updated.

calebmer: \cc @maxcodes

rmosolgo: Oops, error on my part! They can be re-implemented & marked as deprecated, sorry!

rmosolgo: Replacing them in #164

PR #123, opened 2016-03-29
feature: Support inline fragments without a type condition

dylanahsmith:
## Problem

A graphql query with an inline fragment that doesn't have a type condition would be considered an invalid query.

See the [Inline Fragments section of the graphql spec](http://facebook.github.io/graphql/#sec-Inline-Fragments) which shows that the type condition is optional and includes examples of queries with inline fragments without a type condition.

## Solution

There were several places where it was assumed that a fragment always had a type, so I had to update the parser, transformer, static validator and the selection resolution.  Basically if the type condition isn't specified, then we can skip the type check and continue using the parent type.

### DIFF
--- a/lib/graphql/language/parser.rb
+++ b/lib/graphql/language/parser.rb
@@ -26,10 +26,11 @@ module GraphQL
         directives.maybe.as(:optional_directives).as(:directives)
       }
       rule(:spread) { str("...") }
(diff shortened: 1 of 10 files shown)


rmosolgo:
Ohh I see it was added a while back: https://github.com/facebook/graphql/pull/100 I guess I missed it! 

Thanks!

PR #125, opened 2016-03-29
Optional auto-generated input type

brysgo:
Just spit-balling ideas here, what are your thoughts on having something like this. I sometimes find myself wanting to send down an object type in a mutation without having to redefine structure.


brysgo:
Thanks for the feedback, maybe I'll put it into a gem. I'll let you know what I end up doing.

Also, thanks for creating your gems. They are just that!

PR #129, opened 2016-04-01
Middleware Argument Override

theorygeek:
Allows middleware to change the arguments that will be passed to the next middleware.

**Use Case**
Currently, middleware can be used if you want to stop execution and return a value early. For example, this is helpful for authorizing field access and returning an error.

However, sometimes you may simply want to _modify_ the value that is passed to the next middleware. In my case, I have a hash that tells me how fields on my GraphQL schema map to attributes on my models. Sometimes, I need to traverse across an association to produce the value.

This change allows me to load the associations in middleware, and then provide the associated model to the field's resolver, before the attribute access happens.

### DIFF
--- a/lib/graphql/schema/middleware_chain.rb
+++ b/lib/graphql/schema/middleware_chain.rb
@@ -17,7 +17,8 @@ module GraphQL
       end
 
       # Run the next step in the chain, passing in arguments and handle to the next step
-      def call
+      def call(next_arguments = @arguments)
+        @arguments = next_arguments
         next_step = steps.shift
         next_middleware = self
         next_step.call(*arguments, next_middleware)
--- a/spec/graphql/schema/middleware_chain_spec.rb
+++ b/spec/graphql/schema/middleware_chain_spec.rb
@@ -27,5 +27,16 @@ describe GraphQL::Schema::MiddlewareChain do
         assert_equal([1,2], step_values)
       end
     end
+
+    describe "when a step provides alternate arguments" do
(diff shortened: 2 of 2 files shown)


rmosolgo:
This is a really good feature to have! I'm just concerned that the existing middleware implementation isn't very good. 

Rebuilding the whole args array seems error-prone and tedious, I want to think of something more user-friendly. I'm not sure if it should be named args or a `Context` object, or what... hmm....

theorygeek:
Hmm, yeah that makes sense. Sounds like you're thinking that it'd be a breaking change? Another option would be to make the `MiddlewareChain` more aware of how it's being used; perhaps it only lets you modify one particular parameter, instead of all of them.

If you want to give me a little bit more guidance on the change, I don't mind putting some work on it.

rmosolgo:
Maybe I'm just worrying too much. Let's get it out in this form now (better _something_ than nothing :D). 

In the future, maybe an API like this would work: 


```ruby 
class CustomMiddleware 
  def call(parent_object:, next_middleware:, **rest)
    replacement_object = # ... 
    # replace an argument 
    replaced_middleware = next_middleware.replace_argument(:parent_object, replacement_object)
    # continue the execution with the new middleware
    replaced_middleware.call
  end 
end 
```

I prefer it because: 

- Keywords > long list of positional arguments 
- No mutability (lookin at you, Rack `env`)
- Really explicit API 

The only problem is, if the replacement is happening all the time, do we _need_ to mutate existing objects for performance's sake?

Another challenge for another day :)

PR #131, opened 2016-04-05
Fixing bug with list nullability checks

theorygeek:
This fixes a bug where a variable of type `[DairyAnimal!]!` was considered invalid for an argument of type `[DairAnimal!]`.

According to the [GraphQL Spec §5.7.6](https://facebook.github.io/graphql/#sec-All-Variable-Usages-are-Allowed):
> If any list level of _variableType_ is not non‐null, and the corresponding level in _argument_ is non‐null, the types are not compatible.

This seems to suggest that the types _are_ compatible if any list level of _variableType_ is non-null, and the corresponding level in _argument_ is not non-null. (You have to read that sentence twice for it to make sense :P)

So I've added a test to ensure that:
* a variable of type `[DairyAnimal!]!` is compatible with an argument of type `[DairyAnimal!]`
* a variable of type `[[DairyAnimal!]!]!` is compatible with an argument of type `[[DairyAnimal]]`

Oddly enough, the second scenario was already working, but I think only accidentally. I think it's because the [`.wraps?` check inside of `non_null_levels_match`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb#L80) is true for both `Non-Null` and `List`, so you'd get a weird, recursive, off-by-one situation. I'm not sure if that would've actually caused any observable bugs, but at any rate, I think this fixes them also.

### DIFF
--- a/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
+++ b/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
@@ -78,6 +78,9 @@ class GraphQL::StaticValidation::VariableUsagesAreAllowed
     if arg_type.kind.non_null? && !var_type.kind.non_null?
       false
     elsif arg_type.kind.wraps? && var_type.kind.wraps?
+      # If var_type is a non-null wrapper for a type, and arg_type is nullable, peel off the wrapper
+      # That way, a var_type of `[DairyAnimal]!` works with an arg_type of `[DairyAnimal]`
+      var_type = var_type.of_type if var_type.kind.non_null? && !arg_type.kind.non_null?
       non_null_levels_match(arg_type.of_type, var_type.of_type)
     else
       true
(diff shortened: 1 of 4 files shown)


lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
```diff
@@ -78,6 +78,11 @@ def non_null_levels_match(arg_type, var_type)
     if arg_type.kind.non_null? && !var_type.kind.non_null?
       false
     elsif arg_type.kind.wraps? && var_type.kind.wraps?
+      # If var_type is a non-null wrapper for a type, and arg_type is nullable, peel off the wrapper
+      if var_type.kind == GraphQL::TypeKinds::NON_NULL && arg_type.kind != GraphQL::TypeKinds::NON_NULL
```
rmosolgo: Can we use `.non_null?` here instead of comparing to those constants? 

I think its a bit better because it matches the surrounding code and it might be easier to change in the future. (No dependency on specific constants, just the call chain `.kind.non_null?`)

theorygeek:
Whoops... I did not mean to close this PR :P

rmosolgo:
Awesome, thanks for the fix!

PR #132, opened 2016-04-05
Field resolver context

theorygeek:
Before:
```ruby
GraphQL::Field.define do
  resolve -> (obj, args, context) {
    self # instance of GraphQL::Define::DefinedObjectProxy
  }
end
```

After:
```ruby
GraphQL::Field.define do
  resolve -> (obj, args, context) {
    self # instance of GraphQL::Field
  }
end
```


theorygeek:
NVM... It looks like this might break a lot of stuff, particularly variable access from within those resolvers. (Example: https://github.com/rmosolgo/graphql-relay-ruby/blob/master/lib/graphql/relay/mutation.rb#L78)

PR #133, opened 2016-04-07
fix(context) handle empty @values

golmansax:
Thoughts about defaulting `@values = {}` in `Query::Context`?  I ran into an issue where my query would check if `context[:some_key]` exists, but it would raise an error because `@values = nil` if no context is passed into the query.

Thanks!


rmosolgo:
Hi! Sorry I forgot to come back to this after reading it the first time! 

It makes sense to me, since Context explicitly delegates `#[]` to the inner values, inner values should respond to that method in any case.

rmosolgo:
Thanks!

golmansax:
Thanks for merging!

PR #138, opened 2016-04-14
fix(Define::AssignArgument) block definition

kliuless:
Allow Argument `description` and `default_value` to be defined in the block

### DIFF
--- a/lib/graphql/define/assign_argument.rb
+++ b/lib/graphql/define/assign_argument.rb
@@ -10,8 +10,11 @@ module GraphQL
         end
         argument.name = name.to_s
         argument.type = type
-        argument.description = description
-        argument.default_value = default_value
+
+        description && argument.description = description
+        # check nil because false is a valid value
+        default_value != nil && argument.default_value = default_value
+
         target.arguments[name.to_s] = argument
       end
     end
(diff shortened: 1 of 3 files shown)


lib/graphql/define/assign_argument.rb
```diff
@@ -10,8 +10,11 @@ def self.call(target, name, type, description = nil, default_value: nil, &block)
         end
         argument.name = name.to_s
         argument.type = type
-        argument.description = description
-        argument.default_value = default_value
+
+        description && argument.description = description
+        # check nil because false is a valid value
+        default_value != nil && argument.default_value = default_value
```
rmosolgo: would this work? 

```ruby
default_value.nil? || argument.default_value = default_value
```

kliuless: Sure, that's equivalent. Do you want me to amend my commit, since this change is small?

kliuless: Nevermind, that's a bad idea. I can create a new branch with the one commit, or just leave it as 2 commits.

rmosolgo: sorry, it doesn't really matter either way, let's just keep it like this :grinning:

rmosolgo:
thanks!

Issue #139, opened 2016-04-15
Error handling in custom scalars


eapache:
@rmosolgo it's not clear to me from poking through the code what the expected pattern is for handling errors in custom scalar formats (e.g. when I define a `Date` scalar and the input doesn't conform to the required format). Since `coerce_result` is called in `get_finished_value` there doesn't seem to be any mechanism for rescuing an error that is raised or returned. What did you have in mind?

cc @dylanahsmith

dylanahsmith:
It looks like the coerce procs expect nil to be returned for invalid values

eapache:
Except there's no code I could find to distinguish between "nil because the raw input was nil" and "nil because of an error".

dylanahsmith:
Oh, I guess there is only something for input coercion: https://github.com/rmosolgo/graphql-ruby/blob/061d0577a79976ecbb838599f7fe1464fb6ee0ee/lib/graphql/scalar_type.rb#L24

For result coercion, the error isn't really a client error anyways, so should get treated as an internal error.  I think right now that causes that causes the entire query to fail, but we could rescue it to give a partial result that just has `nil` for that field selection.

rmosolgo:
Yes, if the coerce proc returns nil, it's handled as invalid. 

I think the distinguishing code you're looking for is here: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L86

If the incoming value is `nil`, it's never passed to coerce procs, it's just left alone.

Down the road, a non-nil value is passed to the coerce proc:  https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/scalar_type.rb#L23

(Unfortunately, the values are coerced twice. It's a bit inefficent, but it's worth it for the feature (#113) and I think it will be a straightforward refactor to coerce & validate in one pass..!)

I've just pushed a commit with a spec to describe that behavior: https://github.com/rmosolgo/graphql-ruby/commit/061d0577a79976ecbb838599f7fe1464fb6ee0ee

How does that look for your case?

PR #140, opened 2016-04-15
Handle the case where type coersion results in nil

eapache:
The NonNullResolution strategy was missing handling the case where the raw value
was non-null but the inner strategy (usually scalar type coersion) would turn it
into null. This should be handled the same way as when the raw value is null.

@rmosolgo @dylanahsmith part of #139.

### DIFF
--- a/lib/graphql/query/serial_execution/value_resolution.rb
+++ b/lib/graphql/query/serial_execution/value_resolution.rb
@@ -80,7 +80,9 @@ module GraphQL
             wrapped_type = field_type.of_type
             strategy_class = get_strategy_for_kind(wrapped_type.kind)
             inner_strategy = strategy_class.new(value, wrapped_type, target, parent_type, ast_field, execution_context)
-            inner_strategy.result
+            value = inner_strategy.result
+            raise GraphQL::InvalidNullError.new(ast_field.name, nil) if value.nil?
+            value
           end
         end
 
--- a/spec/graphql/query/executor_spec.rb
+++ b/spec/graphql/query/executor_spec.rb
@@ -168,6 +168,21 @@ describe GraphQL::Query::Executor do
       end
     end
 
+    describe 'if a non-nil value coerces to nil for a non-null field' do
+      let(:query_string) {%| query noMilk { cow { name cantBeNullButIsOnCoersion } }|}
+      it 'turns into error message and nulls the entire selection' do
+        expected = {
+          "data" => { "cow" => nil },
+          "errors" => [
+            {
+              "message" => "Cannot return null for non-nullable field cantBeNullButIsOnCoersion"
+            }
+          ]
+        }
+        assert_equal(expected, result)
+      end
+    end
+
     describe "if the schema has a rescue handler" do
       before do
         schema.rescue_from(RuntimeError) { "Error was handled!" }
--- a/spec/support/dairy_app.rb
+++ b/spec/support/dairy_app.rb
@@ -2,6 +2,11 @@ require_relative './dairy_data'
 
 class NoSuchDairyError < StandardError; end
 
+AlwaysNilType = GraphQL::ScalarType.define do
+  name "AlwaysNilType"
+  coerce ->(value) { nil }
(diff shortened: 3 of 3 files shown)


lib/graphql/query/serial_execution/value_resolution.rb
```diff
@@ -80,7 +80,9 @@ def result
             wrapped_type = field_type.of_type
             strategy_class = get_strategy_for_kind(wrapped_type.kind)
             inner_strategy = strategy_class.new(value, wrapped_type, target, parent_type, ast_field, execution_context)
-            inner_strategy.result
+            value = inner_strategy.result
+            raise GraphQL::InvalidNullError.new(ast_field.name, nil) if value.nil?
```
dylanahsmith: I think this check is only needed for ScalarResolution, since ObjectResolution#non_null_result always returns a Hash,  HasPossibleTypeResolution should always delegate to ObjectResolution and ListResolution#non_null_result always returns a list.

eapache: There's no harm in doing it for the other types though I think

dylanahsmith: Well, doing it here means that we silently convert invalid values to `nil` and only report the error for non-null fields.  Doing the check in ScalarResolution#non_null_result means that we can catch those bugs.

spec/graphql/query/executor_spec.rb
```diff
@@ -168,6 +168,21 @@
...
     end
 
+    describe 'if a non-nil value coerces to nil for a non-null field' do
+      let(:query_string) {%| query noMilk { cow { name cantBeNullButIsOnCoersion } }|}
+      it 'turns into error message and nulls the entire selection' do
+        expected = {
+          "data" => { "cow" => nil },
+          "errors" => [
+            {
+              "message" => "Cannot return null for non-nullable field cantBeNullButIsOnCoersion"
```
dylanahsmith: This isn't really a helpful error for either the client or the server developers.  I would prefer if we reported these errors as internal errors.

lib/graphql/query/serial_execution/value_resolution.rb
eapache: `ScalarResolution` doesn't know if it's been wrapped in a `NonNullResolution` or not. When a field is nullable and can't be coerced correctly, returning null without an error is the correct behaviour AFAICT.

spec/graphql/query/executor_spec.rb
eapache: It's reasonably helpful to the server to know the field, and the cause of the error?

lib/graphql/query/serial_execution/value_resolution.rb
dylanahsmith: The [Scalars Result Coercion](http://facebook.github.io/graphql/#sec-Scalars) section of the spec it says:

> Since this coercion behavior is not observable to clients of the GraphQL server, the precise rules of coercion are left to the implementation. The only requirement is that the server must yield values which adhere to the expected Scalar type.

so the graphql spec doesn't say what the correct behaviour is, just that the result can be either coerced if it can be reasonably coerced or a field error must be produced.

The [Error handling](http://facebook.github.io/graphql/#sec-Error-handling) section is clear about field error resulting in the error being added to the "errors" list though

> If an error occurs when resolving a field, it should be treated as though the field returned null, and an error must be added to the "errors" list in the response.

In this case I don't think we want the type coercion to silently change semantics.  I think that should be considered a bug and reported as an internal error for fixing.  If the returning `nil` is appropriate in a field for some garbage value, then we should be explicit about that and return `nil` in the field resolver.

spec/graphql/query/executor_spec.rb
dylanahsmith: But this isn't really the cause of the error, since it sounds like the raw result was `nil` rather than an error in coercion.

eapache:
Going to rework this and #141.

PR #141, opened 2016-04-15
Improve the way scalar coercion happens

eapache:
It is legitimate for a scalar type to want to coerce nils into non-nil values
and vice-versa, in normal (non-error) cases. This means:
- calling `coerce_result` from a method called `non_null_result` is misleading
at best, since it is legitimate for `coerce_result` to return nil
- the short-circuiting that `BaseResolution#result` does on nils is incorrect
for scalars since it doesn't give the coercion an opportunity to handle that
case

Simply override `result` instead of `non_null_result` for `ScalarResolution`.
Still short-circuit when the raw value is an `ExecutionError` but otherwise
trust the coercion proc. Update the string and id scalars to pass through nils
since otherwise they turned them into `""` which is probably undesirable.

@dylanahsmith @rmosolgo

### DIFF
--- a/lib/graphql/boolean_type.rb
+++ b/lib/graphql/boolean_type.rb
@@ -5,5 +5,5 @@ GraphQL::BOOLEAN_TYPE = GraphQL::ScalarType.define do
   name "Boolean"
 
   coerce_input -> (value) { ALLOWED_INPUTS.include?(value) ? value : nil }
-  coerce_result -> (value) { !!value }
+  coerce_result -> (value) { value.nil? ? nil : !!value }
 end
--- a/lib/graphql/float_type.rb
+++ b/lib/graphql/float_type.rb
@@ -1,5 +1,5 @@
 GraphQL::FLOAT_TYPE = GraphQL::ScalarType.define do
   name "Float"
   coerce_input -> (value) { value.is_a?(Numeric) ? value.to_f : nil }
-  coerce_result -> (value) { value.to_f }
+  coerce_result -> (value) { value.nil? ? nil : value.to_f }
 end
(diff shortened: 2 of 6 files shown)


eapache:
CI failure does not look related?

rmosolgo:
Sporadic Rubinius failures ever since rewriting the parser :sob:

rmosolgo:
:+1: from me, if I understand right, the real behavior change here is that nil is passed into scalar `coerce_result` functions, where they used to be short-circuited. 

Should the other scalars' `coerce_result` procs also be updated?

eapache:
Yes, I'd seen the checks for `Numeric` and `ALLOWED_INPUTS` and assumed they were already handled, but that was only for inputs; I've updated the other results coercion methods as well.

dylanahsmith:
> It is legitimate for a scalar type to want to coerce nils into non-nil values
and vice-versa, in normal (non-error) cases

What are those cases?

For instance, I could see a desire to coerce `nil` to `false`, but that would only be desired if the type is wrapped in a non-null type.  This doesn't really allow that.

Also, when would it be appropriate for a type to coerce something to `nil`?  Would that also depend on whether it is wrapped in a non-null type.  For instance, it might be appropriate to coerce a blank string to `nil`, but that depends on whether or not it is wrapped in a non-null type.  If it is wrapped in a non-null type it would be better to coerce `nil` into `""` and if it isn't it might be better to do the opposite coercion of `""` into `nil`.

Is it really best to do this in the type coercion?  Since like these implicit coercions might be better to do in the field resolver, leaving the result coercion to mostly just validate the result.

PR #142, opened 2016-04-20
refactor(InterfaceType) determine possible_types based on the schema

rmosolgo:
Taking a cue from graphql/graphql-js#327

Previously, `InterfaceType#possible_types` was implemented as an array which was mutated for every object that implemented it. This caused bugs and resulted in complex gem code. 

So instead, determine possible types _within_ the schema. This will fix some nasty bugs, support Rails development better, and remove the lingering caveat when people ask "can you make multiple schemas". 

The downer is that it's a breaking change: any orphan type must be passed to `types:` (where previously it was discovered at load-time). 

Before shipping this change, I'll push a minor version with deprecation warnings.

### DIFF
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -50,24 +50,21 @@ module GraphQL
       # Maybe you'll need to override this in your own interfaces!
       #
       # @param object [Object] the object which needs a type to expose it
+      # @param ctx [GraphQL::Query::Context]
       # @return [GraphQL::ObjectType] the type which should expose `object`
-      def resolve_type(object)
-        instance_exec(object, &(@resolve_type_proc || DEFAULT_RESOLVE_TYPE))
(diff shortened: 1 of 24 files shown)


theorygeek:
Nice change! So it sounds like specifically, if your schema has an object type that is only used indirectly (ie, it is a possible type for an interface or union, but is never itself the declared type for any field in the schema), you need to pass it to the `types:` parameter when you call `GraphQL::Schema#new`.

Is that right?

rmosolgo:
Yes, with one clarification:

> used indirectly (ie, it is a possible type for an interface or union)

Unions are ok as-is because they access their possible types during definition, which causes Rails to load those constants.

But interfaces have the problem because they _dont_ cause their members to be loaded by Rails. Instead, those types are loaded sometime later or not at all... and mayhem ensues!

PR #149, opened 2016-05-08
More parsing tests for failure queries

gjtorikian:
While looking over https://github.com/graphql/libgraphqlparser/blob/0fb896cef9a1729ffd3e2b9bb2e33929c13799ea/test/ParserTests.cpp, I noticed that its test coverage for malicious queries is quite high. The goal of this PR is to port some of those tests over into this project for further coverage here.

I've brought the tests over and begun fixing some of the parsing issues individually. For example, before this PR, `;` was reporting an error of "end of document reached," whereas now it's the more comprehendible "Could not parse `;`", as `;` has no meaning in a query.

If you think this is a good idea I can go ahead and try to fix up some of the other gnarly ones.


rmosolgo:
👌 💯 This is great, thank you! Yes, providing a better experience for error cases is definitely an improvement. Thanks for digging in on this!

Just let me know when those other cases are worked out, or if there's anything I can do to help along the way!

gjtorikian:
@rmosolgo This is all set. I tried to keep the commits small and atomic but please let me know if I ought to go into more detail with any of the new parsing tests.

spec/graphql/language/parser_spec.rb
```diff
@@ -304,6 +310,136 @@
...
+      e = assert_raises(GraphQL::ParseError) do
+        GraphQL.parse("{ ...on }")
+      end
+
+      assert_includes(e.message, "Parse error on \"}\"")
+    end
+
+    it "rejects null value" do
+      e = assert_raises(GraphQL::ParseError) do
+        GraphQL.parse("{ fieldWithNullableStringInput(input: null) }")
```
rmosolgo: Is it _legal_ GraphQL to have an enum value named `null`?  Ok, I looked, apparently not: 

image


:+1:

lib/graphql/language/parser.y
```diff
@@ -114,6 +114,12 @@ rule
     | FALSE
     | ON
 
+  name_without_on:
+      IDENTIFIER
+    | FRAGMENT
+    | TRUE
+    | FALSE
+
```
rmosolgo: now that we have this definition, could we update `name` to be: 

```
name: 
    name_without_on 
  | ON 
```

?

lib/graphql/language/lexer.rb
```diff
@@ -653,15 +671,24 @@ def self.emit(token_name, ts, te, meta)
...
-        )
-        meta[:col] += te - ts
+        if value =~ /\\u|\\./ && value !~ ESCAPES
+          meta[:tokens] << GraphQL::Language::Token.new(
+            name: :BAD_UNICODE_ESCAPE,
+            value: value,
+            line: meta[:line],
+            col: meta[:col],
+          )
+        else
```
rmosolgo: do we need to increment the `meta[:col]` counter in this case? I see that the `else` case has `meta[:col] += te - ts` which keeps the column number on track, do we need it here too?

rmosolgo: (Sorry, this comment should be on the line above -- in the `if ... ` block)

rmosolgo:
Looks great, I just had a couple housekeeping questions, can't wait to get this merged!

lib/graphql/language/parser.y
gjtorikian: Yup.

lib/graphql/language/lexer.rb
gjtorikian: Oh, probably!

gjtorikian:
I took all the feedbacks!

rmosolgo:
cool, one last thing: could you rebuild lexer.rb after 0105a92 ? It looks like that changed the Ragel source but not the corresponding Ruby file

gjtorikian:
D'oh, sorry! Done and rebased up into b92d94c.

rmosolgo:
Awesome, thanks for this improvement!!

PR #151, opened 2016-05-13
Add GraphQL query string generation from parsed nodes

josh:
Hey @rmosolgo!

`GraphQL.parse( query_string )` is wonderful, but I was hoping to be able to serialize the parsed result back into a String. My current use case is parsing a GraphQL query, transforming the nodes, and reserializing back to a String. Kinda doing something similar to `Relay.QL` fragment composition.

I started by adding a `to_query_string` method to `Nodes` so that `GraphQL.parse(str).to_query_string` gives you *roughly* the input string back.

I'm not sure if you prefer having this serialization logic on them nodes themselves, or if it should all be moved into a `GraphQL::Language::Generator` module instead.

### DIFF
--- a/lib/graphql/language.rb
+++ b/lib/graphql/language.rb
@@ -1,3 +1,4 @@
+require "graphql/language/generation"
 require "graphql/language/lexer"
 require "graphql/language/nodes"
 require "graphql/language/parser"
--- a/lib/graphql/language/generation.rb
+++ b/lib/graphql/language/generation.rb
@@ -3,160 +3,101 @@ require "graphql/language/nodes"
 module GraphQL
   module Language
     module Nodes
-      class AbstractNode
-        def generate_value(value)
-          case value
-          when AbstractNode
-            value.to_query_string
-          when FalseClass, Float, Integer, NilClass, String, TrueClass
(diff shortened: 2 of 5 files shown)


rmosolgo:
Woah, awesome! This really opens the door to a lot of interesting uses! 

In terms of code organization, I think `document.to_query_string` is a nice public API. But like you said, it would be nice to keep code generation _distinct_ from the AST. What do you think of implementing `to_query_string` roughly like this: 

```ruby 
def to_query_string 
  GraphQL::Language::Generation.generate(self)
end 
```

?

(Besides keeping the node classes simple, it makes it easier for someone to monkey-patch in their own generator 😬 )

josh:
Still a little bit messy, but can you have a look at this new organization direction just to make sure I'm on the right track? Moves everything into a `GraphQL::Language::Generation` module.

rmosolgo:
:+1: works for me, the big `case` is a bit unusual for Ruby, but until they give us pattern matching, we'll have to take what we can get 😆  

Thanks for sharing this awesome feature, I'd love to hear how you end up using it!

josh:
Thanks!

PR #152, opened 2016-05-16
allow passing a pre parsed query to Query.new

xuorig:
Allow passing a `document` to the `Query`.

This can help building a server query cache to avoid parsing the strings everytime.

@rmosolgo 

cc @theorygeek

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -18,7 +18,9 @@ module GraphQL
     # @param debug [Boolean] if true, errors are raised, if false, errors are put in the `errors` key
     # @param validate [Boolean] if true, `query_string` will be validated with {StaticValidation::Validator}
     # @param operation_name [String] if the query string contains many operations, this is the one which should be executed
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: more flexible AND much simpler approach than #76 !

lib/graphql/query.rb
```diff
@@ -28,7 +30,9 @@ def initialize(schema, query_string, context: nil, variables: {}, debug: false,
       @fragments = {}
       @operations = {}
       @provided_variables = variables
-      @document = GraphQL.parse(query_string)
+
+      @document = document
+      @document ||= GraphQL.parse(query_string)
```
rmosolgo: Would this also work:

```ruby
@document = document || GraphQL.parse(query_string) 
```

? 

I'd prefer to avoid reassigning the variable because a later reader may find the _first_ assignment but skip the second one!

xuorig: I like it!

xuorig: changed!

PR #153, opened 2016-05-16
Add ability to modify context hash

gjtorikian:
We were noticing that `context` is not modifiable. Was this intentional? I can't think of a reason why you wouldn't want to be able to modify the context.

/cc @davidcelis

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -32,6 +32,10 @@ module GraphQL
       def [](key)
         @values[key]
       end
+
+      def []=(key, value)
+        @values[key] = value
+      end
     end
   end
 end
--- a/spec/graphql/query/context_spec.rb
+++ b/spec/graphql/query/context_spec.rb
@@ -57,4 +57,14 @@ describe GraphQL::Query::Context do
       assert_equal(nil, context[:some_key])
     end
   end
+
+  describe "assigning values" do
+    let(:context) { GraphQL::Query::Context.new(query: OpenStruct.new(schema: schema), values: nil) }
+
+    it "allows you to assign new contexts" do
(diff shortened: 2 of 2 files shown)


rmosolgo:
This is _kind of_ on purpose. I can describe my thoughts: 

1. I don't want to ever change the signature of `resolve`, because that would be really hard to update -- resolve functions are _everywhere_! But, I know that the needs of `resolve` will change over time (as we learn about how to build good graphql systems). So, I want to pass in a library object (`GraphQL::Query::Context`) which can be altered over time according to our needs.
2. Passing in some application-specific data is essential to query resolution. You must be able to authorize and localized based on the specific query. So, the context should expose some per-query data.
3. Mutability is a pain. Since fields may be resolved in basically _any_ order, managing a stateful context object would be _very_ hard. Imagine you get a trusty ol' `NoMethodError on nil:NilClass` relating to `context` -- to track down this error, you must trace the specific mutations on `context` during that query. Yuck. 
4. Mutability _may_ be necessary. I don't have a use for it, but I'm sure someone will have a use for it. 

So in the end, making `context` read-only was my cop-out. If you want to put state in your context, you can do it. Any of `context`'s keys may be mutable (because Ruby!), so you could have a mutable object "hanging off of" `context`. For example: 

```ruby 
# some resolve proc, mutate a member of `context`:
-> (obj, args, ctx) { ctx[:my_array] << 1 } 

# at query-time, add a mutable object to `context`: 
MySchema.execute(query_string, context: { my_array: [] })
```

This way, an application _may_ mutate application-specific data, but the library doesn't have to support it directly! 

What do you think?

gjtorikian:
Legit. Thanks for the careful write-up!

PR #155, opened 2016-05-17
Better errors when resolve_type fails

theorygeek:
Improves the error message when you have a field that returns an interface or a union, but the field's resolver returns a value that can't be resolved to one of the possible types.

Previously, it would fail with a somewhat cryptic error: 
>undefined method `kind' for nil:NilClass

I'm not sure the best way to construct a test case for this; @rmosolgo any ideas?

### DIFF
--- a/lib/graphql/object_type.rb
+++ b/lib/graphql/object_type.rb
@@ -59,5 +59,26 @@ module GraphQL
         memo.merge!(iface.fields)
       end
     end
+
+
+    # Error raised when the value provided for a field can't be resolved to one of the possible types
+    # for the field.
+    class UnresolvedTypeError < GraphQL::Error
+      attr_reader :field_name, :field_type, :parent_type
+
+      def initialize(field_name, field_type, parent_type)
+        @field_name = field_name
+        @field_type = field_type
+        @parent_type = parent_type
+        super(exception_message)
+      end
+
+      private
+
+      def exception_message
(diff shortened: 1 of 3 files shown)


rmosolgo:
Nice! 

How about adding a dedicated error class like `UnresolvedTypeError`? 

Then I think you could test it by calling `resolve_type` directly: 

```ruby
assert_raises(GraphQL::ObjectType::UnresolvedTypeError) { SomeInterfaceType.resolve_type(:nonsense) }
```

By the way, would it work to change the `.nil?` check to `.is_a?(GraphQL::ObjectType)`? If I understand correctly, that should _always_ return an object type, right?

theorygeek:
I moved the error into a separate class, I also shortened the message a little by removing the value from the message.

Also, I tried testing it with by calling `resolve_type` directly, but the problem is that the error actually happens inside of `ValueResolution`. So I added a test to `value_resolution_spec.rb` instead. Let me know what you think.

rmosolgo:
So to review so late! 

There are a few other uses of `resolve_type`, but I don't think they're covered by this patch: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/type_resolver.rb#L11-L16 Is that right? 

Oh, but now that I'm writing it out, I think I understand the complexity better: in _those_ cases, returning `nil` is appropriate: it's a way of saying "no, this object _can't_ be resolved to a child type".

👌

PR #156, opened 2016-05-18
Deep freeze nodes

josh:
`Node#freeze` freezes children as well so the entire parse tree may be treated as immutable.

An alternative API maybe to add a separate `Node#deep_freeze` if we still wanted to support shallow freezes.

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -48,7 +48,7 @@ module GraphQL
         end
 
         def freeze
-          children.each(&:freeze)
+          self.class.child_attributes.each { |attr_name| public_send(attr_name).each(&:freeze).freeze }
           super
         end
       end
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -46,6 +46,11 @@ module GraphQL
         def to_query_string
           Generation.generate(self)
         end
+
+        def freeze
+          children.each(&:freeze)
+          super
+        end
       end
 
       class WrapperType < AbstractNode
--- a/spec/graphql/language/parser_spec.rb
+++ b/spec/graphql/language/parser_spec.rb
@@ -31,6 +31,16 @@ describe GraphQL::Language::Parser do
       assert document
     end
 
+    it "freeze deep freezes child nodes" do
+      refute document.frozen?
+      refute document.definitions.first.frozen?
+
+      document.freeze
+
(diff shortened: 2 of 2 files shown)


rmosolgo:
Interesting, I'm all for immutability, but what's the goal here? 

I don't know a ton about Ruby's `#freeze` but it doesn't seem to do much for us: 

```ruby 
f = GraphQL::Language::Nodes::Field.new
f.name = "myField"
f.freeze
f.name << "WithModifications"
f.name
# => "myFieldWithModifications"
# 😿
```

rmosolgo:
I guess what i mean is: 

- for sure, any "end user" should treat the AST as read-only 
- do we need-_need_ to enforce that? 
- if we _do_ need to enforce it, how could we?

josh:
> Interesting, I'm all for immutability, but what's the goal here?

I'm storing parsed queries as static constants and want to ensure they aren't modified during runtime.

> I don't know a ton about Ruby's #freeze but it doesn't seem to do much for us:

The base implementation of `Object#freeze` only freezes the instance not anything nested since it doesn't understand the relation between other ivars. So by default, it just prevents `node.name = "something else" # raises an error`. So if you want full immutability, you need to implement it yourself in your class.

D'oh, `children` isn't the actual object we're storing these in. We should probably freeze those arrays too.

> for sure, any "end user" should treat the AST as read-only

I think it should be opt in. This doesn't change anything about the result of `GraphQL.parse("...")`. But if you want make the result immutable, you'd explicitly call freeze.

Its common to freeze constants in Ruby, you usually see patterns like.

``` ruby
module Foo
  BAR = "bar".freeze
end
```

rmosolgo:
Thank you for this suggestion. I've been thinking about it a lot, but I think I have to let it go. I agree that unwitting mutation of the AST would be bad news, but I think we have to rely on well-behaved users in this case.

From a library point of view, it's not needed. The library itself doesn't need to freeze these nodes. If a user wants to use them, then it's up to them whether to mutate them. 

From a defensive coding point of view, it's not _100%_ safe: although the properties can't be reassigned, their values are still mutable. That's Ruby for us :P 

Like you said, users should have the freedom to create valid ASTs. But when it comes to modification, I think we'll rely on good behavior. ("Don't modify things that don't belong to you", can we call it common sense?!)

Fortunately, I think you could implement this behavior with public APIs, so if you need it for your app, it's all yours!

PR #157, opened 2016-05-24
feat(Query): Allow an object for the root type to be specified.

dylanahsmith:
## Problem

There is no way to specify an object to the root query or mutation types. This is something that is supported in graphql-js with a [rootValue option](https://github.com/graphql/graphql-js/blob/v0.6.0/src/graphql.js#L32).

Having a root value can be useful since the default resolving calls methods on the object and there may be a natural root object.

## Solution

Add a `root_value` optional argument when constructing the query which the operation resolution can use for the object of the root type.

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -8,7 +8,7 @@ module GraphQL
       end
     end
 
-    attr_reader :schema, :document, :context, :fragments, :operations, :debug, :max_depth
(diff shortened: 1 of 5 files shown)


rmosolgo:
👌  awesome! I had ignored this feature from graphql-js, thinking any field on `Query` may be looked up from the global namespace, but this is a much better pattern in general!

PR #159, opened 2016-06-03
Make EnumType conform to input coercion

eapache:
It is currently required that `coerce_non_null_input` return `nil` for invalid
inputs to signal an error. `EnumType` was raising a `KeyError` instead.

@rmosolgo @dylanahsmith cc @lreeves

### DIFF
--- a/lib/graphql/enum_type.rb
+++ b/lib/graphql/enum_type.rb
@@ -67,6 +67,7 @@ module GraphQL
     # @param value_name [String] the string representation of this enum value
     # @return [Object] the underlying value for this enum value
     def coerce_non_null_input(value_name)
+      return nil unless @values_by_name.key?(value_name)
       @values_by_name.fetch(value_name).value
     end
 
--- a/spec/graphql/enum_type_spec.rb
+++ b/spec/graphql/enum_type_spec.rb
@@ -8,6 +8,10 @@ describe GraphQL::EnumType do
     assert_equal(1, enum.coerce_input("COW"))
   end
 
+  it "coerces invalid names to nil" do
+    assert_equal(nil, enum.coerce_input("YAKKITY"))
+  end
+
   it "coerces result values to value's value" do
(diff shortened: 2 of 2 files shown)


lib/graphql/enum_type.rb
```diff
@@ -67,6 +67,7 @@ def validate_non_null_input(value_name)
     # @param value_name [String] the string representation of this enum value
     # @return [Object] the underlying value for this enum value
     def coerce_non_null_input(value_name)
+      return nil unless @values_by_name.keys.include?(value_name)
       @values_by_name.fetch(value_name).value
```
dylanahsmith: How about using `@values_by_name[value_name]` instead of explicitly checking if the key is included.  That will be more efficient than creating a keys array then iterating over it to see if the value name is included.

eapache: I did that with a `try!(:value)` and then realized that `try!` is rails-specific; I suppose I could guard it with a ternary.

rmosolgo: :+1: or provide `nil` as `fetch`'s fallback, `@values_by_name.fetch(value_name, nil).value`

rmosolgo: ohhh `.value` ... how about `.key?(value_name)`

eapache: 👍 I'll use `.key?`

dylanahsmith: I was thinking
```ruby
enum_value = @values_by_name[value_name]
enum_value && enum_value.value
```

dylanahsmith:
:+1:

lib/graphql/enum_type.rb
eapache: https://en.wikipedia.org/wiki/There%27s_more_than_one_way_to_do_it

Sometimes I wish we were using python :trollface:

rmosolgo:
Oops, thanks for the fix!

PR #162, opened 2016-06-06
Automatically coerce default inputs

eapache:
The schema now validates that `default_value` values are valid (they pass
through `coerce_non_null_input`) however it wasn't actually coercing them before
passing them to resolve blocks.

For the majority of types this is not a problem (e.g. 42 is a valid int and
works just fine in the context of a resolve block). However, for enums this
required that the default value be in the set of *names* but then was passing
the name unconverted into the resolve block, where the resolve block was
expecting the equivalent *value*.

The fix is to coerce default values before use, the same way we might coerce any
other input literal.

@rmosolgo @dylanahsmith cc @lreeves

### DIFF
--- a/lib/graphql/query/literal_input.rb
+++ b/lib/graphql/query/literal_input.rb
@@ -21,7 +21,7 @@ module GraphQL
             arg_value = coerce(arg_defn.type, ast_arg.value, variables)
           end
           if arg_value.nil?
-            arg_value = arg_defn.default_value
+            arg_value = arg_defn.type.coerce_input(arg_defn.default_value)
           end
           values_hash[arg_name] = arg_value
         end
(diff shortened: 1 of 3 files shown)


lib/graphql/query/literal_input.rb
```diff
@@ -21,7 +21,7 @@ def self.from_arguments(ast_arguments, argument_defns, variables)
             arg_value = coerce(arg_defn.type, ast_arg.value, variables)
           end
           if arg_value.nil?
-            arg_value = arg_defn.default_value
+            arg_value = arg_defn.type.coerce_input(arg_defn.default_value)
```
dylanahsmith: Can we just assign `arg_value = arg_defn.default_value` instead of `arg_value = nil` then avoid this extra condition?

eapache: No; `arg_defn.default_value` is not an AST node instance so calling `LiteralInput.coerce` on it will crash. That's why I'm calling `arg_defn.type.coerce_input` directly here.

dylanahsmith:
:+1:

lreeves:
Good catch, 👍

rmosolgo:
Thanks, good find!

PR #164, opened 2016-06-09
fix(Directive) add previous introspection fields as deprecated

rmosolgo:
Oops, removing them altogether broke some versions of GraphiQL!


maxcodes:
Thanks! Sorry for delay in replying. I found that GraphiQL is broken right now anyway  (see https://github.com/graphql/graphiql/pull/144), so I can't quite use this yet. Still, good to have and thanks for fast fix :)

PR #170, opened 2016-06-17
Allows for one or more possible_types in a union

bswinnerton:
This pull request modifies the validation rules to allow for one or more possible types inside of a union. As mentioned [here](https://goo.gl/Mr0jvQ), a previous version of the GraphQL specification asserted:

> A Union type must define two or more member types.

But has since been [changed](https://git.io/vouAt) to:

> A Union type must define one or more member types.

It appears that the tests and error messages are both already crafted in such a way to support this change. For example, previous to this pull request an error message would read:

```
Field Star.starrable is invalid: must have at least one possible type
```

When at least one possible type was specified. Similarly the validation test also has already been written in such a way to support this change:

```ruby
it "requires at least one possible_types" do
```


rmosolgo:
Thanks! I must've read that change with _half_ my brain ...

PR #173, opened 2016-06-29
Support #to_proc-style field definitions

rmosolgo:
One shortcoming of the current API is defining _whole_ fields programmatically. It's easy to parameterize `resolve` functions, but name, return type and arguments must be specified separately. 

This approach _allows_ you to do that -- to define the whole field in one place, but it's not a very good API ... who thought `to_proc` would be the way to define a field? 

So, this PR is mostly a "thought experiment", trying to find a good way to parameterize _whole_ field definitions

### DIFF
--- a/lib/graphql/define/assign_object_field.rb
+++ b/lib/graphql/define/assign_object_field.rb
@@ -2,7 +2,7 @@ module GraphQL
   module Define
     # Turn field configs into a {GraphQL::Field} and attach it to a {GraphQL::ObjectType} or {GraphQL::InterfaceType}
     module AssignObjectField
-      def self.call(fields_type, name, type = nil, desc = nil, field: nil, deprecation_reason: nil, property: nil, &block)
+      def self.call(fields_type, name = nil, type = nil, desc = nil, field: nil, deprecation_reason: nil, property: nil, &block)
         if block_given?
           field = GraphQL::Field.define(&block)
         else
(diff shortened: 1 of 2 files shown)


theorygeek:
Perhaps another approach could be to change `AssignObjectField::call` so that it checks if the first parameter `.is_a?(GraphQL::Field)`. 

Here's a simple example, for creating simple fields that just access an attribute on a model. It camelizes the name, and figures out the appropriate GraphQL type:
```ruby
module ModelFieldFactory
  def self.generate(model, attribute)
    GraphQL::Field.define do
      name(attribute.to_s.camelize(:lower))
      type(ModelFieldFactory.graphql_type(model, attribute))
      resolve ->(model, args, context) { model.public_send(attribute) }
    end
  end

  def self.graphql_type(model, attribute)
    # check database column type, etc.
  end
end
```

Inside of your GraphQL definition, then you could just call it:
```ruby
GraphQL::ObjectType.define do
  field ModelFieldFactory.generate(MyModelClass, :some_attribute)
end
```

The change might be as simple as:
```ruby
module GraphQL
  module Define
    module AssignObjectField
      def self.call(fields_type, name = nil, type = nil, desc = nil, field: nil, deprecation_reason: nil, property: nil, &block)
        if fields_type.is_a?(GraphQL::Field)
          field = fields_type
        elsif block_given?
          field = GraphQL::Field.define(&block)
        else
          field ||= GraphQL::Field.new
        end
```

rmosolgo:
That's nice, then we could remove the `field:` keyword argument too

PR #177, opened 2016-07-08
Track validation results

gjtorikian:
Internally, we have no idea (really!) what a healthy `max_depth` value ought to be for our schema. We'd like to track and graph the query depths which people are using.

This PR proposes adding a new hash to a schema called `validation_results`. Any validation rules can push key/value pairs to `context.results`. After the validations are done, users can call `schema.validation_results` to inspect that arbitrary data.

In this case, I've added a new key called `depth_reached`. After a query has finished, we can inspect `depth_reached` and log it. We'll use real-world results to determine if our `max_allowed_depth` is too low or too high.

Thoughts on this much appreciated. 

/cc @kdaigle

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -16,7 +16,7 @@ module GraphQL
     DIRECTIVES = [GraphQL::Directive::SkipDirective, GraphQL::Directive::IncludeDirective]
     DYNAMIC_FIELDS = ["__type", "__typename", "__schema"]
 
-    attr_reader :query, :mutation, :subscription, :directives, :static_validator
+    attr_reader :query, :mutation, :subscription, :directives, :static_validator, :validation_results
     attr_accessor :max_depth
     # Override these if you don't want the default executor:
     attr_accessor :query_execution_strategy,
@@ -41,6 +41,7 @@ module GraphQL
       @static_validator = GraphQL::StaticValidation::Validator.new(schema: self)
       @rescue_middleware = GraphQL::Schema::RescueMiddleware.new
       @middleware = [@rescue_middleware]
+      @validation_results = {}
       # Default to the built-in execution strategy:
       self.query_execution_strategy = GraphQL::Query::SerialExecution
       self.mutation_execution_strategy = GraphQL::Query::SerialExecution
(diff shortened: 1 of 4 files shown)


rmosolgo:
This seems great -- we really need more flexibility in the validation phase. Can I put this on hold for the weekend? I want to steal the [Query Reducers](http://sangria-graphql.org/learn/#query-reducers) feature from Sangria this weekend and I think this will be a great fit.

gjtorikian:
Sounds good to me!

kdaigle:
> I want to steal the Query Reducers feature from Sangria this weekend and I think this will be a great fit.

😍 So good!

rmosolgo:
Here's my approach, do you think it would work? https://github.com/rmosolgo/graphql-ruby/blob/query-reducers/guides/security.md#prevent-deeply-nested-queries

My hope is that you could do anything with the resulting value: 

```ruby 
MySchema.query_reducers << GraphQL::Analysis::QueryDepth.new do |depth| 
 write_to_database(value) 
 write_to_logs(value)
 post_to_external_service(value) 
 # ... 
end 
```

Hmm, would it need any extra info in that block? Maybe I should pass the `Query` along ...

gjtorikian:
Yeah, I think passing along query and/or context (if possible) would be fantastic.

If I can be pedantic (sorry), the naming seems a bit weird, though. For example, the field on the schema is called `query_reducers`, and the default reduction is called `QueryDepth`. 

I've just glanced at the code, but it seems like calling it `MySchema.analyzers` would make more sense, especially since you can then construct and append any kind of `GraphQL::Analaysis::*`. The example you've provided seems to confirm that: by logging the depth of the query, you're not really "reducing" anything. You _could_, but it doesn't seem like it's just limited to that kind of work. 

Either way I think this is 🌈 s!

rmosolgo:
Cool, I made an update to #180 to include the query, that way you can get to most anything. 

Hehe, I chose the name _reducer_ because Sangria had it, I think he chose it because it's traversing the AST and producing a single value, a bit like `Array#reduce` in Ruby. I'll think about the naming some more, it would be good for them to match, analysis and reduction, that is.

gjtorikian:
Correct me if I'm wrong but I think #180 vastly supersedes this. Calling it closed. 💀

PR #178, opened 2016-07-08
Mutable context

rmosolgo:
Thanks @gjtorikian for these commits from #153 ... sorry, I wasn't ready to admit my need for them then :P 

I want to write to context from a middleware, in a way the the individual query doesn't have to "worry" about. This runs the risk of key conflicts ... so be it!


gjtorikian:
:tada:

PR #179, opened 2016-07-08
feat(Schema) add TimeoutMiddleware for terminating long executions

rmosolgo:
This will stop executing a query after `max_seconds`. Note the caveat: it doesn't interrupt a `resolve` function, it only checks the time _before_ running the next `resolve` function. As described in the docs for this feature, Ruby's option is `Timeout`, but I've heard quite bad things about it: www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/

Thanks to [ZeeMee](https://zeemee.engineering/) for suggesting and sponsoring this feature 😄


kdaigle:
cc @gjtorikian since we were just looking at this. Possible we could lend a solution here that catches the timeout in line?

rmosolgo:
Sure, I'd love to see another approach!

rmosolgo:
⏰ ?

lib/graphql.rb
```diff
@@ -29,8 +29,10 @@ def self.parse_with_racc(string)
   end
 end
 
+
```
gjtorikian: 💀

lib/graphql/schema/timeout_middleware.rb
```diff
@@ -0,0 +1,63 @@
...
+    #
+    # @example Stop resolving fields after 2 seconds
+    #   MySchema.middleware << GraphQL::Schema::TimeoutMiddleware.new(max_seconds: 2)
+    #
+    class TimeoutMiddleware
+      # This key is used for storing timeout data in the {Query::Context} instance
+      DEFAULT_CONTEXT_KEY = :__timeout_at__
+      # @param max_seconds [Numeric] how many seconds the query should be allowed to resolve new fields
+      # @param context_key [Symbol] what key should be used to read and write to the query context
+      def initialize(max_seconds:, context_key: DEFAULT_CONTEXT_KEY)
```
gjtorikian: `max_seconds ` should almost certainly not be kwarg, or at least, it should have a default value if your prefer it like this. On line 27 below, `Time.now + nil` would blow the world up.

lib/graphql.rb
rmosolgo: 😬

lib/graphql/schema/timeout_middleware.rb
rmosolgo: As it is, a user _must_ pass a value, otherwise :boom: :

```
irb(main):001:0> GraphQL::Schema::TimeoutMiddleware.new
ArgumentError: missing keyword: max_seconds
	from /Users/rmosolgo/projects/graphql/lib/graphql/schema/timeout_middleware.rb:21:in `initialize'
```

Is that the case you're concerned about here? 

Of course, it's still possible to pass nil as 

```ruby 
GraphQL::Schema::TimeoutMiddleware.new(max_seconds: nil)
```

But I'm not sure of a method signature that would prevent passing `nil`. If you think it's important to handle this kind of usage, we could add runtime check!

gjtorikian:
Sorry, I completely spaced on a response here! 

I don't have any suggestions for alternate implementations at the moment. However, if possible, I would really like to see a way to capture the timing of the field resolution. It goes along with a similar statement I made in https://github.com/rmosolgo/graphql-ruby/pull/177: I don't know what a good `max_seconds` value looks like, so I'd rather sit back and collect timing info for all the resolutions. 

One suggestion for that would in fact to be to allow a `nil` `max_seconds` (which goes against what I wrote above, but sit tight! 😁 ). If that were the case, then `call` could look like this:

``` ruby
def call(parent_type, parent_object, field_definition, field_args, query_context, next_middleware)
  timeout_at = query_context[@context_key] ||= Time.now + @max_seconds
  if timeout_at < Time.now
    on_timeout(parent_type, parent_object, field_definition, field_args, query_context)
  else
    after_resolution(timeout_at, parent_type, parent_object, field_definition, field_args, query_context)
    next_middleware.call
  end
end

def after_resolution(timeout_at, parent_type, parent_object, field_definition, field_args, query_context)
end
```

`after_resolution` could be a dummy method (as with `on_timeout`) that can be overridden to allow users to do reporting (or whatever!) with that timing information.

rmosolgo:
Ohh thanks, yes, I understand better now! 

> catches the timeout in line

At first I thought this was "catch" like "rescue an error" but now I see it's "catch" like "capture a value"!

rmosolgo:
For timing individual fields, something like this would do: https://gist.github.com/rmosolgo/2b9355222f9a1a53d1fd95ce1cc6b86b

But for timing the _whole_ execution, I realize you kinda need a hack right now:

```ruby
class TimeTheEntireQueryMiddleware 
  def call(...) 
    is_first_middleware = !context.key?(CONTEXT_KEY)
    if is_first_middleware
      started_at = Time.now 
    end 

    next_middleware.call

    if is_first_middleware 
      ended_at = Time.now 
      log_entire_query_duration(ended_at - started_at)
    end 
  end 
end 
```

For query analyzers, I added `initial_value` and `final_value` hooks, I wonder if that should be supported for middlewares too ... 

I'm going to merge this one as is for now, but let's add a proper timing middleware next...

rmosolgo:
Slight API change: it still has `#on_timeout`, but you can also pass a block to log the error 

```ruby
TimeoutMiddleware.new(max_seconds: 2) { |err, query| ... }
```

PR #180, opened 2016-07-08
Query Analyzers, Max Complexity, Internal Representation

rmosolgo:
## Internal Representation 

Previously, fragments were executed by fetching AST nodes from the top level during execution. This PR adds a transformation step which is performed during validation. The result is a tree of `InternalRepresentation::Node`s which include _all_ selections on an object (including merged-in fragments), the types they apply to (eg, overlapping fragment conditions), which directives affect them (eg, directives from inline fragments apply to the fields inside the fragment), and what type they return. 

So, a query has the following steps: 

- Parse the string into AST 
- Validate AST, meanwhile building IRep tree 
- If there are analyzers to run, analyze the IRep tree 
- Execute the IRep tree
- Return the response 

This had some nice effects on the code base:

- No need to track the type stack during query analysis since the tree has type info "baked in" 
- No need to resolve fragments or merge fields during execution since that's done ahead of time 
- Query analysis and execution work from the same data so it's easy to make query complexity _right_ 

Besides this, I think this will enable more sophisticated "look-ahead" for optimizing database access. 

This has some breaking-ish changes: if you were monkey-patching `SerialExecution` in a way that depended on `attr_reader :ast_node`, this is broken, you should use `irep_node.ast_node` instead. (`graphql-batch` didn't seem broken in my test, see comment below.) `ctx.ast_node` inside resolve functions __is__ supported.

## Query Analyzers 

This is a feature stolen from Sangria (http://sangria-graphql.org/learn/#query-reducers), used to implement another feature stolen from Sangria (http://sangria-graphql.org/learn/#query-complexity-analysis). 😬 

Like middleware for validation, this provides some great flexibility in the validation phase. 

I've moved over `max_depth` and implemented `max_complexity` on top of it. It could also replace `StaticValidation` (which would be _one_ query reducer?). I think that's necessary right now, because otherwise, you're making a double-visit of the AST before running the query.

## `max_complexity` 

This is a nice way to accept _some but not all_ arbitrary queries from the outside world. 

I've also added a "security" guide that combines the options I'm aware of, did I miss anything? 

## Thanks 

Thanks [ZeeMee](https://zeemee.engineering/) for sponsoring this feature! :tada:


rmosolgo:
Migrating type checking to use this flow will be harder than I thought. This branch adds `follow_fragments` to the visitor, which is great for counting depth and complexity, but hard for type checking, because you end up visiting fragment definitions more than once. 

I think that migration will have to come later! I want to do it eventually, because while working on the rewrite, I did find some maybe-bugs or places that could be improved ... There's also a difference between type-checking and query reduction which I didn't notice before: query reduction requires an _actual_ query, with variable values present, so that you can use `args[:limit]`, etc, but type checking doesn't require actual variables. (StaticValidation::Validator _does_ take a query as input, but as far as i can tell, it's just a way of getting the schema and document.)

I did think of one more thing for this branch though: when calculating a query's complexity, we can probably rule out some mutually exclusive query branches. For example, 

```graphql
node(id: $id) { 
  ... on Type1 { 
    # ... 
  } 
  ... on Type2 { 
    # ... 
  } 
}
```

If `Type1` & `Type2` are both object types, we know that the query _can't_ run both branches. 

There's probably more you could figure out with Unions and Interfaces but ... I can't grok it yet!

rmosolgo:
This has an outstanding issue which causes artificially-high scores. Consider a case like this: 

```graphql 
{
  person(id: 1) {
    givenName 

    ... on Nameable {
      givenName
    }

   ... personFields
  }
}

fragment personFields on Person {
  givenName
}
```

In this branch, we'd give it a complexity of 4, one for each of: `person`, `givenName`, `givenName`, `givenName`. However, this doesn't match execution. When execution comes around, only two fields will be _actually_ evaluated: `person` and `firstName` (one time).

The problem is, field de-duplication is [baked into execution](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/selection_resolution.rb#L15-L19). At that point in the query, the problem is simple enough: figure out which fragments apply to the underlying object and merge them. 

However, before we we have the underlying object, it's harder. Consider this query: 

```graphql 
{
  node(id: $nodeId) {
    id

    ... on Person { 
      givenName 
      handedness 
    }
 
    ... on Nameable { 
      givenName 
      familyName
    }

    ... on Pet {
      species 
    }

    ... on Gemstone { 
      hardness 
    } 
  }
}
```

During analysis, we don't know whether the `Node` will be a Person, Pet or Gemstone. So, merging the fields is a different beast at this stage. It involves inspecting the Object Types and figuring out where the interface type fields  can be rewritten. To finish the example, I think the rewrite could be: 

```graphql 
{
  node(id: $nodeId) {
    id

    ... on Person { 
      givenName 
      familyName
      handedness 
    }

    ... on Pet {
      givenName 
      familyName
      species 
    }

    ... on Gemstone { 
      hardness 
    } 
  }
}
```

But, how do you capture this transformation and transfer it to later steps in execution? Should you munge the AST and continue sending the AST down the line? Or should you use the AST to derive an "instruction set", some _other_ tree-like structure which contains these "optimizations" ?

AST manipulation would have an easier migration path. Execution could stay the same. But I'm worried about using AST nodes as the "internal representation" of the query because they're likely to become bloated. For example, we would like to cache the resolved argument values between phases, but where would that go? Would we extend some AST node classes to store resolved argument values? That doesn't seem quite right! 

Creating an "instruction set" representation seems _right_ from a design perspective but requires a lot of new code and it might be a breaking change for graphql-batch which is not good.

😰

rmosolgo:
🎉  doesn't seem to have broken graphql-batch 

```
~/projects/graphql-batch $ git log -n 1
commit e86aa61f1ecfd8152850dd78cb3dd08669ffaffe
Author: Dylan Thacker-Smith <Dylan.Smith@shopify.com>
Date:   Wed Jul 6 12:13:13 2016 -0400

    Release version 0.2.4
~/projects/graphql-batch $ bundle show graphql
/Users/rmosolgo/projects/graphql
~/projects/graphql-batch $ bundle exec rake test
Run options: --seed 20

# Running:

.............................

Finished in 0.022581s, 1284.2724 runs/s, 1771.4102 assertions/s.

29 runs, 40 assertions, 0 failures, 0 errors, 0 skips
```

rmosolgo:
Last checkup: what's the performance overhead for adding `InternalRepresentation`? Here's a simple benchmark: 

```ruby 
require "graphql"
require "./spec/support/dairy_app"
require "benchmark/ips"

document = GraphQL.parse(GraphQL::Introspection::INTROSPECTION_QUERY)
message = "execute query: #{GraphQL::VERSION} #{`git br | grep '*'`}"
Benchmark.ips do |x|
  x.report(message) { DummySchema.execute(document: document) }
  x.compare!
end
```

```
Already on 'master'
~/code/graphql $ ruby -Ilib query_benchmark.rb
execute query: 0.15.3 * master

                         2.000  i/100ms
-------------------------------------------------
execute query: 0.15.3 * master

                         29.390  (± 6.8%) i/s -    148.000
~/code/graphql $ ruby -Ilib query_benchmark.rb
execute query: 0.15.3 * master

                         2.000  i/100ms
-------------------------------------------------
execute query: 0.15.3 * master

                         29.822  (± 3.4%) i/s -    150.000
~/code/graphql $ ruby -Ilib query_benchmark.rb
execute query: 0.15.3 * master

                         3.000  i/100ms
-------------------------------------------------
execute query: 0.15.3 * master

                         30.623  (± 3.3%) i/s -    153.000
~/code/graphql $ git co query-reducers
Switched to branch 'query-reducers'
~/code/graphql $ ruby -Ilib query_benchmark.rb
execute query: 0.15.3 * query-reducers

                         3.000  i/100ms
-------------------------------------------------
execute query: 0.15.3 * query-reducers

                         28.761  (±10.4%) i/s -    144.000
~/code/graphql $ ruby -Ilib query_benchmark.rb
execute query: 0.15.3 * query-reducers

                         3.000  i/100ms
-------------------------------------------------
execute query: 0.15.3 * query-reducers

                         29.790  (± 6.7%) i/s -    150.000
~/code/graphql $ ruby -Ilib query_benchmark.rb
execute query: 0.15.3 * query-reducers

                         2.000  i/100ms
-------------------------------------------------
execute query: 0.15.3 * query-reducers

                         29.450  (± 3.4%) i/s -    148.000
```

Seems like for this query the speed is about the same :)

PR #181, opened 2016-07-15
Add documentation for the scalar types

gjtorikian:
Pretty much what the title says.

### DIFF
--- a/lib/graphql/boolean_type.rb
+++ b/lib/graphql/boolean_type.rb
@@ -3,6 +3,7 @@ GraphQL::BOOLEAN_TYPE = GraphQL::ScalarType.define do
   ALLOWED_INPUTS = [true, false]
 
   name "Boolean"
+  description" The `Boolean` scalar type represents `true` or `false`."
 
   coerce_input -> (value) { ALLOWED_INPUTS.include?(value) ? value : nil }
   coerce_result -> (value) { !!value }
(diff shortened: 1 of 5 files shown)


lib/graphql/boolean_type.rb
```diff
@@ -3,6 +3,7 @@
   ALLOWED_INPUTS = [true, false]
 
   name "Boolean"
+  description" The `Boolean` scalar type represents `true` or `false`."
```
rmosolgo: oops, leading whitespace!

lib/graphql/float_type.rb
```diff
@@ -1,5 +1,7 @@
 GraphQL::FLOAT_TYPE = GraphQL::ScalarType.define do
   name "Float"
+  description "The `Float` scalar type represents signed double-precision fractional values as specified by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point)."
```
rmosolgo: Do these need to start with `"The {x} scalar type represents..."` Could this one just be 

```ruby 

```ruby
description "signed, double-precision fractional value, as specified by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point)"
```

?

gjtorikian: Yeah, I like that. I'll make the change across the types. 👍

rmosolgo:
cool, thanks!

PR #182, opened 2016-07-15
feat(InstanceDefinable) lazy-eval the .define block to handle circular dependencies

rmosolgo:
Previously, self-referencing or mutually-dependent types required a proc to make references: 

```ruby 
PostType = GraphQL::ObjectType.define do 
  # ...
  field :similarPosts, -> { types[PostType] }
end 
```

Now, you can leave of the proc wrapper: 

```ruby 
PostType = GraphQL::ObjectType.define do 
  # ...
  field :similarPosts, types[PostType]
end 
```

The `.define { ... }` block won't be called until one of the defined values is accessed, so `PostType` will be available by then. `-> { }`-style type references are still supported in case you _really_ need to put off calculation until Schema build-time!

### DIFF
--- a/guides/defining_your_schema.md
+++ b/guides/defining_your_schema.md
@@ -27,12 +27,7 @@ CityType = ObjectType.define do
 
   # This returns a list of `PersonType`s
   field :mayors, types[PersonType]
-
-  # To avoid circular dependencies, pass a String or a Proc for the type.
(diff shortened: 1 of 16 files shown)


brandonblack:
Yaaaassssss. Thanks for this. ✨

PR #184, opened 2016-07-18
Bug: Enum value == 'on'

theorygeek:
This PR adds a failing test case. When you define an enum that has a value `'on'`, and then you use it as a field argument:
```graphql
query {
  animals(animalType: on)
}
```

You get a parse error:
```
GraphQL::ParseError:         GraphQL::ParseError: Parse error on "on" (ON) at [3, 29]
            parser.y:265:in `on_error'
            /Users/theorygeek/.rvm/gems/ruby-2.3.0@goco.io/gems/racc-1.4.14/lib/racc/parser.rb:271:in `_racc_do_parse_c'
            /Users/theorygeek/.rvm/gems/ruby-2.3.0@goco.io/gems/racc-1.4.14/lib/racc/parser.rb:271:in `do_parse'
            parser.y:233:in `parse_document'
            parser.y:239:in `parse'
            /Users/theorygeek/goco/graphql-ruby/lib/graphql.rb:28:in `parse_with_racc'
            /Users/theorygeek/goco/graphql-ruby/lib/graphql.rb:24:in `parse'
            /Users/theorygeek/goco/graphql-ruby/lib/graphql/query.rb:45:in `initialize'
            /Users/theorygeek/goco/graphql-ruby/spec/graphql/query_spec.rb:37:in `new'
            /Users/theorygeek/goco/graphql-ruby/spec/graphql/query_spec.rb:37:in `block (2 levels) in <top (required)>'
```

However, it succeeds if you pass it as a variable, rather than as a literal:
```graphql
query($var: DairyAnimal) {
  animals(animalType: $var)
}
```

Thoughts on this? I'm guessing that the parser is confusing this enum literal with the `fragment on SomeType` syntax. 

As best I can tell, `on` is a [legal value for an enum](https://facebook.github.io/graphql/#sec-Enum-Value):
image
image


rmosolgo:
Derp, the old parser even had a comment `TODO: handle "on"` 😬 

I think the new one could handle this by switching `enum_value: IDENTIFIER ...` for `enum_value: name ...` here: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/language/parser.y#L172

IIRC `IDENTIFIER` is a "catch-all" in the lexer, but `name` is a union of keywords _and_ identifiers. (I wonder if `TRUE` and `FALSE` should be _removed_ from `name`...)

rmosolgo:
Thanks for reporting this, fixed in 2c82c02, I ended up testing at the parser level

theorygeek:
Awesome! 🎆

PR #185, opened 2016-07-18
refactor(GraphQL::Relay) merge graphql-relay into graphql

rmosolgo:
I'm thinking about merging these projects, what do you think? 

Here are the pros I see:

- Seems like most (but not all) people use GraphQL because they're using Relay. I bet that will change over time, but right now, that seems to be the lion's share of users. Now they have one less dependency to keep up with.
- Managing compatibility will be much easier. As a maintainer, I can be sure that they're always in sync, and users can't ever have out-of-sync gem versions. 

There are some cons: 

- Adds the `require "graphql/relay"` step for Rails users which is a bit unusual
- More development dependencies, slower test suite 

Other options:

- Don't do it 
- Do it like [`jnunemaker/flipper`](https://github.com/jnunemaker/flipper), where the code is together but they're published to different gems.


metacortex:
Publishing gems like `flipper` looks better to me.

rmosolgo:
Why? (I'm not disagreeing, I'm just trying to understand how it looks to you!)

metacortex:
I think the current way is good. (using 2 separated gem declaration in the Gemfile)
It is more clear and feel comfortable to me.

But as you mentioned, adding `require 'graphql/relay'` is also acceptable.
because I am using relay now anyway :)


```
gem 'graphql'
gem 'graphql-relay'
gem 'graphql-apollo' (maybe someday)
...

gem 'carrierwave'
gem 'carrierwave-aliyun'
...

gem 'mina'
gem 'mina-puma'
gem 'mina-scp'
...

```

rmosolgo:
Thanks for sharing, yes, i see how it fits the pattern there!

rmosolgo:
Lol, you could publish `graphql-relay` which is just one line: 

```ruby
require "graphql/relay"
``` 

😆

metacortex:
Cool

rmosolgo:
Maybe I can just require it by default. It doesn't have any side effects ... _almost_. 

It checks for the presence of ActiveRecord or Sequel at load time, so you have to require those before requiring graphql. We can either document that dependency or expose some API for mounting those by hand if you miss the load-time hook.

I'm going to try this again in a bit -- I want to see if I can preserve the git history too.

PR #191, opened 2016-07-21
Allow the schema IDL to be parsed

dylanahsmith:
## Problem

I would like to have the ability parse a schema printed using in the [GraphQL schema IDL](https://github.com/facebook/graphql/pull/90) format so that I can compare an old version of the schema (stored in git) against a new version to detect breaking schema changes.

## Solution

Currently this PR updates the existing parser so that GraphQL.parse can also be used to parse a schema string.  This is the approach taken by graphql-js of having a single parser for both GraphQL queries and schemas (https://github.com/graphql/graphql-js/pull/163).  The alternative would be to have a separate parser for just the schema.

@rmosolgo If you are happy with this approach and interested in having this feature in this repo, then I'll start adding tests for this feature to this PR.

### DIFF
--- a/lib/graphql/language/generation.rb
+++ b/lib/graphql/language/generation.rb
@@ -1,10 +1,12 @@
 module GraphQL
   module Language
     module Generation
-      def self.generate(node, indent: "")
+      extend self
+
+      def generate(node, indent: "")
         case node
         when Nodes::Document
-          node.definitions.map { |d| generate(d) }.join("\n")
+          node.definitions.map { |d| generate(d) }.join("\n\n")
         when Nodes::Argument
           "#{node.name}: #{generate(node.value)}"
         when Nodes::Directive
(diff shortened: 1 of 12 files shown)


rmosolgo:
:+1: sweet, yeah, go for it! 

It seems a bit weird to use the same parser for such different things, but it looks like they share a lot of code (especially the lexer?) so certainly seems like a good way to start! And if we need to split them later, we can do it later.

dylanahsmith:
This is ready for a review now:
* GraphQL::Query now raises an exception if the query contains schema/type definitions
* The language nodes can now be printed like other language nodes
* GraphQL::Schema::Printer now prints a schema definition to specify the operation roots

lib/graphql/language/generation.rb
```diff
@@ -53,32 +55,73 @@ def self.generate(node, indent: "")
...
+        when Nodes::ObjectTypeDefinition
+          out = "type #{node.name}"
+          out << " implements " << node.interfaces.join(", ") unless node.interfaces.empty?
+          out << generate_field_definitions(node.fields)
+        when Nodes::InputValueDefinition
+          out = "#{node.name}: #{generate(node.type)}"
+          out << " = #{generate(node.default_value)}" unless node.default_value.nil?
+          out
+        when Nodes::FieldDefinition
+          out = node.name
```
rmosolgo: ruh roh!  I think we need a new string here, since it's being mutated below. Otherwise ... chaos!

```ruby
schema_doc = GraphQL.parse("type Mutation { post: Post }")
puts GraphQL::Language::Generation.generate(schema_doc)
# type Mutation {
#   post: Post
# }
puts GraphQL::Language::Generation.generate(schema_doc)
# type Mutation {
#   post: Post: Post
# }
puts GraphQL::Language::Generation.generate(schema_doc)
# type Mutation {
#   post: Post: Post: Post
# }
puts GraphQL::Language::Generation.generate(schema_doc)
# type Mutation {
#   post: Post: Post: Post: Post
# }
```

spec/graphql/language/generation_spec.rb
```diff
@@ -30,7 +31,7 @@
     describe "inputs" do
       let(:query_string) {%|
         query {
-          field(int: 3, float: 4.7e-24, bool: false, string: "☀︎🏆\\n escaped \\" unicode ¶ /", enum: ENUM_NAME, array: [7, 8, 9], object: {a: [1, 2, 3], b: {c: "4"}}, unicode_bom: "\xef\xbb\xbfquery")
+          field(int: 3, float: 4.7e-24, bool: false, string: "☀︎🏆\\n escaped \\" unicode ¶ /", enum: ENUM_NAME, array: [7, 8, 9], object: { a: [1, 2, 3], b: { c: "4" } }, unicode_bom: "\xef\xbb\xbfquery")
```
rmosolgo: Are these whitespace changes needed? Or are they testing something new?

spec/graphql/query_spec.rb
```diff
@@ -124,6 +124,20 @@
...
+      { root }
+
+      type Query { foo: String }
+    '
+    begin
+      GraphQL::Query.new(schema, query_string)
+      flunk "Expected an exception"
+    rescue GraphQL::ExecutionError => exc
+      assert_equal "GraphQL query cannot contain a schema definition", exc.message
+    end
```
rmosolgo: could we use `assert_raises` here? for example 

```ruby 
exc = assert_raises(GraphQL::ExecutionError) { GraphQL::Query.new(schema, query_string) } 
assert_equal "...", exc.message
```

?

If that works, I prefer it because of the `assert_` language!

spec/graphql/language/generation_spec.rb
dylanahsmith: They were so that GraphQL::Language::Generation and GraphQL::Schema::Printer print value literals the same way.  That way the GraphQL::Language::Parser "parses the test schema" test can assert that no information is lost through during parsing by simple doing

```
        schema_string = GraphQL::Schema::Printer.print_schema(schema)
        document = GraphQL::Language::Parser.parse(schema_string)

        assert_equal schema_string, document.to_query_string
```

spec/graphql/query_spec.rb
dylanahsmith: oh, I didn't realize assert_raises returns the exception.

lib/graphql/language/generation.rb
dylanahsmith: Good catch.  I've pushed a fix along with a regression test.

spec/graphql/language/generation_spec.rb
rmosolgo: oh, thanks!

rmosolgo:
:rocket: sweet!

PR #199, opened 2016-08-04
Update comment documentation

gjtorikian:
It's minor, but this tripped me up earlier. 😅


rmosolgo:
Oops! I updated that API after the fact but I guess I missed the docs 😬 thanks!!

PR #205, opened 2016-08-09
Site build improvements

gjtorikian:
This PR:

* Shuffles content into the top-level `site` directory.
* Adds [html-proofer](https://github.com/gjtorikian/html-proofer) to test the legitimacy of links.
* Fixes the deploy script such that the site is built and sent to `gh-pages`.
* Adds `script/server` to boot up the site ([because I'm lazy](http://githubengineering.com/scripts-to-rule-them-all/)).


Rakefile
```diff
@@ -33,15 +35,28 @@ task :build_parser do
...
+    Dir.mktmpdir do |tmp|
+      system "mv output/* #{tmp}"
+      system "git checkout gh-pages"
+      system "rm -rf *"
+      system "mv #{tmp}/* ."
+      system "git add ."
+      system "git commit -am 'deploy site to gh-pages (automatic)'"
+      system "git push origin gh-pages --force"
+      system "git checkout master"
+      system "echo yolo"
```
rmosolgo: just curious, is there a reason to sometimes use backticks, other times use `system`? (I never do this, I have no idea)

rmosolgo:
It looks like this branch removes the guides `.md` source, but adds `/index.html` files, was that intentional? (I think we should keep the `.md` source, maybe some directory wasn't added to the tree?)

Rakefile
gjtorikian: Ah, whoops. I almost always use `system` to be clear of the intent. First line was copy-pasted from somewhere else. 😅

gjtorikian:
> It looks like this branch removes the guides .md source, but adds /index.html files, was that intentional? (I think we should keep the .md source, maybe some directory wasn't added to the tree?)

Whoops! Totally not intentional. I've force-pushed an update to correct that last bad commit.

site/layouts/default.html
```diff
@@ -29,12 +29,12 @@
...
-            <li><%= link_to "Executing Queries", "/graphql-ruby/guides/executing_queries/" %></li>
-            <li><%= link_to "Code Reuse", "/graphql-ruby/guides/code_reuse/" %></li>
-            <li><%= link_to "Security", "/graphql-ruby/guides/security/" %></li>
-            <li><%= link_to "Testing", "/graphql-ruby/guides/testing/" %></li>
+            <li><%= link_to "Introduction", "/guides/introduction/" %></li>
+            <li><%= link_to "Defining a Schema", "/guides/defining_your_schema/" %></li>
+            <li><%= link_to "Executing Queries", "/guides/executing_queries/" %></li>
+            <li><%= link_to "Code Reuse", "/guides/code_reuse/" %></li>
+            <li><%= link_to "Security", "/guides/security/" %></li>
+            <li><%= link_to "Testing", "/guides/testing/" %></li>
```
rmosolgo: I added `graphql-ruby/` to the paths because this was getting deployed to `rmosolgo.github.com/graphql-ruby/`, is that prefix being added another way?

gjtorikian: Nope, I think you're right in that I'm wrong. Reverting.

gjtorikian:
@rmosolgo Hah! I wasn't sure when this was going to ship, but here ya go: https://github.com/blog/2228-simpler-github-pages-publishing

Aside from the test stuff, which can be cherry-picked, I reckon this PR is probably obsolete. What do you think?

rmosolgo:
Woaaah cool! This means that the build step should just output HTML to `docs/`, right?

gjtorikian:
Oh, duh, I just realized this is a nanoc site, not a Jekyll site. This will only work with vanilla Jekyll sites. 

> This means that the build step should just output HTML to docs/, right?

Yep. It doesn't look like you need any nanoc-specific functionality, so you _could_ just use Jekyll to generate the site by keeping the source files in _docs/_ instead of _site/_.

rmosolgo:
:+1: Jekyll 

I thought I'd try learning another static site generator, one with less stuff to configure, but if github pages will build it _for us_, that's too good to say no!

gjtorikian:
Ok. Destroying this branch and _starting over_.

PR #206, opened 2016-08-16
Pass `field_name` to `connection_class` resolver

gjtorikian:
Kind of kicking myself for not doing this earlier. I need the `field_name` for some error message handling, so let's pass it in as a kwarg to the resolver.

### DIFF
--- a/lib/graphql/relay/base_connection.rb
+++ b/lib/graphql/relay/base_connection.rb
@@ -60,10 +60,11 @@ module GraphQL
       # @param Query arguments
       # @param max_page_size [Int] The maximum number of results to return
       # @param parent [Object] The object which this collection belongs to
-      def initialize(nodes, arguments, max_page_size: nil, parent: nil)
+      def initialize(nodes, arguments, max_page_size: nil, field_name: field_name, parent: nil)
         @nodes = nodes
         @arguments = arguments
         @max_page_size = max_page_size
+        @field_name = field_name
         @parent = parent
       end
 
--- a/lib/graphql/relay/base_connection.rb
+++ b/lib/graphql/relay/base_connection.rb
@@ -60,7 +60,7 @@ module GraphQL
       # @param Query arguments
       # @param max_page_size [Int] The maximum number of results to return
       # @param parent [Object] The object which this collection belongs to
-      def initialize(nodes, arguments, max_page_size: nil, field_name: field_name, parent: nil)
+      def initialize(nodes, arguments, field_name: field_name, max_page_size: nil, parent: nil)
         @nodes = nodes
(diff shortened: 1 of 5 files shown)


rmosolgo:
:+1: love the idea,  a couple of suggestions: 

- what about passing the `GraphQL::Field` _itself_ to the connection object? that way you could get `field.name`, but also "anything" else (not sure what folks would need, but you could have `field.type`, `field.metadata` ...)
- how about a test to make sure we maintain this API? One way I can imagine testing is by adding a field to one of the custom connections eg 

  ```ruby 
  field :fieldName, types.String, resolve: -> (obj, args, ctx) { obj.field.name } 
  ```

  then making sure it comes back correctly in a query, eg 
 
  ```graphql 
  { 
    rebels { 
      basesCustomConnection { 
      # ^ not sure if that's the right name, but I think there's one in the test schema already 
        fieldName 
        # ^ this should be "basesCustomConnection" 
      } 
    } 
  }
  ```
        

?

gjtorikian:
@rmosolgo I have done both those things!

lib/graphql/relay/base_connection.rb
```diff
@@ -53,17 +53,18 @@ def register_connection_implementation(nodes_class, connection_class)
...
-      attr_reader :nodes, :arguments, :max_page_size, :parent
+      attr_reader :nodes, :arguments, :max_page_size, :parent, :field
 
       # Make a connection, wrapping `nodes`
       # @param [Object] The collection of nodes
       # @param Query arguments
       # @param max_page_size [Int] The maximum number of results to return
       # @param parent [Object] The object which this collection belongs to
-      def initialize(nodes, arguments, max_page_size: nil, parent: nil)
+      def initialize(nodes, arguments, field: field, max_page_size: nil, parent: nil)
```
rmosolgo: I don't recognize this syntax, `field: field,`, is it meant to be a required keyword, `field:,` or does this default it to `nil`, since it calls the `attr_reader`?

gjtorikian: It's meant to be required, but I could've just written it as `field:`.

rmosolgo:
shipped in 0.18.2

gjtorikian:
Thanks!

panbanda:
When I upgraded from 1.9.14 to 1.9.16 this may have been introduced.  In my rspec I initialize the mutation like this:

```ruby
Mutations::Auth::Login.new(object: nil, context: {}).resolve(args)
```

And all my mutations fail when `field:` is missing.  The following passes:

```ruby
Mutations::Auth::Login.new(object: nil, field: nil, context: {}).resolve(args)
```

PR #207, opened 2016-08-17
Implement loading schema from JSON introspection query dump

josh:
This library currently defines the introspection query (`GraphQL::Introspection::INTROSPECTION_QUERY`) for dumping a schema, but no mechanize for loading it back in elsewhere.

This initial PR is mostly a spike/PoC. Theres still alot of cases to implement, but I wanted to maintainer feedback first before putting more work into it.

``` ruby
json = JSON.load("swapi-schema.json")
schema = GraphQL::Schema::Loader.load(json)

# executing directly against this schema object will
# not work (as expected) since there are no resolvers
schema.execute(...) 

# however, we can still use it to validate queries
query = GraphQL::Query.new(schema, "query { ... }")
validator.validate(query)
```

-
To: @rmosolgo 
CC: @kdaigle @charliesome

### DIFF
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -1,0 +1,124 @@
+module GraphQL
+  class Schema
+    module Loader
+      extend self
+
+      def load(obj)
+        schema = obj.fetch("data").fetch("__schema")
+
+        types = {}
+        type_resolver = -> (type) { -> { resolve_type(types, type) } }
+
+        schema.fetch("types").each do |type|
+          next if type.fetch("name").start_with?("__")
+          type_object = define_type(type, type_resolver)
+          types[type_object.name] = type_object
+        end
+
+        query = types.fetch(schema.fetch("queryType").fetch("name"))
+
+        Schema.new(query: query, types: types.values)
+      end
+
+      private
+
+      def resolve_type(types, type)
+        case kind = type.fetch("kind")
+        when "ENUM", "INTERFACE", "INPUT_OBJECT", "OBJECT", "SCALAR", "UNION"
+          types.fetch(type.fetch("name"))
+        when "LIST"
+          ListType.new(of_type: resolve_type(types, type.fetch("ofType")))
+        when "NON_NULL"
+          NonNullType.new(of_type: resolve_type(types, type.fetch("ofType")))
+        else
+          fail NotImplementedError, "#{kind} not implemented"
+        end
+      end
+
+      def define_type(type, type_resolver)
+        case type.fetch("kind")
+        when "ENUM"
+          EnumType.define(
+            name: type["name"],
+            description: type["description"],
+            values: type["enumValues"].map { |enum|
+              EnumType::EnumValue.new(
+                name: enum["name"],
+                description: enum["description"],
+                deprecation_reason: enum["deprecationReason"],
+                value: enum["value"]
+              )
+            })
+        when "INTERFACE"
+          InterfaceType.define(
+            name: type["name"],
+            description: type["description"],
+            fields: Hash[type["fields"].map { |field|
+              [field["name"], define_type(field.merge("kind" => "FIELD"), type_resolver)]
+            }]
+          )
+        when "INPUT_OBJECT"
+          InputObjectType.define(
+            name: type["name"],
+            description: type["description"],
+            arguments: Hash[type["inputFields"].map { |arg|
+              [arg["name"], define_type(arg.merge("kind" => "ARGUMENT"), type_resolver)]
+            }]
+          )
+        when "OBJECT"
+          ObjectType.define(
+            name: type["name"],
+            description: type["description"],
+            fields: Hash[type["fields"].map { |field|
+              [field["name"], define_type(field.merge("kind" => "FIELD"), type_resolver)]
+            }]
+          )
+        when "FIELD"
+          Field.define(
+            name: type["name"],
+            type: type_resolver.call(type["type"]),
+            description: type["description"],
+            arguments: Hash[type["args"].map { |arg|
+              [arg["name"], define_type(arg.merge("kind" => "ARGUMENT"), type_resolver)]
+            }]
+          )
+        when "ARGUMENT"
+          Argument.define(
+            name: type["name"],
+            type: type_resolver.call(type["type"]),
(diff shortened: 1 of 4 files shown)


rmosolgo:
:+1: :tada:  This is awesome! 

A couple other thoughts: 

- It would be nice to leave API space for a loader based on the "schema language", since that can be parsed since #191. (I don't think it has to happen now, I just figure someone will need it soon, and it would be nice to have a "spot" for it when the time comes.)
- Why `Array(...)` ? I thought we could safely assume that they return arrays, but am I missing something?

josh:
> Why Array(...) ? I thought we could safely assume that they return arrays, but am I missing something?

Thats just a little trick to convert `nil` to `[]`. Though, I think I need to read through the spec to see which of these values are nullable and non-nullable.

josh:
I'm trying to come up with a good way to cleanup and organize the loading logic for each type.

What if each `GraphQL::???Type` define a loader method that could create the type without needing to go through the `define` DSL?

``` ruby
GraphQL::ObjectType.load({ "name" => "Post", "description" => "A blog post", "fields" => [ ... ] })
```

rmosolgo:
`.define(...)` will _almost_ cover those cases -- since 60a3d3dae6b9afe9b550510c59dcaafb6542e82d, `.define` accepts keywords with the same names as the "helper methods" inside the block. So, these are equivalent: 

```ruby 
GraphQL::ObjectType.define do 
  name "Post"
  description "A post on my blog" 
end 

# is the same as 
GraphQL::ObjectType.define(name: "Post", description: "A post on my blog")
```

However, this has a couple of shortcomings: 

- It expects symbols, not strings, so we'd either have to: 
  
  - whitelist & `.to_sym` them (when was symbol GC introduced? I think we have to whitelist)
  - update `.define()` to accept strings too 

- It doesn't support setting all fields at once, ie 
  
  ```ruby
  field :favoriteIceCream, IceCreamType 
  field :favoriteColor, ColorType 
  ```

  doesn't have an analogous keyword. So I think we'd have to add `fields: [...]` which takes an array of `GraphQL::Field`s and reduces them to a `{name => GraphQL::Field}` hash 


If we can make some of those changes, would that work for you? I don't mind taking some time to beef up `.define` if you think it would work for this case.

josh:
> .define(...) will almost cover those cases -- since 60a3d3d, .define accepts keywords with the same names as the "helper methods" inside the block. So, these are equivalent:

Oh, awesome! I didn't realize that.

> If we can make some of those changes, would that work for you?

Yeah, I'll see what I can do.

Thanks so much for the detailed guidance!

josh:
> .define accepts keywords with the same names as the "helper methods" inside the block.

Okay, look into this. It helps a little, but not that much. I think the issue is around type resolution. The flatten static schema basically has string names of types like `{ "name" => "Post" }`. But the type constructors expect already constructed type instances to be assigned. And type constructors themselves have no parent dependency on a schema, so they can't resolve from a type storage either.

rmosolgo:
To solve a similar issue (fetching types from Rails models) I used a proc which would be called lazily 

```ruby
new_field.type = -> { TYPE_NAME_TO_TYPE[field_type_name] }
```

And in the mean time, `TYPE_NAME_TO_TYPE` was being populated, eg 

```ruby
TYPE_NAME_TO_TYPE[type.name] = type 
```

So that when the proc was called, the type would be there! 

I'd like to keep types independent of the schema if possible, I guess well see about that :S

josh:
> I'd like to keep types independent of the schema if possible, I guess well see about that :S

I agree, I think they need to be. Otherwise we're in for more circular dependency headaches.

josh:
This PR is kinda getting out of hand. Trying to cherry pick some smaller helper changes into other PRs.

See https://github.com/rmosolgo/graphql-ruby/pull/222.

josh:
@rmosolgo thanks for merging https://github.com/rmosolgo/graphql-ruby/pull/222

Can you have another review of this PR. I'm feeling stuck again as to how to clean it up further. It seems feels a little messy. Just alot of string -> symbol keyword mapping going on.

lib/graphql/schema/loader.rb
```diff
@@ -0,0 +1,124 @@
+require "set"
```
rmosolgo: Is set still used?

rmosolgo:
Yes, agreed that the mapping unusual, but for my part, I'm not sold on any of the alternatives. Some [other](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/printer.rb#L154) [parts of](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/value_resolution.rb#L97) [this codebase](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/literal_input.rb#L79) use a hash lookup instead of a giant `case`. Previously this was my preference, but now I think, why introduce all the organizational overhead? You want some code to run in case X ... just write it 😆 . (Sure, there are times for better organization, but maybe `case` is fine for these kind of things.)

The double-nested proc (`type_resolver`) is unusual but I think I get the gist of it -- it's a function that, given a type hash, will return the corresponding type definition. 

Since the public API is _literally_ one function, we have unlimited ability to refactor the implementation, so I'm happy to leave it as-is :) 

The only things I would change are: 
- mark everything but `load` as private, either by enforcing it in Ruby or by yardoc `@private`
- ~~use named type kinds instead of string literals~~ nevermind, i see that you're really switching on names, and most of the named constants are objects of some kind or another

lib/graphql/schema/loader.rb
josh: Nope, I'll get rid of that.

josh:
> Since the public API is literally one function, we have unlimited ability to refactor the implementation, so I'm happy to leave it as-is :)

Ha, very true.

> mark everything but load as private, either by enforcing it in Ruby or by yardoc @private

👍

josh:
Alright, updated.

Thanks for all the advice @rmosolgo.

Maybe I'll stop messing around with this and we can iterate as we go.

rmosolgo:
:rocket: so good, thanks!!

rmosolgo:
🚢 in 0.18.6!

josh:
❤️

PR #208, opened 2016-08-18
refactor(Schema) create schemas with Schema.define

rmosolgo:
Schema initialization has slowly grown unwieldy: 

- some values are passed to `.new` (eg `query:`, `mutation:`, `max_depth:`)
- others are reassigned _after_ initialization (eg `schema.query_execution_strategy = `) 
- others are added by mutating the schema (eg `middleware << MyMiddleware.new`).
- others are defined by calling methods on the schema (eg `schema.rescue_from(err_class, &block)`)

To me, this is sloppy and confusing (and hard to maintain). This gem already includes a DSL for configuring GraphQL-related objects, so let's apply it here too. So, you can create schemas using the same `.define` API as types, fields and arguments. 

```ruby 
  query QueryType 
  mutation MutationType 
end 
```

This also provides `Schema#metadata`, which makes schema extensible & ready for schema-level values.

## Todo 

- [x] Update docs 
- [x] Update inline schemas in test

### DIFF
--- a/guides/defining_your_schema.md
+++ b/guides/defining_your_schema.md
@@ -238,11 +238,11 @@ Some parts of schema definition take types as an input. There are two good ways 
 Your schema can be initialized with some options:
 
 ```ruby
-MySchema = GraphQL::Schema.new(
-  query: QueryType,       # root type for read-only queries
-  mutation: MutationType, # root type for mutations
-  max_depth: 7,           # if present, the max depth for incoming queries
-)
+MySchema = GraphQL::Schema.define do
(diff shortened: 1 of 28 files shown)


kdaigle:
If it's not too much of a trouble, could you add some context around "why" you're making this change? :heart:

rmosolgo:
Haha, sorry I was in a hurry, maybe I should have just waited. Description updated!

rmosolgo:
Lol, added one more wacky way to define a schema

PR #209, opened 2016-08-19
Allow interfaces to be compatible with objects that implement said interface

kdaigle:
In `0.18.2` (but it was working in `0.18.1`), it appears some functionality was implicitly broken. The new version seemingly broke the ability to have a query fragment on an interface. For example...

```
{
  node(id:"abc123") {
    ...repoInfo
  }
}

fragment repoInfo on RepositoryInfo {
  owner {
    login
  }
}
```

... where `RepositoryInfo` is an Interface and the `node` returned is an Object that implements the `RepositoryInfo` interface.

I'm not entirely confident I'm not breaking something here or if I'm breaking the letter of GraphQL. I just know the functionality changed. :smile: I put together 79431da and cf65366 to show a failing test case.

### DIFF
--- a/lib/graphql/execution/typecast.rb
+++ b/lib/graphql/execution/typecast.rb
@@ -22,8 +22,10 @@ module GraphQL
           current_type.resolve_type(value, query_ctx) == potential_type
         elsif potential_type.kind.union?
           potential_type.include?(current_type)
+        elsif current_type.kind.interface? && potential_type.respond_to?(:interfaces)
+          potential_type.interfaces.include?(current_type)
         elsif current_type.kind.interface?
-          current_type.resolve_type(value, query_ctx) == potential_type
+          (current_type.resolve_type(value, query_ctx) == potential_type)
         elsif potential_type.kind.interface?
           current_type.interfaces.include?(potential_type)
         else
--- a/lib/graphql/execution/typecast.rb
+++ b/lib/graphql/execution/typecast.rb
@@ -1,11 +1,13 @@
 module GraphQL
   module Execution
-    # GraphQL object `{value, type}` can be cast to `other_type` when:
-    # - `type == other_type`
-    # - `type` is a union and it resolves `value` to `other_type`
-    # - `other_type` is a union and `type` is a member
-    # - `type` is an interface and it resolves `value` to `other_type`
-    # - `other_type` is an interface and `type` implements that interface
+    # GraphQL object `{value, current_type}` can be cast to `potential_type` when:
+    # - `current_type == potential_type`
+    # - `current_type` is a union and it resolves `value` to `potential_type`
+    # - `potential_type` is a union and `current_type` is a member
+    # - `current_type` is an interface, `potential_type` has interfaces,
+    #   and `current_type` is one of those interfaces
+    # - `current_type` is an interface and it resolves `value` to `potential_type`
+    # - `potential_type` is an interface and `current_type` implements that interface
     module Typecast
       # While `value` is exposed by GraphQL as an instance of `current_type`,
       # should it _also_ be treated as an instance of `potential_type`?
--- a/spec/graphql/execution/typecast_spec.rb
+++ b/spec/graphql/execution/typecast_spec.rb
@@ -22,6 +22,10 @@ describe GraphQL::Execution::Typecast do
   end
 
   it "resolve correctly when potential type is an Interface and current type implements it" do
+    assert GraphQL::Execution::Typecast.compatible?(CHEESES[1], EdibleInterface, CheeseType, context)
+  end
+
+  it "resolve correctly when potential type is an Interface and current type implements it" do
     assert GraphQL::Execution::Typecast.compatible?(MILKS[1], EdibleInterface, CheeseType, context)
   end
 
--- a/spec/graphql/execution/typecast_spec.rb
+++ b/spec/graphql/execution/typecast_spec.rb
@@ -1,0 +1,28 @@
+require "spec_helper"
+
+describe GraphQL::Execution::Typecast do
+
+  let(:schema) { DummySchema }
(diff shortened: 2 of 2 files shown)


rmosolgo:
Ahh sorry I forgot about 01c163791986c528eb2ddbdb76deb80b4ca2dc59 ! Thanks for adding tests for this module, this will make our lives a lot easier down the road. 

Did you get a chance to try this on your specific case? I ask because I would have diagnosed the case a bit differently, here's my take: https://github.com/rmosolgo/graphql-ruby/pull/210

kdaigle:
I did try my exact case in this PR and it worked. I've given you the 👍🏻 on your other PR assuming that the tests are accurately testing my concern. I could also pull down your other branch and give it a try if that'd help. Let me know.

I know you're super busy but if you can provide some feedback on why my diagnosis was incongruent with your thinking it'd probably help share some extra knowledge with me and my team in this area. But, I'm mostly just happy we have a resolution to the bug. :sparkles: 👍

rmosolgo:
I started to type this but I realized I was wrong: 

> I was trying to think through the flow of types in this query, like what was "self type" at each point, here's how it seems to me: 
> 
> ```graphql 
> {
>   # QueryType
>   node(id:"abc123") {
>     # NodeInterface
>     ... on RepositoryInfo {
>       # RepositoryInterface 
>       owner {
>         # PersonType 
>         name 
>       }
>     }
>   }
> }
> ```
> So, I would expect the broken typecast to be `NodeInterface => RepositoryInterface` . In case of a typecast like that, both `current_type` & `possible_type` would be `kind.interface?`, and I don't _see_ a case for that, even with your patch.
>

I was missing another typecast of sorts -- [`ValueResolution` resolves the type](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/value_resolution.rb#L57) before going to the inner fields. So I think my end-to-end test must not actually touch the new code in my patch. I'm curious to see if this passes: https://github.com/rmosolgo/graphql-ruby/pull/211

This makes me wonder if an interface's `resolve_type` is not returning the right type for an object. This patch circumvents `resolve_type`, which I don't think is the right thing. For example, only one of these assertions should pass: 

```ruby
  it "resolve correctly when current type is an Interface and it resolves to potential type" do
    assert GraphQL::Execution::Typecast.compatible?(MILKS[1], CheeseType, EdibleInterface, context)
    assert GraphQL::Execution::Typecast.compatible?(MILKS[1], MilkType, EdibleInterface, context)
    assert GraphQL::Execution::Typecast.compatible?(MILKS[1], HoneyType, EdibleInterface, context)
  end
``` 

Is it possible that `RepositoryInfo`'s `resolve_type` function is missing some info? (There's a short [discussion of `resolve_type`](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/defining_your_schema.md#interface-types) in the guides.)

rmosolgo:
I was thinking, "could it be NodeInterface's `resolve_type`?", but I don't think it could, since if that happened, you'd get an explicit unresolved type error from https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/value_resolution.rb#L60

kdaigle:
Yeah, I'd expect it to blow up. I know my PR definitely solved the problem and it worked in `v0.18.1` so I'm relatively confident the refactor broke something. 😕

rmosolgo:
Hmm, let me try to zero in on the behavior change from that refactor...

rmosolgo:
Ok, here's the change that I've had my eye on: 

## Before 

```ruby 
elsif child_type.kind.interface?		
  child_type.resolve_type(target, query_ctx)
```
## After 

```ruby 
elsif current_type.kind.interface?
  current_type.resolve_type(value, query_ctx) == potential_type
```
(`child_type` was renamed to `current_type`, but they're the [same incoming values](https://github.com/rmosolgo/graphql-ruby/commit/01c163791986c528eb2ddbdb76deb80b4ca2dc59#diff-83ca4ff1ce9b41dc1422edc74dfe912cL36), these names are killin' me)


Previously, it was only required that `resolve_type` returned non-nil for an object, and we took that as an :ok_hand:, but now we assert that its return value is the type we were looking for. There's a specific case where this changes the behavior: 

- Two (or more) types implement & declare an interface 
- The interface sometimes returns the "wrong" type for an incoming object

Here's a test case with the correct behavior: https://github.com/rmosolgo/graphql-ruby/compare/typecast-differences-test-from-master

Here's a branch from pre-refactor with the same test case: https://github.com/rmosolgo/graphql-ruby/compare/typecast-differences-test-from-previous (I'm not sure how to diff it against 01c163791986c528eb2ddbdb76deb80b4ca2dc59, which is the parent of that branch)

If I run that test _before_ the refactor, it gives this output: 

image

rmosolgo:
I'm still not sure that the behavior change I described above is causing the bug you're seeing... 

Specifically, my hesitation with this patch is summed up by this test: 

```ruby 
it "resolve correctly when current type is an Interface and it resolves to potential type" do
 assert GraphQL::Execution::Typecast.compatible?(MILKS[1], CheeseType, EdibleInterface, context)
end
```

Should we really ignore the fact that `EdibleInterface` thinks the incoming object is of `MilkType`, and cast it as `CheeseType` _anyway_? 

Maybe we have enough info here that it's irrelevant -- I can't think of a way that this could happen (It should already have `MilkType` from the `resolve_type` call in value resolution) but it doesn't seem right to "throw out" that info about the object itself. :thinking_face: ...

rmosolgo:
Ok, one weird thing is that we only use the value in `Interface => Object` casts, _not_ `Object => Interface` casts ... (added some more tests in #212 )

rmosolgo:
Addressed by #212

PR #210, opened 2016-08-19
Fix interface typecast

rmosolgo:
Casting an interface to an interface was broken since 01c163791986c528eb2ddbdb76deb80b4ca2dc59. The previous implementation was _too_ permissive, the new was too restrictive. 

Allow an object to be cast from one interface to another if: 

- It resolves from the current interface to an object type 
- that object type implements the requested interface type

I suspect we have a similar bug with unions, let me see ....


kdaigle:
@rmosolgo I'm totally happy with this if you are. 👍🏻 Would love to get this cut and merged.

rmosolgo:
I don't think this actually _did_ anything, the tests passed on master at #211 😅

PR #212, opened 2016-08-19
Fix typecast api

rmosolgo:
- Stop checking `value` when typecasting -- only use info from types at hand 
- Reorder values passed to `compatible?`, they were passed in the wrong order :S 
- Include the tests from #209 and add some negative examples

I think issues can arise when different interfaces return different results for `resolve_type` -- but it's hard for me to figure out exactly which cases affect which things. 

Maybe it would be better to adopt a "full-Relay" strategy for the whole schema: the user defines a `resolve_type` function for the _schema_, rather than individual interfaces, that way you never get conflicting results.

### DIFF
--- a/lib/graphql/execution/typecast.rb
+++ b/lib/graphql/execution/typecast.rb
@@ -1,20 +1,24 @@
 module GraphQL
   module Execution
-    # GraphQL object `{value, type}` can be cast to `other_type` when:
-    # - `type == other_type`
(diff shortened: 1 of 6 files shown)


rmosolgo:
@kdaigle I think this includes the main fix from #209 -- switching from `resolve_type` to checking for membership in `.interfaces`

Issue #213, opened 2016-08-20
ArgumentError (missing keyword: field):


gauravtiwari:
Hey @rmosolgo 

Getting follow error when creating a new relation connection object - `ArgumentError (missing keyword: field):`

I see addition of new field argument. Does this require passing a field option?

rmosolgo:
Oops, we forgot about the range add use case! This was added in #206, I guess it should be optional

Definitely need a range add test to avoid this in the future, sorry!

gauravtiwari:
Yeah, I think not needed in all use cases. No worries! 😄

gauravtiwari:
All works! Fixed by #215

panbanda:
When I upgraded from 1.9.14 to 1.9.16 this may have been introduced.  In my rspec I initialize the mutation like this:

```ruby
Mutations::Auth::Login.new(object: nil, context: {}).resolve(args)
```

And all my mutations fail when `field:` is missing.  The following passes:

```ruby
Mutations::Auth::Login.new(object: nil, field: nil, context: {}).resolve(args)
```

rmosolgo:
Thanks for sharing your solution! It's also listed here https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#breaking-changes-1

firedev:
Sorry I don't get it, I did interit from my own `Base` class but I didn't override `initialize`. Why do I have to add `field: nil` everywhere?

rmosolgo:
The method has a new required argument. It's meant to be called by GraphQL-Ruby internals, which I already updated. If you're using it in some other way, you'll have to update it.

firedev:
I am updating all tests. I thought it meant to be an optional argument. Guess it wasn't.

PR #214, opened 2016-08-20
Document passing field to connection class and creating relation object

gauravtiwari:
Add reference to the doc for #213 

Is this good enough?

### DIFF
--- a/guides/relay.md
+++ b/guides/relay.md
@@ -295,11 +295,30 @@ Maybe you need to make a connection object yourself (for example, to return a co
 ```ruby
 items = [...]     # your collection objects
 args = {}         # stub out arguments for this connection object
+field = nil # optional field name
 connection_class = GraphQL::Relay::BaseConnection.connection_for_nodes(items)
-connection_class.new(items, args)
+connection_class.new(items, args, field: field)
 ```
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
I think this is covered by #215 , right?

gauravtiwari:
@rmosolgo Yup, but I guess would be good to keep the relation connection part in readme, folks using AR or Sequel might find it useful. What do you think? I can update.

gauravtiwari:
btw @rmosolgo is that new field option needs documenting -  although it's optional?

rmosolgo:
Yes, happy to take an updated patch! If you keep it, I think it's important to mention `field:` separately, after the main example. eg 

> Additionally, connections may be provided with the `GraphQL::Field` that created them. This may be used for custom introspection or instrumentation. For example, ...

I think most people don't actually need it, so I don't want to present it early in the documentation. 

Getting the field from inside the `resolve` proc is not easy, hmm I can think of only a couple ways: 

```ruby 
Schema.get_field(TodoListType, "todos") 
# => #<GraphQL::Field name="todos">
context.irep_node.definitions[TodoListType] 
# => #<GraphQL::Field name="todos">
# although this one may not work with fields on interfaces
```

gauravtiwari:
@rmosolgo there you go. Good enough?

rmosolgo:
💰 👌  Thanks!!

PR #215, opened 2016-08-20
Connection#initialize(field:) is optional

rmosolgo:
Oops, when I merged #206, I forgot about the use case of creating connection objects during range-add mutations (eg #213) 

In that case, I think it's safe to allow fields to be nil. Previously, I supported requiring it, so we didn't forget it in any internal code, but I think it's important support this case. 

I've also "fixed" the mutation spec to be a proper RANGE_ADD example. 

@gauravtiwari & @gjtorikian, does this look alright to you? Did I miss anything?

### DIFF
--- a/lib/graphql/relay/base_connection.rb
+++ b/lib/graphql/relay/base_connection.rb
@@ -61,7 +61,7 @@ module GraphQL
       # @param field [Object] The underlying field
       # @param max_page_size [Int] The maximum number of results to return
       # @param parent [Object] The object which this collection belongs to
-      def initialize(nodes, arguments, field:, max_page_size: nil, parent: nil)
+      def initialize(nodes, arguments, field: nil, max_page_size: nil, parent: nil)
         @nodes = nodes
         @arguments = arguments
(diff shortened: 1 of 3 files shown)


gjtorikian:
Shoot, I'm sorry about that. This looks good to me. 👍

gauravtiwari:
@rmosolgo awesome! Looks good to me :+1:

rmosolgo:
Thanks for double-checking!

rmosolgo:
lol @gjtorikian I blame the guy who pushed "merge" :P seriously, there should have been a test for this use since I've been "suggesting" it for months

panbanda:
When I upgraded from `1.9.14` to `1.9.16` this may have been introduced.  In my rspec I initialize the mutation like this:

```ruby
Mutations::Auth::Login.new(object: nil, context: {}).resolve(args)
```

And all my mutations fail when `field:` is missing.  The following passes:

```ruby
Mutations::Auth::Login.new(object: nil, field: nil, context: {}).resolve(args)
```

PR #217, opened 2016-08-24
Set exception message in InvalidNullError

charliesome:
I'm seeing some `GraphQL::InvalidNullError` exceptions bubble through to our exception tracker and without any message set it's really hard to figure out where this exception is coming from.

This pull request changes `initialize` to call super with the existing message in the `to_h` method.

### DIFF
--- a/lib/graphql/invalid_null_error.rb
+++ b/lib/graphql/invalid_null_error.rb
@@ -5,13 +5,12 @@ module GraphQL
     def initialize(field_name, value)
       @field_name = field_name
       @value = value
+      super("Cannot return null for non-nullable field #{@field_name}")
     end
 
     # @return [Hash] An entry for the response's "errors" key
     def to_h
-      {
-        "message" => "Cannot return null for non-nullable field #{@field_name}"
-      }
+      { "message" => message }
     end
 
     # @return [Boolean] Whether the null in question was caused by another error


rmosolgo:
:+1: if this is enough, another option would be to improve the error's API to include the parent type (who owns the field) and/or the expected return type. 

I think this error is only raised here: 

https://github.com/rmosolgo/graphql-ruby/blob/66ed2dd389108a4d1eb98a28885c2a9e8e0956a1/lib/graphql/query/serial_execution/value_resolution.rb#L85

It could be updated to 

```ruby 
# parent_type is the field owner 
# definition_name is the field name (without any alias) 
# field_type is the expected type (with a NonNullType wrapper) 
# value is `nil` or `GraphQL::ExecutionError`
raise GraphQL::InvalidNullError.new(parent_type, irep_node.definition_name, field_type, value)
```

Additionally, the message could include `value` -- that might be helpful in cases where the `nil` was caused by a `GraphQL::ExecutionError`.

What do you think, is this good enough for now? Happy to merge & :ship: if you just want to get this going, or wait if we want to add anything else while we're at it.

rmosolgo:
We can always improve on the next go-round!

charliesome:
@rmosolgo Ugh whoops, I don't think I ever got pinged for your reply here. 👍  for improving the error API to reference the parent type and field as well

PR #218, opened 2016-08-25
Query with nothing but comments causes stack overflow

gjtorikian:
We noticed a weird edge case where, if a query is nothing but a comment, an endless loop occurs with the fragment validator. The test I've pushed demonstrates the issue:

<details>
GraphQL::StaticValidation::FragmentsAreUsed::queries that are comments
  test_0001_handles them gracefully                              ERROR (0.25s)
SystemStackError:         SystemStackError: stack level too deep
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/static_validation/type_stack.rb:44:in `block in initialize'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:73:in `call'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:73:in `block in apply_hooks'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:73:in `each'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:73:in `reduce'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:73:in `apply_hooks'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:59:in `begin_visit'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:51:in `visit_node'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/language/visitor.rb:45:in `visit'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/static_validation/validator.rb:33:in `validate'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:134:in `perform_validation'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:86:in `selected_operation'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:99:in `variables'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:139:in `perform_validation'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:86:in `selected_operation'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:99:in `variables'
            /Users/gjtorikian/Development/graphql-ruby/lib/graphql/query.rb:139:in `perform_validation'
[...]

PR #219, opened 2016-08-25
Fix empty queries

rmosolgo:
Currently, a query with no operations causes an infinite loop ( #218 ). This is because query validation is a trainwreck! What I _want_ to do is: 

- Provide enough public API on `Query` that other parts of the system don't have to worry about whether the query was previously validated or not, eg: 
  - `AnalyzeQuery` can just ask for `query.internal_representation` and _get it_.
  - `Executor` can just ask for `query.selected_operation` and _get it_.
- Make validation a lazy operation -- only run it if needed (is this pointless? is there a case where you make a query but don't try to execute it? The only _really_ public API for `GraphQL::Query` is via `Schema#execute`). 
- Make the steps in the process decoupled enough that they can be tested in isolation 

~~These are all 👎  in this patch:~~
- ~~Passing an unvalidated Query to AnalyzeQuery would give `{}` for internal representation, that's bogus~~ 
- ~~Validation is ahead-of-time, not lazy~~
- ~~Validation is not decoupled from the query itself, see added `validate: false` to test suite~~

~~But this is a step in the right direction -- at least the _phases_ of validation have been separated and explicitly put in sequence.~~

~~I think another commit can get validation back to an as-needed operation~~ 

Ok, it's lazy again 😀 

@gjtorikian for whatever reason, returning `nil` from `Query#execute` returns a result of `{}`, I changed your test, is that ok?

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -65,8 +65,8 @@ module GraphQL
       @arguments_cache = Hash.new { |h, k| h[k] = {} }
       @validation_errors = []
       @analysis_errors = []
-      @internal_representation = {}
-      valid?
+      @internal_representation = nil
+      @was_validated = false
     end
 
     # Get the result for this query, executing it once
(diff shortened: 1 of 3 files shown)


gjtorikian:
`{}` is totally ok, thank you for the quick work as always!!!

rmosolgo:
:ship: 0.18.4

PR #221, opened 2016-08-26
Include expected type in validation error message

swalkinshaw:
The inspiration behind this was the more helpful error messages that GraphiQL has.

image

Two questions/possible changes:

1. I could unwrap `arg_defn` but I'm on the fence as to what's more helpful.
2. I initially wanted to output `node.value` as well but it could be nested with arguments. I couldn't find an existing way to get to the underlying value so leaving it out for now.

This is still an improvement regardless.


rmosolgo:
💰 Thanks, this is awesome! I think you're right to leave `arg_defn` as-is. I'll make a to-do item about putting the actual received value in the error message. Thanks for this improvement!

PR #222, opened 2016-08-29
Add additional Type.define convenience arguments

josh:
(Extraction from #207)

Suggestion from https://github.com/rmosolgo/graphql-ruby/pull/207#issuecomment-240608269.

Adds some additional parameters to `Type.define` to allow types to be constructed without a lazy `define do` block.

-
To: @rmosolgo


rmosolgo:
💰 !!

josh:
Thanks! ❤️

PR #223, opened 2016-08-29
Extract reusable parser tests

staugaard:
This moves the parser tests into a reusable module so that they can be used across multiple parsers.

This allows us to use the same tests on `graphql-libgraphqlparser-ruby`


staugaard:
Huh? The tests pass just fine on my machine.

spec/graphql/language/parser_spec.rb
```diff
@@ -1,562 +1,7 @@
 require "spec_helper"
+require 'graphql/Language/parser_tests'
```
rmosolgo: I think this should be a little `l`, not a big `L`! Maybe it passes on OSX where the filesystem is not case sensitive?

staugaard: Oh snap!

dphaener: Just to throw in my two cents, that's exactly the issue. Has bitten me quite a few times.

staugaard:
All green now. Sorry about that.

rmosolgo:
Awesome, thanks!

rmosolgo:
:ship: in 0.18.6!

staugaard:
Thanks!

Issue #224, opened 2016-08-30
Description of mutation


b8in:
Hello guys!

I would be grateful if you could share your thoughts and help on the following thing considering GraphQL and GraphiQL Editor.

I use GraphQL in my application. I provide descriptions for my queries/mutations and use GraphiQL Editor for reading the outcoming documentation.
Currently, mutation description is not displayed in the editor:
http://joxi.ru/nAyqdDWcXJ648A

Here's my code:

```ruby
#mutation_type.rb
MutationType = GraphQL::ObjectType.define do
  name "Mutation"
  description "The mutation root for this schema"
  
  field :createBlogPost, field: Widgets::Blog::CreateBlogPostMutation.field
end

#widgets\blog\create_blog_post_mutation.rb
Widgets::Blog::CreateBlogPostMutation = GraphQL::Relay::Mutation.define do

  # Used to name derived types:
  name "createBlogPost"
  description I18n.t('graphql.mutations.createBlogPost.description')

  # ...
end
```

Here are the results from rails console:

```ruby
>> Widgets::Blog::CreateBlogPostMutation.field.description 
nil

>> Widgets::Blog::CreateBlogPostMutation.description 
"*Create new post in blog widget*"
```

Maybe, description should be set somehow differently. 
If yes, then please, advise on how could I do it correctly.

At this point, I found a way, but it requires changes in gem source, particularly adding parameter `(description: self.description)` into method call `GraphQL::Field.define do` in here:

```ruby
#lib/graphql/relay/mutation.rb

  def field
    @field ||= begin
      # ...
      GraphQL::Field.define(description: self.description) do
      # ...
      end
    end
  end
```

Here's the result after editing the gem: http://joxi.ru/BA0pKJBFBOkRBA

If you like the way I suggested, I can leave a PR.

Thanks and looking forward to your thoughts on this!
Alex

rmosolgo:
Huh, it looks like that description is not used at all, oops! Your fix sounds good to me!

b8in:
Glad to hear that! I'm happy to be of help :)
Then, I'll leave a PR today or tomorrow already.

b8in:
Thanks, updated the gem (v 0.18.7) on my side, everything's ok.

PR #225, opened 2016-08-31
issue #224

b8in:


rmosolgo:
This is great, do you mind adding a test to be sure we maintain this in the future? For example, you might add one here: 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/relay/mutation_spec.rb

that asserts that the mutation field's description is present, eg 

```ruby 
it "applies the description to the derived field" do 
   assert_equal "Add a ship to this faction", IntroduceShipMutation.field.description
end 
```

b8in:
Sure, with pleasure :)

b8in:
I've just commited the test you mentioned above.
I didn't try to write any other checks - I think that one would be enough.

rmosolgo:
awesome, thanks for this improvement!!

rmosolgo:
:ship: in 0.18.7

b8in:
Thanks, updated the gem on my side, everything's ok.

PR #226, opened 2016-08-31
Move the dairy schema into the library for shared testing

staugaard:
This moves the dairy schema into the library, so that it can be used in the shared tests.

I also moved the parser tests to `lib/graphql/testing/parser_tests.rb`


rmosolgo:
:hand: I consider this part of the test suite to be a mistake :( Here are some issues I have: 

- It's too personal: I love cheese, but now I feel embarrassed making people work on this when they add tests 😬 
- It's not localized enough: sometimes, you have to modify this schema for a certain kind of test (for example, create some complex return type and test the result), and your change breaks other tests which were hardcoded to expect a certain (different) structure. 
- As a result of those two, it makes no sense: weird things have been tacked on over time so that the schema has a nonsensical structure and is full of nonsensical data. 

So, I suggest writing _another_ test schema for `GraphQL::Testing`, one whose scope is limited to `GraphQL::Testing`. And then in the same vein, the tests in `GraphQL::Testing` would only use this new schema. This would decouple `GraphQL::Testing` from `graphql-ruby`'s own test suite, that way the errors described above don't "leak" into other projects. 

What do you think? Are those issues above "real"? Would that suggestion work in this case?

rmosolgo:
In other news, :+1: for the move to `GraphQL::Testing` and the rename of `DummySchema => DairySchema`, that was also a mistake!

rmosolgo:
I think this need is covered by `GraphQL::Compatibility` in 1.1.0, please feel free to open another issue if something is still missing!

PR #230, opened 2016-09-05
A few Schema loader bug fixes

josh:
1. Fix "Schema.new is deprecated, use Schema.define instead" deprecation warning.
2. Fix loading `ObjectType`'s interfaces
3. Fix loading other root types: mutation and subscriptions

I discovered the interface bug after switching from `Schema.new` to `Schema.define`. As `Schema.define` discovers all types from the roots, I noticed that the interface in the test file was no longer being discovered. So this case already has test coverage.

### DIFF
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -65,7 +65,7 @@ module GraphQL
             InterfaceType.define(
               name: type["name"],
               description: type["description"],
-              fields: Hash[type["fields"].map { |field|
+              fields: Hash[(type["fields"] || []).map { |field|
                 [field["name"], define_type(field.merge("kind" => "FIELD"), type_resolver)]
               }]
             )
@@ -81,6 +81,9 @@ module GraphQL
             ObjectType.define(
               name: type["name"],
               description: type["description"],
+              interfaces: (type["interfaces"] || []).map { |interface|
+                define_type(interface.merge("kind" => "INTERFACE"), type_resolver)
+              },
               fields: Hash[type["fields"].map { |field|
                 [field["name"], define_type(field.merge("kind" => "FIELD"), type_resolver)]
               }]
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/loader.rb
```diff
@@ -22,9 +22,13 @@ def load(introspection_result)
...
 
-        query = types.fetch(schema.fetch("queryType").fetch("name"))
+        kargs = {}
+        [:query, :mutation, :subscription].each do |root|
+          type = schema["#{root}Type"]
+          kargs[root] = types.fetch(type.fetch("name")) if type
+        end
 
-        Schema.new(query: query, types: types.values)
+        Schema.define(**kargs)
```
rmosolgo: `types` was renamed to `orphan_types` inside `.define` because of the naming conflict with the `types` helper object (eg `types.String`). 

It's too bad that the new name doesn't imply that "it's ok to just put all the types here and we'll sort it out later". Maybe it could be improved!

rmosolgo:
This case was previously covered by the use of `types:`, but I think it's not covered currently: 

```graphql
type Query {
  searchMedia(term: String!): Media
}

interface Media { 
  # whatever fields 
}

type Video implements Media {
  # whatever fields
}

type Audio implements Media {
  # whatever fields 
}  
```

(Starting from `Query`, we can find `Media`, but since `Media` contains no references to its implementing types, we can't find `Audio` or `Video`).

I think it would be covered by using `orphan_types:` where types was previously used, or is it covered another way? 

Two nice-to-have-but-optional changes are: 
- adding a test for this case 
- brainstorming a better name than `orphan_types` since it's not necessarily orphans 

But in my opinion, those are optional, since this PR is just to fix the warning, which is an improvement in itself!

josh:
> adding a test for this case

👍 good call. Added test and fixed that `Media` case by passing `orphan_types`.

rmosolgo:
💰 !

josh:
Thanks!

PR #231, opened 2016-09-05
Implement Node value equality

josh:
In Ruby, `Object#equal?` is true reference equality while `Object#eql?` exists for *value* equality.

This implements a deep value equality check for any `AbstractNode`. This allows you to compare the results from `GraphQL.parse` to see if they produce an identical AST.

`Object#eql?` is also used for `Set#include?` checks, so this allows for:

``` ruby
validated_documents = Set.new

doc = GraphQL.parse(str)

# check if we already validated this AST, otherwise we can skip
if !validated_documents.include?(doc)
  validate_document!(doc)
  validated_documents << doc
end
```

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -26,6 +26,7 @@ module GraphQL
         end
 
         def eql?(other)
+          return true if equal?(other)
           other.is_a?(self.class) &&
             other.scalars.eql?(self.scalars) &&
             other.children.eql?(self.children)
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -32,7 +32,17 @@ module GraphQL
             .flatten
         end
 
+        def scalars
+          self.class.scalar_attributes
+            .map { |attr_name| public_send(attr_name) }
+        end
+
         class << self
+          def scalar_attributes(*attr_names)
+            @scalar_attributes ||= []
+            @scalar_attributes += attr_names
+          end
+
           def child_attributes(*attr_names)
             @child_attributes ||= []
             @child_attributes += attr_names
@@ -50,6 +60,8 @@ module GraphQL
 
       class WrapperType < AbstractNode
         attr_accessor :of_type
+        scalar_attributes :of_type
+
         def initialize_node(of_type: nil)
           @of_type = of_type
         end
@@ -61,6 +73,8 @@ module GraphQL
 
       class NameOnlyNode < AbstractNode
         attr_accessor :name
+        scalar_attributes :name
+
         def initialize_node(name: nil)
           @name = name
         end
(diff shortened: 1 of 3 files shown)


lib/graphql/language/nodes.rb
```diff
@@ -25,14 +25,35 @@ def initialize_node(options={})
...
+            .map { |attr_name| public_send(attr_name) }
+        end
+
         class << self
+          def inherited(subclass)
+            subclass.scalar_attributes(*@scalar_attributes)
+            subclass.child_attributes(*@child_attributes)
+          end
+
+          def scalar_attributes(*attr_names)
```
josh: I declared a new attribute helper similar to `child_attributes` for annotating attributes pointing at scalars like `String`.

lib/graphql/language/nodes.rb
```diff
@@ -25,14 +25,35 @@ def initialize_node(options={})
...
+        def scalars
+          self.class.scalar_attributes
+            .map { |attr_name| public_send(attr_name) }
+        end
+
         class << self
+          def inherited(subclass)
+            subclass.scalar_attributes(*@scalar_attributes)
+            subclass.child_attributes(*@child_attributes)
+          end
```
josh: This was an existing bug in `child_attributes` as well. If a subtype extends another class like `class ListType < WrapperType; end`, it won't inherit class ivars. You always need to copy them. But that ensures they get their own copy of the Array. Which is good.

lib/graphql/language/nodes.rb
rmosolgo: :+1:

lib/graphql/language/nodes.rb
rmosolgo: :+1: for `scalar_attributes` 

For `child_attributes`,  I didn't include it because I didn't need it. Is there any place where that code is (or could be) used? 

If you want to include it for consistency, that's ok with me, but I want to make sure it's tested somehow, either by use "in the wild" (ie, some of the node classes inherit child attributes) or in a unit test with a dummy subclass.  What do you think?

rmosolgo:
I've always been spooked by Ruby's many equality methods, but it sounds great to support cases like the one you described!

lib/graphql/language/nodes.rb
josh: > Is there any place where that code is (or could be) used?

Hmm, I guess you're right, all the existing node subclasses don't have any child attrs. So maybe its fine.

josh:
Thanks for the feedback, @rmosolgo any other change suggestions?

rmosolgo:
Is `subclass.child_attributes(*@child_attributes)` used or tested in any way (either in isolation or by way of being used in the library)? I think that's the only outstanding thing here -- I think it should be removed if it's not used or tested.

josh:
> Is subclass.child_attributes(*@child_attributes) used or tested in any way (either in isolation or by way of being used in the library)?

Ah, okay, I'll add some isolated tests for these guys.

josh:
Okay, pushed a test case.

rmosolgo:
🚀

rmosolgo:
🚢 in 0.18.8

PR #232, opened 2016-09-05
Allow parsing anonymous fragment definitions

josh:
I'm trying to find a good way to work with "anonymous" fragment definitions.

While not part of the official GraphQL grammar, Relay makes heavy use of anonymous fragments.

``` js
Tea = Relay.createContainer(Tea, {
  fragments: {
    tea: () => Relay.QL`
      fragment on Tea {
        name,
        steepingTime,
      }
    `,
  },
});
```

The use case is taking these anonymous fragments and then automatically generating names for them and putting them into a new document to actually use for the query. That latter probably doesn't belong in this library itself, but I'd still love to be able to use the core parser without having to write my own extended syntax parser.

<del>I spiked out a parser extension, though it doesn't actually pass the tests.</del>


rmosolgo:
:+1: for the idea, but we should make sure we have a validation in place to reject queries that contain unnamed fragments, right?

josh:
> but we should make sure we have a validation in place to reject queries that contain unnamed fragments

Good call, added `GraphQL::StaticValidation::FragmentsAreNamed`.

Do you have any idea what's wrong with the parser extension I added. I'm not that familiar with ragel.

josh:
> Do you have any idea what's wrong with the parser extension I added. I'm not that familiar with ragel.

Ah, figured it out.

lib/graphql/static_validation/rules/fragments_are_used.rb
```diff
@@ -38,7 +38,7 @@ def add_errors(context, used_fragments, defined_fragments)
       end
 
       def find_difference(fragments, allowed_fragment_names)
-        fragments.select {|f| !allowed_fragment_names.include?(f.name) }
+        fragments.select {|f| f.name && !allowed_fragment_names.include?(f.name) }
```
rmosolgo: :clap: :clap: on working with StaticValidation! There should be some kind of badge for dealing with this code.

This change (required, totally the right thing to do) is such a code smell. In this design, they're order-dependent in ways that are hard to understand and hard to test. Hoping to fix that one day 😖

rmosolgo:
:ship: in 0.18.7

PR #233, opened 2016-09-06
Resolve __type to null for type that doesn't exist

dylanahsmith:
## Problem

The `__type` introspection field in the GraphQL spec is

```
__type(name: String!) : __Type
```

but graphql-ruby made it a non-null type, which gets a runtime error when the type isn't found.

## Solution

Make the result type nullable and return null if the type doesn't exist


rmosolgo:
Thanks!

rmosolgo:
:ship: in 0.18.7

PR #235, opened 2016-09-08
fix(connection) use the field name, not the module name

### LINKED ISSUE #234: Weird `AssignConnection` leakage
Suppose you have a `connection` defined on your root query. If you list the fields off of this root query, you get an interesting extra object:

```
$ Schema.types["RootQuery"].fields.keys
> ["node", "viewer", "connection", "GraphQL::Define::AssignConnection"]
```

I'm not sure what `GraphQL::Define::AssignConnection` is. It doesn't _seem_ to be exposed when trying to make requests to the GraphQL server, so is it harmless?

rmosolgo:
Derp, this was definitely from a faulty refactor, let me try to find that commit. Fixes #234 

Amazingly, it still worked because `ConnectionField` _modifies_ the incoming field instead of creating a new one. I guess if it had made a new field, we would have noticed this right away though, since tests would have broken.

cc @gjtorikian

### DIFF
--- a/lib/graphql/define/assign_connection.rb
+++ b/lib/graphql/define/assign_connection.rb
@@ -4,7 +4,7 @@ module GraphQL
       def self.call(type_defn, *field_args, max_page_size: nil, **field_kwargs, &field_block)
         underlying_field = GraphQL::Define::AssignObjectField.call(type_defn, *field_args, **field_kwargs, &field_block)
         connection_field = GraphQL::Relay::ConnectionField.create(underlying_field, max_page_size: max_page_size)
-        type_defn.fields[name.to_s] = connection_field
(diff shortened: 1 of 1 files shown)


brandonblack:
Thanks @rmosolgo

rmosolgo:
Wow, it must have been introduced in `graphql-relay` who-knows-when, it was migrated over here _with_ the bug https://github.com/rmosolgo/graphql-ruby/commit/ee1b4528411023a11031d6ca8f2dead638226012

rmosolgo:
Sorry, I wrote the test for this but forgot to commit it! I've pushed it in 26a56f7

rmosolgo:
:ship: in 0.18.8,  let me know if you'd like this backported!

PR #239, opened 2016-09-09
feat(Relay::Mutation) Put a reference to the original mutation inside the derived object

rmosolgo:
Currently, you can access derived objects from the `Mutation` instance. But if you have a derived object, you have no way to "reach back" for the mutation which generated it. This has become valuable since the mutation can hold user-defined metadata. 

So, provide access to the mutation from the derived object:

```ruby 
derived_field.mutation # => #<GraphQL::Relay::Mutation ... >
```


rmosolgo:
🚢  in 0.18.10

brandonblack:
Quick draw @rmosolgo with yet another fast release! 😁  🔫

rmosolgo:
I better slow down, I'm gonna exceed the limit for my free plan for RubyGems 😝

PR #240, opened 2016-09-10
Fix leaked global constant

josh:
Constants defined in blocks are hoisted into the parent module scope. In this `GraphQL::BOOLEAN_TYPE` case, the global `Object` scope.

``` ruby
require 'graphql'
GraphQL::BOOLEAN_TYPE.name
::ALLOWED_INPUTS #=> [true, false]
```

In this case I simply replaced the constant with an equality test. Its probably faster in some micro benchmark since it doesn't need to go through `Array#include?` first.


rmosolgo:
Oopsie! You're right, plain ol' equality sounds like a better choice anyways.

Issue #236, opened 2016-09-08
Document "slicing" API idea


josh:
Hey @rmosolgo!

I have another API *we're* using and I wanted to ask you think it's worth upstreaming to graphql-ruby core.

When working with a single monolith document with all the operations and fragment definitions, its useful to extract slices of the document given a root operation and pulling in all its fragment definitions.

For an example:

``` ruby
document = GraphQL.parse <<-'GRAPHQL' 
query FooQuery { ...FooFragment }
query BarQuery { ...BarFragment }
query BazQuery { ...BarFragment }
mutation FooMutation { ... }
fragment FooFragment on Foo { ... }
fragment BarFragment on Bar { ... }
GRAPHQL

sliced_document = slice(document, operation_name: "FooQuery")
sliced_document == <<-'GRAPHQL'
query FooQuery { ...FooFragment }
fragment FooFragment on Foo { ... }
GRAPHQL
```

Where `sliced_document` is the minimal set of definitions to fulfill the `FooQuery` operation.

With Relay 2 and an attempt to standardize persisted queries, executing operations on large documents will become more common than always posting small queries in production environments.

I have two use cases for this sort of algorithm:

1. If client maintains a concatenated document of all its queries, it may prefer sending a minimal slice of the operation definition to minimal the size of the request body. Rather than sending the entire document every time and specify an operation name parameter.

``` ruby
http_request("POST", "https://example.com/graphql?operation=FooQuery", body: large_document)

# post small document slice instead
small_document = slice(large_document, operation_name: "FooQuery")
http_request("POST", "https://example.com/graphql", body: small_document)
```

2. If the server supports persisted queries, it will likely be storing large GraphQL documents. To aid server error debugging and reporting, the server may want to dump a smaller query in its logs that is sufficient to reproduce the error.

``` ruby
operation_name = request.params["operation_name"]
document = load_document_from_database(request.params["document_id"])

# add simplified document to our exception reporter so we know what original
# query caused the exception.
AppErrorContext.push query: slice(document, operation_name).to_query_string

query = GraphQL::Query.new(MySchema, document: document, operation_name: operation_name)
query.result
```

The gist of the algorithm recursively walks from the operation to any fragment spreads, then each fragment dependency, all of its fragment dependencies.

I was also wondering if this process was already encoded somewhere in the query executor (I assume it does a similar thing) and could be exposed as a public API.

Ha, well, even if it doesn't belong in this lib, hopefully you find this use case interesting.

[...]

PR #241, opened 2016-09-10
Definition slicing

josh:
Follow up discussion from #236.

``` ruby
document = GraphQL.parse <<-'GRAPHQL' 
query FooQuery { ...FooFragment }
query BarQuery { ...BarFragment }
query BazQuery { ...BarFragment }
mutation FooMutation { ... }
fragment FooFragment on Foo { ... }
fragment BarFragment on Bar { ... }
GRAPHQL

sliced_document = document.slice_definition("FooQuery")
sliced_document == <<-'GRAPHQL'
query FooQuery { ...FooFragment }
fragment FooFragment on Foo { ... }
GRAPHQL
```

I think the name "slice" is okay, but I don't love it. Trying to brainstorm some other terminology for this algorithm.

### DIFF
--- a/lib/graphql/language/definition_slice.rb
+++ b/lib/graphql/language/definition_slice.rb
@@ -1,0 +1,29 @@
+module GraphQL
+  module Language
+    module DefinitionSlice
+      extend self
+
+      def slice(document, name)
+        definitions = {}
+        document.definitions.each { |d| definitions[d.name] = d }
+        names = find_definition_dependencies(definitions, name)
+        definitions = document.definitions.select { |d| names.include?(d.name) }
+        Nodes::Document.new(definitions: definitions)
+      end
+
+      private
+
+      def find_definition_dependencies(definitions, name)
+        names = Set.new([name])
+        visitor = Visitor.new(definitions[name])
+        visitor[Nodes::FragmentSpread] << -> (node, parent) {
+          if fragment = definitions[node.name]
+            names.merge(find_definition_dependencies(definitions, fragment.name))
+          end
+        }
+        visitor.visit
+        names
+      end
+    end
+  end
+end
--- a/lib/graphql/language.rb
+++ b/lib/graphql/language.rb
@@ -1,3 +1,4 @@
+require "graphql/language/definition_slice"
 require "graphql/language/generation"
 require "graphql/language/lexer"
 require "graphql/language/nodes"
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -129,6 +129,10 @@ module GraphQL
         def initialize_node(definitions: [])
           @definitions = definitions
         end
+
+        def slice_definition(name)
+          GraphQL::Language::DefinitionSlice.slice(self, name)
+        end
       end
 
       class Enum < NameOnlyNode; end
--- a/spec/graphql/language/definition_slice_spec.rb
+++ b/spec/graphql/language/definition_slice_spec.rb
@@ -1,0 +1,225 @@
+require "spec_helper"
+
+describe GraphQL::Language::DefinitionSlice do
+  let(:document) { GraphQL::Language::Parser.parse(query_string) }
+
+  describe "anonymous query with no dependencies" do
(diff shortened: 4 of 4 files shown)


lib/graphql/language/definition_slice.rb
```diff
@@ -0,0 +1,29 @@
...
+        names = find_definition_dependencies(definitions, name)
+        definitions = document.definitions.select { |d| names.include?(d.name) }
+        Nodes::Document.new(definitions: definitions)
+      end
+
+      private
+
+      def find_definition_dependencies(definitions, name)
+        names = Set.new([name])
+        visitor = Visitor.new(definitions[name])
```
rmosolgo: wow, i'm happily surprised that this use of `Visitor` actually works 😅

rmosolgo:
This is sweet! I wonder what other names it could be ... 

```ruby 
document.pluck("FooQuery") # too Rails-y
document.get_named_operation("FooQuery") # too icky 
document.fetch("FooQuery") # too vague 
document.extract("FooQuery")  # same hash-y feel as `slice`, but ... wrong :P 
``` 

???

rmosolgo:
Just an idea to knock around, how does this implementation compare to one where: 

- the `Document` maintains a "sidecar" object like `DependencyTree`, eg: 
  ```ruby 
  def dependency_tree
    @dependency_tree ||= DependencyTree.new(self) 
  end 
  ```
- During initialization, `DependencyTree` traversed the _whole_ document once, but maintained a reference of _every_ definition's dependencies 
- calling `.slice(op_name)` delegated to `DependencyTree` for the named operation & its dependencies (but after `DependencyTree`'s initialization, it doesn't have to do any traversal) 

In my mind, an implementation like that is better only in a specific case: you have a BIG document which you have to fetch things from over and over. AND it only event makes a difference if it's on a "hot path" -- if it's just during app initialization ... then ... who cares? (Ok, not "who cares", but, "make it work, then make it fast".)

Are there other advantages or downsides of an all-at-once approach that I'm overlooking?

gjtorikian:
> I wonder what other names it could be ...

I like `pick`. You pick up fragments from the floor after you break a vase. 😅

josh:
> the Document maintains a "sidecar" object like DependencyTree, eg: ruby def dependency_tree @dependency_tree ||= DependencyTree.new(self) end

Yeah, I've really wanted some kind of *indexes* on `Document` for a number of cases. Even `document["FooQuery"] => Definition` would be handy.

The one issue with memoizing that *index* is responding to mutations to the node.

``` ruby
document = GraphQL.parse "query Foo { ... }"
document.dependency_tree["Foo"]

other_document = GraphQL.parse "query Bar { ... }"
document.definitions.concat(other_document.definitions)

# How would this work?
document.dependency_tree["Foo"]
document.dependency_tree["Bar"]
```

This is related to Nodes being mutable. I may try to resurrect my [old "deep freeze" PR](https://github.com/rmosolgo/graphql-ruby/pull/156) to provide an optional immutable state for this tree such that we can ensure memoization is safe.

rmosolgo:
> mutations to the node

Oh, I forgot! That's a good reason to use this implementation. 

Another option would be, instead of caching, to provide some object for making that calculation, eg 

```ruby
dependency_tree = DependencyTree.new(document) 
foo_query = dependency_tree.pick("FooQuery") # your method name here :P 
```

Yes, technically, the dependency tree could be held onto while the document is modified, but such is life! At least in this case, we're not _inviting_ mutability bugs 😅 

Happy to keep this implementation since it works for you! The API and implementation are both well-positioned for a later refactor if we end up needing one.

rmosolgo:
Oh, and for any curious reader, here's the dependency resolution used for rewriting ASTs into query trees: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/internal_representation/rewrite.rb#L147-L163

josh:
Cool.

I guess I'm fine with the name. Maybe we'll come up with something better later or roll it into a dependency resolution class if we think of other features that it would enable.

Thanks!

PR #243, opened 2016-09-10
refactor(Schema) move GlobalNodeIdentification to Schema

rmosolgo:
`GraphQL::Relay::GlobalNodeIdentification` was a direct port from `graphql-relay-js`: 

image

As graphql-ruby has matured, and since merging graphql-relay-ruby, it has become out of place and unnecessary -- the schema could be the "source of truth" for these functions. 

So this PR removes `GlobalNodeIdentification` and relocates some of its responsibilities: 

- Many object resolution functions were moved to `Schema`: `to_global_id`, `from_global_id`, `id_separator`, `object_from_id`
- The node lookup field is available from `GraphQL::Relay::Node.field` 
- The node interface is available from `GraphQL::Relay::Node.interface`


theorygeek:
Just some food for thought about global node ID's. We ended up rolling our own node identification system, to support a couple of use-cases. 

#### 1. Sometimes we need more than just Type + ID
Most of our GraphQL types are backed by ActiveRecord models, so node identification is just a matter of know the type and ID of the model. But sometimes, they're "computed" objects. We re-construct these by using the type of the object, and then some arbitrary parameters. A typical example looks like this:

```ruby
class SomeCustomClass
  def to_id_attrs
    # return some hash
  end

  def self.from_id_attrs(attrs)
    # reconstruct the object given the same hash
  end
end
```

We generate the ID's by serializing that info as JSON, and then Base64 encoding it. The only real difference between this and the built-in node ID system is that instead of using only an ID, we use a couple of pieces of information.

#### 2. Preventing ID's from being synthesized by outside code
A second concern was that we didn't want users to be able to arbitrarily request objects from our API, by synthesizing the ID's. So, we cryptographically sign all of our global object ID's, and verify those signatures before we reconstruct the object.

That helps us with security in numerous ways, since you can't get an object from us that you didn't already have. But it also simplifies the code needed to reconstruct the objects, since we can count on having been given correct information.

So having said all of that, I think it's worth considering whether global object ID's really needs to be part of the core `graphql` gem. It's not very difficult to implement, and it seems like one of those features where different apps could have pretty different needs.

Just wanted to give some of that feedback, hopefully it's helpful. :)

rmosolgo:
Thanks for sharing your use of those APIs! 

I hear your point about removing the "default" global ID implementation. It's so easy to implement, you might as well let people implement it themselves, so that the understand how the system works and they'll be able to build a more robust implementation when the time comes. 

Maybe this change is a good time to start removing the default implementation, but provide the same function so current users can opt in. For example 

```ruby 

GraphQL::Schema.define do 
  # ... 
  # please forgive the names
  to_global_id -> (obj, ctx) { GraphQL::GlobalIdFromClassNameAndMethod.call(obj, :id) } 
  from_global_id -> (id, ctx) { GraphQL::FindByClassNameAndId.call(id) } 
end 
```

That way, a reader can see how the ids actually work, and can at least follow the codepath to find the implementation provided by the gem.

rmosolgo:
Also, that's a good point about encryption. That, and _versioning_ for cursors, seem like "next-level" Relay identifier techniques that could be supported out-of-the-box by graphql-ruby!

rmosolgo:
I just pushed a refactor that provides these two hooks on a schema: 

```ruby 
object_from_id -> (id, ctx) { ... } 
id_from_object -> (object, type, ctx) { ... }
```

It supports the old approach (via `GraphQL::Schema::UniqueWithinType`) and more, and removes the useless intermediary `from_global_id`

PR #244, opened 2016-09-10
feat(Introspection) sort lists by name

rmosolgo:
For a normal human, the introspection result is not useful because the results are in a meaningless order. It could be better for normal humans if they're sorted. 

However, this is basically the wrong approach. Assuming your endpoint executes an introspection query _every time_ someone opens up GraphiQL, we should _maintain_ sorted lists instead of sorting them _again_ every time!

__TODO__

- [ ] find a way to not sort every time (each object maintains sorted members? maintain a sorted cache in `Introspection` somewhere?)

### DIFF
--- a/lib/graphql/introspection/arguments_field.rb
+++ b/lib/graphql/introspection/arguments_field.rb
@@ -1,5 +1,5 @@
 GraphQL::Introspection::ArgumentsField = GraphQL::Field.define do
   description "Arguments allowed to this object"
   type !GraphQL::ListType.new(of_type: !GraphQL::Introspection::InputValueType)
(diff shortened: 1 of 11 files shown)


rmosolgo:
Hmm, maybe this is the best we can do ... Or maybe it's not worth it? It's hard to maintain a sorted list of an object's fields because it's hard to tell when `.fields[...]=` is assigned. 

You _could_ make the assumption that types won't be modified after the schema is initialized, then have a schema-level cache for various introspection-y things (like sorted lists of values). I think that's a fair assumption.

rmosolgo:
I'm gonna close this until someone actually wants it.

Issue #186, opened 2016-07-19
Rethink middleware


rmosolgo:
Middleware was designed with a simple, Rack-like API, but turns out that's a bug, not a feature: 
- http://blog.plataformatec.com.br/2012/06/why-your-web-framework-should-not-adopt-rack-api/
- http://tenderlovemaking.com/2011/03/03/rack-api-is-awkward.html

Similar issues affect `graphql-ruby`. For example, if you use graphql-batch, the actual processing is done "out of bounds", not within `next_middleware.call`. With [`DeferredExecution`](https://github.com/rmosolgo/graphql-ruby/pull/168), the problem is even worse: there's no way to detect the _actual_ end of the query, since any deferred fields are executed _outside_ the root node's middleware call. 

So, we need a system that: 
- can intercept field resolves  
- can detect the start & end of queries (even when processing is done "out of bounds", eg `graphql-batch`)
- supports middleware-local state (eg, timing data)
- has a user-friendly API 

Whatever happens, it will be important to support existing middleware somehow. I imagine that will be simple enough, wrapping an existing middleware to port it to the new API, perhaps 

``` ruby
legacy_middleware = CustomMiddleware.new 
new_middleware = LegacyMiddlewareAdapter.new(legacy_middleware)
MySchema.middleware << new_middleware 
```
#### Make it like query analyzers
- `before_query` returns an initial memo 
- subsequent hooks return a new memo 
- **problem:** Previous middlewares could affect field resolution by returning a value and _not_ calling `next_middleware.call`. This implementation requires the return value to be `memo`, how can you intercept a field call? 

``` ruby
MyCustomMiddleware = GraphQL::Middleware.define do 
  before_query -> (memo, env) { 
    { 
      counter: 0
     }
  }

  each_field -> (memo, env) {
    memo[:counter] += 1 
    env[:next_middleware].call 
    memo 
  } 

  after_query -> (memo, env) { 
    puts "Total fields: #{memo[:counter]}"
  }
end 
```
#### Make it an event listener
- Execution strategy would be responsible for triggering events at the right time
- easy to add new events in the future 
- **problem:** again, how to modify field resolutions? (I don't want to [use exceptions for control flow](http://programmers.stackexchange.com/a/189225))

``` ruby
MyCustomMiddleware = GraphQL::Middleware.define do 
  on(:begin_query) { ... } 
  on(:end_query) { ... } 
  on(:begin_field) { ... }
  on(:end_field) { ... }
end 
```
[...]

PR #245, opened 2016-09-10
Add field usage query analyzer

cjoudrey:
Follow up from a discussion over Slack with @rmosolgo.

When maintaining a GraphQL API it's important to know what fields are being used, especially deprecated fields.

This adds a simple query analyzer that exposes both `used_fields` and `used_deprecated_fields`.

These arrays can then be fed into other systems in order to know if removing a deprecated field is _safe_ and resurfaced to API consumers to let them know they are using a deprecated field.

I'm not sure if this belongs in the repo, but at the very least could serve as an example in docs for someone wanting to build a query analyzer.

@rmosolgo any thoughts?


rmosolgo:
:+1: love it! I have a few thoughts: 

- Is it possible to add a query error from the context of this analyzer? Or is there another way to notify the client as a result of using a field? I _think_ you could push an entry into `query.context.errors`, but I forget if that array holds error instances or hashes.
- I noticed you're pushing one string into `used_fields` and another into `used_deprecated_fields`. Can we put the same object in both arrays? I know it's splitting hairs, but this is a "hot path" so I figured I'd ask!
- I wonder if there's benefit to pushing _whole objects_ into the usage arrays, instead of just strings. For example, if you only wanted to log deprecated fields on certain types, then you'd be able to filter the usages. I can imagine introducing some intermediary object (but I can only think of the name `FieldUsage` 😆 ) that contains references to the `field_defn` and `type_defn`. What do you think, any advantage there, or some reason it's no good?

cjoudrey:
> Is it possible to add a query error from the context of this analyzer? Or is there another way to notify the client as a result of using a field? I think you could push an entry into `query.context.errors`, but I forget if that array holds error instances or hashes.

I'm not sure if you'd want to add a query error. Deprecated fields are meant to be an intermediary state between a field that is available and a field that is removed, no? I believe client libraries should add a warning log when a deprecated field is being queried. 

What I meant by "resurfaced to API consumers" was something like a weekly e-mail to API consumers "Your app is using X, Y, Z deprecated fields" or via an admin panel that API consumers have access to.

> Can we put the same object in both arrays? I know it's splitting hairs, but this is a "hot path" so I figured I'd ask!

:+1: Absolutely!

> I wonder if there's benefit to pushing whole objects into the usage arrays, instead of just strings. For example, if you only wanted to log deprecated fields on certain types, then you'd be able to filter the usages.

Good question. I hadn't thought of that as I'm mostly interested in logging all deprecated fields and all used fields and this data is then processed by another system which is why I went with strings. 

Alternatively, I suppose we could allow passing a `filter` lambda when initializing the analyzer. I'm not sure if this should be added right away though. What do you think?

rmosolgo:
Yes, I guess you're right about errors, that's not really a good way to communicate the deprecation thing. 

Happy to leave `FieldUsageEntry` to a later refactor as YAGNI. 

One last thought: 

Right now this uses `irep_node.definitions` which _can_ contain more than one field definiton. For example, I think this one would cause that: 

```graphql
{
  node(id: $myNodeId) {
    ... on Cheese { id }
    ... on Milk    { id }
  }
}
```

I think this would cause: 

```
Used fields: Milk.id, Cheese.id
```

That seems _right_ to me: from the query analysis point of view, you don't _know_ which objects will be used at runtime, so you have to mark both as used. And if either one is deprecated, then you want to know, since it _is_ present in the query. Does that sound right to you, is that what you had in mind also?

rmosolgo:
:metal:

cjoudrey:
🎉 

We should probably revisit this once https://github.com/rmosolgo/graphql-ruby/issues/186#issuecomment-247412158 has landed.

rmosolgo:
:ship: in 0.18.12

PR #246, opened 2016-09-11
Deep freeze nodes

josh:
Return of #156.

Addresses the original issue of scalars not being frozen. https://github.com/rmosolgo/graphql-ruby/pull/156#issuecomment-220033187

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -68,6 +68,20 @@ module GraphQL
         def to_query_string
           Generation.generate(self)
         end
+
+        def freeze
+          self.class.child_attributes.each do |attr_name|
+            public_send(attr_name).freeze.each(&:freeze)
+          end
+
+          self.class.scalar_attributes.each do |attr_name|
+            if object = public_send(attr_name)
+              object.freeze
+            end
+          end
+          
+          super
+        end
       end
 
       class WrapperType < AbstractNode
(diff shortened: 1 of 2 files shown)


rmosolgo:
I guess what I'm missing is, of the the places where mutable data structures _can_ be misused, why try to "batten down the hatches" on this particular one? 

I'm not used to programming with frozen objects, is there other library whose objects work this way that I can see for reference?

josh:
> I guess what I'm missing is, of the the places where mutable data structures can be misused, why try to "batten down the hatches" on this particular one?

My main use case is that I want to be able to share AST subtrees. Where I may have many documents each pointing to the same definition nodes to save memory. Its not always obvious when you pass a document instance to some function (even in ruby-graphql) that you know its not gong to be mutated.

But mutations are useful and I'm making use of the fact they are mutable by default. My typical pattern looks like:

``` ruby
doc = GraphQL.parse(str)
transform1(doc)
transform2(doc)
transform3(doc)

# done making changes, this should no longer be mutable
doc.freeze
```

> I'm not used to programming with frozen objects, is there other library whose objects work this way that I can see for reference?

http://www.rubypigeon.com/posts/a-review-of-immutability-in-ruby/

Looking forward, Ruby 2.3 now has immutable String literals and maintainers are proposing more immutable structs as a basis for Ruby 3 concurrency.

rmosolgo:
Hey, I heard you're on your way back from Ruby Kaigi -- I'm in SF until Friday evening, depending on your jetlag, maybe we can chat about this in person!

josh:
Alright, bailing again, maybe later I'll reopen if third time is the charm 😁

PR #247, opened 2016-09-11
Validation: Report field doesn't exist error on field, not parent

josh:
Unless I'm missing some edge case, doesn't it make more sense to report the line/column of the invalid field than the parent object?

``` graphql
query {          # 1
  viewer {       # 2
    id           # 3          
    missingField # 4
  }              # 5
}                # 6
```

I'd expect `"Field 'missingField' doesn't exist on type 'User'"` on line `4` rather than line `2`.


rmosolgo:
yup!

PR #249, opened 2016-09-14
Tiny fix in security guide

dmitry:


rmosolgo:
tiny fix, but it's a __big__ difference! Thank you!

PR #251, opened 2016-09-15
feat(Arguments#key?) show when arguments were missing from the query

rmosolgo:
Only put keys in the `Arguments` instance if the key was present in the query in some way. And use `#key?` to show whether it was present or missing.

### DIFF
--- a/lib/graphql/query/literal_input.rb
+++ b/lib/graphql/query/literal_input.rb
@@ -16,14 +16,25 @@ module GraphQL
         values_hash = {}
         argument_defns.each do |arg_name, arg_defn|
           ast_arg = ast_arguments.find { |ast_arg| ast_arg.name == arg_name }
-          arg_value = nil
-          if ast_arg
-            arg_value = coerce(arg_defn.type, ast_arg.value, variables)
+          arg_default_value = arg_defn.type.coerce_input(arg_defn.default_value)
+          if ast_arg.nil? && arg_default_value.nil?
+            # If it wasn't in the document,
+            # and there's no provided default,
+            # then don't pass it to the resolve function
+            next
+          else
+            arg_value = nil
+
+            if ast_arg
+              arg_value = coerce(arg_defn.type, ast_arg.value, variables)
+            end
+
+            if arg_value.nil?
+              arg_value = arg_default_value
+            end
+
+            values_hash[arg_name] = arg_value
           end
-          if arg_value.nil?
-            arg_value = arg_defn.type.coerce_input(arg_defn.default_value)
-          end
-          values_hash[arg_name] = arg_value
         end
         GraphQL::Query::Arguments.new(values_hash)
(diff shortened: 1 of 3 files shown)


lib/graphql/query/literal_input.rb
```diff
@@ -17,13 +17,17 @@ def self.from_arguments(ast_arguments, argument_defns, variables)
         argument_defns.each do |arg_name, arg_defn|
           ast_arg = ast_arguments.find { |ast_arg| ast_arg.name == arg_name }
           arg_value = nil
-          if ast_arg
+          if ast_arg.nil?
```
josh: Ah, this was the change I was trying to find.

staugaard:
Is this really the right logic?

I thought it should be:

If the argument is not present, then use the default value.
If the argument is present, then use that value and not the default value even when the value is null.

rmosolgo:
Oops, apparently I totally broke default argument values! Pushed a some fresh code to handle that case, thanks @staugaard 

With that in mind, the logic I implemented is like this: 

- If an argument is present in the query (either in the query string or in the variables) and its value is not `nil`, assign it to the `Arguments`
- Else, if an argument has a default value, assign the default value to the `Arguments`
- Else, don't assign anything to `Arguments` (so that the key is absent)

I think that, currently, `null` is not allowed as a literal input or as a variable input. But that "feature" is being debated here: https://github.com/facebook/graphql/pull/83

staugaard:
I would still argue that the logic is wrong. I don't know the spec but this would be the natural thing to do:

pseudo code:

```
if valuet_in_query? || value_in_variable?
  values_hash[arg_name] = value_from_query || value_from_variable
elsif default_value_present
  values_hash[arg_name] = default_value
end
```

In other words: The default value should only be used if there is no value in the query or variables.

So if I have a default value of 5, I can still override it to be null in my query.

josh:
Actually, I was wrong about this whole thing. `null` isn't literal value in GraphQL. You could receive null in `$variables` but you'd probably need to treat it like the undefined case.

rmosolgo:
@staugaard  I think that's the difference in our understanding: 

> override it to be null

GraphQL has no way to provide "null" as a value. That's the behavior described in the spec, and I'm not sold that we should deviate from it! 

I think this feature is still a good addition -- it makes `#to_h` more accurate. I've just added tests for that behavior, those tests fail on master like this: 

```
  test_0002_works from query literals                             FAIL (0.01s)
Minitest::Assertion:         --- expected
        +++ actual
        @@ -1 +1 @@
        -{"a"=>1, "b"=>2}
        +{"a"=>1, "b"=>2, "c"=>nil, "d"=>nil}
        /Users/rmosolgo/projects/graphql/spec/graphql/query/arguments_spec.rb:72:in `block (3 levels) in <top (required)>'
```

PR #253, opened 2016-09-15
Make a site: part two

gjtorikian:
A successor to https://github.com/rmosolgo/graphql-ruby/pull/205.

The main difference here is that the documentation generation relies solely on Jekyll, which would allow for much easier integration into GitHub Pages. Although I love Nanoc, I don't think it's necessary for a simple Markdown -> HTML site as you have here.

To run the site, you can type `rake serve`, which will build the pages and render everything to `localhost:4000`. I took care to place `baseurl` in instances where URLs need to resolve, which should ensure that the site works on `rmosolgo.github.io/graphql-ruby`, too.

I also added https://github.com/gjtorikian/html-proofer to catch for broken links (which it did!). There's one failing test indicating that `rmosolgo.github.io/graphql-ruby` doesn't work, which is correct. This ought to resolve once this PR is merged. 😉 

Let me know if there's anything else I can do here to help out!


rmosolgo:
Nice, Jekyll seems like a much better fit. :+1: for HTMLProofer. 

One specific requirement from my perspective is a good solution to have both: 

- A website whose docs match the latest RubyGems version 
- Documentation in the `master` branch that stays in sync with the source code 

Does this setup support that arrangement? Or can you think of a way to make it happen?

gjtorikian:
> Does this setup support that arrangement?

It does! There's [recent support](https://github.com/blog/2228-simpler-github-pages-publishing) to allow for better Pages publishing that doesn't involve the `gh-pages` branch. I just realized we'd have to change the folder here from `guides` to `docs`, but that's about it.

rmosolgo:
Let's say all the guides were in `docs/`. Soon, I'll merge a breaking change which will also require a change to the docs (eg #243). So, I update the file in docs in master and push to GitHub. I _thought_ that would trigger a build on GitHub pages, is that correct? (If so, then it's not quite what I'm hoping for, because the website would match _`master`_, not the latest RubyGems version. People would pull from RubyGems, try to set up according to the website ... and :boom: )

gjtorikian:
> is that correct?

That is correct. So, I guess your only option is to keep everything in _guides_ as-is. Write a Rake task that, as part of `bundle rake publish`, also takes those contents and moves them to `gh-pages`, where the site will build automatically. You can continue to update the contents of _guides_ and release a new website whenever the gem is updated.

I'm happy to go ahead and do that if it makes sense.

rmosolgo:
Yeah, I think a build step like that would do the trick! Would it be enough to "just" build like this: 

```
rm -rf docs 
mkdir -p docs 
cp -r guides/ docs/
```
 
So that the new guides just totally replace the old ones?

gjtorikian:
I'd rather not clutter the top-level root, though. Seems best to just build the site and check it straight into `gh-pages`, rather than keep a set of duplicate docs in _guides_ and _docs_.

rmosolgo:
That works for me! I get intimidated by the whole switching branches with Ruby thing, but if you can make it work ... :bow: !

gjtorikian:
Cool. You should be good to go!

`bundle exec rake site:preview` will show you the local state of the site. `bundle exec rake site:publish` will deploy the site. I stole these commands from https://github.com/jekyll/jekyll/blob/7a9427ccec81eb210bb818fc66523c7d57ddb890/rake/site.rake. 

I also tested the output of the site in my fork: https://github.com/gjtorikian/graphql-ruby/tree/589dbb93e0096711856ef8e2e9872db81a8b7a32

rmosolgo:
This looks awesome! I never knew that Jekyll provided a nice Ruby API like that. Let me just give it a try locally tonight before I merge it.

rmosolgo:
:metal: I pushed a couple of tweaks, which included touching `readme.md`, so I rebased, here goes!

rmosolgo:
This is so awesome!

rmosolgo:
it's alive!

http://rmosolgo.github.io/graphql-ruby/

gjtorikian:
:tada:

PR #254, opened 2016-09-16
Gracefully handle mutations when there is no mutation root

cjoudrey:
If a mutation is attempted on a schema that does not have a mutation root a `NoMethodError: undefined method unwrap for nil:NilClass` will occur originating from [lib/graphql/static_validation/type_stack.rb:108](https://github.com/rmosolgo/graphql-ruby/blob/dccbe895ca0cd194a0a8971850a7e13d37c37138/lib/graphql/static_validation/type_stack.rb#L108):

I noticed `graphql-js` handles this case by [returning a GraphQL error](https://github.com/graphql/graphql-js/blob/988508c9ee8e64e549b1028ffebdbbf4d22b4047/src/execution/execute.js#L268-L275).

@rmosolgo is a `StaticValidation` rule the right place for this?

### DIFF
--- a/lib/graphql/static_validation/rules/mutation_root_exists.rb
+++ b/lib/graphql/static_validation/rules/mutation_root_exists.rb
@@ -1,0 +1,20 @@
+module GraphQL
+  module StaticValidation
+    class MutationRootExists
+      include GraphQL::StaticValidation::Message::MessageHelper
+
+      def validate(context)
+        return if context.schema.mutation
+
+        visitor = context.visitor
+
+        visitor[GraphQL::Language::Nodes::OperationDefinition].enter << -> (ast_node, prev_ast_node) {
+          if ast_node.operation_type == 'mutation'
+            context.errors << message('Schema is not configured for mutations', ast_node, context: context)
+            return GraphQL::Language::Visitor::SKIP
+          end
+        }
+      end
+    end
+  end
+end
--- a/lib/graphql/static_validation/rules/subscription_root_exists.rb
+++ b/lib/graphql/static_validation/rules/subscription_root_exists.rb
@@ -1,0 +1,20 @@
+module GraphQL
+  module StaticValidation
+    class SubscriptionRootExists
+      include GraphQL::StaticValidation::Message::MessageHelper
+
+      def validate(context)
+        return if context.schema.subscription
+
+        visitor = context.visitor
+
+        visitor[GraphQL::Language::Nodes::OperationDefinition].enter << -> (ast_node, prev_ast_node) {
+          if ast_node.operation_type == 'subscription'
+            context.errors << message('Schema is not configured for subscriptions', ast_node, context: context)
+            return GraphQL::Language::Visitor::SKIP
+          end
+        }
+      end
+    end
+  end
+end
--- a/spec/graphql/static_validation/rules/mutation_root_exists_spec.rb
+++ b/spec/graphql/static_validation/rules/mutation_root_exists_spec.rb
@@ -1,0 +1,39 @@
+require "spec_helper"
+
+describe GraphQL::StaticValidation::MutationRootExists do
+  let(:query_string) {%|
+    mutation addBagel {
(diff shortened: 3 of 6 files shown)


rmosolgo:
:+1: I'd say so, how about checking for `subscription` too? It's a legal GraphQL root but I doubt that many schemas implement it!

cjoudrey:
Absolutely! I'll add that too.

spec/graphql/static_validation/rules/mutation_root_exists_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  let(:validator) { GraphQL::StaticValidation::Validator.new(schema: schema, rules: [GraphQL::StaticValidation::MutationRootExists]) }
+  let(:query) { GraphQL::Query.new(schema, query_string) }
+  let(:errors) { validator.validate(query)[:errors] }
+
+  it "errors when a mutation is performed on a schema without a mutation root" do
+    assert_equal(1, errors.length)
+    missing_mutation_root_error = {
+      "message"=>"Schema is not configured for mutations",
+      "locations"=>[{"line"=>2, "column"=>5}],
+      "path"=>["mutation addBagel"],
```
cjoudrey: Random question, is `path` relevant here? I was under the impression this should just be `nil`, but [the helper currently defaults to `context.path`](https://github.com/rmosolgo/graphql-ruby/blob/dccbe895ca0cd194a0a8971850a7e13d37c37138/lib/graphql/static_validation/message.rb#L10).

lib/graphql/static_validation/rules/subscription_root_exists.rb
```diff
@@ -0,0 +1,20 @@
+module GraphQL
+  module StaticValidation
+    class SubscriptionRootExists
+      include GraphQL::StaticValidation::Message::MessageHelper
+
+      def validate(context)
+        return if context.schema.mutation
```
rmosolgo: should this be `context.schema.subscription`?

spec/graphql/static_validation/rules/mutation_root_exists_spec.rb
rmosolgo: It generally leads to the field that has an error, like a more human-friendly version of `locations`. To me it makes sense here: some part of your query caused an error, if you want to find it, follow this path. Does it make sense in that context? If you think `nil` is a better fit, feel free to pass it.

lib/graphql/static_validation/rules/subscription_root_exists.rb
cjoudrey: :facepalm: -- yes. Bad copy/paste.

cjoudrey:
Makes sense! :) I'd keep it as is then.

cjoudrey:
@rmosolgo let me know if there's anything else you'd like me to change.

lib/graphql/static_validation/rules/mutation_root_exists.rb
```diff
@@ -0,0 +1,20 @@
...
+    class MutationRootExists
+      include GraphQL::StaticValidation::Message::MessageHelper
+
+      def validate(context)
+        return if context.schema.mutation
+
+        visitor = context.visitor
+
+        visitor[GraphQL::Language::Nodes::OperationDefinition].enter << -> (ast_node, prev_ast_node) {
+          if ast_node.operation_type == 'mutation' && !context.schema.mutation
```
rmosolgo: Is `!context.schema.mutation` already known at this point? (I think it would have caused an early return above.) Technically, someone could mess with the schema in the middle of validation but ... I don't think we need to support that case! (Same for `subscription` below, sorry for missing this earlier!)

rmosolgo:
Sorry for missing this earlier!

cjoudrey:
Good catch, should be fine now. :)

rmosolgo:
:metal:

rmosolgo:
🚢 in 0.18.12

PR #256, opened 2016-09-19
Update @skip and @include processing with specs

jsdnxx:
This PR corrects some of the behavior around @skip and @include directives.

# in summary, this PR
- fixes `@skip` and `@include` execution in graphql-ruby to be in line with spec and graphql-js behavior

## example of a valid graphql query that errors without this PR
```graphql
      query getCheese ($include: Boolean!, $skip: Boolean!) {
        cheese(id: 1) {
          flavor @include(if: false) @skip(if: true),
          withVariables: flavor @include(if: $include) @skip(if: $skip)
        }
      }
```
with variables
```
skip: true,
include: false
```

The expected result is that `flavor` would not be included in the result document, but it is. The spec updates in this PR include other cases.

# details

We encountered this error in production with queries that were being generated with Relay (0.9.2) for queries where the same field in some cases had `@include` directives on one fragment but not on another.

The GraphQL spec [defines the semantics](https://github.com/facebook/graphql/blob/master/spec/Section%203%20--%20Type%20System.md#include) when merging multiple `@skip` and `@include` directives.

After fixing directive validating in #263, I encountered some subtle bugs in the current implementation of `@skip` and `@include` in this gem where the observed behavior differs from the GraphQL spec. The spec mentions that skipped fields do not need to be present during execution, so my strategy was to skip over including these Field nodes in the internal representation when parsing the query.

I wrote specs for this behavior, included in this PR. To validate this behavior, I also implemented these same specs in javascript against Facebook's graphqljs. These other tests can be found at https://github.com/jsdnxx/graphql-js-test

Also, feel free to ping me on the graphql slack, username `@jsdn`

### DIFF
--- a/lib/graphql/internal_representation/rewrite.rb
+++ b/lib/graphql/internal_representation/rewrite.rb
@@ -24,7 +24,7 @@ module GraphQL
         # This tracks dependencies from fragment to Node where it was used
         # { frag_name => [dependent_node, dependent_node]}
         @fragment_spreads = Hash.new { |h, k| h[k] = []}
-        # [Nodes::Directive ... ] directive affecting the current scope
+        # [[Nodes::Directive ...]] directive affecting the current scope
         @parent_directives = []
       end
 
@@ -55,6 +55,7 @@ module GraphQL
               name: node_name,
               definition_name: ast_node.name,
               parent: parent_node,
+              included: false, # may be set to true on leaving the node
             )
           end
           object_type = context.parent_type_definition.unwrap
@@ -64,20 +65,21 @@ module GraphQL
         }
 
         visitor[Nodes::InlineFragment].enter << -> (ast_node, prev_ast_node) {
-          @parent_directives.push([])
+          @parent_directives.push(InlineFragmentDirectives.new)
         }
 
         visitor[Nodes::Directive].enter << -> (ast_node, prev_ast_node) {
           # It could be a query error where a directive is somewhere it shouldn't be
           if @parent_directives.any?
-            @parent_directives.last << Node.new(
+            directive_irep_node = Node.new(
               name: ast_node.name,
               definition_name: ast_node.name,
               ast_node: ast_node,
-              definitions: [context.directive_definition],
+              definitions: {context.directive_definition => context.directive_definition},
               # This isn't used, the directive may have many parents in the case of inline fragment
               parent: nil,
             )
+            @parent_directives.last.push(directive_irep_node)
           end
         }
 
@@ -88,6 +90,7 @@ module GraphQL
             parent: parent_node,
             name: ast_node.name,
             ast_node: ast_node,
+            included: false, # this may be set to true on leaving the node
           )
           # The parent node has a reference to the fragment
           parent_node.spreads.push(spread_node)
@@ -117,8 +120,9 @@ module GraphQL
           # so that they can be applied to fields when
           # the fragment is merged in later
           spread_node = @nodes.pop
-          spread_node.directives.merge(@parent_directives.flatten)
-          @parent_directives.pop
+          applicable_directives = pop_applicable_directives(@parent_directives)
+          spread_node.included ||= GraphQL::Execution::DirectiveChecks.include?(applicable_directives, context.query)
+          spread_node.directives.merge(applicable_directives)
         }
 
         visitor[Nodes::FragmentDefinition].leave << -> (ast_node, prev_ast_node) {
@@ -139,8 +143,9 @@ module GraphQL
           # and record any directives that were visited
           # during this field & before it (eg, inline fragments)
           field_node = @nodes.pop
-          field_node.directives.merge(@parent_directives.flatten)
-          @parent_directives.pop
+          applicable_directives = pop_applicable_directives(@parent_directives)
+          field_node.directives.merge(applicable_directives)
+          field_node.included ||= GraphQL::Execution::DirectiveChecks.include?(applicable_directives, context.query)
         }
 
         visitor[Nodes::Document].leave << -> (ast_node, prev_ast_node) {
@@ -149,12 +154,14 @@ module GraphQL
           while fragment_node = @independent_fragments.pop
             fragment_usages = @fragment_spreads[fragment_node.name]
             while dependent_node = fragment_usages.pop
-              # remove self from dependent_node.spreads
-              rejected_spread_nodes = dependent_node.spreads.select { |spr| spr.name == fragment_node.name }
-              rejected_spread_nodes.each { |r_node| dependent_node.spreads.delete(r_node) }
+              # Find the spreads for this reference
+              resolved_spread_nodes = dependent_node.spreads.select { |spr| spr.name == fragment_node.name }
+              spread_is_included = resolved_spread_nodes.any?(&:included?)
+              # Since we're going to resolve them, remove them from the dependcies
+              resolved_spread_nodes.each { |r_node| dependent_node.spreads.delete(r_node) }
 
               # resolve the dependency (merge into dependent node)
-              deep_merge(dependent_node, fragment_node, rejected_spread_nodes.first.directives)
+              deep_merge(dependent_node, fragment_node, spread_is_included)
               owner = dependent_node.owner
               if owner.ast_node.is_a?(Nodes::FragmentDefinition) && !any_fragment_spreads?(owner)
                 @independent_fragments.push(owner)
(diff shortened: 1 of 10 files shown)


jsdnxx:
Also, I found #124 and the changes here are consistent with the observation that `@skip` and `@include` are special in terms of how they are implemented and how they affect overall graphql resolution.

lib/graphql/internal_representation/rewrite.rb
```diff
@@ -45,6 +45,12 @@ def validate(context)
         visitor[Nodes::Field].enter << -> (ast_node, prev_ast_node) {
           parent_node = @nodes.last
           node_name = ast_node.alias || ast_node.name
+          
+          return :_skip if ast_node.directives.any? do |directive|
+            directive.name == 'skip' && directive.arguments.any? {|arg| arg.value == true } ||
+            directive.name == 'include' && directive.arguments.any? {|arg| arg.value == false }
+          end
```
rmosolgo: Just to make sure I understand correctly, this is here so that, when we _know_ that this field won't be included because of `@skip`/`@include`, we can avoid building an IRep tree, since we _know_ we won't be executing it.  Is that right? 

If this optimization is separate from the document validation issue, do you mind extracting it to a separate PR? In my opinion this kind of optimization deserves a serious look in its own right!

lib/graphql/internal_representation/rewrite.rb
```diff
@@ -138,6 +144,12 @@ def validate(context)
           # Pop this field's node
           # and record any directives that were visited
           # during this field & before it (eg, inline fragments)
+          
+          return if ast_node.directives.any? do |directive|
+            directive.name == 'skip' && directive.arguments.any? {|arg| arg.value == true } ||
+            directive.name == 'include' && directive.arguments.any? {|arg| arg.value == false }
+          end
```
rmosolgo: And this is a corresponding change for the one above, right? Since we shortcutted the `enter` hook, we should also shortcut the `leave` hook?

spec/graphql/static_validation/rules/fields_will_merge_spec.rb
```diff
@@ -45,4 +45,27 @@
...
+          name,
+          ...F0,
+          ...F1
+        }
+      }
+      fragment F0 on Cheese {
+        name @include(if: true)
+      }
+      fragment F1 on Cheese {
+        name @skip(if: false)
```
rmosolgo: How would this behavior be different if this directive was instead: `@skip(if: true)`, would it get a validation error? (Am I overlooking an existing spec for this case?)

rmosolgo:
I'm _really_ interested in this improvement! I love the idea of "compile-time" optimizations.

But I want to take it a bit slow -- I want to think through how this will affect a few things: 

- User experience for `@skip(if: true)` and `@skip(if: $boolVar)` is now very different. We "optimize" away some parts of the document in case of the literal `true`, but for the variable, we don't do so. So, your query could be rejected in the second case, even if not the first case. 
- People writing query analyzers should be aware that they no longer get the _whole_ query. We may "save them the trouble" in some cases. But are there cases where they _want_ the trouble? For example, if a deprecated field is behind `@skip(if: true)`, should we sent the user an warning email to update their query?

What do you think about the behavior changes in those cases?

lib/graphql/internal_representation/rewrite.rb
jsdnxx: It is separate. I included it in this PR because after the document validation change for `@skip` and `@include`, it exposes some cases where graphql-ruby will improperly merge these directives. That error was prevented before by the validation (unless executing with `validation: false`. Happy to split into a separate PR, because as you said, the particular implementation deserves careful thought. From our end, the main thing we're concerned with is correctness, with optimization and efficiency being secondary (but still important).

lib/graphql/internal_representation/rewrite.rb
jsdnxx: Yes - without shortcircuiting the leave hook, the `@nodes` stack gets out of sync with the current location in the AST, which breaks fragment resolution later on.

spec/graphql/static_validation/rules/fields_will_merge_spec.rb
jsdnxx: `@skip(if: true)` is also valid - any combination of `<no directive>`, `@skip(if: true)`, `@skip(if: false)`, `@include(if: true)`, and `@include(if: false)` should be considered valid, either in combination on the same AST field or split across fragments. This is because `@skip` and `@include` have specific semantics in the GraphQL spec, so we know they can merge cleanly - which is the intention of this `fields_will_merge` validation.

jsdnxx:
@rmosolgo thanks for considering this PR! I'm submitting it on behalf of my employer, Zendesk. We have a workaround in our use case, but we're interested in seeing this merged upstream. Correctness of behavior is our primary concern. So, happy to work with you on this issue to address all concerns.

> User experience for @skip(if: true) and @skip(if: $boolVar) is now very different. We "optimize" away some parts of the document in case of the literal true, but for the variable, we don't do so. 

Interesting! I had not considered this case, and did not specifically test for it. I'll update the PR spec to include cases with variables, in addition to using `true` and `false` literals.

> People writing query analyzers should be aware that they no longer get the whole query.

My primary goal is correct behavior, more than efficient parsing. I'm fine with including the nodes in the IRep. The datastructure then needs to support multiple instances of the same directive to be correctly merged later. `deep_merge_child` would need to be updated to merge the `child_node`'s directives with the `parent_node`'s. Also, I'm not sure that the current behavior of merging `extra_directives` is correct, but it was out of the scope of my PR so I didn't investigate it thoroughly.

With optimizations that may affect query analyzers, it might be worthwhile adding a flag for `advanced_optimizations?` or something like that, and then document what the effects are.

jsdnxx:
I've updated the directive spec cases in this PR with variables, and found a few new interesting failures. I've also made the same updates to my reference graphql-js spec, which passes: https://travis-ci.org/jsdnxx/graphql-js-test/builds/161394821 / https://github.com/jsdnxx/graphql-js-test/blob/eecfc039f8a686660d1c683dc77e72cf2fbeeb7c/spec.js

jsdnxx:
rebased after #263

jsdnxx:
I am confident in the specs for this PR, and would like to discuss alternate implementations.

rmosolgo:
Before getting into the weeds on implementation details, I thought I'd share how this works at the moment, in case that's helpful. 

- During execution, when we enter a selection (that is, a group of fields wrapped in `{ ... }`), we only execute its members if their directives allow it ([source](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/selection_resolution.rb#L16).
- [`DirectiveChecks.skip?(irep_node, query)`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/directive_checks.rb#L13-L24) returns `true` if some directive says that this node (and therefore its children) should be skipped. 
- Each directive has its own `include_proc` (eg [`@include`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/directive/include_directive.rb#L7-L9)), which uses its arguments to determine whether its owner node should be skipped. I consider this to be a design error: this degree of generalization is not required; I think this logic should be included in `DirectiveChecks.skip?` instead. 

If the main goal of this PR is to change how we handle incoming directives, I wonder, could it be handed in one of the execution-related objects? For example, could we: 

- Remove directive-specific `include_proc`s 
- Code that logic in `DirectiveChecks.skip?`, with the correct behavior. 

I suppose this will need some other update because previously, `Rewrite` assumed that fields with the same name would also have the _same_ directives, but now, they could have _different_ directives, so we need to union the directives from different fields. 

How would an approach like that suit you? 

It removes the "optimization" elements of this change, but the advantage of that is that we don't have to consider all the consequences of such an "optimization!"

jsdnxx:
I have updated this PR to only include the spec changes so we can see easily on Travis what behavior needs to be addressed.

I agree with consolidating the logic in `DirectiveChecks.skip?`. This is similar to what graphql-js does in their execution with a function called [`shouldIncludeNode`](https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L450-#L485)

Let me take a quick shot at it, I'll update this PR

jsdnxx:
Alright - everything looks good with the DirectivesCheck.skip? / .include?

I found that graphql-ruby appears to be handling multiple fields of the same name in the same selection incorrectly, compared to graphql-js, and I added tests showing that. Once again, this is something that could be split out into a separate PR

Test results: [graphql-ruby](https://travis-ci.org/rmosolgo/graphql-ruby/jobs/161675404) / [graphql-js](https://travis-ci.org/jsdnxx/graphql-js-test/builds/161706578)

rmosolgo:
Ok, I think I just caught up with you! 

Currently, this is handled properly, no field is present: 

```graphql 
{  
  # not evaluated because of @skip(if: true)
  skipped: field @skip(if: true) @include(if: true) 
  # not evaluated because of @include(if: false)
  notIncluded: field @include(if: false) @skip(if: false)
}
```

BUT we have a bug when it comes to merging fields. This _should_ evaluate a field, but it doesn't: 

```graphql 
{
  # this one is skipped 
  field @skip(if: true) 
  # but this one should be allowed 
  field 
}
```

Since we merge directives on fields with the same name, we can't tell that _one_ of the requests for this field has no directives! 

Did I understand that correctly? 

What if we re-work this by adding a new attribute to IRep nodes, such as `Node#included?`, which may start as `false`, but each field we visit may change the value? (I think we should keep `Node#directives` intact, but stop using it for this purpose.)

I think each AST node could yield one of these instructions: 

1. This field has no directives, so it MUST be present 
2. The directives that say it should be skipped, so it MIGHT NOT be present 
3. The directives that say it should be included, so it MUST be present 

So would it be enough to implement it this way?  

```ruby 
# When initializing the node, it gets an initial value from this occurrence in the AST 
node = Node.new(...) 
# This _might_ start as false:
node.included = DirectiveChecks.include_by_directives?(ast_node.directives) 
# ... 
# If we find a later field with this name, maybe switch it to be included 
node.included ||= DirectiveChecks.include_by_directives?(ast_node.directives)
```

That way, it may start out `false`, but any subsequent `true`s will override it. If it starts `true`, no subsequent `false` can override it. 

I think this requires a new function, `DirectiveChecks.include_by_directives?`, which takes the directives for that _specific_ AST node as input, instead of the IRep node, but I might be mistaken.

Then, during execution, we don't need `DirectiveChecks`, we can just ask `node.included?`. 

What do you think, would that do the trick??

jsdnxx:
> which takes the directives for that specific AST node as input, instead of the IRep node

yes, I think this is essentially correct. We're currently evaluating the DirectiveChecks on the IRep node, at which point it's too late (as you said) to see if there was an AST node which had *no* directives and therefore *must* be included

rmosolgo:
Wow, the logic for including fields across fragment spread & inline fragment boundaries turned out to be REALLY hard! 

As a bonus, I fixed a "compromise" in the way directives are treated by `Rewrite`. Previously, a directive on a parent node would "cascade" to all child selections, and all their children, and so on. This didn't matter because, previously, one `@skip` was enough to skip the whole thing, so if we messed with the child selections, it didn't matter! But to implement the right behavior here, I had to be more careful and _only_ take into account directives from immediate parent _spreads_ and _inline fragments_. 

That previous implementation caused bugs with `@stream` and `@defer` too (eg, `@stream` would cascade to nested lists, how unexpected!), so that should "just work" after a rebase :) 

It looks like the tests are turning green, so I'm going to merge this. Please let me know here or in another issue if I've missed something. 

Thanks for working with me on this, and for contributing the GREAT specification of behavior in the tests. I enjoyed working with them, knowing that when everything was ✅, we'd be set to go. I hope this will fix the issue with those Relay queries!!

jsdnxx:
Excellent! I had spent about an hour on this yesterday, but the direction I took started breaking some of other other tests around fragment spreads. Thank you so much for taking the time to help review the behavior, and for your patience with my several stabs at an implementation :)

PR #258, opened 2016-09-19
fix(FragmentsAreFinite) find fragment spreads below the first level of a fragment definition

### LINKED ISSUE #257: Circular references in fragments, but shouldn't!
From https://platform.github.community/t/circular-references-in-fragments/314:

> It seems that requests with circular references in fragments fail abruptly, instead of returning a validation error. See the "No fragment cycles4" validation rule.

True story! The query provided does refer onto itself, and shouldn't, because it results in a `stack level too deep`.

@rmosolgo Think you'll have time to look into this? If not I can pick it up, no worries. 👌

rmosolgo:
Oops, the previous implementation of validating infinite loops only checked the _first level_ of fragment definitions! Now, it checks deeply.

Closes #257

### DIFF
--- a/lib/graphql/static_validation/rules/fragments_are_finite.rb
+++ b/lib/graphql/static_validation/rules/fragments_are_finite.rb
@@ -14,12 +14,23 @@ module GraphQL
       private
 
       def has_nested_spread(fragment_def, parent_fragment_names, context)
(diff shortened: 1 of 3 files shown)


rmosolgo:
🚢 0.18.13

Issue #160, opened 2016-06-03
Provide nicer validation errors?


eapache:
@rmosolgo cc @dylanahsmith @lreeves

Right now if I run a mutation with invalid input (e.g. from a custom scalar definition):

```
mutation {
  doTheThing(input: { myField: "someInvalidValue" }) {
    myField
  }
}
```

I'll get back:

```
{
  "errors": [
    {
      "message": "Argument 'input' on Field 'doTheThing' has an invalid value",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    },
    {
      "message": "Argument 'myField' on InputObject 'DoTheThingInput' has an invalid value",
      "locations": [
        {
          "line": 2,
          "column": 21
        }
      ]
    }
  ]
}
```

There are two things that would be nice here:
- The first error complaining about `input` is, as far as I can tell, entirely redundant and occurs because the input for `myField` is invalid. It would be good if this was not returned since we already have a more precise error for `myField`.
- If the error for `myField` also returned `fields: [String!]` in the form `["doTheThing", "input", "myField"]` for callers to be able to easily (and programatically) determine which field was malformed.

I'm willing to work on this but a quick dive into `argument_literals_are_compatible.rb` didn't make either change look easy so I thought I'd ask for opinions and maybe hints first.

rmosolgo:
I agree with your suspicion that the first error is "cascading"! 

Agreed, it would be nice to include a human-readable path to the validation error instead of just the line & col. 

Feel free to take a crack at it if you're so inclined, but I think `StaticValidation` is ready for an overhaul. If you can survive with it for another month or so, I'll consider this a "feature request" at that time! Otherwise, if you add a spec for the desired behavior, I'll be sure not to break it when I get around to it.

rmosolgo:
Adding `path` to validation errors in #198

rmosolgo:
working on a rewrite now ...

gjtorikian:
> working on a rewrite now ...

Just jumping in, did this happen in a branch somewhere, @rmosolgo ?

rmosolgo:
Yes, #268 , but I gave up. I took some parts from there into `master`, including the part that made infinite loops 😆 

When I got to the part about making up new errors, I realized it was going to be impossible to preserve backwards compatibility. 

In retrospect, I think the right thing to've done would've been to target GraphqQL-CATS error messages.

rmosolgo:
[...]

PR #259, opened 2016-09-19
Add JSON-path to execution errors

josh:
Implements `path` on error messages as seen in graphql-js.

https://github.com/graphql/graphql-js/blob/89f3f1a319bb913616f2910e8a806170155ffaf0/src/error/GraphQLError.js#L42-L48
https://github.com/graphql/graphql-js/pull/396

-
To: @rmosolgo 
CC: @kdaigle


josh:
Hmm, I should probably an add array example. It should look something like `["user", "repositories", "edges", 42, "node"]`.

rmosolgo:
:+1: with a test for array indices 

I'm searching for a centralized approach while rewriting `StaticValidation`, I wonder if I can steal this!

lib/graphql/query/serial_execution/field_resolution.rb
```diff
@@ -35,6 +35,7 @@ def result
         def get_finished_value(raw_value)
           if raw_value.is_a?(GraphQL::ExecutionError)
             raw_value.ast_node = irep_node.ast_node
+            raw_value.path = irep_node.path
```
josh: @rmosolgo I'm a little stumped on how to get the current collection index context here. Just want to double check its not available somehow that I'm just missing. Otherwise I can try to hack some extra state into the execution context.

josh:
> 👍 with a test for array indices

Well, added a failing test 😁

lib/graphql/query/serial_execution/field_resolution.rb
rmosolgo: 😭  sadly it's totally _not_ available. 

Here's the `.map` where the index would come from: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/value_resolution.rb#L48

This feature is a hard requirement for `DeferredExecution` and you can see the implementation here: https://github.com/rmosolgo/graphql-ruby/blob/8e7700c1047722e753f662e8740087b6ec59f3ef/lib/graphql/execution/deferred_execution.rb#L386-L391

Although ... it doesn't really apply here :S I guess we have our pick among so-so options: 

- mutate the `irep_node` for each loop in the `map` 
- copy the `irep_node` and give the copy a new `path` for each loop in the map (`irep_node.dup` _is_ supported -- I think it's used for fragment inlining)
- add some kind of proxy object to wrap the irep node & send it into each loop of the map 
- ????

josh:
@rmosolgo So I think we should actually consider reverting https://github.com/rmosolgo/graphql-ruby/pull/198, or at least renaming that `"path"` key. 

> An array describing the JSON-path into the execution response which corresponds to this error. **Only included for errors during execution**.

https://github.com/graphql/graphql-js/blob/89f3f1a319bb913616f2910e8a806170155ffaf0/src/error/GraphQLError.js#L42-L48

https://github.com/rmosolgo/graphql-ruby/pull/198 defines a *path* format for validation errors. However, with the semantics defined in graphql-js, this should only be used as a path into the JSON result data. Which is there none in a validation error.

josh:
Don't love the collection index implementation, but it appears to *work*. Any suggestions for test cases that would probably break this?

rmosolgo:
Ugh, are either of these covered by the spec? All I can find is: 

> Every error must contain an entry with the key message with a string description of the error intended for the developer as a guide to understand and correct the error.
>
> If an error can be associated to a particular point in the requested GraphQL document, it should contain an entry with the key locations with a list of locations, where each location is a map with the keys line and column, both positive numbers starting from 1 which describe the beginning of an associated syntax element.
> 
> GraphQL servers may provide additional entries to error as they choose to produce more helpful or machine‐readable errors, however future versions of the spec may describe additional entries to errors.

It looks like _everybody_ is trying to be "more helpful", but in conflicting ways.

There _is_ a way that they can be disambiguated: if `data` is `null`, then the path describes a location in the document; if `data` is not `null`, the path describes a location in the response. 

Anyways, I don't _know_ that validation-error-`path` has any tooling built on it, so we can probably rename it safely. I added it because #160 ...

josh:
> Anyways, I don't know that validation-error-path has any tooling built on it, so we can probably rename it safely.

Proposed renaming in https://github.com/rmosolgo/graphql-ruby/pull/264

lib/graphql/query/serial_execution/value_resolution.rb
```diff
@@ -45,7 +45,8 @@ class ListResolution < BaseResolution
           def non_null_result
             wrapped_type = field_type.of_type
             strategy_class = get_strategy_for_kind(wrapped_type.kind)
-            value.map do |item|
+            value.each_with_index.map do |item, index|
+              irep_node.index = index
```
rmosolgo: Mutation is generally error-prone, but I'm open to it here, since it's a hot path and not used intensively. Maybe one good thing would be to put it back to `nil` when we're done, just in case? I can't think of why it would matter though. :S 

An alternative implementation would be to define a delegator, something like 

```
inner_node = IndexedNode.new(irep_node, index)
```

where `IndexedNode` is like 

```ruby 
class IndexedNode 
  extend Forwardable 

  def initialize(irep_node, index) 
    @irep_node = irep_node 
    @index = index 
  end 

  def path 
    @irep_node.path + [@index]
  end 

  def_delegators :@irep_node, :name, :definitions, :directives, # ?? Lots of stuff 
end 
```

I find this implementation less scary, since you never mess with the original `irep_node`, but it's also less efficient. I'm happy to leave it as is and refactor to something like :point_up: that if we _actually_ have issues.

lib/graphql/internal_representation/node.rb
```diff
@@ -72,6 +72,15 @@ def owner
         end
       end
 
+      def path
+        path = parent ? parent.path.dup : []
```
josh: Removed.

lib/graphql/internal_representation/node.rb
```diff
@@ -72,6 +72,15 @@ def owner
...
+        path << name if name
+        path << @index if @index
```
josh: Correct.

In the case of the root IRep, theres no `name` or `@index`.

lib/graphql/query/serial_execution/value_resolution.rb
josh: > Maybe one good thing would be to put it back to nil when we're done, just in case?

Reseting `index` back to `nil` in adf0c0febf44be3e501b32cee04566f428a7df5a

josh:
> Do you want to wait for #264 or can this be merged independently?

Cool. Appears to have no conflicting files either.

rmosolgo:
Nice, thanks for making this happen! Looking forward to see what it can be used for :)

PR #260, opened 2016-09-20
Don't move deprecated fields into `used_fields` segment

gjtorikian:
I feel like this makes more sense, no? 

/cc @cjoudrey as the initial adder.

### DIFF
--- a/lib/graphql/analysis/field_usage.rb
+++ b/lib/graphql/analysis/field_usage.rb
@@ -28,7 +28,7 @@ module GraphQL
         if irep_node.ast_node.is_a?(GraphQL::Language::Nodes::Field) && visit_type == :leave
           irep_node.definitions.each do |type_defn, field_defn|
             field = "#{type_defn.name}.#{field_defn.name}"
-            memo[:used_fields] << field
+            memo[:used_fields] << field unless field_defn.deprecation_reason
             memo[:used_deprecated_fields] << field if field_defn.deprecation_reason
           end
         end


lib/graphql/analysis/field_usage.rb
```diff
@@ -28,7 +28,7 @@ def call(memo, visit_type, irep_node)
         if irep_node.ast_node.is_a?(GraphQL::Language::Nodes::Field) && visit_type == :leave
           irep_node.definitions.each do |type_defn, field_defn|
             field = "#{type_defn.name}.#{field_defn.name}"
-            memo[:used_fields] << field
+            memo[:used_fields] << field unless field_defn.deprecation_reason
             memo[:used_deprecated_fields] << field if field_defn.deprecation_reason
```
swalkinshaw: This might be clearer:

```ruby
if field_defn.deprecation_reason
  memo[:used_deprecated_fields] << field
else
  memo[:used_fields] << field
end
```

cjoudrey:
Hm, interesting.

The way I was thinking about it is deprecated fields still belong in "used_fields" if they were used otherwise the name "used_fields" is sort of incorrect.

rmosolgo:
For my part, I agree with the original design. But I wonder if it's possible to "please everyone" with a generalized approach, eg 

```ruby 
schema.query_analyzers << FilteredFieldUsageAnalyzer.new(
  select: -> (type_defn, field_defn) { 
    # user code may return: 
    #   - `nil`, to do nothing 
    #   - something else, which will be pushed into an array 
  },
  analyze: -> (results) { 
    # `results` is an array with any items returned from `select` above
  } 
)
```

If only there was a way to provide _two_ blocks in Ruby :P

gjtorikian:
I'll just keep it as-is. I'm not mega passionate.

PR #261, opened 2016-09-20
fix(FieldsWillMerge) detect infinite spreads when checking mergeablility

rmosolgo:
Attempt number 2 at #257 

This is why `StaticValidation` needs a rewrite 😖


gjtorikian:
Putting here for posterity: this did fix the issue. 👍

rmosolgo:
:ship: in 0.18.14

Issue #255, opened 2016-09-19
Global analyzers error context?


gjtorikian:
Hello! Now I can use phrases like, "At GitHub, we've..." 😆 

At GitHub, we've got a few query analyzers. Two that report errors back to the user are our scope analyzer and our pagination analyzer. The scope analyzer ensures that, given a query, the requesting user has the right OAuth scopes and access (for example, if your query requests a user's email, you'll need [the `user:email` scope](https://developer.github.com/v3/oauth/#scopes). Our pagination analyzer reports on things like if you're not using `first:` or `last:` on connections, or if your `first` value is insanely large, etc.

We're interested in cascading errors through all our analyzers, such that the `errors` response a user receives reports on all the errors in a query. After your talk with @bswinnerton, we realized we could use `final_value` to report the sum of all the errors, but this is scoped per-query. We're trying to avoid a situation where we provide errors on scoping, the user fixes that, and then we provide errors on missing pagination arguments. Let's just throw 'em all at once!

Would you be open to a patch that handles this? I'm thinking of passing along an ivar through the analyzers, and then defining a new method, like `schema.query_analysis_result_handler`, which a user can then define to do whatever they want.

rmosolgo:
The goal is to support query analyzers _returning_ errors. Here's the code that's _supposed_ to handle one returned error per analyzer: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query.rb#L184-L185

(That is, `#final_value(memo)` may return an AnalysisError, which will be reported to the user & prevent the query.)

The intended use case is just what you described: different analyzers can return their own errors to the user.  However: 

- I don't think the multiple-analyzers case is _actually_ tested so ... maybe it's broken? 
- Maybe it would be an improvement to support _multiple_ errors per analyzer (for example, one error _per_ erroneous pagination)

[...]

PR #262, opened 2016-09-20
Intelligently queue together analyzer errors

gjtorikian:
Closes https://github.com/rmosolgo/graphql-ruby/issues/255

I pretty much went through and did what was discussed in #255!

/fyi @bswinnerton


spec/graphql/analysis/analyze_query_spec.rb
```diff
@@ -18,8 +18,12 @@ def call(memo, visit_type, irep_node)
   describe ".analyze_query" do
     let(:node_counter) {
       -> (memo, visit_type, irep_node) {
-        memo ||= Hash.new { |h,k| h[k] = 0 }
-        visit_type == :enter && memo[irep_node.ast_node.class] += 1
+        memo ||= Hash.new
```
gjtorikian: I'm not mega thrilled I had to switch to `Hash.new` without the block here. But, with the block, this test (and the test below) were pushing `memo[:errors] = 0`, causing the error checking logic to break.

I could be more careful in the error checker--something like `if next_memo.is_a?(Hash) && next_memo[:errors].present? && next_memo[:errors].all? { |e| e.is_a?(GraphQL::AnalysisError}`--but opted not for the verbosity. I can do that if you think it's safer, though!

gjtorikian:
No clue what's going on with RBX in Travis. Has it caused problems before?

guides/defining_your_schema.md
```diff
@@ -341,6 +341,40 @@ If the last value of `memo` (or the return of `.final_value`) is a `GraphQL::Ana
 - `GraphQL::Analysis::QueryDepth` and `GraphQL::Analysis::QueryComplexity` for inspecting query depth and complexity
 - `GraphQL::Analysis::MaxQueryDepth` and `GraphQL::Analysis::MaxQueryComplexity` are used internally to implement `max_depth:` and `max_complexity:` options
 
+### Handling errors
```
rmosolgo: :ok_hand: !!

lib/graphql/analysis/analyze_query.rb
```diff
@@ -21,7 +26,11 @@ def analyze_query(query, analyzers)
         reduce_node(op_node, analyzers_and_values)
       end
 
-      analyzers_and_values.map { |(r, value)| finalize_reducer(r, value) }
+      if !@errors.blank?
+        @errors.flatten
+      else
+        analyzers_and_values.map { |(r, value)| finalize_reducer(r, value) }
```
rmosolgo: In this flow, if _any_ reducer causes an error, then _none_ of the reducers are finalized. Is that the right behavior? 

For example, let's say you have two reducers: 

- one of them measures incoming query complexity and logs it;
- the other checks for valid permissions and turns away queries when people don't have permission

If someone makes a request with invalid permissions, should we still log the query complexity? 

(I started this comment by assuming the answer was yes, but now I'm not sure: since there was an error, we didn't _actually_ run the query, so maybe we shouldn't log it.)

lib/graphql/analysis_error.rb
```diff
@@ -1,4 +1,12 @@
 module GraphQL
   class AnalysisError < GraphQL::ExecutionError
+    def initialize(message, ast_node = nil)
```
rmosolgo: can we "future-proof" this by accepting a named argument instead, eg 

```ruby 
def initialize(message, ast_node: nil)
```

? Our future selves will thank us if this API grows!

spec/graphql/analysis/analyze_query_spec.rb
```diff
@@ -109,11 +115,103 @@ def call(memo, visit_type, irep_node)
...
+
+    before do
+      @previous_query_analyzers = DummySchema.query_analyzers.dup
+      DummySchema.query_analyzers.clear
+      DummySchema.query_analyzers << id_catcher << flavor_catcher
+    end
+
+    after do
+      DummySchema.query_analyzers.clear
+      DummySchema.query_analyzers.push(*@previous_query_analyzers)
```
rmosolgo: No need to change, but I thought I'd share, since writing the previous test, I learned that these are equivalent: 

```ruby
# Modify array_1 by adding the members of array_2 to the end of it
array_1.push(*array_2)
array_1.concat(array_2)
```

lib/graphql/analysis/analyze_query.rb
```diff
@@ -43,7 +52,18 @@ def visit_analyzers(visit_type, irep_node, analyzers_and_values)
...
+
+        begin
+          next_memo = reducer.call(memo, visit_type, irep_node)
+        rescue GraphQL::AnalysisError => e
+          @errors << e
+        end
+
+        if next_memo.is_a?(Hash) && next_memo[:errors].present?
+          @errors << next_memo[:errors]
+          next_memo[:errors] = []
```
rmosolgo: I wonder if it's possible to implement this feature without touching the reducer's "private" state. I think of this flow a bit like a safe deposit box: we manage `memo` on behalf of the reducer, but it is the reducer's _own_ state, it should be reducer-defined how that object is used.

spec/graphql/analysis/analyze_query_spec.rb
```diff
@@ -109,11 +115,103 @@ def call(memo, visit_type, irep_node)
...
+    end
+
+    it "groups all errors together" do
+      data = result["data"]
```
rmosolgo: Could we also assert that these both prevent the query from being executed, eg 

```ruby 
assert_equal nil, data
```

because I think that's also a very important behavior for both approaches to reporting errors!

lib/graphql/analysis/analyze_query.rb
```diff
@@ -21,7 +26,11 @@ def analyze_query(query, analyzers)
         reduce_node(op_node, analyzers_and_values)
       end
 
-      analyzers_and_values.map { |(r, value)| finalize_reducer(r, value) }
+      if !@errors.blank?
```
rmosolgo: `.blank?` and `.present?` are ActiveSupport methods, right? We want to support contexts _without_ ActiveSupport, so I think we should switch these  to `.none?` and `.any?`, respectively.

lib/graphql/analysis/analyze_query.rb
```diff
@@ -1,6 +1,9 @@
 module GraphQL
   module Analysis
     module_function
+
+    attr_reader :errors
```
rmosolgo: :tada: surprise, this is a module 😛 I think instance attributes on it will be shared across threads, resulting in :boom:. Currently, the local variable `analyzers_and_values` contains all state.

gjtorikian:
That didn't take too long to fix up at all. ✨

rmosolgo:
Now for a game we call "what the heck is wrong with Rubinius"

gjtorikian:
> Allow other reducers to finalize even if one of them finds an error (perhaps I'm wrong about this)

I meant to respond to this. I think it's more correct to not finalize on an error. (So, leaving it as is.)

As you noted, the query is not going to execute anyway. If users are graphing the max depths and such, they may find that these values do not end up correlating to queries that actually executed.

rmosolgo:
:tada: so awesome, thank you!!

rmosolgo:
:ship: in 0.18.14 !

PR #263, opened 2016-09-20
Update `fields_will_merge` validation to ignore directives

jsdnxx:
⚠️  blocked on #256 ⚠️ 

This PR skips "fields will merge" validation for @skip and @include directives, since they can be safely merged with behavior defined in the GraphQL spec.

## example of a valid graphql query that errors without this PR
```graphql
query {
  hello,
  ...F0
}
fragment F0 on RootQueryType {
   hello @include(if: true)
}
```

The issue is where directives cannot merge cleanly unless their semantics are known. However, in the case of `@skip` and `@include` in particular, their [semantics are specified](https://github.com/facebook/graphql/blob/master/spec/Section%203%20--%20Type%20System.md#include), so they should not generate these validation errors.

I wrote specs for this behavior, included in this PR. To validate this behavior, I also implemented these same specs in javascript against Facebook's graphqljs. These other tests can be found at https://github.com/jsdnxx/graphql-js-test

### note

With the updated validation logic in this PR, some queries that were previously rejected will be permitted which may still be resolved incorrectly by graphql-ruby due to the way `@skip` and `@include` directives are resolved. Specifically, some fields may be erroneously skipped from the result when one fragment has a field with `@skip(if: true)` and another fragment or query selection has the same field without any directives. See #256 for more details and discussion.

### DIFF
--- a/lib/graphql/static_validation/rules/fields_will_merge.rb
+++ b/lib/graphql/static_validation/rules/fields_will_merge.rb
@@ -76,16 +76,6 @@ module GraphQL
             errors << message("Field '#{name}' has an argument conflict: #{args.map {|a| JSON.dump(a) }.join(" or ")}?", defs.first, context: context)
           end
 
-          directive_names = defs.map { |defn| defn.directives.map(&:name) }.uniq
-          if directive_names.length != 1
-            errors << message("Field '#{name}' has a directive conflict: #{directive_names.map {|names| "[#{names.join(", ")}]"}.join(" or ")}?", defs.first, context: context)
-          end
-
-          directive_args = defs.map {|defn| defn.directives.map {|d| reduce_list(d.arguments) } }.uniq
-          if directive_args.length != 1
-            errors << message("Field '#{name}' has a directive argument conflict: #{directive_args.map {|args| JSON.dump(args)}.join(" or ")}?", defs.first, context: context)
-          end
-
           @errors = errors
         end
 
--- a/spec/graphql/static_validation/rules/fields_will_merge_spec.rb
+++ b/spec/graphql/static_validation/rules/fields_will_merge_spec.rb
@@ -36,8 +36,6 @@ describe GraphQL::StaticValidation::FieldsWillMerge do
 
   it "finds field naming conflicts" do
     expected_errors = [
-      "Field 'id' has a directive conflict: [] or [someFlag]?",                 # different directives
-      "Field 'id' has a directive argument conflict: [] or [{}]?",              # not sure this is a great way to handle it but here we are!
       "Field 'nickname' has a field conflict: name or fatContent?",             # alias conflict in query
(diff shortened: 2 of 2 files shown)


rmosolgo:
Thanks for breaking this out into a separate PR! 

I wonder if I was totally off my rocker on this validation. As I review the spec, I don't see any mention of asserting that directives are equal: 

image

It seems a bit spooky to me ... I mean, according to the spec, this is fair game, right? 

```graphql 
{
  someField 
  someField @defer 
}
```

Perhaps their assumption is: 

- `@skip` & `@include` are the only directives defined in the spec, and they can be safely handled in this way 
- If you make directives of your own, it's totally up to you to handle them safely.

I also found this example (in the context of ordered result sets) that demonstrates the same behavior you're advocating for in this PR: 

image


Are you interested in updating this PR to remove _all_ directive checks from `fields_will_merge.rb`? I think it's in keeping with the spec, and since `graphql-ruby` only supports `@include` and `@skip` right now, the effect on users will be the same. 

I'm also happy to merge as-is, what do you think?

jsdnxx:
The writing of the spec seems to indicate that *no* directives are used in validating whether fields can merge. This intuitively seems right to me, since the meaning of directives is defined by their implementation - the merge validation as specced is basically only type checking the contract of the field's shape in the result document.

Looking at the implementation of graphql-js, which is intended by Lee Byron & co. to be a de facto implementation-as-spec (and the reason I'm using spec cases written against graphql-js to get external validation on these PRs), it looks like their equivalent of the `fields_will_merge` validation rule is `OverlappingFieldsCanBeMerged`, which is commented as:
```
 * Conflicts occur when two fields exist in a query which will produce the same
 * response name, but represent differing values, thus creating a conflict.
```
While the behavior of specific directives is not known or specified at this point, they cannot change the type of a returned field to something outside of its schema definition. (I suppose it's possible for a directive to change the return type within a `union` type, for example).

graphql-js also has a validation rule called KnownDirectives. Their implementation is interesting, and lets each Directive definition specify what valid locations are for that directive: https://github.com/graphql/graphql-js/blob/master/src/validation/rules/KnownDirectives.js#L71-L79

Note that this directive validation checks individual directives, but doesn't do anything to check other directives on that field. No directive validation is done for different selection sets across fragments.

It seems like the correct thing to do would be to skip checking for directives in `fields_will_merge` and instead handle it during execution. It's still worth special-casing `@skip` and `@include`, but it appears that should be outside of the validation step.

jsdnxx:
I've updated this PR to remove directive checking from `fields_will_merge` validation rule

rmosolgo:
🎲  🎲  here goes! 

Thanks for this fix!

rmosolgo:
:ship: in 0.18.14 !

jsdnxx:
excellent- we're going to hold off upgrading until we can address the behavior in #256 Thanks for your quick replies :)

PR #264, opened 2016-09-20
Rename validation error `path` key

josh:
Conflicts with graphql-js **execution only** error JSON-path which uses the same key name.

https://github.com/graphql/graphql-js/blob/89f3f1a319bb913616f2910e8a806170155ffaf0/src/error/GraphQLError.js#L42-L48

Some other naming ideas:

* `queryPath`
* `documentPath`
* `validationPath`

Personally, this error path isn't that useful to me. `locations` tells me what I need to resolve the query AST node in the query text for reporting. But I don't see the harm in keeping it.

-
To: @rmosolgo

### DIFF
--- a/lib/graphql/static_validation/message.rb
+++ b/lib/graphql/static_validation/message.rb
@@ -26,7 +26,7 @@ module GraphQL
         {
           "message" => message,
           "locations" => locations,
-          "path" => path,
+          "fields" => path,
         }
       end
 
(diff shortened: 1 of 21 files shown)


brandonblack:
+1 for `queryPath`

rmosolgo:
I originally added this as part of addressing #160, but personally I don't depend on this name. That issue suggested `fields:` as the key. @eapache are you using this yet? Do you mind if we change it?

eapache:
Just did a quick check, we're OK if you change this it won't be too difficult to migrate.

josh:
> That issue suggested fields: as the key.

Updated with `"fields"`.

rmosolgo:
Sorry, I didn't mean to suggest a rename, only to document the history of the feature! 

Do you prefer the previous name? Otherwise I'll merge as-is.

josh:
> Do you prefer the previous name? Otherwise I'll merge as-is.

I'm indifferent, we'll call it whatever you'd like!

rmosolgo:
:+1: merged as `pathToValidationErrorForYouToFindThePlaceInTheQueryWhereItsMessedUp`

josh:
> 👍 merged as pathToValidationErrorForYouToFindThePlaceInTheQueryWhereItsMessedUp

😆

PR #267, opened 2016-09-20
Alternative: Make default_value behavior consistent

theorygeek:
This PR gets at the same problem as https://github.com/rmosolgo/graphql-ruby/pull/266, but it makes arguments behave the way input field's do (instead of vice-versa).

I think that this ends up being a more natural way to define default values, but some downsides:
- It ends up making validation more complicated. It relies on the `coerce_result` method to detect things like when you have a `String` field and set the default value to `[1, 2, 3]`. But that breaks when you have a list, because list doesn't have a `coerce_result` method.
- This is definitely a breaking change. Perhaps that could be monkey-patched to work correctly?

TODO:
- [x] Fix the broken tests (ie, by fixing `coerce_result`)

### DIFF
--- a/lib/graphql/enum_type.rb
+++ b/lib/graphql/enum_type.rb
@@ -115,8 +115,6 @@ module GraphQL
     end
 
     def coerce_result(value)
-      return nil if value.nil?
-
       ensure_defined
       @values_by_value.fetch(value).name
     end
--- a/lib/graphql/enum_type.rb
+++ b/lib/graphql/enum_type.rb
@@ -115,6 +115,8 @@ module GraphQL
     end
 
     def coerce_result(value)
+      return nil if value.nil?
+
       ensure_defined
       @values_by_value.fetch(value).name
     end
@@ -136,7 +138,7 @@ module GraphQL
         @name = name
         @description = description
         @deprecation_reason = deprecation_reason
-        @value = value
+        @value = value || name
       end
     end
   end
--- a/lib/graphql/enum_type.rb
+++ b/lib/graphql/enum_type.rb
@@ -138,7 +138,7 @@ module GraphQL
         @name = name
         @description = description
         @deprecation_reason = deprecation_reason
-        @value = value || name
+        @value = value
       end
     end
   end
--- a/lib/graphql/introspection/input_value_type.rb
+++ b/lib/graphql/introspection/input_value_type.rb
@@ -7,7 +7,7 @@ GraphQL::Introspection::InputValueType = GraphQL::ObjectType.define do
   field :defaultValue, types.String, "The value applied if no other value is provided" do
     resolve -> (obj, args, ctx) {
       value = obj.default_value
-      !value.nil? ? JSON.dump(obj.type.coerce_result(value)) : nil
+      value.nil? ? nil : JSON.dump(obj.type.coerce_result(value))
     }
   end
 end
--- a/lib/graphql/introspection/input_value_type.rb
+++ b/lib/graphql/introspection/input_value_type.rb
@@ -7,7 +7,7 @@ GraphQL::Introspection::InputValueType = GraphQL::ObjectType.define do
   field :defaultValue, types.String, "The value applied if no other value is provided" do
     resolve -> (obj, args, ctx) {
       value = obj.default_value
-      value ? JSON.dump(obj.type.coerce_result(value)) : nil
+      !value.nil? ? JSON.dump(obj.type.coerce_result(value)) : nil
     }
   end
 end
--- a/lib/graphql/introspection/input_value_type.rb
+++ b/lib/graphql/introspection/input_value_type.rb
@@ -7,8 +7,6 @@ GraphQL::Introspection::InputValueType = GraphQL::ObjectType.define do
   field :defaultValue, types.String, "The value applied if no other value is provided" do
     resolve -> (obj, args, ctx) {
       value = obj.default_value
-      # TODO: Only scalars respond to coerce_result
-      return nil unless obj.type.respond_to?(:coerce_result)
       value ? JSON.dump(obj.type.coerce_result(value)) : nil
     }
   end
--- a/lib/graphql/introspection/input_value_type.rb
+++ b/lib/graphql/introspection/input_value_type.rb
@@ -7,11 +7,9 @@ GraphQL::Introspection::InputValueType = GraphQL::ObjectType.define do
   field :defaultValue, types.String, "The value applied if no other value is provided" do
     resolve -> (obj, args, ctx) {
       value = obj.default_value
-      if value.is_a?(String)
-        "\"#{value}\""
-      else
-        value
-      end
+      # TODO: Only scalars respond to coerce_result
+      return nil unless obj.type.respond_to?(:coerce_result)
+      value ? JSON.dump(obj.type.coerce_result(value)) : nil
     }
   end
 end
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -102,7 +102,7 @@ module GraphQL
               name: type["name"],
               type: type_resolver.call(type["type"]),
               description: type["description"],
-              default_value: type["defaultValue"]
+              default_value: type["defaultValue"] ? JSON.parse(type["defaultValue"], quirks_mode: true) : nil
             )
           when "SCALAR"
             case type.fetch("name")
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -58,7 +58,7 @@ module GraphQL
                   name: enum["name"],
                   description: enum["description"],
                   deprecation_reason: enum["deprecationReason"],
-                  value: enum["value"]
+                  value: enum["name"]
                 )
               })
           when "INTERFACE"
--- a/lib/graphql/schema/validation.rb
+++ b/lib/graphql/schema/validation.rb
@@ -100,10 +100,16 @@ module GraphQL
 
         DEFAULT_VALUE_IS_VALID_FOR_TYPE = -> (type) {
           if !type.default_value.nil?
-            coerced_value = type.type.coerce_result(type.default_value)
+            coerced_value = begin
+              type.type.coerce_result(type.default_value)
+            rescue => ex
+              ex
+            end
 
-            if coerced_value.nil?
-              "default value #{type.default_value.inspect} is not valid for type #{type.type}"
+            if coerced_value.nil? || coerced_value.is_a?(Exception)
+              msg = "default value #{type.default_value.inspect} is not valid for type #{type.type}"
+              msg += " (#{coerced_value})" if coerced_value.is_a?(Exception)
+              msg
             end
           end
         }
--- a/lib/graphql/schema/validation.rb
+++ b/lib/graphql/schema/validation.rb
@@ -106,9 +106,9 @@ module GraphQL
               ex
             end
 
-            if coerced_value.nil? || coerced_value.is_a?(Exception)
+            if coerced_value.nil? || coerced_value.is_a?(StandardError)
               msg = "default value #{type.default_value.inspect} is not valid for type #{type.type}"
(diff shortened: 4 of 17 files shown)


lib/graphql/schema/validation.rb
```diff
@@ -100,7 +100,8 @@ def self.assert_named_items_are_valid(item_name, get_items_proc)
 
         DEFAULT_VALUE_IS_VALID_FOR_TYPE = -> (type) {
           if !type.default_value.nil?
-            coerced_value = type.type.coerce_input(type.default_value)
+            coerced_value = type.type.coerce_result(type.default_value)
```
rmosolgo: Previously we took a stringified default value and turned it into an application value. Now, don't we have an application value _already_? Or, why do we need to `coerce_result` on it?

rmosolgo:
Thanks for tracking this down! I see your point about taking application values as `default_value`. It's definitely a win for custom scalars, too. 

If we wanted to make a breaking change, we could do it, and suggest manually calling `coerce_input` as a migration path: 

```ruby 
argument :days, types[DayOfWeekEnum], default_value: types[DayOfWeekEnum].coerce_input(["MON", "TUES"])
```

I can't remember if `ListType` and `NonNullType` implement `coerce_input` + `coerce_result` -- if they don't, this is a great time to implement those!

lib/graphql/schema/validation.rb
theorygeek: There were a number of tests that would setup arguments like:
```ruby
argument :invalid do
  type GraphQL::STRING_TYPE
  default_value [1,2,3]
end
```
And then assert:
```ruby
it "applies validation to its member Arguments" do
  assert_error_includes invalid_argument_member_field, "default value [1, 2, 3] is not valid for type String"
end
```

This behavior is nice, because it's an easy way to catch developer mistakes. Also, it's still possible to preserve it, because input values are always eventually composed of scalars. Just need to add `coerce_result` behavior for lists and input object types.

theorygeek:
Glad you like this approach! I hadn't thought of suggesting `coerce_input` to deal with the breaking change.

I'll keep going down this path, and get the tests and such fixed-up.

theorygeek:
Summary of the big-ticket changes:
- Default values for arguments and input fields should now always be application values.
- The validation contract for default values is now changed.
  - **Before:** Default values were validated with the same strictness as variables.
  - **After:** Default values are only checked to ensure that `coerce_result` does not return `nil` or raise an exception.
- The `defaultValue` provided at introspection is now a JSON-encoded version of whatever you provided.
  - Similarly, when the schema is re-loaded, the default value is reconstructed by parsing the value as JSON.

Also, there was a bug in the schema loading code, where enum types were not reconstructed properly, fixed in [enum_type.rb](https://github.com/rmosolgo/graphql-ruby/pull/267/files#diff-b80b30c3eb4fbab060885af0f5950d76R141).

@rmosolgo Ready for review :)

theorygeek:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/265

swalkinshaw:
This definitely fixes #265 at least. It loads/parses a schema correctly which has default values of various types including Enums.

spec/graphql/argument_spec.rb
```diff
@@ -15,7 +15,7 @@
     }
 
     expected_error = %|Query is invalid: field "invalid" argument "invalid" default value ["123"] is not valid for type Float|
-    assert_equal expected_error, err.message
+    assert_includes err.message, expected_error
```
rmosolgo: Just curious, why can't we assert equality here? Was an error raised?

lib/graphql/introspection/input_value_type.rb
```diff
@@ -7,11 +7,7 @@
   field :defaultValue, types.String, "The value applied if no other value is provided" do
     resolve -> (obj, args, ctx) {
       value = obj.default_value
-      if value.is_a?(String)
-        "\"#{value}\""
-      else
-        value
-      end
+      !value.nil? ? JSON.dump(obj.type.coerce_result(value)) : nil
```
rmosolgo: This is a lot of calls on one line ... could we at least remove one by reversing the boolean logic? 

```ruby
value.nil? ? nil : JSON.dump(...)
```

?

lib/graphql/enum_type.rb
```diff
@@ -136,7 +138,7 @@ def initialize(name:, description:, deprecation_reason:, value:)
         @name = name
         @description = description
         @deprecation_reason = deprecation_reason
-        @value = value
+        @value = value || name
```
rmosolgo: Can you help me understand why this was needed? I couldn't quite tell just by looking! (Usually this is covered by `GraphQL::Define::AssignEnumValue`)

lib/graphql/schema/loader.rb
```diff
@@ -102,7 +102,7 @@ def define_type(type, type_resolver)
               name: type["name"],
               type: type_resolver.call(type["type"]),
               description: type["description"],
-              default_value: type["defaultValue"]
+              default_value: type["defaultValue"] ? JSON.parse(type["defaultValue"], quirks_mode: true) : nil
```
rmosolgo: Is this what fixes the Enum dump + load bug? If so, can we get a test for this change? For example, I think it could be added to `loader_spec.rb`, you could: 

- update the enum into have a value whose Ruby representation is something custom
- add an argument and use that value as a default value
- then improve `assert_deep_equal` to check for `Argument#default_value` equality

spec/graphql/argument_spec.rb
theorygeek: I modified the message to include the exception information, if there was one. In this case, the exception shows the undefined method `to_f` for array. I thought it would be helpful for debugging if you have a custom scalar, and the default value is being rejected.

This code could assert equality, in newer versions of Ruby, the messaging for undefined method exceptions changed:

Ruby 2.3.1:
```
2.3.0 :001 > ['abc'].to_f
NoMethodError: undefined method `to_f' for ["abc"]:Array
Did you mean?  to_s
               to_a
               to_h
	from (irb):1
	from /Users/theorygeek/.rvm/rubies/ruby-2.3.0/bin/irb:11:in `<main>'
```

Ruby 2.2.3:
```
2.2.3 :001 > ['abc'].to_f
NoMethodError: undefined method `to_f' for ["abc"]:Array
	from (irb):1
	from /Users/theorygeek/.rvm/rubies/ruby-2.2.3/bin/irb:11:in `<main>'
```

lib/graphql/enum_type.rb
theorygeek: So this is the fix for the enum dump + load bug, actually. Here's the bug:

In loader.rb, when an enum is being loaded, it receives a hash that looks like this:
```ruby
{"kind"=>"ENUM",
 "name"=>"Choice",
 "description"=>nil,
 "fields"=>nil,
 "inputFields"=>nil,
 "interfaces"=>nil,
 "enumValues"=>
  [{"name"=>"FOO", "description"=>nil, "isDeprecated"=>false, "deprecationReason"=>nil}, {"name"=>"BAR", "description"=>nil, "isDeprecated"=>false, "deprecationReason"=>nil}],
 "possibleTypes"=>nil}
```

It then instantiates `EnumType::EnumValue` and passes `value: enum['value']` to the constructor. Well, the thing is, application values for these enum's is not included in the introspection query. To my knowledge, even having application values is just an implementation detail that is not defined in the GraphQL Spec. 

The result is that the constructor gets `value: nil` passed in, and that causes all kinds of problems (eg, `@values_by_value` ends up looking like `{ nil => EnumValue<...> }`).

I modified it here, but now that I'm looking at it, this fix should probably be in the loader instead. But does that make sense?

lib/graphql/schema/loader.rb
theorygeek: Dove-tailing off my comment above about the enum loader fix, the application values for enum's is one of those things that gets lost in introspection (just like field resolvers).

At the time that the schema is being reloaded, the data being passed in from `defaultValue` is _always_ going to contain the GraphQL values (because we call `coerce_result` when it's dumped).

My thinking was:
- For enum's, we're OK, because by default, the GraphQL and application values are the same. (And per my comment above, we don't know the application values anyway)
- For the other built-in types, like Int and Float, we're also OK, because `JSON.parse` will give you the right native Ruby types.
- For custom scalars, you'll get whatever the external representation was (potentially any valid JSON value). We probably can't do much better than that, since we don't have the `coerce_input` proc's defined.

So back to your comment about testing, I think where that particular kind of test breaks down is:
- Define an enum with a custom value
- Export the schema via introspection; the custom value is lost
- Reload the schema, the best we can do is make the internal and external values the same
- Assert default value equality, and :boom:

I think the best we can test for is:
- Given different types of default values (enum's, input object types, lists, etc), it doesn't break.
- For custom scalars, the default value is equal to whatever you got from `coerce_result` on the original value.
- If you import a schema, you can export it again without any problems. (This test would probably break right now, since export relies on `coerce_result`, which is left undefined by the loader).

Hopefully that long train of thought made sense. I'll add some more tests for those cases I mentioned. Your thoughts?

spec/graphql/argument_spec.rb
rmosolgo: Ah, that makes sense, thank you for helping me understand!

lib/graphql/enum_type.rb
rmosolgo: Ohhh I see, yeah, I agree with your judgment to fix this in `loader.rb` -- we shouldn't look for a hash key that doesn't exist (and never will)!

lib/graphql/schema/loader.rb
rmosolgo: Oh huh, so is this a "how did it ever work" moment? Because previously the default value would _always_ have been a string, but with this commit, it's properly turned back into a float, int, boolean or, if it's a string, the extra quotes are removed? 

👏 👏 ! I think a test case for that behavior would be great, that way we don't um, re-break it in the future!

swalkinshaw: I'm pretty sure schema loading only ever worked if you either didn't have `defaultValue`'s or they were actually a string. So it didn't really work :) mine blew up right away due to our usage of ints and enum values as default values.

rmosolgo: 😬  Glad to see this fixed! Is this change covered by a change in the tests? (As I review the diff, it's not jumping out at me.)

theorygeek:
I added some more test cases, and I _think_ they cover all of the new behavior. Take a look and see if you think there's something missing.

One thing mentioned specifically was a test for the `JSON.parse` behavior when loading a schema. There was an existing test that broadly covered loading a schema; I added a few others that focus on the default values specifically.

Also, I mentioned a problem that would occur if you loaded a schema with custom scalars that had default values (ie, they wouldn't have `@coerce_result_proc`'s, so validation and exporting would fail). I fixed this by providing [default `coerce_result` behavior](https://github.com/rmosolgo/graphql-ruby/pull/267/files#diff-459636301230b8c63aa8c35b716e4eccR67) for custom scalars - they just return their provided value.

>Is there any documentation that needs updating? (For example, in the guides, or a the top of argument.rb or input_object_type.rb?)

The comments at the top of `argument.rb` and `input_object_type.rb` are not incorrect after this change. However, they don't explain that default values should be _application_ values. (I'm not sure if that's something to add or not.)

As for the guides, they don't really mention default values much. I did find a few code examples that looked like this:
```ruby
argument :moderated, types.Boolean, default_value: true
```

That one lone line doesn't provide very much guidance 😛 . 

It'd probably be good to augment the guides so that they explain:
- custom scalars
- the concept of an _application_ value, in contrast to a _result_ value (is this the right jargon?) 
- how default values are handled (most notably, that they're passed to resolvers without any modifications, such as array-wrapping)

/cc: @rmosolgo Ready for review again.

rmosolgo:
:clap: :clap: thanks for tracking down all the nitty-gritty details on this! It'll be a huge win to make this behavior consistent (and ya know, _actually_ work!). I'll take a stab at updating the docs before I cut a new version, since this is technically a breaking change.

PR #268, opened 2016-09-21
Typecheck rewrite

rmosolgo:
(This isn't done yet -- just a WIP branch in case you're interested in the progress here.)

`GraphQL::StaticValidation` is one of the oldest, dustiest parts of `graphql-ruby`. It makes sense: it was important to do right away, and it's _basically_ worked, so I haven't messed with it. But, I think revisiting this problem could yield dividends. Here are my goals in rewriting that functionality: 
- Support validation _without_ a schema: some validations can be run a query string alone. If we decouple the code properly, we get that behavior at no extra cost! 
- Improve performance: I have doubts about some of the rules, see below. 
- Reduce ad-hoc traversals: some existing rules re-traverse the AST at the end of validation. Besides being a huge waste of time, this is buggy, since each rule has to implement its own prevention of infinite loops (see #257)
- Find _more_ errors: there are many cases where the current implementation finds _one_ error then stops the traversal. Technically, we can do better: we can continue traversing, looking for type-agnostic usage errors, and possibly _regaining_ type information (eg, inline fragment) so we can continue to type-check. 
- Improve maintainability: currently, the rules are order-dependent: some of them require that previous rules halt traversal in case of an error. So, if you reorder them, :boom:. Also, I want to run all tests through the _whole_ validation stack instead of one-validator-at-a-time. This is because there have been lots of bugs about how validators work together (or _don't_ work together)

My approach has a few components: 
- Only one tree traversal. Gather important info during the _one_ traversal, then distribute it to individual rules for their own purposes. 
- Recover from type errors. Use `AnyType`, `AnyField` and friends to inject "magical" types in that case. This requires adding some methods to the "real" types and fields so that they can be easily imitated. 
- Lump more behavior together. It might result in bigger files, but I want to avoid cases where code dependencies are not clear because they've been (wrongfully) extracted to other files. Also, I had a previous goal of matching Ruby files one-to-one with paragraphs in the spec. I don't have that goal anymore -- now my goal is to implement the spec as best I can, in Ruby. 
- Explicit separation of schema-dependent rules and schema-independent rules. It doesn't have to be a _great_ API, but there should be an API for validating a document _only_. (As long as the behavior is there, we can improve the API later.)

Generally, the steps are: 
- Reimplement (and retest 😩 ) document validation 
- Run some benchmarks 
- Make it faster, if it isn't already faster than the current implementation 
- Document the new APIs 
- :ship: 

---

**TODO:**
- [x] Reimplement validations 
- [ ] Refactor: type_check.rb should only hold calls on shared state -- everything else should be extracted 
- [ ] Refactor: rename typecheck modules & functions 
- [ ] Test: run existing tests through new code, make sure there are no regressions 
- [ ] Refactor: Migrate new code to StaticValidation


rmosolgo:
Just finished reimplementing the last rule, naturally, I couldn't resist a preliminary benchmark. TL;DR: 2x faster :tada: !

```ruby 
# validation_benchmark.rb
require "graphql"
require "benchmark/ips"
require "support/dairy_data"
require "support/dairy_app"
document = GraphQL.parse(GraphQL::Introspection::INTROSPECTION_QUERY)
schema = DummySchema
query = GraphQL::Query.new(schema, document: document)

Benchmark.ips do |x|
  x.report("old validation") {
    validator = GraphQL::StaticValidation::Validator.new(schema: schema)
    validator.validate(query)
  }

  x.report("new validation") {
    visitor = GraphQL::Language::Visitor.new(document)
    analysis = GraphQL::StaticAnalysis.prepare(visitor, schema: schema)
    visitor.visit
    analysis.errors
  }
  x.compare!
end
```

```sh
~/code/graphql $ ruby -Ilib -Ispec validation_benchmark.rb
      old validation    43.000  i/100ms
      new validation   104.000  i/100ms
      old validation    430.374  (± 4.0%) i/s -      2.193k in   5.104203s
      new validation      1.059k (± 3.5%) i/s -      5.304k in   5.012636s

Comparison:
      new validation:     1059.5 i/s
      old validation:      430.4 i/s - 2.46x  slower
```

rmosolgo:
I guess it's not quite fair: I haven't re-implemented `queryPath` yet!

rmosolgo:
Some of the lessons-learned from this branch have been merged into master. The performance better and I've removed the ad-hoc recursion from the rules. There are still some goals unaccomplished here (eg, improved handling of undefined types and fields) but the main issues are resolved!

PR #272, opened 2016-09-22
Fix for a small syntax error in the security guide

Willianvdv:


rmosolgo:
Thank you! Bad migration to the new API on my part.

PR #273, opened 2016-09-23
Add schema language directives

cjoudrey:
Back in May, https://github.com/graphql/graphql-js/pull/376 added support for directives in GraphQL IDL.

Deprecations were added as directives to the IDL in https://github.com/graphql/graphql-js/pull/384.

Finally, descriptions were added to the IDL in https://github.com/graphql/graphql-js/pull/464.

This is part 1 of a few pull requests I will be submitting that will add all of these features to the gem.

This pull request includes:

- Add parsing support for directives.
- Test schema parsing/generation against a copy of graphql-js' [`schema-kitchen-sink.graphql`](https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/language/__tests__/schema-kitchen-sink.graphql) to ensure we are consistent.
- Add a new `EnumValueDefinition` AST node to support directives on enum values. **This is a breaking API change** as `GraphQL::Language::Nodes::EnumTypeDefinition#values` used to return an array of strings, now it will return an array of `EnumValueDefinition`.

:eyes: @rmosolgo @dylanahsmith


lib/graphql/language/nodes.rb
```diff
@@ -202,7 +202,7 @@ class FragmentDefinition < AbstractNode
         def initialize_node(name: nil, type: nil, directives: [], selections: [])
           @name = name
           @type = type
-          @directives = directives
+          @directives = directives || []
```
rmosolgo: I see that directives is often falling back to `|| []` -- what's the case where this is needed? (I ask because I'm only aware of `directives_list_opt` in the parser, which already provides `[]` for the empty case.)

lib/graphql/language/generation.rb
```diff
@@ -116,7 +130,7 @@ def generate(node, indent: "")
         when Array
           "[#{node.map { |v| generate(v) }.join(", ")}]"
         when Hash
-          "{ #{node.map { |k, v| "#{k}: #{generate(v)}" }.join(", ")} }"
+          "{#{node.map { |k, v| "#{k}: #{generate(v)}" }.join(", ")}}"
```
rmosolgo: Am I right that this whitespace change was added so that our dump matches the "kitchen sink" example from graphql.js ? It's good by me, I just want to make sure I understand it before adding noise to peoples' diffs :P

cjoudrey: That's correct. I took for granted that Facebook's "kitchen sink" was formatted correctly. Also: https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/language/printer.js#L81

lib/graphql/language/nodes.rb
cjoudrey: Good question. I did this because I saw we had done it for [`ObjectTypeDefinition` `interfaces`](https://github.com/rmosolgo/graphql-ruby/blob/1a7f560212c419f2ab2aa6020528582cc744f5a2/lib/graphql/language/nodes.rb#L346).

I was under the impression we did this so that if someone intentionally passes `nil`, it would get turned into `[]`. It does add quite a bit noise, maybe we can just remove the `|| []`.

rmosolgo: :+1: I'd say if the tests are green without this, then let's not add them. If "people" pass `nil`, they get nil :P (fortunately I think "people" is just `parser.rb`)

cjoudrey: Done 😁

rmosolgo:
Awesome, thanks for this new feature!

PR #275, opened 2016-09-23
Add deprecations to IDL

cjoudrey:
This is part 2 of #273 and is currently based off it. Let's wait for #273 to merge before reviewing this.

This PR adds support for deprecations in the IDL using directives as per graphql/graphql-js#384.

Part 3 will be adding descriptions to IDL as per https://github.com/graphql/graphql-js/pull/464.


lib/graphql/directive.rb
```diff
@@ -16,16 +16,19 @@ class Directive
...
       INLINE_FRAGMENT =     :INLINE_FRAGMENT,
+      ENUM_VALUE =          :ENUM_VALUE,
     ]
 
     def initialize
       @arguments = {}
     end
 
     def include?(arguments)
+      return true if include_proc.nil?
```
rmosolgo: The bad news is, I caused a conflict. 

The good news is, this is implemented directly in `DirectiveChecks` and this method is gone :tada:

cjoudrey:
Haha don't worry! Saw it coming. :D I'll fix tonight.

cjoudrey:
@rmosolgo conflict solved. 🎉

cjoudrey:
Any other comments for this one? :)

lib/graphql/schema/printer.rb
```diff
@@ -44,6 +48,10 @@ def print_schema_definition(schema)
       BUILTIN_SCALARS = Set.new(["String", "Boolean", "Int", "Float", "ID"])
       private_constant :BUILTIN_SCALARS
 
+      def is_spec_directive(directive)
+        ['skip', 'include', 'deprecated'].include?(directive.name)
+      end
```
cjoudrey: Good question.

I ported this from the JS implementation: https://github.com/graphql/graphql-js/blob/5e3d377e6e0881191efe02b9d40720ff0f0b5b0f/src/utilities/schemaPrinter.js#L28-L44

My understanding of the JS code is that when printing the schema, we do not want to include spec directives since they are part of the spec. We still print non-spec directives though.

When printing the introspection query, we only include spec directives.

lib/graphql/schema/printer.rb
```diff
@@ -56,12 +64,24 @@ def print_type(type)
...
-            type.all_fields.map{ |field| "  #{field.name}#{print_args(field)}: #{field.type}" }.join("\n")
+        module DeprecatedPrinter
+          def print_deprecated(field_or_enum_value)
+            return unless field_or_enum_value.deprecation_reason
+
+            if field_or_enum_value.deprecation_reason == ''
+              ' @deprecated'
+            else
+              " @deprecated(reason: #{field_or_enum_value.deprecation_reason.to_s.inspect})"
+            end
```
cjoudrey: Good catch. The code is clearer with the `case` statement. I'll make that change.

lib/graphql/schema/printer.rb
rmosolgo: > non-spec directives 

Part of me wants to leave this out, since there _aren't_ any non-spec directives yet. But ... if we left it out, I can guarantee people would be surprised not to see `@defer` / `@stream` in the query dump if I ever got around to merging that 😬

spec/graphql/schema/printer_spec.rb
```diff
@@ -14,6 +14,7 @@
       value "FOO"
       value "BAR"
       value "BAZ", deprecation_reason: 'Use "BAR".'
+      value "WOZ", deprecation_reason: GraphQL::Directive::DEFAULT_DEPRECATION_REASON
```
rmosolgo: image

rmosolgo:
:metal: Very cool, thanks!

lib/graphql/schema/printer.rb
cjoudrey: > would be surprised not to see `@defer` / `@stream` in the query dump 

Just to make sure we're on the same page. You'd still see `@defer` and `@stream` on the fields, you just don't see their definitions, i.e. `directive @defer`.

I think the other reason they do this in the JS implementation is because `directives` can be specified on the `schema`: https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/type/schema.js#L50-L54

We don't support that yet.

rmosolgo: You're right, maybe I don't understand! If graphql-ruby supports `@stream` and `@defer`, but they aren't part of the GraphQL spec, would they show up in a schema print-out or not?

cjoudrey: Don't worry, I kept confusing myself here too.

There's two concepts: `Directive` and `DirectiveDefinition`.

Example `@skip` `Directive`:

```
query myQuery($someTest: Boolean) {
  experimentalField @skip(if: $someTest)
}
```

Example `@skip` `DirectiveDefinition`:

```
directive @skip(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
```

The `is_spec_directive` method is used to omit the `DirectiveDefinition` of spec directives from getting printed when dumping the schema.

I need to rename `def print_directive(directive)` to `def print_directive_definition(directive)` as that's not named correctly and could be a source of confusion.

In terms of `Directive` we're only printing the `deprecated` directive at the moment, see `DeprecatedPrinter`. We may want to generalize this in the future though.

PR #276, opened 2016-09-25
Remove definition name from error `path`

josh:
Was correct for anonymous query definitions, but I forgot to test what would happen if a query name was set. It incorrectly gets prepended to the `"path"` array.

-
To: @rmosolgo

### DIFF
--- a/lib/graphql/internal_representation/node.rb
+++ b/lib/graphql/internal_representation/node.rb
@@ -82,10 +82,14 @@ module GraphQL
       end
 
       def path
-        path = parent ? parent.path : []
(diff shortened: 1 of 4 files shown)


rmosolgo:
:+1: makes sense, since you want the path to the value in `{ "data" => ... }` instead of the query string, it looks like a couple tests need updating (but good that we have some with a query name, and some without)

josh:
> it looks like a couple tests need updating

D'oh, forgot to run the full test suite locally. Fixed!

rmosolgo:
👌  cool!

PR #277, opened 2016-09-25
Fix LEFT and RIGHT in lexer/parser

cjoudrey:
This made me chuckle. 😀 

The parser rules will now make a bit more sense too as the left brace/paren/bracket will actually be on the left side of definitions.

For reference: https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/language/lexer.js#L89-L99

:eyes: @rmosolgo


rmosolgo:
😅 I've always been bad at left and right ... if you ever ask me for directions, I have to point with my hands while I talk, to make sure what I'm saying matches what's in my mind's eye.

Thanks for fixing this "typo"!

cjoudrey:
:grinning: :tada:

PR #278, opened 2016-09-25
Fix: remove pluralization of argument in docs for object_type

RyanCCollins:
Note, I believe that the guide was misleading for the object_type.  Using the argument `arguments` (plural) seems to have thrown an exception.  I see that in other examples, it is `argument`, so I believe that this may have been an error.


rmosolgo:
Sorry, thanks for fixing that!!

RyanCCollins:
No problem at all! Love the library! Thanks so much for your work.

PR #281, opened 2016-09-27
Adds conversion of a passed block to an argument if needed

gucki:
With this patch it's now possible to call all methods ruby style with blocks:

Before (still works of course):

```
connection :users, UserType.connection_type do
    resolve ->(object, args, ctx){
      User.all
    }
end
```

Now also possible:

```
connection :users, UserType.connection_type do
  resolve do |object, args, ctx|
    User.all
  end
end
```

### DIFF
--- a/lib/graphql/define/defined_object_proxy.rb
+++ b/lib/graphql/define/defined_object_proxy.rb
@@ -13,6 +13,7 @@ module GraphQL
       def method_missing(name, *args, &block)
         definition = @dictionary[name]
         if definition
+          args = args + [block] if args.count == definition.method(:call).arity - 2 && block
           definition.call(@target, *args, &block)
         else
           p "Failed to find config #{name} in #{inspect}"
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: I love this API change! Maybe it will make people feel more at home using the DSL, even though we're still using proper functions under the hood :P 

I restarted the tests, hopefully they'll just magically pass the second time (it looks like there was a problem downloading `rbx`)

gucki:
Can you again restart the tests (error was again rbx install failure)? Seems like I cannot, can I?

rmosolgo:
I can't figure out how to fix `rbx`, so I've added `allow_failures` on master 😢  

A couple thoughts: 

- How should this API change be reflected in the documentation? 
- Since the goal of this PR is to support _both_ block and positional arguments, would you mind adding an example to the tests which use the positional argument? That way, when working with this code later, the side-by-side tests will remind us how it supports both uses.

rmosolgo:
In the end, I'm happy to accept an object that responds to `#call`. If you want to push ahead with block syntax, feel free to reopen this with test & docs, and we can take another look!

PR #284, opened 2016-09-28
Parity with graphql-js' introspection schema

cjoudrey:
While working [on adding descriptions to the schema dumper](https://github.com/graphql/graphql-js/pull/464), I noticed our introspection schema descriptions don't match graphql-js'.

This means we need to maintain [our own copy of the introspection schema](https://github.com/rmosolgo/graphql-ruby/blob/9c6ebbda48b59c0a23467ab9dd45179654279282/spec/graphql/schema/printer_spec.rb#L71).

This PR brings us back to parity with graphql-js by [using their introspection schema](https://github.com/graphql/graphql-js/blob/a725499b155285c2e33647a93393c82689b20b0f/src/utilities/__tests__/schemaPrinter-test.js#L600-L797) in our tests.

As graphql-js evolves this means we'll be able to easily copy over their introspection schema to ensure the gem remains compliant.

👀 @rmosolgo

### DIFF
--- a/lib/graphql/directive/deprecated_directive.rb
+++ b/lib/graphql/directive/deprecated_directive.rb
@@ -7,5 +7,5 @@ GraphQL::Directive::DeprecatedDirective = GraphQL::Directive.define do
     "suggestion for how to access supported similar data. Formatted "\
     "in [Markdown](https://daringfireball.net/projects/markdown/)."
 
-  argument :reason, !GraphQL::STRING_TYPE, reason_description, default_value: GraphQL::Directive::DEFAULT_DEPRECATION_REASON
+  argument :reason, GraphQL::STRING_TYPE, reason_description, default_value: GraphQL::Directive::DEFAULT_DEPRECATION_REASON
 end
--- a/lib/graphql/introspection/type_kind_enum.rb
+++ b/lib/graphql/introspection/type_kind_enum.rb
@@ -1,7 +1,7 @@
 GraphQL::Introspection::TypeKindEnum = GraphQL::EnumType.define do
   name "__TypeKind"
-  description "The kinds of types in this GraphQL system"
+  description "An enum describing what kind of type a given `__Type` is."
   GraphQL::TypeKinds::KIND_NAMES.each do |kind_name|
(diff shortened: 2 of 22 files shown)


lib/graphql/directive/deprecated_directive.rb
```diff
@@ -7,5 +7,5 @@
     "suggestion for how to access supported similar data. Formatted "\
     "in [Markdown](https://daringfireball.net/projects/markdown/)."
 
-  argument :reason, !GraphQL::STRING_TYPE, reason_description, default_value: GraphQL::Directive::DEFAULT_DEPRECATION_REASON
+  argument :reason, GraphQL::STRING_TYPE, reason_description, default_value: GraphQL::Directive::DEFAULT_DEPRECATION_REASON
```
cjoudrey: This PR made me realize I made a mistake in https://github.com/rmosolgo/graphql-ruby/pull/275. `reason` is not required. 😁

lib/graphql/introspection/type_kind_enum.rb
```diff
@@ -1,7 +1,7 @@
 GraphQL::Introspection::TypeKindEnum = GraphQL::EnumType.define do
   name "__TypeKind"
-  description "The kinds of types in this GraphQL system"
+  description "An enum describing what kind of type a given `__Type` is."
   GraphQL::TypeKinds::KIND_NAMES.each do |kind_name|
-    value(kind_name)
+    value(kind_name, GraphQL::TypeKinds::TYPE_KIND_DESCRIPTIONS[kind_name.to_sym])
```
rmosolgo: It looks like `KIND_NAMES` is used here and in `TypeKind` for dynamically generating some methods. 

I think it would be a nice refactor to: 

- Add `TypeKind#description` (bringing it inline with other schema members) and initialize each one with its description
- Remove `KIND_NAMES` and iterate over `TYPE_KINDS` here & in `TypeKind` instead 

Or, is there a benefit to keeping the descriptions in a hash?

rmosolgo:
🚀 !

PR #287, opened 2016-09-30
Remove errant commas and equals in documentation.

mcolyer:
A simple fix but helpful when you're getting started.


rmosolgo:
Thank you! Looks like some failed updates after API changes 😬 !

PR #289, opened 2016-10-02
Eager load schema

rmosolgo:
Previously, schema validation was deferred until the first call to `Schema#types`. This call constructed the `TypeMap` and validated each entry. This means you might start up your app and get an error backtrace that points to validation, which is annoying. 

Also, it might have been the culprit in some thread-safety bugs. If you start accepting requests on different threads, one thread might start building the schema's type map, then the other thread might see that the type map was already there, and try to use it, but it's not complete, so :boom: .

Now, run validations as part of `Schema.define`. This will find errors earlier and reduce the likelihood of thread-unsafety with the `TypeMap`.

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -66,6 +66,7 @@ require "graphql/schema/printer"
 
 require "graphql/analysis_error"
 require "graphql/invalid_null_error"
+require "graphql/unresolved_type_error"
 require "graphql/query"
 require "graphql/internal_representation"
 require "graphql/static_validation"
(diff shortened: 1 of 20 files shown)


rmosolgo:
@cjoudrey A while back we were talking about, "if you had field-level instrumentation functions, _when_ would you attach them?" I think inside `#define`, after `super`, the same place I added this validation, that would be the place (it didn't exist before this).

cjoudrey:
@rmosolgo sweet! :+1: I should be looking into this shortly. I have a couple other use-cases for instrumentation that we can discuss at the same time.

PR #291, opened 2016-10-04
Raise when resolved_type not part of possible types

xuorig:
Raise when Schema resolves an object to a Type, but that Type does not implement the interface. ( Is not part of possilbe_types)

Example in tests :)

@rmosolgo


rmosolgo:
:+1: Thanks! I bet this will be a huge DX improvement.

rmosolgo:
(oh, I should let the tests finish ...)

xuorig:
they're done now 💃

PR #292, opened 2016-10-04
Expose problems to VariableValidationError message.

colepatrickturner:
It took me a while to track down which argument was causing a validation error on my input object. This change exposes the problems (along with path and explanation in a message).

### DIFF
--- a/lib/graphql/query/variable_validation_error.rb
+++ b/lib/graphql/query/variable_validation_error.rb
@@ -8,7 +8,8 @@ module GraphQL
         @validation_result = validation_result
 
         problems = validation_result.problems.map do |p|
-          "#{p['explanation']} on path '#{p['path'].join(" > ")}'"
+          path_str = p['path'].size > 0 ? " on path '#{p['path'].join(" > ")}'" : ""
+          "#{p['explanation']}#{path_str}"
         end
 
         msg = "Variable #{variable_ast.name} of type #{type} was provided invalid value with problems: \n#{problems.join('\n')}"
--- a/lib/graphql/query/variable_validation_error.rb
+++ b/lib/graphql/query/variable_validation_error.rb
@@ -7,7 +7,11 @@ module GraphQL
         @value = value
         @validation_result = validation_result
 
-        msg = "Variable #{variable_ast.name} of type #{type} was provided invalid value"
+        problems = validation_result.problems.map do |p|
+          "#{p['explanation']} on path '#{p['path'].join(" > ")}'"
+        end
+
+        msg = "Variable #{variable_ast.name} of type #{type} was provided invalid value with problems: \n#{problems.join('\n')}"
         super(msg)
         self.ast_node = variable_ast
       end
--- a/spec/graphql/analysis/analyze_query_spec.rb
+++ b/spec/graphql/analysis/analyze_query_spec.rb
@@ -72,7 +72,7 @@ describe GraphQL::Analysis do
 
       it "returns an error" do
         error = query.result["errors"].first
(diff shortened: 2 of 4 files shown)


rmosolgo:
That's a nice improvement! It looks like some tests need updating, too. 

What do you think, is there a better way to handle the case where  the message includes `on path ''` ?

colepatrickturner:
I could think of two scenarios:

A) Returning all the problems, after validation.
B) Returning the first problem, halting validation.

If a single validation holds back a mutation from executing, then it makes sense to halt execution and return that single problem with the specific path: "Failed to coerce this field to this value on path InputObjectType > field_name > argument" along with some expectations ie. "Expected: Float,  Received: String."

On the other hand, it's nice to have all of the problems up front, after validation. This way you can tackle more than one problem in a single go-around. That feels better to me than fixing a problem, restarting and finding another error on the next execution. 

I imagine this really only matters in the case of rescued exceptions, which is what I experienced. In that case, I prefer having all the information available since it's faster than digging deeper to see what's really at fault. Throwing a validation error on the entire input object is not enough information.

rmosolgo:
I totally agree with you when it comes to giving errors back to the user, returning as many as possible is a big win!

But if I understood you correctly (and remember correctly), this code _raises_ on the first error it finds, so it doesn't support flow. Boo hiss! Do you want to take a stab at that refactor now? 

If you do, I'm happy to help out any way I can. If now's not good, I'm happy to merge an improvement like this one, but I think we shouldn't send `path ''` back to the user, I don't think that's very helpful!

colepatrickturner:
Aha I see sorry I misunderstood. This code will return all of the problems for the first ValidationError, correct. That will occur upon the first failed validation result as shown here:
https://github.com/rmosolgo/graphql-ruby/blob/bf4bc89794cd677867b3ea819de0a44671249a9d/lib/graphql/query/variables.rb#L32

I've added another commit to only show that when there is a known path.

rmosolgo:
Looks good to me, do you mind updating those failing tests?

colepatrickturner:
Sure! It may take me a couple tries but I'll get it to pass.

rmosolgo:
Looking again, is this duplicate with the `"problems"` key in the JSON? Would it be sufficient to reorder the hash keys for this? 

Here's an example error: 

image

colepatrickturner:
I'm not sure, it seems to fail at this line: https://travis-ci.org/rmosolgo/graphql-ruby/jobs/165064030#L490

rmosolgo:
Oops, let me rephase that: 

Given that the same information is already available in the `errors` response, is this PR really necessary?

This PR takes the info from the `"problems"` key and adds it to the `"message"`. I can see how this improves the visibility of that info, but could we achieve the same result by reordering the keys in the error's JSON? (_Instead_ of adding it to the message?)

colepatrickturner:
You're right that the error is exposed in the errors returned from Schema.execute, though if you wish to rescue and catch `ExecutionError` or any other error along the way, what's returned from `execute` is irrelevant.

**Variable input_0 of type InsertMyObjectInput! was provided invalid value** is too general. In my case it was coercion of a string to a float error, which is also not easy to pin down given that `graphql-ruby` has strict preference to a number type, not allowing string.

I believe more specific information is pertinent to the error message, and one should not have to dig as deep as I did to learn that the cause was in another property.

rmosolgo:
Sorry, I'm not convinced this info should be in the error twice. If you think the `"problems"` key in the JSON response isn't helpful, maybe we could remove that and show the error some other way!

PR #293, opened 2016-10-04
Add parent type name to the InvalidNullError message

dylanahsmith:
When investigating an invalid null error, the first thing I end up having to do is find the type for the field, which can take a big more work when we have the error occurs on a generic field name.  E.g. multiple object types may have a field named `:name` or `:id`.  Having the type name in the error message would allow us to avoid that extra work to debug an error.


rmosolgo:
Sweet!

PR #294, opened 2016-10-05
Handle nonnullable root query execution errors

josh:
The new `"path"` on errors is looking good, but I think I came up with an edge case that I'm unsure if the correct behavior or not.

When an execution error comes from a root query field that is nonnullable it forces `"data"` to be entirely `null`. That makes sense, but how should `"path"` behave in this case? The current implementation records the field that errored as `"path"`, however it doesn't exist at all.

I was kinda expecting that for any error `"path"`, you could access `data[path[0]][path[1][path[2]...`. Should we ensure this property holds true for this root edge case and return an empty `[]` path?

Would be interesting to check how `graphql-js` handles this.

### DIFF
--- a/spec/graphql/execution_error_spec.rb
+++ b/spec/graphql/execution_error_spec.rb
@@ -171,4 +171,27 @@ describe GraphQL::ExecutionError do
       assert_equal(expected_result, result)
     end
   end
+
+  describe "error on root query field" do
+    let(:query_string) {%|
+    query MilkQuery {
+      root
+      nonnullableExecutionError
+    }
+    |}
+
+    it "data is empty" do
+      expected_result = {
+        "data" => nil,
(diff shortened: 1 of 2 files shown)


josh:
Still not sure what is the *correct answer* here, any guidance @rmosolgo?

rmosolgo:
No, I don't really know either! `"path"` isn't in the spec and personally, I don't have a use for it (and I haven't seen a use, except for what you mentioned a while back).

josh:
Hmm, okay I suppose the current behavior does match graphql-js.

https://github.com/graphql/graphql-js/blob/89f3f1a319bb913616f2910e8a806170155ffaf0/src/__tests__/starWarsQuery-test.js#L474-L530

PR #295, opened 2016-10-05
Support partial ExecutionError in lists

josh:
For single values, resolvers can return a `GraphQL::ExecutionError` instance. It'd be great if lists offered a single ability to fail partially. Its a bit of an implementation edge case since individual list values don't have their own `resolve` function.

For GitHub's GraphQL API, I'd like our root `nodes(ids: [ID!]!): [Node]!` batch lookup to fail partially if a given ID fails to resolve. The returned collection itself is non-nullable, but the list values are nullable `Node`s.

``` graphql
type Query {
  node(id: ID!): Node
  nodes(ids: [ID!]!): [Node]!
}
```


rmosolgo:
:rocket: !

josh:
@rmosolgo ❤️

PR #296, opened 2016-10-06
Move invalid null rescue for graphql-batch execution strategy.

dylanahsmith:
For https://github.com/Shopify/graphql-batch/issues/18

The execution strategy in the graphql-batch gem overrides FieldResolution#get_finished_value because the resolution hasn't been completed, so it chains the promise to passes the resolved result using `super` rather than the promise object.

However, https://github.com/rmosolgo/graphql-ruby/pull/174 added error handling into FieldResolution#result, which only receives a promise from get_finished_value, so the rescue effectively gets skipped when using graphql-batch.

This pull request moves the rescue into FieldResolution#get_finished_value to avoid the need to have to duplicate the invalid null rescue in the graphql-batch execution strategy.

### DIFF
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -15,17 +15,8 @@ module GraphQL
 
         def result
           result_name = irep_node.name
-          begin
-            raw_value = get_raw_value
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: sometime soon I'd like to provide a better API for graphql batch so it's less dependent on "internals" !

PR #297, opened 2016-10-06
Fix bug in schema printer for enum values with custom ruby value.

dylanahsmith:
cc @xuorig & @cjoudrey 

## Problem

When trying to upgrade to the newest release of this gem we noticed that our schema dump changed for enum values.  This can be seen with this PR's test changes for this regression without the changes to the code under test:

```
GraphQL::Schema::Printer::.print_schema
  test_0001_returns the schema as a string for the defined types  FAIL (0.01s)
Minitest::Assertion:         --- expected
        +++ actual
        @@ -26,7 +26,7 @@
         }

         type Query {
        -  post(id: ID!, varied: Varied = {id: \"123\", int: 234, float: 2.3, enum: FOO, sub: [{string: \"str\"}]}): Post
        +  post(id: ID!, varied: Varied = {id: \"123\", int: 234, float: 2.3, enum: foo, sub: [{string: \"str\"}]}): Post
         }

         input Sub {
        @@ -38,6 +38,6 @@
           int: Int
           float: Float
           bool: Boolean
        -  enum: Choice = FOO
        +  enum: Choice = foo
           sub: [Sub]
         }"
        /Users/dylansmith/src/graphql-ruby/spec/graphql/schema/printer_spec.rb:218:in `block (3 levels) in <top (required)>'
```

It looks like this is a regression from https://github.com/rmosolgo/graphql-ruby/pull/267

## Solution

Use coerce_result to convert the ruby value into the name for the GraphQL enum value in the schema printer.

### DIFF
--- a/lib/graphql/schema/printer.rb
+++ b/lib/graphql/schema/printer.rb
@@ -111,7 +111,7 @@ module GraphQL
             when ScalarType, ID_TYPE, STRING_TYPE
               value.to_s.inspect
             when EnumType
-              value.to_s
+              type.coerce_result(value)
             when InputObjectType
               fields = value.to_h.map{ |field_name, field_value|
                 field_type = type.input_fields.fetch(field_name.to_s).type
--- a/spec/graphql/schema/printer_spec.rb
+++ b/spec/graphql/schema/printer_spec.rb
@@ -11,8 +11,8 @@ describe GraphQL::Schema::Printer do
     choice_type = GraphQL::EnumType.define do
       name "Choice"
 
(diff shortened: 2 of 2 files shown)


rmosolgo:
:ok_hand: 

Too bad that was ambiguous in the previous test, thanks for fixing that too!

xuorig:
Mind cutting a release with this one 😅🙏🙏?

rmosolgo:
:ship: 0.19.2 !

xuorig:
❤️

PR #299, opened 2016-10-06
Pass root value to mutation resolve

rmosolgo:
Mutation's `resolve` function now receives the `root_value` given to `Schema#execute`. The signature is changed to: 

```ruby 
-> (obj, input, ctx)  { ... }
```

The previous signature (`(input, ctx)`) is deprecated.

Also use a derived class for mutation results so we don't have to `method_missing`.


lib/graphql/relay/mutation.rb
```diff
@@ -121,24 +127,49 @@ def input_type
...
+        def initialize(client_mutation_id:, result:)
+          @client_mutation_id = client_mutation_id
+          result.each do |key, value|
+            self.public_send("#{key}=", value)
           end
         end
+
+        def self.define_subclass(mutation)
+          subclass = Class.new(self) do
+            attr_accessor(*mutation.return_fields.keys)
```
rmosolgo: I thought I would make sure this was worth the trouble: 

```ruby 
require "benchmark/ips"

class HashLookup
  def initialize(storage)
    @storage = storage
  end

  def method_missing(name, *args, &block)
    if @storage.key?(name)
      @storage[name]
    else
      super
    end
  end
end

class BaseAccessorLookup
  def initialize(storage)
    storage.each do |key, value|
      self.public_send("#{key}=", value)
    end
  end

  def self.define_subclass(accessors)
    Class.new(self) do
      attr_accessor(*accessors)
    end
  end
end

AccessorLookup = BaseAccessorLookup.define_subclass(:hit_key)

hash_lookup = HashLookup.new({hit_key: true})
accessor_lookup = AccessorLookup.new({hit_key: true})

Benchmark.ips do |x|
  x.report("method-missing") { hash_lookup.hit_key }
  x.report("class.new") { accessor_lookup.hit_key }

  x.compare!
end
```

```
~/code/graphql $ ruby lookup_benchmark.rb
      method-missing   193.115k i/100ms
           class.new   266.421k i/100ms
      method-missing      3.545M (±17.0%) i/s -     17.187M in   5.193915s
           class.new      7.769M (±20.4%) i/s -     36.500M in   5.011390s

Comparison:
           class.new:  7768845.5 i/s
      method-missing:  3544535.8 i/s - 2.19x  slower
```

Issue #269, opened 2016-09-21
Schemas with shared members 


rmosolgo:
There's a compelling case for multiple schemas built from the same set of schema members (eg, instances of `GraphQL::ObjectType` or `GraphQL::Field`): 

- Schema A is private and exposes all kinds of data; it's for internal, server-to-server communication 
- Schema B is public and exposes a subset of Schema A; it's for external users (or a JavaScript front-end). 

In this case, some members from Schema A should not even be _visible_ in Schema B: they shouldn't appear in introspection queries and attempts to access them should be treated as accessing a non-existing member. 

Since Schema B is a subset of Schema A, it's desirable to share code between the two. However, a simple problem makes this hard right now: schema members are held in common between the two schemas, so any attempt to modify _one_ of those objects will affect _both_ schemas. 

How can we support this use case? I can think of a few possibilities (and more than a few problems), here they are: 

- __Do nothing.__ Users should write their own Ruby code to describe their system, then build two independent schemas composed of new objects for each schema. This is basically how programmatically-generated fields work currently: you have to regenerate the field each time you use it. Except now we're talking about an entire schema! __Pros__ are: very easy. __Cons__ are: we waste a lot of user time for problem that could be common. Besides that, we essentially ask people to reimplement GraphQL schema definition, which is sad. 
- __Schema deep-copy.__ Provide an API like `Schema#filter(&block)` which recursively copies Schema members, calling `block` to determine whether each member should be included in the new schema. This could work, but we'd probably still end up sharing some user-provided values, like resolve functions or argument default values. (Or, could we copy those too?) __Pros__ are: it would work. __Cons__ are: it will be hard to get right. Depending on what we copy, and how we copy it, it could lead to some surprising behaviors if users modify objects after-the-fact. 
- __Dynamic schema member resolution.__ Accept query-specific input (eg `context[:permission]`) and check that anytime you fetch schema members (types, fields, arguments). __Pros__ are: this is easy for users. You hook up metadata to your objects, then define how that query-specific input should be checked against metadata. __Cons__ are: it will be hard to find all of these places in the gem and update them. Until now, we have assumed that a schema's members are fixed after `Schema.define`. If we introduce query-level schema membership, it will be really hard to find where those assumptions are made and ... unmake them!
- Ok, here's a weird one: __Schema "masks".__ Assuming Schema B is a subset of Schema A, what if we could use Schema B as a "mask" for Schema A, meaning that: 
[...]

PR #300, opened 2016-10-08
Query-level schema restrictions

rmosolgo:
Here's a take on #269

For a given query, make a schema appear _smaller_ than it really is. This way, you can restrict access (even visibility) to certain types, fields, arguments or enum values within a schema. 

```ruby 
# Objects are hidden when `.call(schema_member)` returns truthy
top_secret = -> (schema_member) { schema_member.metadata[:top_secret] }
Schema.execute(query_string, except: top_secret)
```

In fact, the only requirement of a "mask" is that it responds to `#call(member)`, so you can write your own, too!


cjoudrey:
> One reason I'm trying a dynamic approach is because it could lay the groundwork for query-level masks. Once the filtering is in place, you can add the query as an input, too.

Love that you're thinking about this query-based use-case.

@xuorig and I were talking about this last week for feature-flagged fields. Imagine you want to give a specific API client early-access to some feature being built. There is far too many possible permutations to pre-calculate all possible schemas in advance.

Instead you might be able to have a query-analyzer (or more a query validator) to prevent unauthorized clients from resolving the feature-flagged fields and some way to hook into the introspection schema to omit the feature-flagged fields/types when unauthorized client requests schema.

kdaigle:
> Instead you might be able to have a query-analyzer (or a static validator) to prevent unauthorized clients from resolving the feature-flagged fields. You'd also want some way to hook into the introspection schema to omit the feature-flagged fields/types when unauthorized client requests schema.

We've discussed this approach internally as well and become concerned that you'd need resolved objects to possibly check feature flags against which you couldn't do in a query analyzer. I like the idea of the masks quite a bit but, as @cjoudrey pointed out, eventually there would be some crazy multiple of permutations. Hmm.

cjoudrey:
@kdaigle I _think_ you could achieve this with a query analyzer since you have access to the `GraphQL::Query`, the query's `context` and the type/field definitions for the query.

Suppose you added a `GraphQL::Field.accepts_definitions(required_feature_flag: ...)` and have `context[:api_client]` you should be able to write a query analyzer that does something like:

```ruby
def call(memo, visit_type, irep_node)
  if irep_node.ast_node.is_a?(GraphQL::Language::Nodes::Field) && visit_type == :enter
    irep_node.definitions.each do |type_defn, field_defn|
      if field_defn.required_feature_flag && memo[:context].has_feature_flag?(field_defn.required_feature_flag)
        # raise GraphQL::ExecutionError or AnalysisError or context.errors.add
      end
    end
  end
end
```

kdaigle:
Yeah, that'd work assuming you're keying off of something that's always in `context`. Sometimes, we key off of objects that aren't in `context` (for example, a repository and not the current user). We'd need that repository before we could flag it. It's not a deal breaker at all but I figured I'd mention that, for us, we'd likely need to rely heavily on the masks.

rmosolgo:
An analyzer might work, depending on your privacy requirements. For example, if a certain enum value is only visible to some users, you couldn't remove it from an introspection result with an analyzer alone. 

That would tell GraphiQL users that a certain value is valid for them ... even though it's not 😿  

Even if an analyzer-based solution did _work_, it would be bummer to basically reimplement some gem internals (validation, introspection at least) in user code.

That's why I want to provide this in the library. It's a really nice feature for large-scale GraphQL schemas and the best implementation is ... right at the heart of the library :) 

I pushed a hot take on query-level masking: 2260b7e It seems promising!

cjoudrey:
> For example, if a certain enum value is only visible to some users, you couldn't remove it from an introspection result with an analyzer alone.

Correct, you need both. The analyzer was simply to prevent people from using the hidden things. The introspection resolvers still need to be modified in order to hide the things in the first place.

> It's a really nice feature for large-scale GraphQL schemas and the best implementation is ... right at the heart of the library :)

:100:

> I pushed a hot take on query-level masking: 2260b7e It seems promising!

:+1:

lib/graphql/define/defined_object_proxy.rb
```diff
@@ -10,6 +10,11 @@ def types
         GraphQL::Define::TypeDefiner.instance
       end
 
+      # TODO: do I actually want to add this?
```
cjoudrey: Could you think of other use cases where `metadata` can be helpful in the future?

rmosolgo: It's definitely useful -- it's the basis for things like `field_defn.required_feature_flag` in your example above (that is, arbitrary data attached to schema members).

Right now it's only supported via 

```ruby
GraphQL::BaseType.accepts_definitions(
  required_feature_flag: GraphQL::Define.assign_metadata_key(:required_feature_flag)
)
```

and then 

```ruby
  required_feature_flag :top_secret_access 
end 
```

But with this you could "just" do 

```ruby 
  metadata :required_feature_flag, :top_secret_access
end 
```

I added it because I wanted to mess with `metadata` without extending every schema in the test suite! Maybe it would be useful IRL to have a helper like that too.

rmosolgo: But, it's more prone to typos, that's a bit sad

rmosolgo:
You can see the API I sketched out here: https://github.com/rmosolgo/graphql-ruby/blob/bf75ccbcd0b9c4051338a2275b635194025a5c50/lib/graphql/schema/mask.rb#L3

How does that look to you?

I was considering between these two: 

```ruby 
# 1) Yield the schema member 
Schema::Mask.new { |member| ... } 
# 2) Yield the schema member and the query context 
Schema::Mask.new { |member, ctx| ... }
```

I opted for `1)` so that a mask could be independent of a `GraphQL::Query` instance. That way, in the future, you could apply a mask to a schema dump (`Schema::Printer.print_schema`), too.

You can accomplish the equivalent of `2)` by closing over query-level values with a function or an object: 

```ruby 
@current_user = user_from_session
# Access outer scope in the block 
mask = Schema::Mask.new { |member| @current_user.can_see?(member) } 
# Or pass it into another object
mask = CustomUserBasedMask.new(@current_user)

Schema.execute(query_string, mask: mask)
```

jamesmacaulay:
About the `Mask` class: personally I didn't notice at first that the result of the provided block is negated in `Mask`'s implementation of `visible?`. The fact that the block result is the opposite of the method result was confusing to me, and something that I think I would always end up having to look up or figure out through trial and error.

One alternative would be to use bare `Proc` objects without any `Mask` wrapper, and use a keyword param that is more explicit about the interface. For example:

```ruby
top_secret = ->(schema_member) { schema_member.metadata[:top_secret] }
Schema.execute(query_string, except: top_secret)
```

Just an idea.

rmosolgo:
:+1: for `except: <#call(member)>` API! That's much clearer, I'll make the refactor soon.

rmosolgo:
image

image


Also, implementing `only:` would be a breeze from here if we ever need it!

lib/graphql/define/defined_object_proxy.rb
cjoudrey: I asked because I'm always a bit wary with exposing key/value stores like that. lol

> it's more prone to typos, that's a bit sad

Right!

```ruby
  required_feature_flag :top_secret_access 
end
```

Letting people expose this on their own via `accepts_definitions` felt less prone to errors (and typos). It also means we don't need to expose `metadata` in the `define` block.

I could be wrong though. :)

rmosolgo: :+1: I [moved it to `spec_helper.rb`](https://github.com/rmosolgo/graphql-ruby/pull/300/commits/7b91e3ff109369bd681f7e7b05959eace0fc0007#diff-93830fa29d616f7c87903d08b5b1b29aR22) so it's there for testing but not default for the gem :)

rmosolgo:
Oops, realized I missed a spot with _variables_, where it was still possible to pass in hidden enum values or input object fields. If anyone gets a chance to think through any other "leaks", I'd love to hear the outcome of that!

rmosolgo:
I added some docs and lots of caching on `Warden`'s result, but this feature still adds quite a bit of overhead, running the introspection schema is 10% slower 

<details>
```
Switched to branch 'master'
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     33.294  (± 3.0%) i/s -    168.000  in   5.050755s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     33.386  (± 3.0%) i/s -    168.000  in   5.037097s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     34.060  (± 2.9%) i/s -    171.000  in   5.026944s
~/code/graphql $ git co masks
Switched to branch 'masks'
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     2.000  i/100ms
           run query     29.056  (± 6.9%) i/s -    146.000  in   5.039352s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     2.000  i/100ms
           run query     31.288  (± 6.4%) i/s -    158.000  in   5.064336s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     2.000  i/100ms
           run query     29.425  (± 3.4%) i/s -    148.000  in   5.039897s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     26.531  (±15.1%) i/s -    132.000  in   5.067148s
```
</details>

To be honest, it's not even clear that the caching improved the situation, so maybe I should remove it: 

<details>
```
~/code/graphql $ git co 065e9bde5f12d336eb368fc1ed834fdf94fd11a7
HEAD is now at 065e9bd... refactor(BaseType#validate_input) thread the Warden through all input validation
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     30.811  (± 6.5%) i/s -    156.000  in   5.082216s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     31.348  (± 3.2%) i/s -    159.000  in   5.076198s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     32.714  (± 3.1%) i/s -    165.000  in   5.050744s
~/code/graphql $ git co 9de13412b5b16500263299f189db429f22f18434
Previous HEAD position was 065e9bd... refactor(BaseType#validate_input) thread the Warden through all input validation
HEAD is now at 9de1341... feat(Warden) cache generated results too
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     3.000  i/100ms
           run query     31.463  (± 3.2%) i/s -    159.000  in   5.060034s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     2.000  i/100ms
           run query^[[A     31.937  (± 3.1%) i/s -    160.000  in   5.019291s
~/code/graphql $ ruby -Ilib:spec/support introspection_benchmark.rb
           run query     2.000  i/100ms
           run query     31.050  (± 3.2%) i/s -    156.000  in   5.032587s
```
</details>

rmosolgo:
I'm gonna push without caching since it doesn't seem to be much of a win, and make up the time someplace else.

cjoudrey:
:tada:!

PR #303, opened 2016-10-11
[minor] Adding missing descriptions to default relay objects

brandonblack:
We recently noticed that [some default relay objects were missing descriptions](https://developer.github.com/early-access/graphql/interface/node/). This super minor PR pulls in descriptions for edges, nodes, cursors, etc. to match what can be found the reference implemtation for relay over in [graphql/graphql-relay-js](https://github.com/graphql/graphql-relay-js).

This is of course, just default documentation mirroring graphql-relay-js and descriptions can be overridden at will.


rmosolgo:
:metal:

PR #305, opened 2016-10-12
Add descriptions to IDL

cjoudrey:
Implementation wise I've ported a lot of the logic over from https://github.com/graphql/graphql-js/pull/464 including how they break up comments into sublines of ~80 chars.

Tokens now have a reference to prev/next token. This was needed in order to obtain full-line contiguous comment blocks preceding a statement.

Comments are still ignored by the parser (i.e. not emitted) in order to not be too strict about where you can place them as per the spec:

> Comments behave like white space and may appear after any token, or before a line terminator, and have no significance to the semantic meaning of a GraphQL query document.

I've had to add `GraphQL::Language::Comments` to share some line splitting logic between `Schema::Printer` and `Language::Generator`. I'm not :100: convinced on the name of this module so feel free to suggest another name/approach.

👀 @rmosolgo


lib/graphql/language/nodes.rb
```diff
@@ -17,9 +17,6 @@ def initialize(options={})
           if options.key?(:position_source)
             position_source = options.delete(:position_source)
             @line, @col = position_source.line_and_column
-          elsif options.key?(:line)
```
cjoudrey: As far as I can tell this is dead code.

rmosolgo: I thought it might have snuck in there for `graphql-libgraphqlparser` but looks like not: 
https://github.com/rmosolgo/graphql-libgraphqlparser-ruby/blob/master/ext/graphql_libgraphqlparser_ext/visitor_functions.c#L58-L63

lib/graphql/schema/printer.rb
```diff
@@ -84,10 +84,62 @@ def print_deprecated(field_or_enum_value)
           end
         end
 
+        module DescriptionPrinter
+          def print_description(definition, indentation='', first_in_block=true)
+            return '' unless definition.description
+
+            lines = definition.description.split("\n")
+
+            description = indentation && !first_in_block ? "\n" : ""
```
rmosolgo: I've only noticed values of `""` and `"  "` for `indentation`, are there possible falsey values? If not, maybe this could be 

```ruby 
description = first_in_block ? "" : "\n"
```

?

lib/graphql/schema/printer.rb
```diff
@@ -84,10 +84,62 @@ def print_deprecated(field_or_enum_value)
...
+              end
+            end
+
+            description
+          end
+
+          def break_line(line, length)
+            return [line] if line.length < length + 5
+
+            parts = line.split(Regexp.new("((?: |^).{15,#{length - 40}}(?= |$))"))
```
rmosolgo: Can you help me understand this regexp? I'm certainly not opposed to regexps but I can't quite grok this one 😬 

Maybe its ... trying to find natural line break positions?

cjoudrey:
> That's interesting about linking the tokens ... is it basically because the `COMMENT` token can't wrap newlines?

Descriptions are formed by concatenating contiguous comment lines.

For instance:

```graphql
# This WILL NOT be part of the description for Car

# This WILL be part of the description
# This WILL too
type Car {
```

I originally built it by emitting `COMMENT` tokens and prefixing parser rules with `comment_list_opt` for nodes that could accept comments.

Where it got tricky is that `COMMENT` tokens are allowed to be anywhere, it’s just that if they precede certain nodes we consider them special.

That’s when I took a look at `graphql-js` and noticed how they implemented it which is what I implemented here.

I’m not sure if that answers your question.

>  The alternative is to print the user-provided description verbatim, and leave line breaks to them.

Yeah, I’m on the fence about this too. According to the spec, `descriptions` may contain Markdown. While slicing plain text works fine, I’d be worried if we sliced in the middle of a Markdown table or something like that.

I don’t mind reverting the line breaking code and worse case we can re-revert in the future.

lib/graphql/schema/printer.rb
cjoudrey: This was bad porting on my end, `""` is falsey in JS, but in Ruby it isn't.

This should be `indentation != "" && !first_in_block`.

cjoudrey:
@rmosolgo this should be ready for another round of review 👀 .

Let me know if you want me to change anything. 😀

lib/graphql/language/parser.y
```diff
@@ -372,6 +372,24 @@ def next_token
...
+    break if token.nil?
+    break if token.name != :COMMENT
+    break if token.next_token.line != token.line + 1
+
+    comments << token.to_s[2..-1] || ""
+  end
+
+  return nil if comments.empty?
+
+  comments.reverse.join("\n")
```
rmosolgo: I guess since this is only for schemas, this isn't a super hot path, but if we want to avoid the intermediary array from `.reverse` later, I think we could use `.unshift`, eg

```ruby 
loop do 
  prev_token = token 
  token = token.prev_token
  if token && token.name == :COMMENT && prev_token.line == token.line + 1
    comments.unshift(token.to_s.sub(/^#\s*/, "")
  end 
end 

comments.join("\n")
```

lib/graphql/language/parser.y
```diff
@@ -372,6 +372,24 @@ def next_token
...
+  comments = []
+
+  loop do
+    token = token.prev_token
+
+    break if token.nil?
+    break if token.name != :COMMENT
+    break if token.next_token.line != token.line + 1
+
+    comments << token.to_s[2..-1] || ""
```
rmosolgo: This assumes `"# "` for leading a comment, right? Should we support `#` _without_ a space?

lib/graphql/language/token.rb
```diff
@@ -4,13 +4,15 @@ module Language
     # Contains type, value and position data.
     class Token
       # @return [Symbol] The kind of token this is
-      attr_reader :name
+      attr_reader :name, :prev_token, :line
+      attr_accessor :next_token
```
rmosolgo: If we do it this way in `get_description`: 

```ruby 
loop do 
  next_token = token 
  token = token.prev_token 
  # ...
  if next_token.line == token.line + 1 
    # ... 
  end 
end 
```

Could we drop `Token#next_token`?

If that's the only use for `next_token`, we might be able to save ourselves the assignment for nodes that don't need it!

spec/graphql/static_validation/rules/fragments_are_used_spec.rb
```diff
@@ -4,7 +4,7 @@
   include StaticValidationHelpers
   let(:query_string) {"
     query getCheese {
-      name,
+      name
```
rmosolgo: Is this change part of the feature? (If so ... I'm missing something!)

cjoudrey: No, just a typo I noticed. `,` aren't required here.

I could break this up into a separate commit if you prefer.

lib/graphql/language/token.rb
cjoudrey: Great catch! :)

rmosolgo:
:+1: looks good to me, do you mind rebasing?

cjoudrey:
> do you mind rebasing?

Done :+1:

rmosolgo:
🚀 !

PR #308, opened 2016-10-12
Support 7 level introspection query instead of 3

tahnok:
There are some schemas that require deeply nested types, such as a 3-dimensional array of non-null Ints (`[[[Int!]!]!]!`) that the current introspection query does not handle. Instead of 3, change the `ofType` query to go 7 deep.

It looks like graphql-js added support for this https://github.com/graphql/graphql-js/pull/364. As the introspection query in the gem is from that repo, I think it's safe to pull this is an as well.

Is there a better spot for this test?


rmosolgo:
This is great, thanks!

PR #312, opened 2016-10-15
Cleanup order in base connection

farism:
Looking at the API docs and seeing http://www.rubydoc.info/github/rmosolgo/graphql-ruby/GraphQL/Relay/BaseConnection#order-instance_method confused me for a good 15 minutes, so I dug through history and found https://github.com/rmosolgo/graphql-relay-ruby/pull/33.

It doesn't seem like this was cleaned up 100% and artifacts are still showing in the docs, so I'm hoping this is the right PR here. Let me know if it needs any changes, or if I have the wrong understanding. Thanks!


rmosolgo:
Yes, I think you're right, this stuff is left over! Thanks!

PR #306, opened 2016-10-12
Ruby 1.9.3 support for us dinosaurs

astorije:
Do not even think about merging this.
Do not even think about running this in production.

Opening this PR nonetheless to see Travis CI run and give its report. See #148 for more details.

This will be used to run a 24-hour hackathon, and might stay around if folks are willing to experiment and prototype with GraphQL on their outdated projects.

PR #313, opened 2016-10-15
Replace `-> (...)` with `->(...)` to be Ruby 1.9.3 compatible

astorije:
See http://ruby-journal.com/becareful-with-space-in-lambda-hash-rocket-syntax-between-ruby-1-dot-9-and-2-dot-0/ for more info.

Extracted from #306 which will be easier to maintain without that commit.

Note that this is also how http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/Lambda defines lambdas like this as well.

### DIFF
--- a/CHANGELOG-relay.md
+++ b/CHANGELOG-relay.md
@@ -109,7 +109,7 @@
    ```ruby
    field :cities, CityType.connection_type do
      argument :order, types.String, default_value: "name"
-     resolve -> (obj, args, ctx) {
+     resolve ->(obj, args, ctx) {
        obj.order(args[:order])
      }
    end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,13 +17,13 @@
   Previously, it was called with two arguments:
 
   ```ruby
-  resolve -> (inputs, ctx) { ... }
+  resolve ->(inputs, ctx) { ... }
   ```
 
   Now, it's called with three inputs:
 
   ```ruby
-  resolve -> (obj, inputs, ctx) { ... }
+  resolve ->(obj, inputs, ctx) { ... }
   ```
 
   `obj` is the value of `root_value:` given to `Schema#execute`, as with other root-level fields.
@@ -92,11 +92,11 @@
     ```ruby
     MySchema = GraphQL::Schema.define do
       # Fetch an object by UUID
-      object_from_id -> (id, ctx) {
+      object_from_id ->(id, ctx) {
         MyApp::RelayLookup.find(id)
       }
       # Generate a UUID for this object
-      id_from_object -> (obj, type_defn, ctx) {
+      id_from_object ->(obj, type_defn, ctx) {
         MyApp::RelayLookup.to_id(obj)
       }
     end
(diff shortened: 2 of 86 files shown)


rmosolgo:
I'm ok with this, but I think we should add some kind of safeguard to prevent these from slipping back in as files are changed. Otherwise, I'll forget, and start adding more `-> (...`! For example, I currently have a branch of documentation rewrites, and I'm sure it's full of that. 

What do you think about adding something to the unit tests that makes sure this rule is followed? It would be ok with me to do something like:

```ruby 
it "uses 1.9.3-compatible stabby literals" do 
  invalid_stabby_pattern = /->\s+\(/
  assert_project_files_dont_match(invalid_stabby_pattern)
end 
```

(Implementing `assert_project_files_dont_match` is a TODO :P )

Or, if you have another suggestion, I'm open to something else! I just don't trust myself to remember this style.

rmosolgo:
Alternatively, is this transform predictable enough to be maintained as a git hook? I mean, if it's a "simple" find and replace, maybe it wouldn't be so bad to maintain it on the 1.9.3 branch. Just a thought, I'm not a git hook user so I can't really take it farther than that :P

astorije:
> I think we should add some kind of safeguard to prevent these from slipping back in as files are changed.

Fine by me.

> Otherwise, I'll forget, and start adding more -> (...! For example, I currently have a branch of documentation rewrites, and I'm sure it's full of that.

Well, until we enforce that, me rebasing the `1.9.3` branch on master and letting the almighty tests give their judgment would be a good enough non-permanent solution IMO...

> What do you think about adding something to the unit tests that makes sure this rule is followed?

I could do that as a fallback solution, but I don't want to cloud your test suite with something like that.

> Or, if you have another suggestion, I'm open to something else! I just don't trust myself to remember this style.

I looked a bit, and it seems to me that 99% of the web uses `->(`. This is why I opened https://github.com/bbatsov/rubocop/issues/3631. I'll see what the maintainers say on this and maybe we could merge this and pay attention manually until there is a cop for it that we can enforce here?
Sure, it wouldn't fix documentations but I think that would be the cleanest way to do so, the rest would come naturally.

> I mean, if it's a "simple" find and replace, maybe it wouldn't be so bad to maintain it on the 1.9.3 branch.

This doesn't seem very idiomatic in Ruby so I'd rather fix it upstream. Plus I don't know if this is compatible with all implementations of Ruby, such as JRuby.

> is this transform predictable enough to be maintained as a git hook?

Probably, but git hooks are not server side, and they aren't added to the repo, so any contributor would need to have the hook, definitely not something we can do.

Let's wait and see for what the Rubocop guys have to say about this. In the meantime, I am happy to periodically check for this should you decide to merge this without enforcing it automatically.

astorije:
@rmosolgo, author of RuboCop and the Ruby style guide agreed keeping the style without space is better on https://github.com/bbatsov/rubocop/issues/3631. Mind merging this then?

If no one is creating a cop for this, I'll look into how to do that myself when I have a bit of time.

rmosolgo:
:+1: cool, I'm happy to just _do my best_ for the time being! I guess I picked up habits for `->` from coffeescript and elixir, where it's usually padded with spaces.

PR #314, opened 2016-10-15
Use dot-notation for OpenStructs as [] is not available in Ruby 1.9.3

astorije:
This is also how doc (http://ruby-doc.org/stdlib-2.1.0/libdoc/ostruct/rdoc/OpenStruct.html) uses getters in most example.

This will make maintaining the Ruby 1.9.3 branch easier. Any strong opinion on this?
Extracted from #306.


rmosolgo:
:+1: no idea why I did it that way in the first place :P I wouldn't be surprised if they used to be hashes

PR #315, opened 2016-10-15
Add missing requires for OpenStruct objects

astorije:
Std-lib doc shows the needed `require` statement, not sure why the builds pass without it:
http://ruby-doc.org/stdlib/libdoc/ostruct/rdoc/OpenStruct.html

It was failing during the Ruby 1.9.3 experiments.
Extracted from #306.


rmosolgo:
:+1: I bet the one of the test dependencies requires it, so our error went unnoticed!

PR #316, opened 2016-10-16
Use JSON's generate+quirks_mode to allow dumping scalar values

astorije:
From https://github.com/rmosolgo/graphql-ruby/pull/306#issuecomment-254020617:

> One more thing, I'd like to get the `JSON`-related changes in `master`. Recently, someone else reported a bug for that, where another JSON implementation was 💥-ing.

Exported from #306. This was failing on Ruby 1.9.3 as well.


rmosolgo:
:+1: Awesome, I think this will also fix https://github.com/rmosolgo/graphiql-rails/issues/14 !

PR #317, opened 2016-10-16
Fix links to other guides in website

astorije:
Prior to this change, guides on the website were redirecting to the Rubydoc pages.

This also brings the advantage of not leaving the HTTPS world.

Of course, note that with this change, the links in the Rubydoc pages now redirect to the website...
I'll let you decide if this is an issue to you or not. Maybe you're perfectly aware and fine with that, I don't have an opinion, just giving you the option 😅.

### DIFF
--- a/guides/introduction.md
+++ b/guides/introduction.md
@@ -166,5 +166,5 @@ result_hash = Schema.execute(query_string)
 
 Read more in some other guides:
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: Thanks, I used to point to rubydocs _before_ there was a website!

PR #318, opened 2016-10-16
Use a more comprehensive Travis CI configuration

astorije:
This:

- Adds builds for Ruby 2.1.0, minimal version supported by the gem.
- Adds builds for latest Ruby 2.2 as it is supposed to be supported.
- Removes a duplicate job that was re-running Rails 5.
- ~~*Actually* runs an RBX build...~~ Nope, Travis CI definitely 💩 the 🛌 on Rubinius...
- ~~Allows for fast finish if the RBX build fails.~~ I ended up removing this as I removed Rubinius altogether.
- Removes Rubinius/RBX build. Note that since it was in the `allow_failures` section but not the `include` section, it was not even run. On current Travis CI, there is actually no way to successfully test with Rubinius at the moment, and I have tried with all versions returned by `rvm list known` on Travis servers and therefore supposedly supported... Lies! None of them are actually found!
- Mentions specific versions of Rubies because Travis's `rvm` is dumb (`2.2` loads `2.2.0`, `2.3` is reported as not existing...). Latest patch versions of Rubies were used.
- Removes that `rvm get stable` which was actually not helping getting anything 😓 It was even not run as it happens after RVM selects a Ruby version.
- ~~Upgrades versions of Active Record in the Gemfiles.~~ Not anymore.
- Removes Gemfile for Active Record 4.0, as latest release was in January 2015.
- Should make it easier to add further versions to the Travis CI configuration.

As a drawback, there are now a couple more builds running. I think it's worth it, but you tell me.


astorije:
Alright, I had quite a hard time on this one so you probably a loooot of commit activity, but I am now done :-)
Happy to discuss choices if not clear enough in the description.

rmosolgo:
:+1: for more ruby versions but I'm not sure the gem version changes are what I want. 

I used the `~>` so that I could get the most recent version where only the last digit changed. For example, `"activerecord", "~> 4.2.4"` actually installed [`4.2.7.1`](https://travis-ci.org/rmosolgo/graphql-ruby/jobs/168037917#L306), which is the latest `4.2.*` version. Similarly, `5.0.0` actually installed [`5.0.0.1`](https://travis-ci.org/rmosolgo/graphql-ruby/jobs/168037918). 

I chose to do it that way so we could test the latest "patch" release within minor versions. Is there a reason we should lock to more specific versions?

astorije:
I'm not too sure about that dependency bump, I thought it was a good idea to stay latest but actually nevermind. I removed the changes.

Are you OK dropping 4.0 and 4.1 or you want me to keep them? I see last 4.0 was released on January 2015 so I think we're definitely safe, but there was a security upgrade for 4.1 in July 2016 so we should probably keep it around.

I'll re-add here, let me know if you want me to remove again.

astorije:
[`master` currently builds in 2 min 22 sec](https://travis-ci.org/rmosolgo/graphql-ruby/builds/168037909) with 6 jobs. After re-adding Active Record 4.1, [this branch builds in 3 min 35 sec](https://travis-ci.org/rmosolgo/graphql-ruby/builds/168076883) with 11 jobs. I think that's pretty good!
Thanks for your input.

astorije:
Note that this PR will also make it easier for me to maintain a Ruby 1.9.3 branch!

rmosolgo:
:+1: I might remove some items from the `Ruby x Rails` matrix later. I've never _seen_ a case were a certain Ruby + Rails combo failed, and the same Ruby or same Rails passed with another complement. So in my experience, testing each Ruby _once_ and testing each Rails _once_ would be enough. But the only way I could figure out how to do that with Travis is to [write a TON of `exclude:`s](https://github.com/reactjs/react-rails/blob/master/.travis.yml#L24-L52) 😖

astorije:
To be fair, I am less concerned about failing tests than gem incompatibility between 2 values of the matrix (so failures at `bundle install` time).
But more importantly, with a total build time of 3 minutes and 30 seconds, I don't think you should worry too much about optimizing the builds at the moment. Should this time get closer to 10 minutes, improvements would be needed, but until then, I think you're safe 😅.

Feel free to post here if you want to get rid of some combinations later and I'll take a stab at it.

In the meantime, thanks, I'll be updating the `ruby-1.9.3` branch :)

PR #319, opened 2016-10-16
Fixed typo in testing guide

albertomg994:
There should be double curly braces, like in the docs:

http://www.rubydoc.info/github/rmosolgo/graphql-ruby/file/guides/testing.md


rmosolgo:
Oops, thanks!!

PR #320, opened 2016-10-16
Omit schema definition from IDL when type names match convention

cjoudrey:
As per graphql/graphql-js#470, schema definition should be omitted from IDL when root types match convention names: `Query`, `Mutation`, `Subscription`.

The diff might be easier to look at with `&w=1` in query params as I indented a big chunk in `generation_spec.rb`.

:eyes: @rmosolgo

### DIFF
--- a/lib/graphql/language/generation.rb
+++ b/lib/graphql/language/generation.rb
@@ -72,6 +72,12 @@ module GraphQL
         when Nodes::VariableIdentifier
           "$#{node.name}"
         when Nodes::SchemaDefinition
+          if (node.query.nil? || node.query == 'Query') &&
+             (node.mutation.nil? || node.mutation == 'Mutation') &&
+             (node.subscription.nil? || node.subscription == 'Subscription')
+            return
+          end
+
           out = "schema {\n"
           out << "  query: #{node.query}\n" if node.query
           out << "  mutation: #{node.mutation}\n" if node.mutation
--- a/lib/graphql/schema/printer.rb
+++ b/lib/graphql/schema/printer.rb
@@ -18,7 +18,7 @@ module GraphQL
       # Return the GraphQL schema string for the introspection type system
       def print_introspection_schema
         query_root = ObjectType.define do
-          name "Query"
+          name "Root"
         end
         schema = GraphQL::Schema.define(query: query_root)
         print_filtered_schema(schema, method(:is_spec_directive), method(:is_introspection_type))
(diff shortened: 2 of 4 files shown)


lib/graphql/schema/printer.rb
```diff
@@ -33,11 +33,18 @@ def print_filtered_schema(schema, directive_filter, type_filter)
...
         type_definitions = types.map{ |type| print_type(type) }
 
-        [print_schema_definition(schema)].concat(directive_definitions)
+        [print_schema_definition(schema)].compact
+                                         .concat(directive_definitions)
                                          .concat(type_definitions).join("\n\n")
       end
 
       def print_schema_definition(schema)
+        if (schema.query === nil || schema.query.name == 'Query') &&
```
rmosolgo: Why triple equals? (Ported from JS?) Would `.nil?` work here?

lib/graphql/language/generation.rb
```diff
@@ -72,6 +72,12 @@ def generate(node, indent: "")
         when Nodes::VariableIdentifier
           "$#{node.name}"
         when Nodes::SchemaDefinition
+          if (!node.query || node.query == 'Query') &&
```
rmosolgo: we expect this to be of type `nil | String` right? Could we reflect that with 

```ruby
node.query.nil? || node.query == "Query"
```

? That would make it more clear that `false` is not an expected value here.

lib/graphql/schema/printer.rb
cjoudrey: :facepalm: too much JS. lol

cjoudrey:
> possible JS leftovers :P

😭

Fixed 😀

spec/graphql/schema/printer_spec.rb
```diff
@@ -63,14 +63,47 @@
...
+    create_post_mutation = GraphQL::Relay::Mutation.define do
+      name "CreatePost"
+      description "Create a blog post"
+
+      input_field :title, !types.String
+      input_field :body, !types.String
+
+      return_field :post, post_type
+
+      resolve -> (_, _, _) { }
```
astorije: Could you remove spaces between the arrow and the parenthesis in your lambda definitions please? Thanks 😃

astorije: [There is RuboCop already](https://github.com/rmosolgo/graphql-ruby/blob/master/.rubocop.yml) but [that rule doesn't exist yet](https://github.com/bbatsov/rubocop/issues/3631).

rmosolgo:
sweet! 

btw @astorije if you want to make stabby syntax a requirement for changes, please add something to the test suite or CI. I think that will help contributors be more efficient with their time!

cjoudrey:
> btw @astorije if you want to make stabby syntax a requirement for changes, please add something to the test suite or CI. 

:+1: That's what made me open #326. We'll have to wait until Rubocop adds a rule for that: https://github.com/bbatsov/rubocop/issues/3631

PR #322, opened 2016-10-16
Parse implemented interfaces and union members as TypeName instead of String

cjoudrey:
While working on an utility method to convert an `AST` to `Schema` (like [graphql-js' `buildASTSchema.js`](https://github.com/graphql/graphql-js/blob/345400987c67bc67f57d1d07f637bc9acbfd1016/src/utilities/buildASTSchema.js)), I noticed two spots where we parse referenced types as a string instead of a `TypeName` node:

1. We parse an object type definition's interfaces as strings. In graphql-js [they are parsed as a `NAMED_TYPE` node](https://github.com/graphql/graphql-js/blob/345400987c67bc67f57d1d07f637bc9acbfd1016/src/language/parser.js#L816-L825).

2. We parse union members as strings. In graphql-js [they are parsed as a `NAMED_TYPE` node](https://github.com/graphql/graphql-js/blob/345400987c67bc67f57d1d07f637bc9acbfd1016/src/language/parser.js#L929-L935).

On the surface this might not seem problematic, but [it causes code like this](https://github.com/rmosolgo/graphql-ruby/blob/28998fec1a02f79ab2de7be88e67ca17fc2cb490/lib/graphql/schema/type_expression.rb) to break as `ast_node` would be a string instead of a `TypeName` node.

This is a breaking change for anyone relying on `GraphQL::Language::Nodes`.

:eyes: @rmosolgo

### DIFF
--- a/lib/graphql/language/generation.rb
+++ b/lib/graphql/language/generation.rb
@@ -83,7 +83,7 @@ module GraphQL
         when Nodes::ObjectTypeDefinition
           out = "type #{node.name}"
           out << generate_directives(node.directives)
-          out << " implements " << node.interfaces.join(", ") unless node.interfaces.empty?
+          out << " implements " << node.interfaces.map(&:name).join(", ") unless node.interfaces.empty?
           out << generate_field_definitions(node.fields)
         when Nodes::InputValueDefinition
           out = "#{node.name}: #{generate(node.type)}"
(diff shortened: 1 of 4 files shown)


rmosolgo:
:+1: who could this possibly break? I guess `GraphQL.parse(schema_string)` returns a different result, but I don't know of anyone who's using that directly (or any reason to use it directly), do you?

(I think fragment conditions have this same shortcoming, parsing types as strings instead of `TypeName`s, but changing that will require a change in `graphql-libgraphqlparser` too)

cjoudrey:
Correct, I think it would only break for someone using GraphQL.parse
directly.

Good catch for fragment conditions, we might want to fix that in a separate
pull request so that we can coordinate the change in
graphql-libgraphqlparser.

PR #326, opened 2016-10-17
Add Rubocop to CI

cjoudrey:
We currently have `.rubocop.yml` but it is not enforced by CI. This means we will end up having code violations in `master` unless every person thinks about running `rubocop` locally, which isn't realistic.

This PR will make CI fail when Rubocop finds bad code.

I also made a few adjustments to our `.robucop.yml` file. We can discuss each change in the diff.

I also ran `rubocop` and corrected offenses in the codebase.

:eyes: @rmosolgo @astorije

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,11 +5,6 @@ AllCops:
   - 'lib/graphql/language/lexer.rb'
   - 'lib/graphql/language/parser.rb'
 
-# value = do_something
-# if value # ...
-Lint/AssignmentInCondition:
-  AllowSafeAssignment: false
-
 # def ...
 # end
 Lint/DefEndAlignment:
@@ -20,10 +15,6 @@ Lint/DefEndAlignment:
 # end
 Lint/EndAlignment:
   AlignWith: variable
-
-Metrics/ClassLength:
-  CountComments: false
-  Max: 120
 
 Metrics/ParameterLists:
   Max: 6
@@ -47,23 +38,5 @@ Style/LeadingCommentSpace:
 Style/MethodName:
   EnforcedStyle: snake_case
 
-Style/StringLiterals:
-  EnforcedStyle: double_quotes
-
 Style/WordArray:
   EnforcedStyle: brackets
-
-Style/StringLiterals:
-  Enabled: false
-
-Metrics/ClassLength:
-  Enabled: false
-
-Metrics/ModuleLength:
-  Enabled: false
-
-Metrics/MethodLength:
-  Enabled: false
-
-Lint/AssignmentInCondition:
-  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,9 +32,6 @@ Metrics/ParameterLists:
 Style/ClassAndModuleChildren:
   EnforcedStyle: nested
 
-Style/DeprecatedHashMethods:
-  Enabled: true
-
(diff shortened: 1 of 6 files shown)


.rubocop.yml
```diff
@@ -32,9 +32,6 @@ Metrics/ParameterLists:
 Style/ClassAndModuleChildren:
   EnforcedStyle: nested
 
-Style/DeprecatedHashMethods:
```
cjoudrey: This rule does not exist anymore.

.rubocop.yml
```diff
@@ -55,3 +52,18 @@ Style/StringLiterals:
 
 Style/WordArray:
   EnforcedStyle: brackets
+
+Style/StringLiterals:
```
cjoudrey: This is double-quotes versus single-quotes. If you want to take a side, we can enable it and fix the entire code base.

rmosolgo: :no_good_man: I'm a double-quotes guy but more than that, I'm a don't-hassle-contributors-with-quote-drama guy

cjoudrey:
Same. I just don't think the gains of enabling this rule outweigh the pain.

cjoudrey:
@rmosolgo were you happy with the rest of the changes?

.rubocop.yml
```diff
@@ -55,3 +52,18 @@ Style/StringLiterals:
...
+  Enabled: false
+
+Metrics/ClassLength:
+  Enabled: false
+
+Metrics/ModuleLength:
+  Enabled: false
+
+Metrics/MethodLength:
```
rmosolgo: There's `DisabledByDefault: true` above, is that enough for these `Length` cops? (It's ok with me to remove the `ClassLength` cop higher up in the file, too)

.rubocop.yml
```diff
@@ -55,3 +52,18 @@ Style/StringLiterals:
...
+  Enabled: false
+
+Lint/AssignmentInCondition:
+  Enabled: false
```
rmosolgo: I'm willing to give this one up (we don't pass on it anyways), but should we remove the previous definition from above ?

cjoudrey: @rmosolgo I just realized `Lint/AssignmentInCondition` was already defined above.

We have some code that violates this [like assignment in `while`](https://github.com/rmosolgo/graphql-ruby/blob/2e59be6482ce94a6124ce27766ad4a7910e6c589/lib/graphql/internal_representation/rewrite.rb#L154).

Did you prefer we fix those or disable this rule?

rmosolgo:
I'm no lover of styleguides but I knew this day would come, thanks for getting this going! Sorry I didn't get to it til now, busy day at my _real_ job :P

.rubocop.yml
rmosolgo: My _personal_ opinion is "yes" for `while` and "no" for `if`, but that doesn't really seem fair to push on people, so turning it off is ok for me!

.rubocop.yml
cjoudrey: Not sure how I missed that. You're absolutely right. All these `Enabled: false` aren't necessary.

cjoudrey:
> I'm no lover of styleguides but I knew this day would come, thanks for getting this going!

No problem! 😀 

At the very least it'll help us catch bad indentation.

rmosolgo:
:+1:

Issue #311, opened 2016-10-14
create mutations have begin returning "nil cannot be coerced into fixnum"


Plummat:
happened after I updated today to .19.3, any clues?

rmosolgo:
could you share the stack trace?

rmosolgo:
Also your mutation definition, if you don't mind

Plummat:
this is where it breaks:

```
 Prequalification Load (0.6ms)  SELECT "prequalifications".* FROM "prequalifications" WHERE "prequalifications"."deleted_at" IS NULL AND "prequalifications"."opportunity_id" = $1 OFFSET 0  [["opportunity_id", 88019]]
Completed 500 Internal Server Error in 15297ms (ActiveRecord: 357.3ms)

TypeError (nil can't be coerced into Fixnum):
  app/controllers/queries_controller.rb:8:in `create'
```

mutation: 
```ruby
CreatePrequalificationMutation = GraphQL::Relay::Mutation.define do
  name "CreatePrequalificationMutation"
  description "Creates a New Prequalification"

  input_field :id, !types.ID
  input_field :opportunity_id, !types.ID

  return_field :prequalification, PrequalificationType.edge_type
  return_field :opportunity, OpportunityType
  return_field :product, ProductType

  resolve -> (_object, inputs, _ctx) {
    prequalification = Prequalification.find_or_initialize_by(product_id: inputs[:id], opportunity_id: inputs[:opportunity_id])
    if prequalification.new_record?
      prequalification.save!
    end
    prequalification.move_to_interested
    connection = GraphQL::Relay::RelationConnection.new(prequalification.opportunity.prequalifications, {})
    prequal_edge = GraphQL::Relay::Edge.new(prequalification, connection)
    { prequalification: prequal_edge, opportunity: prequalification.opportunity, product: prequalification.product }
  }
end
```

Plummat:
It also saves just fine, it occurs after the mutation returns

rmosolgo:
Nothing else after this? 

```
TypeError (nil can't be coerced into Fixnum):
  app/controllers/queries_controller.rb:8:in `create'
```

(I'm hoping to see some lines of the gem's stack trace there)

rmosolgo:
What version were you on before?

Plummat:
Thats all I have... was 0.18.15 previosuly

rmosolgo:
I suspect I somehow broke these, I've broken this before (and last time I added a test 😖)

[...]

PR #327, opened 2016-10-18
fix(RelationConnection) raise when the edge isn't found

rmosolgo:
Provide a more descriptive error for cases like #311


Plummat:
thanks for working through that with me!

PR #334, opened 2016-10-19
Don't catch invalid null errors unless they are from an ExecutionError

dylanahsmith:
cc @xuorig

## Problem

https://github.com/rmosolgo/graphql-ruby/pull/161 stopped gobbling errors by default for most exceptions, but invalid null errors are still getting caught, even though a non-null field returning `nil` is an internal error.

## Solution

Internally InvalidNullError is used to handle the error propagation behaviour in http://facebook.github.io/graphql/#sec-Error-handling, which still should be handled when the initial `nil` comes from a GraphQL::ExecutionError.  We keep this behaviour in that case, where  GraphQL::InvalidNullError#parent_error? returns `true` to indicate that it came from an ExecutionError.

When GraphQL::InvalidNullError#parent_error? returns `false` this PR now re-raises the error since it is an internal error, and it should be handled similarly to result coercion errors.


dylanahsmith:
cc @eapache since I think you added this behaviour in https://github.com/rmosolgo/graphql-ruby/pull/94 back when internal errors were being caught by this gem

eapache:
Makes sense to me, was probably just missed in #161.

rmosolgo:
This makes sense -- user code should have an opportunity to respond to this error. I'm tinkering locally with a way to preserve the old behavior in the next release ...

rmosolgo:
Ok, I pushed a commit here to accept a custom handler. That way an endpoint can selectively log, re-raise, or push an error. How does that seem to you?

I've defaulted to re-raising, but we could just as easily default to `ctx.errors << err` and then have no breaking change.

dylanahsmith:
I think having a custom handler just for invalid nulls is too narrowly focused.  I would suspect that the reason to preserve the behaviour of catching and rendering the invalid null error would be the same reason for CatchallMiddleware or another middleware for catching and rendering internal errors.  However, middleware doesn't surround coercion, so there is no way to provide similar behaviour for catching coercion errors in general.

Having a custom handler for any coercion error would solve the general issue as well as providing a way for preserving the current behaviour by checking if the error `.is_a?(GraphQL::InvalidNullError)` and turning it into a GraphQL::ExecutionError.

rmosolgo:
Ah, I think I see what you mean: provide a hook for user code to handle the case of "I expected to get a value like X but all I got was Y, what do you want me to do?"

dylanahsmith:
Looks like this is even worse on master with the new executor, since now this error is [being added to the context as a GraphQL::ExecutionError](https://github.com/rmosolgo/graphql-ruby/blob/66f6bc89991f0de8073c2136afdea6d8f448a2ac/lib/graphql/execution/execute.rb#L125) rather than a GraphQL::InvalidNullError as is done from SerialExecution.  This means we can no longer workaround this issue by checking for non-GraphQL::ExecutionError exceptions in the context errors.

rmosolgo:
😖 Somehow I had the idea you were removing them by _message_, not by class. 

I've been dawdling on this one because, on the one hand, re-raising seems like overkill: we _can_ recover from the situation, so I think we should (otherwise, why would the spec describe how to propagate nulls?). On the other hand, the hook I implemented here was "too narrowly focused", so I started exploring how to integrate with Scalar parse functions, but got bogged down. 

So, I think I will re-implement the hook described here and apply it to two type errors: invalid nulls and unresolved union/types. Then in a later version, we can see if other type errors should be passed to the same handler. 

That way, you can implement that behavior with a public, tested, documented API :)

dylanahsmith:
> I've been dawdling on this one because, on the one hand, re-raising seems like overkill: we can recover from the situation, so I think we should

This seems like the opposite of what you decided in https://github.com/rmosolgo/graphql-ruby/pull/161.  I agree that having something like a CatchallMiddleware for coercion and invalid null errors would be useful, but more important than that is to stop swallowing the error so that the server can notify the server developers of the error.  Classifying the error as an ExecutionError makes it sound like a client error, which it isn't in this case.

> we can recover from the situation, so I think we should (otherwise, why would the spec describe how to propagate nulls?)

Because this null propagation also applies to nulls that originate from a field error (i.e. GraphQL::ExecutionError).  Those are from client errors and allows the client to fallback to using the data from a response with an error without worrying about non-null fields have a null value.

> So, I think I will re-implement the hook described here and apply it to two type errors: invalid nulls and unresolved union/types.

I think it would be easier to have it apply to any error raised in get_finished_value, which would apply to those types as well as coercion errors.  Alternatively, it could be a hook that applies to any internal error, then this error handling wouldn't have to also be done in a middleware.

dylanahsmith:
Pushed a rebase of my commit on this PR https://github.com/rmosolgo/graphql-ruby/compare/master...dylanahsmith:re-raise-invalid-null-error-rebased, since you had pushed a commit to this branch.

PR #335, opened 2016-10-19
Fix indentation when printing args that have comments

cjoudrey:
There was a small indentation bug in `Schema::Printer` when printing arguments that have comments.

Here's an example that displays the bug:

**Currently**

```graphql
type Query {
  post(
    # Post ID
    id: ID!
    varied: Varied = {id: \"123\", int: 234, float: 2.3, enum: FOO, sub: [{string: \"str\"}]}
): Post
}
```

**With this PR**

```graphql
type Query {
  post(
    # Post ID
    id: ID!
    varied: Varied = {id: \"123\", int: 234, float: 2.3, enum: FOO, sub: [{string: \"str\"}]}
  ): Post
}
```

:eyes: @rmosolgo


rmosolgo:
nice catch!

PR #340, opened 2016-10-21
Update mutation resolve() signature

holman:
Two arg signatures are deprecated; this little bit in the docs just was missed in all the updates lately, looks like.


rmosolgo:
Thanks! Those relay docs were taken from an old branch, oops!

holman:
Figured as much. :) Liking the expansion of the docs, though!

Issue #302, opened 2016-10-11
Array connection type is used for AR relations in Rails 3.2


spalladino:
Given a model with a `connection` on a `has_many` association, for instance:

```ruby
class Worker < ActiveRecord::Base
  has_many :worked_hours, dependent: :destroy
end

WorkerType = GraphQL::ObjectType.define do
  name "Worker"
  field :id, !types.Int, "The id of this worker"
  connection :entries, EntryType.connection_type, "Worked hours", property: :worked_hours
end

EntryType = GraphQL::ObjectType.define do
  name "Entry"
  field :id, !types.Int, "The id of this entry"
  field :record, !types.String, "The record of this entry"
end
```

When running a query on the connection, the ArrayConnection subclass is used, instead of the RelationConnection. This causes any `limit`s to be issued in memory instead of in the SQL query.

This seems to be happening because the resolver for the Connection subtype relies on the `property`'s class, and Rails 3.2 reports a `has_many` association to be of class `Array`, whereas Rails 4+ returns `AR::Relation`.

```
Worker.first.worked_hours.class
=> Array
```

rmosolgo:
Thanks for sharing the details on this! I'll take a look at that connection-matching issue soon.

rmosolgo:
:metal:  fixed by #343

PR #343, opened 2016-10-21
Resolve connection class by most specific class type

jsdnxx:
This PR updates the way GraphQL::Relay::BaseConnection.connection_for_nodes resolves Connection class implementations from collection types.

The current algorithm is depth-first, so custom Nodes collection classes which inherit from Array will resolve to ArrayConnection even if there is a more specific subclass of BaseConnection registered.

With this PR, it changes to breadth-first to search all registered CONNECTION_IMPLEMENTATIONS for ancestors of the nodes class, starting with the most specific.


rmosolgo:
Nice! I'm hoping this will fix #302 also. Thank you!

PR #345, opened 2016-10-22
Add failing test for relation connection on Rails 3

### LINKED ISSUE #302: Array connection type is used for AR relations in Rails 3.2

rmosolgo:
Trying to fix #302 ... first step, test failure :)

### DIFF
--- a/spec/graphql/relay/relation_connection_spec.rb
+++ b/spec/graphql/relay/relation_connection_spec.rb
@@ -361,4 +361,11 @@ describe GraphQL::Relay::RelationConnection do
       assert_includes err.message, "item not found"
     end
   end
+
+  it "is chosen for a relation" do
+    relation = Base.where(faction_id: 1)
+    assert relation.is_a?(ActiveRecord::Relation)
+    connection = GraphQL::Relay::BaseConnection.connection_for_nodes(relation)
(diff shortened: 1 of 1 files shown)


rmosolgo:
🎉  fixed by #343

PR #346, opened 2016-10-22
Add Schema.from_definition

cjoudrey:
This PR adds `Schema.from_definition` which is similar to `Schema::Loader`, but is used to build a schema from a parsed IDL.

My primary use for this class to have saner tests when different schemas are required for each test. For instance, here's a snippet from the test suite of a `SchemaComparator` PR I'm working on:

```ruby
it "detects if a type was removed or added" do
  schema1 = <<-SCHEMA
    type Type1 {
      field1: String
    }
 
    type Type2 {
      field1: String
    }
  SCHEMA

  schema2 = <<-SCHEMA
    type Type2 {
      field1: String
    }
  SCHEMA
 
  changes = GraphQL::Schema::SchemaComparator.compare(schema(schema1), schema(schema2))

  # ...
```

Secondary use-case is basically everything else you might currently use `Schema::Loader` for, but want to use the IDL instead of the introspection result.

Some discussion points:

  - ~~Should we expose these features as `Schema#build_from_ast` and `Schema#build_from_introspection_schema` instead of `Schema::Loader.load` and `Schema::BuildFromAST.load`?~~
  - We might want to eventually accept a 2nd argument to `load` which would be a hash where the `key` could be a field name and `value` would be a resolver proc for that field.
  - Sangria also moves the _build_ methods into a builder class and `load` can accept a custom builder. I can't think of a use-case for this right now so I think we can add it later.
  - Tests are currently red because of two `rescue`: [1](https://github.com/rmosolgo/graphql-ruby/blob/c78d24bb1f2fbfa5b0e486c4504d2ecc334cb700/lib/graphql/object_type.rb#L50) and [2](https://github.com/rmosolgo/graphql-ruby/blob/fd90fcccacd991a7c202d61b630f9bccc1414412/lib/graphql/union_type.rb#L45). Discussing this with @rmosolgo over Slack. I believe we shouldn't be swallowing all exceptions there but rather those at define time.

:eyes: @rmosolgo


lib/graphql/schema/build_from_ast.rb
```diff
@@ -0,0 +1,260 @@
...
+          when "Int"
+            INT_TYPE
+          when "String"
+            STRING_TYPE
+          when "Float"
+            FLOAT_TYPE
+          when "Boolean"
+            BOOLEAN_TYPE
+          when "ID"
+            ID_TYPE
```
cjoudrey: I'm not super happy with this mapping. `Schema::Loader` has it too. I wonder if we can create a `BUILT_IN_TYPES` hash somewhere that both classes can use.

lib/graphql/schema/build_from_ast.rb
```diff
@@ -0,0 +1,260 @@
...
+              description: field_definition.description,
+              type: type_resolver.call(field_definition.type),
+              arguments: field_arguments,
+              deprecation_reason: build_deprecation_reason(field_definition.directives),
+            )
+          ]
+        end
+      end
+
+      def resolve_type(types, ast_node)
```
cjoudrey: I was considering refactoring [`Schema::TypeExpression.build_type`](https://github.com/rmosolgo/graphql-ruby/blob/c78d24bb1f2fbfa5b0e486c4504d2ecc334cb700/lib/graphql/schema/type_expression.rb) to accept a hash of `types` instead of `schema`.

This would allow us to :fire: this entire method.

lib/graphql/schema/build_from_ast.rb
```diff
@@ -0,0 +1,260 @@
...
+      # Create schema with the AST of a parsed schema.
+      # @param document [GraphQL::Language::Nodes::Document] A Document node obtained from {GraphQL.parse}
+      # @return [GraphQL::Schema] the schema described by `document`
+      def build(document)
+        raise InvalidDocument.new('Must provide a document ast.') if !document || !document.is_a?(GraphQL::Language::Nodes::Document)
+
+        schema_definition = nil
+        types = {}
+        directives = {}
+        type_resolver = -> (type) { -> { resolve_type(types, type) } }
```
cjoudrey: Copied this logic from `Schema::Loader`.

The alternative would be to iterate over `document.definitions` twice. Once to initialize `types` with empty definitions that way when you iterate a second time you can reference objects instead of passing a proc that resolves later.

lib/graphql/schema/build_from_ast.rb
rmosolgo: Would it work to preload the `types` hash with the built-in scalars, so they didn't need "special" treatment?

lib/graphql/schema/build_from_ast.rb
```diff
@@ -0,0 +1,260 @@
+module GraphQL
+  class Schema
+    # You can use the result of {GraphQL.parse} to make a schema.
+    # This schema is missing some important details like
+    # `resolve` functions, but it does include the full type system,
+    # so you can use it to validate queries.
+    module BuildFromAST
+      extend self
+
+      class InvalidDocument < Error; end;
```
rmosolgo: Can we end this classname with `Error`? that would make it consistent with the rest of the errors in the library.

lib/graphql/schema/build_from_ast.rb
```diff
@@ -0,0 +1,260 @@
...
+            raise InvalidDocument.new("Specified mutation type \"#{schema_definition.mutation}\" not found in document.") unless types[schema_definition.mutation]
+            mutation_root_type = types[schema_definition.mutation]
+          end
+
+          if schema_definition.subscription
+            raise InvalidDocument.new("Specified subscription type \"#{schema_definition.subscription}\" not found in document.") unless types[schema_definition.subscription]
+            subscription_root_type = types[schema_definition.subscription]
+          end
+        else
+          query_root_type = types['Query'] if types['Query']
```
rmosolgo: What do we gain by these `if types[...]` suffixes? would it be the same to "just" get `nil` by a hash lookup miss, eg 

```ruby 
# query_root_type = types['Query'] if types['Query']
# is that equivalent to?
query_root_type = types['Query'] 
```

rmosolgo:
:+1: Love the idea. It'd be nice if there was some way to unify this & `Loader` as "front-ends" to the schema builder, but I don't know how much its possible to reuse functionality between them. 

:+1: For methods on `Schema`, it would be nice to remove direct access to these internals from user code. I don't think the method names need the word `build_` in them, what about just `Schema.from_definition(definition_string)` and `Schema.from_introspection(introspection_result)` ?

lib/graphql/schema/build_from_ast.rb
cjoudrey: Hm, yeah, not sure what I was thinking here. lol

lib/graphql/schema/build_from_ast.rb
cjoudrey: Do you mean add all built-in types to the `types` hash above such that we can :fire: this extra logic? If so, yeah that's what I had in mind, but that's why I was wondering if we could have a `BUILT_IN_TYPES` constant hash somewhere so that we can do something like:

```ruby
types = {}
types.merge(GraphQL::Language::BUILT_IN_TYPES)
# ...
```

And :fire: this code.

lib/graphql/object_type.rb
```diff
@@ -46,9 +46,11 @@ def interfaces=(new_interfaces)
     def interfaces
       @clean_interfaces ||= begin
         ensure_defined
-        @dirty_interfaces.map { |i_type| GraphQL::BaseType.resolve_related_type(i_type) }
-      rescue
-        @dirty_interfaces
+        if @dirty_interfaces.respond_to?(:map)
```
cjoudrey: @rmosolgo I'm not :100: happy with this, but can't think of a better way. We seem to want to let it _pass_ here such that we can pick up the issue when validating the schema in `Schema::Validation`. Alternatively I was going to `raise ArgumentError unless Array` in `interfaces=` but this would mean changing this test and behaviour: https://github.com/rmosolgo/graphql-ruby/blob/c78d24bb1f2fbfa5b0e486c4504d2ecc334cb700/spec/graphql/schema/validation_spec.rb#L83-L88

Any ideas?

cjoudrey:
>  For methods on Schema, it would be nice to remove direct access to these internals from user code. I don't think the method names need the word build_ in them, what about just Schema.from_definition(definition_string) and Schema.from_introspection(introspection_result) ?

Love the idea, I'll do that.

Should we deprecate `GraphQL::Loader` first or is it okay to :fire: it right away?

brandonblack:
@rmosolgo this sounds like another use case for multiple schemas with shared definitions. Rather than another loader, could this be resolved with your schema mask proposal?

rmosolgo:
I think the main use case here is a bit different. It's part of the plumbing for answering the question "have I made breaking changes to the schema?" 

It's answered this way: 

- Keep a dump of the schema definition as a string in source control
- Whenever you change the schema, update the definition dump 
- Then, have a utility who can compare the _new_ definition against the _old_ definition and report any breaking changes  

So the `.from_definition`'d schema would never be used for resolution -- just for a structural comparison to another schema (which was also `from_definition`'d). 

@cjoudrey  did I get that right? 😬

cjoudrey:
> @cjoudrey did I get that right? 😬

That's correct!

I fail to see how this is related to schema masking. @brandonblack can you elaborate on that? Maybe I created confusion when I said `accept a 2nd argument that could contain resolvers`.

`Schema.from_definition` is essentially just a tool for people who want to do stuff with their IDL like schema comparison.

rmosolgo:
:+1: Sorry, could you give 'er one more rebase?

cjoudrey:
:+1: Rebased!

Btw, I'm not super familiar with yard, but I think I did it right.

rmosolgo:
:rocket: !

PR #355, opened 2016-10-26
nameless LIST and NON_NULL

### LINKED ISSUE #352: LIST and NON_NULL types should be nameless
The GraphQL spec requires that in an introspection response, types with a kind of "LIST" or "NON_NULL" must only have `kind` and `ofType` values present; all other properties must have null values:

https://facebook.github.io/graphql/#sec-List
https://facebook.github.io/graphql/#sec-Type-Kinds.Non-Null

Right now this library provides the names "List" and "Non-Null" for these types, respectively:

https://github.com/rmosolgo/graphql-ruby/blob/e9cc6c5790eb492bc288ed77541c1a29f94aa0c5/lib/graphql/list_type.rb#L28-L30
https://github.com/rmosolgo/graphql-ruby/blob/e9cc6c5790eb492bc288ed77541c1a29f94aa0c5/lib/graphql/non_null_type.rb#L38-L40

These names are propagated into introspection responses, which is invalid according to the spec. Instead, these types should always have a name of `null`. I propose that the `ListType` and `NonNullType` classes should return `nil` for their `.name`, which would then propagate valid `null` values to introspection responses.

jamesmacaulay:
Fixes #352.

This modifies both `ListType#name` and `NonNullType#name` to return `nil`, so that introspection responses correctly return `null` names for these types as per the GraphQL spec. This is a breaking change, both because it modifies the introspection responses, and because downstream code may be relying on the `name` method on types always returning a string.

### DIFF
--- a/lib/graphql/list_type.rb
+++ b/lib/graphql/list_type.rb
@@ -25,9 +25,8 @@ module GraphQL
   #
   class ListType < GraphQL::BaseType
     include GraphQL::BaseType::ModifiesAnotherType
-    attr_reader :of_type, :name
+    attr_reader :of_type
     def initialize(of_type:)
-      @name = "List"
       @of_type = of_type
     end
 
--- a/lib/graphql/non_null_type.rb
+++ b/lib/graphql/non_null_type.rb
@@ -35,10 +35,6 @@ module GraphQL
       @of_type = of_type
     end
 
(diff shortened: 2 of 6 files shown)


rmosolgo:
👍  thanks for this fix!!

PR #358, opened 2016-10-27
Change documentation enum_type to include ActiveRecord::Base

zyixc:
Add documentation how to use enum types defined with ActiveRecord::Enum.


rmosolgo:
That's great, thank you!

PR #360, opened 2016-10-27
Allow passing arrays as literals for scalar types

jcmfernandes:
From my understanding, the GraphQL spec doesn't dictate that arrays cannot be
used as literals for scalar types, therefore I'm removing this check.

I hit this while defining a scalar to represent two-dimensional coordinates
using an array with two values.

### DIFF
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -12,7 +12,7 @@ module GraphQL
         elsif type.kind.list?
           item_type = type.of_type
           ensure_array(ast_value).all? { |val| validate(val, item_type) }
-        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::AbstractNode) && !ast_value.is_a?(Array)
+        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::AbstractNode)
           type.valid_input?(ast_value, @warden)
         elsif type.kind.enum? && ast_value.is_a?(GraphQL::Language::Nodes::Enum)
           type.valid_input?(ast_value.name, @warden)
--- a/spec/graphql/scalar_type_spec.rb
+++ b/spec/graphql/scalar_type_spec.rb
@@ -22,6 +22,45 @@ describe GraphQL::ScalarType do
     assert_equal(bignum.to_s, custom_scalar.coerce_result(bignum))
   end
 
+  describe "when passing literals for scalar types in input objects" do
+    let(:scalar_type) {
+      GraphQL::ScalarType.define do
+        name "ArrayScalar"
+        coerce ->(value) { value if value.is_a?(Array) }
+      end
+    }
+    let(:dummy_mutation) {
+      scalar_type = scalar_type()
+      GraphQL::Relay::Mutation.define do
+        name "DummyMutation"
+        input_field :input_value, !scalar_type
+        return_field :output_value, !scalar_type
+        resolve -> (obj, inputs, ctx) { {output_value: inputs[:input_value]} }
+      end
+    }
+    let(:root_object) {
+      dummy_mutation = dummy_mutation()
+      GraphQL::ObjectType.define do
+        name "Mutation"
+        field :field, field: dummy_mutation.field
+      end
+    }
+    let(:schema) {
+      GraphQL::Schema.define(mutation: root_object)
+    }
+    let(:query_string) {%|
+      mutation M { field(input: {input_value: [1, 2]}) { output_value } }
+    |}
+    let(:result) {
+      schema.execute(query_string)
+    }
+
+    it "correctly validates arrays" do
+      expected = {"data" => {"field" => {"output_value" => [1, 2]}}}
+      assert_equal(expected, result)
+    end
+  end
+
   describe "custom scalar errors" do
     let(:result) { custom_scalar.validate_input("xyz", PermissiveWarden) }
 


rmosolgo:
To continue to support this behavior, we should have some spec for it. For example, in `scalar_type_spec.rb`, we could have a spec where a small schema is defined with a custom scalar which comes in as an array. Then it could run the query and assert that the custom scalar was parsed & handled correctly. Are you interested in adding that? 

If you're feeling really enterprising, you might also be able to support custom scalars coming in as objects too. I suspect this would require more changes to that `case` statement (and another spec, of course :P).

jcmfernandes:
Haha, challenge accepted for the first part, will work on that tomorrow 😉 let me assess the changes needed for the second and I'll get back to you!

jcmfernandes:
I need to do a brain dump. Beware, ramblings below.

I hit this when I introduced mutations in the mix. Queries always worked fine, that is, coercions where being respected. Example:

```ruby
ArrayScalarType = GraphQL::ScalarType.define do
  name "ArrayScalar"
  coerce ->(value) { value if value.is_a?(Array) }
end

RootObject = GraphQL::ObjectType.define do
  name "QueryType"
  field :field, !ArrayScalarType do
    argument :value, !ArrayScalarType
    resolve -> (obj, args, ctx) { args['value'] }
  end
end

Schema = GraphQL::Schema.define(query: RootObject)

Schema.execute('query Q($value: ArrayScalar!) { field(value: $value) }',
               variables: {'value' => [1, 2]})
=> {"data"=>{"field"=>[1, 2]}}
```

All good! Now, as stated, mutations didn't:

```ruby
ArrayScalarType = GraphQL::ScalarType.define do
  name "ArrayScalar"
  coerce ->(value) { value if value.is_a?(Array) }
end

DummyMutation = GraphQL::Relay::Mutation.define do
  name "DummyMutation"
  input_field :input_value, !ArrayScalarType
  return_field :output_value, !ArrayScalarType
  resolve -> (obj, inputs, ctx) { {output_value: inputs[:input_value]} }
end

RootObject = GraphQL::ObjectType.define do
  name "Mutation"
  field :field, field: DummyMutation.field
end

Schema = GraphQL::Schema.define(mutation: RootObject)

Schema.execute('mutation M { field(input: {input_value: [1, 2]}) { output_value } }')
=> {"errors"=>
  [{"message"=>
     "Argument 'input' on Field 'field' has an invalid value. Expected type 'DummyMutationInput!'.",
    "locations"=>[{"line"=>2, "column"=>20}],
    "fields"=>["mutation M", "field", "input"]},
   {"message"=>
     "Argument 'input_value' on InputObject 'DummyMutationInput' has an invalid value. Expected type 'ArrayScalar!'.",
    "locations"=>[{"line"=>2, "column"=>33}],
    "fields"=>["mutation M", "field", "input", "input_value"]}]}
```

However, when using variables it does work!

```ruby
ArrayScalarType = GraphQL::ScalarType.define do
  name "ArrayScalar"
  coerce ->(value) { value if value.is_a?(Array) }
end

DummyMutation = GraphQL::Relay::Mutation.define do
  name "DummyMutation"
  input_field :input_value, !ArrayScalarType
  return_field :output_value, !ArrayScalarType
  resolve -> (obj, inputs, ctx) { {output_value: inputs[:input_value]} }
end

RootObject = GraphQL::ObjectType.define do
  name "Mutation"
  field :field, field: DummyMutation.field
end

Schema = GraphQL::Schema.define(mutation: RootObject)

Schema.execute('mutation M($value: ArrayScalar!) { field(input: {input_value: $value}) { output_value } }', 
               variables: {'value' => [1, 2]})
=> {"data"=>{"field"=>{"output_value"=>[1, 2]}}}
```

I believe this happens because when using variables, method `GraphQL::StaticValidation::LiteralValidator#validate` tests

```ruby
...
elsif ast_value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
  true
else
...
```

which is indeed true. It seems to me that it assumes that the values of variables are being validated elsewhere.

jcmfernandes:
@rmosolgo I amended my commit and added the requested spec. Feedback is appreciated 😉

theorygeek:
So regarding the problem around literals vs variables, I'm not sure that that's going to be easily solved. The spec itself defines `[` and `]` as [punctuators](https://facebook.github.io/graphql/#sec-Punctuators), and only uses them for [list values](https://facebook.github.io/graphql/#sec-List-Value).

If you were going to embed a custom scalar into the text of a GraphQL query itself, I think the language expects that [it'd be represented as a string](https://facebook.github.io/graphql/#sec-Scalars):
>All GraphQL scalars are representable as strings, though depending on the response format being used, there may be a more appropriate primitive for the given scalar type, and server should use those types when appropriate.

The same scenario arises, for example, if you were going to have a custom scalar that represents arbitrary JSON. When you pass information as _variables_, you can pass along any Ruby object; but with literals within the query text, you're much more limited.

For what it's worth, I think the spec _does_ exclude the possibility of a scalar being something like an array or an object, by virtue of the fact that they're called "scalars." The word scalar means:
>representable by position on a scale or line; having only magnitude

You can't really represent an array of values on a line; you really need one line _per dimension_ to represent it. And in the CS world, the word "scalar" is nearly always used in contrast with "vector," with is synonymous with "array."

But for your present scenario, couldn't you just use a list of a list (ie, `types[types[types.Int]]`)?

jcmfernandes:
This isn't an easy discussion.

I totally get your arguments. With that said, IMO, you're focusing on the meaning of "scalar" as in the physics world. Are strings scalars? Perl says so. Are booleans scalars? They don't convey scale, `true` isn't greater than `false` nor vice-versa.

Ultimately, I can't ignore the etymology of the word, so I'll leave the decision to @rmosolgo 😹 .

jcmfernandes:
Oh, and regardless of the conclusion, the behavior I described is definitely inconsistent.

rmosolgo:
I really see the appeal here. It reminds me of Crystal's type system where you can make aliases of built in data types. You get some implementation wiggle room where the type name can stay the same even when the underlying structure of its members changes.

But I'm not ready to go forward for a few reasons: 

- I don't think other servers support this practice, and I don't have the nerve to test it out. (I could be wrong, and that would make me more open to a change like this) 
- I don't _know_ if clients support this well. (Does someone else know?) I think it's important for graphql-ruby's features to work well with any client. 
- The workaround is sufficient. For example, couldn't a two-dimensional point be modeled with an input type like `{x:, y:}`? This has the advantage of having named attributes, too. 

If you disagree on those points (that the practice isn't widespread, and that the workaround is sufficient), please help me understand the other side!

PR #361, opened 2016-10-28
Non-nullable arguments cannot have default values

cjoudrey:
While working on schema comparisons I realized we currently allow non-nullable types to have default values.

According to [`graphql-js`](https://github.com/graphql/graphql-js/blob/89f3f1a319bb913616f2910e8a806170155ffaf0/src/validation/rules/DefaultValuesOfCorrectType.js#L19) this is not valid.

This PR adds the same validation to `graphql-ruby`.

I'm not sure if this will change in the future due to facebook/graphql#83 but in the meantime it's probably best that we have this.

:eyes: @rmosolgo /cc @theorygeek

### DIFF
--- a/spec/graphql/schema/validation_spec.rb
+++ b/spec/graphql/schema/validation_spec.rb
@@ -200,16 +200,20 @@ describe GraphQL::Schema::Validation do
       end
     }
 
-    let(:invalid_default_argument) {
+    let(:invalid_default_argument_for_non_null_field) {
       GraphQL::Argument.define do
         name "InvalidDefault"
-        type GraphQL::INT_TYPE
-        default_value "abc"
+        type !GraphQL::INT_TYPE
+        default_value 1
       end
     }
 
     it "requires the type is a Base type" do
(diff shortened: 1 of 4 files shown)


spec/graphql/schema/validation_spec.rb
```diff
@@ -200,16 +200,20 @@ def assert_error_includes(object, error_substring)
       end
     }
 
-    let(:invalid_default_argument) {
```
cjoudrey: We forgot to :fire: this back in #267 so removing it now. 😀

rmosolgo:
This is good by me: 

> If an argument is required, it can't have a default value. 

FWIW, the analogous graphql-ruby code to the graphql-js link is here, the same logic applied to variables in a query: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/rules/variable_default_values_are_correctly_typed.rb#L17

I suppose this is _technically_ a breaking change, but it's a very easy upgrade path and the app just won't start if any code is affected. 

:+1: !!

PR #362, opened 2016-10-28
Fixes #356

### LINKED ISSUE #356: Relation connection with `last` but without `before` yields wrong result
When querying connection field query with arguments `first: 2` yields the same result as `last: 2`.

sjchmiela:


rmosolgo:
Thanks for taking a shot at this bug! I'm concerned about the unbounded count issue, described previously here: 

https://github.com/rmosolgo/graphql-relay-ruby/issues/1#issuecomment-143030081

Is there any way to do this without causing a (potentially huge) `COUNT` query?

sjchmiela:
By the way… isn't https://github.com/sjchmiela/graphql-ruby/blob/b0453b6a9b995ff1c98b3ecb1e5be70660703a75/lib/graphql/relay/array_connection.rb#L34 loading whole relation content into memory for `starting_offset = 0` (the case where only `last` is defined)?

It that's so, we could fix the problem in `paged_nodes` method, and instead of always grabbing the `nodes.first(limit)` we could in that case use `nodes.last(limit)`.

rmosolgo:
`ArrayConnection` always has an `Array` as `@nodes`, so it's _already_ in memory!

sjchmiela:
Oh right, I was looking at the wrong file, sorry!

rmosolgo:
Sorry, I can't accept this as-is because of the `count` issue, feel free to reopen this or another PR if you find another approach! I hope I can take a look soon!

sjchmiela:
That's ok, fix for ArrayConnection would be easy, but there is no way to solve this problem for RelationConnection not using `count`.

I personally would prefer slow response than incorrect one, but I understand your decision. 😊

rmosolgo:
Ok, it's too bad to add an unbounded count here. But, apparently not many people take this codepath, since this bug didn't show up for quite a long time! 

I think you're right that the proper behavior is better than a bug here. I have an idea for supporting heavy-duty use cases that I'll try to prepare by next week. 

Thanks for tracking this down!

PR #366, opened 2016-10-30
GraphQL::Compatibility

rmosolgo:
Some parts of GraphQL are (theoretically?) interchangeable. This adds some tests for external users who want to replace some functionality: `GraphQL::Compatibility::{Execution,QueryParser,SchemaParser}Specification`. 

Personally I want this for libgraphqlparser now, it will show us clearly how the Ruby parser differs from libgraphqlparser, here's the PR for that:  https://github.com/rmosolgo/graphql-libgraphqlparser-ruby/pull/20

@staugaard thanks for getting this started with `ParserTests`, I've moved them around a bit but I think I covered all the bases! Some of the behaviors don't match libgraphqlparser so I put them in the gem-specific test file.


rmosolgo:
This also introduces a breaking change ("breaking fix"?) to how we parse string literals. Previously, we wrongly took the sequence `'\' 'u' \d \d \d \d` and converted it to unicode, now we don't. That is, only `"\u"` is treated as the unicode escape character, not `"\\u"`. (I noticed this issue when testing against libgraphqlparser, then re-read the spec. Oops!)

staugaard:
Nice thanks..

rmosolgo:
K, that unicode thing was wrong I think, I will put it back: 

https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/lexer-test.js#L166-L173

PR #367, opened 2016-10-31
Improve API for returning value and error

josh:
Returning an `ExecutionError` from a resolver automatically annotates the error object with the current ast node and path. Super handy.

However, this only works with nullable fields. If you want to return a sensible value on a non-nullable field you must annotate the ast node and path by hand.

This adds `Context#add_error` to automatically annotate resolver errors.

``` ruby
resolve -> (repo, args, ctx) {
  ctx.add_error GraphQL::ExecutionError.new("Repository stars could not be fetched")
  # just return a zero star count to clients
  return 0
}
```

An alternative API might be allowing another special error type that would hold a value and error.

``` ruby
resolve -> (repo, args, ctx) {
  return GraphQL::ExecutionErrorWithValue.new(value: 0, error: "Repository stars could not be fetched")
}
```

<hr>

To: @rmosolgo 
CC: @mcolyer

### DIFF
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -24,17 +24,18 @@ module GraphQL
         # After getting the value from the field's resolve method,
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
-          execution_context.query.context.irep_node = irep_node
-
           case raw_value
           when GraphQL::ExecutionError
-            execution_context.query.context.add_error(raw_value)
+            raw_value.ast_node = irep_node.ast_node
+            raw_value.path = irep_node.path
+            execution_context.add_error(raw_value)
           when Array
             list_errors = raw_value.each_with_index.select { |value, _| value.is_a?(GraphQL::ExecutionError) }
             if list_errors.any?
               list_errors.each do |error, index|
+                error.ast_node = irep_node.ast_node
                 error.path = irep_node.path + [index]
-                execution_context.query.context.add_error(error)
+                execution_context.add_error(error)
               end
             end
           end
--- a/lib/graphql/query/serial_execution/field_resolution.rb
+++ b/lib/graphql/query/serial_execution/field_resolution.rb
@@ -24,18 +24,17 @@ module GraphQL
         # After getting the value from the field's resolve method,
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
+          execution_context.query.context.irep_node = irep_node
+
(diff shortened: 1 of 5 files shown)


rmosolgo:
:+1: can we remove the old assignment of `irep_node=` & `path=` and use this API instead?

josh:
> 👍 can we remove the old assignment of irep_node= & path= and use this API instead?

Yeah, I had that same thought, but it looks like `FieldResolution` has a reference to an `ExecutionContext` but not a regular `Context`.

https://github.com/rmosolgo/graphql-ruby/blob/9d4b6c1f10589de7206b7b41548ed0f344a91db4/lib/graphql/query/serial_execution/field_resolution.rb#L29-L31

Any ideas?

rmosolgo:
ExecutionContext is stupid, I regret everything

```ruby
execution_context.query.context.add_error(err)
```

(IRL I wish I could move a lot of APIs to `Query`, some from Schema, others from these various context objects, not all of them could be moved but lots of them could)

rmosolgo:
Or, refactor and use [`ExecutionContext#add_error`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/execution_context.rb#L31-L33)

or remove it. Either way 😩

rmosolgo:
Hmm ... that didn't magically become beautiful, did it ...

josh:
I still like this API, but realizing it will suffer from this async bug https://github.com/rmosolgo/graphql-ruby/issues/368

lib/graphql/query/serial_execution/field_resolution.rb
```diff
@@ -24,18 +24,17 @@ def result
         # After getting the value from the field's resolve method,
         # continue by "finishing" the value, eg. executing sub-fields or coercing values
         def get_finished_value(raw_value)
+          execution_context.query.context.irep_node = irep_node
```
rmosolgo: Yeah, the other thing that gets me is the extra, possibly-needless assignment on a hot path. It's not cleaner code but maybe it was better the old way for that reason :S

rmosolgo:
That bug is a real bummer. I'm happy to merge this anyways (with a fix for that extra assignment) if you still want it. 

It's definitely time for a re-think of execution for built-in batch support and for that same bug.  Next release ...

josh:
> I'm happy to merge this anyways (with a fix for that extra assignment) if you still want it.

Yeah, sounds good. Should I just revert 27cead57fffb1433041971d9302868db7f17feab?

rmosolgo:
yep revert sounds good to me

josh:
> yep revert sounds good to me

cool, done!

PR #369, opened 2016-11-01
Language support for NullValue

### LINKED ISSUE #364: Missing Null Value support
The GraphQL spec specifies `null` as a possible argument value (https://facebook.github.io/graphql/#sec-Null-Value). But when I pass null into a mutation like so:

```
mutation {
  user_update(id: "1", address: null) {
    id
    address
  }
}
```

I receive this error: `GraphQL::ParseError - Parse error on "null" (error)`

cjoudrey:
Add support for [null value](https://github.com/facebook/graphql/pull/83) and fixes #364.

#### Example of `null` in field argument

```ruby
GraphQL::ObjectType.define do
  name 'Customer'

  field :fullName, types.String do
    argument :something, types.String

    resolve ->(obj, args, ctx) do
      if args.key?('something')
        something = args['something'] # can be `nil` or a string
      end

      # ...
    end
  end
end
```

```graphql
{
  customer {
    fullName(something: null)
  }
}
```

#### Example of `null` in input object

```ruby
GraphQL::Relay::Mutation.define do
  name 'UpdateCustomer'

  input_field :id, !types.ID
  input_field :fullName, types.String

  return_field :customer, !customer_type

  resolve ->(obj, args, ctx) do
    if args.key?('fullName')
      fullName = args['fullName'] # can be `nil` or a string
    end

    # ...
  end
end
```

```graphql
mutation {
  updateCustomer(input: { id: "123", fullName: null }) {
    customer {
      fullName
    }
  }
}
```


spec/support/static_validation_helpers.rb
```diff
@@ -12,7 +13,8 @@
 module StaticValidationHelpers
   def errors
     target_schema = schema
-    validator = GraphQL::StaticValidation::Validator.new(schema: target_schema)
+    target_rules = defined?(rules) ? rules : GraphQL::StaticValidation::ALL_RULES
```
cjoudrey: While working on validation rule tests, I noticed we run validate against `ALL_RULES`.

I think it might be cleaner if each validation rule spec tested only against its own rules that way we could have stricter assertions.

@rmosolgo do you see this being a problem? If not, we could add `let(:rules) { .. }` to every spec and :fire: the `defined?(rules)` here.

rmosolgo: There have been so many bugs where one validator messed with another, that's why I recently switched these to run against all rules 😞

rmosolgo: The real ickiness comes from some validators assuming that previous ones will return `Visitor::SKIP`, otherwise they hit a Ruby error or start some infinite loop (eg https://github.com/rmosolgo/graphql-ruby/pull/330 as the most recent example)

rmosolgo:
Sorry about the merge conflict, could you put the new tests in `Compatibility::QueryParserSpecification` instead?

cjoudrey:
> Sorry about the merge conflict, could you put the new tests in `Compatibility::QueryParserSpecification` instead?

Absolutely! 😀 Love that you split up the `SchemaParser` tests.

lib/graphql/input_object_type.rb
```diff
@@ -75,16 +75,14 @@ def coerce_non_null_input(value)
 
       arguments.each do |input_key, input_field_defn|
         field_value = value[input_key]
-        field_value = input_field_defn.type.coerce_input(field_value)
 
-        # Try getting the default value
-        if field_value.nil?
```
cjoudrey: @rmosolgo or @theorygeek I would love your :eyes: on this particular line.

Was the objective to use the default value if `value[input_key]` was undefined or was the goal to also catch the case where `coerce_input` returns `nil`?

I rewrote this code differently and tests still pass so it makes me wonder if we were missing a test or if the intended behaviour was to only use default value when argument was not provided a value.

rmosolgo: I think it did both, [`coerce_input` short-circuits](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L77) the provided `coerce_*` functions in the case of nil, so this nil check serves a double purpose: 

- if the value was initially `nil` (on a key miss), it would be nil here because of that short-circuit 
- if there was a value but it failed coercion, the coerce function should return nil

Looks to me like your code covers both those cases

cjoudrey: The case I don't cover at the moment is if `coerce_input` returns `nil` because of a failed coercion we used to use the `default_value`. I'm unsure if that was intended since all tests pass.

theorygeek: Is there a discrepancy between arguments and input objects, though?
- If `coerce_input` returns nil for a field argument, the query fails
- If `coerce_input` returns nil for an input object, the application code receives the default value

cjoudrey: > So, I think I'd say, we were probably missing a test for the old behavior. But that said, maybe there's an argument to be made that `field` and `field(arg: null)` should be treated differently? (Or `{ }` and `{ arg: nil }`)

Yeah, they absolutely should be treated differently. One is omitted and one is explicitly `null`. This PR introduces  that difference. 😀 `field(arg: null)` wasn't a thing before.

theorygeek: I think the answer is no: if `coerce_input` is given something that isn't nil, but it returns nil, it means the input was invalid.

lib/graphql/define/assign_argument.rb
```diff
@@ -2,16 +2,21 @@ module GraphQL
...
-      def self.call(target, name, type = nil, description = nil, default_value: nil, &block)
+      def self.call(target, name, type = nil, description = nil, **rest, &block)
         argument = if block_given?
           GraphQL::Argument.define(&block)
         else
           GraphQL::Argument.new
         end
+
+        if (rest.keys - [:default_value]) != []
+          raise ArgumentError.new("unknown keyword: #{rest.keys.first}")
```
rmosolgo: `rest.keys.first` _could_ be `:default_value`, right? Should this be 

```ruby 
unexpected_keys = rest.keys - [:default_value] 
if unexpected_keys.any? 
  raise(ArgumentError, "Unknown argument options: #{unexpected_keys}"
end 
```

?

cjoudrey: Good catch. Yeah absolutely, I meant to grab the first key of the subtraction array.

If we want to be _correct_, we may want to do pluralization:

```
irb(main):006:0> a(one: 1, two: 2, three: 3)
ArgumentError: unknown keywords: two, three
...
irb(main):007:0> a(one: 1, two: 2)
ArgumentError: unknown keyword: two
...
irb(main):008:0>
```

lib/graphql/schema/printer.rb
```diff
@@ -132,16 +132,22 @@ def print_input_value(arg)
           def print_value(value, type)
             case type
             when FLOAT_TYPE
+              return 'null' if value.nil?
```
cjoudrey: This makes me a bit sad, maybe this method is doing too much. It's both validating `type` and coercing `value` to IDL.

cjoudrey:
👀 @rmosolgo I think I've handled all cases. I'm unsure about some of the code. Would love your thoughts.

lib/graphql/argument.rb
```diff
@@ -21,6 +21,15 @@ class Argument
...
     ensure_defined(:name, :description, :default_value, :type=, :type)
 
+    def default_value?
+      !!@has_default_value
+    end
+
+    def default_value=(new_default_value)
+      @has_default_value = true
+      @default_value = new_default_value
+    end
```
rmosolgo: I guess there's no way to _unset_ a default value at this point, but I think that'll be fine, cross that bridge when we get to it

lib/graphql/compatibility/query_parser_specification.rb
```diff
@@ -63,6 +63,9 @@ def test_it_parses_inputs
                   object: {a: [1,2,3], b: {c: "4"}}
                   unicode_bom: "\xef\xbb\xbfquery"
                   keywordEnum: on
+                  nullValue: null
+                  nullValueInList: {a: null, b: "b"}
```
rmosolgo: `nullValueInObject` ?

lib/graphql/define/assign_argument.rb
```diff
@@ -2,16 +2,22 @@ module GraphQL
   module Define
     # Turn argument configs into a {GraphQL::Argument}.
     module AssignArgument
-      def self.call(target, name, type = nil, description = nil, default_value: nil, &block)
+      def self.call(target, name, type = nil, description = nil, **rest, &block)
```
rmosolgo: The alternative here, though I'm not sure it's worth it, is 

```ruby 
NO_DEFAULT = :__graphql_assign_argument_no_default__
def self.call(..., default_value: NO_DEFAULT)
  # ...
    if default_value != NO_DEFAULT 
      # assign the default value 
    end 
end 
```

The advantage is that we get built-in errors for kwargs. But I guess it's a bit unusual, there's probably no "net decrease" in complexity :P

lib/graphql/introspection/input_value_type.rb
```diff
@@ -8,8 +8,12 @@
   field :type, !GraphQL::Introspection::TypeType
   field :defaultValue, types.String, "A GraphQL-formatted string representing the default value for this input value." do
     resolve ->(obj, args, ctx) {
-      value = obj.default_value
-      value.nil? ? nil : JSON.generate(obj.type.coerce_result(value), quirks_mode: true)
+      if obj.default_value?
+        value = obj.default_value
+        value.nil? ? 'null' : JSON.generate(obj.type.coerce_result(value), quirks_mode: true)
```
rmosolgo: If `value.nil?`, what happens when we call 

```
coerced_value = obj.type.coerce_result(nil)
JSON.generate(coerced_value, quirks_mode: true)
```

would that give us `null` ?

lib/graphql/schema/build_from_definition.rb
```diff
@@ -150,13 +150,23 @@ def build_input_object_type(input_object_type_definition, type_resolver)
...
         def build_input_arguments(input_object_type_definition, type_resolver)
           input_object_type_definition.fields.map do |input_argument|
-            default_value = case input_argument.default_value
-            when GraphQL::Language::Nodes::Enum
-              input_argument.default_value.name
-            else
-              input_argument.default_value
+            kwargs = {}
+
+            if !input_argument.default_value.nil?
```
rmosolgo: Here, we check for `nil?` because `null` is represented by `NullValue`, right?

lib/graphql/introspection/input_value_type.rb
cjoudrey: I think the answer there is: it depends.

Every type can implement `coerce_result(nil)` the way they want and I didn't want this behaviour to depend on that.

For instance:

```
irb(main):004:0> GraphQL::ListType.new(of_type: GraphQL::ID_TYPE).coerce_result(nil)
=> [nil]
```

lib/graphql/argument.rb
cjoudrey: Right 😞 I wasn't sure about that either, I figured if it comes up we can introduce a way to unset it somehow.

lib/graphql/schema/printer.rb
rmosolgo: can we put it above the case? 

```ruby 
if value.nil? 
  "null" 
else 
  case ... 
end 
```

? 

(I see that NonNullType doesn't have this early return, but it _does_ make a recursive call, so if value is `nil`, then it will get to that condition _eventually_!)

cjoudrey: Yeah, my worry was that if we early return `if value.nil?` you lose the type-checking for that `type`.

> I see that NonNullType doesn't have this early return

Correct, I did this because a `NonNullType` cannot be `null` but its elements can.

spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
```diff
@@ -65,4 +65,126 @@
...
+          field(arg: [null])
+        }
+      |}
+
+      it "finds no errors" do
+        assert_equal [], errors
+      end
+    end
+
+    describe "list" do
```
rmosolgo: is this a list? should this be input object?

spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
```diff
@@ -65,4 +65,126 @@
     }
     assert_includes(errors, fragment_error)
   end
+
+  describe "null value" do
```
rmosolgo: one thing I'm missing is what happens if you have type `[Int]!` and you provide value `null`? Does it become `[null]`? or is it invalid? (I think `1` would become `[1]` in this above case)

cjoudrey: If the type is `[Int]!` it cannot be `null`, but it could be `[null]`.

As far as I understand, non-null types cannot be `null`.

rmosolgo:
:tada: :tada: :tada: this so awesome! Thanks for the great test coverage, too. I had just a couple questions about cases I didn't understand. 

Down the line, I wonder if we can find a way to dry up 

```ruby
kwargs = {}
if arg.default_value? 
  kwargs[:default_value] = arg.default_value
end 

do_someting(**kwargs)
```

But I think it should wait for another day :)

lib/graphql/introspection/input_value_type.rb
rmosolgo: :+1: sounds better to leave it as is. It _might work_ to send it through coerce_result but you end up printing something other than the user-provided value!!

spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
rmosolgo: Nil check comes before list coercion :+1:

cjoudrey:
Updated PR description with some example usages and addressed review comments. 😀

rmosolgo:
👯  so awesome, thank you for all your work on this!! :dancing_men:

Issue #368, opened 2016-10-31
Accessing resolver state async


josh:
GitHub uses [graphql-batch](https://github.com/Shopify/graphql-batch) pattern for most of its data loading. This leads to most resolvers returning Promises. Thats been great.

However, there are some subtle issues accessing certain resolver `context` properties async.

The resolver process uses many **mutable** values that are updated as the resolve walks the query tree. So unless a resolve block captures the state at the start of the block, it maybe different by the time the promise is resolved.

``` ruby
resolve -> (foo, args, ctx) {
  expected_ast_node = context.ast_node
  loadDataAndReturnAPromise().then {
    actual_ast_node = context.ast_node
    expected_ast_node != actual_ast_node
  }
}
```

I think its going to be a pain, but I think we really need to start ensuring any context values passed to a resolver are immutable.

<hr>

To: @rmosolgo 
CC: @charliesome @mcolyer

rmosolgo:
You ask hard questions! :P 

I think we need some new object, I hate to write it after just disowning ExecutionContext but 


```ruby 
class FieldContext
  def initialize(query_ctx, irep_node, path); end 
  # delegate stuff to context and irep_node 
  attr_reader :irep_node, :path
end 
```

somehow have to prevent going crazy and making those when we don't need 'em though

rmosolgo:
(I guess that's like an `ExecFrame` from `DeferredExecution`, maybe it's time to polish that baby up ...)

rmosolgo:
1.2.0 puts a `FieldResolutionContext` into the resolve function, which not modified as the query continues. I think that should do the trick, let me know if not!

PR #371, opened 2016-11-02
Execution "path" from fragment fields is wrong

josh:
Heh, I keep finding more problems with the implementation of error `"path"`. Basically my fault, I wrote this thing.

The latest is that resolving fields within a fragment context have an incomplete path. They only walk up to the root of the fragment definition rather than the full query stack.

I'm thinking that using `irep_node` is probably the wrong path for recording this path stack. And I'm feeling a little stuck about where to track this stack best.

<hr>

To: @rmosolgo 
CC: @mcolyer

### DIFF
--- a/spec/graphql/execution_error_spec.rb
+++ b/spec/graphql/execution_error_spec.rb
@@ -185,4 +185,63 @@ describe GraphQL::ExecutionError do
       assert_equal(expected_result, result)
     end
   end
+
+  describe "fragment query when returned from a field" do
+    let(:query_string) {%|
+    query MilkQuery {
+      dairy {
+        ...Dairy
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


rmosolgo:
Deferred execution has `path` built-in. I want to support it  properly by changing execution in a big way, and I think it will support batching. But I need some time ...

josh:
> Deferred execution has path built-in.

Oh, what file is that class in? Is it different than the serialize execution implementation?

rmosolgo:
#168  💓

rmosolgo:
I spent some time on that branch last night, it might be hard to ever get it in shape. I'm trying to introduce `Execution::Frame` in #378 which will also be the immutable, field-level context requested in #368

PR #373, opened 2016-11-02
Fix nested typed fields

### LINKED ISSUE #370: Large query causes incorrect returned results
I have a large query to receive notifications to my app with about 18 fragments to return different data for each notification type. When the query is at full size, A2VConversionFailed wrongly returns results similar to, but the other notifications return correct results, including StatusUpdated. Here is the incorrect returned json:
```
[
{
  "id"=>"208", 
  "type"=>"a2v_conversion_failed", 
  "release"=>{
    "file_song_name"=>"Untitled", 
    "artist"=>{
      "name"=>"Zachery"
    }, 
    "id_name"=>"Distribution Release #467", 
    "id"=>"467"
  }
}
]
```
 When I alter the query to have only one fragment:
```
 ... on A2VConversionFailed {
          release {
            file_song_name
            artist {
              name
            }
          }
        }
```
the results are correct:
```
[
{
  "id"=>"208", 
  "type"=>"a2v_conversion_failed", 
  "release"=>{
    "file_song_name"=>"Untitled", 
    "artist"=>{
      "name"=>"Zachery"
    }
  }
}
]
```
I have each notification set up in my schema as orphan types. Each notification is initialized as an ObjectType with specific fields that it is allowed to return:
```
module Vydia
  module GraphQL
    module V1
      module Types
        module Objects
          module Notifications

            A2VConversionFailed = ::GraphQL::ObjectType.define do
              name "A2VConversionFailed"
              description "A Notification with type: a2v_conversion_failed"

              interfaces [Interfaces::Notification]

              field :release, Objects::Release
            end
          end
        end
      end
    end
  end
end
```

```
module Vydia
  module GraphQL
    module V1
      module Types
        module Objects
          module Notifications

            StatusUpdated = ::GraphQL::ObjectType.define do
              name "StatusUpdated"
              description "A Notification with type: status_updated"

              interfaces [Interfaces::Notification]

              field :release, Objects::Release

            end
          end
        end
      end
    end
  end
end
```

Each notification includes the Notification Interface: 
```
module Vydia
  module GraphQL
    module V1
      module Types
        module Interfaces

          Notification = ::GraphQL::InterfaceType.define do
            name "Notification"
            description "A user Notification"

            field :id, !types.ID
            field :type, !types.String

          end
        end
      end
    end
  end
end
```

Here is my query:

```
{
      notifications {
        id
        type
        ... on A2VConversionFailed {
          release {
            file_song_name
            artist {
              name
            }
          }
        }
        ... on ArtistUpdated {
          user_name
          artist_name
        }
        ... on AuthError {
          broadcast {
            unauthorized
            failed
            name
            network {
              name
            }
          }
          release {
            file_song_name
            id_name
            artist {
              name
            }
          }
        }
        ... on DeploymentConfirmed {
          broadcast {
            network {
              name
            }
          }
          release {
            file_song_name
            artist {
              name
            }
          }
        }
        ... on DeploymentScheduled {
          broadcast {
            pretty_scheduled_time
            network {
              name
            }
          }
          release {
            file_song_name
            artist {
              name
            }
          }
        }
        ... on MergeCompleted {
          old_account
          new_account
        }
        ... on MergeConflict {
          old_account
          new_account
        }
        ... on NewAdminMessage {
          message_author_name
          release {
            id
            file_song_name
            artist {
              name
            }
          }
        }
        ... on NewMergeRequest {
          old_account
          new_account
        }
        ... on NewMessage {
          message_author_name
          release {
            id_name
            file_song_name
            artist {
              name
            }
          }
        }
        ... on ProcessingFinished {
          media {
            artist {
              name
            }
          }
          release {
            file_song_name
          }
        }
        ... on PublisherNewComment {
          content
          user_name
          release {
            file_song_name
            artist {
              name
            }
          }
        }
        ... on PublisherNewVideoRequested {
          user_name
          release {
            file_song_name
            artist {
              name
            }
          }
        }
        ... on PublisherNewVideoSubmitted {
          media {
            file_song_name
            artist {
              name
            }
          }
        }
        ... on StatusUpdated {
          release {
            id
            id_name
            file_song_name
            artist {
              name
            }
          }
        }
        ... on UploadFailed {
          max_file_size
        }
        ... on UploadFinished {
          media {
            file_song_name
            artist {
              name
            }
          }
        }
      }
    }
```

rmosolgo:
Fixes #370 

That was a doozy! 

Sadly this is a breaking change to `InternalRepresentation::Node`, but I don't see a way around it. 

The old structure kept _shallow_ typed branches in the form of `definitions[type] # => field_defn`. But the problem was that `children` wasn't branched by type, so when we deeply merged children, we lost information about which type condition that branch applied to. 

Now, `children` is replaced by `typed_children`, whose keys are GraphQL types and values are `{ string => node }` hashes (what `children` was previously). 

There's still an issue: when _executing_ divergent branches, it's possible that we'll do redundant work. For example, this would result in double work: 

```graphql
{
  thing {
    ... on InterfaceA {
      inner1 { inner2 }
    }
 
    ... on InterfaceB {
      inner1 { inner2 }
   }
}
```

We see that `inner1` is not a scalar, so we go execute it, but we get nothing new. 

Somehow this needs to _keep_ type condition info as you merge nested selections, so that we only traverse it once.

### DIFF
--- a/lib/graphql/analysis/analyze_query.rb
+++ b/lib/graphql/analysis/analyze_query.rb
@@ -32,8 +32,10 @@ module GraphQL
     def reduce_node(irep_node, reducer_states)
       visit_analyzers(:enter, irep_node, reducer_states)
 
-      irep_node.children.each do |name, child_irep_node|
-        reduce_node(child_irep_node, reducer_states)
+      irep_node.typed_children.each do |type_defn, children|
+        children.each do |name, child_irep_node|
+          reduce_node(child_irep_node, reducer_states)
+        end
       end
 
       visit_analyzers(:leave, irep_node, reducer_states)
--- a/lib/graphql/analysis/field_usage.rb
+++ b/lib/graphql/analysis/field_usage.rb
@@ -26,10 +26,10 @@ module GraphQL
 
       def call(memo, visit_type, irep_node)
         if irep_node.ast_node.is_a?(GraphQL::Language::Nodes::Field) && visit_type == :leave
-          irep_node.definitions.each do |type_defn, field_defn|
-            field = "#{type_defn.name}.#{field_defn.name}"
-            memo[:used_fields] << field
-            memo[:used_deprecated_fields] << field if field_defn.deprecation_reason
+          field = "#{irep_node.owner_type.name}.#{irep_node.definition.name}"
+          memo[:used_fields] << field
+          if irep_node.definition.deprecation_reason
+            memo[:used_deprecated_fields] << field
           end
         end
 
--- a/lib/graphql/analysis/query_complexity.rb
+++ b/lib/graphql/analysis/query_complexity.rb
@@ -46,7 +46,7 @@ module GraphQL
               type_complexities = memo[:complexities_on_type].pop
               child_complexity = type_complexities.max_possible_complexity
               own_complexity = get_complexity(irep_node, memo[:query], child_complexity)
-              memo[:complexities_on_type].last.merge(irep_node.definitions, own_complexity)
+              memo[:complexities_on_type].last.merge(irep_node.owner_type, own_complexity)
             end
           end
         end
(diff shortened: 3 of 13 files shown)


rmosolgo:
OMG, this is basically the non-deterministic finite automaton => deterministing finite automaton proble. 

BRB, getting a bachelors in CS.

rmosolgo:
I can't solve it :'( 

The benchmark for the normal case is not affected, and slow case is the one that didn't work anyways, so I think it's still an improvement, though not a perfect one. 

I have to update the docs on Irep Node and maybe I can provide some path for backwards compatibility.

rmosolgo:
I put back the old method on `Node` so at least existing analyzers don't break. 

I checked out the built-in analyzers from master and gave 'em a run, they still passed with this refactor. I think there will be cases where the traversal is a bit different -- you "dive in" to selections on a certain type where you might not have before. Before, it would have been wrong :)

rmosolgo:
This is also a "breaking" change to analyzers: selections on type-dependent parts of hte queries will be visited _separately_ instead of together. 

These are separate now: 

```graphql
... on A { inner1 { inner2, inner3 } }
... on B { inner1 { inner2, inner4 } } 
```

Previously you would have visited: 

- `inner1` with 2 entries in `definitions` & 3 entries in `children`
- (then visit each of `children`): 
  - `inner2` with 1 entry in `definitions` & 0 entries in `children`
  - `inner3` with 1 entry in `definitions` & 0 entries in `children`
  - `inner4` with 1 entry in `definitions` & 0 entries in `children`

Now, you visit: 

- `inner1` with 2 entries in `definitions` & 3 entries in `children` & 2 entries in `typed_children`
- (then go down one of `typed_children`'s entries) 
  - `inner2` with 1 entry in `definitions`, 0 entries in `children`, 0 entries in `typed_children`
  - `inner3` with 1 entry in `definitions`, 0 entries in `children`, 0 entries in `typed_children`
 - (then go down the other `typed_children` entry)
  - `inner2` with 1 entry in `definitions`, 0 entries in `children`, 0 entries in `typed_children`
  - `inner4` with 1 entry in `definitions`, 0 entries in `children`, 0 entries in `typed_children`

I think this is _better_ because it avoids the same bug that hit query execution, but it's also _different_! You can see how the built-in analyzers were refactored for this flow.

rmosolgo:
This is gonna have bugs. And it resolves some fields twice when it shouldn't. I'm working on a ... fix for the fix. I can't put this code on rubygems though 😖

PR #376, opened 2016-11-04
Make schema builder an extendable class

xuorig:
This makes `from_definition` take an additional `builder` argument. Allowing you to override the Default schema builder with your own.

This could be used to inject resolvers for example.

```
class MyCustomBuilder << GraphQL::Schema::BuildFromDefinition::Builder
  def build_field(field_definition)
    field = super

    if field.name.include?('Robert')
      field.resolve = ->(_, _, _) { raise NoRobertFieldsAllowed }
    end

    field
  end
end
```
```
Schema.from_definition(definition_string, builder: MyCustomBuilder)
```

Main change is moving `Builder` to class so it's possible to extend it.

I decided to keep a simple `Builder` class instead of having `DefaultBuilder` + `Builder` base class, but I'm open to ideas.

I also exploded `build_types` in smaller methods so it's easier to extend.

(Tests are still in `build_from_definition_spec`)

WDYT ?

@rmosolgo 
cc @cjoudrey


rmosolgo:
Love the idea, not crazy about the [`extends` relationship](http://asserttrue.blogspot.com/2009/02/inheritance-as-antipattern.html) between library code and user code. 

What's the main use case of this? 

- If it's to print a subset of the schema, what about adding an `except:` parameter, like the one `Schema#execute` accepts?
- If it's to assert that the schema contains no invalid members, what about building a hook for schema assertions?

xuorig:
@rmosolgo so the main use case would be generating a usable Schema from an IDL. For example adding resolvers (maybe random data, or trying to be intelligent and finding business logic 👼 ), or really changing the way fields are getting built.

I'm mainly interested by adding resolvers. Another option would be passing some kind of config to the builder.

rmosolgo:
Oh, interesting! It would be a custom extension for code generation, right? Just curious, how would it be supported from the parser?

rmosolgo:
Derp, I have it backwards, this is the ~~loader~~ builder, not the printer

rmosolgo:
Would love to have an API for working with parsed schema members but ... not with `<` 😬

Issue #274, opened 2016-09-23
First-class async support 


rmosolgo:
It would be nice to provide first-class support to asynchronous resolvers. It's true, CRuby can't do too much in parallel, but what it _can_ do (I/O) might be worth supporting. 
#168 (DeferredExecution) may provide a platform for this: the execution strategy maintains a queue of "threads" and resolves them one at a time. Perhaps resolvers could signal to `graphql-ruby` that they're "working", and graphql-ruby could push the thread to the end of the queue. Then, when it's done with other things, it could revisit that thread and see if it's ready (and, if it's the only thread left, wait for it).

I'd like to support this behavior in a general way without requiring any dependencies. I know `graphql-batch` uses `promise.rb`, and personally, I'm interested in `concurrent-ruby`'s `Future`. 

I can imagine general support working like this: 
- You provide a GraphQL-friendly wrapper around the async "primitive" of your choice. Perhaps it's an object that responds to `.ready?`, `.value` and `.wait`. 
- You tell GraphQL that the resolve function will return an async response, eg 
  
  ``` ruby
  field :remote_call, RemoteThingType, async: true do 
    resolve -> (obj, args, ctx) { fetch_from_http(obj.thing_id) }
  end 
  
  # or ... 
  
  field :remote_call, RemoteThingType do 
    resolve_async -> (obj, args, ctx) { fetch_from_http(obj.thing_id) }
  end 
  
  # or ... 
  
  field :remote_call, RemoteThingType do 
    resolve -> (obj, args, ctx) { GraphQL.async(fetch_from_http(obj.thing_id))  }
  end 
  
  # or ... 
  
  field :remote_call, RemoteThingType do 
    resolve GraphQL.async(-> (obj, args, ctx) { fetch_from_http(obj.thing_id) })
  end 
  ```

Another important element to this is that it should also support asynchronous connection resolvers. In that case, GraphQL "messes with" the resolver you give it, so we should be careful to support that case too. 

What do you think, would this kind of support be valuable in the gem? Are there other factors that should be considered?

cjoudrey:
cc @dylanahsmith what I was referring to the other day.

staugaard:
Absolutely. I would even say that this is a high priority "Issue". 

Please don't try to roll your own concurrency code though. I think that this is one of those areas where it is better to pull in a dependency. concurrent-ruby is a good choice.

dylanahsmith:
[...]

PR #378, opened 2016-11-05
In-house batching 

rmosolgo:
For reasons (#274), the gem should have built-in support for batched resolution: one pass to gather needs, then resolve those needs together, then dole out the results to the fields that asked for them. This is what graphql-batch does so well, but moving it to core library code will make things a bit easier to maintain. 

## Todo 

- [ ] Add query-level batching
- [ ] Implement null propagation
- [ ] Support nested loaders 
- [ ] Check graphql-batch tests for any cases I've missed

### DIFF
--- a/lib/graphql/define.rb
+++ b/lib/graphql/define.rb
@@ -1,4 +1,5 @@
 require "graphql/define/assign_argument"
+require "graphql/define/assign_batch_resolve"
 require "graphql/define/assign_connection"
 require "graphql/define/assign_enum_value"
 require "graphql/define/assign_global_id_field"
--- a/lib/graphql/define/assign_batch_resolve.rb
+++ b/lib/graphql/define/assign_batch_resolve.rb
@@ -1,0 +1,10 @@
+module GraphQL
+  module Define
+    module AssignBatchResolve
+      def self.call(field_defn, loader, *loader_args, resolve_func)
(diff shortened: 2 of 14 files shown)


cjoudrey:
cc @dylanahsmith

dylanahsmith:
Why not focus supporting asynchronous resolvers, since batching can be implemented on top of it in addition to concurrent and parallel execution?

For instance, the overrides in graphql-batch are purely to support async field resolution, so bringing that in-house would remove the coupling between the two libraries, which still allowing for experimentation with different ways of doing the execution.

Right now this PR seems to couple batch loading to resolving a field in a way that only allows a single batch load per field.  Since a resolver may need to do multiple queries that should be batched and will need to do something with that result, I suspect you will end up with something more similar to promises/futures to provide the more general solution, so I don't think you should start with the current coupling of a field to a single batch loader.

I'll try to take a stab at providing a minimal solution for async support.

rmosolgo:
Yes, I'd love to see that! It's hard for me to understand quite what the right extension point is, so any help in that direction would be very welcome!

jsdnxx:
I agree that async field resolution is the right extension point here. It should be possible without changing the API or adding an `async: true` flag or anything like that, depending on the promise or future implementation.

rmosolgo:
@dylanahsmith I took another crack at at last night. If you can take a look, I'd love to hear your feedback: #386 

In any case, it's much better than this one!

PR #379, opened 2016-11-06
Fix nested object selections

rmosolgo:
`test_it_only_resolves_fields_once_on_typed_fragments` describes a bug that was added in #373 (but not shipped to rubygems), it's a bad bug. 

Since this messed with `query.context.irep_node`, I took care of a couple other things while I was at it: 

- Use `Context#path` to track error paths (fixes #371 , @josh, is it OK to deprecate `Node#path`? This was the only usage I know)
- Remove `ExecutionContext` from the universe, thread a `Query` through serial execution instead. 

## Todo :

- [x] Make sure graphql-batch didn't depend on execution context APIs 
- [ ] Update query analyzer visitation to visit each object type


lib/graphql/internal_representation/node.rb
```diff
@@ -99,6 +99,7 @@ def owner
       end
 
       def path
+        warn("InternalRepresentation::Node#path is deprecated, use Query::Context#path instead")
```
josh: 👍 `"Deprecated and is totally broken, why are you even using this"` lol

lib/graphql/query/context.rb
```diff
@@ -54,10 +58,9 @@ def add_error(error)
           raise TypeError, "expected error to be a ExecutionError, but was #{error.class}"
         end
 
-        error.ast_node = irep_node.ast_node unless error.ast_node
-        error.path = irep_node.path unless error.path
+        error.ast_node ||= irep_node.ast_node
+        error.path ||= path.dup
```
josh: Nice

lib/graphql/query/serial_execution/field_resolution.rb
```diff
@@ -2,21 +2,25 @@ module GraphQL
...
-          @field = execution_context.get_field(parent_type, irep_node)
-          @arguments = execution_context.query.arguments_for(irep_node, @field)
+          @query = query
+          @field = @query.get_field(parent_type, irep_node.definition_name)
+          @arguments = @query.arguments_for(irep_node, @field)
         end
 
         def result
           result_name = irep_node.name
+          @query.context.path.push(result_name)
```
josh: Idea to make this immutable:

``` ruby
 previous_path = @query.context.path
 @query.context.path = (@query.context.path + [result_name]).freeze
# ...
ensure
  @query.context.path = previous_path
```

rmosolgo: Well, it breaks graphql-batch pretty bad as-is, so maybe now's the time!

rmosolgo:
Ok, added a field-level, immutable (in practice) context thingy. 

But I broke a GraphQL::Batch test along the way, gotta track that down: 

```
 1) Failure:
GraphQL::BatchTest#test_mutation_execution [/Users/rmosolgo/projects/graphql-batch/test/batch_test.rb:311]:
--- expected
+++ actual
@@ -1 +1 @@
-{"data"=>{"count1"=>0, "incr1"=>{"value"=>1, "load_value"=>1}, "count2"=>1, "incr2"=>{"value"=>2, "load_value"=>2}}}
+{"data"=>{"count1"=>0, "incr1"=>{"value"=>1, "load_value"=>#<GraphQL::Batch::Promise:0xXXXXXX @state=:fulfilled, @callbacks=[], @value=1>}, "count2"=>1, "incr2"=>{"value"=>2, "load_value"=>#<GraphQL::Batch::Promise:0xXXXXXX @state=:fulfilled, @callbacks=[], @value=2>}}}
```

rmosolgo:
Huh, apparently they were assigning a value to an accessor that was never used, but now it is used, and that :boom: s. Hooray, coupling!

PR #380, opened 2016-11-06
Add Sinatra example to docs

robinjmurphy:
Hello! Thanks for the great library. I've been building a GraphQL Sinatra example this weekend and have really enjoyed working with it.

I thought it might be useful for other people, so I've added a link to my example in the docs underneath your Rails one.

### DIFF
--- a/guides/related_projects.md
+++ b/guides/related_projects.md
@@ -7,6 +7,7 @@ Want to add something? Please open a pull request [on GitHub](https://github.com
 ## Code
 
 - `graphql-ruby` + Rails demo ([src](https://github.com/rmosolgo/graphql-ruby-demo) / [heroku](http://graphql-ruby-demo.herokuapp.com))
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! I'd love to include a link to a Sinatra example, thank you for sharing. 

This HTML file is actually the generated output of a Markdown file on the master brach: 

https://github.com/rmosolgo/graphql-ruby/blob/master/guides/related_projects.md

Could you update that file instead? Then I'll regenerate the HTML & push to github pages.

robinjmurphy:
Thanks. I've updated the branch. Should be updating the Markdown in master now.

rmosolgo:
Thanks!

rmosolgo:
generated and pushed! http://rmosolgo.github.io/graphql-ruby/related_projects

PR #381, opened 2016-11-06
Add ability to set default ID separator globally

daqo:
While upgrading an old app using version 0.18.5 to 0.19.0, I noticed that there is no way to set the default ID separator. It is somehow a pain in the neck to pass the separator to UniqueWithinType's encode/decode methods.

This PR adds ability to set default ID separator globally in a project.

Please let me know what you think


lib/graphql/schema/unique_within_type.rb
```diff
@@ -1,14 +1,21 @@
...
+      class << self
+        attr_accessor :default_id_separator
+      end
+      self.default_id_separator = "-"
 
       module_function
 
+      def set_default_separator(separator)
+        self.default_id_separator = separator
+      end
```
davidcelis: With an `attr_accessor`, there's not really a benefit to the extra setter method, is there? `GraphQL::Schema::UniqueWithinType.default_id_separator=` would be a public interface

daqo:
@davidcelis Yea. I removed that wrapper. I added it at the first place just to make it explicit

rmosolgo:
Awesome! It used to work this way, I guess I just spaced out when I refactored it. Thanks!

daqo:
@rmosolgo Thank you Robert

PR #382, opened 2016-11-07
Add strategy alias for execution_strategy in FieldResolutionContext

dylanahsmith:
https://github.com/rmosolgo/graphql-ruby/pull/379 changed the API of FieldResolution which broken graphql-batch https://github.com/Shopify/graphql-batch/pull/32 for version 1.2.0 of the graphql gem.  Now I'm not sure what the best way to get access to the execution strategy from a FieldResolution object.

I found that adding an strategy alias in FieldResolutionContext would allow https://github.com/rmosolgo/graphql-ruby/pull/379 to work with older and newer versions of the graphql gem.

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -73,6 +73,7 @@ module GraphQL
         end
 
         def_delegators :@context, :[], :[]=, :spawn, :query, :schema, :warden, :errors, :execution_strategy
+        alias_method :strategy, :execution_strategy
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: sorry about that, I must have run the tests on my branch 😩  Clearing up this category of issues is next on my list (by providing an actually-useful API for this!)

rmosolgo:
Shipped in 1.2.1

Issue #377, opened 2016-11-04
Rubocop extension


cjoudrey:
It would be interesting to provide a set of Rubocop cops in the repo that people can optionally enable in their own `rubocop.yml` to do checks on their schema.

Some example checks:

- Naming conventions for Mutations and other parts of the Schema
- Relay conventions
  - All mutations should include a `clientMutationId` field
  - Mutations that remove nodes should include a `deletedId` field
  - Single `input` argument on mutations
- Detect breaking schema changes by comparing checked-in IDL with current commits IDL
- Detect N+1 (if possible) or provide tools to enable people to check this

Feel free to add more ideas.

cc @xuorig @Willianvdv @brandonblack 

**Edit:** Rubocop might not be a silver bullet here since in some cases we might have to run the Ruby code, for instance detecting N+1s with static code analysis might not be feasible.

kdaigle:
cc @gjtorikian @josh

josh:
Nice. I started working on some graphql-client specific cops already. https://github.com/github/graphql-client/tree/master/lib/rubocop/cop/graphql Would be rad to have some basic ones here as well.

gjtorikian:
We don't have a cop for this (we use tests) but "has a description" would be grand, as well as conventions for resolve styling (`-> {...}` vs `do ... end`).

Willianvdv:
What do you think about a linter that checks if the name of a mutation starts with capital letter?
 So:
```
# Good
AddCommentMutation = GraphQL::Relay::Mutation.define do
  name "AddComment"
end

# Bad
AddCommentMutation = GraphQL::Relay::Mutation.define do
  name "addComment"
end
```

rmosolgo:
Also relevant: https://github.com/cjoudrey/graphql-schema-linter

rmosolgo:
I have no plans to work on this but I would welcome suggestions here or in another issue if someone else wants to do it!

DmitryTsepelev:
I've implemented some _graphql-ruby_ checks in [rubocop-graphql](https://github.com/DmitryTsepelev/rubocop-graphql), would be happy to hear back ideas and feedback

PR #384, opened 2016-11-08
WIP - Add Schema comparison

cjoudrey:
There's still a couple of TODO's in the code, but I figured I'd open this to get some :eyes: on it.

**What?**

The goal of this PR is to provide a way to compare two schemas and obtain a list of changes. 

These changes can either be breaking or not.

This PR adds one class method on `Schema` and an instance method on `Schema`:

```ruby
changes = Schema.compare(schema1, schema2)
```

```ruby
changes = schema1.compare(schema2)
```

**Uses**

First and foremost, this can be used in a linter tool or tests (i.e. #377) to ensure breaking changes aren't accidentally pushed into production.

Assuming an IDL version of your schema is checked in to git, you could do the following:

```ruby
old_schema = Schema.from_definition(File.read('path_to_idl.graphql'))
changes = Schema.compare(old_schema, new_schema)
breaking_changes = changes.keep_if{|change| change[:breaking_change]}
```

Another use case could be to detect new deprecations and surface this to API clients via another system (e-mail, change log, etc...)


lib/graphql/schema.rb
```diff
@@ -47,6 +48,7 @@ module GraphQL
   #
   class Schema
     include BuildFromDefinition
+    include SchemaComparator
```
rmosolgo: I see a couple of includes have found their way here already, but how about implementing methods to call through to the underlying class instead of including a module? For example: 

```ruby 
def self.compare(s1, s2) 
  SchemaComparator.find_changes(s1, s2) 
end 

def compare_to(s2) 
  self.class.compare(self, s2)
end 
```

The advantages I see are:

- The reader knows exactly what methods on `Schema` come from `SchemaComparator` 
- Someone who modifies `SchemaComparator` can easily find its usages in `Schema` 

Would that work in this case?

lib/graphql/schema/schema_comparator.rb
```diff
@@ -0,0 +1,713 @@
...
+        def find_changes_in_interface_type(old_type, new_type)
+          find_changes_in_object_type_fields(old_type, new_type)
+        end
+
+        def type_removed(type)
+          {
+            type: TYPE_REMOVED,
+            description: "`#{type}` type was removed",
+            breaking_change: true,
+          }
```
rmosolgo: Is there an advantage to using Hashes here instead of a dedicated object? The nice thing about objects is that the API is a method call, and a typo results in a no-method-error instead of silent failure.

rmosolgo:
:+1: pumped to see this feature come together!!

rmosolgo:
I added the tag just to help with eyeballing the PR list. Please remove it when those remaining `TODO`s are addressed!

rmosolgo:
I'm interested in focusing on this for 1.6.0 (1.5 isn't out yet 😅 ), but I'm considering approaching at the AST level so that the same diff framework could be applied to schema definitions _and_ query documents. I think the same API could be supported, but it would go to and from string under the hood. 

Are you interested in trying that? Or do you mind if I close this PR?

cjoudrey:
Exciting!

I _think_ catching breaking changes and diff-ing ASTs are two different things, but one definitely builds on the other.

Suppose we had some sort of `diff(document1, document2)`, what did you have in mind for its return value?

I've had this PR in the back of my mind for a while. To be honest, I'm not super happy with it.

I felt like returning non-breaking changes like `DESCRIPTION_CHANGED` was getting a bit out of hand. I also wasn't :100: convinced on what I was returning.

Two use-cases come to mind when I think of schema comparison:

- **Catching breaking/dangerous changes before they make it to production.** - i.e. You could have a bot that listens to PR webhooks and adds an extra status in your pull request to let you know you made a breaking/dangerous change to the schema.

- **Notifying developers of new deprecations.** - i.e. You could have a bot that watches a production schema and whenever a field/type becomes deprecated, it could notify 3rd party developers. This could be pretty powerful when combined with field usage data.

Some references for this: [graphql-js](https://github.com/graphql/graphql-js/blob/578c76cee55a8677e2c3fb39563ea441d053fb51/src/utilities/findBreakingChanges.js) / [sangria](https://github.com/sangria-graphql/sangria/blob/5680b34eff0c086facb3eecf87b347584d06c7cc/src/main/scala/sangria/schema/SchemaComparator.scala)

What are some use-cases for wanting to diff query ASTs? Caching?

rmosolgo:
Here's the API I envisioned: 

```ruby 
diff = GraphQL::Language.diff(doc1, doc2)
# #<GraphQL::Language::Diff>
if diff.breaking_changes.any? 
  puts "There are breaking changes, update the manifest to approve them: "
  diff.breaking_changes.each { |d| puts d }
end 
```

Since it works on the AST level, we define breaking changes for each AST node: 

- Considering schema definitions:
  - Breaking changes for an object type definition node are:
    - removing a field 
    - a breaking change to any of its fields
  - Breaking changes for a field definition node are:
    - new required arguments
    - removing an argument 
    - new return type which is not a supertype of the previous return type 
    - a breaking change to any of its arguments
- In the same way, considering query documents: 
  - Breaking changes to an operation node are: 
    - adding a required variable 
    - breaking changes to any variables (eg, changing the type or name)
    - breaking changes to its selections
  - Breaking changes to a selection set are: 
    - removing a field 
    - breaking changes to any field node (eg, requiring a new variable) 

One part that's a bit tough is fragments: a breaking change to a fragment might be OK if it's covered by corresponding changes at fragment usage sites.

Anyhow, the pros of doing it AST level that I see are: 
  - Similar execution flow: traverse both ASTs, trying to find matching nodes and compare them
  - Minimal dependencies: do we actually _need_ the "real" schema (or query) objects to make a diff, or is the AST sufficient? (I _think_ it's sufficient)

> What are some use-cases for wanting to diff query ASTs? Caching?

Yes, say for example you wanted to [manage all queries in static files](https://rmosolgo.github.io/graphql-ruby/pro/persisted_queries), then the ability to predict (and avoid) breakages would be really helpful when maintaining those files!

cjoudrey:
Some inspiration: http://sangria-graphql.org/learn/#schema-comparison

rmosolgo:
This new project seems to fit the bill: https://github.com/xuorig/graphql-schema_comparator. 

If it becomes preferable to include this behavior in `graphql` proper, I'm open to that, but no hurry from my perspective.

PR #388, opened 2016-11-08
Expose Schema#from_definition directly in schema.rb

cjoudrey:
As per https://github.com/rmosolgo/graphql-ruby/pull/384#discussion_r86922870, this will make it a bit more obvious to the reader what is being exposed in `Schema` as well as where `BuildFromDefinition` is used.

:eyes: @rmosolgo


rmosolgo:
👌 sweet!

PR #391, opened 2016-11-09
Fix validations for dynamic fields

### LINKED ISSUE #387: Missing validation for __type on non-root object
`__schema` and `__type` are only accessible from on the root of a query operation, but this isn't caught during validation, leading to an unhandled exception (e.g. NoMethodError) from doing a query like `{ test { __type } }`

Here is a script that reproduces the exception

```ruby
require 'graphql'

TestType = GraphQL::ObjectType.define do
  name "Test"

  field :unused, types.String
end

  name "Query"

  field :test, TestType do
    resolve -> (obj, args, ctx) {
      {}
    }
  end
end

  query QueryType
end

Schema.execute('{ test { __type } }')
```

which results in the following exception output

```
/Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:151:in `arguments_for': undefined method `arguments' for nil:NilClass (NoMethodError)
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:15:in `initialize'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:11:in `new'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:11:in `block in resolve'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:10:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:10:in `resolve'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/value_resolution.rb:44:in `resolve'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:51:in `get_finished_value'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/field_resolution.rb:21:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:16:in `block in resolve'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:10:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/selection_resolution.rb:10:in `resolve'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution/operation_resolution.rb:6:in `resolve'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/serial_execution.rb:20:in `execute'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/executor.rb:33:in `execute'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/executor.rb:15:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:254:in `get_result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:238:in `call'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:96:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/schema.rb:145:in `execute'
	from tmp/test.rb:23:in `<main>'
```

This is a low priority issue.  Just something that came up from some developer playing around with GraphQL.

cjoudrey:
Fixes #387

Validation of dynamic fields (i.e. `__schema`, `_type`, and `__typename`) seem broken (see tests for examples) because we were early returning for them.

I believe we can treat these fields as any other field since [`Schema.get_field`](https://github.com/rmosolgo/graphql-ruby/blob/1246912b777b5e82665349c592f85f8885c376b0/lib/graphql/schema.rb#L150) is aware of these special fields.

Also added a bit more test coverage around validating these dynamic fields.

:eyes: @rmosolgo

### DIFF
--- a/lib/graphql/static_validation/rules/fields_are_defined_on_type.rb
+++ b/lib/graphql/static_validation/rules/fields_are_defined_on_type.rb
@@ -6,7 +6,6 @@ module GraphQL
       def validate(context)
         visitor = context.visitor
         visitor[GraphQL::Language::Nodes::Field] << ->(node, parent) {
-          return if context.skip_field?(node.name)
           parent_type = context.object_types[-2]
           parent_type = parent_type.unwrap
           validate_field(context, node, parent_type, parent)
@@ -16,7 +15,7 @@ module GraphQL
       private
 
       def validate_field(context, ast_field, parent_type, parent)
-        if parent_type.kind.union?
+        if parent_type.kind.union? && ast_field.name != '__typename'
           context.errors << message("Selections can't be made directly on unions (see selections on #{parent_type.name})", parent, context: context)
           return GraphQL::Language::Visitor::SKIP
         end
--- a/lib/graphql/static_validation/arguments_validator.rb
+++ b/lib/graphql/static_validation/arguments_validator.rb
@@ -17,8 +17,6 @@ module GraphQL
                 return
               end
             end
-          elsif context.skip_field?(parent.name)
-            return
           elsif parent.is_a?(GraphQL::Language::Nodes::Directive)
             parent_defn = context.schema.directives[parent.name]
           else
--- a/lib/graphql/static_validation/rules/fields_have_appropriate_selections.rb
+++ b/lib/graphql/static_validation/rules/fields_have_appropriate_selections.rb
@@ -7,7 +7,6 @@ module GraphQL
 
       def validate(context)
         context.visitor[GraphQL::Language::Nodes::Field] << ->(node, parent)  {
-          return if context.skip_field?(node.name)
           field_defn = context.field_definition
           validate_field_selections(node, field_defn, context)
         }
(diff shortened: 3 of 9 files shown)


lib/graphql/static_validation/rules/fields_are_defined_on_type.rb
```diff
@@ -16,7 +15,7 @@ def validate(context)
       private
 
       def validate_field(context, ast_field, parent_type, parent)
-        if parent_type.kind.union?
+        if parent_type.kind.union? && ast_field.name != '__typename'
```
cjoudrey: As per the spec:

> Because unions do not define fields, fields may not be directly selected from a union‐typed selection set, with the exception of the meta‐field `__typename`.

rmosolgo:
Sweet, thanks for taking care of this!

(Basically, "Validator, stop copping out!")

rmosolgo:
:ship: 1.2.2 !

PR #392, opened 2016-11-11
`input_fields` did not ensure_define

wenley:
Based on the original order of execution, `input_fields` was still aliased to the un-`ensure_defined` version of `arguments`.


rmosolgo:
Thanks! Should we add a test for this behavior so we don't break it in the future?

wenley:
Done!

rmosolgo:
Great, thanks for this fix!

PR #393, opened 2016-11-11
[documentation] fix the logic in the Instrumenter::field setup exampl…

ipoval:
…e to return the intended value

Overview:
- it seems valid to return a previous resolver value in the :field instrumentation example - without it we have an error;
- minor comment change: previous comment mentioned `return` and `copy` which may not be correct because there is no need to return `the field` in `#instrument` and the field is not a copy, but the same reference.

### DIFF
--- a/guides/schema/instrumentation.md
+++ b/guides/schema/instrumentation.md
@@ -27,14 +27,17 @@ class FieldTimerInstrumentation
       old_resolve_proc = field.resolve_proc
       new_resolve_proc = ->(obj, args, ctx) {
         Rails.logger.info("#{type.name}.#{field.name} START: #{Time.now.to_i}")
-        old_resolve_proc.call(obj, args, ctx)
+        resolved = old_resolve_proc.call(obj, args, ctx)
         Rails.logger.info("#{type.name}.#{field.name} END: #{Time.now.to_i}")
+        resolved
       }
 
       # Return a copy of `field`, with a new resolve proc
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for taking a look at this! 

>  field is not a copy, but the same reference

`#redefined` is _meant_ to make a new object and apply the new definitions to it ([spec](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/define/instance_definable_spec.rb#L79)), but is it broken? Why do you say it's not a copy?

ipoval:
@rmosolgo , sorry, bad test on my side. reverted comment change

rmosolgo:
Hmm, is the change in this PR: 

```ruby
# from 
if condition 
  # do stuff 
end 

# to 
return unless condition 
# do stuff
```

I'm not a fan of `unless`, so I don't want to include it in the documentation. 

I think there _is_ a bug here, right? `#instrument` should aways return a `GraphQL::Field`, but the documentation example sometimes returns `nil`. Is that addressed in this PR? (Perhaps I've misread it)

ipoval:
@rmosolgo, updated the PR;
correct, I think there are 2 bugs in the documentation:
- `#instrument` should return a `GraphQL::Field`
- `new_resolve_proc` should return value of `old_resolve_proc`

rmosolgo:
Thank you for fixing this! I've just redeployed the site.

PR #394, opened 2016-11-12
Fix unicode character escape in literal strings

dylanahsmith:
I found out that [EscapedUnicode](http://facebook.github.io/graphql/#EscapedUnicode) was being treated as valid by the lexer, resulting in a `Parse error on bad Unicode escape sequence` GraphQL::ParseError exception.


rmosolgo:
I thought to myself "We have [a spec for this](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/compatibility/query_parser_specification.rb#L77), how did _it_ pass?" 

And I see that it's because we're testing the _whole_ string for those patterns, so if the string contains _any_ valid escapes, then it won't be identified as `BAD_UNICODE_ESCAPE`. I've pushed another test to demonstrate the difference. 

In the short term, it seems like that string matching should be more local (ie, it should look at _each_ escape and assert that it is valid). 

In the long term, I'd like to refactor the Lexer to make replacements on the _first_ pass, like libgraphqlparser does. 

(Also, I found this in the meantime! https://github.com/rmosolgo/graphql-ruby/pull/396)

dylanahsmith:
Pushed a fix for the other bug you found with validating character escapes in string tokens.

rmosolgo:
Sweet, thanks!!

rmosolgo:
:ship: in 1.2.4!

PR #396, opened 2016-11-13
fix(lexer) don't preserve state between runs

rmosolgo:
Oops, this was previously not threadsafe! Fortunately it's only used for schema parsing, which I doubt people are doing on multiple threads 😬


rmosolgo:
This was probably a memory leak as well.  Since `@previous_token` was never cleared, the linked list kept growing and growing, and old tokens were never GCed!

PR #399, opened 2016-11-16
Add instrumenters after schema.define 

rmosolgo:
This allows you to call `#instrument` _after_ `.define`, eg 

```ruby 
MySchema = Schema.define do 
  # ... 
end 

MySchema.instrument(:field, SomeInstrumenter)
```

It points out some issues with this approach: if an attribute wasn't assigned via `.define`, it's not copied by `redefine`. I found some bugs in the gem. I'll keep an eye out for others: if they pop up, I'll either find another way to support this API or investigate `#dup` as an implementation of `#redefine`.

cc @tmeasday


tmeasday:
Awesome, thanks @rmosolgo, this will allow me to delete some code I'm not a big fan of.

PR #401, opened 2016-11-17
refactor(Visitor) remove global hooks; remove sibling skips

rmosolgo:
I think these APIs are unused (or, can be unused with this refactor). These are removed:

- Returning `SKIP` no longer skips sibling nodes (did that actually work? untested.)
- Visitor no longer has global `enter` and `leave` hooks (called for every node)

cc @josh do you care if I remove these? I didn't see them in use at graphql-client. 

(I don't know anyone else consuming `GraphQL::Language::Visitor`)

### DIFF
--- a/lib/graphql/language/visitor.rb
+++ b/lib/graphql/language/visitor.rb
@@ -17,16 +17,9 @@ module GraphQL
       # node right away
       SKIP = :_skip
 
-      # @return [Array<Proc>] Hooks to call when entering _any_ node
-      attr_reader :enter
-      # @return [Array<Proc>] Hooks to call when leaving _any_ node
-      attr_reader :leave
-
       def initialize(document)
         @document = document
         @visitors = {}
-        @enter = []
-        @leave = []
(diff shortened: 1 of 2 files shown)


tmeasday:
I was using it @rmosolgo! https://github.com/apollostack/optics-agent-ruby/blob/master/lib/optics-agent/normalization/query.rb#L19-L26

Will work around.

rmosolgo:
Sorry! I'll be sure to tag you if I come messing with this stuff again!

tmeasday:
Thanks! Here's the workaround if anyone ever runs into this: https://github.com/apollostack/optics-agent-ruby/commit/6af24ea23dede474e785e7107a90b8e468906d09

rmosolgo:
Ironically, I used to be really picky about always deprecating stuff before removing it, but nobody actually used the gem. 

Now people actually use it and I've gotten out of the habit! I'll try to play more conservatively in that area from now on.

PR #403, opened 2016-11-18
convert enum type values to names for error messages

zvkemp:
In error messages for `EnumType#validate_non_null_input`, display the enum value names instead of the default ruby object representation.


rmosolgo:
Thanks! 

I applied your commit with a small refactor (d61e87f): instead of implementing `EnumValue#to_s`, I made an explict call to `.map(&:name)` when building the error message. I did this because, personally, I find `#to_s` hard to refactor because it's hard to find usages of it. You can't grep the project for `to_s` since it's often called implicitly by Ruby!

rmosolgo:
:ship: in 1.2.4!

PR #405, opened 2016-11-18
Returning a GraphQL::ExecutionError in a mutation with return_fields …

aldrinmartoq:
…works as expected.


rmosolgo:
Awesome! Thank you!!

rmosolgo:
I added a small change: ebe644e

this way, `clientMutationId` can still be in the result, in case Relay needs it for cache management (I don't know if it does!)

PR #412, opened 2016-11-22
feat(Query) run after_query hooks even if an error occurs

rmosolgo:
With this change, `after_query` hooks are _always_ run, even if the query raises a Runtime error

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -254,6 +254,7 @@ module GraphQL
       def call
         @instrumenters.each { |i| i.before_query(@query) }
         result = get_result
+      ensure
         @instrumenters.each { |i| i.after_query(@query) }
         result
       end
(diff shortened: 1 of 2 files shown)


rmosolgo:
(I did this with graphql-batch in mind, but it's a good move either way!)

lib/graphql/query.rb
```diff
@@ -254,6 +254,7 @@ def initialize(query, instrumenters)
       def call
         @instrumenters.each { |i| i.before_query(@query) }
         result = get_result
+      ensure
         @instrumenters.each { |i| i.after_query(@query) }
         result
```
dylanahsmith: This line and the `result = ` assignment above isn't needed, since the result of the ensure block isn't returned.

rmosolgo: Oh! Thanks, fixed in 84143b7

PR #415, opened 2016-11-23
Use GraphQL::InvalidNullError in the new executor

dylanahsmith:
## Problem

As mentioned in https://github.com/rmosolgo/graphql-ruby/pull/334#issuecomment-262440736, there is a regression on master where GraphQL::ExecutionError is now used instead of an invalid null error, which prevents this internal error from being found in the context errors by error class for internal error reporting.

E.g.

```ruby
require 'graphql'

  name "Query"

  field :nonNullButIs, !types.String do
    resolve proc { nil }
  end
end

  query QueryType
end

query = GraphQL::Query.new(Schema, '{ nonNullButIs }')
result = query.result
internal_errors = query.context.errors.reject { |err| err.is_a?(GraphQL::ExecutionError) }
puts internal_errors.inspect
```

would output `[#<GraphQL::InvalidNullError: Cannot return null for non-nullable field Query.nonNullButIs>]` but now outputs `[]`

## Solution

Go back to putting a GraphQL::InvalidNullError in the context errors, while we are still putting this internal error in the execution context.

### DIFF
--- a/lib/graphql/compatibility/lazy_execution_specification.rb
+++ b/lib/graphql/compatibility/lazy_execution_specification.rb
@@ -53,7 +53,7 @@ module GraphQL
             {
               push(value: 2) {
                 push(value: 3) {
-                  fail1: push(value: 13) {
+                  fail1: push(value: 14) {
                     value
                   }
                   fail2: push(value: 14) {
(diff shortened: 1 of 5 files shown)


rmosolgo:
Sorry for introducing this incompatibility, thanks for the fix!

PR #416, opened 2016-11-24
Type error hook

rmosolgo:
There are a few judgment calls made by library internals that could be customized by users: 

- What to do when we get `nil` for a non-null type (currently, the library adds an error to the errors key) 
- What to do when a union/interface object can't be resolved to a member type (currently, the library :boom:s) 

This PR adds a hook for those two cases, with the default implementation being the current behavior. 

In the future, this hook could be expanded to handle more cases, like scalar coercion.

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -122,9 +122,7 @@ module GraphQL
       def resolve_value(parent_type, field_defn, field_type, value, selection, field_ctx)
         if value.nil?
           if field_type.kind.non_null?
-            error = GraphQL::InvalidNullError.new(parent_type.name, field_defn.name, value)
-            error.set_backtrace(caller)
-            field_ctx.errors << error
+            field_ctx.schema.type_error(value, field_defn, parent_type, field_ctx)
             PROPAGATE_NULL
           else
             nil
@@ -185,7 +183,7 @@ module GraphQL
             possible_types = query.possible_types(field_type)
 
             if !possible_types.include?(resolved_type)
-              raise GraphQL::UnresolvedTypeError.new(selection.definition_name, field_type, parent_type, resolved_type, possible_types)
+              field_ctx.schema.type_error(value, field_defn, parent_type, field_ctx)
             else
               resolve_value(
                 parent_type,
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -162,7 +162,7 @@ module GraphQL
             result
           when GraphQL::TypeKinds::NON_NULL
             wrapped_type = field_type.of_type
-            resolve_value(
+            inner_value = resolve_value(
               parent_type,
               field_defn,
               wrapped_type,
@@ -170,6 +170,11 @@ module GraphQL
               selection,
               field_ctx,
             )
+            if inner_value.nil?
+              PROPAGATE_NULL
+            else
+              inner_value
+            end
           when GraphQL::TypeKinds::OBJECT
             resolve_selection(
               value,
@@ -184,6 +189,7 @@ module GraphQL
 
             if !possible_types.include?(resolved_type)
               field_ctx.schema.type_error(value, field_defn, parent_type, field_ctx)
+              PROPAGATE_NULL
             else
               resolve_value(
                 parent_type,
--- a/lib/graphql/query/serial_execution/selection_resolution.rb
+++ b/lib/graphql/query/serial_execution/selection_resolution.rb
@@ -13,7 +13,7 @@ module GraphQL
               query_ctx
             ).result
 
-            if field_result.values.any? { |v| v == GraphQL::Execution::Execute::PROPAGATE_NULL }
+            if field_result.values[0] == GraphQL::Execution::Execute::PROPAGATE_NULL
               selection_result = nil
               break
             else
--- a/lib/graphql/query/serial_execution/selection_resolution.rb
+++ b/lib/graphql/query/serial_execution/selection_resolution.rb
@@ -6,12 +6,19 @@ module GraphQL
           selection_result = {}
 
           selection.each_selection(type: current_type) do |name, subselection|
-            selection_result.merge!(query_ctx.execution_strategy.field_resolution.new(
+            field_result = query_ctx.execution_strategy.field_resolution.new(
               subselection,
               current_type,
               target,
               query_ctx
-            ).result)
+            ).result
+
+            if field_result.values.any? { |v| v == GraphQL::Execution::Execute::PROPAGATE_NULL }
+              selection_result = nil
+              break
+            else
+              selection_result.merge!(field_result)
+            end
           end
 
           selection_result
--- a/lib/graphql/schema/default_type_error.rb
+++ b/lib/graphql/schema/default_type_error.rb
@@ -1,0 +1,19 @@
+module GraphQL
+  class Schema
+    module DefaultTypeError
+      def self.call(value, field, parent_type, query_ctx)
+        if field.type.kind.non_null?
+          if value.nil?
+            query_ctx.errors << GraphQL::InvalidNullError.new(parent_type.name, field.name, value)
+          else
+            # it was caused by GraphQL::ExecutionError
+          end
+        else
+          resolved_type = query_ctx.query.resolve_type(value)
+          possible_types = query_ctx.query.possible_types(field.type.unwrap)
+          raise GraphQL::UnresolvedTypeError.new(field.name, field.type, parent_type, resolved_type, possible_types)
+        end
+      end
+    end
+  end
+end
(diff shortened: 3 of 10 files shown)


lib/graphql/execution/execute.rb
```diff
@@ -164,14 +162,19 @@ def resolve_value(parent_type, field_defn, field_type, value, selection, field_c
...
+            inner_value = resolve_value(
               parent_type,
               field_defn,
               wrapped_type,
               value,
               selection,
               field_ctx,
             )
+            if inner_value.nil?
+              PROPAGATE_NULL
```
dylanahsmith: When can this happen?  If I remove this `if` condition that you added the tests still pass.  Because it seems like if we did get here, then `field_ctx.schema.type_error` should be used to avoid ignoring the invalid null

rmosolgo: This can happen for non-null union or interface types. The non-null call to `resolve` passes the value to the union or interface type, but then the object fails type resolution, so it gets replaced with `nil`.  Here's the [spec](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/compatibility/execution_specification.rb#L357). I added some [similar code](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/serial_execution/value_resolution.rb#L53-L57) to make SerialExecution pass and must have assumed it was required here too! 

But I think this is handled by `FieldResult` in this case. Removed in 4b02c42. Thanks!

lib/graphql/schema/default_type_error.rb
```diff
@@ -0,0 +1,19 @@
+module GraphQL
+  class Schema
+    module DefaultTypeError
+      def self.call(value, field, parent_type, query_ctx)
+        if field.type.kind.non_null?
+          if value.nil?
+            query_ctx.errors << GraphQL::InvalidNullError.new(parent_type.name, field.name, value)
+          else
+            # it was caused by GraphQL::ExecutionError
```
dylanahsmith: The type_error hook isn't called for a GraphQL::ExecutionError, nor should it.  So I think this is just an unused code path, since this hook is only being used for invalid null and unresolved type errors.

dylanahsmith:
I ran the graphql-batch tests and found that this PR caused a regression as seen from this graphql-batch test failure:

```
  1) Failure:
GraphQL::GraphQLTest#test_non_null_field_promise_raises [/Users/dylansmith/src/graphql-batch/test/graphql_test.rb:116]:
--- expected
+++ actual
@@ -1 +1 @@
-{"data"=>nil, "errors"=>[{"message"=>"Error", "locations"=>[{"line"=>1, "column"=>3}], "path"=>["nonNullButPromiseRaises"]}]}
+{"data"=>{"nonNullButPromiseRaises"=>:__graphql_propagate_null__}, "errors"=>[{"message"=>"Error", "locations"=>[{"line"=>1, "column"=>3}], "path"=>["nonNullButPromiseRaises"]}]}
```

rmosolgo:
😖

lib/graphql/schema/default_type_error.rb
```diff
@@ -0,0 +1,19 @@
+module GraphQL
+  class Schema
+    module DefaultTypeError
+      def self.call(value, field, parent_type, query_ctx)
+        if field.type.kind.non_null?
```
dylanahsmith: The parameters to this hook don't actually give you the information about the type that had the error in the case of wrapping types.  For example, the field may be of type `[String!]` in which case the field itself is nullable and drop to the else branch where it assumes it is an unresolved type and try to resolve the `nil` type.

E.g. this script reproduces the bug

```ruby
require 'graphql'

  name "Query"

  field :invalid_null_list, types[!types.String] do
    resolve proc { [nil] }
  end
end

  query QueryType
end

query = GraphQL::Query.new(Schema, '{ invalid_null_list }')
puts query.result
```

which results in the output

```
/Users/dylansmith/src/graphql-ruby/lib/graphql/schema.rb:218:in `resolve_type': Can't determine GraphQL type for: nil, define `resolve_type (obj, ctx) -> { ... }` inside `Schema.define`. (NotImplementedError)
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:198:in `resolve_type'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/schema/default_type_error.rb:12:in `call'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/schema.rb:284:in `type_error'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:129:in `resolve_value'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:156:in `block in resolve_value'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:148:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:148:in `each_with_index'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:148:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:148:in `map'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:148:in `resolve_value'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:116:in `continue_resolve_field'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:87:in `resolve_field'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:30:in `block in resolve_selection'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/internal_representation/selection.rb:40:in `block in each_selection'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/internal_representation/selection.rb:37:in `each'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/internal_representation/selection.rb:37:in `each_selection'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:28:in `resolve_selection'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/execution/execute.rb:8:in `execute'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/executor.rb:33:in `execute'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query/executor.rb:15:in `result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:272:in `get_result'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:256:in `call'
	from /Users/dylansmith/src/graphql-ruby/lib/graphql/query.rb:102:in `result'
	from tmp/test.rb:16:in `<main>'
```


It would be much nicer to just get the exception object (e.g. GraphQL::InvalidNullError or GraphQL::UnresolvedTypeError) which could inherit from a common base class like GraphQL::TypeError and expose the relevant values that you are currently passing into this lambda.

rmosolgo: That's a great idea! I was trying to think about what to inject, and it didn't strike me to use an error object. That will make it much easier for people to tell "what went wrong" AND it will make it easier to send new info into the function in a non-breaking way in the future!

lib/graphql/query/serial_execution/selection_resolution.rb
```diff
@@ -6,12 +6,19 @@ def self.resolve(target, current_type, selection, query_ctx)
...
-            selection_result.merge!(query_ctx.execution_strategy.field_resolution.new(
+            field_result = query_ctx.execution_strategy.field_resolution.new(
               subselection,
               current_type,
               target,
               query_ctx
-            ).result)
+            ).result
+
+            if field_result.values[0] == GraphQL::Execution::Execute::PROPAGATE_NULL
```
dylanahsmith: The problem mentioned in https://github.com/rmosolgo/graphql-ruby/pull/416#issuecomment-263365631 is likely related to this check assuming that the selection has been resolved, but in graphql-batch it lazily resolves fields so would only have the promise value as a placeholder at this point.  Instead, get_finished_value should remain the place to check for nulls to propagate.

staugaard:
Have you guys considered merging graphql-batch into graphql-ruby?

rmosolgo:
#386 covers my needs in that area: it copies the resolution strategy, but not the logic for composing & executing `Promise`s

What I really need is to grasp this assumption about when errors are raised, and write a test for _that_!

lib/graphql/query/serial_execution/selection_resolution.rb
dylanahsmith: Actually, doing the same thing in get_finished_value won't really allow the null to propagate for objects, since at that point only the object has resolved and not its fields.  It is probably easiest to just stick with exceptions, especially if you are just going to pass exceptions to the type_error callback.

PR #421, opened 2016-11-25
fix(ConnectionField) use redefine to capture connection behaviors

### LINKED ISSUE #417: Instrumenters break relay connections
Using the simple field instrumenter documented [here](https://rmosolgo.github.io/graphql-ruby/schema/instrumentation), `connection` fields lose their arguments and fail with

```
{"errors"=>[{"message"=>"Field 'people' doesn't accept argument 'first'", "locations"=>[{"line"=>1, "column"=>3}], "fields"=>["query", "people", "first"]}]}
```

I believe the reason is that this [line](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/relay/connection_field.rb#L34) is incompatible with `field.redefine` (i.e. it gets forgotten).

[This workaround](https://github.com/apollostack/optics-agent-ruby/commit/5311c3dc975880e6979998e1c91d3e4ecc867596) appears to work.

rmosolgo:
Fixes #417

### DIFF
--- a/lib/graphql/relay/connection_field.rb
+++ b/lib/graphql/relay/connection_field.rb
@@ -24,17 +24,17 @@ module GraphQL
         memo
       end
 
-      # Turn A GraphQL::Field into a connection by:
+      # Build a connection field from a {GraphQL::Field} by:
       # - Merging in the default arguments
       # - Transforming its resolve function to return a connection object
       # @param underlying_field [GraphQL::Field] A field which returns nodes to be wrapped as a connection
       # @param max_page_size [Integer] The maximum number of nodes which may be requested (if a larger page is requested, it is limited to this number)
-      # @return [GraphQL::Field] The same field, modified to resolve to a connection object
+      # @return [GraphQL::Field] A redefined field with connection behavior
       def self.create(underlying_field, max_page_size: nil)
-        underlying_field.arguments = DEFAULT_ARGUMENTS.merge(underlying_field.arguments)
+        connection_arguments = DEFAULT_ARGUMENTS.merge(underlying_field.arguments)
         original_resolve = underlying_field.resolve_proc
-        underlying_field.resolve = GraphQL::Relay::ConnectionResolve.new(underlying_field, original_resolve, max_page_size: max_page_size)
-        underlying_field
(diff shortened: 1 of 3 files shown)


rmosolgo:
This is the second time I've run into bugs where schema members get attributes _outside_ of `.define`,  so they aren't carried over by `.redefine`. I'm conflicted: 

The benefit of `.new` + replaying `.define` blocks is that the two instances don't share any state: each object starts with fresh hashes for `arguments`, `fields` and enum values, etc.

I briefly tried using `#dup` for this behavior, so that `redefine` = `.dup` + `define` . But the problem is, the two objects (old and new) would _share_ some state (eg, `arguments` hash), so modifying the new instance would also modify the old instance. 

The current arrangement is a bit dangerous: it's possible to modify these objects _outside_ of `.define` blocks, but if you do so (that is, if the _library itself_ does so 😩 ), we're all gonna have a bad time. On the other hand, the out-of-the-box tool for shallow copying is not sufficient for this task. 

So I can see some options: 

- Prevent modification outside of `.define`. It's Ruby, this is hard to do. Also, I don't love the thought of forcing this "DSL" on everyone. Sometimes you just want to assign attributes to things!
- Support a robust, `.define`-agnostic copy operation for schema members. This sounds good, but I'm afraid that it will be hard to maintain: what if you add some internal state for a new feature but forget the corresponding duplication logic? 
- Just leave it as is (:-1:)

rmosolgo:
I'm gonna apply this as-is and keep thinking :S

tmeasday:
@rmosolgo what kind of data structures are being used here? Sort of struggling to imagine a scenario where a "deep `dup`" wouldn't work but forcing `define` would. Also your option 3 would share many of the same properties as 2--if you add a new feature that uses internal state, you'd have to remember to make it shareable and `dup`-agnostic also, which would be an easy thing to forget and get wrong.

rmosolgo:
>  a scenario where a "deep dup" wouldn't work but forcing define would

only a case of user error 😆  which is equally possible, if not more, in the other option

> option 3 would share many of the same properties as 2

good point!

PR #423, opened 2016-11-26
Get validation errors for invalid (string, number, integer) input for InputObject

mvgijssel:
It would be cool if InputObject arguments would have validation when receiving something different from a Hash. Currently it will crash with something like `NoMethodError: undefined method `each' for "":String`, when for example a user is tampering with the request payload to the server. 

Failing line is https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/input_object_type.rb#L57, so is it acceptable to build some sort of guard before that statement? 

I have no problem trying to make the added specs pass, but not sure if it's expected behaviour or not. Or if I should just create an input object type with custom coercion.


rmosolgo:
I think a guard there is a great solution!

lib/graphql/input_object_type.rb
```diff
@@ -51,6 +51,18 @@ def kind
     def validate_non_null_input(input, warden)
       result = GraphQL::Query::InputValidationResult.new
 
+      if input.is_a? Array
```
mvgijssel: Did a check for Array and Enumerable instead of Hash, because you use `MinimumInputObject` (spec/support/minimum_input_object.rb) in your specs. So I'm not sure if input could be something else than a Hash. I could rewrite MinimumInputObject to be a Hash or leave it like this. Your call!

rmosolgo: Agreed `is_a?(Hash)` is not a great solution. Rails 5 params fail that check, right? 

How does this approach compare to something like `rescue NoMethodError` ? Is there a viable solution that doesn't make assumptions about the input's inheritance chain? :thinking:

rmosolgo: Sorry to go quiet on this one, I've just been stumped by it! 

Yes, `respond_to?(:each)` isn't even _quite_ right. I can think of two options, like the ones you just described: 

- Check for `#each`, and assume that it will yield key-value pairs. (This assumption may be wrong, eg Array) 
- Check for `#each`, and inside the block, assert that two values were yielded. I don't think this is really possible: the second yield arg may be `nil`, in which case, we can't tell whether the user _provided_ nil, or whether nil is expressing "nothing was yielded"
- "Normalize" the input with something like `val_h = value.is_a?(Hash) ? value : value.to_h`. This could work, I wish it was more "duck type"-y, but I can't think of anything really _wrong_ with it. 

What do you think of those?

mvgijssel: Haha no worries! I think the normalize solution is best. Calling `.to_h` on an array will only work if it's an array of tuples:
```
[1].to_h
=> TypeError: wrong element type Fixnum at 0 (expected array)

[[1,2,3,4]].to_h
=> ArgumentError: wrong array length at 0 (expected 2, was 4)
```
So the entire solution would be something like
```
val_h = value.is_a?(Hash) ? value : (value.to_h rescue nil)
return if val_h.nil?
```

lib/graphql/input_object_type.rb
```diff
@@ -57,18 +57,25 @@ def kind
...
       # Items in the input that are unexpected
-      input.each do |name, value|
+      input_h.each do |name, value|
         if visible_arguments_map[name].nil?
           result.add_problem("Field is not defined on #{self.name}", [name])
         end
       end
 
       # Items in the input that are expected, but have invalid values
-      invalid_fields = visible_arguments_map.map do |name, field|
```
mvgijssel: removed `invalid_fields = ` because assignment wasn't used

lib/graphql/input_object_type.rb
```diff
@@ -57,18 +57,25 @@ def kind
     def validate_non_null_input(input, warden)
       result = GraphQL::Query::InputValidationResult.new
 
+      input_h = input.is_a?(Hash) ? input : (input.to_h rescue nil)
+
+      if input_h.nil?
+        result.add_problem("Expected #{JSON.generate(input, quirks_mode: true)} to be a hash")
```
rmosolgo: can we say `be a Hash or respond to #to_h`? That way, if someone gets here, they might find a way to help themselves.

rmosolgo:
Ick, I just remembered that Rails 5 ActionController::Parameters don't do well with `to_h`:

```ruby 
params = ActionController::Parameters.new("a" => 1, "b" => { "c" => 2 })
# => <ActionController::Parameters {"a"=>1, "b"=>{"c"=>2}} permitted: >
params.is_a?(Hash)
# => false
params.to_h
# => {}
params.to_unsafe_h
# => {"a"=>1, "b"=>{"c"=>2}}
```

but `each` works fine:

```ruby 
params.each { |k, v| p([k, v]) }
# ["a", 1]
# ["b", <ActionController::Parameters {"c"=>2} permitted: >]
```

So, this will be a breaking change for users on Rails 5 who pass in params without calling `to_unsafe_h`. Maybe it's not a big deal, you probably only need to add `to_unsafe_h` in _one_ place, and it's on everyone's radar anyways since it params changed so much from Rails 4 to 5 ... 

Do you mind clarifying that error message a bit, to mention `#to_h`? I think this will be ok to merge after that, and I will consider the `params` situation separately!

mvgijssel:
I changed the strategy to try and coerce to a Hash `.to_h rescue nil`, which in that case just uses the original `.each` implementation. Also took the liberty of adding specs for ActionController::Parameters, hope it's to your liking! 

UNRELATED: Didn't know the `appraisal` gem, it's pretty cool :).

rmosolgo:
That's great, thanks for working through this!!

PR #424, opened 2016-11-28
Fixes an `NoMethodError` in GraphQL::ScalarType when validating a que…

nevesenin:
…ry string against a schema

### DIFF
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -51,7 +51,7 @@ module GraphQL
     end
 
     def coerce_non_null_input(value)
-      @coerce_input_proc ? @coerce_input_proc.call(value) : value
+      @coerce_input_proc.call(value)
     end
 
     def coerce_input=(proc)
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -51,7 +51,7 @@ module GraphQL
     end
 
     def coerce_non_null_input(value)
-      @coerce_input_proc.call(value)
+      @coerce_input_proc ? @coerce_input_proc.call(value) : value
     end
 
     def coerce_input=(proc)
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -120,7 +120,9 @@ module GraphQL
               ScalarType.define(
                 name: type["name"],
                 description: type["description"]
-              )
+              ) do
+                coerce ->(value) { value }
+              end
             end
           when "UNION"
             UnionType.define(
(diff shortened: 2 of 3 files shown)


rmosolgo:
Thanks for this patch! Can you help me understand why it is useful? I mean, why would you have a scalar type without a coercion function? 

Perhaps it's coming from a parsed schema definition. Would it be better to provide a "no-op" coercion function in that case?

nevesenin:
Hi,

there is no scalar type without a coercion function in my schema.

To be honest, I didn't investigate that much what exactly causes this error. It happens when I try to validate a query string against the schema on client side:
```
validator = GraphQL::StaticValidation::Validator.new(schema: @schema)
query = GraphQL::Query.new(@schema, query_string: query_string)

validator.validate(query)
```
The query itself works, when submitting it to the graphql endpoint.

rmosolgo:
Oh, interesting! Where does `@schema` come from in the example above? How is it constructed?

nevesenin:
`@schema` comes from `GraphQL::Schema::Loader.load(downloaded_with_introspection_query_json)`. 

I'll try to provide a complete example.

rmosolgo:
That provides the info I need, thanks! 

When you create a schema with `Loader`, it uses built-in scalars if possible. But if the schema contains a custom scalar, it builds a new one with `name` and `definition` only: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/loader.rb#L120-L123

In that case, it doesn't have a coerce function. 

I'd rather solve this problem by adding "dummy" coerce functions in `Loader`. That way, `ScalarType` can continue to assume that it was properly defined.

rmosolgo:
This change looks great!

 But we should have a test for it, so it continues to work in the future. This is a pain to test ... I wonder if we could make a _fake_ introspection result and use that, for example: 

```ruby
stubbed_introspection_result = { 
  "__schema" => {
    "types" => [
      {
        "name" => "DateTime",
        "kind" => "SCALAR",
      }
    ]
  }
}

schema = GraphQL::Schema::Loader.load(stubbed_introspection_result)
custom_scalar = schema.types["DateTime"]
assert_equal true, custom_scalar.valid_input?("anything", PermissiveWarden)
assert_equal true, custom_scalar.valid_input?(12345, PermissiveWarden)
```

Do you mind adding a test for this?

cjoudrey:
`FromDefinition` might have the same issue. 😔

nevesenin:
I started writing tests, but as you said, it wasn't that easy. This issue doesn't occur when the type is defined on 'Client-Side' so yes, stubbing the introspection result should solve this.

rmosolgo:
Thanks for tracking this down! I grabbed your commit in #445 and added some tests as well as a fix for `FromDefinition`.

nevesenin:
Thanks for that. Sorry that I didn't add tests. I haven't had time.

PR #425, opened 2016-11-28
feat(Execution::Lazy) support lazy connection values

### LINKED ISSUE #420: Handle lazy-resolved values from connection resolve functions
Starting in 1.3.0, graphql-ruby says it will handle certain classes lazily. But this is still broken for connections, since the lazy value is wrapped with a connection (I think?). 

I need to test this behavior and make sure it's uniform with the rest of the gem (ie, connections should Just Work :tm:). 

This issue was first described here: https://github.com/Shopify/graphql-batch/issues/26

rmosolgo:
The forthcoming `lazy_resolve` feature should play nice with connections: if you provide a connection resolve function and it's a registered lazy object, we'll treat it lazily, then continue with a connection object wrapping the lazily-resolved nodes.

Fixes #420

### DIFF
--- a/lib/graphql/compatibility/lazy_execution_specification.rb
+++ b/lib/graphql/compatibility/lazy_execution_specification.rb
@@ -108,6 +108,34 @@ module GraphQL
             ]
             assert_equal expected_pushes, pushes
           end
+
+          def test_it_resolves_lazy_connections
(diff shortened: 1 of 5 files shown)


minardimedia:
Any code example for this

PR #427, opened 2016-11-29
Uphold spec for reserved `__` naming

cjoudrey:
Ref: https://github.com/graphql/graphql-js/pull/600

> GraphQL type system authors must not define any types, fields, arguments, or any other type system artifact with two leading underscores.

Spec: http://facebook.github.io/graphql/#sec-Naming-conventions

---

This PR enforces type, argument, and field names cannot begin with `__`.

:eyes: @rmosolgo

### DIFF
--- a/lib/graphql/schema/validation.rb
+++ b/lib/graphql/schema/validation.rb
@@ -86,7 +86,6 @@ module GraphQL
           }
         end
 
-
         FIELDS_ARE_VALID = Rules.assert_named_items_are_valid("field", ->(type) { type.all_fields })
 
         HAS_ONE_OR_MORE_POSSIBLE_TYPES = ->(type) {
@@ -155,6 +154,17 @@ module GraphQL
             # :ok_hand:
           end
         }
+
+        RESERVED_TYPE_NAME = ->(type) {
+          return unless type.name.start_with?('__')
+          return if INTROSPECTION_TYPES[type.name] && INTROSPECTION_TYPES[type.name] == type
+          "Name #{type.name.inspect} must not begin with \"__\", which is reserved by GraphQL introspection."
+        }
+
+        RESERVED_NAME = ->(named_thing) {
+          return unless named_thing.name.start_with?('__')
+          "Name #{named_thing.name.inspect} must not begin with \"__\", which is reserved by GraphQL introspection."
+        }
       end
 
       # A mapping of `{Class => [Proc, Proc...]}` pairs.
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/validation.rb
```diff
@@ -155,6 +154,17 @@ def self.assert_named_items_are_valid(item_name, get_items_proc)
...
           end
         }
+
+        RESERVED_TYPE_NAME = ->(type) {
+          return unless type.name.start_with?('__')
+          return if INTROSPECTION_TYPES[type.name] && INTROSPECTION_TYPES[type.name] == type
+          "Name #{type.name.inspect} must not begin with \"__\", which is reserved by GraphQL introspection."
+        }
+
+        RESERVED_NAME = ->(named_thing) {
```
cjoudrey: This is a naïve check since there is currently no argument or field name that begins with `__` in the introspection schema. When the day comes, we can build an array of possible values at runtime and compare against that.

lib/graphql/schema/validation.rb
```diff
@@ -201,6 +214,17 @@ def self.assert_named_items_are_valid(item_name, get_items_proc)
           Rules::SCHEMA_CAN_GENERATE_IDS,
         ],
       }
+
+      INTROSPECTION_TYPES = Hash[[
```
cjoudrey: I had initially tried building this list dynamically with `Schema::ReduceTypes` but it was already _too late_ since validation kicks in at that point.

On the flip side, any new type we add to the introspection schema will necessarily have to show up here otherwise tests will fail so we won't forget.

Is there a more elegant solution?

rmosolgo: Lol, at first I thought, how about 

```
def introspection?
  name.start_with?("__")
end
```

😝 derp 

This seems fine for now, if it becomes a maintenance problem we can fix it then!

rmosolgo:
🚀 thanks!

chollier:
makes me want to cry, I've been using `__id` for all our ID fields that are not the Relay Global Node ID ones

rmosolgo:
😖  sorry to hear it! you might be able to minimize the refactor by using aliases, eg 

```graphql
node(id: $nodeId) {
  ... on Thing {
    __id: nonRelayId
  }
}
```

rmosolgo:
I kept hearing [Rich Hickey say "this is how you break software"](https://www.youtube.com/watch?v=oyLBGkS5ICk) so I chickened out and made this a `warn` in the next release!

chollier:
@rmosolgo thanks for that! I guess I'll rename everything _id or something

PR #431, opened 2016-11-30
feat(Schema#type_error) use TypeError to pass info to type_error hook

rmosolgo:
Pass a `TypeError` instance to `Schema#type_error` instead of a long list of marginally useful arguments 😬  Includes assertions about those error objects' APIs.

cc @dylanahsmith thanks for this suggestion!

### DIFF
--- a/lib/graphql/query/serial_execution/value_resolution.rb
+++ b/lib/graphql/query/serial_execution/value_resolution.rb
@@ -7,8 +7,10 @@ module GraphQL
           case value
           when GraphQL::ExecutionError, NilClass
             if field_type.kind.non_null?
-              type_error = GraphQL::InvalidNullError.new(parent_type, field_defn, value)
-              query_ctx.schema.type_error(type_error, query_ctx)
+              if value.nil?
+                type_error = GraphQL::InvalidNullError.new(parent_type, field_defn, value)
+                query_ctx.schema.type_error(type_error, query_ctx)
+              end
               GraphQL::Execution::Execute::PROPAGATE_NULL
             else
               nil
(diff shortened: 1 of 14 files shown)


lib/graphql/schema/default_type_error.rb
```diff
@@ -2,17 +2,14 @@
...
-      def self.call(value, field, parent_type, query_ctx)
-        if field.type.kind.non_null?
-          if value.nil?
-            query_ctx.errors << GraphQL::InvalidNullError.new(parent_type.name, field.name, value)
-          else
-            # it was caused by GraphQL::ExecutionError
+      def self.call(type_error, ctx)
+        case type_error
+        when GraphQL::InvalidNullError
+          if !type_error.parent_error?
```
dylanahsmith: I think this should be handled outside of the type_error callback, so the callback can only be called when there is an internal error to report.  Also, this check is only needed for the serial execution strategy, which is an inconsistency that shouldn't be exposed to the application code.

lib/graphql/query/serial_execution/value_resolution.rb
```diff
@@ -7,7 +7,8 @@ def self.resolve(parent_type, field_defn, field_type, value, selection, query_ct
           case value
           when GraphQL::ExecutionError, NilClass
             if field_type.kind.non_null?
-              query_ctx.schema.type_error(value, field_defn, parent_type, query_ctx)
+              type_error = GraphQL::InvalidNullError.new(parent_type, field_defn, value)
+              query_ctx.schema.type_error(type_error, query_ctx)
```
dylanahsmith: This is where you shouldn't be calling type_error if the value is an GraphQL::ExecutionError.

rmosolgo:
💰 moved that logic for skipping user-provided errors from user hook to library code, bingo! thanks!

Issue #418, opened 2016-11-24
Argument conflict on interface type fragments


eapache:
Currently it is not possible to pass different args to the same field depending on the concrete type. The following example results in an error about an argument conflict:
```graphql
foo {
  ... on TypeOne { image(maxWidth: 200) }
  ... on TypeTwo { image(maxWidth: 300) }
}
```

However, the spec (specifically the bottom of https://facebook.github.io/graphql/#sec-Field-Selection-Merging) gives an example very similar to this as a validly mergeable query since the two fields both cannot be encountered against the same object, so they are safe.

cc @dylanahsmith @swalkinshaw 

Also see https://github.com/graphql/graphql-js/issues/53 and https://github.com/graphql/graphql-js/pull/229

cjoudrey:
I find [the current test](https://github.com/rmosolgo/graphql-ruby/blob/7e2a0ba3ae38937befa97273fff8b7f921e69067/spec/graphql/static_validation/rules/fields_will_merge_spec.rb) for this rule a bit hard to follow.

I'll probably take a stab at breaking it up into multiple tests using the examples provided in the spec.

I'll probably look into fixing this at the same time if nobody gets to it before me.

rmosolgo:
BTW I'm checking out a fix for this now. Before executing the query, we build a tree of nodes that groups fields by each possible object type. I think if I do that grouping sooner, I could use the resulting tree for this validation. Roughly, for each node in tree, make sure all of its backing AST nodes have the same name and same arguments. It introduces an extra visit over that tree, but I don't think it can be helped! (I _think_ the graphql-js validation is also recursive, but I can't quite tell -- I can't nail down the recursive function chain.)

rmosolgo:
I think this is fixed by #512, which will be released with 1.5.0! Please reopen if you still experience this bug with that version.

PR #437, opened 2016-12-05
Better test coverage for FieldsWillMerge validation

cjoudrey:
I ported over a bunch of tests from [graphql-js](https://github.com/graphql/graphql-js/blob/ba422261bfec5ad653a6812fab4718f4f22101c2/src/validation/__tests__/OverlappingFieldsCanBeMerged-test.js).

I also made a change to how we report variable names in error messages. It was previously broken as we were calling `print_arg` on `arg` which was always a `String`.

Before: `Field 'doesKnowCommand' has an argument conflict: {"dogCommand":"SIT"} or {"dogCommand":"dogCommand"}?`

After: `Field 'doesKnowCommand' has an argument conflict: {"dogCommand":"SIT"} or {"dogCommand":"$dogCommand"}?`

There is a few tests that I didn't add because they caused the build to fail (for example: #418). We can either add them to the spec commented out, or I can open an issue for each failing test.

:eyes: @rmosolgo

### DIFF
--- a/lib/graphql/static_validation/rules/fields_will_merge.rb
+++ b/lib/graphql/static_validation/rules/fields_will_merge.rb
@@ -77,7 +77,7 @@ module GraphQL
 
           args = defs.map { |defn| reduce_list(defn.arguments)}.uniq
           if args.length != 1
-            errors << message("Field '#{name}' has an argument conflict: #{args.map {|a| print_arg(a) }.join(" or ")}?", defs.first, context: context)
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: this seems like a win as-is, could you just push a branch with those failing tests? I'd be happy to take a crack at a few of them.

cjoudrey:
:+1: will do!

PR #450, opened 2016-12-09
refactor(Schema::Validation) only warn on __

rmosolgo:
As I was preparing the changelog for 1.3.0, I realized that we could make this transition more gradual. I don't want to put a blocker in front of people who want to upgrade! 

In a later version, we can make this fail by default but provide an option to use the old way. Then maybe someday we can remove it altogether. (I didn't take that approach now because `Schema::Valdiation`'s design is not conducive to it.)

@chollier, this one is for you :)

### DIFF
--- a/lib/graphql/schema/validation.rb
+++ b/lib/graphql/schema/validation.rb
@@ -178,14 +178,23 @@ module GraphQL
         }
 
         RESERVED_TYPE_NAME = ->(type) {
-          return unless type.name.start_with?('__')
-          return if INTROSPECTION_TYPES[type.name] && INTROSPECTION_TYPES[type.name] == type
-          "Name #{type.name.inspect} must not begin with \"__\", which is reserved by GraphQL introspection."
+          if type.name.start_with?('__') && INTROSPECTION_TYPES[type.name] != type
+            # TODO: make this a hard failure in a later version
+            warn("Name #{type.name.inspect} must not begin with \"__\", which is reserved by GraphQL introspection.")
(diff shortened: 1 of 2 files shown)


cjoudrey:
> [...] but provide an option to use the old way. 

Do we need to offer that option? I worry that we're encouraging people to not follow the spec by doing this.

Do you think we could include a date in the warning message and just give people X weeks/months to rename their `__` fields and we revert this PR at that time?

rmosolgo:
I agree that spec compliance is best. Future features may _depend_ on `__` being reserved. 

But, hot on the heels [Rich Hickey's chastisement](https://www.youtube.com/watch?v=oyLBGkS5ICk) and my own recent breakages (eg #415, #401), I'm trying to _not break_ people's stuff. I think it's especially true on this project: adopting GraphQL is a significant risk (because of its newness, mostly), and I don't want adopters to get bit for early adoption, even if the the adoption becomes "wrong" from a future perspective.

For this _particular_ change, isn't it a bit early to break existing code? Or, is there a real penalty to encroaching on the reserved space?

Looking at future versions, I'm open to a breakage down the road. I'd like to give this version a few months to settle in so we can see how big of a problem reserving `__` will be.

cjoudrey:
> For this particular change, isn't it a bit early to break existing code?

I agree with you there.

I was mostly challenging if we could go straight from `warn` to breakage if we waited the proper amount of time (let it be months, that is okay) and if this time was properly documented in the warning.

PR #451, opened 2016-12-09
Add Connection.nodes shorthand

josh:
While the `Connection`/`Edge`/`Node` model provides a useful place to attach join model annotations, most of @github's usage has just been enumerating a connection's node list. Cursor data is simply accessed via `PageInfo.endCursor`.

This change adds and additional `Connection.nodes` as a shortcut for accessing `Connection->edges->node`.

I understand this isn't part of the Relay spec specifically, so maybe it could be opt-in if you think its  necessary?

I'm hoping I'm remember this right, but I believe @dschafer (@facebook) mentioned this alias to me at a GraphQL conference/meetup thing.

### DIFF
--- a/lib/graphql/relay/connection_type.rb
+++ b/lib/graphql/relay/connection_type.rb
@@ -2,8 +2,14 @@
 module GraphQL
   module Relay
     module ConnectionType
+      class << self
+        attr_accessor :default_nodes_field
+      end
+
+      self.default_nodes_field = false
+
       # Create a connection which exposes edges of this type
-      def self.create_type(wrapped_type, edge_type: nil, edge_class: nil, nodes_field: false, &block)
+      def self.create_type(wrapped_type, edge_type: nil, edge_class: nil, nodes_field: ConnectionType.default_nodes_field, &block)
         edge_type ||= wrapped_type.edge_type
         edge_class ||= GraphQL::Relay::Edge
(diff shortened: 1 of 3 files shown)


rmosolgo:
Yeah, I think opt-in is a good choice, that way people who track changes via a schema printout don't get a bunch of noise when they update. Maybe in a later version, we can make this the default behavior and allow people to opt _out_.

josh:
> Yeah, I think opt-in is a good choice,

Revised to support `define_connection(nodes_field: true)`.

rmosolgo:
👌 good by me, but one more in case you want it: do you want a default, something like `GraphQL::Relay::ConnectionType.default_nodes_field`, so you can apply this to your whole schema at once? 

Happy to merge as is if you don't want it or you'd rather do it separately.

josh:
> do you want a default, something like GraphQL::Relay::ConnectionType.default_nodes_field

Oh! Great idea. Yep, let me get that in first.

josh:
Added `GraphQL::Relay::ConnectionType.default_nodes_field`. I guess its global for the entire environment. I didn't see a great way to scope it to a schema instance. Doesn't seem so bad though.

rmosolgo:
Good by me!!

josh:
❤️

PR #452, opened 2016-12-09
Bug with ArrayConnection

theorygeek:
This fixes a bug with the ArrayConnection. The failure condition is if the field resolver returns an array with N elements, and you then query against that connection with the starting offset of N+1.

```
2.3.1 :004 > x = (0...20).to_a
 => [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19] 
2.3.1 :005 > x[21..-1]
 => nil 
```

The `sliced_nodes` method ends up returning `nil`. And then `paged_nodes` tries to do:
```ruby
items.first(limit)
```
And it 💥  because of a null reference. This can happen if the connection _used_ to have plenty of elements in it, but now it doesn't anymore.

### DIFF
--- a/lib/graphql/relay/array_connection.rb
+++ b/lib/graphql/relay/array_connection.rb
@@ -32,7 +32,7 @@ module GraphQL
 
       # Apply cursors to edges
       def sliced_nodes
-        @sliced_nodes ||= nodes[starting_offset..-1]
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks! I added a couple of specs to make sure that the case was handled correctly.

PR #462, opened 2016-12-16
Optimize Query Execution

theorygeek:
This is a work in progress to try to optimize query execution. For the schema in my app, this eliminates about 73k object allocations for the introspection query.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -358,14 +358,6 @@ module GraphQL
       end
     end
 
-    def middleware_chain
-      @middleware_chain ||= begin
-        steps = middleware.dup
-        steps << GraphQL::Execution::Execute::FieldResolveStep
-        GraphQL::Schema::MiddlewareChain.new(steps: steps)
-      end
-    end
-
     protected
 
     def rescues?
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -60,7 +60,7 @@ module GraphQL
       directives: ->(schema, directives) { schema.directives = directives.reduce({}) { |m, d| m[d.name] = d; m  }},
       instrument: -> (schema, type, instrumenter) { schema.instrumenters[type] << instrumenter },
       query_analyzer: ->(schema, analyzer) { schema.query_analyzers << analyzer },
-      middleware: ->(schema, middleware) { schema.user_middleware << middleware },
+      middleware: ->(schema, middleware) { schema.middleware << middleware },
       lazy_resolve: ->(schema, lazy_class, lazy_value_method) { schema.lazy_methods.set(lazy_class, lazy_value_method) },
       rescue_from: ->(schema, err_class, &block) { schema.rescue_from(err_class, &block)}
 
@@ -69,8 +69,13 @@ module GraphQL
       :query_execution_strategy, :mutation_execution_strategy, :subscription_execution_strategy,
       :max_depth, :max_complexity,
       :orphan_types, :directives,
-      :query_analyzers, :user_middleware, :instrumenters, :lazy_methods,
+      :query_analyzers, :instrumenters, :lazy_methods,
       :cursor_encoder
+
+
+    # @return [MiddlewareChain] MiddlewareChain which is applied to fields during execution
+    attr_accessor :middleware
+
 
     class << self
       attr_accessor :default_execution_strategy
@@ -88,7 +93,7 @@ module GraphQL
       @orphan_types = []
       @directives = DIRECTIVES.reduce({}) { |m, d| m[d.name] = d; m }
       @static_validator = GraphQL::StaticValidation::Validator.new(schema: self)
-      @user_middleware = MiddlewareChain.new(final_step: GraphQL::Execution::Execute::FieldResolveStep)
+      @middleware = MiddlewareChain.new(final_step: GraphQL::Execution::Execute::FieldResolveStep)
       @query_analyzers = []
       @resolve_type_proc = nil
       @object_from_id_proc = nil
@@ -108,8 +113,7 @@ module GraphQL
       @orphan_types = other.orphan_types.dup
       @directives = other.directives.dup
       @static_validator = GraphQL::StaticValidation::Validator.new(schema: self)
-      @user_middleware = other.user_middleware.dup
-      @middleware = nil
+      @middleware = other.middleware.dup
       @query_analyzers = other.query_analyzers.dup
 
       @possible_types = GraphQL::Schema::PossibleTypes.new(self)
@@ -347,11 +351,6 @@ module GraphQL
     # @return [Boolean] True if this object should be lazily resolved
     def lazy?(obj)
       !!lazy_method_name(obj)
-    end
-
-    # @return [MiddlewareChain] MiddlewareChain which is applied to fields during execution
-    def middleware
-      @user_middleware
     end
 
     protected
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -358,6 +358,14 @@ module GraphQL
       end
     end
 
+    def middleware_chain
+      @middleware_chain ||= begin
+        steps = middleware.dup
+        steps << GraphQL::Execution::Execute::FieldResolveStep
+        GraphQL::Schema::MiddlewareChain.new(steps: steps)
+      end
+    end
+
     protected
 
     def rescues?
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -349,7 +349,7 @@ module GraphQL
       !!lazy_method_name(obj)
     end
 
-    # @return [MiddlewareChain] Middlewares suitable for MiddlewareChain, applied to fields during execution
+    # @return [MiddlewareChain] MiddlewareChain which is applied to fields during execution
     def middleware
       @user_middleware
     end
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -88,7 +88,7 @@ module GraphQL
       @orphan_types = []
       @directives = DIRECTIVES.reduce({}) { |m, d| m[d.name] = d; m }
       @static_validator = GraphQL::StaticValidation::Validator.new(schema: self)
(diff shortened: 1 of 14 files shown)


lib/graphql/schema.rb
```diff
@@ -358,6 +358,14 @@ def middleware
       end
     end
 
+    def middleware_chain
+      @middleware_chain ||= begin
+        steps = middleware.dup
+        steps << GraphQL::Execution::Execute::FieldResolveStep
+        GraphQL::Schema::MiddlewareChain.new(steps: steps)
+      end
```
theorygeek: If middleware can be changed _after_ the schema is defined, this variable would have to be cleared after each change.

theorygeek:
The next thing I'm looking at are all of the object allocations for FieldResolutionContext. I'm honestly not sure that this _could_ be reduced, unless the context were lazily spawned. 

One thing that has me a little puzzled is that the `FieldResolutionContext#spawn` method is doing almost zero work:
```ruby
def spawn(key:, selection:, parent_type:, field:)
  FieldResolutionContext.new(
    context: self,
    path: path + [key],
    selection: selection,
    parent_type: parent_type,
    field: field,
  )
end
```

Yet when I look at the profiler result:
image

The `Class#new` call accounts for 77% of the work inside of `#spawn`. Where is the other 23% going? The only work I can see is the `path + [key]` call.

rmosolgo:
:tada: this is awesome!! 

> If middleware can be changed after the schema is defined... 

One way to handle that is to build one chain per _query_. It's not quite as efficient -- 1 allocation instead of 0 -- but compared to 73k, it's a big win! Plus we don't have to figure out a compatibility plan. 

I think it is important to keep a compatibility plan because in previous versions of graphql, the _only_ way to add a middleware was to use `<<`, eg: 

```ruby
MySchema.middleware << something 
```

In fact, that's what my app still does! We could deprecate it, but I don't think we should break it right away. (It _probably_ wouldn't be broken -- I doubt people add and remove middlewares in between queries!)

It's doable though: you could have `Schema#middleware` be the _chain itself_, so `MiddlewareChain#<<`  could handle the added middleware.

lib/graphql/schema/middleware_chain.rb
```diff
@@ -8,21 +8,15 @@ class MiddlewareChain
...
 
       # Run the next step in the chain, passing in arguments and handle to the next step
-      def call(next_arguments = @arguments)
-        @arguments = next_arguments
-        next_step = steps.shift
-        next_middleware = self
-        next_step.call(*arguments, next_middleware)
+      def invoke(index, arguments)
+        next_step = steps[index]
+        call_next = ->(next_args = arguments) { invoke(index + 1, next_args) }
```
rmosolgo: Due to the [overhead of function literals](https://bugs.ruby-lang.org/issues/12901), I think we'd be better off with a callable object here, eg: 

```ruby 
module CallNext
  def initialize(chain, next_args, next_offset)
    @chain = chain 
    @next_args = next_args 
    @next_offset = next_offset 
  end 
  def call(next_args = @next_args) 
    @chain.call(@next_offset, next_args) 
  end 
end 

# ... 

call_next = CallNext.new(self, next_args, index + 1)
```

Ever since I learned about that overhead, I've been trying to keep them out of performance-sensitive code! What do you think, would that work here?

theorygeek:
>One way to handle that is to build one chain per _query_.

I moved the middleware chain method to the query context, since that was easiest for now.

I did some benchmarks running the introspection query on the DummySchema. Here's what I got running against [master](https://github.com/rmosolgo/graphql-ruby/commit/5c1269d594ba0150ac67d32aa1e0f953c1daf4ef):
```
               query     29.525  (± 3.4%) i/s -    148.000  in   5.019254s
               query     29.868  (± 3.3%) i/s -    150.000  in   5.029728s
               query     29.361  (± 6.8%) i/s -    148.000  in   5.059944s
```

Running with these changes:
```
               query     32.367  (± 3.1%) i/s -    162.000  in   5.010069s
               query     32.273  (± 3.1%) i/s -    162.000  in   5.022943s
               query     32.299  (± 3.1%) i/s -    162.000  in   5.021201s
```

The interesting thing is that this code _is_ faster, but it actually _increases_ the number of object allocations. In my app, running the introspection query against the original code showed 230,899 calls to `Class#new`.

This code shows 525,168 (more than double!) calls to `Class#new`:
image

Locally, I removed the two commits related to the Warden, to see if the improvements really _were_ because of the middleware changes. And the middleware _looks_ like it really is helping:
```
               query     32.187  (± 3.1%) i/s -    162.000  in   5.038905s
               query     31.564  (± 3.2%) i/s -    159.000  in   5.045532s
               query     31.058  (± 6.4%) i/s -    156.000  in   5.036511s
```

theorygeek:
So I did an experiment where middlewares accept a _block_ instead of a callable object, and it shows the most significant speed improvement so far:
```
               query     36.701  (± 2.7%) i/s -    186.000  in   5.071567s
               query     36.968  (± 2.7%) i/s -    186.000  in   5.037552s
               query     36.166  (± 5.5%) i/s -    183.000  in   5.070882s
```

The other nice advantage is that it avoids all of the extra object allocations as well:
image

The downside, of course, is that it's a breaking change. But fortunately it's a relatively easy one to live with, you just have to put an `&` in front of the `next_middleware` argument:
```ruby
def call(graphql_type, object, field_definition, args, context, &next_middleware)
  next_middleware.call
end
```

Or switch to using the `yield` statement instead:
```ruby
def call(graphql_type, object, field_definition, args, context)
  yield
end
```

rmosolgo:
Thanks so much for reporting your findings! I find this stuff fascinating. 

>  increases the number of object allocations

I wonder about that, because you'd think, it has to allocate _something_, even if it's a `Proc`. But it may do so without a call to `Proc.new`. 

Digging deeper, I see there are some cases Ruby [reuses a proc somehow](https://github.com/ruby/ruby/blob/2a546caae63960a269cb58731e61fea4c7e9797c/proc.c#L668) (although what's actually going on there is beyond me!)

[memory_profiler](https://github.com/SamSaffron/memory_profiler) might provide another angle (it counts object allocations, not just calls to `Class.new`), but it's not worth it in my opinion. 

> block instead of a callable object

Very nice! I always heard it was faster to `yield`, but I didn't know it was still faster if you captured the block with `&`. 

> it's a breaking change

This can be helped: 

- When building the MiddlewareChain, inspect each step 
- If its arity is 6, wrap it with an intermediary object that: 
  - emits a warning from `#initialize`
  - in `#call`, it captures the block and passes it to the "inner" step as a callable 

(PS I learned the hard way that [arity checks are special for Procs](https://github.com/rmosolgo/graphql-ruby/blob/80fd507d9e56692c05e23db615e6333389752a08/lib/graphql/relay/mutation.rb#L137-L144)!)

That way, all the steps of the MiddlewareChain have arity 5. If a user provides a step with arity of 6, we wrap it with an adapter object and suggest that they update for a perf win!

theorygeek:
>Very nice! I always heard it was faster to `yield`, but I didn't know it was still faster if you captured the block with `&`.

I actually don't _know_ that it's faster if you capture the block with `&`, because I updated the middleware in the gem (eg, RescueMiddleware, TimeoutMiddleware) to use ordinary `yield`'s. I _suspect_ that it's still faster, I just don't know for sure.

Keep in mind that the metrics I posted here are showing the introspection query against the dairy app schema. I did also time the introspection query against my app's schema (which has 4 middlewares), and it's showing big improvements (~ 30%) with the block. But again, I updated it to use `yield`.

>it counts object allocations, not just calls to `Class.new`
I should have been more specific in saying that the number of _class_ allocations increased, rather than _object_ allocations. I think neither proc's, arrays, nor hashes will show up as calls to `Class.new`.

I'll make the change with the arity checks, clean up the code a bit, and then call this PR good to go. I was hoping to also take a stab at the field resolution contexts, but I think that's going to require a bigger change.

rmosolgo:
👏  sounds like a great plan, ping me when it's ready!!

theorygeek:
@rmosolgo Alright, I think this is ready to go. Summary of changes, compared to master:
- Warden caching (I think you could explain this better than me)
- Middleware chain is created once per schema, instead of once per field, and it's faster
- Middlewares should accept a block and `yield` instead of accepting a `next_middleware` argument

The code attempts to detect middleware that expects the `next_middleware` argument, wrap it, and provide a deprecation warning. But it won't work for middleware whose signature looks like:
```ruby
def call(*args)
```

lib/graphql/schema.rb
```diff
@@ -349,13 +349,9 @@ def lazy?(obj)
...
 
-    # @return [Array<#call>] Middlewares suitable for MiddlewareChain, applied to fields during execution
+    # @return [MiddlewareChain] MiddlewareChain which is applied to fields during execution
     def middleware
-      @middleware ||= if @rescue_middleware
-        [@rescue_middleware].concat(@user_middleware)
-      else
-        @user_middleware
-      end
+      @user_middleware
```
rmosolgo: I remember this song and dance had something to do with `#redefine`, making sure that middlewares get duped correctly. 

Do you know of this handles in properly? (I can't remember if I wrote a proper test for it 😬 , I'll take a look tonight!)

theorygeek: Is this code testing that behavior? 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/schema_spec.rb#L302:
```ruby
it "copies internal state" do
  schema_2 = schema.dup
  refute schema_2.types.equal?(schema.types)

  refute schema_2.instrumenters.equal?(schema.instrumenters)
  assert_equal schema_2.instrumenters, schema.instrumenters

  refute schema_2.middleware.equal?(schema.middleware)
  assert_equal schema_2.middleware, schema.middleware
end
```

rmosolgo: Ah, that test was only good enough when it was a plain array. I added another spec in 0bb022b6c736c6f01ccf2f3199b1e005c1455b24

rmosolgo:
:tada: :clap: !!!

kdaigle:
cc @charliesome

PR #463, opened 2016-12-18
Schema filter improvements

rmosolgo:
This adds some oomph to dynamic schema filtering feature: 

- Pass `ctx` to filter function (`.call(member, ctx)`) 
- Accept whitelist filters with `only:` (previously, only blacklists accepted via `except:`) 
- Add `Schema#default_mask` as default blacklist function 
- Add `Schema#to_definition` which prints the definition and accepts filters (`only:`/`except:`) and a `context:`. This allows you to preview how your runtime filters will be applied.


rmosolgo:
(oops, fixed title :P )

lib/graphql/query.rb
```diff
@@ -38,11 +30,19 @@ def call(member)
...
-    def initialize(schema, query_string = nil, document: nil, context: nil, variables: {}, validate: true, operation_name: nil, root_value: nil, max_depth: nil, max_complexity: nil, except: NullExcept)
+    def initialize(schema, query_string = nil, document: nil, context: nil, variables: {}, validate: true, operation_name: nil, root_value: nil, max_depth: nil, max_complexity: nil, except: nil, only: nil)
       fail ArgumentError, "a query string or document is required" unless query_string || document
 
       @schema = schema
-      @warden = GraphQL::Schema::Warden.new(schema, except)
+      mask = if except
+        wrap_if_legacy_mask(except)
+      elsif only
+        InvertedMask.new(wrap_if_legacy_mask(only))
```
brandonblack: Thank you for doing this. 👍 We ended up inverting the `except` logic in GitHub's implementation since it was easier to reason with and fit the model we wanted better.

lib/graphql/schema/printer.rb
```diff
@@ -11,35 +11,76 @@ module Printer
       extend self
 
       # Return a GraphQL schema string for the defined types in the schema
+      # @param context [Hash]
+      # @param only [<#call(member, ctx)>]
+      # @param except [<#call(member, ctx)>]
       # @param schema [GraphQL::Schema]
-      def print_schema(schema)
-        print_filtered_schema(schema, lambda { |n| !is_spec_directive(n) }, method(:is_defined_type))
+      def print_schema(schema, context: nil, only: nil, except: nil)
```
brandonblack: Thanks so much for taking care of this as well. ❤️

rmosolgo:
Great point about built-in types. I think a handful of methods on `BaseType` might be a start in the right direction: 

- `BaseType#introspection?`
- `BaseType#default_scalar?`
- `BaseType#default_relay?`

I think those would really simplify the filters in this PR, too

rmosolgo:
I'm going to add those new methods in another branch

vergenzt:
@brandonblack I just wanted to add one qualification to what you said [earlier](https://github.com/rmosolgo/graphql-ruby/pull/463#pullrequestreview-14251927):

> ```
> DEFAULT_TYPES = {
>   ...
>   "Int"                 => GraphQL::INT_TYPE,
>   "String"              => GraphQL::STRING_TYPE,
>   "Boolean"             => GraphQL::BOOLEAN_TYPE,
>   ...
> }
> ```
>
> These are types we include in all versions of the schema and in our custom mask, we [should] return a pass on the filter check when `member` is one of these types.
>
> ...
>
> When it comes to building these masks, I think most folks won't care about the built-in types and will end up doing something similar to what we're doing.

For what it's worth I definitely care about the ability to mask string, int and boolean types. Those could be anything. :)

Huge +1 to @rmosolgo's approach of just including helper methods on the types to detect when they're builtin, introspective, etc. I'm currently trying to implement a role-based schema whitelist and am running into the problem that the introspective queries aren't allowed, so I can't explore in `GraphiQL` what my filtered schema actually looks like! I'd love to be able to simply `return true if schema_member.introspective?` or something like that.

rmosolgo:
A [handful of those methods](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L37-L50) were added in https://github.com/rmosolgo/graphql-ruby/pull/473, please open an issue if there are others that you think would be useful!

vergenzt:
Yeah sorry, I found that and edited. :) After playing with the methods for a few hours though I ended up opening #544 because I found I couldn't do everything I needed to.

PR #473, opened 2017-01-03
Type reflections

rmosolgo:
When metaprogramming a GraphQL schema, it comes in handy to check for special types: introspection types, built-in scalars and built-in relay types. 

Now, `BaseType` responds to checks for those things. 

Additionally, built-in directives are flagged as such. (There's not really a way to add directives from user-space yet, but ... someday 😎 .)

### DIFF
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -6,6 +6,7 @@ module GraphQL
     include GraphQL::Define::InstanceDefinable
     accepts_definitions :name, :description,
         :introspection,
+        :default_scalar,
         {
           connection: GraphQL::Define::AssignConnection,
           global_id_field: GraphQL::Define::AssignGlobalIdField,
@@ -19,6 +20,7 @@ module GraphQL
       @connection_type = nil
       @edge_type = nil
       @introspection = false
+      @default_scalar = false
     end
 
     # @return [String] the name of this type, must be unique within a Schema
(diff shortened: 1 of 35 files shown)


lib/graphql/directive.rb
```diff
@@ -8,10 +8,12 @@ module GraphQL
   #
   class Directive
     include GraphQL::Define::InstanceDefinable
-    accepts_definitions :locations, :name, :description, :arguments, argument: GraphQL::Define::AssignArgument
+    accepts_definitions :locations, :name, :description, :arguments, :default, argument: GraphQL::Define::AssignArgument
```
brandonblack: For the sake of readability, what do you think about calling this `default_directive`? It has more symmetry with the rest of these changes and seeing things like `default = false` in isolation might be confusing.

lib/graphql/schema/printer.rb
```diff
@@ -42,34 +42,24 @@ def print_introspection_schema
...
-
-      BUILTIN_SCALARS = [
-        GraphQL::STRING_TYPE,
-        GraphQL::BOOLEAN_TYPE,
-        GraphQL::INT_TYPE,
-        GraphQL::FLOAT_TYPE,
-        GraphQL::ID_TYPE,
-      ]
-
       # By default, these are included in a schema printout
```
brandonblack: Huge ❤️  for making this change in the printer as well.

lib/graphql/base_type.rb
```diff
@@ -24,6 +34,24 @@ def initialize_copy(other)
     # @return [String, nil] a description for this type
     attr_accessor :description
 
+    # @return [Boolean] Is this type a predefined introspection type?
+    def introspection?
```
cjoudrey: We may be able to change this https://github.com/rmosolgo/graphql-ruby/blob/895d8b7c198d3db6d290bcc95f37e3767ab6b5f5/lib/graphql/schema/validation.rb#L181 now that we have `introspection?`.

rmosolgo: 💰  i'll give it a try

lib/graphql/directive.rb
rmosolgo: Oh yeah, that's much more consistent!

lib/graphql/base_type.rb
rmosolgo: 👌 donezo

rmosolgo:
Thanks for taking a look at this!

PR #476, opened 2017-01-09
Fix grouped connection with only `last` argument specified

sjchmiela:
If a relation contains a `.group` clause, a `.count` will return a `Hash`. So to know a connection length we have to count also the `Hash`'s content.

_Another solution would be to use `.length` instead of `.count`, but that would not ask the databse, but load all the records into the memory, which… is bad._

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -17,7 +17,7 @@ module GraphQL
       end
 
       def has_next_page
-        !!(first && sliced_nodes.limit(limit + 1).count > limit)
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks! This looks good to me but it needs a spec to make sure we don't break it in the future. 

I think we need to add another `connection` in `star_wars_schema.rb` which returns a grouped relation, then test this code by executing a query which hits that new connection and validates the response. 

Do you want to take a crack at adding a test here?

sjchmiela:
Check added. :)

rmosolgo:
You rock, thanks! I just added a commit to split the query string into two parts, to make this new test a little more independent.

PR #478, opened 2017-01-09
Provide default docs for connections

gjtorikian:


lib/graphql/relay/connection_type.rb
```diff
@@ -7,9 +7,11 @@ def self.create_type(wrapped_type, edge_type: nil, edge_class: nil, &block)
         edge_type ||= wrapped_type.edge_type
         edge_class ||= GraphQL::Relay::Edge
         connection_type_name = "#{wrapped_type.name}Connection"
+        description "The connection model for #{wrapped_type.name}."
```
rmosolgo: Should this be "The connection __type__ ..." ?

Since this is not inside `.define { ... }`, this should probably be 

```ruby
connection_type_description = "..."
```

right?

gjtorikian: >  this should probably be

😊 

I swear I know Ruby.

gjtorikian: Fixed!

rmosolgo:
:metal: sweet!

PR #479, opened 2017-01-11
fix(Query::Variables) don't assign nil for absent variables

### LINKED ISSUE #472: args hash contains a key which wasn't part of mutation variables
As [requested](https://graphql.slack.com/archives/ruby/p1483405917000724) by @rmosolgo, opening a bug about this.

In my app, I have an update form. I have that form set up to submit a mutation with `variables` consisting of only the atttributes that have actually changed. Here's a simple example of that query, though you could imagine it having many more arguments:

```graphql
mutation updateObject($id: IDType!, $name: String, $enabled: Boolean) {
  updateObject(id: $id, name: $name, enabled: $enabled) {
    ...
  }
}
```

If I toggle `enabled` on my form, this is the JSON body of the mutation request:

```json
{"query":"mutation updateWatcher($id: IDType!, $name: String, $enabled: Boolean) {
  updateWatcher(id: $id, name: $name, enabled: $enabled) {
    watcher {
      id
      name
      enabled
      __typename
    }
    __typename
  }
}
","variables":{"enabled":true,"id":"27tlpoPeXm6odAxj3paGQP"},"operationName":"updateWatcher"}
```

In particular, note that `variables` is `{"enabled":true,"id":"27tlpoPeXm6odAxj3paGQP"}`. The `name` attr was not changed, and thus was not provided in the `variables`.

However, on my graphql-ruby server, this same request shows the following args:

```ruby
[2] pry(#<UpdateWatcherResolver>)> args.to_h
=> {"id"=>"27tlpoPeXm6odAxj3paGQP", "name"=>nil, "schedule_every"=>60}
[3] pry(#<UpdateWatcherResolver>)> args.key?("name")
=> true
[4] pry(#<UpdateWatcherResolver>)> args.key?(:name)
=> true
[5] pry(#<UpdateWatcherResolver>)> args.class
=> GraphQL::Query::Arguments
```

Note the `"name" => nil` bit. I did not provide a `name` key in my `variables`, but graphql-ruby added it to the args hash, presumably because the variable was referenced in the mutation itself.

I wasn't sure if this was the correct behavior, and @rmosolgo seems to think it might be a bug.

What I'd like to be able to do is have a single statically-defined query in the frontend that includes all the variables I *might* update for that object on that specific mutation. I'd like that one query to work as a PATCH-style update for any valid combination of `variables` and have my backend only see the vars that I actually provided as `variables`.

I'm going to try using an input object for now, and hopefully that gets me the behavior I want.

rmosolgo:
Fixes #472 ... I think :)

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -319,7 +319,7 @@ module GraphQL
         scalar_attributes :name, :type, :default_value
 
         # @!attribute default_value
-        #   @return [String, Integer, Float, Boolean, Array] A Ruby value to use if no other value is provided
+        #   @return [String, Integer, Float, Boolean, Array, NullValue] A Ruby value to use if no other value is provided
 
         # @!attribute type
         #   @return [TypeName, NonNullType, ListType] The expected type of this value
--- a/lib/graphql/query/variables.rb
+++ b/lib/graphql/query/variables.rb
@@ -14,38 +14,31 @@ module GraphQL
         @provided_variables = provided_variables
         @errors = []
         @storage = ast_variables.each_with_object({}) do |ast_variable, memo|
+          # Find the right value for this variable:
+          # - First, use the value provided at runtime
+          # - Then, fall back to the default value from the query string
+          # If it's still nil, raise an error if it's required.
+          variable_type = @schema.type_from_ast(ast_variable.type)
           variable_name = ast_variable.name
-          memo[variable_name] = get_graphql_value(ast_variable)
+          default_value = ast_variable.default_value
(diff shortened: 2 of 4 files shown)


bgentry:
Looks great afaict! :v:

cjoudrey:
:+1: Makes sense!

While poking at the issue, I wrote this test which you might want to cherry-pick: https://github.com/cjoudrey/graphql-ruby/commit/6bd4da2d582f13e01a0346b885dff45770fc727a

It's unrelated to this issue, but related to null values.

rmosolgo:
Thanks for the free test case, cherry-picked!

PR #480, opened 2017-01-11
fix(Warden) also support hiding root types

rmosolgo:
Oops, @bgentry discovered that hiding _root_ types didn't quite work right!

Now, `root_type_for_operation` is correctly filtered.

### DIFF
--- a/lib/graphql/introspection/schema_type.rb
+++ b/lib/graphql/introspection/schema_type.rb
@@ -10,15 +10,15 @@ GraphQL::Introspection::SchemaType = GraphQL::ObjectType.define do
   end
 
   field :queryType, !GraphQL::Introspection::TypeType, "The type that query operations will be rooted at." do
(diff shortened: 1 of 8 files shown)


bgentry:
Awesome! :clap:

The response now looks like:

```json
{
  "errors": [
    {
      "message": "Schema is not configured for mutations",
      "locations": [
        {
          "line": 1,
          "column": 1
        }
      ],
      "fields": [
        "mutation myMutation"
      ]
    },
    {
      "message": "Variable $name is declared by myMutation but not used",
      "locations": [
        {
          "line": 1,
          "column": 1
        }
      ],
      "fields": [
        "mutation myMutation"
      ]
    }
  ]
}
```

I also confirmed that my mask is only being called once when this rejection happens.

rmosolgo:
Sweet!

PR #486, opened 2017-01-15
docs(filters) improve schema filtering documentation

### LINKED ISSUE #481: two terms (filter + mask) for the same concept
After upgrading to try out #480, I ran into a new deprecation notice:

```
.Schema.execute(..., except:) filters now accept two arguments, `(member, ctx)`. One-argument filters are deprecated.
```

It seems that the docs only use the term "mask" to describe these, but the deprecation notice and lots of internals call them "filters". If I'm correct that these are the same thing, maybe it makes sense to consolidate on a single term so it's less confusing?

rmosolgo:
Fixes #481 I hope :)


bgentry:
Much better, thanks! 🙌🏻

PR #492, opened 2017-01-16
docs(types_and_fields) Fix api_doc usage

swalkinshaw:
This was being rendered with the HTML of a link instead of an actual link.


rmosolgo:
oops, thanks fixing that!

PR #493, opened 2017-01-17
Fix documentation for Query#initialize "except" keyword argument

cjoudrey:
As per [this warning](https://github.com/cjoudrey/graphql-ruby/blob/40da6dc463bb0efc04574a92225eadfb6e6d3011/lib/graphql/query.rb#L302), `except` method should accept two arguments. This is just a quick change to the docs to reflect this.

As a side note, I noticed [rubydoc](http://www.rubydoc.info/github/rmosolgo/graphql-ruby/file/CHANGELOG.md) is outdated. Is there anything we need to do to force a rebuild of the docs?

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -32,6 +32,7 @@ module GraphQL
     # @param max_depth [Numeric] the maximum number of nested selections allowed for this query (falls back to schema-level value)
     # @param max_complexity [Numeric] the maximum field complexity for this query (falls back to schema-level value)
     # @param except [<#call(schema_member, context)>] If provided, objects will be hidden from the schema when `.call(schema_member, context)` returns truthy
+    # @param only [<#call(schema_member, context)>] If provided, objects will be hidden from the schema when `.call(schema_member, context)` returns false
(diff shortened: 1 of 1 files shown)


rmosolgo:
To refresh the github-based docs, you can click this little swirly icon in the search results: 

image

(I've just clicked it to make sure it did what I remembered it doing :P) 

The rubygems-based docs have the same problem, different pages point to different versions of the gem (1.3.0, 1.4.0, 1.4.1): 

image

image


image

But I don't know how to fix that one 😖  There's no refresh button: 

image

cjoudrey:
> To refresh the github-based docs, you can click this little swirly icon in the search results

Interesting! I didn't know that.

I did a bit of Googling and looks like we can automate all this:

screen shot 2017-01-17 at 3 49 20 pm

:tada:!

rmosolgo:
Added! 

I'll merge this PR in a bit. I see that `only` is also missing and I'd like to use this open PR as a reminder to add it :)

cjoudrey:
Good catch, I hadn't noticed. I'll add it to the PR.

rmosolgo:
Thanks, that works too!!

PR #495, opened 2017-01-19
Add Schema#validate

rmosolgo:
Make it less awkward to validate a query without running it. I think graphql-client wants this, and I want it for benchmarking.


rmosolgo:
Sad, this PR's target was `node-path`, not master 😩

PR #496, opened 2017-01-19
feat(AbstractNode) add #path to get AST path

rmosolgo:
Make AST nodes aware of their parents, so you can ask them for their `#path`. Then, refactor StaticValidation to use that `#path`. (This allows getting the path sometime _later_, after the first validation pass. You can see some areas where `context.path` was cached for later usage; that's not needed anymore.)

This has some potetially-breaking changes: 

- `"fields"` on validation error has changed: 
  - Previously, directives were not included. Now they are. 
  - For some validations on arguments, previously, the argument itself was not present in `"fields"`, now it is 
  - Previously, variable identifiers were not present in `"fields"`, now they are. 
  - I guess you could say "fields" is more of a "path" now, but `"path"` is used for runtime errors
- For some validations on arguments, the error location now points at the _argument_ node, not the _field_ (owner) node
- The validation `message(...)` helper's call signature has changed: kwargs `context` and `path` have been removed, since that information is now available from the AST nodes.

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -10,11 +10,13 @@ module GraphQL
       # - `to_query_string` turns an AST node into a GraphQL string
       class AbstractNode
         attr_accessor :line, :col
+        attr_accessor :parent
 
         # Initialize a node by extracting its position,
         # then calling the class's `initialize_node` method.
         # @param options [Hash] Initial attributes for this node
         def initialize(options={})
+          @parent = nil
           if options.key?(:position_source)
             position_source = options.delete(:position_source)
             @line, @col = position_source.line_and_column
@@ -28,13 +30,34 @@ module GraphQL
           raise NotImplementedError
         end
 
+        def path
+          # TODO: someday this could be a bug.
+          # If you move a node someplace else in the AST,
+          # this cached value will become stale.
+          @path ||= begin
+            parent_path = @parent ? @parent.path : []
+            new_path = parent_path.dup
+            (key = path_key) && new_path.push(key)
+            new_path
+          end
+        end
+
+        def path_key
+          nil
+        end
+
+        def assign_parent_to_each(nodes)
+          nodes.each { |n| n.parent = self }
+        end
+
         # Value equality
         # @return [Boolean] True if `self` is equivalent to `other`
         def eql?(other)
-          return true if equal?(other)
-          other.is_a?(self.class) &&
+          equal?(other) || (
+            other.is_a?(self.class) &&
             other.scalars.eql?(self.scalars) &&
             other.children.eql?(self.children)
+          )
         end
 
         # @return [Array<GraphQL::Language::Nodes::AbstractNode>] all nodes in the tree below this one
@@ -98,6 +121,7 @@ module GraphQL
       class NameOnlyNode < AbstractNode
         attr_accessor :name
         scalar_attributes :name
+        alias :path_key :name
 
         def initialize_node(name: nil)
           @name = name
@@ -112,6 +136,7 @@ module GraphQL
       class Argument < AbstractNode
         attr_accessor :name, :value
         scalar_attributes :name, :value
+        alias :path_key :name
 
         # @!attribute name
         #   @return [String] the key for this argument
@@ -122,6 +147,7 @@ module GraphQL
         def initialize_node(name: nil, value: nil)
           @name = name
           @value = value
+          value.is_a?(InputObject) && (value.parent = self)
         end
 
         def children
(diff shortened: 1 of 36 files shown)


rmosolgo:
@josh Do these changes to StaticValidation look ok to you? It was easiest to just allow these changes to happen because I don't _know_ of anything that depends on them, what do you think?

josh:
I'd probably say I'm not in favor of this. The big issue is that is prevents shallow copies and sharing AST nodes between multiple trees. I like'd that I could treat AST nodes as a kinda-persistent-data-structure.

rmosolgo:
Is that just an implementation concern? What if I removed the caching?

josh:
> Is that just an implementation concern?

It creates an API inconsistency. Theres now an *undefined behavior* when AST nodes are created without or an incorrect a `parent` reference.

For an example, if an AST tree is created by hand or by some other means (aka graphql-client), and that AST is passed to the query interface, is it going to bomb on `parent` node being `nil`?

rmosolgo:
Ohhh right, if you build a Document in some way other than `GraphQL.parse`, then you're totally hosed. I forgot about that! And in those cases, the same node might have different `path`s in two different trees.

Thanks for taking a look. I'll see if I can refactor these other changes so they don't depend on this one.

rmosolgo:
I think I'll be fine without ☺️

ianks:
@rmosolgo @josh (nice snag on the username btw!)

This feature would be incredibly useful for something I am attempting to do. I would like to compile the AST into a simple SQL style string. In order to know which "table" I would need to select from, I have to access to direct parent of a given AST node. I can't think off the top of my head a way to do this currently... This might be better explained by an

## Example

Given you have a model like this:

```ruby
module Types
  Post = GraphQL::ObjectType.define do
    field :insights do
      type types[InsightsType]

      argument :time_ranges, !types[TimeRangeType]

      resolve ->(obj, args, ctx) {
        # have access to the AST node, but no way to get information about the parent :/
        current_node = ctx.ast_node

        # would like to be able to do something like
        compile_to_sql(ast_node.parent) # walk the tree, do things
      }
    end
  end
end
```

Given some query like this:

```graphql
{
  post(id: "123") {
    insights(time_range: { start: "2016-01-01", end: "2017-01-01" }) {
      views,
      likes
    }
}
```

It does not seem possible to generate something like, as we do not have access to the parent of the `insights` node.

```sql
SELECT views, likes FROM POST_INSIGHTS_REPORT DURING "2016-01-01,2017-01-01
```

## A note

As a quick note, a solution for this would not require a full path, just a property `parent` on every `GraphQL::Language::Node`. @josh would this still prevent shallow copies and sharing AST nodes between multiple trees?

ianks:
Update: I found a (suboptimal) solution to my current problem. I can do something like this to find the parent. However, it requires an unneccesary traversal of the tree.

```ruby
visitor = GraphQL::Language::Visitor.new(ctx.query.document)
visitor[GraphQL::Language::Nodes::Field] << ->(node, parent) {
  if node == ctx.ast_node
    # bingo!
  end
}
visitor.visit
```

rmosolgo:
A couple of thoughts: 


- Depending on your control over incoming query strings, using the AST directly is not an easy way to process the query structure. For example, inside fragment definitions, the `parent` will be a fragment definition, not another field: 

  ```graphql 
  query GetPost($id: ID) {
    post(id: $id) {
      id 
      ...PostFields
    }
  }

  fragment PostFields on Post { 
    title
  }
  ```

  `id`'s parent will be `post(id: $id)`, but `title`'s parent will be `fragment PostFields on Post`. 

  I use `InternalRepresentation::Node`s as a normalized query tree. It is "flattened" to handle fragments. You can access the current irep node with `ctx.irep_node`. The only downside is that the API is a bit unstable: I'm still figuring out how to use that rewritten tree most effectively, so it occasionally changes.

  However, if you have tight control over the incoming query strings, maybe you can be sure that this fragment issue will never arise!
- ~~You _could_ keep a local stack of nodes in `ctx`~~ Nevermind, that doesn't work because there's no way to pop previous nodes when exiting a field 😖 
- Another place to traverse the tree is the [query analysis step](http://rmosolgo.github.io/graphql-ruby/queries/authorization#query-analyzers) -- you might be able to gather the metadata and put it into `ctx`, then execute the database query based on that metadata in your root field. Again, not sure if that fits your case, just a thought!

Hope some of that helps! I think this kind of usage of GraphQL is really interesting so I'd love to hear what you find.

ianks:
@rmosolgo Thank you so much for the information here! It will probably save me some serious headaches :)

I will report back once I get something working.

Issue #508, opened 2017-01-25
Negative offset breaks/crashes PostgreSQL


b8in:
Hello guys!

I've recently noticed that there's an issue:
PG::InvalidRowCountInResultOffsetClause: ERROR: OFFSET must not be negative

It appears here: http://joxi.ru/bmoW8eMhxygJEm when the value of param 'last' is more then the number of records in the appropriate table. I used ActiveRecord ORM.
I tried to write a test for that case, but passes flawlessly. I think it's because it's SQLite, not PostgreSQL.

I fixed this issue locally like this: http://joxi.ru/KAg8OD1F4KPl5r

P.S. If needed, I can leave a PR.

Thanks, would be great to hear your thoughts on this!
Alex

rmosolgo:
Thanks for the details on this error! A PR would be great.  

Funny about the difference between SQLite and PostgreSQL, I didn't know that.

b8in:
PR #510

b8in:
#510

rmosolgo:
Fixed by 508 👍

karki011:
I keep running to this error, is there a solution for this.

I tried to look at past solution but the link is not found.

ActiveRecord::StatementInvalid · PG::InvalidRowCountInResultOffsetClause: ERROR: OFFSET must not be negative : SELECT DISTINCT "positions".* FROM "positions" WHERE "positions"."deleted_at" IS NULL ORDER BY "positions"."id" DESC LIMIT 10 OFFSET -10

Any help would be helpful.

thanks

PR #510, opened 2017-01-26
Fix negative offset

b8in:
Hey @rmosolgo ,

Here's my PR: 

I added a couple of tests and included them into a separate commit, please have a look at them, if they are needed or not - you can exclude that commit.

Regards,
Alex


b8in:
#508

rmosolgo:
This is great, thank you!!

PR #512, opened 2017-01-27
refactor(Rewrite) flatten selections on object types ahead of time

rmosolgo:
I'm hoping this will help solve #418, but gotta make sure: 

- [x] Actually can fix the problem 
- [x] Isn't slower 
- [x] Make the code not really ugly
- [x] Maintain compatibility on irep_node interface


rmosolgo:
It doesn't seem any slower, here's the benchmark of `Schema.validate` against master: 

```
               query    272.653  (± 3.7%) i/s -      1.377k in   5.057757s
               query    282.764  (± 3.2%) i/s -      1.431k in   5.065801s
               query    268.418  (± 3.7%) i/s -      1.352k in   5.044627s
```

And on this branch:

```
               query    290.411  (± 7.6%) i/s -      1.456k in   5.052828s
               query    293.638  (± 3.4%) i/s -      1.484k in   5.059482s
               query    295.286  (± 3.7%) i/s -      1.479k in   5.015950s
```

rmosolgo:
Quick port of fields-will-merge resulted in a big regression: 

```
               query    143.126  (± 4.2%) i/s -    715.000  in   5.003695s
               query    136.824  (± 4.4%) i/s -    689.000  in   5.046835s
               query    139.750  (± 5.0%) i/s -    702.000  in   5.036222s
```

I'll look into why! Lots of leads in the profile result.

rmosolgo:
:metal: Removing the need for `deep_copy` did the trick: 

```
               query    350.029  (± 3.7%) i/s -      1.768k in   5.057584s
               query    335.753  (± 7.4%) i/s -      1.700k in   5.095224s
               query    344.055  (± 3.5%) i/s -      1.734k in   5.045805s
```

rmosolgo:
There are breaking changes to InternalRepresentation::Node: 

- Buggy, deprecated `#path` was removed (as was `#index=` / `#index` and `#parent`)
- `#directives` was removed, you can check `ast_nodes` and `ast_spreads` instead 
- Buggy, deprecated `#children` was removed

gjtorikian:
Was it intentional for `parent` to be removed? As it stands, given an `irep_node`, it seems that one can only go down (via `typed_children`), instead of up as well.

rmosolgo:
I think it was a mistake ... I guess I didn't need it for anything so I forgot to add it back, but I didn't document the remove in CHANGELOG so I probably just missed it! Do you want it back?

gjtorikian:
I do. 🙈  I'm in the process of upgrading and we have a query analyzer that checks a field's parent as well as its children. I could work around it but I wasn't sure if `parent` was removed for perf concerns or what.

rmosolgo:
No, I can't imagine it's a perf problem at all, it only costs a couple of hash table entries! Let me see about putting it back...

PR #513, opened 2017-01-27
Add Schema#validate

rmosolgo:
Because #495 was into the wrong base branch 😬


arathunku:
@rmosolgo https://github.com/blog/2224-change-the-base-branch-of-a-pull-request :D

rmosolgo:
But that's only for _before_ you merge it, right? I merged #495 a few days ago, then today tried to benchmark `Schema#validate` on master and ... no worky!

PR #516, opened 2017-02-02
Set UTF-8 encoding other parsed tokens

josh:
So I was trying to track down why most AST `Language::Node`s had UTF-8 tagged strings and why a few others were untagged as binary.

Stuff like `Language::Node#name` gets surfaced in different ways on introspection queries. This leads to some binary tagged strings leaking into execution result.

I figured fixing the encoding at the source would be the best way to solve this. It looks like some of these strings are already being retagged after they are repacked. https://github.com/rmosolgo/graphql-ruby/blob/571ceab99f3076f269a3d92cc133d559151e3bfd/lib/graphql/language/lexer.rl#L190

Does the parser even support parsing input strings in anything else besides UTF-8?

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -990,7 +990,7 @@ end
       def self.record_comment(ts, te, meta)
         token = GraphQL::Language::Token.new(
           name: :COMMENT,
-          value: meta[:data][ts...te].pack("c*"),
+          value: meta[:data][ts...te].pack("c*").force_encoding("UTF-8"),
           line: meta[:line],
           col: meta[:col],
           prev_token: meta[:previous_token],
@@ -1004,7 +1004,7 @@ end
       def self.emit(token_name, ts, te, meta)
         meta[:tokens] << token = GraphQL::Language::Token.new(
           name: token_name,
-          value: meta[:data][ts...te].pack("c*"),
+          value: meta[:data][ts...te].pack("c*").force_encoding("UTF-8"),
           line: meta[:line],
           col: meta[:col],
           prev_token: meta[:previous_token],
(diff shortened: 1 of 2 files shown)


rmosolgo:
> anything else besides UTF-8

:tada: It's undefined behavior 😬   Do you think it's worth specifying in the test suite, or rejecting non-UTF-8 strings?

josh:
It looks like the parser always operates on the string assuming UTF-8 bytes regardless of how the string is tagged. It's probably *fine*. Most of these query strings will likely be coming from IO somehow and people usually screw up tagging those correctly. I feel like the strict thing to do would be to raise unless its given a UTF-8 tagged string as its required by the spec. But being liberal with the inputs is probably acceptable here. As long as the parsed AST results in UTF-8 strings, its sane for an application to deal with.

lib/graphql/language/lexer.rb
```diff
@@ -1004,7 +1004,7 @@ def self.record_comment(ts, te, meta)
       def self.emit(token_name, ts, te, meta)
         meta[:tokens] << token = GraphQL::Language::Token.new(
           name: token_name,
-          value: meta[:data][ts...te].pack("c*"),
+          value: meta[:data][ts...te].pack("c*").force_encoding("UTF-8"),
```
josh: Actually, it might be better to just change this to repack as UTF-8 with `.pack("U*")`. That results in the string being tagged correctly.

josh: Nevermind, that screws up some unicode BOM parsing. Tried it and got some failing tests.

rmosolgo:
Added a test for UTF-8 comments in 11b8c74

rmosolgo:
Thanks for tracking this down!

rmosolgo:
:ship: !

josh:
Thanks!

This is going to be a fun upgrade to try to get running on github.com. I'm sure theres a number of spots we're screwing this up. Git data + encodings = fun times.

josh:
Oh--wrong thread. Thought I was replying to https://github.com/rmosolgo/graphql-ruby/pull/517.

PR #517, opened 2017-02-02
Ensure returned String types are encoded as UTF-8

josh:
Today its possible to return binary or non-utf8 compatible strings as GraphQL String scalars. If this happens it will likely crash later when the actual JSON serialization step. This sort of thing is really hard to track down, especially if most of the time you're returning valid but untagged UTF-8 data and invalid data slips in crashing the entire response.

Since UTF-8 is a requirement, I think it would be good to enforce at the scalar casting phase.

Test failure depends on https://github.com/rmosolgo/graphql-ruby/pull/516.

My concern is that this is a backwards incompatible change. Should we add some kind of flag to opt into this behavior?

Also, I'm letting US-ASCII data through as it should be a proper subset of UTF-8. An alternative would be to upgrade the encoding to UTF-8. Or we could just enforce strict UTF-8. This should be fine in most case, but I discovered a number of introspection APIs return Ruby `Class#name` strings which are encoded as US-ASCII by default. We could encode those at the callsites, but it was just more code.

### DIFF
--- a/lib/graphql/string_type.rb
+++ b/lib/graphql/string_type.rb
@@ -3,7 +3,11 @@ GraphQL::STRING_TYPE = GraphQL::ScalarType.define do
   name "String"
   description "Represents textual data as UTF-8 character sequences. This type is most often used by GraphQL to represent free-form human-readable text."
 
-  coerce_result ->(value) { value.to_s }
+  coerce_result ->(value) {
+    str = value.to_s
+    str.encoding == Encoding::US_ASCII || str.encoding == Encoding::UTF_8 ? str : nil
+  }
+
   coerce_input ->(value) { value.is_a?(String) ? value : nil }
   default_scalar true
 end
--- a/spec/graphql/string_type_spec.rb
+++ b/spec/graphql/string_type_spec.rb
@@ -8,6 +8,13 @@ describe GraphQL::STRING_TYPE do
     assert_equal(true, string_type.default_scalar?)
   end
 
+  describe "coerce_result" do
+    it "requires string to be encoded as UTF-8" do
+      binary_str = "\0\0\0foo\255\255\255".dup.force_encoding("BINARY")
+      assert_equal nil, string_type.coerce_result(binary_str)
+    end
+  end
+
   describe "coerce_input" do
     it "accepts strings" do
       assert_equal "str", string_type.coerce_input("str")


rmosolgo:
:+1: I think this is fine, we can invite people to make custom scalars for non-UTF-8 strings. I'll let this PR mellow for a couple of days just in case someone thinks of some reason not to do it this way.

> upgrade the encoding to UTF-8

This would be 

```ruby 
-> (v) { v.to_s.force_encoding("UTF-8") } 
```

, right? That seems ok with me, because if it fails, it will fail loudly. 

Or, is there a reason that failing silently (returning nil) is better? 

> incompatible

I think we could document how to get the old behavior, something like 

```ruby 
GraphQL::STRING_TYPE.coerce_result = ->(v) { v.to_s }
MySchema = GraphQL::Schema.define { ... }
```

josh:
> , right? That seems ok with me, because if it fails, it will fail loudly.

Yeah-ish, but, I think we'd use `encode()` to attempt a transcoding. Otherwise it won't fail and the JSON encode will basically just crash reading invalid bytes as UTF-8 later.

The issue with that is `encode()` accepts options to decide how to handle invalid data. So its best left to the application to figure out how to deal with it. It might also be *slow* on some large text. Just enforcing a UTF-8 return value is a cheap as it gets and puts the concern on the app.

josh:
https://github.com/rmosolgo/graphql-ruby/pull/516 is related. I think its probably a good idea to resolve even if we didn't want to merge this specific PR.

josh:
Alright, looks like this is green now.

@rmosolgo any other thoughts on:

* How to handle US-ASCII?
* If there should be an opt-in flag?

rmosolgo:
I agree with your treatment of US-ASCII -- since it's a subset of UTF-8, we know it will be 👌  for serialization. 

I'm still not sold on the use of `nil` for the failing case. My concern is that, when that case is used, people will be unhappily surprised and they won't be given any helpful feedback. I can see in my mind's eye someone asking, "I'm trying to read this file into a GraphQL string, but it's coming through as `nil`, what's up?" and I'd like to help them help themselves. 

Do you think that's a real possibility? Is there ever a case where falling back to `nil` is a desirable behavior? Or, will people _always_ just fix it by properly encoding their strings?

Another option is to _require_ properly-encoded strings and `raise` when we get something unacceptable. That would alert the developer that their schema was trying to do something invalid and must be fixed.

josh:
> I'm still not sold on the use of nil for the failing case. My concern is that, when that case is used, people will be unhappily surprised and they won't be given any helpful feedback.

Yeah, I originally was just raising an error but it looks like all the other [`coerce_input`](https://github.com/rmosolgo/graphql-ruby/blob/a7f1bd8bbb04decdce47790b60c634c63480f765/lib/graphql/boolean_type.rb#L6) fallback to `nil`.

I'd prefer going back to raising an error. What part of the error hierarchy should it stem from? Looks like `GraphQL::Error` is the base, but is something like `RuntimeTypeError` better?

rmosolgo:
Sorry I'm hemming and hawing on this. I was thinking to myself, "What if someone wanted to use a non-JSON transport format and wanted to send binary data as a string?" And I though, "After all, the spec doesn't _require_ UTF-8 strings, does it?" So I looked, and it does: 

> The String scalar type represents textual data, represented as UTF‐8 character sequences.

https://facebook.github.io/graphql/#sec-String 

Ok, so if people want to send binary data, they can make their own `Binary` scalar, that seems fine. Let me just re-read it in the morning and think about handling that failing case ...

josh:
> Ok, so if people want to send binary data, they can make their own Binary scalar, that seems fine.

Yeah, thats how we're (@github) approaching it as well. I think our `Binary` scalar will be a base64 encoded string.

rmosolgo:
Here goes! 🙀

PR #519, opened 2017-02-03
fix proc syntax ordering on validation errors

bgentry:
An actual definition would look like this:

```ruby
resolve_type -> (obj, ctx) { ... }
```

with the dash rocket coming before the argument list.


rmosolgo:
Thanks!

Issue #511, opened 2017-01-26
Improve Rails adoption process 


rmosolgo:
Currently, the first several steps of adding GraphQL to your Rails app involve copy-pasting lots of boilerplate code: 

- controller 
- graphiql-rails gem 
- schema file 
- query type 

Then, expanding the schema also requires repetitive work: opening files and typing boilerplate code. 

What could be done to improve this process? 

If we were to provide an opinionated setup & organization strategy, what would it be? 

Could we also improve the client-side experience? (Eg, configuring apollo-client to send Rails CSRF tokens?) 

Rails generators could solve a lot of problems here. It's important to keep Rails as a totally optional dependency for this library.

jsdnxx:
One of the things we do in our controller is create a standard context object, which includes some things like an error and performance profiling accumulator, which gets flushed to the respective logging services (new relic, statsd, etc) at in an after filter.

rmosolgo:
Very cool, seems like at a minimum, a generator could guide the new user in the right direction with something like: 

```ruby
query_context  = {
  # You can access these values anywhere 
  # in GraphQL execution, so use them to 
  # scope the query or accumulate information. 
  # eg: 
  # current_user: current_user, 
}

{Generated}Schema.execute(query_string, {
  context: query_context, 
  variables: query_variables, 
})
```

nettofarah:
Here are a few ideas:

# Controller/Route setup
- Create a /graphql route in routes.rb
```ruby
post '/graphql' => 'graphql#query'
```
- Create a `app/controllers/graphql_controller.rb` controller

# GraphQL Specific folder 
Something like this:
```
app/graph
 types/
    query_type.rb
 mutations/
 fields/ # not sure about this one
 schema.rb
```
- Add app/graph to `config.autoload_paths`

jsdnxx:
ah, yep, folder structure! ours looks something like

```
./app/
  graph/
    loaders/
    mutations/
    resolvers/
    types/
    schema.rb
```

where `/loaders` is objects from GraphQL::Batch::Loader and `/resolvers` has POROs mapping to the fields in the `/types` dsl

rmosolgo:
Are mutations "just" resolvers with side effects?

rmosolgo:
Another option would be adding Sprockets-ready versions of apollo and Relay. I might be the only Sprockets user left on earth, and Relay might not even _work_ with Sprockets, but I thought I'd mention it 😆

[...]

PR #521, opened 2017-02-05
Add Rails Generators

rmosolgo:
I think generators could make the Rails experience much better! We started chatting about them over at #511 . I've sketched out a few possibilities here, what do you think? 

__TODO__:
- [x] Settle on naming conventions 
- [x] Update yardocs 
- [x] Write a guide


dphaener:
:+1: This looks like a great start. I think I would also vote for a generator that ties a bunch of those together. I typically like to create the model/type/mutations/resolvers all at the same time.

rmosolgo:
That's a good point about tying them together -- I know Rails has hooks for `rails g model` (that's how Rspec makes tests and factory girl makes factories) so we could look into those too!

dphaener:
My personal opinion on that would be to not automatically tie it in to a model generation, but rather wrap up the model generation in a "Higher Order" generator of sorts. Something like what I'm doing here: https://github.com/DNACodeStudios/react-rails-bootstrap/blob/master/lib/generators/graphql_resource_generator.rb

That makes it more of an opt in API, rather than an opt out. 

`rails g graphql_resource User email:string password_digest:string` could then generate a model, type, mutations, etc. Or you could just generate each thing separately.

vergenzt:
I know @theorygeek is working on [goco-inc/graphql-activerecord](https://github.com/goco-inc/graphql-activerecord)--would that be the most appropriate place for Rails- and/or ActiveRecord-specific code?

lib/graphql/generators/install_generator.rb
```diff
@@ -0,0 +1,44 @@
...
+module GraphQL
+  module Generators
+    # Add GraphQL to a Rails app with
+    # `rails g graphql:install`.
+    #
+    # Setup a folder structure for GraphQL:
+    #
+    # ```
+    # - app/
+    #   - graphql/
```
vergenzt: +💯 

(re: https://github.com/rmosolgo/graphql-ruby/issues/511#issuecomment-279773516)

lib/graphql/generators/install_generator.rb
```diff
@@ -0,0 +1,44 @@
...
+    # # config/routes.rb
+    # post "/graphql", to: "graphql_queries#create"
+    # ```
+    #
+    # Accept a `--relay` option which adds
+    # The root `node(id: ID!)` field.
+    #
+    # Accept a `--graphiql` option which adds
+    # `graphiql-rails` to the Gemfile and mounts the
+    # engine in `routes.rb` (Should this be the default?)
```
vergenzt: > Should this be default?

I'd think so. :) It seems like someone _not_ wanting GraphiQL will likely be the exception and not the norm.

lib/graphql/generators/loader_generator.rb
```diff
@@ -0,0 +1,15 @@
...
+module GraphQL
+  module Generators
+    # Generate a batch loader by name.
+    #
+    # ```
+    # rails g graphql:loader RecordLoader
+    # ```
+    #
+    # Should we have a loader for the
+    # nearly-universal foreign key loader?
```
vergenzt: > Should we have a loader for the nearly-universal foreign key loader?

:-1: Personally I think any code that's "nearly-universal" should probably be in a gem rather than generated into the application codebase. Not sure whether `graphql-ruby` is the most appropriate gem or not--but maybe `graphql-activerecord`?

lib/graphql/generators/install_generator.rb
```diff
@@ -0,0 +1,44 @@
...
+    #
+    # ```
+    # - app/
+    #   - graphql/
+    #     - resolvers/
+    #     - types/
+    #       - query_type.rb
+    #     - loaders/
+    #     - mutations/
+    #     - {app_name}_schema.rb
```
cjoudrey: Is there a reason for prefixing with `{app_name}_` besides the fact that otherwise we'd be conflicting with `GraphQL::Schema`?

lib/graphql/generators/install_generator.rb
```diff
@@ -0,0 +1,44 @@
...
+    #
+    # ```
+    # app/controllers/graphql_queries_controller.rb
+    # ```
+    #
+    # Add a route for that controller:
+    #
+    # ```ruby
+    # # config/routes.rb
+    # post "/graphql", to: "graphql_queries#create"
```
cjoudrey: Why not just `graphql#create`?

lib/graphql/generators/install_generator.rb
rmosolgo: :thinking: I did this in my project to avoid file picker conflicts with `db/schema.rb`. (I thought it was a real Ruby naming conflict, but looking back, I see that it's not.) Maybe that's not a good enough reason to do so!

lib/graphql/generators/install_generator.rb
rmosolgo: That could be good, then you could even do something like 

```ruby 
resource :graphql, only: :create
```

lib/graphql/generators/install_generator.rb
vergenzt: Personally I'm a fan of namespacing the schema name. Rails autoloading expects the filename to match the constant name, and having the app-wide constant for the schema simply be called `Schema` doesn't seem right to me. :/

rmosolgo:
I pushed an install generator along the same lines as initially described, how do you think in played out? 

I still need to give it a run with `rails new` and write some docs.

rmosolgo:
Thanks for you input along the way! I've tried to cover all the ground here. Does anyone want to take another look before I get to documenting it? Any reasons not to recommend this kind of setup? 

ps @dphaener Thanks for clarifying that. I think I'll start with some "simple" generators and then explore those options!

lib/graphql/generators/install_generator.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+RUBY
+
+      def create_folder_structure
+        create_dir("app/graphql/mutations")
+        create_dir("app/graphql/resolvers")
+        create_dir("app/graphql/types")
+        template("query_type.erb", "app/graphql/types/query_type.rb")
+        template("schema.erb", "app/graphql/#{schema_name.underscore}.rb")
+        template("graphqls_controller.erb", "app/controllers/graphqls_controller.rb")
```
cjoudrey: Could this be `graphql_controller.rb` instead? Pluralized `GraphQL` feels a bit weird to me.

lib/graphql/generators/templates/graphqls_controller.erb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private
+
+  # Handle form data, JSON body, or a blank value
+  def ensure_hash(ambiguous_param)
+    case ambiguous_param
+    when String
+      if ambiguous_param.present?
+        JSON.parse(ambiguous_param)
```
cjoudrey: This won't necessarily be a hash if `ambiguous_param` is `"[]"`. We might want to deal with that?

lib/graphql/generators/templates/object.erb
```diff
@@ -0,0 +1,5 @@
+<%= type_ruby_name %> = GraphQL::ObjectType.define do
+  name "<%= type_graphql_name %>"
+<% if options.node %>  interfaces [GraphQL::Relay::Node.interface]
```
cjoudrey: We can use `implements GraphQL::Relay::Node.interface` now. 🎉

rmosolgo: 🙌

lib/graphql/generators/install_generator.rb
rmosolgo: Could be, then we'd have to switch:

```diff 
- resource :graphql, only: :create 
+ post "/graphql", to: "graphql#create"
```

I guess it's not really a resource anyways, old habits die hard :P

lib/graphql/generators/templates/schema.erb
```diff
@@ -0,0 +1,27 @@
...
+    # type_name, item_id = GraphQL::Schema::UniqueWithinType.decode(id)
+    #
+    # Then, based on `type_name` and `id`
+    # find an object in your application
+    # ...
+  }
+<% end %><% if options[:batch] %>
+  # GraphQL::Batch setup:
+  lazy_resolve(Promise, :sync)
+  instrument(:query, GraphQL::Batch::Setup)
```
cjoudrey: Somewhat off-topic, but I wonder if we could turn this into a single line, say:

```ruby
GraphQL::Schema.define do
  # ...
  use_graphql_batch
end
```

And within `graphql-batch` we could do:

```ruby
GraphQL::Schema.accepts_definitions(
  use_graphql_batch: ->(schema) {
    return if schema.metadata[:graphql_batch_setup]
    schema.lazy_resolve(Promise, :sync)
    schema.instrument(:query, GraphQL::Batch::Setup)
    schema.metadata[:graphql_batch_setup]
  }
)
```

cc @dylanahsmith

lib/graphql/generators/install_generator.rb
cjoudrey: or maybe:

```ruby
post "/graphql", to: "graphql#create"
```

lib/graphql/generators/templates/graphqls_controller.erb
rmosolgo: I guess we could assert that the parsed value is a Hash. 

I just realized that this won't work on Rails 5 since ActionController::Parameters won't match `when Hash` 😖 

I'll take a look at whether I can return the parameters as-is or whether they need to be `to_unsafe_hash`ed.

lib/graphql/generators/install_generator.rb
delkopiso: I wonder then if we should keep `create` as the action. Sounds a bit off, maybe something custom like execute?

lib/graphql/generators/templates/graphqls_controller.erb
rmosolgo: Oh, it worked ok with GraphiQL since it sends the body as a string. I'll improve the messaging a bit in case someone does send random input

lib/graphql/generators/templates/object.erb
rmosolgo: update

lib/graphql/generators/install_generator.rb
rmosolgo: That's nice, I updated to `post "/graphql", to: "graphql#execute"`!

lib/graphql/generators/templates/graphqls_controller.erb
rmosolgo: donezo

rmosolgo:
I added a guide and yanked `ResolverGenerator` -- I'll add a `FunctionGenerator` along with #545  instead

PR #522, opened 2017-02-05
Correct URL

bparanj:
I have corrected the URL for the screencast link in the related projects file. Please merge.


rmosolgo:
Thanks, that's great!

PR #525, opened 2017-02-07
[Relay] Nodes field definition

xuorig:
Adds a helper method to define the `nodes` field.

Was recently added to `graphql-relay-js`: https://github.com/graphql/graphql-relay-js/pull/141

> The nodes field should accept a single argument ids: [ID!]!. It should return one result per input id, in an order matching the input arguments. For ids that do not exist or cannot be fetched, it should return null in order to ensure the ordering requirement."'

Field is talked about in https://github.com/facebook/relay/issues/1496

Suggestions welcome for a better name 💃

### DIFF
--- a/guides/relay/object_identification.md
+++ b/guides/relay/object_identification.md
@@ -89,3 +89,16 @@ QueryType = GraphQL::ObjectType.define do
   # ...
 end
 ```
+
+### `nodes` field
+
+You can also provide a root-level `nodes` field so that Relay can refetch objects by IDs. Similarly, it is provided as `GraphQL::Relay::Node.plural_field`:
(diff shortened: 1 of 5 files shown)


rmosolgo:
:+1: that's great, how about `GraphQL::Relay::Node.plural_field`? I think it fits alongside `Node.field` a bit better, you?

xuorig:
Great idea, Renamed!

rmosolgo:
:clap: :tada:

rmosolgo:
:ship: !

hkjorgensen:
I'm a little late to the party, thanks @xuorig!

PR #533, opened 2017-02-10
Allow specifying return interfaces for mutations

riffraff:
It is simple to define informal interfaces in mutation results
but it's nicer to have this supported in the type system.

It is currently possible to specify that a mutation payload
type supports certain interfaces but that requires implementing
the whole Payload type from scratch, with this change it's as simple
as declaring interface support for other types, while still
allowing the same simple definition for non-interface fields.

(NOTE: I guess this could be useful for input types too, but I haven't needed it yet)


spec/graphql/relay/mutation_spec.rb
```diff
@@ -167,6 +167,70 @@
...
+      mutation_field = mutation.field
+
+      mutation_root = GraphQL::ObjectType.define do
+        name "Mutation"
+        field :custom, mutation_field
+      end
+
+      GraphQL::Schema.define do
+        mutation(mutation_root)
+        resolve_type ->(obj, ctx) { "not really used" }
```
riffraff: this is required by the schema checks which requires a `resolve_type`  if an interface is defined, but AFAICT it should be needed only if the interface is specified explicitly as a field type.

rmosolgo:
Thanks for sharing this code! I didn't quite understand the issue before, but now I do: A mutation generates an object type based on return fields, but doesn't let you specify which interfaces that object type will implement. This PR allows that, :ok_hand:! 

I'm thinking about going one step further here. I'd like to add generated return types and generated input objects to _all_ fields, with an API like 

```ruby 
return_type do 
  # This block is just like ObjectType.define 
  field :name, types.String 
  interfaces [ErrorInterface, SuccessInterface]
end 

input_type do 
  argument :id, !types.ID
  argument :name, !types.String 
end 
```

Do you mind if I take a try at that now? I'll push a couple of commits to this branch if that's ok.

riffraff:
yep, seems like a great idea, as it also makes the mutation DSL more regular.

rmosolgo:
Hmm, there's a difference in the current implementation. 

Mutations have a wrapper object that serves fields based on hash keys. That's what supports returning a hash from a mutation `resolve` function. 

But fields don't do anything like that. So there's kind of a mismatch between `type do ... end` in a mutation vs in a field. 

We could just make it _all_ the same by requiring `field, name, type, hash_key: ...` for mutations (when they use the new `type { ... }` syntax).

riffraff:
to me it looks reasonable to require the embedded type definition to require the same, i.e.  `field name, type, hash_key: ...`. 

Since this is new syntax anyway it won't affect existing users, and it avoids adding another special case.  

(I'd actually appreciate the ability to define a way to define a default resolver per type, to be used IIF  nothing more specific is specified for a field, but it's a different topic :D)

rmosolgo:
The problem is when `interfaces` come in the mix,  then you have to redefine those fields with `hash_key: ...`, which is a drag.

I'm not crazy about how this has turned out, so I'm going to remove my commits and merge yours. I've been thinking of a new abstraction, `GraphQL::Function`, which could support _both_ fields and relay mutations, so maybe I'll try out this type-generation stuff over there!

rmosolgo:
Thanks for your work on this!

PR #535, opened 2017-02-12
Weird duplicate connection error

xuorig:
I've got this weird error going on in my sample app where I get a Duplicate type definition error on any connection used more than once.

I was able to reproduce in the tests:
`/graphql-ruby/lib/graphql/schema/type_map.rb:24:in `[]=': Duplicate type definition found for name 'ShipConnection'`

If i add anything that touches the object in `def connection_type` on `BaseType`,  (https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L142) then the error goes away :thinking:

This makes me think theres something lazily loaded / something is resetting `@connection_type` to `nil`, voiding the memoization.

@rmosolgo

### DIFF
--- a/lib/graphql/relay/connection_type.rb
+++ b/lib/graphql/relay/connection_type.rb
@@ -10,14 +10,15 @@ module GraphQL
 
       # Create a connection which exposes edges of this type
       def self.create_type(wrapped_type, edge_type: nil, edge_class: nil, nodes_field: ConnectionType.default_nodes_field, &block)
-        edge_type ||= wrapped_type.edge_type
         edge_class ||= GraphQL::Relay::Edge
-        connection_type_name = "#{wrapped_type.name}Connection"
-        connection_type_description = "The connection type for #{wrapped_type.name}."
+        edge_type ||= wrapped_type.edge_type
 
-        connection_type = ObjectType.define do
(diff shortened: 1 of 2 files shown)


xuorig:
For what its worth, this PR: https://github.com/GraphQLAcademy/swapi-graphql-ruby/pull/24

has no self-referential connections but still fails in the same way 🤔

rmosolgo:
I think the bug has to do with `ensure_defined` getting called on the same type twice in the same call stack, so _any_ cycle will cause it to happen. 

A self-referential cycle is the shortest cycle 

```

 .----.
 |    |
 A  <-'

but I think the same bug would appear in a multi-type cycle: 

```
 .-----> B
 |       |
 A  <----'

  .--> B --.
 |          |
 A <------- C
```

rmosolgo:
Something has to be made lazier: the starting type is trying to build its `fields`, which triggers the call to `#connection_type`, which also has to build its `fields`, which points back to the starting type ... 

I hope there's some way to return the connection type without immediately building its fields, I can probably dig in more tomorrow!

xuorig:
nice :D

xuorig:
Thanks robert!

PR #537, opened 2017-02-13
Allow public access to context for Relay Connection

mvgijssel:
I need access to the context in a custom EdgeClass so we can do authorization checks on the edge resource. Adding `context` to the `attr_reader` seems like the easiest approach.


rmosolgo:
Weird, seems like one of the CI builds is frozen and won't unstick. Looks good to me, thanks!!

Issue #534, opened 2017-02-10
Expose Printer#print_type


swalkinshaw:
I'm working on documentation for GraphQL schemas and thought it might be useful to have an option to show the definition of a type of IDL. (kind of like a "view source" for normal language API docs).

The easiest way to do this is to call `GraphQL::Schema::Printer#print_type` except it's a private method.

Would you be open to some minor changes to make this a public method? If I manually construct a warden and call the method via `send`, it seems to do exactly what I'm looking for.

I originally thought I could just pass in an `only` lambda with the type name, but it doesn't return the full type definition.

Example:

```ruby
GraphQL::Schema::Printer.print_schema(schema, only: ->(m, ctx) { m.name == 'Product' })
=> "schema {\n}\n\ntype Product {\n\n}"
```

```ruby
GraphQL::Schema::Printer.send(:print_type, warden, type)
=>
# returns full definition
```

xuorig:
maybe add `ObjectType.to_definition` with that :)

swalkinshaw:
That's an even better idea :)

rmosolgo:
:+1: sounds like a great idea! `BaseType#to_definition` seems like a nice API. Optional `only:`/`except:` would be consistent, but feel free to put that off until later.

rmosolgo:
Addressed by #539

PR #539, opened 2017-02-13
[WIP] Add to_definition to BaseType to print definition

swalkinshaw:
Ref https://github.com/rmosolgo/graphql-ruby/issues/534

Refactors `Printer` from a module to a class to make it easier to print out the definition of a single type.

@rmosolgo this is a WIP approach to exposing `print_type`. Wanted to see if you like it before going much further.

Turns out that adding `to_definition` is not that easy for a few reasons:

1. Warden is used in all the printers
2. Warden is scoped to a schema

So adding a `#to_definition` *without* requiring a schema/warden would result in a lot of duplication of the printers. It also might be a bad thing to expose printers without the ability to mask anything.

The one downside to this API is that calling `to_definition` requires a schema.

This was the most direct refactor which also maintains the existing `print_schema` and `print_introspection_schema` methods.

A simpler alternative is to simply make the existing `print_type` a public method so if this approach isn't good, that's a simple fallback.

/cc @xuorig

### DIFF
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -159,6 +159,14 @@ module GraphQL
       GraphQL::Relay::EdgeType.create_type(self, **kwargs, &block)
     end
 
+    # Return a GraphQL string for the type definition
+    # @param schema [GraphQL::Schema]
+    # @param printer [GraphQL::Schema::Printer]
+    # @param context [Hash]
+    # @param only [<#call(member, ctx)>]
+    # @param except [<#call(member, ctx)>]
+    # @param warden [GraphQL::Warden]
+    # @return [String] type definition
     def to_definition(schema, printer: nil, **args)
       printer ||= GraphQL::Schema::Printer.new(schema, **args)
       printer.print_type(self)
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -158,5 +158,10 @@ module GraphQL
     def define_edge(**kwargs, &block)
       GraphQL::Relay::EdgeType.create_type(self, **kwargs, &block)
     end
+
+    def to_definition(schema, printer: nil, **args)
+      printer ||= GraphQL::Schema::Printer.new(schema, **args)
+      printer.print_type(self)
+    end
   end
 end
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -162,10 +162,7 @@ module GraphQL
     # Return a GraphQL string for the type definition
     # @param schema [GraphQL::Schema]
     # @param printer [GraphQL::Schema::Printer]
-    # @param context [Hash]
-    # @param only [<#call(member, ctx)>]
-    # @param except [<#call(member, ctx)>]
-    # @param warden [GraphQL::Warden]
+    # @see {GraphQL::Schema::Printer#initialize for additional options}
     # @return [String] type definition
     def to_definition(schema, printer: nil, **args)
       printer ||= GraphQL::Schema::Printer.new(schema, **args)
--- a/lib/graphql/schema/printer.rb
+++ b/lib/graphql/schema/printer.rb
@@ -43,17 +43,13 @@ module GraphQL
       # @param context [Hash]
       # @param only [<#call(member, ctx)>]
       # @param except [<#call(member, ctx)>]
-      # @param warden [GraphQL::Warden]
-      def initialize(schema, context: nil, only: nil, except: nil, warden: nil)
+      # @param introspection [Boolean] Should include the introspection types in the string?
+      def initialize(schema, context: nil, only: nil, except: nil, introspection: false)
         @schema = schema
         @context = context
 
-        if warden.nil?
-          blacklist = build_blacklist(only, except)
-          warden = GraphQL::Schema::Warden.new(blacklist, schema: @schema, context: @context)
-        end
-
-        @warden = warden
+        blacklist = build_blacklist(only, except, introspection: introspection)
+        @warden = GraphQL::Schema::Warden.new(blacklist, schema: @schema, context: @context)
       end
 
       # Return the GraphQL schema string for the introspection type system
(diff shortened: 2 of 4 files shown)


lib/graphql/schema/printer.rb
```diff
@@ -7,37 +7,55 @@ class Schema
...
-    module Printer
-      extend self
+    class Printer
+      attr_reader :schema, :warden
+
+      def initialize(schema, context: nil, only: nil, except: nil, warden: nil)
+        @schema = schema
+        @context = context
+        @only = only
+        @except = except
```
rmosolgo: Is it important to persist `@only` and `@except` in private state? If we don't _need_ them, I'd rather not save them. That way, we can avoid any mutable state bugs or untested downstream dependencies. For example, could we do something like: 

```ruby
if warden.nil?
  blacklist = build_blacklist(only, except)
  warden = # ... 
end 

@warden = warden
```

spec/graphql/schema/printer_spec.rb
```diff
@@ -590,4 +590,20 @@
...
+# A blog post
+type Post {
+  id: ID!
+  title: String!
+  body: String!
+  comments: [Comment!]
+  comments_count: Int! @deprecated(reason: \"Use \\\"comments\\\".\")
+}
+SCHEMA
+      assert_equal expected.chomp, GraphQL::Schema::Printer.new(schema).print_type(schema.types['Post'])
```
rmosolgo: Since the public API is `BaseType#to_definition`, how about calling that instead? (In a sense, `Schema::Printer` is an implementation detail of that API.) This test could even go in `base_type_spec`.

swalkinshaw: Will do, it needed more tests anyway, this was just a sanity check for myself 👍

swalkinshaw: I added a `to_definition` test to `base_type_spec` as well but left this as well for now. `print_type` is a public method as well but I'll gladly remove this if you want

swalkinshaw:
@rmosolgo I've updated all the doc strings as well

lib/graphql/base_type.rb
```diff
@@ -158,5 +158,18 @@ def edge_type
...
     end
+
+    # Return a GraphQL string for the type definition
+    # @param schema [GraphQL::Schema]
+    # @param printer [GraphQL::Schema::Printer]
+    # @param context [Hash]
+    # @param only [<#call(member, ctx)>]
+    # @param except [<#call(member, ctx)>]
+    # @param warden [GraphQL::Warden]
+    # @return [String] type definition
```
swalkinshaw: @rmosolgo do you want me to enumerate out the params here like I did? Or just put `args` with a comment about it?

rmosolgo: :+1: this is totally fine by me, IMO you could also use `@see`

lib/graphql/schema/printer.rb
```diff
@@ -3,41 +3,98 @@ module GraphQL
...
+    #
+    #     field :id, !types.ID
+    #     field :title, !types.String
+    #     field :body, !types.String
+    #   end
+    #
+    #   MySchema = GraphQL::Schema.define(query: query_root)
+    #
+    #   printer = GraphQL::Schema::Printer.new(MySchema)
+    #   puts printer.print_type(post_type)
```
rmosolgo: Can we direct people to use the `#to_definition` methods? Or do you think we need to keep `Printer.new` as a public API too?

lib/graphql/schema/printer.rb
```diff
@@ -3,41 +3,98 @@ module GraphQL
...
       # @param except [<#call(member, ctx)>]
-      # @param schema [GraphQL::Schema]
-      def print_schema(schema, context: nil, only: nil, except: nil)
-        blacklist = if only
-          ->(m, ctx) { !(IS_USER_DEFINED_MEMBER.call(m) && only.call(m, ctx)) }
-        elsif except
-          ->(m, ctx) { !IS_USER_DEFINED_MEMBER.call(m) || except.call(m, ctx) }
-        else
-          ->(m, ctx) { !IS_USER_DEFINED_MEMBER.call(m) }
+      # @param warden [GraphQL::Warden]
```
rmosolgo: I would like to keep `warden`s out of the public eye, and instead expose `only:` and `except:` everywhere they are used. Do you mind if I push a commit to make that refactor? (I know some of the class methods will need adjustments.) Or, is there a use case where you need to pass a `warden` in particular? (Where `only` or `except` won't work?)

spec/graphql/schema/printer_spec.rb
rmosolgo: 👌  I'm not the type to say no to an extra spec 😆

lib/graphql/schema/printer.rb
swalkinshaw: Nope, go right ahead 👍 

I did this to easily maintain backwards compat for `print_introspection_schema`. It wouldn't work like this:


```def
  query_root = ObjectType.define(name: "Root")
  schema = GraphQL::Schema.define(query: query_root)
  blacklist = ->(m, ctx) { m == query_root }

  printer = new(schema, only: blacklist)
  printer.print_schema
end
```

Because the `IS_USER_DEFINED_MEMBER` gets applied to `only`. So my way around that was to allow an instance of `warden` to be passed in.

lib/graphql/schema/printer.rb
swalkinshaw: I can't think of a good reason for someone to use one over the other.

```ruby
# via printer instance
printer = Graph::Schema::Printer.new(schema)
schema.types.each {|t| printer.print_type(t) }

# via to_def with a printer
printer = Graph::Schema::Printer.new(schema)
schema.types.each {|t| t.to_definition(schema, printer: printer) }

# via to_def with just a schema
schema.types.each {|t| t.to_definition(schema) }
```

I wanted to avoid instantiating a printer for each type but that's possible anyway with the optional `printer` param.

lib/graphql/schema/printer.rb
rmosolgo: Ohhh I see the difficulty now ... I added `introspection:` to the arguments which seems to still work but I feel better about the implementation wiggle room :)

rmosolgo:
:clap: bravo, thanks!

swalkinshaw:
Thanks 🎉

PR #541, opened 2017-02-13
refactor(Field) Rebuild the default NameResolve when name is changed

xuorig:
Name changes when redefining a field now reset the resolve proc instead of raising.

@rmosolgo


rmosolgo:
:tada: !

Issue #540, opened 2017-02-13
Specify property and/or encoding for arguments.


xuorig:
Thought about this while working on https://github.com/rmosolgo/graphql-ruby/issues/538

It would be pretty useful if you could specify a different name for internal use of arguments vs client facing. For example:

```ruby
field :addThing, !types.String do
  argument :myArgument, !types.String, as: :my_argument

  resolve ->(o, args, ctx) {
    AddThingService.new(args[:my_argument]).result
  }
end
```

It would be pretty useful to shape the arguments the way you want to to adapt to your business logic, but also would be useful for this issue: https://github.com/rmosolgo/graphql-ruby/issues/538

Other more complex cases could allow you to change argument values before they get to your resolver. Relay Global ids are a good example.

```ruby
field :updateThing, !types.String do
  argument :id, !types.ID, decoder: RelayGlobalIDDecoder

  resolve ->(o, args, ctx) {
    # id is already an active record obj id because RelayGlobalIDDecoder decoded the global id
    Thing.find(args[:id])
  }
end
```

WDYT ?

rmosolgo:
:+1:

lucaspolonio:
@rmosolgo any thoughts on @xuorig's second idea to add a `decoder` option to arguments? I'd be willing to work on a PR if you like the idea.

My current use case is argument reusability. I have a few custom arguments using `GraphQL::Field.accepts_definitions`and `Define::AssignArgument`. All they do now is provide default types and descriptions. It would be very useful if they could also provide a block as a `decoder` option (maybe `resolve` or `prepare`, or any other name) to validate/transform the argument value during execution before the resolver is called. One example could look something like this:

```ruby
field :my_field do
  date_argument :date
  resolve -> (obj, args, ctx) {
    args[:date]
  }
end
```

```ruby
module AssignDateArgument
  def self.call(target, name, type, description, **rest, &block)
    description = "...."
    prepare = ->(arg) { ValidateDateArgument.call(arg) }
    ...
    GraphQL::Define::AssignArgument.call(target, name, type, description, prepare: prepare, **rest, &block)
  end
end

GraphQL::Field.accepts_definitions(date_argument: AssignDateArgument)
```

```ruby
module ValidateDateArgument
  def self.call(arg)
    raise GraphQL::ExecutionError.new("Invalid date format") unless ...
  end
end
```

rmosolgo:
I love the idea, I think `prepare:` is a great name for it, please give it a try!

[...]

PR #542, opened 2017-02-14
feat(Arguments) Allow specifying an alternate key for arguments, used in resolvers

xuorig:
related to https://github.com/rmosolgo/graphql-ruby/issues/540

Allow specifying a key that will present in the `Arguments` at resolve time instead of the specified argument name.

```ruby
field :post, PostType do
  argument :postId, types.Id, as: :id
  resolve ->(obj, args, ctx) {
    Post.find(args['id'])
  }
end
```

good for review @rmosolgo, I'm open to other names for the arg 🤔

### DIFF
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -10,8 +10,14 @@ module GraphQL
       def initialize(values, argument_definitions:)
         @original_values = values
         @argument_values = values.inject({}) do |memo, (inner_key, inner_value)|
-          string_key = inner_key.to_s
-          arg_defn = argument_definitions[string_key]
+          arg_defn = argument_definitions[inner_key.to_s]
+
+          string_key = if arg_defn.as != nil
+            arg_defn.as.to_s
+          else
+            inner_key.to_s
+          end
+
           arg_value = wrap_value(inner_value, arg_defn.type)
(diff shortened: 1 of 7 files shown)


lib/graphql/query/arguments.rb
```diff
@@ -10,8 +10,14 @@ class Arguments
       def initialize(values, argument_definitions:)
         @original_values = values
         @argument_values = values.inject({}) do |memo, (inner_key, inner_value)|
-          string_key = inner_key.to_s
-          arg_defn = argument_definitions[string_key]
+          arg_defn = argument_definitions[inner_key.to_s]
+
+          string_key = if arg_defn.as != nil
+            arg_defn.as.to_s
```
rmosolgo: Do we need to `to_s` this at runtime, or could we do it during definition, like 

```
argument.as = rest[:as].to_s
```

?

xuorig: good question. I'm tempted to leave at run time. I feel it might be confusing if someone is for example, comparing arguments together, or doing something like `if argument.as == :aSymbol`

rmosolgo: Hmm, maybe we can have _both_. I've taken a try in 9df2c346, what do you think?

xuorig:
LGTM!

rmosolgo:
Very cool :metal:

PR #543, opened 2017-02-14
Docs on mutations

xuorig:
There's not really any examples on how to add a mutation root and basic mutations on the website. (Only Relay AFAIK).

This adds a simple example on how to do so!

### DIFF
--- a/guides/schema/types_and_fields.md
+++ b/guides/schema/types_and_fields.md
@@ -141,6 +141,45 @@ resolve ->(obj, args, ctx) {
 }
 ```
 
+## Mutations
+
+Registering a mutation root allows to define fields that can mutate your data.
+
+```
+Schema = GraphQL::Schema.define do
+  mutation MutationRoot
+end
+
+MutationRoot = GraphQL::ObjectType.define do
+  name "Mutation"
+
+  field :addPost, Post do
+    description "Find dairy products matching a description"
+
+    # Use Input Types to define complex argument types
+    argument :post, PostInputType
+    resolve ->(t, args, c) {
+      title = args['post']['title']
+      description = args['post']['description']
+      Post.create(title: title, description: description)
+    }
+  end
+end
+
+PostInputType = GraphQL::InputObjectType.define do
+  name "PostInputType"
+  description "Properties for creating a Post"
+
+  input_field :title, !types.String do
+    description "Title of the post."
+  end
+
+  input_field :description, types.String do
+    description "Description of the post."
+  end
+end
+```
+
 ## Extending type and field definitions
 
 Types, fields, and arguments have a `metadata` hash which accepts values during definition.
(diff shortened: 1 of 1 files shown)


guides/schema/types_and_fields.md
```diff
@@ -141,6 +141,45 @@ resolve ->(obj, args, ctx) {
...
+```ruby
+Schema = GraphQL::Schema.define do
+  mutation MutationRoot
+end
+
+MutationRoot = GraphQL::ObjectType.define do
+  name "Mutation"
+
+  field :addPost, Post do
+    description "Find dairy products matching a description"
```
rmosolgo: I don't think you'll find any dairy products on this field 😆

guides/schema/types_and_fields.md
```diff
@@ -141,6 +141,45 @@ resolve ->(obj, args, ctx) {
...
+      Post.create(title: title, description: description)
+    }
+  end
+end
+
+PostInputType = GraphQL::InputObjectType.define do
+  name "PostInputType"
+  description "Properties for creating a Post"
+
+  input_field :title, !types.String do
```
rmosolgo: Can we recommend [`argument`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/input_object_type.rb#L30) instead of `input_field`? I would like to start recommending that one since it's the same as field `argument`s

guides/schema/types_and_fields.md
xuorig: loool

guides/schema/types_and_fields.md
xuorig: 👍

xuorig:
made the changes @rmosolgo !

guides/schema/types_and_fields.md
```diff
@@ -141,6 +141,45 @@ resolve ->(obj, args, ctx) {
 }
 ```
 
+## Mutations
+
+Registering a mutation root allows to define fields that can mutate your data.
+
+```ruby
+Schema = GraphQL::Schema.define do
+  mutation MutationRoot
```
vergenzt: Don't all schemas have to define a `query` root too?

xuorig: > The query type must always be provided, and is an Object base type. The mutation type is optional; if it is null, that means the system does not support mutations. If it is provided, it must be an object base type.

You are right sir! I'll add a fake query too to the example.

rmosolgo:
Sweet, thanks for giving these docs some love!

rahilsondhi:
FYI looks like this is not live @ https://rmosolgo.github.io/graphql-ruby/schema/types_and_fields

rmosolgo:
Yes, sorry, I think that's because it was mixed in with some changes that don't apply to the latest release. So if I regenerate & deploy the site, it would reflect master, not the latest release :S

rmosolgo:
That could be addressed by checking out the 1.4.x branch and cherry-picking these commits to it. If you do the PR, I'll do the deploy!

PR #545, opened 2017-02-14
feat(GraphQL::Function) add a code-reuse container with resolve + type info 

rmosolgo:
Reusing `resolve` functions is OK, but it has a big problem: the resolve function gets separated from its argument and return type specifications. 

What I really want is a unit of code reuse that: 

- carries type information 
- carries the resolve behavior 
- can be parameterized 
- works equally well for a plain `field` or a `Relay::Mutation` 

So `GraphQL::Function` is a take on that. Examples are in the docs or specs.

What do you think? Is this a real problem? Does this approach have any pros or cons?


lib/graphql/function.rb
```diff
@@ -0,0 +1,101 @@
...
+  #
+  #     def resolve(obj, args, ctx)
+  #        @model.find(args.id)
+  #     end
+  #   end
+  #
+  #   QueryType = GraphQL::ObjectType.define do
+  #     name "Query"
+  #     field :post, function: FindRecord.new(model: Post, type: PostType)
+  #     field :post, function: FindRecord.new(model: Comment, type: CommentType)
```
swalkinshaw: `field :comment` I assume?

gjtorikian:
Wow, this is looking pretty nice already. At GitHub, we've just started (like, a couple of weeks ago) trying to write "global" resolvers, which seems to be what this PR is aimed at doing. 

We've started doing stuff like this:

``` ruby
connection :pullRequests, -> { !Connections::PullRequests } do
  description "A list of pull requests that the user owns."

  arguments Resolvers::PullRequests.default_arguments

  resolve Resolvers::PullRequests
end
```

With the expectation that the `pullRequests` connection could hang off of `User` or a Repository. `Resolvers::PullRequests.call` than just decides the right Rails arel call to make.

A couple of questions:

* I believe `connections` are just `field`s in the end with additional data attached, right? It would be helpful if the tests had a test for `connections`, too--just in case.
* Is it possible for defined attributes to override those defined in a `Function`? For example, for `:pullRequests`, the different definitions might read:

    ``` ruby
    connection :pullRequests, "A list of pull requests that the user owns." function: PullRequest.new(model: PullRequest, type: Objects::PullRequest)

   ...

   connection :pullRequests, "A list of pull requests that the repository owns." function: PullRequest.new(model: PullRequest, type: Objects::PullRequest)
    ```

    That way, connections can make use of the same function, but have, for example, differing definitions.

   Another way this may come in handy is additional arguments, such as including field-specific ones  versus global defaults.

lib/graphql/function.rb
rmosolgo: lol, oops!

rmosolgo:
Yeah, that approach at resolvers is just what I'm going for! An object that responds to `#call` is _not quite_ enough, since `#call` depends on a certain set of arguments and probably has a consistent return type.

Good point about `connection`, I hope it will be straightforward but anything that we add will definitely support that too.

I also considered "overriding", where the function is treated as a starting point, and any local configurations would override or add to it. I didn't think of any real use cases, but you're right about description and argument default values, those are both very good applications.

Thanks for taking a look, glad to know I'm not the only one looking for a little something more in this department!

rmosolgo:
Rebased, doc'd, added a generator ... I think this is basically ready. Anything else?

guides/schema/code_reuse.md
```diff
@@ -4,76 +4,84 @@ title: Schema — Code Reuse
...
+```
 
-  def call(obj, args, ctx)
-    obj.public_send(@method_name)
+- `#arguments => Hash<String => GraphQL::Argument>`
+- `#type => GraphQL::BaseType`
+- `#call(obj, args, ctx) => Object`
+- `#complexity => Integer, Proc`
+- `#description => String, nil`
+- `#deprecation_reason => String, nil`
```
gjtorikian: An introduction here might be useful. Is this a list of methods available to `GraphQL::Function`?

guides/schema/code_reuse.md
```diff
@@ -4,76 +4,84 @@ title: Schema — Code Reuse
...
+class FindRecord < GraphQL::Function
+  # ...
 end
 
-# ...
-
-field :name, types.String do
-  resolve(resolve_with_method(:name))
+# 👌 Arguments, description, etc are inherited as usual:
+class BatchedFindRecord < FindRecord
```
gjtorikian: 😍

guides/schema/code_reuse.md
```diff
@@ -4,76 +4,84 @@ title: Schema — Code Reuse
...
 end
-
-# ...
-field :name, types.String do
-  resolve(MethodCallResolver.new(:name))
-end  
 ```
 
-Or, you can generate the `resolve` Proc dynamically:
+Note that `types.` is _not_ available. Instead, you should reference GraphQL's built-in {{ "GraphQL::ScalarType" | api_doc }}s directly.
```
gjtorikian: These docs also don't mention the "extensibility" of the feature. I thought this might have been removed, but it appears not!

What I mean is, the ability to add additional `arguments` or whatever to the global function.

rmosolgo: Oh, I just forgot it! I'll add a bit for that.

guides/schema/code_reuse.md
rmosolgo: 😬  seems like I lost a line in there. Those are the methods a Function has to implement. I'll add that back!

guides/schema/code_reuse.md
rmosolgo: pushed!

PR #547, opened 2017-02-16
Better error msg mutation result

xuorig:
In a Relay Mutation with a generated payload type using `return_field`, if you make the mistake of returning anything but a Hash, you'll get an exception when building the result. `NoMethodError each`.

This is pretty hard to debug for someone new to the gem, would be great to raise a helpful exception here.

@rmosolgo

cc @cjoudrey

### DIFF
--- a/lib/graphql/relay/mutation.rb
+++ b/lib/graphql/relay/mutation.rb
@@ -199,6 +199,11 @@ module GraphQL
           end
 
           if @wrap_result
+            unless mutation_result == nil || mutation_result.is_a?(Hash)
+              raise StandardError, "Expected `#{mutation_result}` to be a Hash."\
+                " Return a hash when using `return_field` or specify a custom `return_type`."
+            end
+
             @mutation.result_class.new(client_mutation_id: args[:input][:clientMutationId], result: mutation_result)
           else
             mutation_result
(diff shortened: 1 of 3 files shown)


lib/graphql/relay/mutation.rb
```diff
@@ -199,6 +199,11 @@ def build_result(mutation_result, args, ctx)
           end
 
           if @wrap_result
+            unless mutation_result == nil || mutation_result.is_a?(Hash)
```
cjoudrey: It might just be me, but `unless` with `||` has always hurt my brain.

I think you can turn this line into:

```ruby
if mutation_result && !mutation_result.is_a?(Hash)
```

rmosolgo: Haha, I'm in the same boat, in my head I translate `unless` to "if not" every time. This could even be

```ruby
if !mutation_result.is_a?(Hash)
```

since everything responds to `is_a?`.

xuorig: made the change 👍

rmosolgo:
👏  Thanks for this improvement!

xuorig:
oops last change seems to have broken something 👀

xuorig:
Fix: https://github.com/rmosolgo/graphql-ruby/pull/549

Issue #546, opened 2017-02-15
Add ObjectType#implements


rmosolgo:
It would be nice to specify interfaces one-at-a-time, like 

```ruby 
GraphQL::ObjectType.define do 
  name "Page" 
  implements GraphQL::Relay::Node.interface 
  implements TitledEntityInterface
  # ...
end 
```

This would push entries into the `@interfaces` array. `ObjectType#interfaces` definition should be made additive, so that it works alongside `implements`.

PR #548, opened 2017-02-16
Add ObjectType#implements

cjoudrey:
As per #546, this adds `ObjectType#implements` as a way to add an interface to a type.

**Usage**

```ruby
GraphQL::ObjectType.define do 
  name "Page" 
  implements TitledEntityInterface
  # ...
end 
```

You can also pass multiple interfaces at once:

```ruby
GraphQL::ObjectType.define do 
  name "Page" 
  implements GraphQL::Relay::Node.interface, TitledEntityInterface
  # ...
end 
```

:eyes: @rmosolgo

### DIFF
--- a/guides/relay/object_identification.md
+++ b/guides/relay/object_identification.md
@@ -58,7 +58,7 @@ To implement the node interface, include `GraphQL::Relay::Node.interface` in you
 PostType = GraphQL::ObjectType.define do
   name "Post"
   # Implement the "Node" interface for Relay
-  interfaces [GraphQL::Relay::Node.interface]
+  implements GraphQL::Relay::Node.interface
   # ...
 end
 ```
--- a/lib/graphql/object_type.rb
+++ b/lib/graphql/object_type.rb
@@ -48,6 +48,7 @@ module GraphQL
     end
 
     # @param new_interfaces [Array<GraphQL::Interface>] interfaces that this type implements
+    # @deprecated Use `implements` instead of `interfaces`.
     def interfaces=(new_interfaces)
       @clean_interfaces = nil
       @dirty_interfaces = new_interfaces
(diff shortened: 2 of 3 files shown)


rmosolgo:
I _think_ you're gonna have a bad time if you do `implements` followed by `interfaces`:

```ruby 
GraphQL::ObjectType.define do 
  name "Page" 
  implements TitledEntityInterface
  interfaces [GraphQL::Relay::Node.interface, AnotherInterface]
  # ...
end 
```

But maybe that's ok, since changing how `interfaces` works would be a "breaking" change anyways.

cjoudrey:
Yeah.. I wouldn't encourage people to use `interfaces` and `implements` together.

I was thinking `implements` could accept an array of interfaces and it keeps appending and we deprecate `interfaces`.

xuorig:
wonder if one day we could have something like

```
implements MyInterface, define_fields: true
```

which would define the interface fields on your object type for you 🤔 
I know this have been talked about a lot of though :D

rmosolgo:
:+1: for a splat of interfaces (and deprecating `interfaces`), that would be awesome

:+1: for `define_fields:` (maybe `inherit_fields:`?) but I wonder, should we allow the interface to specify that its fields are "abstract", or should we let the implementer decided? (You might say "oh heck, why not inherit the fields" only to discover that the interface doesn't actually provide implementations)

xuorig:
> interface to specify that its fields are "abstract"

Yep thats a great idea. We have something custom but similar like an

```
on_implements do
  # actual definitions that can be inherited
  field :...
end
```

cjoudrey:
If we decide interfaces can specify resolvers, we need to make sure there is way to override those.

I remember running into an issue at Shopify where `Node` was an interface you'd implement and it would add the `:id` field on your type for you, but at the time you had no way of modifying that resolver, which caused issues.

cjoudrey:
> 👍 for a splat of interfaces (and deprecating interfaces), that would be awesome

Okay, I'll add that to this PR.

rmosolgo:
Awesome! Sorry for the delay on merging!

PR #549, opened 2017-02-16
Mutation error: dont raise when mutation is nil

xuorig:
@rmosolgo 

Should not raise when mutation returns nil, only if its returning something that is not a hash.


rmosolgo:
Thanks! Sorry I broke the build 😬  You had it right the first time, [my suggestion](https://github.com/rmosolgo/graphql-ruby/pull/547#discussion_r10 1592356) was wrong!

PR #550, opened 2017-02-16
feat(Relay::Node) allow passing a custom resolver to the node and nod…

xuorig:
…es fields

Context:

Wanted to batch the `node` and `nodes` field which requires you to return a lazy object in `object_from_id`

If you do that, by default the `nodes` field would return an array of promises / lazy things `[Promise, Promise, Promise]`.

Of course, `graphql-ruby` cant magically guess this is an array of lazy objects, and not all implementations have something like `Promise.all([...])` so that would be hard to add.

With this PR this would allow somebody to override the default resolve when something like this:

```ruby
GraphQL::Relay::Node.plural_field(
  resolve: ->(_, args, ctx) { Promise.all(args[:ids].map { |id| ctx.query.schema.object_from_id(id, ctx) }) }
)

### DIFF
--- a/lib/graphql/relay/node.rb
+++ b/lib/graphql/relay/node.rb
@@ -4,7 +4,7 @@ module GraphQL
     # Helpers for working with Relay-specific Node objects.
     module Node
       # @return [GraphQL::Field] a field for finding objects by their global ID.
-      def self.field
+      def self.field(resolve: nil)
         # We have to define it fresh each time because
         # its name will be modified and its description
         # _may_ be modified.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rmosolgo** [comment]
Awesome! For some reason I thought that was already possible but obviously not! 

I'll merge this as-is but a long-term win would be to accept _any_ definition from those methods, something like 

```
def field(**kwargs, &block)
  node_field = GraphQL::Field.define { 
    # ... existing definition here 
  }
  # Then apply user-specified overrides:
  node_field.redefine(kwargs, &block)
end 
```

xuorig:
Great idea @rmosolgo, I've implemented it here: https://github.com/rmosolgo/graphql-ruby/pull/552

PR #552, opened 2017-02-17
feat(Relay::Node) extend node & nodes fields using definition block

xuorig:
As mentioned in https://github.com/rmosolgo/graphql-ruby/pull/550 by @rmosolgo. It think its even nicer to allow anything to be redefined on the standard node and nodes field.

### DIFF
--- a/spec/graphql/relay/node_spec.rb
+++ b/spec/graphql/relay/node_spec.rb
@@ -9,43 +9,18 @@ describe GraphQL::Relay::Node do
   end
 
   describe ".field" do
-    describe "with custom resolver" do
-      it "executes the custom resolve instead of relay default" do
-        id = "resolver_is_hardcoded_so_this_does_not_matter"
+    it 'accepts a custom definition' do
+      faction = StarWars::DATA['Faction'][1]
 
-        result = star_wars_query(%|{
-          nodeWithCustomResolver(id: "#{id}") {
-            id,
-            ... on Faction {
-              name
-              ships(first: 1) {
-                edges {
-                 node {
-                   name
-                   }
-                }
-              }
-            }
-          }
-        }|)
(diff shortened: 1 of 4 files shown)


spec/graphql/relay/node_spec.rb
```diff
@@ -9,43 +9,18 @@
...
-
-        assert_equal(expected, result)
+      node_field = GraphQL::Relay::Node.field do
+        name "nod3"
+        description "The Relay Node Field"
+        resolve ->(_, _ , _) { faction }
       end
+
+      node_field.name = "nod3"
+      node_field.description = "The Relay Node Field"
```
rmosolgo: I noticed these assign the same values as assigned above, are these meant to be assertions?

spec/graphql/relay/node_spec.rb
```diff
@@ -131,53 +106,18 @@
...
 
-        assert_equal(expected, result)
+      node_field = GraphQL::Relay::Node.plural_field do
+        name "nodez"
+        description "The Relay Nodes Field"
+        resolve ->(_, _ , _) { factions }
       end
+
+      node_field.name = "nodez"
+      node_field.description = "The Relay Nodes Field"
```
rmosolgo: (same as above)

rmosolgo:
:+1: looks good to me! How about keeping the old test just as an end-to-end test, and to make sure that kwargs are also handled properly?

spec/graphql/relay/node_spec.rb
xuorig: oh. yes. I need ☕️

xuorig:
Added back the tests!

rmosolgo:
Sweet :D

PR #553, opened 2017-02-17
Analyzer context

xuorig:
Opening a PR to get your opinion on this @rmosolgo:

It would be nice to have access to query context inside analyzers. WDYT ?


rmosolgo:
This can be accomplished without a breaking change: 

```ruby
      def initial_value(query)
        {
          ctx: query.context,
          # ... 
        }
      end
```

Is that sufficient? Or is there another benefit to passing it directly?

xuorig:
oh! Well that answers the question. Forgot about `initial_value`. Closing this!

cjoudrey:
Wow! TIL. We should add an example of this to the Guides if there isn't one already.

rmosolgo:
It's mentioned a bit here, but not very discoverable :S 

http://rmosolgo.github.io/graphql-ruby/queries/authorization#query-analyzers

PR #554, opened 2017-02-18
feat(Rewrite) add interceptor API

rmosolgo:
This allows user code to interact with the "codegen" phase of GraphQL execution. A crucial part of custom directives is the ability to alter the query tree (that's what `@skip` and `@include` do), this is an API to allow that. 

This is like the fourth time I've written a middleware-y API and it reflects some lessons-learned: 

- User code needs access to _everything_ (query, schema, context, plus visited items). 
- User code needs an isolated place to store state. 
- User code needs an initialization hook. (Do we need a finalization hook?)

If it works out well, I may try to conform the other middleware-y type things to this. It's a pain how they're all different! 

__TODO:__
- [ ] Test some elements of this API 
- [ ] Document Rewrite interceptor and custom directives
- [ ] Unify other middleware-y APIs? 
- [ ] Make sure it's not much slower


rmosolgo:
This ain't gonna cut it. `@defer` needs access to the runtime object and this doesn't quite get us there.

PR #555, opened 2017-02-19
auto-camelize option for field & argument names.

xuorig:


spec/support/camelized/schema.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  QueryType = GraphQL::ObjectType.define do
+    name "Query"
+    field :shop, Camelized::Shop, resolve: ->(_, _, _) { Camelized::ShopObject }
+  end
+
+  Schema = GraphQL::Schema.define do
+    camelize(true)
```
xuorig: Suggestions welcome for a better name / API for this :)

graphql.gemspec
```diff
@@ -41,6 +41,9 @@ Gem::Specification.new do |s|
   s.add_development_dependency "sequel"
   s.add_development_dependency "sqlite3"
 
+  # TODO(xuorig) probably can in house it:
+  s.add_runtime_dependency "activesupport"
```
xuorig: Only used for camelizing, may be brought into `graphql-ruby`?

spec/support/camelized/schema.rb
rmosolgo: :thinking:  I don't have any better suggestions, here are some brainstorming ideas ... 

```ruby 
# Do we want to allow other options someday?
transform_identifiers :camelize
rename :camelize 
camelize_fields(true) 
camelize_fields_and_arguments(true) 
```

graphql.gemspec
rmosolgo: :+1: since we're only using one function, we can probably avoid forcing activesupport on people.

One option would be to use a built-in implementation by default, but support user-provided transforms, something like 

```
rename :camelize # => uses the built-in transformation 
rename MyCustomRename # => uses `.call(name)` to get a new name 
```

That way, people who _want_ some more extensive inflector features could still use them. (For example, it looks like ActiveSupport tries to detect acronyms.)

Another option would be to do both: If ActiveSupport is present, use it, otherwise, use a local implementation. I wonder if being "smart" would just be annoying though, if people noticed it behaving differently at different times.

lib/graphql/schema/type_reducer.rb
```diff
@@ -1,10 +1,19 @@
 # frozen_string_literal: true
+
+require 'active_support/inflector'
+
 module GraphQL
   class Schema
-    module ReduceTypes
+    class TypeReducer
```
rmosolgo: Hmm I don't [see any third-party code](https://github.com/search?utf8=%E2%9C%93&q=ReduceTypes+language%3ARuby+extension%3Arb&type=Code&ref=advsearch&l=Ruby&l=) depending on this, but still, I've been surprised at my capacity to break code by changing things which I failed to mark `@api private` early on.

It seems like this _could_ be implemented in a backwards-compatible way, right? Adding `camelize: false` to `.reduce` and adding `camelize:` to all the functions it calls? It's definitely more verbose, but are there other downsides to that approach?

xuorig: Ah damn, you're right about backwards-compatibility 🤔 

Yeah I'll just add `camelize` to all functions not a big deal for now!

xuorig:
I like `transform_identifiers :camelize` a lot!

gjtorikian:
I like the concept, but as a team writing a schema grows, couldn't this devolve into a situation where various people are writing things in various cases? Even if the renamer magically takes care of it, it could still lead to a messy codebase. 

Instead, it seems like we ought to enforce that as the schema is being built. Field names not camelcased? Your schema is broken and invalid. We (cc @brandonblack) have been talking about submitting our test cases that validate naming conventions. If you agree on that direction @rmosolgo we can probably submit a PR here for it pretty easily.

xuorig:
>  it seems like we ought to enforce that as the schema is being built.

That'd be pretty cool. The hard thing with this is keeping the per-schema concept. Shared object types / you have no idea which schema you type is going to end up in.

> test cases that validate naming conventions

We love those too :)

rmosolgo:
@gjtorikian I totally agree with you: I'd much rather have the source code match the public API than introduce a transform which causes them to diverge. But this feature is often requested (I guess ppl are unwilling to right camel-case identifiers in Ruby, or unwilling to serve graphql in underscore case), and I'm capitulating 😆 

A framework for schema-level style assertion sounds like a win in either case, I'd love to see what yall have put together!

brandonblack:
I can see an argument both ways here, but my feedback would be to make the unpopular choice and push people towards the GQL way of doing things. 

Early on, we were pretty fixated on trying to make everything more Ruby-like in our GQL implementation. The goal was to make it seamless for consumers and contributors alike and ease on-boarding for folks new to GQL. 

Instead though, we made a conscious choice to push GQL concepts and conventions rather than obscure them with syntactical sugar and layers of indirection in Ruby. A year later, I can say that I think this was the right choice and the net effect of this approach is that everyone working with our implementation has a better understanding of the core, underlying GQL concepts.

While popular and definitely easy, I think the approach of letting schema contributors adopt any convention they want and then transforming that into GQL conventions will lead to much inconsistency and variety for schemas with many contributors.

If the goal is consistency, my recommendation would be to not support this feature and instead push validation that schema definitions adhere to already established GQL conventions. 

I'd be happy to turn what we've built internally into a rubocop extension for others to use if we want to go this direction and think that would be helpful. What we have currently was primarily built with the intention of helping eliminate a lot of overhead and repetition when reviewing new schema contributions (eg. CI enforces style for us and we can focus on the functional changes). I'm sure if we repacked it for broader consumption, this would be useful to others as well.

lib/graphql/schema/default_renamer.rb
```diff
@@ -0,0 +1,13 @@
+module GraphQL
+  class Schema
+    module DefaultRenamer
```
ianks: What are the performance costs of dispatching a method for renaming? I would imagine not much; but it would be good to know.

rmosolgo: This would be part of schema definition, not query execution, so personally, I'm not too concerned about it. If it becomes a problem, we can always add complexity in order to gain performance later. Or, if you'd like to check it out now, I could be convinced to prioritize performance _now_ instead of later by seeing a benchmark!

anazar:
any plans to merge this in?

graphql.gemspec
pabloh: You could also consider adding this gem as a dependency mbj/inflecto is way smaller that active-support, and used by many other projects.

pabloh:
Is there something that needs to be done before merging this?

rmosolgo:
I think there are a few things to address:

- Is this a good idea? (Personally I don't want it, I'm really into "what you see is what you get". But since many others _do_ want it and it's not the default, I'm willing to accept it.)
- Can we remove the dependency?

xuorig:
Im waiting for input on "is this a good idea" and then ill address the
concerns hehe

pabloh:
@rmosolgo, I'm ok with this feature if is deactivated by default, I would also suggest that we should allow the possibility to override this option it for particular, types mutation or fields, otherwise you may end up in cases where you have the same type camelized and not camelized is they are used on different schemas.

About the second bullet that's the easy part we could easy replace `active_support` by just bringing in string transformation code we need.

lib/graphql/schema/reduce_types.rb
```diff
@@ -28,32 +36,46 @@ def self.reduce_type(type, type_hash, context_description)
...
         end
         if type.kind.input_object?
-          type.arguments.each do |argument_name, argument|
-            reduce_type(argument.type, type_hash, "Input field #{type.name}.#{argument_name}")
+          type.arguments.keys.each do |argument_name|
+            original_argument = type.arguments.delete(argument_name)
+            argument = renamer.rename_argument(original_argument)
+
+            reduce_type(argument.type, type_hash, "Input field #{type.name}.#{argument_name}", renamer)
+            type.arguments[argument.name] = argument
```
theorygeek: Could probably save a little bit of work (delete and then re-insert the same thing) here by checking if the renamer isn't provided.

theorygeek:
Hmm hmm. I'm on the fence about it. I used to be _super_ annoyed with manually camelizing field names, but I guess not so much any more.

The only real concern I have is that `camelize` and `underscore` are not always true inverses of each other. For example:
```ruby
"address_1".camelize.underscore
# => "address1"

# Unless you add "QL" or "GraphQL" as an acronym in your inflections initializer...
"GraphQL".underscore.camelize
# => "GraphQl
```

So in order to really make this bullet-proof, the code would have to keep a hash of what the value was before/after renaming. I think today it'd be fine as-is, but I imagine it won't be too long before someone comes along with a PR that makes the renaming code way more complex in order to fix these scenarios. (Or worse, they come along with a GitHub _issue_, and someone else has to fix it 😆 )

I do think it's more important to look at this objectively (instead of just, "raise your hand if you like this feature"). The objective considerations for this feature I can think of are things like:
- How does it impact the maintainability of the codebase going forward? _Seems like it's probably not very big._
- If things go sideways on the end user, how hard will it be to figure out what's wrong? _Maybe needs some more work on this front._
- How does this impact runtime performance? _Not sure._

I think if I were going to decide, I'd probably say: 
- Let's add support for it in the library
- I'd probably make it possible to have a stateful renamer (pass in the `GraphQL::Field` or `GraphQL::Argument` object as well as the value to the renamer). The renamer can keep a hash of how it did the renaming if it wants to.
- And probably: the user has to supply the renamer themselves (but give a code example of how to build one using ActiveSupport)

All of that said, I don't think I'd end up using it myself. But maybe I will, you never know 😛  Plus I really love the contributions that the guys at Shopify (@xuorig and others) have made, and I don't want to be the guy who says no to a feature that they find useful 😄

xuorig:
Closing this! I think a IDL first approach or understanding GraphQL approach will serve better on long term :).

rmosolgo:
Thanks again for your work here, sorry it wasn't merged, but the conversation was really valuable!

emilebosch:
Awww, don't close! i'm pro camelizing. Now i need to map all my fields manually in mutations and such. Can we please just have this feature but turning it off by default? It doesn't feel very DRY to me. Furthermore i think that while it may help with onboarding people, it also enforces more  work onto people in the long run. ActiveRecord and Rails is full of magic/abstraction but its not magic when convention is understood.

Case 1: 

I really want my fields to be underscored in rails and camalized in JS. Lets take a practical example of a rails mutation.

```graphql
addVideoToPost(postId:2, video: {shortExcerpt: "Blabla", leaderVideo:"bla.mpg")
```

Normally i could just write in the resolver:

```ruby
resolve: (o,a,c) -> {
  post = post.find(args[:post_id]) 
  post.videos << Video.new(args[:video]) 
  post.save
}
```

So instead of mapping video and `shortExcerpt` to `short_excerpt` and `leaderVideo` to `leader_video`  things just work.

If i now need to manually map it and also need to change all the input fields, im also increasing LOC, and possible mapping errors. Also when i need to refactor that field, i need to change it in many places.

Case 2:

With IDL changes and upcoming resolver maps in 1.6 there is no way to add `property_name` or `as` in arguments anymore i.e. (because types are generated by `build_from_definition`) and this will result in lots of mapping code.

Also, i think that 80% of the people making a graphql api in ruby will also use activerecord, for me personally the one of the main reasons why i use ruby as graphql server and not go all JS.

Personally i'd go for camelizing with inflector (maybe lazing loading it if present, i..e no hard dependency).

pabloh:
I think doing the proper refactors to use an external renamer would be a good compromise.

xuorig:
So I thought about this a bit more.

Right now, schema definition and resolvers are defined in a single place,
at the same time.

If the gem was to evolve into something like Robert briefly talked about
the we would maybe have something like this:

- Define schema using the IDL (has to be camel case here since were
directly using the langage)
- Define resolution through either conventions, a resolver Map, classes,
objects, anything.

With this, I think we can have the proper mechanism for our resolution
layer to receive everything snake case if we want to, while keeping the
schema camel cased with the IDL.

I feel that then we get the best of both worlds: we teach proper graphql
because you need to know the langage to design your schema, but the ruby
resolvers & actual logic can use snake case since it will just be simple
ruby code!

What do you all think? If the gem doesnt end up going that way I'd be more
open to work on this PR but since schema definition might change im a bit
on the fence :)

emilebosch:
Think so to! In my pr already there is an adjustment to resolver map that uses idl and resolves scalars and unions.

I write IDL the Js way. In order to make it super workable id love to add snake case conversion/mapping. Which would using the args almost directly fit Active-record and the other coding snake_case standards in ruby world.

rmosolgo:
:+1: for applying camelization during `BuildFromDefinition`, I think you're right that it's a nice place to do it. It works around a lot of weirdness since all the _actual_ type objects are internal to the schema, not referenced in user code. 

Maybe `Schema.from_definition(defn, convert_case: true)` ?

mull:
So how are people actually handling camelCase<->snake_case now?

emilebosch:
:property and :as

mull:
@emilebosch thank you! :heart: I spent waaay too much time looking for middleware/custom definition solutions

timscott:
@emilebosch `as:` works for arguments, not for auto field resolution. Without middleware requires a ton of ugly resolvers for scalars that would normally auto-resolve, AFAIK.

rafbgarcia:
Sorry, not sure I get how to implement this case conversion, I've tried adding `as: :fieldName` but got `GraphQL::Field can't define 'as'` =s

@mull did you get it working?
@emilebosch mind explaining a bit more how to do that?

Thank y'all

theorygeek:
@rafbgarcia Try using `property:` on your `field`’s

rafbgarcia:
@theorygeek thanks, I've been using that one.

Does that means there's no way to do the auto-convert?

Thanks

rafbgarcia:
well, my (not so) temporary solution - until it doesn't work anymore - was to do

```rb
query = params[:query].underscore
```

in the controller :P 

I'd love to know if there's a better solution

PR #556, opened 2017-02-21
refactor(RelationConnection) check for next page when fetching items

rmosolgo:
The code is a mess, but I wanted to try this out, apparently this is how join-monster does it!

Before:

```sql
StarWars::Base Load (0.2ms)  SELECT  "bases".* FROM "bases" WHERE "bases"."id" IN (SELECT "bases"."id" FROM "bases" WHERE "bases"."faction_id" = ?) LIMIT ? OFFSET ?  [["faction_id", 2], ["LIMIT", 2], ["OFFSET", 0]]
(0.2ms)  SELECT COUNT(count_column) FROM (SELECT  1 AS count_column FROM "bases" WHERE "bases"."id" IN (SELECT "bases"."id" FROM "bases" WHERE "bases"."faction_id" = ?) LIMIT ? OFFSET ?) subquery_for_count  [["faction_id", 2], ["LIMIT", 3], ["OFFSET", 0]]
```

After: 

```sql
StarWars::Base Load (0.2ms)  SELECT  "bases".* FROM "bases" WHERE "bases"."id" IN (SELECT "bases"."id" FROM "bases" WHERE
```

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -17,7 +17,7 @@ module GraphQL
       end
 
       def has_next_page
-        !!(first && count(sliced_nodes.limit(limit + 1)) > limit)
+        !!(first && paged_nodes && @has_next_page)
       end
 
       def has_previous_page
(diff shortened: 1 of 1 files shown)


korbin:
Awesome work! Looks exactly like how they do it, should kill a ton of queries!

rmosolgo:
I realized there's still one place that the old `count` method is used. I'll have to figure out how to refactor that out.

rmosolgo:
I don't think I can remove that other `count()` -- it's for calculating the new offset when paginating from the end. There's no way to know how far backwards to go except by getting the whole length of the collection!

PR #475, opened 2017-01-06
test on Ruby 2.4

rmosolgo:
Let's see if this works... need to: 

- [x] Get updated versions of dev dependencies with native extensions 
- [x] Watch for any bugs that may affect us 
  - https://bugs.ruby-lang.org/issues/13107

PR #557, opened 2017-02-21
Ruby 2.4 is deprecating Bignum and Fixnum

daemonsy:
## What's up? 

Hi, we upgraded an app that uses graphql-ruby to **Ruby 2.4** and we're getting a deprecation message on the ID type. 

`/app/vendor/bundle/ruby/2.4.0/gems/graphql-1.4.4/lib/graphql/id_type.rb:9: warning: constant ::Fixnum is deprecated`

https://bugs.ruby-lang.org/issues/12005

## Missing things: 
- Tests: Seems like the `id_type_spec.rb` covered that pretty well, asserting the coercion of integers, doesn't feel there's a need to write a new case unless you feel otherwise
- This might break with Ruby < 2.4, so probably should have an accompanying `gemspec` change.

### DIFF
--- a/lib/graphql/id_type.rb
+++ b/lib/graphql/id_type.rb
@@ -6,7 +6,7 @@ GraphQL::ID_TYPE = GraphQL::ScalarType.define do
   coerce_result ->(value) { value.to_s }
   coerce_input ->(value) {
     case value
-    when String, Fixnum, Bignum
+    when String, Integer
       value.to_s
     else
       nil


rmosolgo:
Thanks for looking into this! I'm hoping to officially support Ruby 2.4 via #475 but it's been a bit of a struggle getting CI working across all Rubies. Does that branch look like it would cover your needs, too?

daemonsy:
Ahh, somehow miss that while browsing the PRs. Yes! that looks good, so far we've only seen one deprecation warning pop up. 

Closing this and seeing if there's anything I can contribute to #475

PR #560, opened 2017-02-22
Add support for functional overrides

gjtorikian:
This PR demonstrates how overriding function definitions might work. 

The DSL is a little bleh, but, given a `function: kwarg`, I couldn't think of any other way to support overrides other than passing in a block, too. I didn't think that supporting

``` ruby
field :overwrittenDescription, "A description", function: TestFunc.new
```

made much sense.

Happy to address any feedback or changes here. ✌️


rmosolgo:
Thanks, this a great start! I'm going to see if I can get full compatibility with the DSL ...

PR #562, opened 2017-02-22
Tweak lazy find sample code

dtaniwaki:
We should remove pending IDs once they get retrieved to remove redundant DB requests.
Also avoid duplicate IDs to reduce SQL length.


rmosolgo:
Makes sense to me, thanks!

rmosolgo:
It may take a few days to get this change on the website -- I have changes on master that won't apply until the next release, so I can't just `rake site:publish` :S

PR #565, opened 2017-02-24
Add DefinedObjectProxy#use

cjoudrey:
@rmosolgo @xuorig @dylanahsmith 

As per the discussion in https://github.com/Shopify/graphql-batch/pull/49, this adds a way to write plugins that have access to the DSL.

**Some example usages:**

```ruby
GraphQL::Schema.define do
  use GraphQL::Batch # This will call GraphQL::Batch.use(self)
end

module GraphQL
  module Batch
    # ...

    def self.use(schema_defn)
      schema_defn.lazy_resolve(Promise, :sync)
      schema_defn.instrument(:query, GraphQL::Batch::Setup)
    end
  end
end
```

You can also accept arguments and keyword arguments:

```ruby
GraphQL::Schema.define do
  # This will call GraphQLPro::Authorization.use(self, strategy: :devise)
  use GraphQLPro::Authorization, strategy: :devise
end
```

This will work on any object that supports `.define` so you could use it on object types and fields too:

```ruby
GraphQL::ObjectType.define do
  field :cheeses, !types[!Cheese] do
    use Pagination
    # ...
  end
end
```

**Questions:**

Plugins currently have no way of knowing the type of the object being defined.

We could either expose `DefinedObjectProxy#target` allowing the plugin to do:

```ruby
def use(schema_defn)
  if schema_defn.target.class != GraphQL::Schema
    raise "GraphQL::Batch plugin can only be used on GraphQL::Schema"  
  end

  # ...
end
```

Or we could pass the `target` as an argument to `use`.

Or any other ideas?

### DIFF
--- a/lib/graphql/define/defined_object_proxy.rb
+++ b/lib/graphql/define/defined_object_proxy.rb
@@ -11,6 +11,15 @@ module GraphQL
         GraphQL::Define::TypeDefiner.instance
       end
 
+      def use(plugin, **kwargs)
+        # https://bugs.ruby-lang.org/issues/10708
+        if kwargs == {}
+          plugin.use(self)
+        else
+          plugin.use(self, **kwargs)
+        end
+      end
+
       def method_missing(name, *args, &block)
         definition = @dictionary[name]
         if definition
--- a/lib/graphql/define/defined_object_proxy.rb
+++ b/lib/graphql/define/defined_object_proxy.rb
@@ -2,6 +2,8 @@
 module GraphQL
   module Define
     class DefinedObjectProxy
+      attr_reader :target
+
       def initialize(target)
         @target = target
         @dictionary = target.class.dictionary
--- a/lib/graphql/define/instance_definable.rb
+++ b/lib/graphql/define/instance_definable.rb
@@ -21,18 +21,22 @@ module GraphQL
     #
     # @example Make a class definable
     #   class Car
-    #     attr_accessor :make, :model
+    #     include GraphQL::Define::InstanceDefinable
+    #     attr_accessor :make, :model, :doors
     #     accepts_definitions(
     #       # These attrs will be defined with plain setters, `{attr}=`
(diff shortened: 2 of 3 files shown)


lib/graphql/define/defined_object_proxy.rb
```diff
@@ -11,6 +11,15 @@ def types
         GraphQL::Define::TypeDefiner.instance
       end
 
+      def use(plugin_class, *args, **kwargs)
```
rmosolgo: Is there a use case in mind for `*args`? If not, I think we should drop it until we see a clear need. I can't _think_ of anything where `**kwargs` wouldn't be sufficient for now.

lib/graphql/define/defined_object_proxy.rb
```diff
@@ -11,6 +11,15 @@ def types
...
+        # https://bugs.ruby-lang.org/issues/10708
+        if kwargs == {}
+          plugin_class.use(self, *args)
+        else
+          plugin_class.use(self, *args, **kwargs)
```
rmosolgo: We don't expect this to be a `Class`, right? Only something that responds to `use`?

gjtorikian:
Can we get some documentation for this too please?

cjoudrey:
I got sidetracked by a few things. I'll look into writing a guide for this feature shortly.

Any thoughts on `target` being passed to the plugin vs accessible via `defn.target`?

rmosolgo:
✋  my vote is for adding a `#target` method. It might conflict with a user definition but I doubt it!

If we get a bug report, we can make it backwards-compatible with something like 

```ruby 
def target(*args, &block)
  if args.any? || block 
    send_to_definition_function(args, block)
  else 
    @target 
  end 
end 
```

rmosolgo:
This is rad! Thanks!!

rmosolgo:
pushed yardoc in 7bfdb7dd

Issue #440, opened 2016-12-06
Feature ideas for graphql-ruby 2.0


theorygeek:
I'm not sure if there's already a backlog for upcoming work on the gem, but I thought I'd write out a couple of ideas that I've had that could improve the library, and get some feedback.

(I labeled this as 2.0 because I _think_ these are all breaking changes.)

## Passing arguments to resolver methods
Normally, field resolvers have access to three things:
- Backing-object for the type (eg, ActiveRecord model, hash, something similar)
- Arguments for the field
- Query context

But when you use the `property:` argument when declaring the field, none of these parameters are passed to the method being called:

```ruby
# some_method doesn't get the arguments or the context
field :some_field, !types.String, property: :some_method
```

But it'd be nice to be able to access those things from within the method. I can think of two ways this could work. First, they could be passed as-is, nothing fancy:
```ruby
# MethodResolve class
def call(obj, args, ctx)
  obj.public_send(@method_name, args, ctx)
end
```

Second, we could could actually pass in the arguments from the field as keyword arguments. This would feel a little more Ruby-like:
```ruby
# Example field definition:
field :some_field, !types.String, property: :some_method do
  argument :some_arg, !types.Int
  argument :some_other_arg, !types.String
end

# Example method:
def some_method(some_arg:, some_other_arg:)
end

# Or maybe the method takes context *and* arguments:
def some_method(ctx, some_arg:, some_other_arg:)
end
```

## Camelizing fields and arguments
It seems nearly ubiquitous in the GraphQL world to use camelCase for fields and arguments. But in Ruby, it's equally ubiquitous to use snake_case. I think these two things are so common, it probably warrants a feature in the gem to handle.

Perhaps it's something that would be configured:
```ruby
Schema.define do
  camelize true
end
```

If that's enabled, all fields and arguments would appear to be camelized to the outside world, but would be snake_case everywhere in application code.

It might be possible to do this right now, with the instrumentation API - I'm not sure. (If someone has solved this, it'd be awesome to see the code!) But it seems like it'd bubble into various parts of the library code (NameResolve, HashKeyResolve, for example).

Ideally, we'd be able to avoid situations like this:
```ruby
field :someField, !types.String, property: :some_field
```

## Helper methods on GraphQL types
Preface: Is this an anti-pattern?

[...]

PR #567, opened 2017-02-26
Passing arguments to root_value (and sub-fields) fields when they are methods.

f:
This will allow us to build method-like fields just like the example on http://graphql.org/graphql-js/.

With this, I will be able to use class methods as simple resolvers. It makes simple and modular usages possible and it's more intuitive.

```ruby
Todo = Struct.new(:text)

class Root
  attr_accessor :todos

  def initialize
    @todos = [
      Todo.new("Pay the bills."),
      Todo.new("Go to dentist."),
      Todo.new("Buy milk.")
    ]
  end

  def todoAdd(args)
    todo = Todo.new(args[:text])
    @todos << todo
    todo
  end

  def allTodos
    @todos
  end
end

app = Root.new

Schema.execute("mutation {
  todoAdd(text: \"Heyo\") {
     text
  }
}", root_value: app)

result = Schema.execute("query {
  allTodos {
    text
  }
}", root_value: app)
```

Since it covers all the method-like argument accepting fields, it will work on any level.

### DIFF
--- a/lib/graphql/field/resolve.rb
+++ b/lib/graphql/field/resolve.rb
@@ -9,7 +9,7 @@ module GraphQL
       # @return [Proc] A resolver for this field, based on its config
       def create_proc(field)
         if field.property
-          MethodResolve.new(field.property.to_sym)
+          MethodResolve.new(field)
         elsif !field.hash_key.nil?
           HashKeyResolve.new(field.hash_key)
         else
@@ -23,12 +23,17 @@ module GraphQL
 
       # Resolve the field by `public_send`ing `@method_name`
       class MethodResolve < BuiltInResolve
-        def initialize(method_name)
-          @method_name = method_name
+        def initialize(field)
+          @field = field
+          @method_name = field.property.to_sym
         end
 
         def call(obj, args, ctx)
-          obj.public_send(@method_name)
+          if @field.arguments.any?
+            obj.public_send(@method_name, args, ctx)
+          else
+            obj.public_send(@method_name)
+          end
         end
       end
 
@@ -51,20 +56,11 @@ module GraphQL
         end
 
         def call(obj, args, ctx)
-          if accepts_arguments?(obj)
+          if @field.arguments.any?
             obj.public_send(@field.name, args, ctx)
           else
             obj.public_send(@field.name)
           end
-        end
-
-        private
-        
-        def accepts_arguments?(obj)
-          !obj.is_a?(Struct) &&
-          !obj.is_a?(OpenStruct) &&
-          @field.arguments.any? &&
-          obj.method(@field.name).parameters.any?
         end
       end
     end
--- a/lib/graphql/field/resolve.rb
+++ b/lib/graphql/field/resolve.rb
@@ -35,7 +35,7 @@ module GraphQL
 
         def call(obj, args, ctx)
           if method_with_arguments?(@method_name, obj, args)
-            obj.public_send(@method_name, args)
+            obj.public_send(@method_name, args, ctx)
           else
             obj.public_send(@method_name)
           end
@@ -62,7 +62,7 @@ module GraphQL
 
         def call(obj, args, ctx)
           if method_with_arguments?(@field.name, obj, args)
-            obj.public_send(@field.name, args)
+            obj.public_send(@field.name, args, ctx)
           else
             obj.public_send(@field.name)
           end
--- a/lib/graphql/field/resolve.rb
+++ b/lib/graphql/field/resolve.rb
@@ -19,6 +19,12 @@ module GraphQL
 
       # These only require `obj` as input
       class BuiltInResolve
+        protected
+        def method_with_arguments?(method, obj, args)
+          args.to_h.keys.size > 0 &&
+          obj.methods.include?(method.to_sym) &&
+          obj.method(method).parameters.size > 0
+        end
       end
 
       # Resolve the field by `public_send`ing `@method_name`
@@ -28,7 +34,11 @@ module GraphQL
         end
 
         def call(obj, args, ctx)
-          obj.public_send(@method_name)
+          if method_with_arguments?(@method_name, obj, args)
+            obj.public_send(@method_name, args)
+          else
+            obj.public_send(@method_name)
+          end
         end
       end
 
@@ -51,7 +61,11 @@ module GraphQL
         end
 
         def call(obj, args, ctx)
-          obj.public_send(@field.name)
+          if method_with_arguments?(@field.name, obj, args)
+            obj.public_send(@field.name, args)
+          else
+            obj.public_send(@field.name)
+          end
         end
       end
     end
--- a/lib/graphql/field/resolve.rb
+++ b/lib/graphql/field/resolve.rb
@@ -19,12 +19,6 @@ module GraphQL
 
       # These only require `obj` as input
       class BuiltInResolve
-        protected
-        def method_with_arguments?(method, obj, args)
-          args.to_h.keys.size > 0 &&
-          obj.methods.include?(method.to_sym) &&
-          obj.method(method).parameters.size > 0
-        end
       end
 
       # Resolve the field by `public_send`ing `@method_name`
@@ -34,11 +28,7 @@ module GraphQL
         end
 
         def call(obj, args, ctx)
-          if method_with_arguments?(@method_name, obj, args)
-            obj.public_send(@method_name, args, ctx)
-          else
-            obj.public_send(@method_name)
-          end
+          obj.public_send(@method_name)
         end
       end
 
@@ -61,11 +51,20 @@ module GraphQL
         end
 
         def call(obj, args, ctx)
-          if method_with_arguments?(@field.name, obj, args)
+          if accepts_arguments?(obj)
             obj.public_send(@field.name, args, ctx)
           else
             obj.public_send(@field.name)
           end
+        end
+
+        private
+        
+        def accepts_arguments?(obj)
+          !obj.is_a?(Struct) &&
+          !obj.is_a?(OpenStruct) &&
+          @field.arguments.any? &&
+          obj.method(@field.name).parameters.any?
         end
       end
     end
--- a/spec/graphql/field_spec.rb
+++ b/spec/graphql/field_spec.rb
@@ -35,27 +35,30 @@ describe GraphQL::Field do
   it "passing arguments to root_value when fields accepting arguments" do
     schema = GraphQL::Schema.from_definition(" 
       type Todo {text: String}
-      type Query { allTodos: [Todo]}
-      type Mutation { todoAdd(text: String!): Todo}
+      type Query { all_todos: [Todo]}
+      type Mutation { todo_add(text: String!): Todo}
     ")
     Todo = Struct.new(:text)
     class RootResolver
       attr_accessor :todos
+
       def initialize
         @todos = [Todo.new("Pay the bills.")]
       end
-      def allTodos
+
+      def all_todos
         @todos
       end
-      def todoAdd(args) # this is a method and accepting arguments
+
+      def todo_add(args) # this is a method and accepting arguments
         todo = Todo.new(args[:text])
         @todos << todo
         todo
       end
     end
     root_values = RootResolver.new
-    schema.execute("mutation { todoAdd(text: \"Buy Milk\") { text } }", root_value: root_values)
-    result = schema.execute("query { allTodos { text } }", root_value: root_values)
+    schema.execute("mutation { todoAdd: todo_add(text: \"Buy Milk\") { text } }", root_value: root_values)
+    result = schema.execute("query { allTodos: all_todos { text } }", root_value: root_values)
     assert_equal(result.to_json, '{"data":{"allTodos":[{"text":"Pay the bills."},{"text":"Buy Milk"}]}}')
   end
 
--- a/spec/graphql/field_spec.rb
+++ b/spec/graphql/field_spec.rb
@@ -34,11 +34,11 @@ describe GraphQL::Field do
 
   it "passing arguments to root_value when fields accepting arguments" do
     schema = GraphQL::Schema.from_definition(" 
-      type Todo {text: String}
+      type Todo {text: String, from_context: String}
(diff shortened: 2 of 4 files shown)


emilebosch:
I like this too in order to make something like graphql-tools quick api: 

https://github.com/apollographql/graphql-tools

```
const schema = `
type Author {
  id: Int! # the ! means that every author object _must_ have an id
  firstName: String
  lastName: String
  posts: [Post] # the list of Posts by this author
}

type Post {
  id: Int!
  title: String
  author: Author
  votes: Int
}

# the schema allows the following query:
type Query {
  posts: [Post]
}

# this schema allows the following mutation:
type Mutation {
  upvotePost (
    postId: Int!
  ): Post
}

# we need to tell the server which types represent the root query
# and root mutation types. We call them RootQuery and RootMutation by convention.
schema {
  query: Query
  mutation: Mutation
}
`;

export default schema;
```
and then

```
onst resolverMap = {
  Query: {
    posts() {
      return posts;
    },
  },
  Mutation: {
    upvotePost(_, { postId }) {
      const post = find(posts, { id: postId });
      if (!post) {
        throw new Error(`Couldn't find post with id ${postId}`);
      }
      post.votes += 1;
      return post;
    },
  },
  Author: {
    posts(author) {
      return filter(posts, { authorId: author.id });
    },
  },
  Post: {
    author(post) {
      return find(authors, { id: post.authorId });
    },
  },
};

export default resolverMap;
```

f:
This is a sub-level example that it uses `Todo` class methods as resolvers:

```ruby
Schema = GraphQL::Schema.from_definition(" 
  type Todo {
    text: String
    subResolver(name: String!): String
  }

  type Query {
    allTodos: [Todo]
  }

  type Mutation {
    todoAdd(text: String!): Todo
  }
")

class Todo
  attr_accessor :text
  def initialize(text)
    @text = text
  end
  def subResolver(args)
    args[:name].upcase
  end
end

class RootResolver
  attr_accessor :todos

  def initialize
    @todos = [
      Todo.new("Pay the bills.")
    ]
  end
  def allTodos
    @todos
  end
  def todoAdd(args)
    todo = Todo.new(args[:text])
    @todos << todo
    todo
  end
end

app = RootResolver.new

Schema.execute("mutation { todoAdd(text: \"Buy milk\") { text } }", root_value: app)
Schema.execute("query { allTodos { id, text, subResolver(name: \"foo bar\") } }", root_value: app)
```

The result will be, `subResolver` will be resolved all `Todo` models:
```ruby
{"data"=>{
  "allTodos"=>[
    {"text"=>"Pay the bills.", "subResolver"=>"FOO BAR"},
    {"text"=>"Buy milk", "subResolver"=>"FOO BAR"}
  ]
}}
```

lib/graphql/field/resolve.rb
```diff
@@ -51,7 +61,11 @@ def initialize(field)
         end
 
         def call(obj, args, ctx)
-          obj.public_send(@field.name)
+          if method_with_arguments?(@field.name, obj, args)
+            obj.public_send(@field.name, args)
```
cjoudrey: I might be missing something, but wouldn't you want context too?

f: Yes we would, but in this kind of example I use class itself as context. I didn't feel any use case so I did not pass it. But anyhow we may need context.

cjoudrey: Ah yes, of course! 🙃 Passing it when you initialize the class works just fine.

xuorig:
I love the idea. I might be wrong, but `root_value` should probably be used to pass a business logic object (not a graphql resolver) to the root types?

Right now it feels like you're passing a resolver to a resolver, which makes me wonder if this is the right thing to do. Maybe I'm not getting the use case right but why not just define resolvers that way instead of passing an object that does exactly that?

For something like @emilebosch linked, I think you want a way to define your resolvers as classes, instead of passing custom objects to your resolvers 🤔 

I tried something a while ago: https://github.com/xuorig/graphql-tools-ruby/blob/master/spec/tools_spec.rb

Not convinced by my approach either but food for thought 💭

f:
@xuorig You are right, but actually my first approach was making the example on http://graphql.org/graphql-js/ was possible with GraphQL-Ruby.

Also this is not against GraphQL specs and it says Root resolvers may have arguments and context: http://graphql.org/learn/execution/#root-fields-resolvers

lib/graphql/field/resolve.rb
f: On specs (via http://graphql.org/learn/execution/#root-fields-resolvers) it says root resolvers may have context. Maybe I should pass the context to follow the specs, WDYT?

f: @cjoudrey I updated the pull request to send context beside arguments.

This is an example code that accepts arguments and context:
```rb
def todo_add(args, ctx) # this is a method and accepting arguments
    todo = Todo.new(args[:text], ctx[:context_value])
    @todos << todo
    todo
end
```

emilebosch:
I do like the resolver map approach too a lot of graphql-js and makes it all look a bit cleaner.

I can't see completely whats needed to implement this at the moment but i'd take that as the the way to go.

f:
@emilebosch 

Did not tested it but with this case what you are doing above with JS is similar to following code:

```
schema = GraphQL::Schema.from_definition("
type Author {
  id: Int! # the ! means that every author object _must_ have an id
  firstName: String
  lastName: String
  posts: [Post] # the list of Posts by this author
}

type Post {
  id: Int!
  title: String
  author: Author
  votes: Int
}

# the schema allows the following query:
type Query {
  posts: [Post]
}

# this schema allows the following mutation:
type Mutation {
  upvotePost (
    postId: Int!
  ): Post
}
")

class Post < ApplicationRecord
  belongs_to :author

  def upvote!
    update(votes: votes+1)
  end
end

class Author < ApplicationRecord
  has_many :posts
end

class App
  def upvotePost(args)
    Post.find_by_id(args[:postId]).upvote!
  end
  
  def posts
    Post.all
  end

  def run(query, variables = {}, context = {})
    Schema.execute(query, root_value: self, variables: variables, context = {})
  end
end

app = App.new
posts = app.run("query { posts }")
app.run("mutation { upvotePost(postId: 1) { title } }")
```

lib/graphql/field/resolve.rb
```diff
@@ -51,7 +61,11 @@ def initialize(field)
         end
 
         def call(obj, args, ctx)
-          obj.public_send(@field.name)
+          if method_with_arguments?(@field.name, obj, args)
+            obj.public_send(@field.name, args, ctx)
+          else
+            obj.public_send(@field.name)
+          end
```
rmosolgo: I expected to find a noticeable performance regression from this addition but it was very small (within the margin of error, so maybe nothing) 

```
~/code/graphql $ git co passing-arguments-to-root-fields
Switched to branch 'passing-arguments-to-root-fields'
Your branch is up-to-date with 'f/passing-arguments-to-root-fields'.
~/code/graphql $ be rake bench:query
               query     3.000  i/100ms
               query     38.469  (± 7.8%) i/s -    192.000  in   5.029717s
~/code/graphql $ be rake bench:query
               query     4.000  i/100ms
               query     39.447  (± 2.5%) i/s -    200.000  in   5.076236s
~/code/graphql $ be rake bench:query
               query     3.000  i/100ms
               query     35.082  (±20.0%) i/s -    168.000  in   5.036288s
Switched to branch 'master'
~/code/graphql $ be rake bench:query
               query     3.000  i/100ms
               query     39.593  (± 5.1%) i/s -    198.000  in   5.008667s
~/code/graphql $ be rake bench:query
               query     4.000  i/100ms
               query     40.610  (± 4.9%) i/s -    204.000  in   5.031973s
~/code/graphql $ be rake bench:query
               query     3.000  i/100ms
               query     40.077  (± 5.0%) i/s -    201.000  in   5.023988s
```

f:
In my humble opinion, `resolver_map` is really a good idea and intuitive. (But seems we'll need to wait for a while when someone implements it)

The two-level hash idea is also nice and useful, but more than two levels would be confusing, since we define all the types on flat, and nest them via GraphQL. (Couldn't find any use case to make them more-than-two-level resolver mapping)

In this case, I didn't really think about how can we cover `resolve_type` use cases when using `Union` and `Interface` types. How JS version does this?

Besides, this change covers a huge need when to compare the change itself (I believe).

Thanks! 💎

emilebosch:
In the docs of graphql-js it's done like this: (http://dev.apollodata.com/tools/graphql-tools/resolvers.html)

```
const resolverMap = {
  Vehicle: {
    __resolveType(obj, context, info){
      if(obj.wingspan){
        return 'Airplane';
      }

      if(obj.licensePlate){
        return 'Car';
      }

      return null;
    },
  },
};
```

f:
I also realized that **"root_value fields can be methods"** and **"resolvers can be two level of objects"** are different topics which seems close to each other. One is about **data** and the other one is about **logic**. The examples we give misleading us but maybe we should discuss about these in seperate issues/topics.

rmosolgo:
That's a good point about these being two different issues! Maybe I'm getting ahead of myself. A similar feature is requested here: https://github.com/rmosolgo/graphql-ruby/issues/440

Could we do this detection ahead-of-time instead of doing it at runtime? For example, if a field is defined to have any arguments, can we assume that we should provide `args, ctx` to the method? (My goal is to keep runtime work to a minimum.)

It would be a bit tough since the `Field` object is mutable, but if it seems like it would work, I bet we can find a way 😎

f:
It does make sense! And it was actually my first approach to achieve it. Generating a new class called `...ResolverWithArguments` and detect if the field have arguments and pass `args` and `ctx`. But I couldn't make it done and I switched to this to make it work. Maybe I can give it another try and let's see if I can make it now :)

f:
I think there's a problem with schema parsing, or I'm missing somethin on `field/resolve.rb`

```ruby
# Line 9...
      def create_proc(field)
        if field.arguments.size > 0 ## Field arguments.size is always 0
          if field.property
            MethodResolveWithArgs.new(field.property.to_sym)
          elsif !field.hash_key.nil?
            HashKeyResolveWithArgs.new(field.hash_key)
          else
            NameResolveWithArgs.new(field)
          end
        else
          if field.property
            MethodResolve.new(field.property.to_sym)
          elsif !field.hash_key.nil?
            HashKeyResolve.new(field.hash_key)
          else
            NameResolve.new(field)
          end
        end
      end
# ...
```

If I can achieve this, the argument-passing won't be on runtime but it will be defined on creating schema.

The strange part is,

```rb
schema = GraphQL::Schema.from_definition(" 
      type Todo {text: String, from_context: String}
      type Query { all_todos: [Todo]}
      type Mutation { todo_add(text: String!): Todo}
    ")
```

When I inspect this, I can see field arguments, but I cannot see them when defining fields. Do they assigned lazily or something?

f:
I got what's wrong. Resolver initialized before definition:

```rb
# field.rb
    def initialize
      @complexity = 1
      @arguments = {} # <-- no arguments
      @resolve_proc = build_default_resolver # <-- default resolver thinks there's no argument
      @lazy_resolve_proc = DefaultLazyResolve
      @relay_node_field = false
      @default_arguments = nil
    end
``` 

So according to default resolver, there's no arguments yet.

f:
@rmosolgo I created a **[branch](https://github.com/f/graphql-ruby/commit/d760de35714f3c706376be0fa679d214743b9122)** that checks arguments on definition and decide how resolver behave.

But this has another issue that we cannot detect DSL created arguments.

In following example, `todo_add` field cannot be detected as a field with arguments.
```rb
MutationType = GraphQL::ObjectType.define do
  name "MutationType"
  field :todo_add, TodoType do
    argument :text, !types.String
  end
end
```

rmosolgo:
:thinking: thanks for taking a crack at this, you're right about them being assigned lazily. It's all a bit weird!

Maybe the best we can do is simplify the runtime check. Do you think this would work? 

```ruby 
def call(obj, args, ctx)
  if @field.arguments.any?
    obj.public_send(@field.name, args, ctx) 
  else 
    obj.public_send(@field.name)
  end 
end 
```

f:
@rmosolgo I made the runtime usage minimum in that way -- I added Struct check since they are method-like but not actually accepting argument methods: 

```ruby
      class NameResolve < BuiltInResolve
        def initialize(field)
          @field = field
        end

        def call(obj, args, ctx)
          if accepts_arguments?(obj)
            obj.public_send(@field.name, args, ctx)
          else
            obj.public_send(@field.name)
          end
        end

        private
        
        def accepts_arguments?(obj)
          !obj.is_a?(Struct) &&
          !obj.is_a?(OpenStruct) &&
          @field.arguments.any? &&
          obj.method(@field.name).parameters.any?
        end
      end
```

rmosolgo:
Sorry for being a bit slow, but can you help me understand the use case for a GraphQL field that accepts arguments with a Ruby method that _doesn't_ accept arguments? For example, an arrangement like this: 


```
field :items, types[Types::ItemType] do 
  argument :limit, types.Int
end 

# Backed by: 

def items
  Item.all 
end 
```

Why should we support it? To me, it seems reasonable to say that if the field has arguments, we will pass them to the method. Otherwise, what's the use of having arguments? (Maybe there _is_ a use, but I can't think of it!)

f:
I don't want you to support that case and also this PR doesn't try to make it supported.

Simply you call `public_send` without passing arguments, so you drop the arguments when you call the method. That's the problem to be fixed I think :) The method will never know about the arguments.

Here is a simple use case, how can you make it possible?
```rb
Schema = GraphQL::Schema.from_definition("
    type Query {
      world(arg: Int): Int
    }
")

class Hello
  def world
    # I want to access argument `arg`, how? <--
  end
end

puts Schema.execute("query {
  world(arg: 1234)
}", root_value: Hello.new)
```

rmosolgo:
Sorry, I think I'm not expressing myself well! Here's what I'm suggesting: 

https://github.com/f/graphql-ruby/pull/1

It's possible that we raise some ArgumentErrors in that case, but I think it would be ok. What do you think of that implementation?

f:
This looks simple and slick :+1:

rmosolgo:
:metal:  thanks for making this happen!

PR #578, opened 2017-03-06
Additional specs for #575 (complementary case)

khamusa:
Proposing complementary specs for #575. 

I'm not aware whether or not this is already tested on other specs, but I thought it could lbe nice to ensure we're not messing with the other behavior. I also took the liberty of updating some spec descriptions that were created when we were still diagnosing the issue.


rmosolgo:
Great, thanks, I agree it's great to have those side-by-side!

PR #580, opened 2017-03-07
Fix(generators) rename base class so it's not a generator

rmosolgo:
Oops, this was showing as a false positive:

```
$ rails g graphql:type -h
Usage:
  rails generate graphql:type TypeName [options]

Runtime options:
  -f, [--force]                    # Overwrite files that already exist
  -p, [--pretend], [--no-pretend]  # Run but do not make any changes
  -q, [--quiet], [--no-quiet]      # Suppress status output
  -s, [--skip], [--no-skip]        # Skip files that already exist

Description:
    Create graphql files for type generator.
```

now, appropriately: 

```
$ rails g graphql:type -h
Could not find generator 'graphql:type'. Maybe you meant 'graphql:object', 'graphql:loader' or 'graphql:union'
Run `rails generate --help` for more options.
```

cc @cjoudrey

### DIFF
--- a/lib/generators/graphql/interface_generator.rb
+++ b/lib/generators/graphql/interface_generator.rb
@@ -9,7 +9,7 @@ module Graphql
     # ```
     # rails g graphql:interface NamedEntityType name:String!
     # ```
-    class InterfaceGenerator < TypeGenerator
+    class InterfaceGenerator < TypeGeneratorBase
       desc "Create a GraphQL::InterfaceType with the given name and fields"
       source_root File.expand_path('../templates', __FILE__)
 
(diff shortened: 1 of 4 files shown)


cjoudrey:
:heart: :+1:

I must be missing something super obvious here, but why does renaming to `TypeGeneratorBase` magically solve the issue?

rmosolgo:
I think it's Rails ✨  magic ✨ , it takes:

```
rails generate #{namespace}:#{action}
```

And then:

- Loads the generator: `require "#{namespace}/generators/#{action}_generator"`
- Finds a constant named `#{namespace.camelize}/#{action.camelize}Generator`
- Executes that constant as a generator (somehow with Thor, I don't quite get that part)

So, from Rails' perspective, `graphql:type` _was_ a perfectly good generator 😆

PR #584, opened 2017-03-08
feat(Schema.from_definition) accept custom resolve behavior for runnable schemas

rmosolgo:
With this patch: 

- Previous behavior for DSL-created schemas is unchanged 
- New default behavior for IDL-created schemas is unchanged 
- Custom execution for IDL-created schemas is possible 

I've included an example in the specs how field resolution can be specified any way you can imagine. In that example, resolve functions are provided in a two-level map, `{ type_name => { field_name => resolve_fn } }`. What do you think, is this a good fit?

Also, lexical scopes are fun :) 

cc/ @f @tberman @theorygeek

### DIFF
--- a/lib/graphql/field/resolve.rb
+++ b/lib/graphql/field/resolve.rb
@@ -24,16 +24,11 @@ module GraphQL
       # Resolve the field by `public_send`ing `@method_name`
       class MethodResolve < BuiltInResolve
         def initialize(field)
-          @field = field
           @method_name = field.property.to_sym
         end
 
         def call(obj, args, ctx)
-          if @field.arguments.any?
-            obj.public_send(@method_name, args, ctx)
-          else
-            obj.public_send(@method_name)
-          end
+          obj.public_send(@method_name)
         end
       end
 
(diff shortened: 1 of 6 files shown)


rmosolgo:
I'm also interested to see how this could support convention-based GraphQL schemas. For example, you could define the schema in IDL along with a class structure (accessible by Ruby constants) and it would "just work". (All you have to do is write that callable for `default_resolve:`!)

lib/graphql/field/resolve.rb
```diff
@@ -24,16 +24,11 @@ class BuiltInResolve
...
           @method_name = field.property.to_sym
         end
 
         def call(obj, args, ctx)
-          if @field.arguments.any?
-            obj.public_send(@method_name, args, ctx)
-          else
-            obj.public_send(@method_name)
-          end
+          obj.public_send(@method_name)
```
theorygeek: Nice, I definitely like changing this back.

theorygeek:
This looks nice! It looks like a good solution for IDL-defined schemas. I'm still hoping to find a way to bring similar functionality to DSL-defined schemas; some sort of way to swap out the `BuiltInResolve` behavior. Hoping to PR for it soon!

rmosolgo:
I couldn't help myself, I pushed support first-class support for a two-level hash 😊

f:
This is awesome!

PR #588, opened 2017-03-09
Update Relay mutation docs for new RangeAdd method

sgwilym:
Updates the Relay mutation docs so that it goes into a little more detail about the fields that certain Relay mutations expect, and updates the comment mutation example so that it uses the new `GraphQL::Relay::RangeAdd` API.


rmosolgo:
:metal: thanks!

sgwilym:
Always a pleasure!

PR #598, opened 2017-03-12
Hung query?

rmosolgo:
Trying to replicate an issue on 1.5.0

[Schema definition](https://github.com/rmosolgo/graphql-ruby/blob/b596ab852294d9068bcb68938108960157252c04/spec/graphql/static_validation/definition_dependencies_spec.rb#L18-L47), [Query from Relay](https://github.com/rmosolgo/graphql-ruby/blob/b596ab852294d9068bcb68938108960157252c04/spec/graphql/static_validation/definition_dependencies_spec.rb#L57-L100), [assertions](https://github.com/rmosolgo/graphql-ruby/blob/b596ab852294d9068bcb68938108960157252c04/spec/graphql/static_validation/definition_dependencies_spec.rb#L104-L117)

I could use some more info: 

- Definition of `Comment` & `Commentable` type in Ruby or in GraphQL IDL 
- If you use `schema.validate`, does it still hang? (This method is in 1.5.0 only)
- Could you check my schema & query to make sure I didn't overlook something from the earlier report?

### DIFF
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -27,10 +27,7 @@ module GraphQL
         end
 
         def call(type, field, obj, args, ctx)
-          @resolve_hash
-            .fetch(type.name)
-            .fetch(field.name)
-            .call(obj,args,ctx)
+          @resolve_hash[type.name][field.name].call(obj,args,ctx)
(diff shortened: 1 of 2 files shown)


rmosolgo:
Addressed in #599

PR #600, opened 2017-03-12
Large Schema to test fix to the new fragment resolution

jrhicks:
1.5.0 code had a couple of bugs that resulted in GIANT arrays of the same AST node, I've obfuscated  the schema that was originally triggering the issue so it can be included in tests.

### DIFF
--- a/spec/graphql/static_validation/fragment_resolution_spec.rb
+++ b/spec/graphql/static_validation/fragment_resolution_spec.rb
@@ -1,0 +1,1894 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe GraphQL::StaticValidation::DefinitionDependencies do
+  describe "fragment resolution bug GIANT arrays of the same AST node" do
+    let(:resolvers) {
+      resolvers Hash.new
+    }
+    let(:schema_defn) { <<-GRAPHQL
+      schema {
+  query: Query
+  mutation: RootMutation
+}
+type Ac10339026 implements Node, Re17665062, Co68529874 {
+  id: ID!
+  ra52455944: Int!
+  to39952341: String!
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks! This is rad. I really want to adapt this into the current benchmark code. There were to factors in the query that exposed the bug: 

- Fragments on abstract types 
- Lots of implementers of that abstract type 

In order to fix #418, I track the current scope as a set of _concrete_ object types, which means finding each implementor of the interface and preparing the execution path for it. This greatly simplified query execution and improved spec compliance, but came with a runtime penalty. I backported the benchmark tools to v1.4.5 and saw that validating this query increased from ~1.2ms to ~3ms:

```
# backported 1.4.5:
~/code/graphql $ be rake bench:validate
            validate    83.000  i/100ms
            validate    844.240  (± 7.7%) i/s -      4.233k in   5.044954s
# 1.5.0: 
~/code/graphql $ be rake bench:validate
            validate    29.000  i/100ms
            validate    301.780  (± 7.6%) i/s -      1.508k in   5.024896s
```

Although it's a regression (:cry:) I don't consider it a show-stopper because the total amount of time involved is quite small and I speculate that we would see end-to-end gains in query execution (but I haven't prepared a benchmark for that). 

I want to address this regression in two ways: 

- Add a schema & query like this one to the benchmark tooling to detect future regressions
- Fix the regression by more subtle scope handling during `Rewrite`. I think I can remove the overhead by getting tricky with sets of types 😅 

I'll keep this PR open in the meantime to track progress, thanks again for this contribution!

rmosolgo:
Thanks again for your help tracking this down. I've added a test schema & fixed the regression in #622 which I'll release shortly in 1.5.4!

Issue #590, opened 2017-03-09
Fuzzing: Relay Connection Argument Exceptions


stephencelis:
Giving junk arguments for Relay pagination results in uncaught exceptions. _E.g._,

``` graphql
me {
  followers(first: -1) { # or `last: -1`
    nodes {
      id
    }
  }
}
```

This line is the culprit:

https://github.com/rmosolgo/graphql-ruby/blob/a15142a0c176df7ef9da9dc216eba87db7428527/lib/graphql/relay/array_connection.rb#L26

Not a big deal, but we probably don't want to completely blow up on bad client requests.

rmosolgo:
Yes, I agree that client errors shouldn't result in Ruby exception! Thanks for reporting this one.

rmosolgo:
This'll be fixed in 1.5.2!

stephencelis:
🙌

PR #612, opened 2017-03-16
fix(BaseConnection) replace negative pagination args with 0

rmosolgo:
This scrubs out `first: -1` before it hits the connection implementation. In a subsequent PR I may apply max_page_size at the same time, but that will be a bit bigger change.


rmosolgo:
Fixes #590

PR #614, opened 2017-03-17
Provide a mechanism to wrap around the execution

gjtorikian:
I mentioned this briefly in Slack. In general, prior to a query being executed, I need to find a way to determine ultimately which type of query is being acted on, and then do something. Specifically, it would be the difference between selecting a read-only database versus one for writes, like:

``` ruby
Database.use_readonly_db do
  # execute query
end

Database.use_write_db do
  # execute mutation
end
```

There's a way to do this by looking at the `operation_name` and parsing the query for `OperationDefinition `, but it felt very cumbersome, and I wished the gem had a mechanism for it. I imagine this to work as follows:

``` ruby
execute_wrapper = Proc.new do |query, resolver|
                                   if query.mutation?
                                     Database.use_write_db do
                                       return resolver.call
                                      end
                                   else
                                     Database.use_readonly_db(resolver) do
                                       return resolver.call
                                      end
                                   end
                                 end
```

My test fixture for `execute_wrapper` is just a goofy little example because I didn't care about the results at all. I just wanted to show that the query type is being passed and an execution is being made.

Happy to go into more detail or change any aspect of this as necessary!

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -6,19 +6,23 @@ module GraphQL
       PROPAGATE_NULL = :__graphql_propagate_null__
 
       def execute(ast_operation, root_type, query)
-        yield query.before_execute.call(query) unless query.before_execute.nil?
+        resolver = Proc.new do
+          resolve_selection(
+            query.root_value,
+            root_type,
+            query.irep_selection,
+            query.context,
+            mutation: query.mutation?
+          )
+        end
 
-        result = resolve_selection(
-          query.root_value,
-          root_type,
-          query.irep_selection,
-          query.context,
-          mutation: query.mutation?
-        )
+         if query.execute_wrapper.nil?
+           result = resolver.call
+         else
+           query.execute_wrapper.call(query) { result = resolver.call }
+         end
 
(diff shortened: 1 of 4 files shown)


rmosolgo:
Would it be possible to accomplish this without a block? If so, could you use `instrument(:query, MyQueryInstrumenter)` with 

```ruby 
module MyQueryInstrumenter 
  module_function 
  def before_query(query) 
    if query.mutation? 
      begin_permit_writes
    else 
      begin_read_only 
    end 
  end 

  def after_query(query) 
    if query.mutation?
      end_permit_writes 
    else 
      end_read_only
    end 
  end 
end 
```

?

I didn't strive for a block-based API (`around_query` ?) in the current implementation because I wasn't confident I could really support it in a (future) world with streaming responses. But if you'd like to add a block-based API, do you think we could fit it into the current instrumentation API?

gjtorikian:
Oh, shucks! `{before|after}_query` would certainly do exactly what I need. Thanks!

> But if you'd like to add a block-based API, do you think we could fit it into the current instrumentation API?

I _could_, but, it's not required for what I need, and given what you said I'd rather we focus on getting streaming working first before adding something to the gem that might complicate that.

PR #619, opened 2017-03-19
Make sure the variant name of an enum is a string

hibariya:
Hello. Thank you for implementing & maintaining graphql-ruby.

I'd like to ensure that names of enum values are strings. I think that passing a symbol for the 2nd argument (`name`) is right since it is also used for the default value of `value`. However, the variant name can't be other than a string in GraphQL-land.


rmosolgo:
:+1: Makes sense to me, thanks!

PR #621, opened 2017-03-21
fix(InternalRepresentation#Node) add reference to parent node

rmosolgo:
oops, I removed this API unwittingly in #512 

cc @gjtorikian


rmosolgo:
It's still buggy https://github.com/rmosolgo/graphql-ruby/compare/frag-bench...frag-parent

PR #622, opened 2017-03-21
benchmark(Rewrite) add a case for fragments on abstract types

rmosolgo:
This is a case that experienced a regression in `1.5.x`. (It was a showstopper before #599). The problem is that Rewrite treats each fragment as a _new_ scope, and reifies it anew. 

1.5.x

```
~/code/graphql $ be rake bench:validate
Warming up --------------------------------------
validate - introspection
                        31.000  i/100ms
validate - abstract fragments
                        24.000  i/100ms
Calculating -------------------------------------
validate - introspection
                        331.501  (± 5.7%) i/s -      1.674k in   5.067422s
validate - abstract fragments
                        241.949  (± 5.4%) i/s -      1.224k in   5.073226s

Comparison:
validate - introspection :      331.5 i/s
validate - abstract fragments:      241.9 i/s - 1.37x  slower
```

1.4.x

```
~/code/graphql $ be rake bench:validate
Warming up --------------------------------------
validate - introspection
                        27.000  i/100ms
validate - abstract fragments
                        63.000  i/100ms
Calculating -------------------------------------
validate - introspection
                        270.351  (± 3.7%) i/s -      1.350k in   5.001078s
validate - abstract fragments
                        629.590  (± 4.3%) i/s -      3.150k in   5.012732s

Comparison:
validate - abstract fragments:      629.6 i/s
validate - introspection :      270.4 i/s - 2.33x  slower
```

I'm still noodling on exactly how to fix it...

PR #624, opened 2017-03-21
refactor(Rewrite) use abstract types as scopes during rewrite

rmosolgo:
An attempt to address the regression identified by #622. I've got some positive movement on the worst case, but a big regression in the best case: 

```
~/code/graphql $ be rake bench:validate
validate - introspection
                        19.000  i/100ms
validate - abstract fragments
                        36.000  i/100ms
validate - introspection
                        198.033  (± 7.6%) i/s -    988.000  in   5.017588s
validate - abstract fragments
                        371.484  (± 6.7%) i/s -      1.872k in   5.062809s

Comparison:
validate - abstract fragments:      371.5 i/s
validate - introspection :      198.0 i/s - 1.88x  slower
```

There's some wanton `node.dup`s in there that should be optimized, I think in many cases, it's safe to use the same node over again (leaf nodes on concrete owner types, perhaps others).

### DIFF
--- a/benchmark/run.rb
+++ b/benchmark/run.rb
@@ -32,12 +32,17 @@ module GraphQLBenchmark
   def self.profile
     # Warm up any caches:
     SCHEMA.execute(document: DOCUMENT)
+    CARD_SCHEMA.validate(ABSTRACT_FRAGMENTS)
 
     result = RubyProf.profile do
-      SCHEMA.execute(document: DOCUMENT)
+      CARD_SCHEMA.validate(ABSTRACT_FRAGMENTS)
+      # SCHEMA.execute(document: DOCUMENT)
     end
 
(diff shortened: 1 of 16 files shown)


rmosolgo:
Ok, did some refactoring, looks like we're back at 1.4.x levels 😎 

```
~/code/graphql $ be rake bench:validate
validate - introspection
                        29.000  i/100ms
validate - abstract fragments
                        62.000  i/100ms
validate - introspection
                        301.424  (± 6.0%) i/s -      1.508k in   5.021396s
validate - abstract fragments
                        659.310  (± 5.9%) i/s -      3.286k in   5.001450s

Comparison:
validate - abstract fragments:      659.3 i/s
validate - introspection :      301.4 i/s - 2.19x  slower
```

PR #625, opened 2017-03-21
Fix docs about returning connections in mutations

polmiro:
Noticed an inconsistency in the documentation about returning connections in a mutation.


rmosolgo:
Thanks! I'll push that right up!

Issue #628, opened 2017-03-22
Runtime error mutating lazy method map on JRuby


tjsousa:
While using `graphql-ruby` with JRuby 9.1.0.7 we're seeing what it looks like a concurrency issue (only happens randomly but it is frequent enough to become problematic) when calling `Graphql.execute` inside the lib's code (and before executing any application-specific code).

I'll try to dig a bit deeper into this, but at first sight [it doesn't look like this code is thread-safe](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/lazy/lazy_method_map.rb#L12-L19). The problem is that multiple threads can be calling this code concurrently and while one is iterating (`each` and `each_key` are called in this class), another may be mutating the hash.

Here's a simpler snippet that reproduces the error
```ruby
irb(main):057:0>  $hash = Hash.new do |h, key|
irb(main):058:1*   value = h.each_key.find { |key| sleep 1 }
irb(main):059:1>   if value.nil?
irb(main):060:2>     h[key] = nil
irb(main):061:2>   else
irb(main):062:2*     h[key] = value
irb(main):063:2>   end
irb(main):064:1> end
irb(main):112:0> 100.times do
irb(main):113:1*    Thread.new do
irb(main):114:2*      begin
irb(main):115:3*        $hash[rand()]
irb(main):116:3>     rescue => e
irb(main):117:3>       puts e
irb(main):118:3>     end
irb(main):119:2>   end
irb(main):120:1> end
=> 100
irb(main):082:0> can't add a new key into hash during iteration
can't add a new key into hash during iteration
can't add a new key into hash during iteration
can't add a new key into hash during iteration
can't add a new key into hash during iteration
...
irb(main):122:0* $hash.size
=> 4
```

Exception trace:
<img width="1231" alt="screen shot 2017-03-22 at 13 48 13" src="https://cloud.githubusercontent.com/assets/389917/24200841/410f0c96-0f06-11e7-9310-59834ac8baa2.png">

tjsousa:
My question here is if the `Schema` is supposed to be thread-safe, or should I have one per thread anyway?

Using a [`Concurrent::Hash`](http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Hash.html) should be enough to solve this in the lib:

```ruby
irb(main):056:0> require 'concurrent'
irb(main):057:0>  $hash = Concurrent::Hash.new do |h, key|
[...]

PR #631, opened 2017-03-22
fix(LazyMethodMap) wrap hash modification with a mutex

rmosolgo:
This mapping of `class => lazy_method_name` was not threadsafe because two threads could run the hash's default value block at the same time, resulting in a modification during enumeration, which is not allowed. 

So, I've wrapped the modifications in a mutex. Key _hits_ in `.get` won't be synchronized, is that safe?

I originally wrote a test for this: 

```ruby 
# frozen_string_literal: true
require "spec_helper"

describe GraphQL::Execution::Lazy::LazyMethodMap do
  let(:lazy_method_map) { GraphQL::Execution::Lazy::LazyMethodMap.new }
  it "handles multithreaded access" do
    a = Class.new
    b = Class.new(a)
    c = Class.new(b)
    lazy_method_map.set(a, :a)
    1000.times.map do |i|
      Thread.new {
        d = Class.new(c)
        lazy_method_map.get(d)
      }
    end.map(&:join)
  end
end
```

But on CRuby, I couldn't make it fail without putting an explicit `Thread.pass` in the `.find` block. So I didn't commit the test, since it didn't _really_ work. Is there another way I could test this?

### DIFF
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency "benchmark-ips"
   s.add_development_dependency "codeclimate-test-reporter", "~>0.4"
+  s.add_development_dependency "concurrent-ruby", "~>1.0"
   s.add_development_dependency "guard", "~> 2.12"
   s.add_development_dependency "guard-bundler", "~> 2.1"
   s.add_development_dependency "guard-minitest", "~> 2.4"
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -42,6 +42,4 @@ Gem::Specification.new do |s|
 
   # website stuff
   s.add_development_dependency "github-pages"
-
-  s.add_dependency 'concurrent-ruby', '~> 1.0'
 end
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -42,4 +42,6 @@ Gem::Specification.new do |s|
 
   # website stuff
   s.add_development_dependency "github-pages"
+
+  s.add_dependency 'concurrent-ruby', '~> 1.0'
 end
--- a/lib/graphql/execution/lazy/lazy_method_map.rb
+++ b/lib/graphql/execution/lazy/lazy_method_map.rb
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 require 'thread'
+begin
+  require 'concurrent'
+rescue LoadError
+  # no problem, we'll fallback to our own map
+end
 
 module GraphQL
   module Execution
@@ -67,6 +72,19 @@ module GraphQL
           def compute_if_absent(key)
             @semaphore.synchronize {
               @storage.fetch(key) { @storage[key] = yield }
+            }
+          end
+
+          def initialize_copy(other)
+            @semaphore = Mutex.new
+            @storage = other.copy_storage
+          end
+
+          protected
+
+          def copy_storage
+            @semaphore.synchronize {
+              @storage.dup
             }
           end
         end
--- a/lib/graphql/execution/lazy/lazy_method_map.rb
+++ b/lib/graphql/execution/lazy/lazy_method_map.rb
@@ -1,29 +1,19 @@
 # frozen_string_literal: true
 require 'thread'
 
+
 module GraphQL
   module Execution
     class Lazy
       # {GraphQL::Schema} uses this to match returned values to lazy resolution methods.
       # Methods may be registered for classes, they apply to its subclasses also.
       # The result of this lookup is cached for future resolutions.
+      # Instances of this class are thread-safe.
       # @api private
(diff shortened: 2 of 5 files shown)


ivoanjo:
Ok so I would say that for sharing a partially-unsynchronized hash instance as you're doing here the issues that may be observed are:

1. initializing the same key/value multiple times (as testing before writing is not synchronized) --- this also entails computing the initialization multiple times
2. reading some key and not finding some value because underlying data structures are being updated by other thread)
3. corrupting the underlying hash or causing exceptions as we saw in #628 

I believe this PR does fix 3). I'm not entirely sure that 1) and 2) are covered; yet, in this case since this is basically a cache, there's no problem if you don't find a key, because that just means you'll compute the value; and same for initializing the value multiple times, and this may be rare enough that you don't care.

I would say that the main thing to look out for in this approach is that when the cache is cold (at application start), locking while iterating means that all threads will be competing for the iteration, so performance may suffer for a bit. They may even all at the same time request the same key and repeat the iteration and initialization for the same class, one at a time.

As i'm a big fan of concurrent-ruby (even contributed a few patches), I would recommend instead using a [`Concurrent::Map`](https://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Map.html) (not `Concurrent::Hash`) to back your `@storage`. As per the documentation

> Concurrent::Map is a hash-like object and should have much better performance characteristics, especially under high concurrency, than Concurrent::Hash.

`Map` allows you to solve all three issues I pointed out above, as it's always Ok to iterate and to modify, and furthermore you can use `compute_if_absent` to ensure that you only initialize each value once (e.g. it solves problem 1 completely).

In terms of performance use, on MRI the result is probably similar as your Hash but for JRuby, TruffleRuby or Rubinius `Map` provides the best performance.

But yeah, on the other side, it's an added dependency. I think it's really worth it (:wink::wink::wink:), but whatever you go for, it's awesome to see a fix!

rmosolgo:
I tried writing separate storage backends (`HashStorage` and `ConcurrentMapStorage`) but I got a bit hung up looking for docs on `Concurrent::Map` :S Theres http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Map.html but it doesn't include any method docs

rmosolgo:
I think most CRuby servers are forking, so I wonder if these cached results are really short-lived anyways, since the underlying structures are copied whenever they're modified: 

```ruby
# $ irb
require "graphql"
# => true
map = GraphQL::Execution::Lazy::LazyMethodMap.new
# => #<GraphQL::Execution::Lazy::LazyMethodMap:0x007ffcf9a5af98 @storage={}>
map.set(Integer, nil)
# => nil
map.instance_variable_get(:@storage).keys
# => [Integer]
fork { map.set(String, nil); p map.instance_variable_get(:@storage).keys }
# => [Integer, String]
map.instance_variable_get(:@storage).keys
# => [Integer]
```

The forked map gets a new key, but the original is unchanged. I think forked processes _are_ reused, so each forked process gets a cache and uses that.

ivoanjo:
At @Talkdesk we mostly use [puma](http://puma.io/) which allows a mix of both threads and processes on our MRI systems (only allows threads on JRuby).

As for the map, you're right, the documentation is very lacking :/ . It is mostly modeled after https://docs.oracle.com/javase/8/docs/api/java/util/Map.html but yeah it wouldn't be the first time I had to look at the source to make sure I'm using something right.

I should have some time today, I'll take a stab at implementing this with a map so that you can evaluate both choices.

ivoanjo:
I've opened an alternate PR ( #645 ) using `Concurrent::Map` so we can compare both and decide. Let me know what you think :)

rmosolgo:
Thanks, that's really helpful! 

I've cherrypicked that commit here and added a fallback strategy to Hash + Mutex. How does it look to you?

ivoanjo:
Looks great! I would only suggest two improvements:

* Try to load concurrent-ruby if it is around:

```ruby
begin
  require 'concurrent-ruby'
rescue LoadError
  # no problem, we'll fallback to our own map
end

module GraphQL
  module Execution
    # ...
```

This way we make sure that the only thing needed by clients to use concurrent-ruby is to add it to their `Gemfile`. Otherwise it may be confusing that this gem may use different backends depending on if some other file that required `concurrent-ruby` was around or not.

* I am not entirely sure in what cases do a `LazyMethodMap` is  `dup`'d, but it may be safer for the `ConcurrentishMap` to implement a dup that grabs the lock

```
def dup
  @semaphore.synchronize { super }
end
```
(or something similar)

Other than that :+1: and feel free to close my PR :)

rmosolgo:
:+1: will do, good catch on `#dup`, it has to go one level deeper for concurrent-ish map!

tjsousa:
Awesome work! Thank you both

PR #633, opened 2017-03-23
Give mutations a more prominent spot in the docs, and add nested mutations

emilebosch:
Hey all,

Maybe mutations need to be more in the spotlight. Since its super common to do so, also i've added nested mutations which are super hawt 🙌 .

Mutations code /w help from @theorygeek.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,166 @@
+Robert Mosolgo <rdmosolgo@gmail.com>
+Christian Joudrey <cmallette@gmail.com>
+Garen Torikian <gjtorikian@gmail.com>
+Joshua Peek <josh@joshpeek.com>
+Dylan Thacker-Smith <Dylan.Smith@shopify.com>
+Marc-André Giroux <mgiroux0@gmail.com>
+theorygeek <theorygeek@gmail.com>
+charlieschwabacher <charlieschwabacher@gmail.com>
+Dave Qorashi <dave@zeemee.com>
+Evan Huus <eapache@gmail.com>
+mvgijssel <maarten@hackerone.com>
+Chris Hanks <christopher.m.hanks@gmail.com>
+Jérémie Astori <jeremie@astori.fr>
+Scott Walkinshaw <scott.walkinshaw@gmail.com>
+Gaurav Tiwari <gaurav@gauravtiwari.co.uk>
+kliu <kliu@tribpub.com>
+Samuel Gwilym <sam@comique.co>
+Brian Krausz <bkrausz@stripe.com>
+Charlie Somerville <charlie@charliesomerville.com>
+alex <alesssandro@ukr.net>
+Kristian Kvarnsbacke <kristian@kmja.com>
+Todd Berman <tberman@github.com>
+Stanisław Chmiela <sjchmiela@gmail.com>
+Willian van der Velde <mail@willian.io>
+Keith Pitt <me@keithpitt.com>
+Fatih Kadir Akın <fakin@protein.tech>
+jden <jden@zendesk.com>
+Juan Lulkin <maiz@lulk.in>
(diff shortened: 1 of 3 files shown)


guides/schema/mutations.md
```diff
@@ -0,0 +1,65 @@
...
+  # Create an input type for each item
+  TodoItemInputObjectType = GraphQL::InputObjectType.define do
+    name "TodoItem"
+    input_field :name, !types.String
+    input_field :starred, !types.Boolean
+  end
+
+  # Mutation takes an array of them:
+  input_field :todos, !types[!TodoItemInputObjectType]
+    end
```
Willianvdv: Unexpected end

emilebosch: I'll fix it in a minute and squash the commits 👍  thanks for catching bit of a hasty thing

guides/schema/mutations.md
```diff
@@ -0,0 +1,65 @@
...
+  name "AddTodoList"
+ 
+  # Create an input type for each item
+  TodoItemInputObjectType = GraphQL::InputObjectType.define do
+    name "TodoItem"
+    input_field :name, !types.String
+    input_field :starred, !types.Boolean
+  end
+
+  # Mutation takes an array of them:
```
Willianvdv: `them` should be `those`, right?

guides/schema/mutations.md
```diff
@@ -0,0 +1,65 @@
...
+  input_field :todos, !types[!TodoItemInputObjectType]
```
Willianvdv: Does the second `!` make a difference?

AUTHORS
```diff
@@ -0,0 +1,166 @@
...
+Wenley Tong <wenley.tong@gmail.com>
+Wesley Ellis <tahnok@gmail.com>
+Zach Holman <zach@zachholman.com>
+Zach Kemp <zvkemp@gmail.com>
+Zoltan Olah <zol@percolatestudio.com>
+aviscasillas <aviscasillas@gmail.com>
+empunkt <mfa@gpredictive.de>
+exAspArk <exaspark@gmail.com>
+jordanielewski <jordanielewski@users.noreply.github.com>
+willian <willian@hackerone.com>
```
Willianvdv: I'm listed multiple times with the same email address

guides/schema/mutations.md
rmosolgo: Yes, the first one says "there will definitely be a list here (not `null`)" and the second one says "the members of the list will definitely be `TodoItemInputObjectType`s (there will be no `null`s in the list)"

AUTHORS
```diff
@@ -0,0 +1,166 @@
...
+Mick Staugaard <mick@staugaard.com>
+Netto Farah <nettofarah@gmail.com>
+Nick Poorman <nickpoorman@gmail.com>
+Oguz Bilgic <fisyonet@gmail.com>
+Pol Miro <polmiro@gmail.com>
+Richard Lee <dlackty@gmail.com>
+Robin Murphy <robin@robinmurphy.co.uk>
+Robert Mosolgo <rdmosolgo@gmail.com>
+Christian Joudrey <cmallette@gmail.com>
+Garen Torikian <gjtorikian@gmail.com>
```
gjtorikian: This list, while nice, is going to be impossible to keep updated.

I'm also really uncomfortable with the idea of a list that's just ripe for spammers to pluck from: "Hello I am working with an early stage dark horse pre-seed startup and we are using the GraphQLs to help revolutionize chicken farmers! Would you be open for a ten minute chat?"

Can we remove this?

emilebosch: Sure, but its auto generated? Or is this maybe something from git-extras?

emilebosch: Wait it is not. I see whats the problem.

emilebosch:
Wait

PR #634, opened 2017-03-23
Give mutation its own section and nested mutations

emilebosch:
Ok sorry for the last one, that was messy and a bit too hasty.
Here a second attempt to get these docs in order.

In this PR i propose:

- Give mutations its own spot in schema
- Add nested mutations syntax for embedding

### DIFF
--- a/guides/schema/mutations.md
+++ b/guides/schema/mutations.md
@@ -1,0 +1,63 @@
+---
+title: Schema — Mutations
+---
+
+## Mutations
+
+Registering a mutation root allows to define fields that can mutate your data.
+
+```ruby
+Schema = GraphQL::Schema.define do
+  query QueryRoot
+  mutation MutationRoot
+end
+
+MutationRoot = GraphQL::ObjectType.define do
+  name "Mutation"
+
+  field :addPost, Post do
+    description "Adds a Post."
+
+    # Use Input Types to define complex argument types
+    argument :post, PostInputType
+    resolve ->(t, args, c) {
+      title = args['post']['title']
+      description = args['post']['description']
+      Post.create(title: title, description: description)
+    }
+  end
+end
+
+PostInputType = GraphQL::InputObjectType.define do
+  name "PostInputType"
+  description "Properties for creating a Post"
+
+  argument :title, !types.String do
+    description "Title of the post."
+  end
+
+  argument :description, types.String do
+    description "Description of the post."
+  end
+end
+```
+
+## Nested input types
+
+You can also nest input types. Let take this a todo list as an example.
+
+```ruby
+AddTodoList = GraphQL::Relay::Mutation.define do
+  name "AddTodoList"
+ 
+  # Create an input type for each todo item
+  TodoItemInputObjectType = GraphQL::InputObjectType.define do
+    name "TodoItem"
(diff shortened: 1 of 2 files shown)


guides/schema/mutations.md
```diff
@@ -0,0 +1,63 @@
...
+    description "Title of the post."
+  end
+
+  argument :description, types.String do
+    description "Description of the post."
+  end
+end
+```
+
+## Nested mutations
```
xuorig: Maybe have better wording for this since this seems to imply Nested Mutation in the sense of (not in spec):

```graphql
myMutation(input: {...}) {
  myNestedMutation(input: {...}) {
     ...
  }
}
```

and and not nested inputs?

rmosolgo: Yes, I was confused by this too. I thought that "nested mutations" referred to this: 

https://github.com/facebook/graphql/issues/252

(Which is the same as the example posted above.)

I would have described this as "mutations with complex inputs". However, I may be wrong to think that since at least the Graph.cool folks _have_ called this "nested mutations": https://blog.graph.cool/improving-dx-with-nested-mutations-af7711113d6#.syi1rk3o8

xuorig: > Graph.cool folks have called this "nested mutations"

Interesting 🤔 im not convinced of the wording

rmosolgo: Me either, but it wouldn't be the first time that my perspective as a graphql implementer was off track of the rest of the community 😖

emilebosch: I agree that they are not nested mutations but embedded inputs or complex input. Please let me know so i can adjust accordingly!

emilebosch: The use case would be that whenever creating items in bulk this really helps.
An added benefit of this is that it allows you to to easily wrap it in an transaction, when creating multiple records, making a mutation transaction safe. Which is for what i was using it in the first place. 

Is that something that we should maybe mention? I.e. with ruby sample code? I know graphql is db agnostic, but still graphql/activerecord is a common use case.

emilebosch:
Fixed the naming to nested input types :)

guides/schema/mutations.md
```diff
@@ -0,0 +1,63 @@
...
+
+  argument :description, types.String do
+    description "Description of the post."
+  end
+end
+```
+
+## Nested input types
+
+You can also nest input types. Let take this a todo list as an example.
```
Willianvdv: Copy suggestion:
> Let's create a mutation that adds multiple todo items at once

guides/schema/mutations.md
```diff
@@ -0,0 +1,63 @@
...
+    description "Title of the post."
+  end
+
+  argument :description, types.String do
+    description "Description of the post."
+  end
+end
+```
+
+## Nested input types
```
Willianvdv: Have you considered making a nested mutation that creates multiple post items? The example above already defines the type so it might be easier for the reader to continue with this example set.

rmosolgo:
Sorry for the slow response on this one, thanks, it looks great!

rmosolgo:
Published :tada: http://rmosolgo.github.io/graphql-ruby/schema/mutations

Issue #630, opened 2017-03-22
default_value of InputObjectType isn't serialised like graphql-js breaking tools like Relay


mvgijssel:
When you currently define a field as InputObject with a default value other than an empty Hash the introspection query will return something like:
```json
"name": "order_by",
"description": null,
"type": {
  "kind": "INPUT_OBJECT",
  "name": "ResourceOrder",
  "ofType": null
},
"defaultValue": "{\"direction\":\"ASC\"}"
```
where the default value isn't an object but a JSON string (https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/introspection/input_value_type.rb#L21).

When running the introspection query on https://github.com/graphql/graphql-js it returns something slightly different (note the missing quotes around the key):
```json
"name": "order_by",
"description": null,
"type": {
  "kind": "INPUT_OBJECT",
  "name": "ResourceOrder",
  "ofType": null
},
"defaultValue": "{direction:\"ASC\"}"
```
This is because graphql-js isn't generating a JSON string but something else: https://github.com/graphql/graphql-js/blob/master/src/type/introspection.js#L344. This implementation calls the printer which prints without quotes around the keys: https://github.com/graphql/graphql-js/blob/master/src/language/printer.js#L83. 

Currently when using default_values in graphql-ruby we do the following to not break Relay:

```ruby
argument :order_by, -> { ResourceOrder }, default_value: NoJsonHash.new(direction: 'ASC')
```
where `NoJsonHash` implements `to_json` to always return {}. Which is a terrible fix 🙈

mvgijssel:
After doing some more work realised that `NoJsonHash` totally doesn't work 😅. So the following "fix" works:
```ruby
module GraphQL
  module Language
    # @api private
    def self.serialize(value)
      if value.is_a? Hash
        JSON.generate({}, quirks_mode: true)
      else
        JSON.generate(value, quirks_mode: true)
      end
    end
  end
end
```
That way Relay doesn't crash but shouldn't be considered a solution.

rmosolgo:
Derp, this one is always getting me! 

> defaultValue may return a String encoding (using the GraphQL language)

[spec](https://facebook.github.io/graphql/#sec-The-__InputValue-Type)

For scalars, this was fine, so I used `JSON`, but it broke enums (fixed recently #605) and apparently input objects are wrong too, figures! 

Thanks for reporting this, I wonder if a suitable fix is to modify that handle hashes specially, something like 

[...]

PR #635, opened 2017-03-23
Remove quotes from defaultValue when it is a Hash

esjee:
Hi

This is an attempt to fix https://github.com/rmosolgo/graphql-ruby/issues/630. I hope this helps.

### DIFF
--- a/lib/graphql/loadtime_type_error.rb
+++ b/lib/graphql/loadtime_type_error.rb
@@ -1,5 +1,0 @@
-# frozen_string_literal: true
-module GraphQL
-  class LoadtimeTypeError < GraphQL::Error
-  end
-end
--- a/lib/graphql/loadtime_type_error.rb
+++ b/lib/graphql/loadtime_type_error.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+module GraphQL
+  class LoadtimeTypeError < GraphQL::Error
+  end
+end
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -109,10 +109,25 @@ module GraphQL
             kwargs = {}
             if type["defaultValue"]
               kwargs[:default_value] = begin
-                JSON.parse(type["defaultValue"], quirks_mode: true)
-              rescue JSON::ParserError
-                # Enum values are not valid JSON, they're bare identifiers
-                type["default_value"]
+                default_value_str = type["defaultValue"]
+                dummy_query_str = "query getStuff($var: InputObj = #{default_value_str}) { __typename }"
+
+                # Returns a `GraphQL::Language::Nodes::Document`:
+                dummy_query_ast = GraphQL.parse(dummy_query_str)
+
+                # Reach into the AST for the default value:
+                input_value_ast = dummy_query_ast.definitions.first.variables.first.default_value
+
+                case input_value_ast
+                when String, Integer, Float, TrueClass, FalseClass
+                  input_value_ast
+                when GraphQL::Language::Nodes::Enum
+                  input_value_ast.name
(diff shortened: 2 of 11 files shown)


rmosolgo:
:+1: This looks like a great approach, thanks for taking a try! 

We have a schema loader which takes introspection JSON as input and builds a GraphQL::Schema from it. It [expects inputs to be valid JSON](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/loader.rb#L110-L116) (expect enum values) but this was a mistake. It seems like we need a way to parse those strings as GraphQL values.

The trick is, the parser in this library accepts query documents only, not plain input objects. So one approach to solve this problem would be to make a fake query document, then extract the value from it: 

```ruby 
default_value_str = type["defaultValue"]
dummy_query_str = "query getStuff($var: InputObj = #{default_value_str}) { __typename }"
# Returns a `GraphQL::Language::Nodes::Document`:
dummy_query_ast = GraphQL.parse(dummy_query_str)
# Reach into the AST for the default value:
input_value_ast = dummy_query_ast.operations.first.variables.first.default_value
default_value = input_value_ast.to_h
```

Do you want to try adding that to make those tests pass?

Also, you can see that some validation errors also use `GraphQL::Language.serialize`. Those tests should be updated to reflect the new, correct behavior which you've implemented here. Do you mind updating those tests?

esjee:
I'll give it a try. Thanks for the pointers!

esjee:
Hey @rmosolgo, I ran into some trouble with different node types not supporting `.to_h`, and arrays not cooperating with the (de)serialization. I believe the commits I just pushed should fix that. Let me know what you think!

This turned out to be quite the adventure in the internals, and, well, unexplored territory for me. Please check that everything makes sense to you. 😅

lib/graphql/schema/loader.rb
```diff
@@ -109,10 +109,25 @@ def define_type(type, type_resolver)
...
+                input_value_ast = dummy_query_ast.definitions.first.variables.first.default_value
+
+                case input_value_ast
+                when String, Integer, Float, TrueClass, FalseClass
+                  input_value_ast
+                when GraphQL::Language::Nodes::Enum
+                  input_value_ast.name
+                when GraphQL::Language::Nodes::NullValue
+                  nil
+                else
```
rmosolgo: I forgot about all these other possibilities ... is there one more, do we need a `when Array` for cases like `$id: [ID!]! = ["a", "b"]` ?

esjee: You're right! I missed that. Will fix, and add a test.

lib/graphql/loadtime_type_error.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+module GraphQL
+  class LoadtimeTypeError < GraphQL::Error
```
esjee: I'm not sure if a seperate error makes sense. Thoughts?

rmosolgo: In other cases like this one, I've used a bare `raise` call for cases like this. 

Here's my thought: this error is only raised when something is _very_ wrong, and the user program should not continue. I can't think of any use case for rescuing this error, can you? 

If that's true, then it's important that we provide a good message (like the one you added below) and just crash with a good message. 

Adding an error class has a downside which I'd like to avoid: it adds a bit of API surface area, which I'll have to keep in mind when modifying this code. 

What do you think, would it be ok to use a plain `raise` (or `fail`, if you prefer) here?

esjee: That makes sense to me. Thanks!

rmosolgo:
Thanks for taking care of this, it's awesome!

PR #637, opened 2017-03-24
nil for GraphQL::Language::Nodes::NullValue input

nevesenin:
Hi,

I created a query type ```fooCollection(uuid: [Uuid!]....)```.
When executing a query like this
```
query bar($uuid: [Uuid!]) {
    fooCollection(uuid: $uuid) {
      totalCount
      edges {
        node {
          uuid
        }
      }
    }
  }
```
with variables ```{uuid: nil}``` the query arguments look like this:
```
#<GraphQL::Query::Arguments:0x00000005468430>
{"uuid"=>nil}
```
Same when executing the query via GraphiQL interface with variables ```{"uuid": null}```. This is what I expected

But when I execute this
```
{
  fooCollection(uuid: null) {
    totalCount
    edges {
      node {
        uuid
      }
    }
  }
}
```
query arguments is
```
#<GraphQL::Query::Arguments:0x000000054f9700>
{"uuid"=>[nil]}
```
which is not what I expected.

This PR would fix that, but I'm not sure if the fix is at the right place.

### DIFF
--- a/lib/graphql/query/literal_input.rb
+++ b/lib/graphql/query/literal_input.rb
@@ -7,6 +7,8 @@ module GraphQL
         case ast_node
         when nil
           nil
+        when Language::Nodes::NullValue
+          nil
         when Language::Nodes::VariableIdentifier
           variables[ast_node.name]
         else
--- a/spec/graphql/query_spec.rb
+++ b/spec/graphql/query_spec.rb
@@ -470,4 +470,90 @@ describe GraphQL::Query do
       }
     end
   end
+
+  describe 'NullValue type arguments' do
+    let(:schema_definition) {
+      <<-SCHEMA_DEFINITION
+        schema {
+          query: Query
+        }
+
+        type Query {
+          foo(id: [ID]): Foo
+        }
+
+        type Foo {
+          bar: String!
+        }
+      SCHEMA_DEFINITION
+    }
+    let(:expected_args) { [] }
+    let(:default_resolver) do
+      {
+        'Query' => { 'foo' => ->(_obj, args, _ctx) { expected_args.push(args) } },
+        'Foo' => { 'bar' => ->(_obj, _args, _ctx) { 'baz' } }
+      }
+    end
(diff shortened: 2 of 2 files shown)


rmosolgo:
Hi, thanks for this contribution! I agree that the previous behavior was incorrect and I think this is a very good fix for it. In order to merge this change, I would like to add a test which covers the behavior. Do you want to take a crack at adding the test? 

I don't think there's a `_spec` file for this code yet, so you could add a new `describe` block to `query_spec.rb` instead. 

One way to test it would be to define a [prototype schema](http://rmosolgo.github.io/blog/2017/03/17/prototyping-a-graphql-schema-from-definition-with-ruby/) in the test itself, then run some queries on that schema to make sure that the `args` object has the right value. 

For example, you could create an array on the _outside_ of the schema, then push `args` into the array: 

```ruby 
arg_values = [] 

# Make a test schema:
defn_string = # ... 
resolvers = { "Query" => { "argTest" => ->(obj, args, ctx) { arg_values << args } } } 
schema = GraphQL::Schema.from_definition(defn_string, default_resolve: resolvers) 

# Run a query that demonstrates the bug
schema.execute(query_str)

# Assert that the `args` value from above has the _correct_ value:
assert_equal nil, arg_values[0][:testArgument]
```

Of course, if you have another approach to test this code, feel free to try it! I just want to make sure we don't break this again in the future :)

nevesenin:
Thanks for your feedback. I'll add some tests next week.

rmosolgo:
That sounds great, let me know if I can help out on it!

nevesenin:
I was curious about the schema from definition stuff so I couldn't wait. The first one is the one that fails without the fix.

rmosolgo:
Awesome, thanks again for taking care of it! I pushed a few changes to reduce the boilerplate here in case you're interested: 7e90a7e6

I'm hoping to track down a few more bugs before releasing 1.5.5 but this will be out soon!

PR #639, opened 2017-03-25
Fix query specs for GraphQL::Language::Nodes::NullValue arguments

nevesenin:
I'm sorry. There is an embarrassing mistake in the specs for #637.


rmosolgo:
Oops, I missed it too, good catch! Thanks!

PR #640, opened 2017-03-26
Raise `Query::UserError` when appropriate

gjtorikian:
I noticed some `ExecutionError`s popping up in our logs, so I started to investigate. It turns out, queries like the following were being executed:

``` graphql
query {
  viewer {
    login
  }
}

type Foo {
  name: String
}
```

We reserve `ExecutionError`s for catastrophic 500s--things that our GraphQL system should have known how to handle, but didn't.

In this case,  this query is the result of bad input, not anything a server should be able to fix. I think this error is much more appropriate as a user error. Implementors should catch `UserError` and find a way to make it appear in the `errors` key.

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -13,6 +13,9 @@ module GraphQL
   class Query
     extend Forwardable
 
+    class UserError < GraphQL::ExecutionError
+    end
+
     class OperationNameMissingError < GraphQL::ExecutionError
       def initialize(name)
         msg = if name.nil?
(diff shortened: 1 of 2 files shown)


rmosolgo:
woah... I was just looking at this, trying to make sure that the error doesn't propagate into the application, but apparently it does?!

What's your end goal here? It _seems_ like this should go back to the client, like a validation error, right?

gjtorikian:
Exactly. I would like the server to `rescue` this, but ultimately, it needs to go back to the client. And it does propagate into the app, because `ExecutionError` is raised in several places, and sometimes, it actually can be a server issue. 

We don't _need_ to do it like this--perhaps there's an easier mechanism to fail instantly and report to the client?--but it was the quickest solution I saw.

There needs to be a distinction between execution errors that's the fault of the server and those of the client, basically.

rmosolgo:
Cool, I totally agree! That was something I missed on the first pass of this gem: if the client does something wrong, the server doesn't need to know about it.

Let me see if there's a way to handle this without communicating via `raise ... rescue` ...

rmosolgo:
What do you think of #647?

gjtorikian:
Oh heck yes. I'd much rather prefer doing this via the parser. Wish I'd thought of that!

PR #641, opened 2017-03-27
Run tests on JRuby

MSNexploder:
This pull requests enables JRuby for all tests on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ rvm:
 - 2.2.5
 - 2.3.3
 - 2.4.0
+- jruby-9.1.5.0
 
 gemfile:
 - gemfiles/rails_3.2.gemfile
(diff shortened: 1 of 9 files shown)


rmosolgo:
Wow, awesome!! Thanks for taking a crack at this. It would be great to formally support JRuby.

It looks to me like the failures are unrelated to your changes here, right? Maybe some CI configs just need to be updated (but I wonder why they pass on master ...)

rmosolgo:
Ok, I fixed CI on master, so if you rebase, you should get a proper result!

MSNexploder:
I rebased to the latest master and fixed Travis picking JRuby 1.7 instead of the current 9k builds.

rmosolgo:
Thanks! Wow, strange failures there. I guess we can take a look at those next 😬

PR #642, opened 2017-03-28
Add graphql-query-resolver

nettofarah:
GraphQL::QueryResolver is an add-on to graphql-ruby that allows your field resolvers to minimize N+1 SELECTS issued by ActiveRecord.


rmosolgo:
👍  Thanks!

nettofarah:
Thank you!

PR #643, opened 2017-03-28
Add presentations section

nettofarah:
We can use this section to add talks about graphql-ruby.


rmosolgo:
Nice, I'll try to round up a few more soon.

nettofarah:
Thank you!

PR #645, opened 2017-03-28
Make LazyMethodMap thread-safe using Concurrent::Map as backing storage

### LINKED ISSUE #628: Runtime error mutating lazy method map on JRuby

ivoanjo:
#### This is my proposal for an alternative to #631 using the concurrent-ruby gem instead of explicit locking to deal with #628.

As identified in #628, the previous LazyMethodMap implementation was not thread-safe, which could cause issues when using graphql and a ruby threaded webserver (runtime errors and potentially initializing the same location more than once).

This affected both MRI and JRuby.

To fix this, this commit switches LazyMethopMap's backing storage from a regular ruby Hash into a Concurrent::Map from the concurrent-ruby gem (adding this gem as a dependency), thus making this class thread-safe.

Fixes #628

### DIFF
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -44,4 +44,6 @@ Gem::Specification.new do |s|
 
   # website stuff
   s.add_development_dependency "github-pages"
+
+  s.add_dependency 'concurrent-ruby', '~> 1.0'
 end
--- a/lib/graphql/execution/lazy/lazy_method_map.rb
+++ b/lib/graphql/execution/lazy/lazy_method_map.rb
@@ -1,22 +1,19 @@
 # frozen_string_literal: true
+
+require 'concurrent'
+
 module GraphQL
   module Execution
     class Lazy
       # {GraphQL::Schema} uses this to match returned values to lazy resolution methods.
       # Methods may be registered for classes, they apply to its subclasses also.
       # The result of this lookup is cached for future resolutions.
+      # Instances of this class are thread-safe.
       # @api private
       # @see {Schema#lazy?} looks up values from this map
       class LazyMethodMap
         def initialize
-          @storage = Hash.new do |h, value_class|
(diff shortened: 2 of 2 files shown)


rmosolgo:
Thanks for figuring this out, I've cherry-picked this commit into #631!

PR #646, opened 2017-03-29
feat(Arguments) accept a prepare function to modify or validate arguments

lucaspolonio:
This is an attempt to implement the argument `prepare` function discussed on https://github.com/rmosolgo/graphql-ruby/issues/540. I'm not too familiar with this code base, so I'm open to any suggestions you may have. 

Also, my specs for `LiteralInput.from_arguments` look more like end-to-end tests than unit tests. Let me know if you'd like to move them to a better spec file or test the feature in a different way.

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -40,6 +40,15 @@ module GraphQL
 
     ensure_defined(:name, :description, :default_value, :type=, :type, :as, :expose_as, :prepare)
 
+    # @api private
+    module DefaultPrepare
+      def self.call(value); value; end
+    end
+
+    def initialize
+      @prepare_proc = DefaultPrepare
+    end
+
     def initialize_copy(other)
       @expose_as = nil
     end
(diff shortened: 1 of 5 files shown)


lib/graphql/argument.rb
```diff
@@ -54,9 +72,22 @@ def expose_as
       @expose_as ||= (@as || @name).to_s
     end
 
+    # @param value
+    # @return [Object] The prepared `value` for this argument or `value` itself if no `prepare` function exists.
+    def prepare(value)
+      return value unless @prepare_proc.respond_to?(:call)
+      @prepare_proc.call(value)
```
rmosolgo: How about removing this logical branch by providing a default, no-op `@prepare_proc`? For example:

```ruby
module DefaultPrepare 
  def self.call(value); value; end 
end 

def initialize 
  # ... 
  @prepare_proc = DefaultPrepare
end 

def prepare(value)
  @prepare_proc.call(value)
end 
```

It requires some setup, but it simplifies the runtime code since we can always assume that a `@prepare_proc` is present.

lucaspolonio:
Hey @rmosolgo, thanks for reviewing the PR. I like your suggestion. Done 👍 

Also, I'll look into using `Argument.from_dsl` for mutations in a new PR. Hopefully I can get it done soon. I'll let you know if I have any questions.

rmosolgo:
Very cool, thanks for this feature!!

PR #650, opened 2017-03-31
Fix mutations using :as option on arguments

pabloh:
Using `:as` on mutations arguments is ignored:

```ruby
LoginMutation = GraphQL::Relay::Mutation.define do
  name 'Login'
  input_field :username, !types.String, as: :user_name # The :as option has no effect
  input_field :password, !types.String

  # ...
end
```

### DIFF
--- a/lib/graphql/relay/mutation.rb
+++ b/lib/graphql/relay/mutation.rb
@@ -169,9 +169,8 @@ module GraphQL
             input_field :clientMutationId, types.String, "A unique identifier for the client performing the mutation."
             relay_mutation.arguments.each do |input_field_name, field_obj|
               kwargs = {}
-              if field_obj.default_value?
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks! I'll take another look at this soon, since it will break again after merging #646

pabloh:
@rmosolgo , may be I should add some test.

rmosolgo:
If you want to, that'd be great! But since I was about to cut a release, I wanted to include this fix anyways 😁

pabloh:
Cool, I'll try to send you the test before the release, if not I could really use this fix anyway.

rmosolgo:
It's already released :) https://rubygems.org/gems/graphql/versions/1.5.5

Feel free to add tests in another PR.

pabloh:
:stuck_out_tongue: ok, next one...

PR #652, opened 2017-04-04
Remove useless assignments

hibariya:
Hello.

I noticed there are some assignments which look unnecessary. I think that removing them makes codes more readable since it avoids unnecessary guessing. That's why I removed some assignments and added [a rule](https://github.com/bbatsov/rubocop/blob/master/manual/cops_lint.md#lintuselessassignment) to `.rubocop.yml`.

Thank you!

### DIFF
--- a/spec/graphql/base_type_spec.rb
+++ b/spec/graphql/base_type_spec.rb
@@ -39,10 +39,11 @@ describe GraphQL::BaseType do
     it "resets connection types" do
       # Make sure the defaults have been calculated
       obj_edge = obj_type.edge_type
+      obj_conn = obj_type.connection_type
       obj_2 = obj_type.dup
       obj_2.name = "Cheese2"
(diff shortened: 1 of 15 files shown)


spec/graphql/base_type_spec.rb
```diff
@@ -39,7 +39,6 @@
     it "resets connection types" do
       # Make sure the defaults have been calculated
       obj_edge = obj_type.edge_type
-      obj_conn = obj_type.connection_type
```
rmosolgo: The issue here is a typo on line 46: it should assert `refute_equal(obj_conn, obj_2.connection_type)`

hibariya: Oh, I see! I've left this line and updated line 46 on b6965b4

rmosolgo:

Issue #653, opened 2017-04-04
[Question] How to create a mutation without `input` field?


cpunion:
I want to create a mutation, likes:

```graphql
mutation createPost($title: String!, $body: String!) {
  createPost(title: $title, body: $body) {
    id
  }
}
```

When I create by `rails g graphql:mutation CreatePost`, and then use it with:

```ruby
Types::MutationType = GraphQL::ObjectType.define do
  name 'Mutation'

  field :createPost, field: Mutations::CreatePost.field
end
```

It generate a mutation:

```graphql
mutation createPost($input: CreatePost!) {
  createPost(input: $input) {
    id
  }
}
```

It just has one argument, needs less code, sometimes easy to write. But I want it seems like a function with some arguments.
**How to expand the input argument to its fields when use GraphQL::Relay::Mutation?**

If I define with below:

```ruby
Types::MutationType = GraphQL::ObjectType.define do
  name 'Mutation'

  field :createPost do
    argument :title, !types.String
    argument :body, !types.String
    resolve -> (...) {...}
  end
end
```

**How to separate fields to files in this way?**

rmosolgo:
[All Relay mutations have an `input` argument](https://facebook.github.io/relay/graphql/mutations.htm#sec-Mutation-inputs), so you can't use `Relay::Mutation` for this. 

I think you might like `GraphQL::Function` ([API](http://www.rubydoc.info/github/rmosolgo/graphql-ruby/GraphQL/Function), [guide](https://rmosolgo.github.io/graphql-ruby/schema/code_reuse#functions)), how does that look?

cpunion:
@rmosolgo thanks. I will try it.

rmosolgo:
feel free to reopen this if it doens't work for you!

cpunion:
I tried and read document about functions. It only can use builtin types for arguments. So if I want use custom types, and I don't want see `input`, do I must define the field with a block likes below?

```ruby
  field :createPost do
    argument :title, !types.String
    argument :body, !types.String
    resolve -> (...) {...}
  end
```

rmosolgo:
> only can use builtin types

Is that right? It should support _any_ type!

cpunion:
This is my code:

```ruby
class Functions::SignUp < GraphQL::Function
  type Types::UserInfoType

  argument :email, !types.String
  argument :password, !types.String

  def call(obj, args, ctx)
    UserInfo.new
  end
end
```

```ruby
[...]

PR #654, opened 2017-04-04
Provides shorthand access to GraphQL's built-in types within function classes

cpunion:
See https://github.com/rmosolgo/graphql-ruby/issues/653#issuecomment-291553315


rmosolgo:
:metal: awesome, thanks!

PR #655, opened 2017-04-05
Test against latest ruby on Travis CI

hibariya:
Hello. According to the `.travis.yml`, on 2.2 or later, latest rubies of each minor version have been being used. I've updated them.


rmosolgo:
Thanks!

PR #660, opened 2017-04-06
Fix InputObjectType input_fields

lucaspolonio:
Both `Mutation` and `InputObjectType` input_fields already use `Define::AssignArgument`, which should be the only way to create `Argument`s.

But when a `Mutation` builds its `InputObjectType`, it recreates `Argument`s (`input_field`s) using properties from the `Mutation`'s own `Argument`s. It needs to explicitly copy all the options from one argument to another, which caused https://github.com/rmosolgo/graphql-ruby/pull/650.

This PR gets rid of this recreation process and just merges the `Mutation`'s `input_fields` into the new `InputObjectType`'s `input_fields`.
This change should prevent any divergence between fields and mutation arguments/input_fields.

Note 1: As you can see in the changed specs, the `input_field`s order changed. `clientMutationId` now comes after custom `input_field`s. If the order is important, I can update the PR accordingly.
Note 2: https://github.com/rmosolgo/graphql-ruby/pull/646 can now be safely merged.


rmosolgo:
:+1: I like your approach to just reassign the objects we already have! I pushed a small refactor with two changes: 

- Don't add to `input_object`'s public API (otherwise I will have to maintain it forever 😆 )
- Keep input fields in the same order (some people keep a dump of their schema in version control, I don't want to introduce needless noise to their diffs!)

lucaspolonio:
@rmosolgo Looks good to me! Thanks for tidying it up

Issue #657, opened 2017-04-05
Better visibility for non-UTF8 strings


gjtorikian:
https://github.com/rmosolgo/graphql-ruby/pull/517 mandated that all strings come back UTF-8 encoded, which is great. But in development practice, I and several other colleagues have been bitten by the mysterious `null` that comes back from a part of the Rails application that otherwise works fine.

Could I amend this check to do something like `raise ExecutionError if Rails.env.development? || Rails.env.test?`? Basically, when building a GraphQL schema around existing DB values, it would be extremely helpful to know that the values need to be force encoded. In production this would continue to return `nil`.

cc @rmosolgo @josh for thoughts
fyi @cheshire137

rmosolgo:
:+1: 

As a stop-gap, you can override the built-in function:

```ruby 
GraphQL::STRING_TYPE.coerce_result = ->(value) { 
    str = value.to_s
    if str.encoding == Encoding::US_ASCII || str.encoding == Encoding::UTF_8 
      str
    elsif Rails.env.development? || Rails.env.test? 
      raise "Gotta be utf-8"
    else 
      nil
    end 
}
```

I'm open to a louder 💥  in the failing case, but as you saw on #517, I didn't have a great time selecting one 😖

josh:
> I'm open to a louder 💥 in the failing case, but as you saw on #517, I didn't have a great time selecting one 😖

Ha, same!

dylanahsmith:
I would prefer to get the exception in production too, since I don't want this to be hard to debug in production either

gjtorikian:
> I didn't have a great time selecting one

I keep rolling this over in my head; I'm almost tempted to have it `raise` _always_. Is that crazy?

I looked back on the old thread and came across [this comment](https://github.com/rmosolgo/graphql-ruby/pull/517#issuecomment-281554702). Given that the spec requires UTF-8 I feel pretty good about making this 💥  always. What do you think?

dylanahsmith:
Also, I think ExecutionError is for client errors.  In this case we would want a non-ExecutionError so that it would be treated as an internal server error, where usually that error would get reported to something like errbit.

rmosolgo:
:+1: for some hard crash & non `ExecutionError` as @dylanahsmith suggested, so that schema owners can track down and address the issue. 

[...]

PR #661, opened 2017-04-06
Raise on UTF-8 encoding error

gjtorikian:
Aims to close https://github.com/rmosolgo/graphql-ruby/issues/657!

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -9,6 +9,9 @@ module GraphQL
   class Error < StandardError
   end
 
+  class CoercionError < Error
+  end
+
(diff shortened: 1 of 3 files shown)


rmosolgo:
Give me just a bit, I want to see if I can layer a change to scalar coercion so that `ctx` is threaded through, that will make this not-a-breaking-change

rmosolgo:
Thanks again, I applied a bit more plumbing to this error (sent it through `Schema#type_error`) and renamed it (to `StringEncodingError`), that way people who _want_ the old behavior can still have it!

PR #665, opened 2017-04-08
refactor(Schema) raise definition errors lazily to avoid messing with Rails constant loading

### LINKED ISSUE #651: Code Reloading w/ Rails in Development
Using some of the very basic examples I'm running into an issue where query execution would hang whenever I change code in development. And even if I don't change code, on all subsequent requests I would get the error `LoadError (Unable to autoload constant Types::QueryType, expected /Users/randuin/code/pw/pwpback/app/graphql/types/query_type.rb to define it):`

rmosolgo:
Fixes #651 

I don't know why, but raising an error during `Schema.define` causes `Types` to appear in `ActiveSupport::Dependencies.autoloaded_constants` _twice_. As a result, `ActiveSupport.remove_unloadable_constants!` goes into a loop and the server hangs. Sad! 

Moving the `raise` into `#execute` circumvents the issue by allowing the constant to load properly. The user experience _should_ be the same: you can't execute a query on an invalid schema. 

(Technically, this arrangement is a bit more flexible: you _could_ define an invalid schema, then `.redefine` it to become valid. This was impossible before, but obviously it doesn't matter much 😆 )

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -58,6 +58,7 @@ module GraphQL
       :query_execution_strategy, :mutation_execution_strategy, :subscription_execution_strategy,
       :max_depth, :max_complexity,
       :orphan_types, :resolve_type, :type_error, :parse_error,
+      :raise_definition_error,
       :object_from_id, :id_from_object,
       :default_mask,
       :cursor_encoder,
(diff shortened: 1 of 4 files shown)


cpunion:
Does this PR also solve the stack trace?

Currently, if an error raised, can't see the stack trace:
```
GraphQL::Schema::InvalidTypeError (Mutation is invalid: field "signIn" type must return GraphQL::BaseType, not NilClass (nil)):

app/graphql/api_schema.rb:1:in `<top (required)>'
app/controllers/graphql_controller.rb:19:in `execute'
...
```
Sometimes can't see useful messages.

rmosolgo:
No :( 

That's possible but hard, as far as I know. You don't have access to `caller` during schema definition or during validation. You'd have to capture it in the DSL methods. Possible, but hard!

PR #667, opened 2017-04-09
Optionally pass `ctx` to scalar coerce functions

rmosolgo:
Pass ctx to scalar coerce functions, pass coercion errors to `Schema#type_error` handler 

__TODO__ 

- [x] Thread `ctx` through coercion functions 
- [x] Maintain `ctx`-less API for compatibility
- [x] Pass coerce errors to type_error handler 
- [x] Make `coerce_non_null_input` private (there are no hits on github)

~~In a later (non-patch) release, I'll mark the old `ctx`-less APIs as deprecated.~~ Doin' it now

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -9,9 +9,6 @@ module GraphQL
   class Error < StandardError
   end
 
-  class CoercionError < Error
-  end
-
   class ParseError < Error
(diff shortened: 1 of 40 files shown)


rmosolgo:
Oops, fixed in 8b26dfa7 to _return_ the value from `type_error`

PR #668, opened 2017-04-10
Argument conflict on nested interface type fragments

arthurschreiber:
I ran into an issue where fetching the same field with different arguments for different types using different fragments that used the same interface would fail with a validation error. (That's a really long winded sentence!)

I think the issue is a bit hard to describe, so I modified one of the existing test cases to reproduce the failure I was running into. I'm not even sure whether this is allowed by the GraphQL specification or not, and I couldn't find anything in the spec. It definitely feels like this should just work. 😆 

This is similar to https://github.com/rmosolgo/graphql-ruby/issues/418.

### DIFF
--- a/spec/graphql/static_validation/rules/fields_will_merge_spec.rb
+++ b/spec/graphql/static_validation/rules/fields_will_merge_spec.rb
@@ -422,12 +422,20 @@ describe GraphQL::StaticValidation::FieldsWillMerge do
       {
         pet {
           ... on Dog {
-            name(surname: true)
+            ...X
           }
           ... on Cat {
-            name
+            ...Y
           }
         }
+      }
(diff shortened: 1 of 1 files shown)


rmosolgo:
Wow, thanks for the failing test case, that's really helpful! I think you're right that this should "just work", I'll take a look ...

rmosolgo:
What a doozie! Thanks again for the test case, I've cherry-picked it into #680 and I'll merge it there.

rmosolgo:
BTW the problem was during "fragment inlining": we didn't keep track of the scope where fragment spreads were used, so when they were inlined, they lost the surrounding context (in this case, constraints from the non-overlapping fragment spreads)

PR #672, opened 2017-04-11
feat(Subscriptions) add subscriptions

rmosolgo:
Let's support [GraphQL Subscriptions](https://github.com/facebook/graphql/pull/267) in Ruby! 

Some library changes are required: 

- Special behavior for executing subscriptions: 
  - On the first execution, no fields are executed. Instead, each [root field + arguments pair](https://github.com/facebook/graphql/issues/282) is registered with the subscription backend. 
  - On later executions, only root field + argument pairs which match the triggered event are executed. The root value for that execution is an application-provided object. (eg, if `commentAdded` is triggered, the root object is the `Comment` provided by the application.)
  - Add `Query::Result` to simplify re-evaluation #898  

My implementation has made an assumption about how root fields map to application events: 

```ruby 
event_key = "#{field.name}(#{JSON.dump(args.to_h)})"
# commentAdded({"postId":1})
```

Is this sufficient for capturing the "unit of subscription"? Or does it miss some cases? 

See the guides in this PR for documentation. 

---- 

Working demo: https://gist.github.com/rmosolgo/ba31acf93f07f8007d99ba365a662d8f


rmosolgo:
Feature-wise I think this is ready, but as I write the docs, I think the API can be better. Something more ... _object-oriented_ 🙀

rmosolgo:
Working on an ActionCable client https://github.com/rmosolgo/graphql-ruby-client/pull/2

blevine:
Not quite sure what the status of subscription support is at this point. Would I be able to pull the 1.7.0 branch to start implementing subscriptions now?

rmosolgo:
@blevine not yet, i'm hoping to merge this into 1.7 today

rmosolgo:
Here goes nothin! Feel free to take the 1.7 branch for a spin.

PR #673, opened 2017-04-12
Add some details to hasNextPage/hasPreviousPage descriptions.

tiegz:
According to the [Relay spec](https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo.Fields), the `hasNextPage` and `hasPreviousPage` fields in PageInfo require `first` or `last` arguments to be meaningful:

> hasPreviousPage is only meaningful when last is included, as it is always false otherwise. hasNextPage is only meaningful when first is included, as it is always false otherwise. When both first and last are included, both of the fields are set according to the above algorithms, but their meaning as it relates to pagination becomes unclear. This is among the reasons that pagination with both first and last is discouraged.

(NB this was decided [for performance purposes](https://github.com/graphql/graphql-relay-js/issues/58#issuecomment-169542843))

Since this can cause some [confusion](https://github.com/graphql/graphql-relay-js/issues/58#issuecomment-169542843) from getting a meaningless `false` value back, I propose we mention this in these fields' descriptions.

### DIFF
--- a/lib/graphql/relay/page_info.rb
+++ b/lib/graphql/relay/page_info.rb
@@ -5,10 +5,10 @@ module GraphQL
     PageInfo = GraphQL::ObjectType.define do
       name("PageInfo")
       description("Information about pagination in a connection.")
-      field :hasNextPage, !types.Boolean, "Indicates if there are more pages to fetch (only meaningful if `first` argument is present)", property: :has_next_page
-      field :hasPreviousPage, !types.Boolean, "Indicates if there are any pages prior to the current page (only meaningful if `last` argument is present)", property: :has_previous_page
-      field :startCursor, types.String, "When paginating backwards, the cursor to continue", property: :start_cursor
-      field :endCursor, types.String, "When paginating forwards, the cursor to continue", property: :end_cursor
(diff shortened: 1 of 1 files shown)


rmosolgo:
Sorry for the delay on this, these descriptions were originally added for parity with the JS implementation. 

It seems like the best thing is to bring some i18n-like framework to graphql so that users can customize the built-in messages, but that's a big effort 😅 . I'm not sure if we should deviate from JS in the meantime or not.

tiegz:
i18n sounds good for the future! I'd be happy to help if you need some assistance with that.

Also wanna throw out there that the `graphl-relay-js` library's descriptions for these implies the forward/backward direction too: https://github.com/graphql/graphql-relay-js/blob/373f2dab5fc6d4ac4cf6394aa94cbebd8cb64650/src/connection/connection.js#L133-L153

rmosolgo:
Thanks for the link, I guess I mis-remembered the bit about matching their descriptions! What do you think about taking the descriptions from JS? I'd be happy to merge that.

tiegz:
Sounds good! Just pushed that change.

rmosolgo:
Thanks, sweet!

PR #675, opened 2017-04-14
fix(Rewrite) root nodes have nil definition

rmosolgo:
Oops, `node.definition` raised on root nodes which broke some query analyzers.


rmosolgo:
It actually passed I just cancelled a build:

image

PR #676, opened 2017-04-15
Transcode string results to UTF-8 when possible

dwhitlow:
Fixes an issue where strings such as "f2f962bd-2216-4c00-a8b0-8f9c71818285"
encoded as ASCII-8BIT would raise GraphQL::StringEncodingError when they
could be easily transcoded.

Strings containing code points that cannot be mapped to UTF-8 will still
raise GraphQL::StringEncodingError.

### DIFF
--- a/lib/graphql/string_encoding_error.rb
+++ b/lib/graphql/string_encoding_error.rb
@@ -4,7 +4,7 @@ module GraphQL
     attr_reader :string
     def initialize(str)
       @string = str
-      super("String \"#{str}\" was encoded as #{str.encoding}! GraphQL requires UTF-8 encoding.")
+      super("String \"#{str}\" was encoded as #{str.encoding}! GraphQL requires an encoding compatible with UTF-8.")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


rmosolgo:
This approach was [discussed](https://github.com/rmosolgo/graphql-ruby/pull/517#issuecomment-277103377) a while back but skipped over because we didn't have a way to let the application handle invalid strings. Now we _do_ have a way for that, @josh how does this approach look to you?

dwhitlow:
@rmosolgo @josh ping?

cberkom:
We're having this exact issue as well. Staying on `1.5.5` until this can be merged in.

rmosolgo:
Sorry for the delay. This approach is fine with me, and it sounds like nobody objects!

rmosolgo:
I'll make a patch release tomorrow or so, let me take a quick look at other open bugs...

rmosolgo:
🚢 !

cberkom:
Awesome! By the way, thanks for the hard work on this gem; it's been a pleasure to work with!

PR #677, opened 2017-04-15
Fix accessing input fields in Relay mutations guide

juno:
Thank you for maintaining such a great gem!

The example code in Relay mutation guide which accessing input fields seems to be not up to date.
Example uses `args` to access input field, but it causes error.

```
NoMethodError (GraphQL::Relay::Mutation can't define 'args'):
```

I think this should to be `inputs`. 🤔


rmosolgo:
Oops, you're right, thanks for the fix!!

PR #681, opened 2017-04-19
fix(Warden) return fields with instrumentation

rmosolgo:
Oops, it was returning fields without instrumentation, not the instrumented fields which are present on `Schema`


rmosolgo:
There's technically still a bug here: if you change a field's return type during instrumentation, it won't show up in the type_map. I think the real fix is to merge reduce_types & instrumented_field_map into one pass.

PR #687, opened 2017-04-21
feat(RakeTask) add tasks for dumping the schema

### LINKED ISSUE #593: Rake tooling
I'd like to provide a set of rake tasks for managing a GraphQL system. I don't know a lot about Rake yet but I can imagine something like Rubocop's API:

```ruby 
require "graphql/rake_task"

# Define a bunch of tasks that will load the schema as needed
GraphQL::RakeTask.new(schema_name: "MyAppSchema")
```

One challenge is loading the schema: we may need to depend on `:environment` in Rails, or require a certain file to find the schema. Perhaps the user could define the task: 

```ruby 
GraphQL::RakeTask.new do 
  # On Rails:
  task load_schema: :environment do 
    MyAppSchema 
  end 
  # Off Rails: 
  task :load_schema do 
    require "app/my_app_schema" 
    MyAppSchema 
  end 
end 
```

Also, should we support multiple schemas out of the box? 

Which tasks should we implement? I can only think of schema dumping:

- Dump Schema to IDL 
- Dump Schema to JSON

rmosolgo:
Closes #593

### DIFF
--- a/lib/graphql/rake_task.rb
+++ b/lib/graphql/rake_task.rb
@@ -1,0 +1,137 @@
+# frozen_string_literal: true
+module GraphQL
+  # A rake task for dumping a schema as IDL or JSON.
+  #
+  # By default, schemas are looked up by name as constants using `schema_name:`.
+  # You can provide a `load_schema` function to return your schema another way.
+  #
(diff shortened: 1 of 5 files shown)


gjtorikian:
Yay!

Issue #526, opened 2017-02-07
Custom directives


defrex:
It seems that custom directives are _almost_ possible with `graphql-ruby`, but they are missing something akin to `resolve`. Ideally the resolve for a directive would be able to either modify the output of the original resolve, or prevent it from being executed.

Something like this would be nice.

```ruby
GraphQL::Directive.define do
  name 'bananaMe'
  description 'Does what it needs to.'
  locations [GraphQL::Directive::FIELD, GraphQL::Directive::FRAGMENT_SPREAD, GraphQL::Directive::INLINE_FRAGMENT]

  argument :force, !types.Boolean, 'Forces banana.'

  resolve ->(original_resolve, obj, args, ctx) {
    if args[:force]
      'Banana'
    else
      original_resolve(obj, args, ctx)
    end
  }
end
```

Right now that logic seems to exist [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/directive_checks.rb). It would take a real hack to change it without support.

rmosolgo:
> almost possible 

I agree! I was looking at options while rewriting the query transformation last week but alas, left it as [a `# TODO`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/internal_representation/node.rb#L31).

I'm also interested in this because I want to implement `@defer` and `@stream` (and maybe `@live`!) in a non-intrusive, non-[rewrite-everything](https://github.com/rmosolgo/graphql-ruby/pull/168) way. 

Here's kind of what I was imagining: 

1. Add a hook to schema definitions where people can opt into directives: 

  ```ruby 
  MySchema = GraphQL::Schema.define do 
    # map `@banana` to user-defined `BananaDirective`
    directive banana: BananaDirective 
  end 
  ``` 

2. Add an API for directives to modify flagged nodes (like `resolve` in your example above). I'm not _sure_ if wrapping `resolve` is enough for defer and stream because, when nodes are deferred, they should _entirely_ absent from the response, not `nil`. So I may need to add a way to modify the query itself.

3. Port the current `@skip` and `@include` to this pattern, and add some kind of global `GraphQL::Schema.default_directives` map which includes them and applies them by default. 

What do you think of something like that?

defrex:
That sounds great!

I don't think having the modification API be a little complex is a huge deal either, as long as the logic can be easily contained. Writing a directive is a pretty rare activity.

rmosolgo:
[...]

PR #688, opened 2017-04-22
Support skipping fields altogether from user code

rmosolgo:
This will help on subscriptions, where untriggered root fields will be skipped. I think it'll also help for `@defer`-like behavior where some fields should be entirely absent from the result, but we have to do the skip at runtime in order to have access to the backing `obj`.

__TODO:__

- Improve `SelectionResult#delete` to not be O(n). I'm feeling a lot of pain here in the parallel trees of `Context` and `Result` objects, I think they should be merged.
- Test for lists of scalars? Or this is unsupported for lists? (The whole list must be `ctx.skip`?)

### DIFF
--- a/lib/graphql/compatibility/execution_specification.rb
+++ b/lib/graphql/compatibility/execution_specification.rb
@@ -15,6 +15,7 @@ module GraphQL
     # - Typecasting
     # - Error handling (raise / return GraphQL::ExecutionError)
     # - Provides Irep & AST node to resolve fn
+    # - Skipping fields
     #
     # Some things are explicitly _not_ tested here, because they're handled
     # by other parts of the system:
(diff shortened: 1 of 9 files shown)


rmosolgo:
I was thinking about rewriting `@skip` to use this API, but I realized it's not enough. Consider a case where some nested fields are skipped and others are not: 

```graphql
{
  thing {
    a
  }
  thing @skip(if: true) {
    b
  }
}
```

Certainly `data.thing.a` should be present while `data.thing.b` should be absent. 

This is currently implemented during the rewrite phase, where the AST is flattened into a query plan. This is the perfect time to do it since skipping is lexically scoped: children of skipped nodes are also skipped.

Later, it's too late. If you didn't remove skipped nodes while rewriting, you'd end up with a plan like: 

```graphql 
{
  thing { # two AST nodes, one @skip, one not
    a
    b 
  }
}
```

But you can't distinguish which of `a` and/or `b` were skipped, because they've lost their original lexical scope. 

So I'm not sure how to do it, but I think if you could provide some API handling this, you'd open the door to custom directives (#526).

Issue #406, opened 2016-11-19
Merging GraphQL Queries Together


theorygeek:
For those of use who use Relay, we sometimes encounter N+1 node queries:
```graphql
query {
  node(id: "...") { ... }
}
```

Relay might execute hundreds of these queries at once. For example, you might be displaying a table of data on the page, and maybe the user is able to add a new column. Relay might fetch the data for the new column for each row individually.

I wrote helper for GraphQL that can take these queries and combine them together. You can check it out here: https://gist.github.com/theorygeek/85b2f7df6931a0eb3f06141ef1e57f30

Example usage:
```ruby
query_1_text = "..."
query_2_text = "..."
query_1_vars = {}
query_2_vars = {}

merged = GraphQL::Language.merge_queries(
  [query_1_text, query_2_text],
  [query_1_vars, query_2_vars]
)

# Returns an array of responses, in the same order as the queries you provided
result = merged.execute(YourSchema, context: your_context)
```

So I thought I'd ask for some feedback:
- Would it be useful for this to be part of the gem's codebase?
- Any problems (or other feedback) with how it's implemented?

One particular challenge is merging queries that specify different operations (eg, a mutation with a query). And I'm not really sure the impact on subscriptions, `@defer`, `@stream`, etc.

rmosolgo:
I really like the idea of providing some support for "messing" with queries! It's _possible_ to do for people who learn the AST classes, but that's a lot to ask of someone. Besides that, I think some AST-based operations implemented in library externals could be extracted to public (or at least reusable) units of code. 

Here are some things I can think of: 

- Merging multiple queries into one (as described here) 
- Splitting operations out of a collection of definitions (already implemented by this library) 
- Resolving fragment dependencies (implemented in StaticValidation _and_ InternalRepresentation, and in the `typecheck` branch) 
- Adding type information to a query (implemented in TypeStack, requested in some capacity for graphql-client) 
- Programmatically changing selections sets (I think graphql-client adds `__typename` in some cases)
- Asserting the validity of a document _apart_ from a GraphQL::Schema (eg, names are unique, fragments don't form cycles) implemented in `typecheck` branch, I think graphql-client could benefit from that. 

I want to brainstorm any other foreseeable uses of the AST and try to find a framework for covering them. (But I also suspect that, when the tools are available, people will find cool new ways to use them!) 

ps @josh this might be relevant to your interests!

jsdnxx:
[...]

PR #691, opened 2017-04-25
feat(Schema#multiplex) run multiple queries in the same batching context; validate complexity on the multiplex

rmosolgo:
A hot take on #406 , but at the runtime-level instead of AST level. I've also reimplemented single-query execution to be a special case of `Multiplex`: a case with only _one_ query. 

TODO: 

- [x] Add multiplex-level analysis, make query complexity a multiplex-level check


rmosolgo:
Adding a multiplex-level analysis phase is hard 😓  

I want to go ahead and add it now because that's where the query complexity check should go. The _sum_ of complexities across the multiplex should used, the complexity of _each_ query (otherwise, you could send as many complex queries as you want!) 

But I need to tease apart the validation steps. Right now, calling `Query#result` kicks off the flow of validation etc, but now we need to do trigger these steps from outside code (from the multiplex execution).

rmosolgo:
💰 Schema's `max_complexity` is applied to the multiplex, `Schema#multiplex` accepts a `max_complexity:` override

PR #694, opened 2017-04-27
Fix bundler source URL for pro

toolmantim:
Trying to use the old shorthand version of Gem sources results in the following error on recent Bundler versions:

```
[!] There was an error parsing `Gemfile`: The source must be an absolute URI. For example:
source 'https://rubygems.org'. Bundler cannot continue.

 #  from Gemfile:5
 #  -------------------------------------------
 >  source "gems.graphql.pro" do
 #    gem "graphql-pro"
 #  -------------------------------------------
```


rmosolgo:
Oh, thanks! I better update my bundler, too!

PR #695, opened 2017-04-27
Function docs: update types usage

swalkinshaw:
`types` is exposed in functions as of `1.5.6` (in https://github.com/rmosolgo/graphql-ruby/pull/654)

I figured the whole section on Types wasn't even needed anymore?


swalkinshaw:
Found two more function related things to update 👌

rmosolgo:
Awesome, thanks for giving this some attention! It looks like the Function generator tests may need an update for the new output as well.

swalkinshaw:
Updated 🎉

rmosolgo:
:metal: published, thanks again!

PR #697, opened 2017-04-27
Update graphql-batch usage in generators

swalkinshaw:
`v0.3.2` includes the `use` functionality

Not sure if the `# GraphQL::Batch setup:` comment is still needed or not?


rmosolgo:
Sweet! Let's keep the comment just for verbosity's sake.

Issue #491, opened 2017-01-16
Ambiguous Type Resolution?


MatthewChang:
Forgive my ignorance if this is not a true issue.

It appears as this implementation of graph ql departs from the standard by having a single type resolution function defined on the schema. This in itself is not a problem, but the context passed into that function seems to not be the context at the evaluating node. I.e. 
```ruby
GraphQL::Schema.define do
  query QueryType
  resolve_type -> (object,ctx) {
    puts ctx.irep_node
  }
end
```
is always resulting in nil in my schema (ctx.ast_node raises a NoMethodError). I would guess that this is the context of the schema.

Given that, if I were to have two union types which operate on the same kind of resolved object, such as unions for query and create operations, it is ambiguous simply from the type of the object which type to resolve to. This could be circumvented by embedding some meta information in the object as to which kind of operation it is a part of, but that seems like a patently bad design pattern. 
Have I missed something with the proper way to specify type resolution? Thanks for your help!

rmosolgo:
You raise a good point! Yes, that's the query-level context. It contains any keys that you passed as `context:`, but it doesn't have the field-level values. It seems like it should, huh? 

Here's the bad line of code: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/execute.rb#L178

It calls through to `query`, which provides its own context. But that method signature could be improved to `query.resolve_type(value, field_ctx)`, that way it would work as you expect! 

Perhaps another way to improve it would be to inject the composite type which is being resolved, eg: 

```ruby
resolve_type ->(obj, type, ctx) { 
  # 
}
```

Where `type` is a union or interface type. (If `ctx` were improved as described above, I think you could get this value via `ctx.irep_node.owner_type`.)

BTW, the current API was designed to _prevent_ the same feature requested here 😆  It was previously possible to resolve the _same_ objects to _different_ types -- but when that happened accidentally, it resulted in nasty bugs! From your example, I see how that ability can be _useful_ in the right situation!

MatthewChang:
Ah I see. Thanks for the prompt and thorough response! I haven't quite needed this functionality yet (and designing around it does seem like the preferable option) but may make a PR if I end up using it. Cheers!

khamusa:
[...]

PR #699, opened 2017-04-28
Make resolvemap work more like graphql-tools

emilebosch:
Work in progress! Need some help from @xuorig and @rmosolgo 

In this PR i will try to make resolvemap behave a little more like graphql-tools. I.e. make it easy to build resolvers from a hash much like the JS version - https://github.com/apollographql/graphql-tools. The end goal is to clone all functionality from its Js counterpart for rapid api prototyping.

- [x] Use symbols instead of strings
- [x] When no resolver specified try to call `field.name` on `obj`
- [x] Call said resolver with args, if given
- [x] Cache all the responds_to? etc?
- [x] Fix empty resolver `Counter`
- [x] Implement `__resolveType` for unions @rmosolgo any idea on how to implement this? see: http://dev.apollodata.com/tools/graphql-tools/resolvers.html#Unions-and-interfaces 
- [x] Implement custom scalar resolving @rmosolgo any idea on how to implement this? see: http://dev.apollodata.com/tools/graphql-tools/scalars.html
- [x] Handle subscriptions?

See for context https://gist.github.com/rmosolgo/ba31acf93f07f8007d99ba365a662d8f#file-app-rb-L126-L174

From this map:

```ruby
 Resolvers = {
    "Mutation" => {
      "incrementCounter" => ->(o, a, c) {
        counter = App::Counter.find(a["id"])
        counter.increment
        API::Schema.subscriber.trigger("counterIncremented", a, counter)
        counter
      }
    },
    "Query" => {
      "counter" => ->(o, a, c) { App::Counter.find(a["id"]) }
    },
    "Counter" => {
      "value" => ->(o, a, c) { o.value },
      "id" => ->(o, a, c) { o.id },
    }
  }
```

to this map:

```ruby
Resolvers = {
  Mutation: {
    incrementCounter: ->(o, a, c) {
      counter = App::Counter.find(a["id"])
      counter.increment
      API::Schema.subscriber.trigger("counterIncremented", a, counter)
      counter
    }
  },
  Query: {
    counter: ->(o, a, c) { App::Counter.find(a["id"]) }
  }
}
```

@xuorig You've been making something like this too right, what do you think?

### DIFF
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -46,6 +46,13 @@ module GraphQL
               @resolve_hash[type_name_s] = fields
             end
           end
+
+          # Check the normalized hash, not the user input:
+          if @resolve_hash.key?("resolve_type")
+            define_singleton_method :resolve_type do |type, ctx|
+              @resolve_hash.fetch("resolve_type").call(type, ctx)
+            end
+          end
         end
 
         def call(type, field, obj, args, ctx)
@@ -54,7 +61,6 @@ module GraphQL
         end
 
         def after_build_schema(schema)
-          hookup_union(schema)
           hookup_scalars(schema)
         end
 
@@ -96,12 +102,6 @@ module GraphQL
               # Call through this time
               resolver.call(obj, args, ctx)
             end
-          end
-        end
-
-        def hookup_union(schema)
-          if @resolve_hash.key?("__resolve_type")
-            schema.resolve_type = @resolve_hash["__resolve_type"]
           end
         end
 
@@ -189,7 +189,11 @@ module GraphQL
             mutation mutation_root_type
             subscription subscription_root_type
             orphan_types types.values
-            resolve_type NullResolveType
+            if default_resolve.respond_to?(:resolve_type)
+              resolve_type(default_resolve.method(:resolve_type))
+            else
+              resolve_type(NullResolveType)
+            end
 
             directives directives.values
           end
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -22,22 +22,42 @@ module GraphQL
 
       # @api private
       class ResolveMap
+        
         def initialize(resolve_hash)
-          @resolve_hash = resolve_hash
+          @resolve_hash = convert_keys_to_strings(resolve_hash)
+        end
+
+        def convert_keys_to_strings(h)
+          Hash[
+            h.map {|k, v| 
+              v = convert_keys_to_strings(v) if v.is_a?(Hash)
+              [k.to_s, v] 
+            }
+          ]
         end
 
         def call(type, field, obj, args, ctx)
-          type_hash = @resolve_hash[type.name.to_sym]
-          type_hash && (resolver = type_hash[field.name.to_sym])
+          type_hash = @resolve_hash[type.name]
+
+          if !type_hash
+            type_hash = @resolve_hash[type.name] = {}
+          end
+
+          type_hash && (resolver = type_hash[field.name])
 
           if resolver.nil?
-            if !obj.respond_to? field.name
-              raise(KeyError, "resolver not found for #{type.name}.#{field.name}")
-            else
-              obj.send field.name
-            end
+            raise(KeyError, "resolver not found for #{type.name}.#{field.name}") unless obj.respond_to?(field.name)
+            resolver = type_hash[field.name] = build_resolver(type, field, obj)
+          end
+
+          resolver.call(obj, args, ctx)
+        end
+
+        def build_resolver(type, field, obj)
+          if obj.method(field.name.to_sym).arity > 0 
+            return -> (o, a, c) { o.send(field.name, a) }
           else
-            resolver.call(obj, args, ctx)
+            return -> (o, a, c) { o.send(field.name) }
           end
         end
       end
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -55,9 +55,9 @@ module GraphQL
 
         def build_resolver(type, field, obj)
           if obj.method(field.name.to_sym).arity > 0 
-            return -> (o, a, c) { o.send(field.name, a) }
+            return ->(o, a, c) { o.send(field.name, a) }
           else
-            return -> (o, a, c) { o.send(field.name) }
+            return ->(o, a, c) { o.send(field.name) }
           end
         end
       end
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require "graphql/schema/build_from_definition/resolve_map"
+
 module GraphQL
   class Schema
     module BuildFromDefinition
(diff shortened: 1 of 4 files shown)


rmosolgo:
:+1: Cool! Thanks for taking a crack at this and outlining those goals, they all sound like wins to me. I think we can make a few modifications to the current schema builder, and some additions for `resolveType` and scalars, and we'll be on our way :) 

Can you help me understand this one: 

> Call said resolver with args, if given

For you, how do you think it should work?

rmosolgo:
A few thoughts: 

- To support symbols or strings for user-provided maps, how about normalizing everything to a string during the initial build? That way, we don't have to do any conversions at runtime. We only need to use the `.name`s to look up values in the map. 
- Default resolve behavior is ok with me. Should we still require the type name in the resolver map if all the resolves will use the default behavior? For example, in the example above, could we remove `Counter: {}` ?
- For backwards compatibility, we could check the user-provided map for [`Hash#default_proc`](https://ruby-doc.org/core-2.4.0/Hash.html#method-i-default_proc) and if it's present, don't provide a new default.
- Some approaches to caching the `respond_to?` checks: 
  - Inspect the schema at build-time to provide the right default
  - Make the check on the first execution, but then modify the resolver map with a new function which _doesn't_ include the check 
- `__resolveType` will be tough because we don't have type-level resolve_type functions yet, see #491 
- Perhaps scalars could be implemented as: 

  ```ruby 
  "MyScalar" => {
    "coerce_input" => ->(val, ctx) { ... },
    "coerce_result" => ->(val, ctx) { ... },
  }
  ```

emilebosch:
🙌  🔥 Awesome that you like it too!

> To support symbols or strings for user-provided maps, how about normalizing everything to a string during the initial build? That way, we don't have to do any conversions at runtime. We only need to use the .names to look up values in the map.

Will do that

> Default resolve behavior is ok with me. Should we still require the type name in the resolver map if all the resolves will use the default behavior? For example, in the example above, could we remove Counter: {} ?

Totally, my idea too

> For backwards compatibility, we could check the user-provided map for Hash#default_proc and if it's present, don't provide a new default.

Can you elaborate a bit on that? WHen is it needed in what cases and what do i need to check?

> Inspect the schema at build-time to provide the right default

You mean looping trough the schema, check the classes provided in the schema and check if those classes have methods on them that correspond to `reponds_to?` we might missing out the `missing_method` dynamic methods there then. Or am i misunderstanding your approach.

> Make the check on the first execution, but then modify the resolver map with a new function which doesn't include the check

Makes most sense for me, i think i'l go this route.

> __resolveType will be tough because we don't have type-level resolve_type functions 

Ok, will go and hack it with a global resolvetype

> Perhaps scalars could be implemented as:

Fine by me, not sure about the naming yet though. Is there also a way we can access the AST? via context?

In JS they do this magic

```javascript
  parseLiteral(ast) {
      if (ast.kind === Kind.INT) {
        return parseInt(ast.value, 10); // ast value is always in string format
      }
      return null;
    },
```

Oh, and i'll fix the burning tests ofc, 🚒

emilebosch:
@rmosolgo  ok, i've come to scalar types now. I see that the standard is a `NullScalarCoerce = ->(val, _ctx) { val }` in build_from_definition.rb. However i need to hook up that the 

```
 def build_scalar_type(scalar_type_definition, type_resolver)
          GraphQL::ScalarType.define(
            name: scalar_type_definition.name,
            description: scalar_type_definition.description,
            coerce: NullScalarCoerce,
          )
        end
```

Somehow ends up in my ResolveMap. Also i think maybe my resolve map wants to know about schema, so it can hook up all the resolvers, scalars and union resolvers.

emilebosch:
Alright so i got scalars and unions working via the following hack:

This is the resolver map
```ruby
 Resolvers = {
    __resolve_type: -> (obj, ctx) {
      ctx.schema.types['A'] # Or whatever you think obj should be
    },
    Mutation: {
      incrementCounter: ->(o, a, c) {
        counter = App::Counter.find(a["id"])
        counter.increment
        API::Schema.subscriber.trigger("counterIncremented", a, counter)
        counter
      }
    },
    A: {
      id: -> (o,a,c) {
        3
      }
    },
    Date: {
      coerce_input:  -> (value, ctx) { Time.at(Float(value)) },
      coerce_result: -> (value, ctx) { value.to_f }      
    },
    Query: {
      counter: ->(o, a, c) { App::Counter.find(a["id"]) }
    }
  }
```

And then setting the resolvers on the schema afterwards. Ofcourse, this need to be somehow implemented in the `build_from_definition`. But no idea where yet.

```ruby
Schema = GraphQL::Schema.from_definition(Definition, default_resolve: ResolveMap.new(Resolvers)).redefine do
  use GraphQL::Subscriptions,
    transports: { "stream" => App::StreamTransport },
    queue: App::Queue,
    store: App::SubscriptionDatabase
end

Schema.resolve_type                         = Resolvers[:__resolve_type]
Schema.types['Date'].coerce            = nil
Schema.types['Date'].coerce_input  = Resolvers[:Date][:coerce_input]
Schema.types['Date'].coerce_result = Resolvers[:Date][:coerce_result]
```

@rmosolgo any idea how we can automate this? I was maybe thinking of just pasing ResolveMap something like schema and not just be a resolve map but more like a schema configurator setter upper?

emilebosch:
Alrighty everything works now but hacked it together in regards to hooking up scalar resolving and union resolving. 

https://github.com/rmosolgo/graphql-ruby/pull/699/commits/9481f70e94197bc464c48ea377d846e9c0327fe2

rmosolgo:
Hehe, well, at least it works :P In order to consider merging this patch, it's important for me that these new features be described in the specs. Otherwise, later changes to the code may result in silent regressions! And I might be working in this area again soon -- the duplication between `Loader` and `BuildFromDefinition` is a pet peeve of mine 😆 

Are you interested in adding that?

emilebosch:
Sure, i'll add some tests :) Expect them to arrive soon

spec/graphql/schema/build_from_definition_spec.rb
```diff
@@ -766,14 +832,14 @@ def todo_add(args, ctx) # this is a method and accepting arguments
...
-        end
-      end
+      # describe "wihtout defaults" do
+      #   let(:base_hash) { {} }
+      #   it "raises a KeyError" do
+      #     assert_raises(KeyError) do
+      #       schema.execute("mutation { todoAdd: todo_add(text: \"Buy Milk\") { text } }", context: {context_value: "bar"}, root_value: todos)
+      #     end
+      #   end
+      # end
```
rmosolgo: It's ok with me to remove this behavior, but can we remove the test altogether? Any reason to keep it commented out?

lib/graphql/schema/build_from_definition.rb
```diff
@@ -22,18 +22,60 @@ def self.call(type, field, obj, args, ctx)
...
+          end
+
+          resolver.call(obj, args, ctx)
+        end
+
+        def build_resolver(type, field, obj)
+          if obj.method(field.name.to_sym).arity > 0
+            return ->(o, a, c) { o.send(field.name, a) }
+          end
+          ->(o, a, c) { o.send(field.name) }
```
rmosolgo: Can we use `.public_send` here? If possible, I prefer it, because it prevents calling things like `Kernel#exit` by mistake

lib/graphql/schema/build_from_definition.rb
```diff
@@ -22,18 +22,60 @@ def self.call(type, field, obj, args, ctx)
...
+          end
+
           type_hash && (resolver = type_hash[field.name])
 
           if resolver.nil?
-            raise(KeyError, "resolver not found for #{type.name}.#{field.name}")
-          else
-            resolver.call(obj, args, ctx)
+            raise(KeyError, "resolver not found for #{type.name}.#{field.name}") unless obj.respond_to?(field.name)
+            resolver = type_hash[field.name] = build_resolver(type, field, obj)
```
rmosolgo: I wonder if we could work some of this into the `@resolve_hash` itself, for example, if it was defined like this: 

```ruby
@resolve_hash = Hash.new do |h, k| 
  # For each type name, provide a new hash if one wasn't given: 
  h[k] = Hash.new do |h2, k2| 
    # For each field, provide a resolver that: 
    # - Makes a runtime check on the object then: 
    #    - if it passes, replaces itself & call through 
    #    - if it fails, raises a keyerror 
    h2[k2] = ->(obj, args, ctx) {
      if !obj.respond_to?(k2) 
        raise KeyError, "..."
      else  
        method_arity = obj.method(k2).arity 
        resolver = case method_arity
        when 0
          ->(o, a, c) { o.public_send(k2) } 
        when 1 
          ->(o, a, c) { o.public_send(k2, a) } 
        when 2 
          ->(o, a, c) { o.public_send(k2, a, c) } 
        else 
          raise "Unexpected resolve arity: #{method_arity}. Must be 0, 1, 2"
        end 
        # Call the resolver directly next time 
        h2[k2] = resolver 
        # Call through this time
        resolver.call(obj, args, ctx)
      end 
    } 
  end 
end 
```

Then, `#call` can be: 

```ruby
def call(type, field, obj, args, ctx) 
  @resolver_hash[type.name][field.name].call(obj, args, ctx)
end 
```

that keeps the default `call` behavior pretty simple and lets us use Ruby built-ins for providing fallback values in the hash. What do you think, would it work here?

lib/graphql/schema/build_from_definition.rb
```diff
@@ -115,6 +157,12 @@ def build(document, default_resolve: DefaultResolve)
 
             directives directives.values
           end
+
+          if default_resolve.respond_to? :after_build_schema
```
rmosolgo: Hmm, here's an alternate approach, what do you think: 

- default_resolve objects _may_ implement `#resolve_type`, and if they do, it's hooked up to the schema: 

  ```ruby 
  Schema.define do 
    if default_resolve.respond_to?(:resolve_type)
      resolve_type ->(obj, ctx) { default_resolve.resolve_type(obj, ctx) } 
    end 
  end 
  ```
  
  Then, `ResolveMap` could implement it to fetch from the provided hash: 
  
  ```ruby
  def resolve_type(obj, ctx) 
    @resolver_hash.fetch("__resolve_type").call(obj, ctx)
  end 
  ```

  Alternatively, you could check the hash during `#initialize` and conditionally define the method with `define_singleton_method`. This would be nice because you'd get `GraphQL::Schema`'s built-in check for the presence of `resolve_type`. Either way. 

- Ok, I don't have a good idea for scalars yet ... let me think on it ...

lib/graphql/schema/build_from_definition.rb
```diff
@@ -22,18 +22,60 @@ def self.call(type, field, obj, args, ctx)
 
       # @api private
       class ResolveMap
+        
         def initialize(resolve_hash)
-          @resolve_hash = resolve_hash
+          @resolve_hash = convert_keys_to_strings(resolve_hash)
+        end
+
+        def convert_keys_to_strings(h)
```
rmosolgo: Can some of these new methods be hidden under a `private` declaration? That will make it a bit easier to maintain, since the reader knows those methods are only called within the class itself.

lib/graphql/schema/build_from_definition.rb
rmosolgo: Refactoring that default logic a bit, extracting it to a class: 

```ruby 

@resolve_hash = Hash.new do |h, k| 
  # For each type name, provide a new hash if one wasn't given: 
  h[k] = Hash.new do |h2, k2| 
    # For each field, provide a resolver that will 
    # make runtime checks & replace itself 
    h2[k2] = SelfReplacingDefaultResolveFunction.new(h2, k2)
  end 
end

class SelfReplacingDefaultResolveFunction 
  def initialize(field_map, field_name) 
    @field_map = field_map
    @field_name = field_name 
  end 

  # Make some runtime checks about 
  # how `obj` implements the `field_name`.
  # 
  # Create a new resolve function according to that implementation, then: 
  #   - update `field_map` with this implementation 
  #   - call the implementation now (to satisfy this field execution)
  #
  # If `obj` doesn't implement `field_name`, raise an error.
  def call(obj, args, ctx) 
    if !obj.respond_to?(k2) 
      raise KeyError, "..."
    else  
      method_arity = obj.method(k2).arity 
      resolver = case method_arity
      when 0
        ->(o, a, c) { o.public_send(k2) } 
      when 1 
        ->(o, a, c) { o.public_send(k2, a) } 
      when 2 
        ->(o, a, c) { o.public_send(k2, a, c) } 
      else 
        raise "Unexpected resolve arity: #{method_arity}. Must be 0, 1, 2"
      end 
      # Call the resolver directly next time 
      @field_map[@field_name] = resolver 
      # Call through this time
      resolver.call(obj, args, ctx)
    end 
  end 
end 
```

emilebosch:
@rmosolgo Ok, moved stuff to private, added arity checks.

- I didn't implement the `resolve_type` yet, since its because of scalars we still need some sort of after hook or something more elegant. 
- I didn't implement the fancy hash replace self function, because i found it a bit hard to follow, but feel free to add it!

rmosolgo:
Ok, thanks for waiting for me to try some things! Here are some changes I just pushed, what do you think? 

- API change: use `resolve_type` key instead of `__resolve_type`. I don't think Ruby projects often use double-underscore methods, do you agree? Can we use a plain key here? (It means that nobody can name a type `"resolve_type"`, that's ok with me 😆 )
- API change: when a scalar's coerce functions aren't provided, return the Ruby value as-is 
- Refactor: instead of looking for `#after_schema_build`, a `default_resolve` object may define `#resolve_type`, `#coerce_input` and `#coerce_result` methods. (This doesn't change any behavior, but it sets up an API for using Ruby namespaces instead of hash keys for resolvers.)
- Refactor: move `ResolveMap` & friends into their own files

emilebosch:
🎉🎉🎉✌️️✨good changes! Elegant as ever!

Issue #459, opened 2016-12-11
Fragment-based codegen 


rmosolgo:
I use Immutable.js and it would be nice to generate `Immutable.Record`s from the GraphQL queries that populate them. 

Similarly, you could use fragments to define Ruby classes or Protobuf messages. 

Should graphql-ruby contain tools for this? Or should we wait for a third-party tool (perhaps the "language service") to do it for us?

armstrjare:
Relevant: https://github.com/apollostack/apollo-codegen

rmosolgo:
Cool! I hadn't seen that before, thanks for sharing the link.

emilebosch:
Also relevant: https://github.com/swagger-api/swagger-codegen

rmosolgo:
I'm not able to really work on this, so I'm closing it.

PR #701, opened 2017-04-28
Expose IRep Node fragment definition roots

josh:
👋 Hi @rmosolgo,

So I started working some algorithms for graphql-client to do typed document flattening. I realized that I was basically reimplementing graphql-ruby's internal node representation. `InternalRepresentation::Node#typed_children` was exactly what I needed and was trying to implement.

* First, I was wondering how much of an internal class `GraphQL::InternalRepresentation::Node` is? Should I depend on its API? I believe `irep_node`s are exposed to resolvers via `context`. So its probably something that should be considered stable, maybe?
* The APIs to get hold of the internal representation are a little cumbersome. It seems like the easily way is to call a validator and it from the result. That actually works *okay* for graphql-client as we do invoke the validator when parsing queries, so I can just snag it at that time and avoid multiple calls to the rewriter. https://github.com/github/graphql-client/blob/8d72a7992417fd5e7829d039948b7c888a33dc2c/lib/graphql/client.rb#L155-L162
* Lastly, graphql-client handles parsing fragments without an associated operation. `FragmentsAreUsed` is skipped when validating. Thats all fine, but the rewriter discards `fragment_definitions` state after rewriting. This PR purposes holding that reference so it can be used.

I'm really wondering if all this is just a horrible abuse of an internal API? Is there another way you've always dream that the internal representation algorithms could be exposed in a way that would be nice for me to use? 😁


rmosolgo:
I think it would be great to think about this API a bit and stabilize it. This data is really powerful but I haven't really had the bandwidth to document it or explore it much myself!

:+1: for including fragment definitions in the result, I don't see why not anyways. 

:+1: for decoupling validation from rewrite, but I think it's important to do them both at the same time when possible. The current implementation of rewriting is tricky though: it uses variable values in order to bypass `@skip`ped nodes. This needs to be fixed somehow -- it won't scale to custom directives anyways. Maybe we can find a way to make the rewrite a static operation but the trick is keeping `@skip`ped children separate from non-skipped ones (https://github.com/rmosolgo/graphql-ruby/pull/688#issuecomment-296379546)

:+1: for improving documentation & stability of these APIs. Can you help me understand your use case? I'm also hoping to write a more powerful analysis layer on top of it soon, so maybe we can keep these both in mind.

josh:
> The current implementation of rewriting is tricky though: it uses variable values in order to bypass @skipped nodes. This needs to be fixed somehow -- it won't scale to custom directives anyways. Maybe we can find a way to make the rewrite a static operation but the trick is keeping @skipped children separate from non-skipped ones

I haven't thought too much about this, but I need a similar annotation client side as well. On the client, we need to process the query before its executed, so no variables are available to evaluate those directives.

I also agree about custom directives and I struggle to come up with practical hooks to make `@skip` implemented as an extension rather than tooling knowing about it as a first class directive. Maybe there needs to be a hook for "directive ast processing" where we visit all irep nodes with directives and execute custom directive extensions that could manipulate the ast. I haven't really thought this through.

> Can you help me understand your use case? I'm also hoping to write a more powerful analysis layer on top of it soon, so maybe we can keep these both in mind.

Yeah! So I've realized I've implemented a very buggy field flattening algorithm and wanted to use the `typed_children` helper here.

Given (this github specific schema example)

``` graphql
fragment on Repository {
  owner {
    __typename
    id
    avatarURL

    ... on User {
      name
    }

    ... on Organization {
      companyName
    }
  }
}
```

Lets say we want to generate a response wrapper class for the `Repository.owner` here. In this case `owner` returns a `RepositoryOwner` interface with a field guaranteed fields like `id` and `avatarURL`. Both `User` and `Organization` implement this.

Theres two possible response types that could be returned, a `User` or `Organization`. Both have different combinations of fields.

``` ruby
User = Struct.new(:id, :avatar_url, :name)
Organization = Struct.new(:id, :avatar_url, :company_name)
```

Using `InternalRepresentation::Node`, `typed_children` returns:

``` ruby
{
  UserType => [<Node id>, <Node avatarURL>, <Node name>],
  OrganizationType => [<Node id>, <Node avatarURL>, <Node companyName>],
}
```

Which is easy to recursively map over and create those wrapper structs.

rmosolgo:
👌  that seems like a perfect use case, basically codegen, but instead of the pure AST (a la #459), it's based on the processed AST.

What's the API we need? Something like: 

```ruby 
query_plan = MySchema.plan(query_string)
PlannedClass = PlannedClass.build(query_plan)
```

?

How would you handle `@skip` etc in the derived class? Are they `nil`-able attributes (even if the field is non-null)? Or do something like booby-trap the accessor so hitting it raises an error?

josh:
I don't think graphql-ruby needs to implement the struct generation code. Right now graphql-client is doing all that already. I mean, maybe we can upstream it someday. But exposing a typed query flattening API would enable it.

> How would you handle @skip etc in the derived class? Are they nil-able attributes (even if the field is non-null)? Or do something like booby-trap the accessor so hitting it raises an error?

In either case you just need some way to detect it. The classes would have a `irep_node` reference. As is, you check `Node#ast_nodes`, `klass.irep_node.ast_nodes.any? { |n| n.directives.map { |d| is_skip_directive?(d) } }`

rmosolgo:
> struct generation code

Oh, my bad! I should have posted this: 

```
# from graphql-ruby:
query_plan = MySchema.plan(query_string)
# user code does whatever:
PlannedClass = PlannedClass.build(query_plan)
```

Is that enough from the gem? I guess there can't be variables, context, operation name -- those are all runtime considerations.

josh:
> Is that enough from the gem? I guess there can't be variables, context, operation name -- those are all runtime considerations.

Yeah, it could even accept a parsed AST document instead of a raw query string.

rmosolgo:
(Or, maybe a straightforward API to the current behavior wouldn't do any good because of that static/runtime issue described above 😖 )

josh:
Did some testing around with this proposed change within graphql-client to see if it would work. Good news, it does! Heh. At minimum just changing `fragment_definitions` to be tracked as `@ fragment_definitions` is most helpful.

Even though the API is strange, I don't mind this data structure being a side effect of the validation process. I need to run that anyway, so reusing it means I don't have to invoke the rewriter twice. Though, it seems feel like some little internal secret that only I'm aware of.

Another idea, maybe this IRep AST should be exposed off of `GraphQL::Query`. Say `Query#irep_definitions`. Instead of [`StaticValidation::Validator` newing up a `Rewrite`](https://github.com/rmosolgo/graphql-ruby/blob/44b7052350fb839da1d4449e970c1c85f294d0b4/lib/graphql/static_validation/validator.rb#L26), it would just use the cached `query.irep_definitions`. That avoids the rewriter ever being called multiple terms per query instance. I'm mostly unsure of what other side effects there are of invoking `Rewrite` and holding a reference to its irep ast. Does that seem safe?

rmosolgo:
Yeah, exposing that from `GraphQL::Query` seems good by me, I'm happy to support either one. But, I'd appreciate it if you'd help me help you by adding a spec for whichever API you end up depending on in graphql-client. I think this output is mostly treated as private so far and I'll probably break it if it doesn't have a spec of its own 😬

rmosolgo:
Wouldn't have to be much, just something so that if I end up reworking it, I get a reminder, "hey ya dummy, somebody's using this!"

josh:
> But, I'd appreciate it if you'd help me help you by adding a spec for whichever API you end up depending on in graphql-client.

Ah, yeah, sorry, I was mostly waiting for API direction before polishing this up. Alright, let me see what can be done by moving it to `Query`.

josh:
Oh, hey, `Query#internal_representation` already exists! See revisions.

lib/graphql/internal_representation/rewrite.rb
```diff
@@ -137,9 +141,9 @@ def validate(context)
         # can be shared between its usages.
         context.on_dependency_resolve do |defn_ast_node, spread_ast_nodes, frag_ast_node|
           frag_name = frag_ast_node.name
-          fragment_node = fragment_definitions[frag_name]
+          fragment_node = @definitions[frag_name]
 
-          if fragment_node
+          if fragment_node && fragment_node.ast_node.is_a?(Nodes::FragmentDefinition)
```
rmosolgo: Does something break without this check? (I wonder what _happens_ if we have operations and fragments with the same name ... :S )

josh: Crap. I think you're right. I've been assuming **definition** names where a unique set up until this point. I can't find anything in the GraphQL spec about this. It only mentions operation uniqueness and fragment definition uniqueness.

I will change this back.

It this is all true, I think theres probably a few graphql-ruby implementation bugs around this issue. I just tested on the github.com API and ran into this:

``` graphql
query Foo {
  viewer {
    ...Foo
  }
}

fragment Foo on User {
  login
}
```

``` json
{
  "data": null,
  "errors": [
    {
      "message": "Fragment Foo contains an infinite loop",
      "locations": [
        {
          "line": 7,
          "column": 1
        }
      ]
    }
  ]
}
```

josh:
If we can might a slight API change in 1.6.0, it'd be nice to change the return type of`Query#internal_representation`. Since a basic `Hash[String: IRep::Node]` is not sufficient to represent clashing operation and definition names.

Maybe we could return a new `IRep::Document` type that has separate `#operations` and `#fragments` as `Hash[String => IRep::Node]`?

rmosolgo:
:+1: Breaking change is fine with me in that area. Backwards compat is possible (by delegating `#[]` to operations) but not important in my opinion. :+1: for introducing a proper class for it!

josh:
> Backwards compat is possible (by delegating #[] to operations) but not important in my opinion.

Haha, that's pretty clever.

josh:
> 👍 for introducing a proper class for it!

Done.

I like how `InternalRepresentation::Document` is looking. I added deprecated aliases for all the old APIs being used internally. Is `warn` your preferred style to print deprecation warnings?

josh:
Cool, tests still pass after merging https://github.com/rmosolgo/graphql-ruby/pull/706.

rmosolgo:
I used `warn` because I don't know anything better 😬 Do you have another favorite technique? Anything's fine by me as long as people will see it!

josh:
> Do you have another favorite technique?

Its probably *fine*. Usually I like to offer something people can disable or change to `raise` instead of `warn`. With Rails, `ActiveSupport::Deprecation` has these features.

Another thing that is helpful is to bring a stack trace. Basically just `warn(caller)` so people can see what is calling the deprecated method.

rmosolgo:
🙌

josh:
🎉 thanks ❤️!

PR #704, opened 2017-05-01
Preserve InternalRepresentation::Node#return_type original wrapping

josh:
As `InternalRepresentation::Node`s are constructed, their `return_type` is assigned to an unwrapped type. These feels like a loss of information as its no longer possible to tell `#<Node User.repositories -> Repository>` was actually `#<Node User.repositories -> [Repository]>`.

Initially implemented as change to preserve the originally wrapping type on `return_type` as its easy to `unwrap` and get the inner object type. However, I'd like to get some feedback on this being a problematic backwards incompatible change. We could possibly maintain the original behavior of `return_type` but add new `wrapped_return_type` accessor. `return_type` could be implemented as `wrapped_return_type.unwrap`.

Let me know which direction you think I should try going here and I can add tests and documented the expected behavior. Thanks!


rmosolgo:
Good point about the data loss, it totally is a misnomer. I thought I had a dependency on this in graphql-pro, but I don't. Ironically, I go around my elbow to reach my thumb: 

```ruby 
irep_node.definition.type.unwrap 
# derp, could've used #return_type 
```

I'd be happy to take either of these approaches: 

- Add a new accessor now and release it soonish 
- Change the behavior of `#return_type` and release it in 1.6.0, hopefully Wednesday or Thursday. I think a good entry in the changelog will make the upgrade quite smooth.

Implementer's choice! 😉

josh:
Ah, `irep_node.definition.type.unwrap ` is a good workaround.

> Change the behavior of #return_type and release it in 1.6.0, hopefully Wednesday or Thursday. I think a good entry in the changelog will make the upgrade quite smooth.

I'll go for the behavior change here and polish this up. Since theres a workaround, I can wait for 1.6.

josh:
Alright, @rmosolgo, ready for review again. I'm good to have this merged into 1.6.0.

rmosolgo:
Awesome! I'll take a look at benchmarks and memoize that unwrapped result if it makes a difference.

rmosolgo:
(no regression on `1.6.x` compared to `master`)

PR #705, opened 2017-05-01
Improve pretty print debugging of ListType and NonNullType

josh:
Its bugged me that `p GraphQL::STRING_TYPE` gives a nice `String` output while `p !types.String` just prints *nothing* for a strange reason.

It looks like these subclasses forgot to realias the `inspect` method. `p !types.String` now outputs `String!`. 😎


rmosolgo:
:+1: This was a side-effect of the fix in #355 but your patch gives us the best of both worlds!

PR #706, opened 2017-05-02
Support Operation and Fragment definitions of the same name

josh:
As per @rmosolgo's wise comment on https://github.com/rmosolgo/graphql-ruby/pull/701#discussion_r114229506.

It turns out operations and fragment names live in separate namespaces. An operation can share a fragments name. I tested and confirmed on the reference implementation (graphql-js) and it does indeed work.

### DIFF
--- a/lib/graphql/static_validation/definition_dependencies.rb
+++ b/lib/graphql/static_validation/definition_dependencies.rb
@@ -12,17 +12,19 @@ module GraphQL
       end
 
       def initialize
+        @node_paths = {}
+
         # { name => node } pairs for fragments
         @fragment_definitions = {}
 
         # This tracks dependencies from fragment to Node where it was used
-        # { frag_name => [dependent_node, dependent_node]}
+        # { fragment_definition_node => [dependent_node, dependent_node]}
         @dependent_definitions = Hash.new { |h, k| h[k] = Set.new }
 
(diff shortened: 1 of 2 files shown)


josh:
Current failure:

```
Failure:
GraphQL::Query::Executor::operation and fragment defintions of the same name#test_0001_runs the named operation [spec/graphql/query/executor_spec.rb:78]
+++ actual
@@ -1 +1,4 @@
-{"data"=>{"cheese"=>{"flavor"=>"Brie"}}}
+{"errors"=>
+  [{"message"=>"Fragment Cheese contains an infinite loop",
+    "locations"=>[{"line"=>4, "column"=>7}],
+    "fields"=>["fragment Cheese"]}]}
```

josh:
Wow, that was tricky.

I basically refactored the definition dependency tracker to manage nodes rather than string names so both operation definition and fragment definition named `"Foo"` can be tracked in `@immediate_dependencies` separately. `NodeWithPath` got tricky as we don't know `context.path` until the visit runs but we need to know all the `FragmentDefinition`s ahead of the traversal. 😣

josh:
None of the changes here are intended to change the APIs of the class itself so it should be backwards compatible. But putting it into the next minor release is fine also.

rmosolgo:
👏 !!

PR #708, opened 2017-05-02
fix(Relay::RangeAdd) initialize connection with field

rmosolgo:
Oops, this field wasn't populated for the range_add helper's Connection object!


rmosolgo:
😖 Forgot the _test_, this doesn't actually work. (Or, it does _work_, but it's not useful, because in a case of a mutation, who knows where the actual connection field will be.)

Issue #524, opened 2017-02-06
Topical documentation


rmosolgo:
I think a few "recipes" could help folks work through common problems with GraphQL + Ruby. A few that come to mind are: 

- File uploads with Rails + a GraphQL client (Apollo?)
- CSRF tokens with GraphQL clients (maybe this doesn't apply with Relay?)
- Activating & deactivating parts of the schema with Flipper 
- ... others?

These could be added to the gem's documentation or written elsewhere and linked from "Related Projects."

hd719:
Hello I am new to open source and I would like to contribute to this project. 

-Thank you!

github0013:
I could manage file-uploading using graphql-ruby + apollo-client + apollo-upload-client
https://gist.github.com/github0013/d79fa651be3d7450adcd447676d01921

gist is a concept of how you do it.

PR #711, opened 2017-05-03
Improve documentation

emilebosch:
Hi!

I'd love to make the documentation on the website better. I think that all information exists in one form or the other on the website but seems hard to find in certain areas.

I'd propose to sketch out an outline of different topics and then start writing the actual content. Anyone caring to jump in with ideas? Or fill in where they had blanks when starting with graphql ruby?

In general:

- Documentation should highlight the 3 parts of a working graphql server,  1 the schema implementation (DSL/IDL), 2 the resolver implementation and the graphql query sent to the graphql server. I think every code sample would need to show this.
- Would be nice to have a sample app displaying some of the features of graphql-ruby. Is this app still relevant? Maybe to confusing since it also encompasses relay? https://github.com/rmosolgo/graphql-ruby-demo/
- Sometimes hard to see the difference between relay things and "pure" graphql things, for instance mutations vs relay mutations etc.

I would like to have a topic on mutations covering the following:

- What mutations are (link to graphql docs)
- How to do simple mutations with arguments 
- With Input types (i.e create a post /w activerecord)
- With with nested input types (i.e. create a todolist with items)
- How to deal with mutation errors, validations and some patterns on how to integrate with rails/ar.
- How to do relay mutations

Some other toptics that might need some love:

- How to implement unions
- How to implement scalars and use them
- How to use enums

On resolvers:

- How to implement resolvers with the DSL
- How to implement resolvers with a resolver map aka graphql-tools

On tie in with rails:

- How to use cancan/authorization
- Error handeling /w active record
- How to do/not do file uploads

Anyone more to add?


rmosolgo:
:+1: totally agree that the docs could use some love! Thanks for pointing out those areas for improvement. 

Sometime before 1.6.0 I'd like to redo the layout and structure of the website too, to make things easier to find (perhaps better table of contents or better titles, something).

rmosolgo:
Here are some of my thoughts on these documentation topics: 

- > sample app
  
  I would love to have this, but as you can see, I can't manage to maintain it :( 

- > difference between relay things and "pure" graphql things

  At first I tried to make this distinction, but I've almost given up. Many people use Relay concepts even for non-Relay clients, since they're generally very robust. What's more, Apollo is _trying_ to make Relay concepts seem like "pure" GraphQL, for example, ["GraphQL Connections"](https://dev-blog.apollodata.com/explaining-graphql-connections-c48b7c3d6976): 
  
  > connections are not specific to Relay at all
 
  😕 

  So, I don't know how to navigate it anymore: on the one hand, I'm sad when people avoid GraphQL because they think it's "for Relay and all that stuff", but on the other hand, I'm swimming against the current to decouple them. 

- > 3 parts of a working graphql server,
  
  I only found two numbered parts, what's the third one there?

- > other topics that might need some love

  Thanks again for identifying those, I'll take a look as part of #717

emilebosch:
> I would love to have this, but as you can see, I can't manage to maintain it :(

If you tell me what needs to be done, i can take a shot at it.

> connections are not specific to Relay at all 😕

Hmm. Okay, seems like a case of the "nested mutations". I think we should stick to whatever thing holds true. Ie. "Implementing Relay style cursors" maybe? For me it was super confusing and intimidating because Relay is super abstract, and graphql is as about concrete as you can get. And the naming such as edges, nodes etc might be a bit intimidating. Keeping it pure and then adding relay later sorta helped me. 

> I only found two numbered parts, what's the third one there? 

1. IDL schema/DSL schema 
2. Resolver implementation
3. The query to be send to graphql via graphiql for instance

rmosolgo:
One way I'm trying to improve documentation is by including the API docs in the search results. That should improve coverage of unions, enums and scalars, for example: 

image

image

rmosolgo:
I'd like to get this into #717 and release it soon, so I'm merging this PR. I'd love to keep an eye on the quality of docs for the issues you listed here, do want to add the outstanding topics to #524 ?

PR #712, opened 2017-05-04
BuildFromDefinition: add parser option

swalkinshaw:
This adds a callable `parser` option to `from_definition`.

The main use case for changing the default `GraphQL.parse` is if you're using `graphql-libgraphqlparser-ruby`. That gem overrides `GraphQL.parse` which does not support parsing an IDL string.

The workaround for this would be to pass in a callable that uses `GraphQL.parse_without_libgraphqlparser`.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -399,9 +399,10 @@ module GraphQL
     # Create schema from an IDL schema.
     # @param definition_string [String] A schema definition string
     # @param default_resolve [<#call(type, field, obj, args, ctx)>] A callable for handling field resolution
+    # @param parser [Object] An object for handling definition string parsing (must respond to `parse`)
     # @return [GraphQL::Schema] the schema described by `document`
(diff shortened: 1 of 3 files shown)


rmosolgo:
:+1: I love the approach! Here's one thought: what about swapping `parser.call(string)` for `parser.parse(string)`. The advantage there is that both `GraphQL::Language::Parser` and `GraphQL::Libgraphqlparser` work out of the box: 

```ruby
GraphQL::Schema.build_from_definition(str, parser: GraphQL::Language::Parser)
# or: 
GraphQL::Schema.build_from_definition(str, parser: GraphQL::Libgraphqlparser)
```

What do you think? Would that work for your case?

swalkinshaw:
Updated 👍 . I didn't love the callable either but forgot about going through the other namespaces vs just `GraphQL`.

rmosolgo:
Sweet!

PR #715, opened 2017-05-07
feat(Arguments#fetch) add args.fetch(key)

### LINKED ISSUE #453: Prevent typos in `args` by whitelisting defined arguments
Sometimes I'm working on a `resolve` and it doesn't work ... and I realize it's because I made a typo in my `args[key]` call 😖 

This error could be avoided altogether because the set of valid arguments is known ahead of time. We could "help" the developer avoid typos by raising an error when they try to access an undefined argument. 

It could be implemented at least two ways: 

- `Arguments#[](key)` could assert that `key` is a defined argument (possibly breaking)
- `Arguments` classes could be defined on a field-by-field basis with _real methods_ for each defined argument (not breaking, but possible naming conflicts with existing methods)

However, I'm not sure if this is worth the trouble: 

- Do other folks have this problem? 
- Do other folks _depend_ on the fact that undefined `args[key]` calls return `nil`?

rmosolgo:
Fixes #453

### DIFF
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -37,6 +37,13 @@ module GraphQL
         @unwrapped_values ||= unwrap_value(@original_values)
       end
 
+      # Returns the value for `key` or raises `KeyError` if `key` isn't present
+      # @param key [String, Symbol]
+      # @return [Object] the argument at `key`
+      def fetch(key)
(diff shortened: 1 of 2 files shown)


rmosolgo:
This wasn't quite right -- I want it to handle defined-but-not-provided arguments with nil, not keyerror

PR #716, opened 2017-05-08
Fix: List of Input Objects

theorygeek:
According to the [GraphQL Spec](https://facebook.github.io/graphql/#sec-Lists):
> If the value passed as an input to a list type is not a list and not the null value, it should be coerced as though the input was a list of size one, where the value passed is the only item in the list. This is to allow inputs that accept a “var args” to declare their input type as a list; if only one argument is passed (a common case), the client can just pass that value rather than constructing the list.

As best I can tell, this should apply to both lists of scalars or lists of input objects. One of the recent refactors modified `GraphQL::ListType` so that instead of using `ensure_array(value)`, it would use `Array(value)`.

The problem is that when you pass a hash into the `Array` method, you get the equivalent of calling `to_a` on the hash:
```ruby
Array({ "hello" => "world" })
 => [["hello", "world"]] 
```
I'm guessing that the change was so that the user could pass in an array-like object, instead of an actual array, as inputs to list types. So I took a cue from [`Array#wrap`](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/array/wrap.rb#L37) and wrote a custom `ensure_array` method that will attempt to call `to_ary` on provided objects. 

This may not be the best fix, but maybe it's a good starting point.

### DIFF
--- a/spec/graphql/list_type_spec.rb
+++ b/spec/graphql/list_type_spec.rb
@@ -8,6 +8,10 @@ describe GraphQL::ListType do
     assert_equal([1.0, 2.0, 3.0].inspect, float_list.coerce_isolated_input([1, 2, 3]).inspect)
   end
 
+  it "converts items that are not lists into lists" do
+    assert_equal([1.0].inspect, float_list.coerce_isolated_input(1.0).inspect)
+  end
+
   describe "validate_input with bad input" do
     let(:bad_num) { "bad_num" }
     let(:result) { float_list.validate_isolated_input([bad_num, 2.0, 3.0]) }
@@ -28,6 +32,22 @@ describe GraphQL::ListType do
       expected = GraphQL::FLOAT_TYPE.validate_isolated_input(bad_num).problems[0]["explanation"]
       actual = result.problems[0]["explanation"]
       assert_equal(actual, expected)
+    end
+  end
+
+  describe "list of input objects" do
+    let(:input_object) do
+      GraphQL::InputObjectType.define do
+        name "SomeInputObjectType"
+        argument :float, !types.Float
+      end
+    end
+
+    let(:input_object_list) { input_object.to_list_type }
+
+    it "converts hashes into lists of hashes" do
+      hash = { 'float' => 1.0 }
+      assert_equal([hash].inspect, input_object_list.coerce_isolated_input(hash).map(&:to_h).inspect)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/graphql/list_type_spec.rb
```diff
@@ -8,6 +8,10 @@
     assert_equal([1.0, 2.0, 3.0].inspect, float_list.coerce_isolated_input([1, 2, 3]).inspect)
   end
 
+  it "converts items that are not lists into lists" do
+    assert_equal([1.0].inspect, float_list.coerce_isolated_input(1.0).inspect)
```
rmosolgo: Why the use of `.inspect` here (and below)? Did the test fail without it? 

If it works without it, I'd rather remove it, that way we can be sure we're testing the return value of this method, not the return value's `#inspect` method.

theorygeek: I imagine it works without it; I was just following the pattern of the other tests in this file, which also compare the results of `inspect`. I'll switch remove those calls, though.

theorygeek:
>I'm happy to include the to_ary check but I'm curious, did you add it because of a certain case that you found? Or is it "just in case"?

It was just in case; I thought there might have been a behavioral reason behind switching from `ensure_array` to `Array` on the previous refactor, and I thought this might be a good way to preserve that.

But, I'm just as happy to go back to the original version:
```ruby
def ensure_array(value)
  value.is_a?(Array) ? value : [value]
end
```

spec/graphql/list_type_spec.rb
rmosolgo: Oops, I didn't check the rest of the file! I suppose that's a fine reason to leave it.

rmosolgo:
> I thought there might have been a reason

That's very generous 😬 I think I just learned about `Array(...)` in the meantime and thought I'd apply it here. Apparently I didn't learn _enough_ about `Array(...)`! It's ok with me to drop `to_ary` support until we need it for something.

rmosolgo:
I'm about to release 1.5.11 so I went ahead with the update to `ensure_array`, I hope that's ok! 

Also, I realized the reason for `.inspect` in those previous tests: it was to make sure that floats were coerced to ints, where `==` wasn't a strict enough check.

theorygeek:
Ah, ok gotcha, and yeah, I'm very OK with removing the `to_ary` check!

PR #718, opened 2017-05-08
Camelize type name

mtsmfm:
Thank you for this great gem!

I think it is better to care lower case same as Rails' generator.

For example, `bin/rails g model foo` will generate:

```
class Foo < ApplicationRecord
end
```


rmosolgo:
That's much more user-friendly, thank you!

PR #722, opened 2017-05-11
fix(Relay::Mutation) handle errors raised in resolve function

rmosolgo:
There was no test for this and it changed 1.5.7 -> 1.5.8


rmosolgo:
This adds a strange disparity: 

- Errors during eager resolution turn the result into `nil`
- Errors during lazy resolution turn the result into hash where all keys are `nil`

The new behavior added here is the _correct_ behavior. I'll update them to match in 1.6.0 but it will be a breaking change.

PR #724, opened 2017-05-11
Permit valueless definitions

gjtorikian:
I'd like to not necessarily require values when adding my own metadata definitions.

For example, the comment in this file shows

``` ruby
# Use it in a definition
subaru_baja = Car.define do
 # ...
 all_wheel_drive true
end
```

Wouldn't it be much nicer like this? 😀 

``` ruby
# Use it in a definition
subaru_baja = Car.define do
 # ...
 has_all_wheel_drive
end
```

### DIFF
--- a/lib/graphql/define/instance_definable.rb
+++ b/lib/graphql/define/instance_definable.rb
@@ -279,7 +279,7 @@ module GraphQL
           @key = key
         end
 
-        def call(defn, value)
+        def call(defn, value = true)
           defn.metadata[@key] = value
         end
       end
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: would merge with a test, feel free to update the existing one according to your example!

gjtorikian:
🎉

gjtorikian:
Bump for eyes~ 😅

rmosolgo:
Thanks for the bump, somehow I missed that followup commit!

PR #728, opened 2017-05-18
Fix ArrayConnection and RelationConnection issues

arthurschreiber:
Fix the ArrayConnection to work as described in the Relay Connection spec.

* Both `before` and `after` can be specified now.
* Both `first` and `last` can be specified now. Note that doing so is discouraged and the result is confusing.

This also fixes an issue where `max_page_size` in combination with `before` and no `last` would return results from the end of the list instead of from the beginning.

### DIFF
--- a/spec/graphql/relay/array_connection_spec.rb
+++ b/spec/graphql/relay/array_connection_spec.rb
@@ -89,6 +89,27 @@ describe GraphQL::Relay::ArrayConnection do
       assert_equal(false, result["data"]["rebels"]["ships"]["pageInfo"]["hasPreviousPage"])
     end
 
+    it 'works with before and after specified together' do
+      result = star_wars_query(query_string, "first" => 1)
+      assert_equal(["X-Wing"], get_names(result))
+
+      first_cursor = get_last_cursor(result)
+
+      # There is no records between before and after if they point to the same cursor
+      result = star_wars_query(query_string, "before" => first_cursor, "after" => first_cursor, "last" => 2)
+      assert_equal([], get_names(result))
+
+      result = star_wars_query(query_string, "after" => first_cursor, "first" => 2)
+      assert_equal(["Y-Wing", "A-Wing"], get_names(result))
+
+      # After the last result, find the next 2:
+      second_cursor = get_last_cursor(result)
+
+      # There is only 2 results between the cursors
+      result = star_wars_query(query_string, "after" => first_cursor, "before" => second_cursor, "first" => 5)
+      assert_equal(["Y-Wing", "A-Wing"], get_names(result))
+    end
+
     it 'handles cursors beyond the bounds of the array' do
       overreaching_cursor = Base64.strict_encode64("100")
       result = star_wars_query(query_string, "after" => overreaching_cursor, "first" => 2)
@@ -157,13 +178,13 @@ describe GraphQL::Relay::ArrayConnection do
 
       it "applies to queries by `last`" do
         last_cursor = "Ng=="
-        second_to_last_two_names = ["Death Star", "Shield Generator"]
+
         result = star_wars_query(query_string, "last" => 100, "before" => last_cursor)
-        assert_equal(second_to_last_two_names, get_names(result))
+        assert_equal(["Death Star", "Shield Generator"], get_names(result))
         assert_equal(true, get_page_info(result)["hasPreviousPage"])
 
         result = star_wars_query(query_string, "before" => last_cursor)
-        assert_equal(second_to_last_two_names, get_names(result))
+        assert_equal(["Yavin", "Echo Base"], get_names(result))
         assert_equal(false, get_page_info(result)["hasPreviousPage"], "hasPreviousPage is false when last is not specified")
 
         third_cursor = "Mw=="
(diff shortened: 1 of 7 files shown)


arthurschreiber:
I thought it might also be helpful to link to the specification I'm talking about. 😁

This changes `ArrayConnection` to conform to the slicing/pagination algorithm described here: https://facebook.github.io/relay/graphql/connections.htm#sec-Pagination-algorithm

rmosolgo:
Thanks for this patch! I'll take a close look soon but the changes you describe sound good to me. However, the tests for Rails 3.2 fail 😖

arthurschreiber:
@rmosolgo That is really weird! I'll take a look into why they are failing. 😞

arthurschreiber:
@rmosolgo After tinkering around a bit, I __finally__ have a version of this that works and is "more correct" than before. Would you mind taking another look?

I'm not 100% happy about the code here, but working with the offsets and limits is just really messsy and I'm not sure how to make that part nicer. 🤔

spec/graphql/relay/relation_connection_spec.rb
```diff
@@ -400,4 +420,148 @@ def get_last_cursor(result)
...
+
+        connection = GraphQL::Relay::RelationConnection.new(StarWars::Base.offset(2).limit(3), { first: 2, last: 5 })
+        assert_equal [StarWars::Base.find(3), StarWars::Base.find(4)], connection.edge_nodes
+
+        connection = GraphQL::Relay::RelationConnection.new(StarWars::Base.offset(2).limit(3), { first: 1, last: 5 })
+        assert_equal [StarWars::Base.find(3)], connection.edge_nodes
+
+      end
+    end
+  end
```
rmosolgo: :clap: !

spec/graphql/relay/array_connection_spec.rb
```diff
@@ -157,13 +178,13 @@ def get_page_info(result)
...
-        second_to_last_two_names = ["Death Star", "Shield Generator"]
+
         result = star_wars_query(query_string, "last" => 100, "before" => last_cursor)
-        assert_equal(second_to_last_two_names, get_names(result))
+        assert_equal(["Death Star", "Shield Generator"], get_names(result))
         assert_equal(true, get_page_info(result)["hasPreviousPage"])
 
         result = star_wars_query(query_string, "before" => last_cursor)
-        assert_equal(second_to_last_two_names, get_names(result))
+        assert_equal(["Yavin", "Echo Base"], get_names(result))
```
rmosolgo: Hmm, can you help me understand what's happening here? I _think_ the behavior of `"before"` without `"last"/"first"` is changing: 

- Previously, we implied an implicit `last: max_page_size`
- Now, we apply an implicit `first: max_page_size`

Is that correct? 

(Re-reading the spec, this may be correct, but I want to make sure to think through the compatibility issues in rolling this out.)

spec/graphql/relay/page_info_spec.rb
```diff
@@ -75,8 +75,8 @@ def get_last_cursor(result)
       result = star_wars_query(query_string, "last" => 1, "first" => 1, "before" => cursor_of_last_base)
       assert_equal(true, get_page_info(result)["hasNextPage"])
       assert_equal(true, get_page_info(result)["hasPreviousPage"])
-      assert_equal("Mg==", get_page_info(result)["startCursor"])
-      assert_equal("Mg==", get_page_info(result)["endCursor"])
+      assert_equal("MQ==", get_page_info(result)["startCursor"])
+      assert_equal("MQ==", get_page_info(result)["endCursor"])
```
rmosolgo: It looks like this cursor value changed by 1, is that correct? Can you help me understand the _meaning_ of this change? (What did it _used_ to point to, what does it point to now?)

spec/graphql/relay/array_connection_spec.rb
arthurschreiber: Yes, you're right.

Previously, using `before` implied `last: max_page_size` automatically (unless you specified `first`). This is not according to the relay connection spec, which specifies that `before` and `after` are not tied to `first` and `last` in any way.

This might be a breaking change to users, but I'm not sure. 🤔  The previous behaviour was confusing and wrong, so I hope people don't rely on it.

spec/graphql/relay/page_info_spec.rb
arthurschreiber: If I remember correctly the `star_wars_query` here without any of `first`/`last`/`before`/`after` will return 3 bases: `Death Star`, `Shield Generator` and `Headquarters` (in this order).

The `cursor_of_last_base` points to `Headquarters`, so without `first`/`last` the result would contain `Death Star` and `Shield Generator`.

According to the spec, `first`/`last` have to be applied one after the other. In this case here, applying `first: 1` first reduces the result set to `Death Star` only (cursor value: `MQ==`), and applying `last: 1` does not change the result.

The previous code would return ``

Previously, the cursor for `Shield Generator` was returned, which did not match the spec.

rmosolgo: :+1: Confirmed the changed offset in the test matches the bases you named! 

image

rmosolgo:
Thanks again for this fix, and helping me understand it! I've pointed this change at 1.6.x to avoid surprising people on a patch-version update.

PR #738, opened 2017-05-26
Enum types documentation fix

Tonkpils:
LanguageType is not mentioned should be LanguageEnum


rmosolgo:
Oops, thanks for this fix!

PR #741, opened 2017-05-27
Fix some deprecations

rmosolgo:


rmosolgo:
One of these was covered by #743, I'll get the other one separately.

PR #742, opened 2017-05-27
fix(UniqueWithinType): Allow separator to be used in the ID

ThisIsMissEm:
BREAKING CHANGE:

Previously you could not use the separator character in the ID (object_value) argument, which meant, by default you could not use this with UUIDs. By limiting the split in decoding to returning two values, the object_value can now include the separator.

This change also now guarantees that you won't receive back more than two values from decode, which prevents accidental behaviour (for instance, it could be possible to get the server to decode an ID returning a huge array of strings, consuming more memory than values ever used).

This is a breaking change, but should be entirely backwards compatible, and work with existing documented usage.

This is as discussed on Twitter: https://twitter.com/thisismissem/status/868268773031251969


rmosolgo:
🤘 This looks great, what a win!

rmosolgo:
Thanks again, just released in 1.5.14!

PR #743, opened 2017-05-27
fix(spec): Update usage of Sequel in test schema

ThisIsMissEm:
This change just follows what the error message says, upgrading to the newer syntax.

I encountered this error when working on #742, and it was a simple straight-forward fix.


rmosolgo:
Sweet, thanks!

PR #747, opened 2017-05-29
Prevent name error when ActiveRecord not exists.

marocchino:
The error was NameError: uninitialized constant ActiveRecord.


rmosolgo:
Oops, sorry about that! Thanks for pushing this fix!

PR #749, opened 2017-05-30
Avoid ActionController::InvalidAuthenticityToken

Tonkpils:
Setting `protect_from_forgery` to avoid authenticity token exceptions on a vanilla GraphQL controller.

### DIFF
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -1,4 +1,6 @@
 class GraphqlController < ApplicationController
+  protect_from_forgery
+                                               
   def execute
     variables = ensure_hash(params[:variables])
     query = params[:query]


rmosolgo:
Hi, can you help me understand this patch? 

How were you sending queries to the server that gave you `InvalidAuthenticityToken`? Did this patch fix it in your local controller?

Tonkpils:
Yeah, I was setting up an initial graphql relay modern. First request I made with relay returned 422 due to that exception. This fixed it as API should use protect from forgery with null session

rmosolgo:
Is null session the right behavior, though? I mean, that will cause logged-in users to appear as logged-out, right? 

Here are a couple of other options, what do you think of them? 

- Provide some documentation about how to use Rails CSRF tokens with popular GraphQL client libraries 
- Update the generator to include your fix here, but make it commented out, and include a comment describing why someone might want to use it

Tonkpils:
> Provide some documentation about how to use Rails CSRF tokens with popular GraphQL client libraries

I feel this is the better approach. In order to not remove the session from the request, a better option would be to define `csrf_meta_tags` in the rendered view and have the client's network layer inject that csrf token in every request as a header. Where would be the best place to put this documentation?

rmosolgo:
I think a fresh markdown document inside `guides` would be a good spot for it. 

Here are some options for supporting CSRF for Apollo: https://stackoverflow.com/questions/42723989/invalid-auth-token-with-rails-graphql-apollo-client

Did you find something that worked with Relay + Rails?

Tonkpils:
> Did you find something that worked with Relay + Rails?

I haven't test it but I believe that same concept should work. It's just extracting the csrf token from a meta tag and applying it as a header. Seems like those answers are very similar to what I was going to attempt to do in Relay. 

I'll test it out and write up the docs for it!

rmosolgo:
> extracting the csrf token from a meta tag 

:+1: :+1: :+1: !

rmosolgo:
Glad we found something that works here, I hope we can write client-specific docs soon :)

Tonkpils:
> I hope we can write client-specific docs soon :)

Definitely, I'll get to it soon. I've been a bit swamped lately

PR #750, opened 2017-05-30
fix(Argument#prepare) guides to reflect 1.6.0 changes

lucaspolonio:
As per https://github.com/rmosolgo/graphql-ruby/pull/730, Argument prepare functions receive the context as a second argument. This PR updates the guides to reflect this change.


rmosolgo:
My update flow involves fixing the tests ... I have to remember to find-and-replace next time too 😬  Thanks for the fix, just published it!

Issue #751, opened 2017-05-31
Implement a default max_page_size


esjee:
(As mentioned over Slack)

`max_page_size` is something that can be used to control how many objects people are allowed to request in one go: http://graphql-ruby.org/relay/connections.html#maximum-page-size

Implementing a "global" `max_page_size` that can be used as a default value seems helpful!

I hope to implement this.

PR #752, opened 2017-05-31
Implement "default_max_page_size"

esjee:
https://github.com/rmosolgo/graphql-ruby/issues/751

Usage:

```ruby
  # ...
  default_max_page_size 100
  # ...
end
```

For any relation where `max_page_size` is not defined, or if it is `nil`, the `default_max_page_size` will be used instead.

Also, partially for my own benefit, here's the complete list of bases in order:

| Name | Cursor |
| -----|-------|
| Yavin      | MQ==       |
| Echo Base | Mg== |
| Secret Hideout | Mw== |
| Death Star | NA== |
| Shield Generator | NQ== |
| Headquarters | Ng== |

(Yavin being the first base, Headquarters being the last).

### DIFF
--- a/spec/graphql/relay/array_connection_spec.rb
+++ b/spec/graphql/relay/array_connection_spec.rb
@@ -187,12 +187,12 @@ describe GraphQL::Relay::ArrayConnection do
         assert_equal(["Yavin", "Echo Base"], get_names(result))
         assert_equal(false, get_page_info(result)["hasPreviousPage"], "hasPreviousPage is false when last is not specified")
 
-        fourth_cursor = "Mw=="
+        third_cursor = "Mw=="
         first_and_second_names = ["Yavin", "Echo Base"]
-        result = star_wars_query(query_string, "last" => 100, "before" => fourth_cursor)
+        result = star_wars_query(query_string, "last" => 100, "before" => third_cursor)
         assert_equal(first_and_second_names, get_names(result))
 
-        result = star_wars_query(query_string, "before" => fourth_cursor)
+        result = star_wars_query(query_string, "before" => third_cursor)
         assert_equal(first_and_second_names, get_names(result))
       end
     end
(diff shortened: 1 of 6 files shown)


Willianvdv:
@esjee would it be valuable to add it to the security section of the readme?

esjee:
I think that makes sense! Would it also be helpful to have a sane default value here instead of `nil`?

Willianvdv:
@esjee that will be a breaking change plus I'm not sure what a sane value would be.

rmosolgo:
:+1: for keeping a default of `nil`, I'd rather keep the existing behavior as default, at least so we don't surprise anyone! 

BTW, regarding cursors, I often debug with Chrome console, in case this trick helps: 

image

spec/graphql/relay/array_connection_spec.rb
```diff
@@ -187,14 +187,74 @@ def get_page_info(result)
         assert_equal(["Yavin", "Echo Base"], get_names(result))
         assert_equal(false, get_page_info(result)["hasPreviousPage"], "hasPreviousPage is false when last is not specified")
 
-        third_cursor = "Mw=="
+        fourth_cursor = "Mw=="
```
rmosolgo: On the table you posted in the description, `"Mw=="` is the third row. Are you sure we should rename this?

spec/graphql/relay/relation_connection_spec.rb
```diff
@@ -206,23 +206,82 @@ def get_last_cursor(result)
 
       it "applies to queries by `last`" do
         second_to_last_two_names = ["Death Star", "Shield Generator"]
-        first_and_second_names = ["Yavin", "Echo Base"]
+        first_second_and_third_names = ["Yavin", "Echo Base"]
```
rmosolgo: There are only two items in this array, why should it be renamed to have `and_third` in the variable name?

spec/graphql/relay/array_connection_spec.rb
esjee: Nope! This is a rename gone badly. I was a bit too eager with the find-and-replace. Sorry!

spec/graphql/relay/relation_connection_spec.rb
esjee: Same as above, this is a find-and-replace error. Sorry!

rmosolgo:
Makes sense, thanks for following up!

PR #753, opened 2017-05-31
Update instrumentation.md

dazoakley:
Fix a typo in the code sample


rmosolgo:
Oopsie! Thanks for pointing this out. The preceeding sample uses`QueryTimerInstrumentation`, should we also use that name here?

dazoakley:
Ah d'oh! Good spot, yep, we should have the same name in both. :)

dazoakley:
And thank you for the awesome gem BTW - it's very useful!

dazoakley:
All updated :)

rmosolgo:
Thanks, just published the changes!

PR #758, opened 2017-06-02
Allow rescue_from and remove_handler to take multiple error classes

### LINKED ISSUE #757: rescue_from method does not accept arrays
Right now the rescue middleware only handles a single error class at a time. Sometimes it might be useful to accept an array of error classes that will all be handled in the same way. It's something I've run into a couple times. Is there any interest in a pull request adding that capability?

littlelazer:
Resolves #757


littlelazer:
@rmosolgo I figured I would change the remove_handler method as well, just to keep them in sync, and I went with map in order to somewhat preserve what the return value for each of those methods was previously. Let me know if that's ok

rmosolgo:
This looks great! Thanks for the great tests and updating the docs, too :) I'll probably be releasing a new version this afternoon, just a couple more things to check ...

PR #763, opened 2017-06-05
BackwardsCompatibility arity check: support Method

swalkinshaw:
* Adds support fo `Method` to `get_arity`.
* Updates MiddlewaChain to use `BackwardsCompatibility.get_arity`

### DIFF
--- a/lib/graphql/backwards_compatibility.rb
+++ b/lib/graphql/backwards_compatibility.rb
@@ -25,7 +25,7 @@ module GraphQL
 
     def get_arity(callable)
       case callable
-      when Proc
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Was there any tests I missed for this?

rmosolgo:
Thanks! 

Sadly, no tests. Usually my flow for deprecation is: 

- Add the compatibility code & run the old tests 
- Update the tests 

So I end up testing at the first step, but there aren't dedicated tests for this module (😬 ).  For example, there are some old-style middlewares still in the tests, so at least _they_ are calling through it!

PR #765, opened 2017-06-06
fix(Multiplex) correctly check for any overridden execution strategy

rmosolgo:
Oops, if you overrode _one_ strategy but not _all_ of them, you'd get `undefined method "new" for NilClass`

cc @swalkinshaw


swalkinshaw:
Tested this and it fixed the bug we had at least 👍

rmosolgo:
Thanks for giving it a try! Sorry about that 😅

Issue #769, opened 2017-06-07
`rails g graphql:mutation` doesn't add all boilerplate for mutations


ThisIsMissEm:
Say you do:

```
rails g graphql:mutation createAccount
```

Then you have a few bits of extra work to do, from the result:

1) You have to fix the module name in `app/graphql/mutations/createAccount.rb` from `Mutations::createAccount` to `Mutations::CreateAccount`
2) You have to add the mutation root, which requires creating:

```
Types::MutationType = GraphQL::ObjectType.define do
  name "Mutation"

  field :createAccount, Mutations::CreateAccount.field
end
```

3) You have to add `mutation(Types::MutationType)` to the `Schema` file in `app/graphql/`

Whilst adding the `field :createAccount` in isn't strictly necessary, it is nice to do, considering that the `GraphQL::Relay::Mutation.define` returns an object that isn't like `GraphQL::ObjectType.define`

I'm not sure about the camelCasing issue in 1) though, I suspect that's just a bug and easily fixed.

Would be happy to do a pull request if these are changes the community would find helpful.

rmosolgo:
:+1: for all the changes you recommend here!

esjee:
Was all of this fixed in https://github.com/rmosolgo/graphql-ruby/pull/821?

rmosolgo:
Yes, thanks for following up!

PR #772, opened 2017-06-07
Feature/better api only support

ThisIsMissEm:
This resolves #768, see there for the discussion.

PR #773, opened 2017-06-08
WIP: Improved mutation generator

ThisIsMissEm:
Currently a work in progress for #769 

There is a pretty major change here, in that I'm no longer using NamedBase for the mutation generator. Considering that if you're to do: `rails g graphql:mutation createUser` you get the wrong output.

Now if you do:

```
rails g graphql:mutation createUser
rails g graphql:mutation CreateUser
rails g graphql:mutation Create_User
rails g graphql:mutation create_user
```

All will now result in the exact same mutation being created. The field added to the root will be the lower-camelcase of the NAME, the module will be upper-camelcase of the NAME, and the file name will be lower-underscore-case of the NAME. This makes the output more reliable a predictable.

p.s., don't look at the code to NamedBase, it's confusing because it's so messy. A lot of the rails generator code seem to be just randomly dumped all over the place with no actual design.

I'm still not happy with this 100%, especially my "core" module, but I did need somewhere to start to put common functions used between each generator.

When you initially do `rails g graphql:install`, if you choose not to setup mutations, then when you do the first `rails g graphql:mutation NAME`, it'll automatically setup the mutation root type for you, add it to the schema, and add all the correct structure.

Additionally, it'll automatically add the field to the mutation root type, so following the examples above, you'd get this:

```rb
# MutationType.rb
Types::MutationType = GraphQL::ObjectType.define do
  name "Mutation"

  field :createUser, Mutations::CreateUser.field
end
```

If you choose to create mutations on install, then this file will initially look like:

```rb
# MutationType.rb
Types::MutationType = GraphQL::ObjectType.define do
  name "Mutation"

  # TODO: Add Mutations as fields
end
```

The code for `rails g graphql:mutation` is smart enough to remove that line.

### DIFF
--- a/lib/generators/graphql/install_generator.rb
+++ b/lib/generators/graphql/install_generator.rb
@@ -105,9 +105,22 @@ RUBY
             route(GRAPHIQL_ROUTE)
           end
         end
+
+        if gemfile_modified?
+          say "Gemfile has been modified, make sure you `bundle install`"
+        end
       end
 
       private
+
+      def gemfile_modified?
+        @gemfile_modified
+      end
+
+      def gem(*args)
+        @gemfile_modified = true
+        super(*args)
+      end
 
       def create_dir(dir)
         empty_directory(dir)
--- a/lib/generators/graphql/install_generator.rb
+++ b/lib/generators/graphql/install_generator.rb
@@ -98,7 +98,12 @@ RUBY
           say("  You may wish to use GraphiQL.app for development: https://github.com/skevy/graphiql-app")
         elsif !options[:skip_graphiql]
           gem("graphiql-rails", group: :development)
-          route(GRAPHIQL_ROUTE)
+
+          # This is a little cheat just to get cleaner shell output:
+          log :route, 'graphiql-rails'
+          shell.mute do
+            route(GRAPHIQL_ROUTE)
+          end
         end
       end
 
--- a/lib/generators/graphql/install_generator.rb
+++ b/lib/generators/graphql/install_generator.rb
@@ -68,6 +68,11 @@ module Graphql
         default: false,
         desc: "Include GraphQL::Batch installation"
 
+      # These two options are taken from Rails' own generators'
+      class_option :api,
+        type: :boolean,
+        desc: "Preconfigure smaller stack for API only apps"
+
 
       GRAPHIQL_ROUTE = <<-RUBY
 if Rails.env.development?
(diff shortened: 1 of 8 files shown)


ThisIsMissEm:
I'd also appreciate feedback on writing tests for this stuff, because I'm a bit frazzled figure that out.

ThisIsMissEm:
Oh, and yes, this does branch off #772 — seemed to make sense to start from that foundation.

rmosolgo:
These improvements look great to me! Crazy thing from CI: 

> No output has been received in the last 10m0s

Do the generator specs pass for you locally? I wonder if it's getting stuck in a slow process (I think some of the generator tests do a `rails new` 😬 ) or if there's a loop someplace ...

ThisIsMissEm:
Looks like there's a crash / bug. I'll circle back to this in a bit

ThisIsMissEm:
@rmosolgo I think there's a race condition or something happening. It's basically hanging on opening a specific file.

ThisIsMissEm:
Found the bug, I just have no idea how to fix it: essentially the generator depends on a certain file existing, and otherwise this fails.

rmosolgo:
Hmmm weird stuff ... can you tell what file it depends on?

ThisIsMissEm:
One of the tests is that it correctly updates a file that another generator creates.

lib/generators/graphql/install_generator.rb
```diff
@@ -76,42 +84,52 @@ class InstallGenerator < Rails::Generators::Base
 RUBY
 
       def create_folder_structure
-        create_dir("app/graphql/mutations")
         create_dir("app/graphql/types")
+        template("schema.erb", schema_file_path)
+
+        # Note: Yuo can't have a schema without the query type, otherwise introspection breaks
```
Willianvdv: s/Yuo/You

Willianvdv:
@ThisIsMissEm the install_generator_spec does a Rails setup (https://github.com/rmosolgo/graphql-ruby/blob/a15142a0c176df7ef9da9dc216eba87db7428527/spec/generators/graphql/install_generator_spec.rb#L10-L16) I think the mutation generator spec passes when the installation_generator_spec does the Rails setup. 

I think the solution would be to also do the Rails setup in the mutation spec.

rmosolgo:
I replicated the issue locally, this hangs:

```
be rake test TEST=spec/generators/graphql/install_generator_spec.rb
```

The setup runs, and the test case starts, but `run_generator([])` hangs :S

Willianvdv:
@rmosolgo it's your lucky day, I just fixed this issue in https://github.com/Unobvious-Technology/graphql-ruby/pull/1/files

rmosolgo:
Thanks for these improvements! I merged your commits as a part of #821.

PR #775, opened 2017-06-08
Stable Schema Printing

theorygeek:
Sorts fields, arguments, etc. by name when printing the IDL. (In my code base, these are getting flip-flopped around sometimes).

I think this is justified by the spec:
- [Arguments are unordered](http://facebook.github.io/graphql/#sec-Language.Arguments)
- [Input object fields are unordered](http://facebook.github.io/graphql/#sec-Input-Object-Values)

There isn't a statement in the spec for some of the other items, but it seems safe.

### DIFF
--- a/lib/graphql/schema/printer.rb
+++ b/lib/graphql/schema/printer.rb
@@ -186,7 +186,7 @@ module GraphQL
             end
 
             out = "(\n".dup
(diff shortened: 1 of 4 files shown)


rmosolgo:
This is awesome! People will have one noisy commit when they upgrade, but it will be a much smoother ride after that 😁

PR #777, opened 2017-06-12
Update docs for static validator

piotrze:


rmosolgo:
Thanks for taking care of this! Sorry the docs fell behind 😅 

Just curious, did you find a new use for the validator?

piotrze:
@rmosolgo I'm using it to validate queries stored in db, to keep them consistent with schema.

PR #782, opened 2017-06-13
[typo] Add missing colon in mutation documentation

graysonwright:


rmosolgo:
Good eye, thanks for the fix!

graysonwright:
Of course! Thanks for all of your work with the library, I'm really enjoying it! It's my first time using GraphQL and `graphql-ruby` makes it super easy to dive right in. 👍

PR #783, opened 2017-06-14
Refactor BaseConnection

ianks:
Recently, I have had to implement a few connections to paginate different API endpoints. Overall, I would say the experience was challenging because of the number of dependencies in BaseConnection. Setting up tests requires some internal knowledge of the library,  and rather painful to setup. 

By reducing some of the unnecessary dependencies in BaseConnection, I believe it will make it easier for users to implement their own connections by make implementation and testing more straightforward.

### Changes

- PR removes access to `context` to the BaseConnection. Overall, `context` contains a huge number of dependencies that are not required when implementing a connection. The two things `context` was used for was `max_page_size` and `coder`. I've made it so these arguments are instead passed explicitly. As a result, now I can use a custom coder without having to pass in a mock context and a mock schema. This makes testing, and thus implementing a connection much more straight forward.

- Renames cursor_encoder to cursor_coder. `cursor_encoder` is actually responsible for encoding *and* decoding cursors. Thus, it is semantically more precise and descriptive of its responsibilities
to call it a `cursor_coder`. This will be a deprecation that will only affect users who have set a custom `cursor_encoder`


rmosolgo:
Hi, thanks for taking a look at this! 

I agree that the custom connection situation is a real mess. But the other consideration is _compatibility_: some folks may depend on the APIs which have been changed here, and we should think very carefully before breaking their code. Connections are hard enough as-is, I don't want to give people the nasty surprise that we've broken them :( 

Here's another suggestion: what if we created a _new_ connection API? Something that could replace the current one, while keeping the existing API in place. This part of the library is a bit ... dusty ... and I think we could do a lot better now, after some experience! Here are some things that we could target:

- Reduce the requirements for implementing a connection (as you mention here) 
- First-class API for versioned cursors & pagination algorithms (this is part of GraphQL::Pro but I would like to build it into the gem -- it was very hard to put together for that project!)
- Better isolation of `PageInfo`: in my custom connections, I find I have to mess with `PageInfo` options which is annoying. I wonder if we could abstract that away. 
- Anything that needs to change to support IDL-based schema definition? (not sure on this one)

ianks:
> some folks may depend on the APIs which have been changed here

I tried to design this so alll of their code should still work, but with deprecation warnings. Did I potentially miss something?

>  what if we created a new connection API? 

Yeah @rmosolgo I definitely think a new public API would be nice!

rmosolgo:
Sorry for the late reply!

I see that you're right, compatibility is preserved here. I didn't read carefully enough 😬 

Again, I agree that the current Ruby API is not user-friendly. But I don't want to keep investing in this implementation yet. I'd rather try a new approach and help people migrate rather than push folks along with this setup. 

I'll open an issue for a new approach soon, what I have in mind is to decouple two things: 

- Cursor-based pagination 
- Relay-style `Connection` types & fields 

This way, we could support cursor-based pagination without Relay-related overhead. Two folks suggested that to me last week after looking at Graph.cool's "simple" vs "relay" pagination approaches.

PR #789, opened 2017-06-17
[IDL] Support accepts_definitions functions in IDL 

rmosolgo:
In the DSL, the best way to customize parts of a GraphQL schema is with custom definitions. 

It would be nice to have a similar feature for the IDL, why not use the _same feature_? 

Shout out to @theorygeek for first suggesting the `parse -> apply -> remove` pipeline for extracting behavior from IDL descriptions. 

See to-dos and caveats below.

On this branch, these are equivalent: 

### Field 

DSL 

```ruby 
field :thing,  types.String, "A thing", property: :other_thing, authorize: :admin
```

IDL 

```graphql 
# A thing 
# @property :other_thing
# @authorize :admin 
thing: String
```

### Type 

DSL
 
```ruby 
Types::ThingType = GraphQL::ObjectType.define do 
  name "Thing" 
  description "A thing"
  access :viewer
  # ...
end 
```

IDL 

```graphql
# A thing 
# @access :viewer
type Thing {
  # ...
}
```

### Schema 

DSL 

```ruby 
  max_complexity 1000
  object_from_id Helpers::ObjectFromId
  use GraphQL::Batch
end 
```

IDL 

```graphql 
# @max_complexity 1000 
# @object_from_id Helpers::ObjectFromId
# @use GraphQL::Batch
schema {
  query Query 
}
```

## Caveats 

- Input is passed to `instance_eval` which is unsafe. This is ok for safe input, but it's a foot-gun if turn user input into a schema. I should lock down the parser a bit.
- TODO: lossless schema dump (ie, `.to_definition` _should_ include `@`-definitions)
- TODO: somehow give a context-aware backtrace if something fails.
- TODO: support all customizable objects (eg `GraphQL::Argument` for `prepare:` handlers)
- TODO: don't reuse `instrument` -- find a new name for this
- TODO: don't use `@`, since it's already used for directives?


ThisIsMissEm:
Supporting declarations through IDL would be very nice!

rmosolgo:
I really like this approach, but I'm going to put it on the backburner and try #871 first.

PR #795, opened 2017-06-19
Simplifies the test setup a tiny bit

Willianvdv:
Almost not PR worthy, but while debugging the issues on #773 I stumbled upon this minor improvement.

### DIFF
--- a/spec/generators/graphql/install_generator_spec.rb
+++ b/spec/generators/graphql/install_generator_spec.rb
@@ -3,20 +3,19 @@ require "spec_helper"
 require "generators/graphql/install_generator"
 
 class GraphQLGeneratorsInstallGeneratorTest < Rails::Generators::TestCase
-
   tests Graphql::Generators::InstallGenerator
   destination File.expand_path("../../../tmp/dummy", File.dirname(__FILE__))
 
   setup do
     prepare_destination
-    FileUtils.cd(File.expand_path("../../../tmp", File.dirname(__FILE__))) do
-      `rm -rf dummy`
+
+    FileUtils.cd(File.join(destination_root, '..')) do
       `rails new dummy --skip-active-record --skip-test-unit --skip-spring --skip-bundle`
     end
   end
 
   test "it generates a folder structure" do
-    run_generator([])
+    run_generator
 
     assert_file "app/graphql/types/.keep"
     assert_file "app/graphql/mutations/.keep"


spec/generators/graphql/install_generator_spec.rb
```diff
@@ -3,20 +3,19 @@
...
 
 class GraphQLGeneratorsInstallGeneratorTest < Rails::Generators::TestCase
-
   tests Graphql::Generators::InstallGenerator
   destination File.expand_path("../../../tmp/dummy", File.dirname(__FILE__))
 
   setup do
     prepare_destination
-    FileUtils.cd(File.expand_path("../../../tmp", File.dirname(__FILE__))) do
-      `rm -rf dummy`
```
Willianvdv: prepare_destination already removes the dummy app

rmosolgo:
👌  nice! just curious, did you find a good resource to learn generator tests, or have you been learning as you go? I've struggled to find a proper way to learn :S

Willianvdv:
My resource is this Gem 😆

spec/generators/graphql/install_generator_spec.rb
```diff
@@ -3,20 +3,19 @@
...
+
+    FileUtils.cd(File.join(destination_root, '..')) do
       `rails new dummy --skip-active-record --skip-test-unit --skip-spring --skip-bundle`
     end
   end
```
ThisIsMissEm: I think it'd be good to get this into base generator test or something. It's common to all our tests

ThisIsMissEm:
@rmosolgo @williamvdv I've been learning from the Rails, Devise and Omniauth generators.

spec/generators/graphql/install_generator_spec.rb
Willianvdv: yes!

rmosolgo: It's really slow to do this! 

If we have a lot of tests that need a Rails app, we should keep a dummy app in `spec/dummy`, this has worked well for react-rails anyways

Willianvdv: I thought you meant some spec helper that you can call when dealing with a spec that needs Rails power.

rmosolgo: :+1:

PR #797, opened 2017-06-19
Use prepare function to cap limit

timscott:
Cap limit using prepare function instead of in the resolve body.


rmosolgo:
Nice, thank you!!

rmosolgo:
Published :D http://graphql-ruby.org/fields/limits.html

PR #800, opened 2017-06-20
[WIP] Fix #774: "Cannot override max complexity"

### LINKED ISSUE #774: Cannot override max_complexity in Schema.execute anymore
I just upgraded from 1.4.x to 1.6 and it seems this code 

```
opts = {max_complexity: 300}
MySchema.execute(query, opts)
```

does not override max_complexity anymore as it previously did.

As far as I can tell, this is because we have this sequence:
```
def execute(query_str = nil, **kwargs) # kwargs is {max_complexity: 300}
...  # invokes
multiplex([kwargs])  # sends [{max_complexity: 300]
... # gets to
def multiplex(*args) # args is [[{max_complexity: 300}]]
... # invokes 
GraphQL::Execution::Multiplex.run_all(self, *args)
... # in there
def run_all(schema, query_options, *rest)
  queries = query_options.map { |opts| GraphQL::Query.new(schema, nil, opts) }
  run_queries(schema, queries, *rest)
end
... # query_options is [{max_complexity: 300}] but "rest" is [], Query gets created with these options, but run_queries expects a max_complexity which can't be set
def run_queries(schema, queries, context: {}, max_complexity: nil)
```

Maybe I just need to change the way I invoke `MySchema.execute` but I can't see a way to do this.

esjee:


Willianvdv:
@esjee this spec looks like a duplicate of https://github.com/rmosolgo/graphql-ruby/blob/df83ff81f11a133e17ac779302974038fdb60692/spec/graphql/analysis/max_query_complexity_spec.rb#L60 / https://github.com/rmosolgo/graphql-ruby/blob/df83ff81f11a133e17ac779302974038fdb60692/spec/graphql/analysis/max_query_complexity_spec.rb#L74

esjee:
Closed as a correct failing spec was added in #774

Issue #798, opened 2017-06-19
ExecutionError raised in argument's prepare function doesn't work with GraphQL::Batch


cmatheson:
When an object is loaded with a `GraphQL::Batch::Loader`, errors returned by a field argument for that object are not handled properly.  I've included a minimal test case for the following schema:

```
type Query {                 
  user(_id: ID!, async: Boolean): User    
}                            

type User {                  
  name(someArg: ID): String  
}
```
The full code sample (requires graphql-ruby and graphql-batch) is here (things to note are that the `someArg`'s prepare function just returns an error, and that the user resolve function will either directly return an object or return one via GraphQL::Batch::Loader depending on the async argument for that field)

```ruby
class DummyUser
  attr_reader :name

  def initialize(name)
    @name = name
  end
end

class UserLoader < GraphQL::Batch::Loader
  def initialize(_)
  end

  def perform(ids)
    ids.each do |id|
      fulfill(id, DummyUser.new("async loaded user"))
    end
  end
end

UserType = GraphQL::ObjectType.define do
  name "User"

  field :name, types.String do
    argument :someArg, types.ID,
      prepare: ->(s, _) { GraphQL::ExecutionError.new("uh oh") }
    resolve ->(user, args, _) {
      user.name
    }
  end
end

QueryType = GraphQL::ObjectType.define do
  name "Query"

  field :user, UserType do
    argument :_id, !types.ID
    argument :async, types.Boolean
    resolve ->(_, args, ctx) {
      # everything works as expected unless you use the +UserLoader+
      # (GraphQL::Batch) class
      args[:async] ?
        UserLoader.new(:asdf).load(args[:_id]) :
        DummyUser.new("test user")
    }
  end
end

TestSchema = GraphQL::Schema.define do
  query(QueryType)

  lazy_resolve(Promise, :sync)
  instrument(:query, GraphQL::Batch::Setup)
end
```
querying `{ user(_id: "5") { name } }` returns `{"data"=>{"u1"=>{"name"=>"test user"}}}` as expected.

querying `{ user(_id: "5") { name(someArg: "whatever") } }` returns `{"data"=>{}, "errors"=>[{"message"=>"uh oh", "locations"=>[{"line"=>1, "column"=>23}]}]}` (also expected).

[...]

PR #801, opened 2017-06-20
Check that default value isn't an execution error

esjee:


Willianvdv:
@esjee this fixes #798, right?

esjee:
I believe so, yes!

lib/graphql/query/literal_input.rb
```diff
@@ -77,6 +77,10 @@ def self.from_arguments(ast_arguments, argument_defns, variables)
             # `context` isn't present when pre-calculating defaults
             if context
               value = arg_defn.prepare(value, context)
+              if value.is_a?(GraphQL::ExecutionError)
```
Willianvdv: Nice bug 😄

rmosolgo:
:+1: Awesome, thanks for taking care of this!

PR #803, opened 2017-06-20
fix(variables) Fix error when InputObjectType receives array as provided value

xuorig:
Gem would crash when an InputObject variable was provided an array as a value:

```
gems/graphql-1.6.3/lib/graphql/input_object_type.rb:136[]	
gems/graphql-1.6.3/lib/graphql/input_object_type.rb:136block in validate_non_null_input	
gems/graphql-1.6.3/lib/graphql/input_object_type.rb:135each	
gems/graphql-1.6.3/lib/graphql/input_object_type.rb:135map	
gems/graphql-1.6.3/lib/graphql/input_object_type.rb:135validate_non_null_input
```

### Cause

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/input_object_type.rb#L108-L122

The Input Object class validates that the provided value responds to `to_h` or `to_unsafe_h` (Parameters). However, `Array` does respond to `to_h`, but `array["string"]` would cause the error to happen.

### Possible Fix

Instead of checking for `to_h`. This PR assigns the value of the provided value to `value.to_h`. I'm not 100% sure it's the right approach but it at least handles arrays and `ActionController::Parameters`.

Can anybody think of when this could fail?

### DIFF
--- a/lib/graphql/input_object_type.rb
+++ b/lib/graphql/input_object_type.rb
@@ -105,22 +105,35 @@ module GraphQL
       warden = ctx.warden
       result = GraphQL::Query::InputValidationResult.new
 
+      # Handle arrays specifically
+      # .to_h will handle key value arrays [["a", 1]]
+      # but will raise when array is not in the good format
+      if input.is_a?(Array)
+        begin
+          # Actual coerce array to hash so it does not crash
+          # with string keys.
+          input = input.to_h
+        rescue
+          result.add_problem(INVALID_OBJECT_MESSAGE % { object: JSON.generate(input, quirks_mode: true) })
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


xuorig:
broke some actioncontroller param tests 💃

xuorig:
ah:

```
> <ActionController::Parameters {"source"=>"COW", "fatContent"=>0.4} permitted: >.to_h
> {}
```

xuorig:
@rmosolgo modified the PR so it just bails out on arrays instead, much simpler and really we dont want to handle that 🤔

rmosolgo:
Nice find, thanks for taking care of this!

PR #804, opened 2017-06-21
Add spec showing you cannot increase max complexity at query time

riffraff:


rmosolgo:
This commit is included in (and _fixed by_) #812, thanks again for demonstrating this!!

PR #805, opened 2017-06-21
improve error handling in argument prepare procs

### LINKED ISSUE #798: ExecutionError raised in argument's prepare function doesn't work with GraphQL::Batch

cmatheson:
This change fixes #798, but also improves the error and data results from synchronous queries.  Specifically, errors that happen during argument preparation get path info now, and also less of the result is nil'ed out.

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -76,8 +76,8 @@ module GraphQL
             selection: selection,
           )
 
-          arguments = query.arguments_for(selection, field)
           raw_value = begin
+            arguments = query.arguments_for(selection, field)
             query_ctx.schema.middleware.invoke([parent_type, object, field, arguments, field_ctx])
           rescue GraphQL::ExecutionError => err
             err
@@ -116,7 +116,7 @@ module GraphQL
 
           case raw_value
           when GraphQL::ExecutionError
-            raw_value.ast_node = field_ctx.ast_node
+            raw_value.ast_node ||= field_ctx.ast_node
             raw_value.path = field_ctx.path
             query.context.errors.push(raw_value)
           when Array
--- a/spec/graphql/query/literal_input_spec.rb
+++ b/spec/graphql/query/literal_input_spec.rb
@@ -5,8 +5,8 @@ describe GraphQL::Query::LiteralInput do
   describe ".from_arguments" do
     describe "arguments are prepared" do
       let(:schema) {
-        query = GraphQL::ObjectType.define do
-          name "Query"
+        type = GraphQL::ObjectType.define do
+          name "SomeType"
 
           field :addToArgumentValue do
             type !types.Int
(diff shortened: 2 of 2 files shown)


lib/graphql/execution/execute.rb
```diff
@@ -76,8 +76,8 @@ def resolve_field(owner, selection, parent_type, field, object, query_ctx)
             selection: selection,
           )
 
-          arguments = query.arguments_for(selection, field)
```
cmatheson: exceptions raised here weren't being handled which is how the ExecutionError ends up in the query results as described in #798.

spec/graphql/query/literal_input_spec.rb
```diff
@@ -35,40 +35,56 @@
           end
         end
 
+        query = GraphQL::ObjectType.define do
+          name "Query"
+
+          field :top, type, resolve: ->(_, _, _) { true }
```
cmatheson: I moved the fields into this useless looking "top" field so that the path info / results could be shown better (before the whole query would be nil and the errors would have no path info).

lib/graphql/execution/execute.rb
Willianvdv: Nice find!

rmosolgo:
Awesome, thank you for getting to the bottom of this!

PR #808, opened 2017-06-22
Fix description for schema -> dynamic definition

peterp:


rmosolgo:
Thanks!

rmosolgo:
aaand published, thanks for the fix!

PR #810, opened 2017-06-23
Support .erb files in Schema.from_definition

rmosolgo:
If we start using `.graphql` files as the source of truth, it would be nice to have a way to reduce boilerplate code. 

Here's a crazy idea: use ERB as a preprocessor. It's familiar to Ruby devs and the processing model is fairly simple: the ERB has to output valid GraphQL code, just like an ERB template in Rails has to output valid HTML. 

We can include some built-in helpers (I've added a few, can you think of others?) as well as a way to take a module of user-provided methods to mix in at runtime (the `helpers:` option). 

Is this worth a try?

```ruby 
# Load a schema from definition:
MySchema = GraphQL::Schema.from_definition("schema.graphql.erb")
```

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -444,13 +444,19 @@ module GraphQL
       GraphQL::Schema::Loader.load(introspection_result)
     end
 
-    # Create schema from an IDL schema.
-    # @param definition_string [String] A schema definition string
+    # Create schema from an IDL schema or file containing an IDL definition.
+    # @param definition_or_path [String] A schema definition string, or a path to a file containing the definition
     # @param default_resolve [<#call(type, field, obj, args, ctx)>] A callable for handling field resolution
     # @param parser [Object] An object for handling definition string parsing (must respond to `parse`)
     # @return [GraphQL::Schema] the schema described by `document`
-    def self.from_definition(string, default_resolve: BuildFromDefinition::DefaultResolve, parser: BuildFromDefinition::DefaultParser)
-      GraphQL::Schema::BuildFromDefinition.from_definition(string, default_resolve: default_resolve, parser: parser)
+    def self.from_definition(definition_or_path, default_resolve: BuildFromDefinition::DefaultResolve, parser: BuildFromDefinition::DefaultParser)
+      # If the file ends in `.graphql`, treat it like a filepath
+      definition = if definition_or_path.end_with?(".graphql")
+        File.read(definition_or_path)
+      else
+        definition_or_path
+      end
+      GraphQL::Schema::BuildFromDefinition.from_definition(definition, default_resolve: default_resolve, parser: parser)
     end
 
     # Error that is raised when [#Schema#from_definition] is passed an invalid schema definition string.
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -449,6 +449,7 @@ module GraphQL
     # @param definition_or_path [String] A schema definition string, or a path to a file containing the definition
     # @param default_resolve [<#call(type, field, obj, args, ctx)>] A callable for handling field resolution
     # @param parser [Object] An object for handling definition string parsing (must respond to `parse`)
+    # @param helpers [Module] Extra helper methods for rendering ERB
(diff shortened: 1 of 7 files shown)


lib/graphql/schema/template.rb
```diff
@@ -0,0 +1,58 @@
...
+        @result = ""
+      end
+
+      # @return [String] Rendered ERB
+      def run
+        ERB.new(@erb_template, 0, "", "@result").result(binding)
+        @result
+      end
+
+      def connection(type_name)
```
theorygeek: maybe: `connection_type`

lib/graphql/schema/template.rb
```diff
@@ -0,0 +1,58 @@
...
+        conn = type("#{type_name}Connection", edges: "[#{type_name}Edge!]!", pageInfo: "PageInfo!")
+        edge = type("#{type_name}Edge", cursor: "ID!", node: "#{type_name}!")
+        conn + "\n" + edge
+      end
+
+      def connects(fields)
```
theorygeek: maybe: `connection_field`

lib/graphql/schema.rb
```diff
@@ -444,13 +445,22 @@ def self.from_introspection(introspection_result)
...
+      # If the file ends in `.graphql`, treat it like a filepath
+      definition = if definition_or_path.end_with?(".graphql")
+        File.read(definition_or_path)
+      elsif definition_or_path.end_with?(".erb")
+        erb_template = File.read(definition_or_path)
+        GraphQL::Schema::Template.run(erb_template, helpers: helpers)
+      else
+        definition_or_path
+      end
+      GraphQL::Schema::BuildFromDefinition.from_definition(definition, default_resolve: default_resolve, parser: parser)
```
theorygeek: Maybe this could support an _array_ of definitions/paths. That way you could split your schema up into several files and pass all of them in together:
```ruby
def self.from_definition(definition_or_path, helpers: nil, default_resolve: BuildFromDefinition::DefaultResolve, parser: BuildFromDefinition::DefaultParser)
  definition_or_path = [definition_or_path] unless definition_or_path.is_a?(Array)

  schemas = definition_or_path.map do |value|
    # If the value ends in `.graphql` or `.erb`, treat it like a filepath
    if value.end_with?('.graphql')
      File.read(value)
    elsif value.end_with?('.erb')
      erb_template = File.read(value)
      GraphQL::Schema::Template.run(erb_template, helpers: helpers)
    else
      value
    end
  end

  GraphQL::Schema::BuildFromDefinition.from_definition(schemas.join("\n"), default_resolve: default_resolve, parser: parser)
end
```

Example:
```ruby
GraphQL::Schema.from_definition(Dir["schema/**/*.graphql", "schema/**/*.graphql.erb"])
```

rmosolgo: :+1: I was also thinking to take a `glob`, so just 

```ruby
GraphQL::Schema.from_definition("schema/**/*.graphql*")
``` 

Let me take a try at that ...

rmosolgo: Oh, I think I want to handle glob/arrays separately. 

It's because I want to make sure errors are handled well: if there's a parse error or an invalid or duplicate definition, I want to have the _original_ filename in the error message, so the developer can properly track it down!

lib/graphql/schema/template.rb
rmosolgo: good call, those names are much more predictable. Brb!

rmosolgo:
Another funny thing is how to handle automatically-added types like `PageInfo`. 

If you use the DSL, it's added to your schema without any action on your part. 

What if you use the IDL?  We could:

- Require manual addition, eg `<%= page_info %>`
- Detect whether or not `connection_*` helpers were called, and if they were, add any missing types automatically.

I _think_ it would be nice to add them automatically, so for example: 

```graphql 
type Artist {
  <%= connection_field(paintings: "Painting") %> 
}
# We automatically add: 
# <%= connection_type("Painting") %>
# <%= page_info %>
```

rmosolgo:
I'm not sure this is really needed; I cherry-picked the first commit into #872, we can add `.erb` support later if we find we need it.

PR #811, opened 2017-06-23
Document usage of default_value for field argument

pengwynn:
I was looking for the DSL syntax to add a default value for a field argument and couldn't find it in the docs. This patch adds a quick usage example to the [field arguments docs](http://graphql-ruby.org/fields/arguments.html).


rmosolgo:
Thanks for fixing this, published!

PR #812, opened 2017-06-26
Fix 774: cannot increase max complexity at query time

esjee:
#774 

I cherry picked the failing test from #804 

Much to my surprise it appears the analyzer in the test was being run twice! If I understand things correctly, there's an analyzer being run for the query itself (analyzer no. 1) and an analyzer being run for the collection of all queries -- the multiplex -- (analyzer no. 2). It was this second analyzer that was causing the failure described in #774, as it could not be overridden by providing arguments to `execute`, and so would always default to the schema defined `max_complexity`. 

The analyzer no. 2, the one for the multiplex, gets added here: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/multiplex.rb#L165

So, yes. A lot of figuring out, very little actual code changes :smile: . The issue is fixed by passing `max_complexity` all the way down from `execute` to `run_queries`.

### DIFF
--- a/lib/graphql/execution/multiplex.rb
+++ b/lib/graphql/execution/multiplex.rb
@@ -35,17 +35,18 @@ module GraphQL
       end
 
       class << self
-        def run_all(schema, query_options, *args, **kwargs)
+        def run_all(schema, query_options, *args)
           queries = query_options.map { |opts| GraphQL::Query.new(schema, nil, opts) }
-          run_queries(schema, queries, *args, **kwargs)
+          run_queries(schema, queries, *args)
         end
 
         # @param schema [GraphQL::Schema]
         # @param queries [Array<GraphQL::Query>]
         # @param context [Hash]
-        # @param max_complexity [Integer]
+        # @param max_complexity [Integer, nil]
         # @return [Array<Hash>] One result per query
-        def run_queries(schema, queries, context: {}, max_complexity: nil)
+        def run_queries(schema, queries, context: {}, max_complexity: schema.max_complexity)
+
           if has_custom_strategy?(schema)
             if queries.length != 1
               raise ArgumentError, "Multiplexing doesn't support custom execution strategies, run one query at a time instead"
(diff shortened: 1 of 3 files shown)


lib/graphql/execution/multiplex.rb
```diff
@@ -161,7 +161,7 @@ def with_instrumentation(schema, queries, context:, max_complexity:)
           end
 
           multiplex_analyzers = schema.multiplex_analyzers
-          if max_complexity ||= schema.max_complexity
```
esjee: This breaks `GraphQL::Analysis::MaxQueryComplexity::across a multiplex#test_0001_returns errors for all queries:`!

graphql.gemspec
```diff
@@ -28,6 +28,9 @@ Gem::Specification.new do |s|
   s.add_development_dependency "guard-rake"
   s.add_development_dependency "guard-rubocop"
   s.add_development_dependency "listen", "~> 3.0.0"
+  s.add_development_dependency "pry"
+  s.add_development_dependency "pry-stack_explorer"
+  s.add_development_dependency "m"
```
Willianvdv: Intentional change?

esjee: Yes. I'll extract this to a different PR once this is ready to go (currently waiting for CI)

spec/graphql/analysis/max_query_complexity_spec.rb
```diff
@@ -50,7 +50,7 @@
     end
   end
 
-  describe "when complexity is overriden at query-level" do
```
Willianvdv: @esjee does it matter if you decrease or increase at query level? If not, you can use the original description.

Willianvdv: Nevermind, there are two specs now!

rmosolgo:
Thanks for digging in on this! I added one commit that uses a fix similar to your first try, being smarter with the way that `schema.max_complexity` is applied as a default.

Also, this way, when you run a single query, the complexity analyzer is only run _once_ since we use `nil` as max_complexity.

I also clipped out a few of the dev-dependency commits since we're taking care of that elsewhere.

PR #813, opened 2017-06-26
Add pry gem in an attempt to make development on this gem easier

esjee:
Hi,

Oftentimes I'd like to run a single test. The syntax for this is clunky using just rake, and I don't believe it's possible to specify a test by line number. A solution for this appears to be the succintly named gem "m": https://github.com/qrush/m

You can use it like so:
```
$ bundle exec m spec/graphql/execution/multiplex_spec.rb:52
# Running tests with run options -n "/^(test_0001_runs\\ multiple\\ queries\\ in\\ the\\ same\\ lazy\\ context)$/" --seed 33536:

.

Finished tests in 0.004828s, 207.1371 tests/s, 207.1371 assertions/s.


1 tests, 1 assertions, 0 failures, 0 errors, 0 skips
```

or

```
$ bundle exec m spec/graphql/execution/multiplex_spec.rb
# Running tests with run options -n "/^(test_0001_runs\\ multiple\\ queries\\ in\\ the\\ same\\ lazy\\ context|test_0001_returns\\ a\\ mix\\ of\\ errors\\ and\\ values|test_0001_is\\ provided\\ as\\ context:|test_0001_runs\\ query\\ instrumentation\\ for\\ each\\ query\\ and\\ multiplex\\-level\\ instrumentation)$/" --seed 36965:

....

Finished tests in 0.022976s, 174.0960 tests/s, 174.0960 assertions/s.


4 tests, 4 assertions, 0 failures, 0 errors, 0 skips

```

Perhaps worth noting is that it doesn't have support for full paths. So, for instance, this breaks:

```
$ bundle exec m /home/esjee/work/graphql-ruby/spec/graphql/execution/multiplex_spec.rb
Failed loading test file:
cannot load such file -- .//home/plague/work/graphql-ruby/spec/graphql/execution/multiplex_spec.rb
There were no tests found.
```

In addition to "m", I like using pry a lot to debug failing tests. I find it a very valuable tool for code spelunking, and being to see what's going on under the hood.

To add a breakpoint for pry to step over, you can modify the source code and add "binding.pry". For example:

```diff
diff --git a/lib/graphql/schema.rb b/lib/graphql/schema.rb
index 53e73210..54aafd4f 100644
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -223,6 +223,7 @@ module GraphQL
     # @see {Query#initialize} for arguments.
     # @return [Hash] query result, ready to be serialized as JSON
     def execute(query_str = nil, **kwargs)
+      binding.pry
       if query_str
         kwargs[:query] = query_str
       end
```

I hope you're open to adding these dependencies to the gem!


esjee:
Note that there's some troubles getting pry-stack_explorer (specifically its dependency "binding_of_caller") to work on JRuby. This has to do with the optimizations that JRuby does, which removes the possibility of binding_of_caller to, well, figure out what a caller of a function is. To solve this, I've moved the dependencies from the gemspec to the gemfile, and specified that JRuby should not install pry-stack_explorer.

rmosolgo:
Thanks for taking a look at the development workflow! I'm sure it could use some love 😬 

Adding `pry` is :+1: from me. 

Regarding adding `m`, there are some solutions in the current bundle for running tests in isolation, what do you think of these?

- __single test__: `minitest-focus` provides the `focus` helper, which can isolate one or more `it ...` examples: 

   ```ruby 
    focus 
    it "does something" do 
    ```

   Then, when you run the tests, if _any_ tests were `focus`'d, then only the focused tests are executed.

- __single file__: Use the `TEST=` env var: 
 
  ```
  rake test TEST=spec/graphql/schema_test.rb
  ```

  docs: http://www.rubydoc.info/gems/rake/Rake/TestTask

- __`guard`__: I use guard to manage my development flow:

  ```
  be guard 
  ```

  It watches the files in the project:

  - Whenever a file in `lib/` is modified, the corresponding file in `spec/` is executed and `rubocop` checks the file 
  - Whenever a file in `spec/` is modified, it is executed 
  - Whenever `parser.y` or `lexer.rl` is modified, the parser & lexer are rebuilt 
  - Pressing enter runs everything: rebuild lexer and parser, run all tests, rubocop 

If we improve the docs on using `minitest-focus`, `TEST=` and `guard`, would that do the job?

esjee:
Thanks! I think that works well enough to not require the m gem. Will you add this helpful somewhere, or can I help out in that?

rmosolgo:
I'll add some docs for it!

PR #815, opened 2017-06-27
[Masking] Hide types when no members of that type are visible

xuorig:
Currently, object types and input types that are not queryable because fields or arguments of that type are hidden still show up in the introspection.

That is because the warden does not make that check in `get_type`.

This PR attempts to solve this in an efficient way by building an inverse mapping of `Type => Array<Field|Argument>` at schema build time. This means by looking up a type you get an array of schema members *of* that type.

The warden can then verify that if an object is "referenced" by fields, there must be at least one that is visible to show that type.

@rmosolgo I'd still like to add more tests to this PR but would love to get your opinion on this approach!

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -17,6 +17,7 @@ require "graphql/schema/unique_within_type"
 require "graphql/schema/validation"
 require "graphql/schema/warden"
 require "graphql/schema/build_from_definition"
+require "graphql/schema/type_references_map"
 
 module GraphQL
   # A GraphQL schema which may be queried with {GraphQL::Query}.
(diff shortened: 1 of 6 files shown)


eapache:
Does this handle types which are only exposed via fields typed as an interface?

xuorig:
@eapache has a good point.

`field :a, SomeInterface, visibility: hidden`

Should hide types that implement that interface 🤔 will work on adding this in in this PR too.

eapache:
I was also wondering about the inverse case: if you have
```graphql
field :a, SomeType, visibility: hidden
field :b, SomeInterface
```
where `SomeType` implements `SomeInterface`, then `SomeType` should still be visible in the schema.

rmosolgo:
I'm refactoring these same files in #819 but don't worry about the conflicts -- if you get something worked out here, I'd be happy to help with the merge.

rmosolgo:
Closing this because of inactivity, feel free to reopen it if you pick this up again!

PR #817, opened 2017-06-28
Fix bug concerning `as` and `prepare` on arguments

phyrog:
When using `as` and `prepare` in argument blocks, they were overwritten by
the default values:

```ruby
argument :fooBar, types.String do
  as :foo_bar # will be overwritten by the default value (nil)
end
```

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -109,9 +109,10 @@ module GraphQL
       if default_value != NO_DEFAULT_VALUE
         argument.default_value = default_value
       end
-      argument.as = as
-      argument.prepare = prepare
-
+      as && argument.as = as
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi @phyrog, thanks for getting to the bottom of this! 

Do you mind adding a spec to verify that this behavior works now? I think you could do something like: 

```ruby
arg = GraphQL::Argument.define do 
  name "arg" 
  as "arg_name"
end 

assert_equal "arg_name", arg.as
```

Just to be sure that block-style definition works properly. Do you want to take try at it?

phyrog:
Sure, done.

rmosolgo:
Awesome, thanks again!

PR #822, opened 2017-07-04
Consistent root type naming in code examples

bwillis:
I believe this is the more consistent naming used throughout the guide.


rmosolgo:
Thanks for fixing this! Just published to graphql-ruby.org :)

PR #823, opened 2017-07-05
test(VariableUsagesAreAllowed) add failing test for list dimension

rmosolgo:
Oops, some cases of list variables are not handled properly. Here's the failing case: 

- `imageUrl(sizes: ...)` expects an input of type `[ImageSize!]`
- Here are two valid query string:

  ```graphql
  # Variable is list type: 
  query GetImage($sizes: [ImageSize!]) { 
    imageUrl(sizes: $sizes)
  }
  ```

  ```graphql 
  # Variable is _not_ list type: 
  query GetImage($size: ImageSize!) { 
    # But it is wrapped in a list where it is used:
    imageUrl(sizes: [$size])
  }
  ```

- Here's an invalid query string:

  ```graphql 
  # Variable is list type: 
  query GetImage($sizes: [ImageSize!]) { 
    # AND it is wrapped in a list where it is used:
    imageUrl(sizes: [$sizes])
  }
  ```

  It's being passed as a _double_ list (`[[ImageSize!]]`), which is _not_ valid. But, graphql-ruby doesn't give a validation error!

Closes #816

PR #824, opened 2017-07-05
Fix list variable validation

esjee:
See https://github.com/rmosolgo/graphql-ruby/pull/823 for more info

Just creating a new PR to get it to run on Travis!

### DIFF
--- a/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
+++ b/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
@@ -12,8 +12,8 @@ module GraphQL
         }
 
         context.visitor[GraphQL::Language::Nodes::Argument] << ->(node, parent) {
-          node_values = Array.wrap(node.value)
-          return if !(node_values.all? { |value| value.is_a? GraphQL::Language::Nodes::VariableIdentifier })
+          node_values = Array.wrap(node.value).select { |value| value.is_a? GraphQL::Language::Nodes::VariableIdentifier }
+          return if node_values.none?
 
           arguments = nil
           case parent
--- a/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
+++ b/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
@@ -18,7 +18,6 @@ module GraphQL
           node_values.each do |node_value|
             arguments = nil
 
-
             case parent
             when GraphQL::Language::Nodes::Field
               arguments = context.field_definition.arguments
--- a/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
+++ b/lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
@@ -15,22 +15,22 @@ module GraphQL
           node_values = Array.wrap(node.value)
           return if !(node_values.all? { |value| value.is_a? GraphQL::Language::Nodes::VariableIdentifier })
 
-          node_values.each do |node_value|
-            arguments = nil
-
-            case parent
-            when GraphQL::Language::Nodes::Field
-              arguments = context.field_definition.arguments
-            when GraphQL::Language::Nodes::Directive
-              arguments = context.directive_definition.arguments
-            when GraphQL::Language::Nodes::InputObject
-              arg_type = context.argument_definition.type.unwrap
-              if arg_type.is_a?(GraphQL::InputObjectType)
-                arguments = arg_type.input_fields
-              end
-            else
-              raise("Unexpected argument parent: #{parent}")
+          arguments = nil
+          case parent
+          when GraphQL::Language::Nodes::Field
+            arguments = context.field_definition.arguments
+          when GraphQL::Language::Nodes::Directive
+            arguments = context.directive_definition.arguments
+          when GraphQL::Language::Nodes::InputObject
+            arg_type = context.argument_definition.type.unwrap
+            if arg_type.is_a?(GraphQL::InputObjectType)
+              arguments = arg_type.input_fields
             end
+          else
+            raise("Unexpected argument parent: #{parent}")
+          end
+
+          node_values.each do |node_value|
             var_defn_ast = declared_variables[node_value.name]
             # Might be undefined :(
             # VariablesAreUsedAndDefined can't finalize its search until the end of the document.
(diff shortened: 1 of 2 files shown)


lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
```diff
@@ -7,32 +7,36 @@ class VariableUsagesAreAllowed
...
-            arg_type = context.argument_definition.type.unwrap
-            if arg_type.is_a?(GraphQL::InputObjectType)
-              arguments = arg_type.input_fields
+          node_values = Array.wrap(node.value)
+          return if !(node_values.all? { |value| value.is_a? GraphQL::Language::Nodes::VariableIdentifier })
+
+          node_values.each do |node_value|
+            arguments = nil
+
+
```
Willianvdv: Hello extra newline :)

spec/graphql/static_validation/rules/variable_usages_are_allowed_spec.rb
```diff
@@ -80,4 +80,67 @@
...
+    describe 'list is in the argument' do
+      let(:query_string) {
+        <<-GRAPHQL
+        query ($size: ImageSize!) {
+          imageUrl(sizes: [$size])
+        }
+        GRAPHQL
+      }
+
+      it "works" do
```
Willianvdv: "is a valid query"?

lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
```diff
@@ -7,32 +7,35 @@ class VariableUsagesAreAllowed
...
-            arg_type = context.argument_definition.type.unwrap
-            if arg_type.is_a?(GraphQL::InputObjectType)
-              arguments = arg_type.input_fields
+          node_values = Array.wrap(node.value)
+          return if !(node_values.all? { |value| value.is_a? GraphQL::Language::Nodes::VariableIdentifier })
+
+          node_values.each do |node_value|
+            arguments = nil
+
+            case parent
```
rmosolgo: This case statement is inside the `node_values.each` loop, but does its result depend on `node_value`? 

If we know that the result of this case statement is always the same, can we move it _outside_ the loop? That way, we evaluate it exactly once (rather than evaluating it for each item in `node_values`).

lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
```diff
@@ -7,32 +7,35 @@ class VariableUsagesAreAllowed
...
-          when GraphQL::Language::Nodes::Field
-            arguments = context.field_definition.arguments
-          when GraphQL::Language::Nodes::Directive
-            arguments = context.directive_definition.arguments
-          when GraphQL::Language::Nodes::InputObject
-            arg_type = context.argument_definition.type.unwrap
-            if arg_type.is_a?(GraphQL::InputObjectType)
-              arguments = arg_type.input_fields
+          node_values = Array.wrap(node.value)
+          return if !(node_values.all? { |value| value.is_a? GraphQL::Language::Nodes::VariableIdentifier })
```
rmosolgo: What happens if the query string is like this: 

```graphql 
query GetStuff($int1: Int, $int2: Int) {
  stuff(ints: [1, 2, $int1, $int2])
}
```

where there is a _mix_ of literal values (`1, 2`) and variables (`$int1, $int2`)? I think it would fail this check, since they aren't _all_ variables. 

I wonder, could we still validate those by _filtering_ the node values: 

```ruby 
variable_values = node_values.select { |n| n.is_a?(GraphQL::Language::Nodes::VariableIdentifier) }
if variable_values.any? 
  # ... 
  variable_values.each do ... 
```

Do you think that would work here?

lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
```diff
@@ -66,6 +71,29 @@ def create_error(error_message, var_type, ast_var, arg_defn, arg_node, context)
         message("#{error_message} on variable $#{ast_var.name} and argument #{arg_node.name} (#{var_type.to_s} / #{arg_defn.type.to_s})", arg_node, context: context)
       end
 
+      def wrap_var_type_with_depth_of_arg(var_type, arg_node)
```
rmosolgo: :+1:

lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
```diff
@@ -66,6 +71,29 @@ def create_error(error_message, var_type, ast_var, arg_defn, arg_node, context)
...
+      end
+
+      def depth_of_array(some_array)
+        result = 0
+        while (flattened = some_array.flatten) != some_array
+          result += 1
+          some_array = flattened
+        end
+        result
+      end
```
rmosolgo: I'm a bit worried by the duplication of the array in `some_array.flatten`. In query strings with deeply nested lists, this could result in a lot of Ruby allocations! Would it be possible to measure the depth without duplicating the array?

For example: 

```ruby 
# @return [Integer] Returns the max depth of `array`, or `0` if it isn't an array at all
def depth_of_array(array) 
  case array 
  when Array 
    max_child_depth = 0 
    array.each do |item| 
      item_depth = depth_of_array(item) 
      if item_depth > max_child_depth 
        max_child_depth = item_depth 
      end 
    end 
    1 + max_child_depth 
  else 
    0
  end 
end 
```

(Since this implementation returns `1` for a one-dimensional array, we could remove the `1 +` in `wrap_var_type_with_depth_of_arg`)

esjee: I've used your suggestion. Alternatively, could go for a solution that uses `flatten!` instead. That avoids the recursive method call, but would need to run benchmarks to see if that's helpful or not.

lib/graphql/static_validation/rules/variable_usages_are_allowed.rb
esjee: I tried to add a failing test similiar to your example, but it passess 🙈 

Could you help me write a red test?

rmosolgo:
Sorry for the delay on getting back to you! I've just pushed a failing test (and a couple of passing tests for other cases, just to be sure :) )

rmosolgo:
This is great! Thanks again for all your work on this. It was a tricky one :)

PR #828, opened 2017-07-07
Mention dataloader as other batching solution

sheerun:
Disclaimer: I'm an author.

[facebook/dataloader](https://github.com/facebook/dataloader) accepted my solution as alternative implementation of their library, so I though I could add it here as well.

Dataloader tries to mimic facebook/dataloader api as best as possible. In contrast to `batch-graphql` it offers more general solution that can be used even without graphql-ruby. The loaders created with `dataloader` also compose better as their return type is A+ [promise](https://github.com/lgierth/promise.rb).

EDIT: Link for better access:
https://github.com/sheerun/dataloader

### DIFF
--- a/guides/schema/lazy_execution.md
+++ b/guides/schema/lazy_execution.md
@@ -14,6 +14,8 @@ With lazy execution, you can optimize access to external services (such as datab
 
 [`graphql-batch`](https://github.com/shopify/graphql-batch) provides a powerful, flexible toolkit for lazy resolution with GraphQL.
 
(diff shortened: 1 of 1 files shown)


ianks:
This is a javascript library, yeah? It might not be fit for mentioning alongside `graphql-batch`, which is a Ruby implementation.

rmosolgo:
I think it's a _Ruby_ implementation of the same idea. Let me just take a quick look before merging, it looks really cool!

sheerun:
Indeed, it is Ruby implementation with very similar API to the original.

rmosolgo:
Thanks for the update!

PR #829, opened 2017-07-08
Add UnionType#resolve_type for more ganular control in unions

### LINKED ISSUE #529: Suggestion: Resolve type definition at Field or Interface/Union level
Currently my GraphQL schema doesn't necessarily match the underlying runtime object schema. 

I'm using Interfaces/Union types to define the GraphQL API, but from an implementation perspective there isn't a clear & obvious way to map runtime types to the appropriate `Type` definition at the `Schema` level without extracting various logic into that global `resolve_type` call. 

Likewise, the resolve type depends on the context it's being resolved to (i.e. the same object may resolve to TypeA in UnionAB, and TypeX in UnionXY).

Suggestion:

- Allow unions & Interfaces to define a `resolve_type` handler that will be attempted first, before the global handler.
- Should this be extended to the field definition as well? eg. `field :scope, ScopeUnion, resolve: -> (obj) { obj }, resolve_type: -> { obj.range? ? RangeScopeType : StandardScopeType }`

@rmosolgo What do you think of this? Or am I barking up the wrong tree and this doesn't make sense for some reason that I'm missing?

colepatrickturner:
## Purpose
This feature allows a union type to resolve to more specific types when handling a value. This is useful when working with objects that may fit into multiple types. When executing the query
will call `resolve_type(value, field_ctx)` on either Unions or Interfaces.

Union type's call can be definable via `resolve_type ->(value, ctx) { }`. Interface type's call returns `ctx.query.resolve_type(value)` which is the default implementation for resolving both types.

### Tests:
- Added a new test that raises an error if a returned type is not within possible_types
- Added an implementation in the dummy schema, though I would consider altering this by adding a new type derived from MilkType (for example KefirType) in which it is re-casted as a MilkType - but this would alter a significant portion of other tests.

### Guides
- I have not updated any guides or documentation on the static site. If there are any I should do for this PR please advise.

```
831 tests, 4720 assertions, 0 failures, 0 errors, 0 skips
Running RuboCop...
Inspecting 340 files
....................................................................................................................................................................................................................................................................................................................................................

340 files inspected, no offenses detected
```

Edit:

Resolves #529

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -210,7 +210,7 @@ module GraphQL
               )
             when GraphQL::TypeKinds::UNION, GraphQL::TypeKinds::INTERFACE
               query = field_ctx.query
-              resolved_type = query.resolve_type(value)
+              resolved_type = field_type.resolve_type(value, field_ctx)
               possible_types = query.possible_types(field_type)
 
               if !possible_types.include?(resolved_type)
--- a/lib/graphql/union_type.rb
+++ b/lib/graphql/union_type.rb
@@ -24,13 +24,14 @@ module GraphQL
   #  }
   #
   class UnionType < GraphQL::BaseType
-    accepts_definitions :possible_types
+    accepts_definitions :possible_types, :resolve_type
     ensure_defined :possible_types
 
     def initialize
       super
       @dirty_possible_types = []
       @clean_possible_types = nil
+      @resolve_type_proc = nil
     end
 
     def initialize_copy(other)
@@ -64,8 +65,25 @@ module GraphQL
       end
     end
 
+    def resolve_type(value, ctx)
+      if !@resolve_type_proc.nil?
+        resolved_type = @resolve_type_proc.call(value, ctx)
+        if !include?(resolved_type)
+          raise(NotImplementedError, "Unrecognized possible type kind: #{resolved_type}")
+        end
+
+        return resolved_type
+      end
+
+      ctx.query.resolve_type(value)
+    end
+
+    def resolve_type=(new_resolve_type_proc)
+      @resolve_type_proc = new_resolve_type_proc
+    end
+
     protected
 
-    attr_reader :dirty_possible_types
+    attr_reader :dirty_possible_types, :resolve_type_proc
   end
 end
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -86,7 +86,7 @@ module GraphQL
 
     # Find out which possible type to use for `value`.
     # Returns self if there are no possible types (ie, not Union or Interface)
-    def resolve_type(value)
+    def resolve_type(value, ctx)
       self
     end
 
--- a/lib/graphql/interface_type.rb
+++ b/lib/graphql/interface_type.rb
@@ -42,6 +42,10 @@ module GraphQL
       GraphQL::TypeKinds::INTERFACE
     end
 
+    def resolve_type(value, ctx)
+      ctx.query.resolve_type(value)
+    end
+
     # @return [GraphQL::Field] The defined field for `field_name`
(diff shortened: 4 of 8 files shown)


lib/graphql/union_type.rb
```diff
@@ -24,19 +24,21 @@ module GraphQL
...
       @dirty_possible_types = []
       @clean_possible_types = nil
+      @resolve_type_proc = nil
     end
 
     def initialize_copy(other)
       super
       @clean_possible_types = nil
       @dirty_possible_types = other.dirty_possible_types.dup
+      @resolve_type_proc = other.resolve_type_proc
```
rmosolgo: I think this is the default behavior of `#initialize_copy`, to assign each ivar of the original to to the new copy, so we can remove this line and let `super` handle it.

lib/graphql/union_type.rb
```diff
@@ -64,8 +66,25 @@ def possible_types
...
     end
 
+    def resolve_type(value, ctx)
+      if !@resolve_type_proc.nil?
+        resolved_type = @resolve_type_proc.call(value, ctx)
+        if !include?(resolved_type)
+          raise(NotImplementedError, "Unrecognized possible type kind: #{resolved_type}")
+        end
+
+        return resolved_type
```
rmosolgo: It's possible to [hide types and fields](http://graphql-ruby.org/schema/limiting_visibility.html) for individual queries. Should we do anything special here if the user-provided `@resolve_type_proc` returns a type which _ought_ to be hidden for this query?

lib/graphql/execution/execute.rb
```diff
@@ -210,7 +210,7 @@ def resolve_value(owner, parent_type, field_defn, field_type, value, selection,
               )
             when GraphQL::TypeKinds::UNION, GraphQL::TypeKinds::INTERFACE
               query = field_ctx.query
-              resolved_type = query.resolve_type(value)
+              resolved_type = field_type.resolve_type(value, field_ctx)
```
rmosolgo: :+1: Seems right! Is this change covered by a corresponding change in the tests, to ensure that we _keep_ using the type's `resolve_type` in future versions?

colepatrickturner: I don't believe so, I'll add a new one for that.

lib/graphql/union_type.rb
colepatrickturner: Ok! Will remove

colepatrickturner:
> We should add a specific test for the new feature (that objects of the same Ruby class can resolve to different types at different times).

I have rebased in a test and some changes to the schema that don't change much while trying to use the same schema.

- Restructured `COW` from OpenStruct to Struct for the purpose of being usable in a Union
I was worried about potential conflict here changing from OpenStruct to Struct, because previous tests used `COW` in a `GraphQL::Function` to specify a type. However that does not affect the use of functions to derive type within testing.

- Added `Goat`, `GOATS` and `COWS`, `AnimalUnion` and `AnimalAsCowUnion`
These are used in the dummy schema derive types within unions.
`AnimalUnion` tests the default implementation using the same types that `AnimalAsCowUnion` will type-cast into `CowType`.

- Added `Query->allAnimal` and `Query->allAnimalAsCow`
`allAnimal` returns both the Cow and Goat while `allAnimalAsCow` will typecast a goat into a cow.

I hope I covered the tests per your request. My intention was to use the same schema because I believe this change to be imperative to testing a whole schema to ensure it's compatibility between schema-level and union-level resolution. Thanks for your review and time :)

rmosolgo:
This looks great! Nice work coercing that wacky schema to suit your needs :)

colepatrickturner:
Thank you 🙏

broskoski:
👏 Nice work, thank you!

akaKuruma:
Thank you! 🍻

PR #830, opened 2017-07-09
Add search_object_graphql to related projects list.

RStankov:
`SearchObject::Plugin::GraphQL` is a plugin for [SearchObject](https://github.com/rstankov/SearchObject), for defining search resolvers. 

Example:

```ruby
class PostResolver
  include SearchObject.module(:graphql)

  type types[PostType]

  scope { Post.all }

  option(:name, type: types.String)       { |scope, value| scope.where name: value }
  option(:published, type: types.Boolean) { |scope, value| value ? scope.published : scope.unpublished }
end
```


rmosolgo:
Thanks for adding this! Cool project :)

PR #831, opened 2017-07-10
Enforce EnumValue name uniqueness

swalkinshaw:
Raises an exception when trying to add a value that already exists with the same name.

From the [spec](https://facebook.github.io/graphql/#sec-Enum):

> There must be at least one and they must have unique names.

Two questions:

1. is this a good place for the validation or should it be done during schema validation?
2. is there a better existing exception to raise?


rmosolgo:
I think this is a great solution! Maybe you could defer the error until schema validation but honestly, a native Ruby error will provide a better DX 😬

PR #836, opened 2017-07-12
Faster CI?

rmosolgo:
Builds [used to take 5 mins](https://travis-ci.org/rmosolgo/graphql-ruby/builds/252937570) but now they take [25 mins](https://travis-ci.org/rmosolgo/graphql-ruby/builds/252852853). Looking back through history, I don't see any smoking guns for where the jump came from. It looks like they just slowly crept up! 

But looking at the current build matrix, I got two ideas: 

- Use [travis's suggestion](https://docs.travis-ci.com/user/languages/ruby/#JRuby%3A-Installation-takes-a-long-time-on-Standard-and-Trusty-environments) for faster jruby installation 
- Stop testing ruby 2.2 & ruby 2.3 (is there any reason to test those since we already have 2.1 and 2.4?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,12 @@ env:
 before_install:
   - gem update --system --no-doc
   - gem install bundler
-
+addons:
+  apt:
+    packages:
+      - haveged
 rvm:
-- 2.1.0 # Lowest version officially supported by that gem
-- 2.2.7
-- 2.3.4
+- 2.1.0 # Lowest version officially supported by the gem
 - 2.4.1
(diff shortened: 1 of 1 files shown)


swalkinshaw:
What about using `fast_finish`? It doesn't speed up builds but at least shows PR failure quicker.

https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/

rmosolgo:
that's nice, at least we'll spend less time finishing broken builds :)

rmosolgo:
Try again next time :S

Issue #839, opened 2017-07-14
Query execution fails because Array doesn't have the wrap method


opan:
Hi all! I have a little problem here. I am trying to use an `Argument` on my query. Here is my code:

```ruby

# frozen_string_literal: true

module ProcuraEngine
  module Presenter
    class MeasurementUnitGraphql < GraphqlBase
      MeasurementUnitType = GraphQL::ObjectType.define do
        name 'Measurement Unit'

        field :id, !types.ID
        field :ident_name, !types.String
        field :name, !types.String
        field :symbol, types.String
        field :desc, types.String
        field :created_at, DateTimeType
        field :updated_at, DateTimeType
        field :updated_by_id, !types.Int
        field :created_by_id, !types.Int
      end

      QueryType = GraphQL::ObjectType.define do
        name 'Query'

        field :items do
          type types[MeasurementUnitType]
          argument :id, types.ID, 'Measurement Unit ID'
          resolve -> (obj, args, context) do
            # binding.pry
            context[:measurement_unit_repo].query
          end
        end
        field :resource do
          type !types.String
          resolve -> (_obj, _args, _context) do
            'measurement_unit'
          end
          # resolve -> (_obj, _args, _context) { 'measurement_unit' }
        end
        field :total_server_items, !types.Int do
          resolve -> (_obj, _args, context) { context[:measurement_unit_repo].count }
        end
      end

      Schema = GraphQL::Schema.define do
        query QueryType
      end

      def execute(measurement_unit_repo, graphql_str = nil)
        graphql_str ||= %{
          query {
            resource
            total_server_items
            items {
              ident_name
              name
              symbol
              desc
              created_at
              created_by_id
            }
          }
        }
        binding.pry
        Schema.execute(graphql_str, context: { measurement_unit_repo: measurement_unit_repo })
      end
    end
  end
end


# Execution code

      query = %{
        query {
          resource
          total_server_items
          items(id: 1) {
            ident_name
            name
            symbol
            desc
            created_at
            created_by_id
          }
        }
      }

ProcuraEngine::Presenter::MeasurementUnitGraphql.new.execute(measurement_unit_repo, query)

```

I run these through rspec and the errors I get is:

```
 Failures:

  1) Main::Application /measurement_unit - GET filter should working properly
     Failure/Error: Schema.execute(graphql_str, context: { measurement_unit_repo: measurement_unit_repo })

     NoMethodError:
       undefined method `wrap' for Array:Class
[...]

PR #840, opened 2017-07-14
fix: remove usage of Rails' Array.wrap

esjee:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/839

It looks like active_support happens to be included in the tests, which is why
this succeeded in the first place. Nevertheless, we want to avoid this dependency!
Introduced in https://github.com/rmosolgo/graphql-ruby/pull/824


rmosolgo:
ick, concurrency failure on JRuby 😖  not good, but not a blocker here. 

Thanks for taking care of this! I'll cut another patch version later today.

PR #842, opened 2017-07-14
Add a Travis build without Rails installed

### LINKED ISSUE #841: Create a travis build that doesn't install Rails to make sure that any PR that introduce a dependency on Rails/active_support will break
Creating this issue as a response to https://github.com/rmosolgo/graphql-ruby/issues/839

Rails is added as a test dependency, and there are several tests that rely on it being installed. Unfortunately, that also means that any accidental dependencies will easily go unnoticed. In order to resolve this, let's create a travis build that specifically does *not* install Rails to make sure that such mistakes are caught in the future. It will be necessary to avoid running tests that depend on Rails being installed on this specific build.

esjee:
Fixes #841


rmosolgo:
Seems like this is getting really close!

esjee:
Sorry, I have been a bit distracted lately and unable to give this PR the attention it deserves. I'll try to get it wrapped up in the upcoming week/weekend.

rmosolgo:
😻  Is this ... _working_ ?!

rmosolgo:
It looks like you added a new gemfile, but there aren't any changes to the `Appraisals` file. Can you update that file too, so that if we regenerate the gemfiles, we don't lose `without_rails.gemfile`?

esjee:
Finally! :tada: 
Thanks for the help @rmosolgo!

rmosolgo:
Woot! This is great, hopefully ActiveSupport-only errors are behind us :)

PR #843, opened 2017-07-14
Print Enum name when raising duplicate value error.

xuorig:
Had trouble finding which enum had the problem :)

cc @swalkinshaw


xuorig:
woops 😂

rmosolgo:
I cancelled a build

PR #849, opened 2017-07-17
Prepare arguments for nested input types

skorfmann:
When using `prepare` in a nested input type, the provided block wasn't called. This PR fixes it.


rmosolgo:
Oops, thanks for tracking this down 😬  Looks like a good fix and thanks for adding tests too!

Issue #727, opened 2017-05-17
[Discussion] Schema Definition using the IDL


xuorig:
Discussion on a PR to auto-camelize fields recently has brought up some valid concerns:
https://github.com/rmosolgo/graphql-ruby/pull/555#issuecomment-281194713
https://github.com/rmosolgo/graphql-ruby/pull/555#issuecomment-281482442

This seems to be part of a bigger discussion on wether or not we want users write actual `graphql`, or ruby when defining a GraphQL Schema.

Currently, defining a schema usually looks like this:

```ruby
QueryType = GraphQL::ObjectType.define do
  name "Query"
  description "The query root of this schema"

  field :posts do
    type types.[PostType]
    description "Get all Posts"
    resolve ->(obj, args, ctx) { Post.all }
  end
end
```

The resulting IDL for this, the source of truth, looks like this:

```graphql
# The query root of this schema
type Query {
  # Get all Posts
  posts: [Post!]!
}
```

With this approach, the ruby defined schema != resulting IDL. A good thing about it is that users can write a schema without knowing the GraphQL Langage. At the same time, like discussed in the issue linked above, we don't help users learning GraphQL.

Another argument against this approach is that we're mixing the Schema / Interface with the business logic. Maybe what we want is using the IDL (the best tool we got to define a Schema) and then using plain Ruby to define the resolvers / business logic. With a proper set of conventions, this could be quite powerful. A mix of plain `.graphql` files and their ruby object counter parts.

# Defining a Schema using the IDL

A simple approach to this could be as simple as having something like this:

```graphql
# post.graphql

type Post {
  title: String!
  description: String
}
```

```ruby
# post.rb

class Post < GraphQL::Resolver
  def initialize(post, args, ctx)
    @post = post
    @args = args
    @ctx = ctx
  end

  # These can even be delegated instead by default
  def title
    post.title
  end

  def description
    post.description
  end
end
```

But this approach limits us in certain ways. @ Shopify, we extracted a lot of common logic into field definition arguments. For example, if a field needs to be batched or preloaded, it may look like something like this:

```ruby
field :posts, [Post], preload: :posts
```

It's also possible to extend graphql-ruby's DSL using `accepts_definition` and Instrumenters to come up with something like this:

```ruby
field :posts do
  type types.[Post]
  preloads :posts
end
```

However, with an IDL approach, we lose that DSL and we need a way to extend resolvers. What if I wanted to preload `posts` in this example?

```ruby
[...]

PR #871, opened 2017-08-06
[RFC] Schema Implementation API

rmosolgo:
I want to provide a new API that is more familiar to Ruby developers. I hope that less DSL will make people more productive and happier 😀 See all the advantages and opinions discussed here: https://github.com/rmosolgo/graphql-ruby/issues/727

The implementation is currently quite sloppy. But I've started writing docs and tests for what I _want_ to build, so I'll be working towards that and cleaning up as I go. So, if anyone has feedback on those, please do share it!


xuorig:
Have you thought of ways that could enable someone to collocate definition & implementation? My fear is the context switching between a very large `.graphql` file and smaller impl classes might be a bit hard. wdyt?

guides/schema/implementation.md
```diff
@@ -0,0 +1,146 @@
...
+end
+```
+
+## Combine Them
+
+Build a schema by loading the structure from a file and passing the namespace as `implementation:`
+
+```ruby
+MySchema = GraphQL::Schema.from_definition(
+  "app/graphql/my_schema.graphql",
```
cjoudrey: `from_definition` currently accepts the IDL as an argument, was this meant to be `File.read('app/graphql/my_schema.graphql')` or were you thinking about making `from_definition` support paths?

cjoudrey:
> Have you thought of ways that could enable someone to collocate definition & implementation?

That or adding some documentation to show how one could split one large `schema.graphql` into multiple smaller `.graphql` files. I suppose if `from_definition` accepts the IDL as a string, people could stitch the files together manually, but it might be useful to provide a simpler way.

xuorig:
> That or adding some documentation to show how one could split one large schema.graphql into multiple smaller .graphql files

That'd be great. HoweverI  still think that if it was possible to quickly see a definition and an implementation without too much context switching it would be great!

xuorig:
Have you thought of ways to extend the schema / implementation with common behaviour? Let's take my usual `field :a, :string, preloads: b` as an example. How would you implementsomething like a custom `preloads` argument with thisimplementation?

guides/schema/implementation.md
rmosolgo: I merged this behavior separately in #872 👌

cjoudrey:
> Let's take my usual field `:a, :string, preloads: b` as an example. How would you implementsomething like a custom `preloads` argument with this implementation?

It would still be possible to do this in the resolver function, no?

i.e.

```ruby
def a
  MyPreloader.preload(:b).then {
    # stuff
  }
end
```

Maybe just a bit more verbose, but that might not be a bad thing. 🤔

xuorig:
> Maybe just a bit more verbose, but that might not be a bad thing. 🤔

Hmm agreed but this might get very verbose and repetitive. We've got some fields that will

`preload: { something: [ { something: something {] }, visibility: ..., filters: []`

That would result in possible, huge resolvers, and having it directly in the resolver wouldnt let us use Instrumenters or other ways of reusing logic across fields / resolvers

rmosolgo:
> split one large `schema.graphql` into multiple smaller `.graphql` files

:+1: to this, but I think it can be a separate task from this one. (Make `.from_definition` accept a glob and parse each matching file)

>  collocate definition & implementation?

I think it's interesting, but I'm scared of re-introducing traversal-based schema building. The current implementation has issues with Rails autoloading, cyclical dependencies, and modifying schema objects.

So, it's ok with me to start small (parse a single `.graphql` file) and build from there. 

I can think of a few ways to collocate them which can be explored later, are there other approaches to consider? 

- __Ruby__: make another DSL for describing GraphQL with Ruby

  ```ruby 
  class Blog 
    field :authors, { first: "Int", after: "String" }, "AuthorConnection"
    field :topics, "[Topic!]"
    # ... 
  end 
  ``` 

  Maybe the DSL could be better, but it's not exciting to me, I'd rather leverage the GraphQL IDL.

- __Inline GraphQL, Ruby String__: Put the GraphQL definition _in_ the `.rb` file, as a string 

  ```ruby 
  class Blog 
    structure <<-GRAPHQL 
      authors(first: Int, after: String): AuthorConnection
      topics: [Topic!]
    GRAPHQL 
  end 
  ```

- __Inline GraphQL, Ruby Comment__: Put the GraphQL definition _in_ the `.rb` file, as a comment
 
  ```ruby 
  class Blog 
    =begin GRAPHQL 
    type Blog {
      authors(first: Int, after: String): AuthorConnection
      topics: [Topic!]
    }
    =end 
  end 
  ```

  We parse the file and extract the relevant bit. Could be buggy, but we can use the GraphQL IDL as-is, with no Ruby involved.

rmosolgo:
> Have you thought of ways to extend the schema / implementation with common behaviour?

I like your class-method based suggestions here and in your first issue, I think something like this would be ok: 

```ruby 
# When accessing the posts field, preload the ratings
preloads posts: :ratings 
def posts(first:, after:) 
  # ... 
end 
```

I think we could implement this on top of ActiveSupport-like callbacks, `before_resolve first: :preload_a`, etc, to call methods before/after resolving. This would be a more human-friendly version of field instrumentation. 

How does that sound?

swalkinshaw:
```ruby
# When accessing the posts field, preload the ratings
preloads posts: :ratings 
def posts(first:, after:) 
  # ... 
end
```

🤔  are you saying that instead of specifying all preloads at the top of the file, you just locate these above each method? Never thought of it and it solves a problem I had which came up in the other issue.

ie: `before_action :whatever, only: %i(show)`

This works pretty well because the number of actions in a controller, or lifecycle callbacks in a model, are pretty limited. Some controllers could have a lot of actions, but most are CRUD based.

Using the same type of solution for GraphQL is a little different since it's pretty common to have a lot of fields which could quickly get out of control.

rmosolgo:
My suggestion is a lot like controller hooks, just a different order of arguments:

```ruby 
before_resolve :posts, :do_thing 
# is like 
before_action :do_thing, only: :posts 
```

It could be anywhere in the class definition. It requires a lot of hackery to make it work without an explicit method name. A nice side-effect of the explicit name is that you could apply `before_resolve` to fields which _don't_ have corresponding methods (that is, they're implemented by delegating to the proxied object).

lib/graphql/schema/implementation/method_call_implementation.rb
```diff
@@ -0,0 +1,37 @@
...
+              when :context
+                method_args[:context] = ctx
+              when :irep_node
+                method_args[:irep_node] = ctx.irep_node
+              when :ast_node
+                method_args[:ast_node] = ctx.ast_node
+              end
+            end
+
+            proxy.public_send(@method_name, **method_args)
```
theorygeek: This would probably be over-optimization at this point of development, but I wonder if eventually the `call` method here could be constructed at build-time by constructing the method as a string and then `instance_eval`'ing it. The idea is to avoid constructing the hash and doing this logic here - it may not even make a difference on performance, though.

Something like:
```ruby

call_args = @graphql_arguments.map { |a| "#{a}: args[:#{a}]" }.join(', ')
# "letter: args[:letter]"

method_body = <<~RUBY
  def call(proxy, args, ctx)
    proxy.#{@method_name}(#{call_args})
  end
RUBY

# def call(proxy, args, ctx)
#   proxy.suit(letter: args[:letter])
# end

instance_eval(method_body)
```

rmosolgo: I'll definitely look into it, I had a [similar TODO](https://github.com/rmosolgo/graphql-ruby/pull/871/files#diff-0e778e4bfa4af9b9ca9735e3fd803cd3R11) myself :) 

If it's better on the benchmarks, I'll go for it!

lib/graphql/schema/implementation/build_resolve.rb
```diff
@@ -0,0 +1,62 @@
...
+                  graphql_args << name
+                elsif SPECIAL_ARGS.include?(name)
+                  special_args << name
+                else
+                  raise InvalidImplementationError.new(
+                    "unexpected keyword #{name.inspect}, see #{impl_class.name}##{method_name} (#{field_method.source_location.join(":")})"
+                  )
+                end
+              end
+            end
```
theorygeek: So I'm thinking about `:rest` and `:keyrest` args:
```
# :rest
def my_resolver(*args)
  # => args == [{ :some_argument1 => 'some value 1' }]
end

# :keyrest
def my_other_resolver(**kwargs)
  # => kwargs == { :some_argument1 => 'some value 1' }
end
```

A couple of thoughts on this:
- It seems like we would want to support those kinds of arguments. And if they exist, we'd basically only check to make sure that there are no positional args, and no unknown keyword args.
- Thinking through that case, if a user _did_ use `*args` or `**kwargs`, I'm not sure whether they would expect to get the special args or not.
- It seems like it'd be most predictable if we could guarantee that every call into a resolver method would be done with the exact same signature. But that would require us to take a stance that either all methods will accept `:ast_node` & friends, or no methods would accept them.

That third point is really about avoiding brittle scenarios, where changing from seemingly identical method signatures ends up changing behavior. For example:
```ruby
def my_resolver(some_arg1:, irep_node:, ast_node:)
end

# Several months later, someone refactors this into:
def my_resolver(**kwargs)
end
```

Perhaps the best option right now would be to raise an `InvalidImplementationError` upon encountering a `:rest` or `:keyrest` parameter.

rmosolgo: > raise an InvalidImplementationError upon encountering a :rest or :keyrest parameter.

👏

lib/graphql/schema.rb
```diff
@@ -467,8 +468,8 @@ def self.from_introspection(introspection_result)
     # @param default_resolve [<#call(type, field, obj, args, ctx)>] A callable for handling field resolution
     # @param parser [Object] An object for handling definition string parsing (must respond to `parse`)
     # @return [GraphQL::Schema] the schema described by `document`
-    def self.from_definition(string, default_resolve: BuildFromDefinition::DefaultResolve, parser: BuildFromDefinition::DefaultParser)
-      GraphQL::Schema::BuildFromDefinition.from_definition(string, default_resolve: default_resolve, parser: parser)
+    def self.from_definition(string, default_resolve: BuildFromDefinition::DefaultResolve, parser: BuildFromDefinition::DefaultParser, implementation: nil)
+      GraphQL::Schema::BuildFromDefinition.from_definition(string, default_resolve: default_resolve, parser: parser, implementation: nil)
```
swalkinshaw: Implementation isn't working right now since it's hardcoded to `nil` here

rmosolgo:
I'm not actively working on this branch. I decided to try experimenting with GitHub's schema a bit to see if I can find something that would work, then try it out a bit before upstreaming it. 

I'm worried about building something from scratch, then releasing it & applying it, but learning that it doesn't really solve our problems 😖

PR #885, opened 2017-08-10
Check if ActiveRecord::Relation is defined

### LINKED ISSUE #884: Fix hidden dependency on active_record
I believe there's an hidden dependency on `active_record` that's causing a
 `NameError: uninitialized constant GraphQL::Relay::RelationConnection::ActiveRecord`

When querying for a connection page info like:
```graphql
{
 user {
   friends(first:3) {
     pageInfo {
       hasNextPage
       hasPreviousPage
     } 
   }
 }
}
```
I got:
```
{
  "status": 500,
  "error": "Internal Server Error",
  "exception": "#<NameError: uninitialized constant GraphQL::Relay::RelationConnection::ActiveRecord>",
...
``` 

If one doesn't depend on `active_record` at all it should all still work without issues.

jpserra:
Fixes #884

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -100,8 +100,7 @@ module GraphQL
 
       # If a relation contains a `.group` clause, a `.count` will return a Hash.
       def relation_count(relation)
-        count_or_hash = case relation
-        when ActiveRecord::Relation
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for taking care of this! A CI fix is in the works at #842

PR #889, opened 2017-08-12
Fix selected operation name when there are no selections in a query

xuorig:
Noticed Query would 💥 in two places when a query with no operations is executed, for example a query with only comments.

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -161,7 +161,10 @@ module GraphQL
     end
 
     def irep_selection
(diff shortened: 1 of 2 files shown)


rmosolgo:
🤔  I had some _reason_ for making these two methods but now I can't remember what it was ...

rmosolgo:
🤷‍♂️  I forget, I thought it was a graphql-pro thing but it tests ok with this change. Thanks for fixing it!

PR #890, opened 2017-08-13
Use argument node instead of parent in ArgumentsAreDefined errors

cjoudrey:
The `node` reported when an error occurs in `ArgumentsAreDefined` seems incorrect.

Suppose the `cheese` field does not accept the argument `silly` we currently report this error:

```graphql
query getCheese {
  cheese(silly: false, id: 2) { source }
  ^^^^^^
}
 ```

In reality I think we want:

```graphql
query getCheese {
  cheese(silly: false, id: 2) { source }
         ^^^^^
}
 ```

I'm not sure if this was intentional, but with this fix we are more aligned with the reference implementation: https://github.com/graphql/graphql-js/blob/6953f9714b785bbcd89f0103fd73401c06d5f6fe/src/validation/__tests__/KnownArgumentNames-test.js#L118-L126

cc @rmosolgo

### DIFF
--- a/lib/graphql/static_validation/rules/arguments_are_defined.rb
+++ b/lib/graphql/static_validation/rules/arguments_are_defined.rb
@@ -7,7 +7,7 @@ module GraphQL
         if argument_defn.nil?
           kind_of_node = node_type(parent)
           error_arg_name = parent_name(parent, defn)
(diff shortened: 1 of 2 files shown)


rmosolgo:
Nice find! Thanks for taking care of it.

PR #892, opened 2017-08-13
Fix Sequel::Error when limit is 0

### LINKED ISSUE #887: Sequel::Error Limits must be greater than or equal to 1
Using the following query, if I get the `cursor` from the first `user` and use it as before argument, I get a `500 "#<Sequel::Error: Limits must be greater than or equal to 1>"`.
```graphql
{
  account(id: 1) {
    users {
      edges {
        cursor
      }
    }
    beforeUser: users(before: "MQ==") { # cursor of the first user
      edges { node { __typename } }
    }
  }
}
```
```
{
  "status": 500,
  "error": "Internal Server Error",
  "exception": "#<Sequel::Error: Limits must be greater than or equal to 1>",
  "traces": {
    "Application Trace": [
     ...
    ],
    "Framework Trace": [
      {
        "id": 0,
        "trace": "sequel (4.48.0) lib/sequel/dataset/query.rb:653:in `limit'"
      },
      {
        "id": 1,
        "trace": "graphql (1.6.6) lib/graphql/relay/relation_connection.rb:129:in `sliced_nodes'"
      },
      {
        "id": 2,
        "trace": "graphql (1.6.6) lib/graphql/relay/relation_connection.rb:56:in `paged_nodes'"
      },
...
```

Looking at the code, 

```ruby
## graphql-ruby/lib/graphql/relay/relation_connection.rb

111       # Apply cursors to edges¬
112       def sliced_nodes¬
113         return @sliced_nodes if defined? @sliced_nodes¬
114 ¬
115         @sliced_nodes = nodes¬
116 ¬
117         if after¬
118           offset = (relation_offset(@sliced_nodes) || 0) + offset_from_cursor(after)¬
119           @sliced_nodes = @sliced_nodes.offset(offset) if after¬
120         end¬
121 ¬
122         if before && after¬
123           if offset_from_cursor(after) < offset_from_cursor(before)¬
124             @sliced_nodes = @sliced_nodes.limit(offset_from_cursor(before) - offset_from_cursor(after) - 1)¬
125           else¬
126             @sliced_nodes = @sliced_nodes.limit(0)¬
127           end¬
128         elsif before¬
129           binding.pry¬
130           @sliced_nodes = @sliced_nodes.limit(offset_from_cursor(before) - 1)¬
131         end¬
132 ¬
133         @sliced_nodes¬
134       end¬
```

in line `130` `@sliced_nodes` is a dataset and we're calling `.limit(0)` on it since `offset_from_cursor(before) - 1 = 0`

Here's a simple example:
```ruby
[11] () <GraphQL::Relay::RelationConnection>: 0> @sliced_nodes
=> #<Sequel::Postgres::Dataset: "<<<Super secret dataset>>>">
[12] () <GraphQL::Relay::RelationConnection>: 0> offset_from_cursor(before) - 1
=> 0

[13] () <GraphQL::Relay::RelationConnection>: 0> @sliced_nodes.limit(offset_from_cursor(before) - 1)
Sequel::Error: Limits must be greater than or equal to 1
```

jpserra:
Sequel, contrarily to ActiveRecord, does not support `limit(0)` queries
and raises a `Sequel::Error` instead.

To fix this, we need to deal with this differently by using `where(false)`
instead when we are attempting to perform a query for zero rows in Sequel.

Fixes #887

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -121,15 +121,24 @@ module GraphQL
 
         if before && after
           if offset_from_cursor(after) < offset_from_cursor(before)
-            @sliced_nodes = @sliced_nodes.limit(offset_from_cursor(before) - offset_from_cursor(after) - 1)
+            @sliced_nodes = limit_nodes(@sliced_nodes,  offset_from_cursor(before) - offset_from_cursor(after) - 1)
           else
-            @sliced_nodes = @sliced_nodes.limit(0)
+            @sliced_nodes = limit_nodes(@sliced_nodes, 0)
           end
+
         elsif before
-          @sliced_nodes = @sliced_nodes.limit(offset_from_cursor(before) - 1)
+          @sliced_nodes = limit_nodes(@sliced_nodes, offset_from_cursor(before) - 1)
         end
 
         @sliced_nodes
+      end
+
+      def limit_nodes(sliced_nodes, limit)
+        if limit > 0 || defined?(ActiveRecord::Relation) && sliced_nodes.is_a?(ActiveRecord::Relation)
+          sliced_nodes.limit(limit)
+        else
+          sliced_nodes.where(false)
+        end
       end
 
       def sliced_nodes_count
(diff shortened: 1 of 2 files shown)


rmosolgo:
Nice work, thanks for tracking down the fix on this one!!

PR #900, opened 2017-08-20
plural nodes is not correct here

aomran:
I assume `nodes` here is a typo otherwise it should define `GraphQL::Relay::Node.plural_field`.


rmosolgo:
Oops, thank you!

PR #901, opened 2017-08-20
documentation has Schema#resolve_type deprecation warning

aomran:
Documentation has the older/deprecated syntax. I changed the example to something a bit more OO.

Thanks for the great gem and documentation!


rmosolgo:
Thanks for updating this! Personally, I can't recommend the duck-typing approach, so I'm going to revert the example, but thank you for fixing the syntax there, I guess I missed a spot 😅

PR #906, opened 2017-08-23
Fix unnecessary `hasNextPage` COUNT() query

### LINKED ISSUE #899: Connections generating unnecessary COUNT query
For connection queries that request both `edges` and `hasNextPage`, a COUNT query is run even if the number of edges returned is less than the requested limit.

In queries like this for example, if less than 1000 users are returned, then hasNextPage should always be false:
```
{
  query {
    users(first: 1000) {
      edges {
        ...
      }
      pageInfo {
        hasNextPage
      }
    }
  }
}
```

It looks like optimization was added in [February](https://github.com/rmosolgo/graphql-ruby/commit/03464e02e38f9950ae33a29481242f8199d8e347) but taken out in [May](https://github.com/rmosolgo/graphql-ruby/commit/838b403e1c146ed9b491b93fee1e696aade93828), presumably because it doesn't work when both `first` and `last` are set.  Could this functionality be returned to `RelationConnection#sliced_nodes_count` in the scenario where only `first` is set?

ddgromit:
Fixes #899 

## Change

This PR changes `has_next_page` to first check the length of `paged_nodes` to see if its less than the number of entities requested through `first` to avoid a separate COUNT query when know from the `paged_nodes` query that there is no next page.

The only negative performance impact is that in the rare case `hasNextPage` is called without `edges` the full paged nodes query will be run instead of just a COUNT query.

## Notes

In a future optimization, we could possibly request `last + 1` nodes and avoid the COUNT(*) query altogether.

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -24,11 +24,11 @@ module GraphQL
       end
 
       def has_next_page
-        !!(first && sliced_nodes_count > first)
+        !!(first && paged_nodes_length >= first && sliced_nodes_count > first)
       end
 
       def has_previous_page
-        !!(last && sliced_nodes_count > last)
+        !!(last && paged_nodes_length >= last && sliced_nodes_count > last)
       end
 
       def first
@@ -155,6 +155,14 @@ module GraphQL
       def paged_nodes_array
         return @paged_nodes_array if defined?(@paged_nodes_array)
         @paged_nodes_array = paged_nodes.to_a
+      end
+
+      def paged_nodes_length
+        if paged_nodes.respond_to?(:length)
+          paged_nodes.length
+        else
+          paged_nodes_array.length
+        end
       end
     end
 


lib/graphql/relay/relation_connection.rb
```diff
@@ -156,6 +156,14 @@ def paged_nodes_array
         return @paged_nodes_array if defined?(@paged_nodes_array)
         @paged_nodes_array = paged_nodes.to_a
       end
+
+      def paged_nodes_length
+        if paged_nodes.respond_to?(:length)
```
ddgromit: AFAIK in ActiveRecord the Relation can be simply checked for length, but in Sequel the query must first be run with `.all` and converted to an array - so I'm reusing `pages_nodes_array` here for that use case.

rmosolgo:
Nice find, and thanks for taking care of this! Yes, the `last + 1` would be a nice trick to apply here. I've used it in some more recent implementations but haven't had a chance to revisit this one 😓

lib/graphql/relay/relation_connection.rb
yaphetsglhf: is there anyway to avoid running with .all in sequel.  in my biz, lt is becoming a bottleneck now.

PR #916, opened 2017-08-25
Do not allow double non-null constraints

### LINKED ISSUE #826: Schema can contain non-null(non-null(foo))
Schema generated by `graphql-ruby` can become invalid by wrapping a non-null (required) type in non-null again.

```ruby
field :foo, !!String
```

The introspection query will then return this part:

```json
            {
              "name": "foo",
              "description": null,
              "args": [

              ],
              "type": {
                "kind": "NON_NULL",
                "name": null,
                "ofType": {
                  "kind": "NON_NULL",
                  "name": null,
                  "ofType": {
                    "kind": "SCALAR",
                    "name": "String",
                    "ofType": null
                  }
                }
              },
              "isDeprecated": false,
              "deprecationReason": null
            },
```

Type is inspected as `String!!`.

---

This fails to load in (for example) GraphiQL with the error `Error: No nesting nonnull` as it explicitly checks for this problem.

esjee:
Fixes #826

### DIFF
--- a/lib/graphql/non_null_type.rb
+++ b/lib/graphql/non_null_type.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 module GraphQL
+  class DoubleNonNullTypeError < GraphQL::Error
+  end
+
   # A non-null type modifies another type.
   #
   # Non-null types can be created with `!` (`InnerType!`)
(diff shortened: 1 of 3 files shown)


rmosolgo:
👏  sweet, thanks!

PR #918, opened 2017-08-25
Implements .fetch on Arguments

esjee:
A suggested solution to https://github.com/rmosolgo/graphql-ruby/issues/453. Let me know what you think!

### DIFF
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -33,10 +33,10 @@ module GraphQL
         @argument_values.key?(key_s)
       end
 
-      def fetch(key, default = (default_provided = true; nil))
+      def fetch(key, default = (no_default_provided = true; nil))
         key_s = key.is_a?(String) ? key : key.to_s
         if @argument_definitions.key?(key_s)
-          if default_provided
+          if no_default_provided
             @argument_values.fetch(key_s, NULL_ARGUMENT_VALUE).value
           else
             if @argument_values.key?(key_s)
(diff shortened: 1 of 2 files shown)


mvgijssel:
LGTM @Willianvdv plz accept.

esjee:
Before merging this, I think it might be interesting to look at some way to use methods instead. So:

```
args.id
```

rather than

```
args[:id]
```

rmosolgo:
Just to give my two cents, in my opinion, the _behavior_ is right but the _name_ is off. 

`Hash#fetch(key)` raises an error if `key` isn't present in the hash. But, the `#fetch` method introduced here _may not_ raise if the argument is defined by not present. I'm hesitant about using the same method name to describe different behavior. 

So, what name could we give this method?

```ruby 
args.get(:something) # ? 
args[:something] # change the behavior of this? could be breaking? 
args.something # use method call syntax for this behavior?
```

Method calls sound promising, but to avoid `method_missing`, we'd have to generate a classes at schema build time!

volkanunsal:
@rmosolgo This sounds like an OpenStruct. I always wrap my args in an [`immutable-struct`](https://github.com/stitchfix/immutable-struct). It provides a nice immutable implementation of an OpenStruct, sans the assignment operator, and has a [simple implementation](https://github.com/stitchfix/immutable-struct/blob/master/lib/immutable-struct.rb).

```ruby
args = ImmutableStruct.new(:foo)
args.new(foo: 1)
args.foo
# => 1
args.foo = 1
NotMethodError
```

rmosolgo:
yes, very similar, _except_ we can define all the methods ahead of time, since we know the schema!

esjee:
#924 seems more promising, closing this one out

PR #919, opened 2017-08-25
[Warden] Some edge cases

xuorig:
Had some time to work on https://github.com/rmosolgo/graphql-ruby/pull/815 again.

### Added to schema

  - #references_to
  - #union_memberships
  - #root_types
  - #unions

`@unions`, `@union_memberships` and `@type_reference_map` are now populated during schema build / Traversal.

### New behaviours: 

a type should be visible if:
  - Its own visibility is true

and one of these:
  - It's a root type
  - It's an introspection type
  - its referenced by a field or argument
  - it has a visible abstract type (part of Union or Interface)
  - It is an abstract type and has visible members


rmosolgo:
👏  These behaviors all sound right to me, I'll review the code closely soon!

lib/graphql/schema.rb
```diff
@@ -221,6 +221,22 @@ def types
...
+      rebuild_artifacts unless defined?(@type_reference_map)
+      @type_reference_map.fetch(type_name, [])
+    end
+
+    # Returns a list of Union types in which a type is a member
+    # @param type [GraphQL::ObjectType]
+    # @return [Array<GraphQL::UnionType>] The field named `field_name` on `parent_type`    def union_memberships(type)
+    def union_memberships(type)
+      rebuild_artifacts unless defined?(@unions)
+      @unions.select { |union| possible_types(union).include?(type) }
```
rmosolgo: Can we cache this calculation? For example, could `@unions` be a 

```ruby
Hash.new { |h, k| h[k] = [] }
```

Where the keys are object types and the values are lists of union types which contain that object type? 

(A similar cache is built by `warden.rb`, but could we cache it at schema-level instead/also?)

lib/graphql/schema.rb
```diff
@@ -221,6 +221,22 @@ def types
       end
     end
 
+    # Returns a list of Arguments and Fields referencing a certain type
+    # @param type_name [String]
+    # @return [Hash]
+    def get_members_of_type(type_name)
```
rmosolgo: What do you think of calling this `#get_references_to` (or `#references_to`)? Interfaces and Unions have "members", but I don't think of `User.handle` as a member of `String`, how does that sound to you?

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,53 @@ def root_type_for_operation(op_name)
...
+      end
+
+      def referenced_by_visible_members?(type_defn)
+        members = @schema.get_members_of_type(type_defn.unwrap.name)
+        # Types should be visible if there are
+        # no member definitions with that type
+        return true unless members.any?
+        # Only hide the type when there are fields
+        # but none are visible.
+        members.any? { |member| visible?(member) }
```
rmosolgo: Just a matter of taste but I thought I'd share, another approach is: 

```ruby 
# Show this type if: 
# - it has no references to begin with 
# - it has references, and some of them are still visible 
members.empty? || members.any? { |m| visible?(m) } 
```

Personally, I prefer it because: 
- The method has only one expression which returns, at it's at the end of the method body (no explicit `return` part-way through)
- There's no inverted conditional logic (no `unless`)

Of course, it's 👌  as is, just thought I'd share!

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,53 @@ def root_type_for_operation(op_name)
...
+      end
+
+      def member_or_implements?(type_defn)
+        interfaces(type_defn).any? || union_memberships(type_defn).select { |u| visible?(u) }.any?
```
rmosolgo: I think `.any? { ... }` is preferable to `.select { ... }.any?` since it will return after finding a first match instead of rebuilding a new array, would that work here?

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,53 @@ def root_type_for_operation(op_name)
...
+
+        if type_defn.kind.object? && has_abstract_type?(type_defn)
+          visible_implementation?(type_defn)
+        else
+          referenced_by_visible_members?(type_defn)
+        end
+      end
+
+      def visible_implementation?(type_defn)
+        if orphan?(type_defn)
```
rmosolgo: It's _possible_ that a type would be: 

- Specified as an orphan _AND_ 
- Referenced somewhere in the schema 

I don't think graphql-ruby disallows that. For example, it might happen if a type _used to_ be an orphan but was eventually used somewhere else in the schema. 

In that case, we should still check for references to the type, right? If so, then we should always use the `else` branch for this method, so the whole method body would be: 

```ruby
referenced_by_visible_members?(type_defn) || member_or_implements?(type_defn)
```

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,53 @@ def root_type_for_operation(op_name)
...
+
+      def visible_type?(type_defn)
+        return false unless visible?(type_defn)
+
+        if type_defn.kind.object? && has_abstract_type?(type_defn)
+          visible_implementation?(type_defn)
+        else
+          referenced_by_visible_members?(type_defn)
+        end
+      end
```
rmosolgo: I _think_ another valid implementation of this method is: 

```ruby 
# Type `t` is visible if: 
# - It's visible in its own right AND 
# - Either: 
#   - Some parts of the schema reference it OR 
#   - Some of its abstract types are visible 
def visible_type?(t) 
  visible?(t) && (
      referenced_by_visible_members?(t) || 
      member_or_implements?(t)
    )
end 
```

What do you think of that implementation? Is it valid? Are there other upsides/downsides? 

Personally I prefer it because we can remove some of intermediary methods and I find the control flow a bit easier to follow. I don't _think_ it would have much worse performance, if worse at all. What do you think?

xuorig: The problem with this one is that is would return true in the case where for example:

  - orphan type A implements interface B
  - interface B is hidden

I think type A should be hidden in that case, but currently `referenced_by_visible_members?` will return true because type A has no references (we show it in those cases)

Something that would allow use this simpler logic would be to hide always when there are no references, but that seems like a weird default that would confuse people 🤔

xuorig: That brings a question though, should types that are not part of any abstract types and not referenced by anything even be allowed?

xuorig:
Great comments @rmosolgo, I've addressed most of them. There's just one question left :)

lib/graphql/schema/warden.rb
rmosolgo: > `referenced_by_visible_members?` will return true

ohhh i missed that! 🙇 

> not part of any abstract types and not referenced by anything

I _think_ it's ok, i mean, from a client's point of view, what's the difference compared to now? 

A type which is not part of an abstract type can't be spread anywhere, so that's ok. 

I guess `__type(name: "...")` would start returning `nil`, but I think that's ok too. 

Any reason _not_ to do it?

lib/graphql/schema.rb
```diff
@@ -221,6 +222,22 @@ def types
...
+    def references_to(type_name)
+      rebuild_artifacts unless defined?(@type_reference_map)
+      @type_reference_map.fetch(type_name, [])
+    end
+
+    # Returns a list of Union types in which a type is a member
+    # @param type [GraphQL::ObjectType]
+    # @return [Array<GraphQL::UnionType>] list of union types of which the type is a member
+    def union_memberships(type)
+      rebuild_artifacts unless defined?(@unions)
```
eapache: Not entirely clear to me if we're expected to be able to call `rebuild_artifacts` more than once (the name kind of suggests yes), but if so it should probably clear `@union_memberships` or that data will be stale/cached?

lib/graphql/schema/warden.rb
xuorig: I'll update the PR to have that behaviour, I think it makes sense, just afraid it's confusing to users 🤔

rmosolgo: >  If none of your fields use Int, should Int be hidden?

I _think_ this is the current behavior!

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,44 @@ def root_type_for_operation(op_name)
...
+        @unions ||= read_through { |obj_type| @schema.union_memberships(obj_type).select { |u| visible?(u) } }
+        @unions[obj_type]
+      end
+
       def visible_field?(field_defn)
-        visible?(field_defn) && visible?(field_defn.type.unwrap)
+        visible?(field_defn) && visible_type?(field_defn.type.unwrap)
+      end
+
+      def visible_type?(type_defn)
```
xuorig: Quite a conditional, but basically a type should be visible if:

  - Its own visibility is `true`

and one of these:

  - It's a root type
  - It's an introspection type
  - its referenced by a field or argument
  - it has a visible abstract type (part of Union or Interface)
  - It is an abstract type and has visible members

xuorig:
Turns out Directives also can reference types 👯

spec/graphql/schema/build_from_definition_spec.rb
```diff
@@ -69,6 +69,8 @@ def build_schema_and_compare_output(definition)
 
 # What a great type
 type Hello {
+  anEnum: Color
```
xuorig: Any idea why the Enum was printed before if it wasnt referenced by anything? 🤔

xuorig: Probably the way build_from_definition works :/

xuorig:
@rmosolgo ready for another round of review 💃

lib/graphql/schema.rb
```diff
@@ -545,6 +562,9 @@ def rebuild_artifacts
         traversal = Traversal.new(self)
         @types = traversal.type_map
         @instrumented_field_map = traversal.instrumented_field_map
+        @type_reference_map = traversal.type_reference_map
+        @unions = traversal.unions
+        @union_memberships = {}
```
rmosolgo: I think we should try to populate `@union_memberships` _here_ instead of gradually growing it at runtime. 

The reason is because of how it will behave in threaded and forked environments. In a forking server, every fork will end up with its own copy of `@union_memberships` since each fork will be writing to it, it's not the end of the world, but it's not what we want. In threaded environments, `@union_memberships` becomes shared mutable state, since different threads may be adding to it at different times. In CRuby, the GIL keeps us safe, but in JRuby, this can raise an error.

So, do you think it's possible to pre-fill this hash now instead of fill-as-we-go?

spec/graphql/schema/build_from_definition_spec.rb
rmosolgo: Woah, good find! Yeah, I guess because it takes _all_ the types in the `.graphql` and passes them as `orphan_types`: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/build_from_definition.rb#L100

lib/graphql/schema/warden.rb
rmosolgo: :+1: 😎 

It's a lot of syntax but for my part, I appreciate it!

lib/graphql/schema.rb
xuorig: ah good catch! will fix.

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,44 @@ def root_type_for_operation(op_name)
...
+            root_type?(type_defn) ||
+            type_defn.introspection? ||
+            referenced?(type_defn) ||
+            visible_abstract_type?(type_defn) ||
+            possible_types?(type_defn)
+          )
+      end
+
+      def root_type?(type_defn)
+        [@schema.query, @schema.mutation, @schema.subscription].include?(type_defn)
```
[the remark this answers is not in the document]
xuorig: I'm tempted of instead do:

```
    # in BaseType
    def root?
      @root ||= ROOT_NAMES.include?(name)
    end
```

Do you think this is valid logic? I'd need to take a look at the spec but IIRC this should indeed indicate that a type is root 🤔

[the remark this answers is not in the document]
xuorig: Guess I was wrong, can't find anywhere that roots have to be named something specific 🤔

lib/graphql/schema/warden.rb
```diff
@@ -126,8 +126,53 @@ def root_type_for_operation(op_name)
...
+      end
+
       def visible_field?(field_defn)
-        visible?(field_defn) && visible?(field_defn.type.unwrap)
+        visible?(field_defn) && visible_type?(field_defn.type.unwrap)
+      end
+
+      def visible_type?(type_defn)
+        visible?(type_defn) &&
+        has_visible_fields?(type_defn) && (
```
xuorig: thought of this scenario ^ if all fields are hidden on a type, the type should be hidden 🙈

xuorig:
Updated description to match the current state.

lib/graphql/schema/warden.rb
eapache: This seems ripe for an infinite loop if a type returns a field of itself.

xuorig: good catch 🤔 🙉 

Maybe this should be cached

lib/graphql/schema.rb
```diff
@@ -544,7 +568,11 @@ def rebuild_artifacts
         @rebuilding_artifacts = true
         traversal = Traversal.new(self)
         @types = traversal.type_map
+        @root_types = [query, mutation, subscription]
         @instrumented_field_map = traversal.instrumented_field_map
+        @type_reference_map = traversal.type_reference_map
+        @unions = traversal.unions
```
rmosolgo: is this ivar now unused? can we remove it (Traversal#unions)?

xuorig: yup they're both useless now! ✂️

PR #921, opened 2017-08-25
update limits guide to include the context parameter

codebender:
without this, I was seeing:

```
TypeError:
        no implicit conversion of nil into String
```


rmosolgo:
Thanks for the fix! Sorry the documentation fell behind there 😬

codebender:
No prob, if I run into any more I'll create PRs 😺

PR #924, opened 2017-08-28
Arguments as functions

esjee:
Alternative to #918


spec/support/dummy/schema.rb
```diff
@@ -394,7 +394,7 @@ def call(obj, args, ctx)
       argument :input, !ReplaceValuesInputType
       resolve ->(o, args, ctx) {
         GLOBAL_VALUES.clear
-        GLOBAL_VALUES.push(*args[:input][:values])
+        GLOBAL_VALUES.push(args.input.values)
```
rmosolgo: (I think the specs are failing because we lost a `*` here)

esjee: Aha! Good seeing!

Willianvdv:
@rmosolgo @esjee are you want to use the same technique for the `ctx`?

rmosolgo:
not in my opinion, since ctx is arbitrary keys, but argument keys can be known at build time. 

That's the last thing for this approach -- instead of using `define_singleton_method` at runtime, we should make classes at build-time, one class for each field with arguments 😬  happy to pair on that sometime

esjee:
To get a sense of how difficult it would be to migrate to this appproach, I tried to do this for our codebase (@hacker0x01).

The following sed command did most of the heavy lifting: `sed -ri 's args\[:([[:alnum:]_]+)\] args.\1 g' app/graphql/**/*.rb`
It doesn't catch _all_ cases: it doesn't handle nesting, or something like `args[:values].each { |a| a[:foo] }`. But thankfully those happen to be the exception.

The second thing I was running into was unit tests. We passed simple hashes to resolvers rather than proper arguments, so this was causing trouble. Luckily it's easy to wrap a hash if you have access to the argument definitions:

```
def prepare_args(mutation, some_hash)
  GraphQL::Query::Arguments.new(some_hash, argument_definitions: mutation.arguments)
end
```

If that fails, you can simply use something like OpenStruct, so at least the argument quacks like a `GraphQL::Query::Arguments`.

lib/graphql/input_object_type.rb
```diff
@@ -96,7 +96,8 @@ def coerce_non_null_input(value, ctx)
         end
       end
 
-      GraphQL::Query::Arguments.new(input_values, argument_definitions: arguments)
+      arguments_class ||= GraphQL::Query::Arguments.construct_arguments_class(argument_definitions: arguments)
```
esjee: I need this for a single test. I'm not sure if this is a better approach than defining than generating the classess during traversal of the schema. Thoughts?

lib/graphql/query/arguments.rb
```diff
@@ -1,12 +1,51 @@
...
+      private
+
+      def arg_class
+        @arg_class ||= begin
+          klass = Class.new(GraphQL::Query::Arguments).instance_exec(self) do |static_arguments|
+            static_arguments.argument_definitions.each do |_arg_name, arg_definition|
+              expose_as = arg_definition.expose_as.to_s
+
+              # Don't define a helper method if it would override something.
+              next if instance_methods.include?(expose_as)
```
swalkinshaw: Wondering if a warning should be logged here? Otherwise it could be confusing.

esjee: That seems helpful. What would be a good way to communicate that to the user? `warn(...)`?

swalkinshaw: Not sure, this gem doesn't have a logger or even use `puts` very much for output like this /shrug

Willianvdv: @esjee @swalkinshaw would it make sense to warn or even raise an exception if we hit this condition? I think it will result in a nicer upgrade path when folks have to resolve an error instead of debugging weird behaviour caused by calling the original method.

Willianvdv:
@esjee can you also update the CHANGELOG?

rmosolgo:
(No need to update the changelog -- I'll write it when I prepare the release. That way, we can be sure it ends up with the right version and we can avoid any conflicts.)

lib/graphql/input_object_type.rb
rmosolgo: I don't think we should create a new class here, we should be sure to do it all ahead of time. Which test fails? I'd be happy to take a try at another approach.

lib/graphql/query/literal_input.rb
```diff
@@ -86,7 +86,9 @@ def self.from_arguments(ast_arguments, argument_defns, variables)
           end
         end
 
-        GraphQL::Query::Arguments.new(values_hash, argument_definitions: argument_defns)
+        GraphQL::Query::Arguments
+          .construct_arguments_class(argument_definitions: argument_defns)
```
rmosolgo: Is this code called at runtime? I'm worried that we'll be creating classes on-the-fly instead of reusing ones that are already made. 

Maybe we should refactor `.from_arguments` to take different input. Instead of `argument_defns`, we should take `arguments_owner` (`GraphQL::Field` or `GraphQL::InputObjectType`) then we can get the arguments class from there. (Disclaimer: I don't remember _exactly_ how this method is used so it might not work!!)

lib/graphql/schema/traversal.rb
```diff
@@ -56,6 +56,12 @@ def visit(member, context_description)
               type_defn.arguments.each do |name, arg|
                 visit(arg.type, "Input field #{type_defn.name}.#{name}")
               end
+
+              # Construct arguments class here, which is later used to generate GraphQL::Query::Arguments
+              # to be passed to a resolver proc
+              type_defn.arguments_class = GraphQL::Query::Arguments.construct_arguments_class(
+                argument_definitions: type_defn.arguments,
+              )
```
rmosolgo: 👌

lib/graphql/schema/traversal.rb
```diff
@@ -74,9 +80,16 @@ def visit_fields(type_defn)
...
+
           instrumented_field_defn.arguments.each do |name, arg|
             visit(arg.type, "Argument #{name} on #{type_defn.name}.#{instrumented_field_defn.name}")
           end
+
+          # Construct arguments class here, which is later used to generate GraphQL::Query::Arguments
+          # to be passed to a resolver proc
+          instrumented_field_defn.arguments_class = GraphQL::Query::Arguments.construct_arguments_class(
+            argument_definitions: instrumented_field_defn.arguments,
+          )
```
rmosolgo: 👌

lib/graphql/query/arguments.rb
rmosolgo: > debugging weird behaviour caused by calling the original method.

This also worries me. I would prefer defining methods by default, does that sound viable? 

Alternatively, we could: 

- `warn(...)` when we encounter a naming conflict, then: 
  - accept either `define_method:` which creates the method in case of a conflict OR 
  - `silence_warning:` which suppresses the warning in case of a conflict (to avoid log noise)

lib/graphql/query/literal_input.rb
esjee: Yeah, I'm pretty sure this one's in runtime. There didn't seem to be an easy way to improve that, so I might need some help to look into that :grin:

lib/graphql/query/arguments.rb
esjee: > does that sound viable?

Unsure; I was running into trouble where some methods are defined _after_ this block, which would cause the methods to be overridden silently. Seems a bit flimsy. I'm unsure what a good solution is here; I'll add a `warn` for now.

rmosolgo:
Oh, now I see the failing test you were talking about: 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/list_type_spec.rb#L48-L51

IMO it's fine to just assign manually inside the test case. 

Maybe it's itching to extract some shared code, like a `#assign_arguments_class` method, but either way!

lib/graphql/query/literal_input.rb
```diff
@@ -39,11 +39,11 @@ def self.defaults_for(argument_defns)
         end
       end
 
-      def self.from_arguments(ast_arguments, argument_defns, variables)
+      def self.from_arguments(ast_node, argument_defns, variables)
```
rmosolgo: I think this refactor was the right idea, but applied in the wrong spot. 

Instead of switching `ast_arguments` to `ast_node`, then using `ast_node.arguments_class`, we should apply the same kind of refactor the _second_ parameter in this method. 

This method should accept `argument_owner_defn` instead of `argument_defns`, then reach for `argument_owner_defn.arguments_class` at line 89 below. Then, the callers should also be updated: instead of `type.arguments,` it should pass `type`, and instead of `definition.arguments`, it should pass `definition`.

That way we can re-use the class which was derived when the schema was built. 

The thing about AST nodes is that they're constructed fresh for every parsed query. So for people who accept dynamic queries via a GraphQL API, they'll be defining a new class for each field with arguments in the query 😬 

If that refactor _works_, then I think we can remove the `arguments_class` accessor from `Language::Nodes` classes, right?

lib/graphql/schema/traversal.rb
```diff
@@ -87,17 +103,30 @@ def visit(member, context_description)
...
+        instrumented_field_defn.arguments.each do |name, arg|
+          @type_reference_map[arg.type.unwrap.to_s] << arg
+          visit(arg.type, "Argument #{name} on #{type_defn.name}.#{instrumented_field_defn.name}")
+        end
+
+        # Construct arguments class here, which is later used to generate GraphQL::Query::Arguments
+        # to be passed to a resolver proc
+        instrumented_field_defn.arguments_class = GraphQL::Query::Arguments.construct_arguments_class(
+          argument_definitions: instrumented_field_defn.arguments,
+        )
```
rmosolgo: This bit is of code is used for Fields, InputObjects and Directives, do you think it's possible to refactor this into one method, maybe something like 

```ruby 
# This method: 
# - reads `.arguments 
# - constructs a class 
# - assigns it to `.arguments_class=`
GraphQL::Query::Arguments.construct_arguments_class(argument_owner)
```

esjee:
Thanks a bunch for the help @rmosolgo ! I tried to dry it up a bit. Hopefully this is what you had in mind.

rmosolgo:
That looks great, 👏 !

PR #928, opened 2017-08-29
Add more information in UnresolvedTypeError message

xuorig:
Adds a bit more information on how to fix an `UnresolvedTypeError` by hinting to `type_from_object` and displaying the value trying to be resolved to a type.

I've seen confusion from teammates trying to implement Interfaces and unions and we've added similar errors, thought it might be useful here too 🤷‍♂️


rmosolgo:
That's nice, thanks for upstreaming it!

Issue #678, opened 2017-04-18
Better error message for malformed function definitions


gjtorikian:
I had defined a field like this:

``` ruby
connection :labels, function: Platform::Functions::Labels do
   ...
end
```

See the problem? Me either! When I ran the schema, I received the following error: `RuntimeError: Invalid complexity: nil on labels`

It took me some time to realize I had neglected the `.new` when defining `function: Platform::Functions::Labels.new do`.

I think it'd be pretty useful to bomb immediately if a `function` kwarg is not an instantiated object of some type.

rmosolgo:
:+1: I guess `responds_to` isn't a good enough check because in the above example, the object _did_ respond to the method. (Maybe this was a design error in GraphQL::Function, it can be used as a class _or_ instance)

rmosolgo:
(TBH the whole schema validation _thing_ needs a redo. The output doesn't help you find the right place  and it should include a backtrace)

rmosolgo:
Functions are out, Resolvers are in, and Resolvers work as expected with regards to the snippet above

PR #931, opened 2017-09-01
Failing spec - Complexity on root level connections with functions

Willianvdv:
This change results in:
```
Error:
GraphQL::Analysis::QueryComplexity::simple queries#test_0002_wont fail:
RuntimeError: Invalid complexity: nil on connectionWithoutComplexity
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/query_complexity.rb:64:in `get_complexity'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/query_complexity.rb:37:in `call'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/reducer_state.rb:15:in `call'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:68:in `block in visit_analyzers'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:67:in `each'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:67:in `visit_analyzers'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:63:in `reduce_node'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:59:in `block (2 levels) in reduce_node'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:58:in `each'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:58:in `block in reduce_node'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:57:in `each'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:57:in `reduce_node'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:43:in `block in analyze_query'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:42:in `each'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:42:in `analyze_query'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:11:in `block in analyze_multiplex'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:9:in `map'
    /Users/willian/src/graphql-ruby/lib/graphql/analysis/analyze_query.rb:9:in `analyze_multiplex'
    /Users/willian/src/graphql-ruby/lib/graphql/execution/multiplex.rb:169:in `with_instrumentation'
    /Users/willian/src/graphql-ruby/lib/graphql/execution/multiplex.rb:59:in `run_queries'
    /Users/willian/src/graphql-ruby/lib/graphql/execution/multiplex.rb:40:in `run_all'
    /Users/willian/src/graphql-ruby/lib/graphql/schema.rb:256:in `block in multiplex'
    /Users/willian/src/graphql-ruby/lib/graphql/schema.rb:554:in `with_definition_error_check'
    /Users/willian/src/graphql-ruby/lib/graphql/schema.rb:255:in `multiplex'
    /Users/willian/src/graphql-ruby/lib/graphql/schema.rb:232:in `execute'
    /Users/willian/src/graphql-ruby/spec/graphql/analysis/query_complexity_spec.rb:43:in `block (3 levels) in <top (required)>'
```

Found by @mvgijssel


xuorig:
Hey @Willianvdv I dug into this. It turns out Functions should always be instanciated:

`function: SomeConnectionFunction.new`

It's failing currently because the function class also has a complexity attribute, which may be `nil`...

Wonder how we could make that more obvious. Either way I think in your case the fix is simply instantiating the function :)

rmosolgo:
sounds like https://github.com/rmosolgo/graphql-ruby/issues/678

xuorig:
😂 oh. Yep.

lib/graphql/define/assign_object_field.rb
```diff
@@ -18,16 +18,18 @@ def self.call(owner_type, name, type_or_field = nil, desc = nil, function: nil,
...
-            complexity: function.complexity,
             name: name_s,
             type: function.type,
             resolve: function,
             description: function.description,
             function: function,
             deprecation_reason: function.deprecation_reason,
-          )
+          }
+          kwargs[:complexity] = function.complexity if function.complexity
```
xuorig: Complexity can't be null here if youre passingr a function instance.

The reason its null is because you're passing a Function class and not an instance.

I think we should probably fail hard in that case since it would probably end up failing elsewhere if we accept Function classes here?

Willianvdv: Ah! That makes sense, I totally overlooked that issue! I guess it would make sense to fail if you're not passing an instance.

PR #941, opened 2017-09-05
`connection?` is not properly set when parsing the IDL

gjtorikian:
When using ` GraphQL::Schema.from_definition`, it seems that `connection?` is always `false`. I've added a failing test, but have yet to find a root cause.

### DIFF
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -280,6 +280,9 @@ module GraphQL
               resolve: ->(obj, args, ctx) { default_resolve.call(field, obj, args, ctx) },
               deprecation_reason: build_deprecation_reason(field_definition.directives),
             )
+
+            type_name = resolve_type_name(field_definition.type)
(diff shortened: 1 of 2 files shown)


gjtorikian:
I'm having a _very_ hard time figuring out how to fix this. 😓

rmosolgo:
Good catch! Here's where `GraphQL::Field`s are created when parsing the IDL: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/build_from_definition.rb#L275-L282

I think you could assign [`field.connection=`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/field.rb#L181) _after_  building it, eg: 

```ruby
field.connection = field_definition.type.ends_with?("Connection") # or is it `String#end_with?`, I always forget
```

In any case, I think that's the spot of code that needs a fix!

gjtorikian:
Ok, based on that hint I think I got it.

rmosolgo:
👏 Sweet! Good call on the non-null handling.

PR #944, opened 2017-09-06
Find types from directives

xuorig:
Noticed this while working on visibility: Types may be discovered through Directive arguments.

WDYT @rmosolgo ?


rmosolgo:
Yes, definitely!

rmosolgo:
🚀

Issue #904, opened 2017-08-22
RFC: Re-raise errors with additional GraphQL backtrace


rmosolgo:
I just learned you can provide a backtrace when raising an error as the third argument to `raise(error_or_class, message, backtrace)`. So, you could amend a backtrace before raising an error, for example:

```ruby
def raise_error
  raise KeyError, "Something bad happened"
end

begin
  raise_error
rescue
  backtrace_additions = ["(added):1", "(added:2)"]
  raise $!, $!.message, backtrace_additions + $!.backtrace
end
# KeyError: Something bad happened
# 	from (added):1
# 	from (added:2)
# 	from (irb):2:in `raise_error'
# 	from (irb):6
# 	from /usr/bin/irb:12:in `<main>'
```

So, could we track graphql execution and _add_ entries to the backtrace when errors come from graphql resolve functions?

theorygeek:
Are you thinking that we would add things like the path in the query?
```graphql
query MyQuery {
  node(id: "...") { someField { someOtherField } }
}
# maybe other queries in the document too...
```

```ruby
# Show the name of the operation, and the path:
#       in someOtherField
#       in someField[3]
#       in node(id: "...")
# 	in MyQuery
# 	from (irb):2:in `raise_error'
# 	from (irb):6
# 	from /usr/bin/irb:12:in `<main>'
```

rmosolgo:
Yeah exactly! Maybe it could be optional while we try it out, eg 

```ruby 
# lib/graphql/error.rb
  def backtrace 
    if GraphQL.path_backtrace && @context 
      backtrace_from_context(@context) + super 
    else 
      super 
    end 
  end 
```

rmosolgo:
Here's another approach to modifying the backtrace: 

https://github.com/ko1/pretty_backtrace

The approach looks like:

- Add a [handler for raised errors](https://github.com/ko1/pretty_backtrace/blob/master/lib/pretty_backtrace.rb#L17)
- Use the [debug inspector](https://github.com/ko1/pretty_backtrace/blob/master/lib/pretty_backtrace.rb#L35) to backtrack through the RubyVM stack 
- [Gather the local variables](https://github.com/ko1/pretty_backtrace/blob/master/lib/pretty_backtrace.rb#L38-L42) and inspect them
- [Rebuild the stack trace](https://github.com/ko1/pretty_backtrace/blob/master/lib/pretty_backtrace.rb#L53) with new metadata

[...]

PR #946, opened 2017-09-08
Backtrace annotations

rmosolgo:
We can use `GraphQL::Tracing` to accumulate info about execution, then re-merge it with the backtrace when errors happen. 

Also, add `ctx.backtrace` so you can print the backtrace anytime with 

```ruby
puts ctx.backtrace 
# Loc  | Field             | Object    | Arguments | Result
# 10:9 | Query.pushContext | "rootval" | {}        | 1
# 2:7  | query             | "rootval" | {}        | {backtrace: {...}, pushContext: 1}
```

## Before 

```
NoMethodError: undefined method `each' for :not_a_list:Symbol
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:192:in `resolve_value'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:155:in `continue_resolve_field'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:129:in `resolve_field'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:71:in `block (2 levels) in resolve_selection'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:40:in `block in trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:71:in `call_tracers'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:40:in `trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:70:in `block in resolve_selection'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:63:in `each'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/execute.rb:63:in `resolve_selection'
```

## After 

```
GraphQL::Backtrace::TracedError: Unhandled error during GraphQL execution:

  undefined method `each' for :not_a_list:Symbol

Use #cause to access the original exception (including #cause.backtrace).

GraphQL Backtrace:
Loc  | Field            | Object     | Arguments | Result
1:29 | Thing.listField  | :something | {}        | (error)
1:20 | Query.field1     | nil        | {}        | {}
1:1  | query BrokenList | nil        | {}        | {}


    /Users/rmosolgo/code/graphql-ruby/lib/graphql/backtrace.rb:58:in `rescue in trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/backtrace.rb:61:in `trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:73:in `call_tracers'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:40:in `trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/multiplex.rb:50:in `run_queries'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/multiplex.rb:40:in `run_all'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/schema.rb:256:in `block in multiplex'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/schema.rb:560:in `with_definition_error_check'
```

## Performance impact 

There is a small perf impact, here's the output from a benchmark on my machine 

### Without backtrace 

```
~/code/graphql-ruby $ be rake bench:query
               query     2.000  i/100ms
               query     25.646  (± 3.9%) i/s -    128.000  in   5.000069s
~/code/graphql-ruby $ be rake bench:query
               query     2.000  i/100ms
               query     26.298  (± 3.8%) i/s -    132.000  in   5.023957s
~/code/graphql-ruby $ be rake bench:query
               query     2.000  i/100ms
               query     26.349  (± 3.8%) i/s -    132.000  in   5.016218s
```

### With backtrace 

```
~/code/graphql-ruby $ be rake bench:query
               query     2.000  i/100ms
               query     24.853  (± 4.0%) i/s -    124.000  in   5.007307s
~/code/graphql-ruby $ be rake bench:query
               query     2.000  i/100ms
               query     24.750  (± 8.1%) i/s -    124.000  in   5.031750s
~/code/graphql-ruby $ be rake bench:query
               query     2.000  i/100ms
               query     24.201  (± 4.1%) i/s -    122.000  in   5.058803s
```


rmosolgo:
fixes #904

eapache:
IIUC from looking at the usage of `TracePoint` this would intercept and annotate all exceptions within the ruby process? At least for us, this isn't suitable because GraphQL is only one piece of our server. We would much prefer to simply rescue all exceptions just before they "exit" the gem, annotate them there, and then reraise, leaving non-graphql-related exceptions untouched.

rmosolgo:
Yeah, you're totally right 😬  I think I got worked up after reading the `pretty_backtrace` source! I'll refactor this into a `rescue` block which should do the job just fine. (Now, to figure out where that `rescue` block should go ...)

rmosolgo:
Thanks again for the review, @eapache! I took a try at moving to `rescue`, but one problem is tracking the call stack. 

The current implementation uses a tracepoint on method & block calls & returns to maintain a stack of GraphQL context-related objects (Query, Field, etc) which parallels the call stack. Whenever the call stack changes (via call or return), this context stack is modified in the same way. 

This allows us to merge GraphQL context into the error backtrace -- starting at the top of the backtrace, there's one sister entry in the GraphQL context stack. (Eventually, the GraphQL context stack runs out, and there are several unannotated entries in the backtrace.)

The problem is that, when using `begin ... rescue`, the tracepoint registers returns _before_ the error is annotated, so the GraphQL context stack is empty (even though the error backtrace is populated). 

This is demonstrated in 1e4fe0f, which has some "logging" on the stack events. Here it is: 

```
"Push execute_multiplex"
"Push analyze_multiplex"
"Push validate"
"Pop validate"
"Push analyze_query"
"Pop analyze_query"
"Pop analyze_multiplex"
"Push execute_query"
"Push execute_field"
"Push execute_field"
"Pop execute_field"
"Pop execute_field"
"Pop execute_query"
"Annotating This is broken: Boom"
"Pop execute_multiplex"
```

Notice that _most_ of the context is popped before the error is annotated. Because of this, it's _too late_ to annotate the error! 

Contrast this with the tracepoint-based approach's output: 

```
"Push execute_multiplex"
"Push analyze_multiplex"
"Push validate"
"Pop validate"
"Push analyze_query"
"Pop analyze_query"
"Pop analyze_multiplex"
"Push execute_query"
"Push execute_field"
"Push execute_field"
"Annotating This is broken: Boom"
"Pop execute_field"
"Pop execute_field"
"Pop execute_query"
"Pop execute_multiplex"
```

Notice that `Annotating...` comes _before_ the `pop`s, so the context is still present at that point.

What can we do? I can think of two things: 

- Find another way to manage GraphQL context so that it can be re-merged into the error backtrace
- Keep using tracepoint despite the downside (and try to find a more production-ready implementation in the future) (This was done in b6d37e9a)

I think the current implementation is _good enough_ for test and development, what do you think? Or, is there another way to avoid the all-encompassing tracepoint approach?

xuorig:
Small other thing to consider, although maybe not the biggest deal, is that I think some might look at the backtrace for things like exception classification. Modifying the backtrace now possibly becomes a "breaking change"?

rmosolgo:
> I think some might look at the backtrace for things like exception classification

Can you help me understand what you mean by that, perhaps an example? 

Definitely don't want to break tooling!

xuorig:
If we kept the path of execution in the graphql query context, could we add that on top on the stack trace at the end?

like the example you provided in the issue:

```
in field #product
in field #shop
in field #query
... rest of backtrace
```

xuorig:
@rmosolgo for example we rely on backtraces to classify exception reporting into different channels / how it should be handled 🤷‍♂️

rmosolgo:
> could we add that on top on the stack trace at the end?

yes, i think so.

Another option to keep them separate would be to serve `#graphql_backtrace` as a separate method. Then we coudl either add it to the backtrace or add it to the message, something like that, but leave the Ruby backtrace lines alone.

eapache:
Ya, I like the idea of leaving it a separate method on the exception object and let users merge or process otherwise as they see fit.

rmosolgo:
I can think of a few implementations: 

- Monkey-patch `StandardError` (:-1:)
- Rescue an error, then re-raise a `GraphQL::TracedError` (?), and make sure that `err.cause` returns the original exception 
- Stash the GraphQL context someplace else? `GraphQL::Backtrace.current`? 

Or, something else?

eapache:
Wrapping and using `err.cause` makes sense; it's what that field is for.

xuorig:
random ideam: what if `GraphQL::ResolutionError` (any error that happens during resolution) took `ctx` always, and had a `#graphql_backtrace` method? Then all other specific errors can inherit?

eapache:
We'll still need to catch+wrap all the errors (`NameError`, `ArgumentError`) that aren't intentionally raised.

rmosolgo:
Refactored to raise a wrapper error, `GraphQL::Backtrace::TracedError`. 

Currently it: 

- exposes the original error as `cause`
- exposes the GraphQL trace as `graphql_backtrace`
- includes the original error + graphql trace in `.message`
- leaves all Ruby backtraces as-is 

But I think I can do one better. Since #923, we can backtrack through `context` objects, so we could render a trace which includes arguments & returned objects, too. Let me take a crack at that.

lib/graphql/tracing.rb
```diff
@@ -35,24 +37,40 @@ class << self
       # @param metadata [Hash] Event-related metadata (can be anything)
       # @return [Object] Must return the value of the block
       def trace(key, metadata)
-        if @tracer
-          @tracer.trace(key, metadata) { yield }
-        else
-          yield
-        end
+        call_tracer(0, key, metadata) { yield }
```
xuorig: should it be `call_tracers` now?

rmosolgo: oh, i spose so!

rmosolgo:
Started rendering the backtrace as a table, whadd'ya think?

```
GraphQL Backtrace:
Event             | Field                                  | Object     | Arguments           | Result
execute_field     | Thing.raiseField @ [3:13] as boomError | :something | {"message"=>"Boom"} | nil
execute_field     | Query.field1 @ [2:11]                  | nil        | {}                  | {}
execute_query     | query @ [1:9]                          | nil        | {"msg"=>"Boom"}     |
execute_multiplex | query @ [1:9]                          | nil        | {"msg"=>"Boom"}     |
```

In fact, I might depart even further from the current design. 

To render a table like this, we only need `ctx` 😎 

We can also render a table for errors from promises, since we can backtrack using ctx. I'm going to work on it a bit more.

rmosolgo:
Wow, i think it can be refactored to _just_ `rescue`, using backtracking instead of managing the stack...

rmosolgo:
updated the description with the new output

rmosolgo:
sporadic jruby fails :/

rmosolgo:
oops, i need to update the guide for this.

PR #947, opened 2017-09-10
Add validation for Query Name

ashkan18:
# Problem
We should validate `name` of `Query`. Having space and other invalid characters causes issues when using schema with GraphiQL. Doing this we'll also follow GraphQL reference implementation [here](https://github.com/graphql/graphql-js/blob/eb01a23c/src/utilities/assertValidName.js#L11).

# Solution
Added new `GraphQL::NameVlidator` class which can `validate!` a name. Updated `ObjectType` to override `name=` method which first validates using `NameValidator` and then sets `@name`.

Switched `EnumType` to also use the same validator, if you think it's ok for `EnumType` to raise newly added `GraphQL::InvalidNameError` then we can switch it to use `NameValidator.validate!` instead of `NameValidator.valid?`.


lib/graphql/name_validator.rb
```diff
@@ -0,0 +1,17 @@
...
+      unless valid?(name)
+        raise(
+          GraphQL::InvalidNameError,
+          "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but '#{name}' does not"
+        )
+      end
+    end
+
+    def self.valid?(name)
+      name =~ /^[_a-zA-Z][_a-zA-Z0-9]*$/
```
alloy: You can make this a const and then use it in the error message.

```
irb(main):004:0> regexp = /^[_a-zA-Z][_a-zA-Z0-9]*$/
=> /^[_a-zA-Z][_a-zA-Z0-9]*$/
irb(main):005:0> "Names must match #{regexp.inspect} but 'Root Query' does not"
=> "Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'Root Query' does not"
```

spec/graphql/object_type_spec.rb
```diff
@@ -17,6 +17,20 @@
...
+      InvalidNameObject = GraphQL::ObjectType.define do
+        name "Three Word Query"
+
+        field :id, !types.Int, "id field"
+      end
+
+      # Force evaluation
+      InvalidNameObject.name
+    }
+    assert_equal("Names must match (?-mix:^[_a-zA-Z][_a-zA-Z0-9]*$) but 'Three Word Query' does not", exception.message)
```
alloy: If you call `#inspect` on the regexp and interpolate that instead then you won’t get the `(?-mix:…)` part.

ashkan18: Good point, I thought maybe its useful to have it there, but i agree it's cleaner with `inspect`! thanks.

rmosolgo:
:+1: This looks great, thanks for pushing a fix! I totally agree that we should enforce spec-compliant behavior. 

Here are my thoughts:

- I agree that `#inspect` is a more user-friendly output for the error message.
- What do you think about replacing `EnumNameInvalidError` with your new `InvalidNameError`? I think it makes sense to use the same error here and it's ok with me to make this "breaking" change. Also, you could refactor the error message by putting inside `InvalidNameError#initialize`. What do you think?

ashkan18:
thanks @rmosolgo! both of your points make total sense, I've already done the `inspect` change in last commit. 

Going to replace `EnumNameInvalidError` to use new generic `InvalidNameError`, wasn't sure if we should do it since it was breaking but it would be a nice cleanup 👍

rmosolgo:
Very nice, thanks again for making it happen!

ashkan18:
Thank you @rmosolgo and @alloy for taking time and reviewing it 👍

Issue #933, opened 2017-09-01
Reuse for GraphQL::Argument


rmosolgo:
Maybe you want to share an argument definition between fields. 

Using `GraphQL::Function`, you can share arguments, but if the resolve behavior is totally different, Function might not make sense. 

It should be straightforward to add support for reuse via `.define`, eg 

```ruby 
# This already works 
arg = GraphQL::Argument.define do 
  type(types.Int) 
  description("...")
  prepare ->(...) { ... }
end 

field :field1, Type1 do 
  # This is what we need: 
  argument(:my_arg, arg) 
end 
```

`GraphQL::Field` can already be shared this way, it can be used for inspiration!

rmosolgo:
Fixed by #948

PR #948, opened 2017-09-11
Existing argument def

xuorig:
https://github.com/rmosolgo/graphql-ruby/issues/933

Allows defining an argument from an existing argument a bit like it's already possible with fields.

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -96,7 +96,11 @@ module GraphQL
 
     NO_DEFAULT_VALUE = Object.new
     # @api private
-    def self.from_dsl(name, type = nil, description = nil, default_value: NO_DEFAULT_VALUE, as: nil, prepare: DefaultPrepare, **kwargs, &block)
+    def self.from_dsl(name, type_or_argument = nil, description = nil, default_value: NO_DEFAULT_VALUE, as: nil, prepare: DefaultPrepare, **kwargs, &block)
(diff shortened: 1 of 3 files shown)


lib/graphql/argument.rb
```diff
@@ -96,15 +96,19 @@ def prepare=(prepare_proc)
 
     NO_DEFAULT_VALUE = Object.new
     # @api private
-    def self.from_dsl(name, type = nil, description = nil, default_value: NO_DEFAULT_VALUE, as: nil, prepare: DefaultPrepare, **kwargs, &block)
+    def self.from_dsl(name, type_or_argument = nil, description = nil, default_value: NO_DEFAULT_VALUE, as: nil, prepare: DefaultPrepare, **kwargs, &block)
+      if type_or_argument.is_a?(GraphQL::Argument)
+        return type_or_argument.redefine(name: name.to_s)
+      end
+
       argument = if block_given?
```
xuorig: should we handle a block when provided an existing arg? Redefine attributes in the block 🤔 ? That might be confusing.

rmosolgo: It might come in handy, but I'm happy to leave it for later. 

We [do it for fields](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/define/assign_object_field.rb#L37-L41) but it makes more sense there because some things (like `argument`) _must_ be defined in the block.

rmosolgo:
Thanks for taking care of this one!

PR #953, opened 2017-09-15
Update arguments.md

dherault:
Fix smalltypo in docs

current would throw: 
GraphQL::Schema::InvalidTypeError (Query is invalid: field "foo" argument "category" Variable id of type "String!" is required and will not use the default value. Perhaps you meant to use type "String".)


rmosolgo:
Thanks! Oopsie 😊

Issue #959, opened 2017-09-18
graphql-ruby-client doesn't play well with custom authorization strategies


kevinjalbert:
I ran into this situation when setting up [graphql-ruby-client](https://github.com/rmosolgo/graphql-ruby-client) with graphql-pro.

I have a [Custom Authorization Strategy](http://graphql-ruby.org/pro/authorization.html#custom-authorization-strategy) for my schema, which looks for a specific auth bearer token.

When running the `graphql-ruby-client` the sync will fail on the server with the following backtrace:

```
NoMethodError (undefined method `allowed?' for nil:NilClass):

graphql-pro (1.5.3) lib/graphql/pro/access/schema_filter.rb:32:in `call'
graphql (1.6.8) lib/graphql/filter.rb:13:in `call'
graphql (1.6.8) lib/graphql/filter.rb:12:in `call'
graphql (1.6.8) lib/graphql/schema/warden.rb:43:in `block in initialize'
graphql (1.6.8) lib/graphql/schema/warden.rb:138:in `block in read_through'
graphql (1.6.8) lib/graphql/schema/warden.rb:134:in `visible?'
graphql (1.6.8) lib/graphql/schema/warden.rb:120:in `root_type_for_operation'
graphql (1.6.8) lib/graphql/static_validation/rules/mutation_root_exists.rb:8:in `validate'
graphql (1.6.8) lib/graphql/static_validation/validator.rb:34:in `block in validate'
graphql (1.6.8) lib/graphql/static_validation/validator.rb:33:in `each'
graphql (1.6.8) lib/graphql/static_validation/validator.rb:33:in `validate'
graphql-pro (1.5.3) lib/graphql/pro/operation_store/validate.rb:20:in `validate'
graphql-pro (1.5.3) lib/graphql/pro/operation_store/add_operation.rb:14:in `call'
graphql-pro (1.5.3) lib/graphql/pro/operation_store.rb:101:in `add'
graphql-pro (1.5.3) lib/graphql/pro/operation_store/endpoint.rb:40:in `block (2 levels) in call'
graphql-pro (1.5.3) lib/graphql/pro/operation_store/endpoint.rb:37:in `each'
graphql-pro (1.5.3) lib/graphql/pro/operation_store/endpoint.rb:37:in `block in call'
activerecord (5.1.2) lib/active_record/connection_adapters/abstract/database_statements.rb:225:in `block in transaction'
[...]

PR #967, opened 2017-09-20
Fix #961 Expose typename in persisted query sync

### LINKED ISSUE #961: graphql-ruby-client needs to ability to expose `__typename` on persisted queries
I encountered this using `graphql-ruby-client` version 1.0.2.

When using the `sync` command the client ends up saving the GraphQL queries as expected. The main problem is that in my specific React application I'm using [react-apollo](https://github.com/apollographql/react-apollo) with the [`addTypename` option](http://dev.apollodata.com/core/apollo-client-api.html#ApolloClient.constructor) enabled in the `ApolloClient`. This means that I can write my queries without any `__typename` and they get added automatically. The `__typename` is needed for Apollo to cache entities correctly.

Seems like the `graphql-ruby-client` needs an option to automatically add the `__typename` fields before sending the queries to be persisted. On the JS side I saw that this exists for [apollographql/persistgraphql](https://github.com/apollographql/persistgraphql#build-tool-semantics). I imagine that we could implement a similar interface on the CLI?

In my digging it seems like we'd have to add the fields somewhere [here](https://github.com/rmosolgo/graphql-ruby-client/blob/d05fb3bcd849ad76566f3f004698b7a4042b24f3/sync/prepareProject.js#L42-L61)? The AST visitor could probably follow some similar implementation as to how it's done in [`apollo-client`](https://github.com/apollographql/apollo-client/blob/fec6457746b1cb63c759f128349e66499328ae43/src/queries/queryTransform.ts#L22-L56)?

kevinjalbert:
This should fix #961, by adding a new `--add-typename` option on the `graphql-ruby-client` CLI.

We're taking advantage of the work that [apollo-client](https://github.com/apollographql/apollo-client/blob/fec6457746b1cb63c759f128349e66499328ae43/src/queries/queryTransform.ts#L22-L56) has for adding the typenames to a GraphQL AST.

I've added this approach to both the project and isolated approaches.

I've refactored the isolated approach to take advantage of the visitor approach with `graphql`.

I've added and refactored tests in `graphql-ruby-client` to take advantage of [jest's snapshot testing](https://facebook.github.io/jest/docs/en/snapshot-testing.html).

-----

Side note, I haven't actually done a _full end-to-end_ test on my application yet as I was waiting for #959 (which has since been fixed).

Let me know if there is anything you want me to change, or expand on.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.snap linguist-generated
+*.lock linguist-generated
--- a/guides/javascript_client/sync.md
+++ b/guides/javascript_client/sync.md
@@ -41,6 +41,7 @@ option | description
 `--client` | Client ID ({% internal_link "created on server", "/operation_store/client_workflow" %})
 `--secret` | Client Secret ({% internal_link "created on server", "/operation_store/client_workflow" %})
 `--outfile` | Destination for generated JS code
+`--add-typename` | Add `__typename` to all selection sets (for use with Apollo Client)
 
 You can see these and a few others with `graphql-ruby-client sync --help`.
 
--- a/javascript_client/__tests__/__snapshots__/syncTest.js.snap
+++ b/javascript_client/__tests__/__snapshots__/syncTest.js.snap
@@ -1,0 +1,289 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`sync operations Input files Merges fragments and operations across files 1`] = `
+Array [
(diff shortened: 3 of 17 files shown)


rmosolgo:
This PR looks awesome! I love that you applied Jest snapshot testing. I'll take a close look later today and get back to you.

rmosolgo:
Thanks again for taking care of this, I've just released it as `1.1.1` on NPM!

PR #968, opened 2017-09-20
Fix wrong yarn command

cpunion:


rmosolgo:
oops, thanks for fixing it!

PR #976, opened 2017-09-21
Add Apollo Query Batching example to Multiplexing docs

matsimitsu:
In Slack there were a few questions about using Apollo batching with GraphQL-Ruby. 

I _believe_ Apollo batching is the most common use case of multiplexing, so maybe adding an example of how to handle these batches would be a nice addition to the documentation.


rmosolgo:
Thanks for sharing your experience on this one! Yeah, Apollo is the only use case I know of, so it's great to have a specific example. I just updated the site :)

albandiguer:
thanks so much for that

PR #977, opened 2017-09-21
refactor(Query::Arguments) remove intermediary StaticArguments class

rmosolgo:
Oops, I realized I left one race condition in this code, `@args_class ||=`. So I refactored to build the class eagerly instead of lazily. (In fact, I didn't experience a bug from this, but I noticed it while chasing something unrelated.)


rmosolgo:
cc @esjee just FYI, a little change here !

PR #988, opened 2017-09-26
Support custom coercion error messages for scalars

### LINKED ISSUE #981: Custom error messages for scalar's input coercion
Currently there's no way to specify a custom error message from a scalar type's `coerce_input` method and have it returned in the response.

The closest you can get is using `ctx.add_error` but it only appends to the existing default ones.

Use case: we want to provide better feedback for invalid Global IDs. The default is:

```json
    {
      "message": "Argument 'id' on InputObject 'FooInput' has an invalid value. Expected type 'ID!'.",
      "locations": [
        {
          "line": 2,
          "column": 26
        }
      ],
      "fields": [
        "mutation",
        "fooDelete",
        "input",
        "id"
      ]
    }
```

Ideally we could return or raise an `ExecutionError`. Is there a reason why this isn't possible? Or just not implemented yet?

cc @eapache

swalkinshaw:
Closes #981

Error messages for invalid scalar input values were statically set during validation.

This allows for a scalar to raise a `GraphQL::CoercionError` within `coerce_input` and its error message will be used in the error response.

### DIFF
--- a/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
+++ b/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
@@ -6,10 +6,15 @@ module GraphQL
         return if node.value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
         arg_defn = defn.arguments[node.name]
         return unless arg_defn
-        if !context.valid_literal?(node.value, arg_defn.type)
-          kind_of_node = node_type(parent)
-          error_arg_name = parent_name(parent, defn)
-          context.errors << message("Argument '#{node.name}' on #{kind_of_node} '#{error_arg_name}' has an invalid value. Expected type '#{arg_defn.type}'.", parent, context: context)
+
+        begin
+          if !context.valid_literal?(node.value, arg_defn.type)
+            kind_of_node = node_type(parent)
+            error_arg_name = parent_name(parent, defn)
+            context.errors << message("Argument '#{node.name}' on #{kind_of_node} '#{error_arg_name}' has an invalid value. Expected type '#{arg_defn.type}'.", parent, context: context)
+          end
+        rescue GraphQL::CoercionError => err
+          context.errors << message(err.message, parent, context: context)
         end
       end
     end
--- a/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
+++ b/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
@@ -8,14 +8,20 @@ module GraphQL
         return unless arg_defn
 
         begin
-          if !context.valid_literal?(node.value, arg_defn.type)
-            kind_of_node = node_type(parent)
-            error_arg_name = parent_name(parent, defn)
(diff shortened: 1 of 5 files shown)


lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
```diff
@@ -6,10 +6,15 @@ def validate_node(parent, node, defn, context)
...
-          error_arg_name = parent_name(parent, defn)
-          context.errors << message("Argument '#{node.name}' on #{kind_of_node} '#{error_arg_name}' has an invalid value. Expected type '#{arg_defn.type}'.", parent, context: context)
+
+        begin
+          if !context.valid_literal?(node.value, arg_defn.type)
+            kind_of_node = node_type(parent)
+            error_arg_name = parent_name(parent, defn)
+            context.errors << message("Argument '#{node.name}' on #{kind_of_node} '#{error_arg_name}' has an invalid value. Expected type '#{arg_defn.type}'.", parent, context: context)
+          end
+        rescue GraphQL::CoercionError => err
```
eapache: this might be clearer if it *only* wrapped the `valid_literal?` call and then the default message was more obviously a fallback when no message is provided via `CoercionError`?

eapache:
Overall I like this approach, 👍

swalkinshaw:
Pushed a commit with @eapache's idea. I think it's better 👍

rmosolgo:
👌  Love it! Let me just take one more closer look tomorrow or tonight.

rmosolgo:
Thanks for this feature!

ianks:
Great feature! One small issue @swalkinshaw is that it does not play well with graphql backtrace, and ends up dumping a massive query introspection into the console. This issue is created for me by enabling `GraphQL::Backtrace`, then raising a coercion error in a custom scalar type:

```ruby
module GraphQL
  module Types
    NonEmptyString = GraphQL::ScalarType.define do
      name 'NonEmptyString'
      description <<~DESC
        A subset of the `String` type which cannot be blank
      DESC

      coerce_input ->(v, ctx) {
        raise GraphQL::CoercionError, 'string cannot be blank' if v.blank?
        GraphQL::STRING_TYPE.coerce_input(v, ctx)
      }

      coerce_result ->(v, ctx) { GraphQL::STRING_TYPE.coerce_result(v, ctx) }
    end
  end
end
```

`#<RuntimeError: Unexpected get_rows subject [#<GraphQL::Query:0x0055f227f343b8 ....`

For reference, here is the case that is not handled: https://github.com/rmosolgo/graphql-ruby/blob/17a497d0ed4514b7af90ee1abf5e1ce18e9b12a1/lib/graphql/backtrace/table.rb#L115

ianks:
Ok, so the above mentioned only happens when the type is a VariableIdentifier rather than a scalar

swalkinshaw:
@ianks was just coming back to this, and realized I encountered the same thing separately too. See https://github.com/rmosolgo/graphql-ruby/pull/1071 for a fix which is in progress.

PR #990, opened 2017-09-27
createActionCableHandler should handle many subscriptions properly, use eslint and build with babel es6

michaelstockton:
The current action cable implementation creates a new “channel” for each relay requestSubscription call. Internally, action cable tracks each channel which is identified by the channel name “GraphqlChannel”.

When a message is received, js action cable looks for all the local js channels subscribed to “GraphqlChannel” and delivers the server payload to each one.

Unfortunately, this means that for each payload received, it will be sent to the updated callback and observer of every single requestSubscription call.

This change addresses that by asking the client to forward a channel_operations_id to the server. This is similar to how subscriptions-transport-ws works. The server is responsible for sending
the channel operations id back to client. The client uses this id internally to map a specific observer to an operation and set of variables.

I made a few other changes as well. 

Eslint is pretty handy for keeping code quality up. And es6 is a better developer experience.

This is broken into four commits.

1. Building the client with babel
2. The addition of eslint and it's configuration
3. The result of eslint's autofix feature (eslint --fix)
4. A little extra help where --fix doesn't work and disabling eslint where I didn't want to spend time refactoring but the functionality still works.

I'd like to get some feedback on my implementation before I write tests and documentation. And I have a few questions:

1. These changes mean one websocket will only have one instance of a channel. Rather than N channels. What does that mean for the subscription_id? I'm not sure how that would be exposed properly. Is this entirely up to the developer? And how would it be used? 

See: https://github.com/michaelstockton/graphql-ruby/blob/master/lib/graphql/subscriptions/action_cable_subscriptions.rb#L95

2. The current suggested implementation of GraphqlChannel suggests returning the result of the first query. However, from I can tell, the first run is empty. Can you explain your thinking here?

See: https://github.com/michaelstockton/graphql-ruby/blob/master/lib/graphql/subscriptions/action_cable_subscriptions.rb#L44

3. On a websocket disconnection, I'm calling onCompleted on each subscription. I think we could resend the subscription requests, but I'm not sure if that's desired. What do you think?


rmosolgo:
Hi, thanks for all your work on the JS code! As you can see, JavaScript is _not_ my strong suit 😆 

I totally forgot about an `operation_id` 😖 .  I even experimented with ActionCable a while back, and [used an id](https://github.com/rmosolgo/graphql-streaming/blob/master/lib/graphql/streaming/clients/graphql-streaming/graphql_channel.js) but missed it this time! 

So it sounds like all your changes are good, just give me some time to make a good review. Thanks!

rmosolgo:
Sorry for the slow turnaround here! Here are some answers to your questions: 

> 1. One websocket will only have one instance of a channel. Rather than N channels. 

I'm not sure I _quite_ understand this one. But maybe this is what you mean: 

In my previous implementations, I tried "multiplexing" subscriptions over a single channel, so that each subscription query would add a `stream_from`. But I had a problem _closing_ the individual subscriptions, because I couldn't find a way to "un-`stream_from`" a specific broadcasting. Here's my unanswered SO post: https://stackoverflow.com/questions/39000931/how-to-stop-streaming-from-a-specific-broadcasting-in-actioncable

So in this approach, I figured, one channel per subscription query, to avoid that. But evidently I didn't do _that_ right either 😬 

> 2. ... returning the result of the first query.

There are two things I had in mind here: 

- It can return `errors` if there were any 
- The same codepath can be used for queries and mutations, if desired 

Sending queries and mutations over ActionCable can simplify client development. 

> 3. On a websocket disconnection, I'm calling onCompleted 

Do you mean, for example, if a server crashes or is restarted, so the client loses its connection? 

I think in that case we want to re-send the subscription query. That way, the client application can recover from the disconnect. (I _think_ the current implementation will do that because it's ActionCable's default behavior, does that sound right to you?)

rmosolgo:
Thanks again for raising this issue, I added a channel ID to the action cable subscription in #1004: 

https://github.com/rmosolgo/graphql-ruby/pull/1004/files#diff-fe5db0463355d5e6d7b1fdfd7bd436f7
https://github.com/rmosolgo/graphql-ruby/pull/1004/files#diff-631f0a7d0ea99a4d59f03b45ebcca7c9

So, I think the bug should be addressed 😅 

If you're still interested in adding eslint & babel, I'd be happy to check that out in another PR!

michaelstockton:
The problem with one channel per subscription query -- or from the client's perspective -- creating a handler (via createActionCableHandler) for each request is the responses aren't handled correctly.

ActionCable processes payloads by the channel's name (i.e. GraphqlChannel). It doesn't take it account the operation id. As a result, the payload for a subscription is sent to every handler (created by createActionCableHandler). 

Imagine two subscription queries with different types:

`
subscription {
  newPost {
    ...
  }
`

`
subscription {
  newComment {
     ...
  }
}
`

If you're subscribed to both and rails triggers a subscription update for newPost, newComment and newPost's mutation updater will be called with the payload of newPost. 

If we need to unsubscribe from individual queries, I suspect we'll have to implement some protocol within ActionCable. I suppose we could implement the subscription protocol used by apollo. https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L607.

rmosolgo:
Thanks for following up, I have a little test case for end-to-end subscriptions over action cable, so I just added a step to unsubscribe one of them: 

https://github.com/rmosolgo/graphql-ruby/pull/1010

How does that look to you? Does it seem like a good test, or is anything missing there? (Feel free to respond on that other PR)

PR #994, opened 2017-09-30
feat(Tracing) open-source graphql-pro monitoring 

### LINKED ISSUE #991: Pro monitoring loads relations before pagination
In order to measure the database load time, graphql-pro's `monitoring` calls `.load` on relations. But since a refactor to how connections are handled (1.6.0, I think), it can call `.load` _before_ pagination, resulting in big DB queries. 

This can be really bad, since it might load a huge association into memory (and then measure the wrong thing). 

There's not really a work-around except to turn of monitoring for a while. 

My suggested "fix" is to: 

- Reimplement monitoring in graphql-ruby (open-source) 
- Build it on top of `GraphQL::Tracing`

I will look at this soon and update this issue! 😖 

cc @mvgijssel thanks for reporting this

### LINKED ISSUE #993: Pro monitoring doesn't handle multiple queries per controller action
Several of the supported platforms of graphql-pro's monitoring feature don't play nicely if you have multiple GraphQL executions in one controller action (eg, with graphql-client). 

This is because they rename the transaction after the currently-executing query. So, the first query is overridden by the first one.

rmosolgo:
This is a re-implementation of GraphQL-Pro's monitoring, open-source. 

Why? 

- We _need_ it this way: we need more eyes on performance (and how the performance is measured)
- It can be better: this is re-implemented on `GraphQL::Tracing`, which was not available for the previous implementation 
- The previous implementation has bugs (see points above, 😆 )

Fixes #993  
Fixes #991 

TODO: 

- [ ] need schema- & query-local tracers (the test has side-effects ATM)
- [ ] should I include other platforms in the initial release


exAspArk:
Hey, @rmosolgo! Thank you so much for improving and open-sourcing more features in graphql-ruby! 

I'd like to ask you a few questions about `GraphQL::Tracing` which was introduced in https://github.com/rmosolgo/graphql-ruby/issues/896. This is so funny that a few days before we created an open-sourced gem called [graphql-tracing](https://github.com/uniiverse/graphql-tracing), also during the hackathon 😂 The gem is a simple implementation of the [apollo-tracing](https://github.com/apollographql/apollo-tracing) format.

Since the same `GraphQL::Tracing` is already in the `graphql` gem, I think we'd need to rename it to `apollo-tracing` or something like this. But in general, what is your plan for the `GraphQL::Tracing`:

* Stability. Are you planning to change it significantly? It seems like it's an [experimental feature](http://graphql-ruby.org/queries/tracing.html) for now?
* API. In my opinion, ideally, `GraphQL::Tracing` can provide some rich interface for other tools (without `ActiveSupport:: Notifications`, please). So, for example, I could build an `apollo-tracing` format support (or collect and save metrics I want manyally) by using the interface. What do you think?

rmosolgo:
Great questions! Here's what I've got: 

> Stability. Are you planning to change it significantly?

Yes, I have some improvements planned for the next release here: #996 

Eventually, the current (non-thread-safe) implementation will go away.

> API. So, for example, I could build an apollo-tracing format support (or collect and save metrics I want manually) by using the interface. What do you think?

I think this would be great 😊  that's just what I'm hoping to do: provide enough information for comprehensive metrics in whatever format you like. 

The current implementation _supports_ `AS::N` but doesn't depend on it. 🍻  Here's hoping that graphql-ruby _never_ depends on ActiveSupport 😆 . It's a very convenient library, but I respect Ruby programmers who want to avoid all the mess, too!

PR #985, opened 2017-09-26
(WIP) Serilaize/deserialize GlobalID::Identification

cpunion:
Make SubscriptionType's field receive deserialized model object instead of Hash object.

See https://github.com/rmosolgo/graphql-ruby/issues/969

Issue #984, opened 2017-09-25
Improve backtrace printout when entire result is compressed


rmosolgo:
The backtrace table prints the in-progress graphql response for each field in the "Result" column. 
It minifies nested objects to reduce visual noise, eg: 

```ruby
{id: "MDEzOkNvbW1pdENvbW1lbnQ0OA==", author: {...}}
```

However, it's still a bit off when the result is a list of objects, eg: 

```ruby
[{...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}, {...}]
```

I think that this should be printed as simply 

```ruby 
[...]
```

Or, if someone has a better suggestion, I'd love something even _better_ :)

rmosolgo:
This would be fine but nobody else seems to care and I'm never going to get around to it.

PR #997, opened 2017-10-02
Serialize Subscription Scopes

awmichel:
Adds support for serializing the `subscription_scope` used for GraphQL subscriptions. The new `serialize_scope` method is based on [`broadcasting_for`](https://github.com/rails/rails/blob/012196bbc78cf1c9243d0c2ceaa1e509a9a34a11/actioncable/lib/action_cable/channel/broadcasting.rb#L18-L27) used by ActionCable for a similar purpose.

This solves a bug when using an ActiveRecord model object to scope the subscription. `GraphQL::Subscriptions::Event.serialize` calls `JSON.dump([scope, name, sorted_h])` which appears to call `to_s` on each of the array elements. In the case of `scope`, it returns a string like `"#<User:0x00000005e01c80>"` containing the `object_id` and breaking scoped subscriptions.

Example:
```ruby
  subscription(SubscriptionType)
end

SubscriptionType = GraphQL::ObjectType.define do
  field :activityFeed, ActivityType do
    subscription_scope :current_user
  end
end

context = {
  current_user: User.find(1)
}

Schema.execute(query: 'subscription { activityFeed() { id } }', context: context)
```

### DIFF
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -22,7 +22,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "benchmark-ips"
   s.add_development_dependency "codeclimate-test-reporter", "~>0.4"
   s.add_development_dependency "concurrent-ruby", "~>1.0"
-  s.add_development_dependency "globalid", "~> 0.4"
   s.add_development_dependency "guard", "~> 2.12"
   s.add_development_dependency "guard-bundler", "~> 2.1"
   s.add_development_dependency "guard-minitest", "~> 2.4"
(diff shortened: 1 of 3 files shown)


rmosolgo:
Hi! Thanks for paying some attention to this. I really like the direction your patch is going. We have a similar improvement underway (for the object, not the scope) in #984. 

I'll make a closer review of these patches soon and release a new version with better subscription support. 

It looks like this will need a little more work to get the tests passing, since GlobalID is supported on Rails 4.2+, but we test on Rails 3 and we test without Rails at all. Feel free to take a look at the configuration for that, or wait until after I give it another review :)

awmichel:
Happy to help, subscription support landed at just the right time to start using in a new project.

I fixed the broken test builds. I didn't see any strict Rails test files, so I simply wrapped the relevant specs in a conditional similar to `execute_spec.rb`.

rmosolgo:
That looks great, thanks! I'm going to pull your commits and see about working them in with #985

rmosolgo:
Thanks again for your work on this one, I cherry-picked your commits and merged them in #1004 !

PR #1002, opened 2017-10-04
Add options parameter to GraphQL::ExecutionError

gfx:
There are cases where I'd like to add an arbitrary data structure to error objects, like `{ "code": "payment_required" }`.

What do you think of it?

### DIFF
--- a/lib/graphql/execution_error.rb
+++ b/lib/graphql/execution_error.rb
@@ -11,8 +11,12 @@ module GraphQL
     # response which corresponds to this error.
     attr_accessor :path
 
-    def initialize(message, ast_node: nil)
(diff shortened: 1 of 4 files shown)


rmosolgo:
:+1: I like the idea! Thanks for the suggestion and the patch. Could you add a test for it?

gfx:
@rmosolgo Thanks. I'll add tests tonight!

gfx:
Added tests: 92470c43

rmosolgo:
That's great, thanks! I'll be releasing a new version soon and this is a really nice feature.

gfx:
Thanks!

 I'd love to use it in my project!

PR #1003, opened 2017-10-04
Don't propogate nil to sibling query selections

esjee:
The problem we ran into is that we want to provide as much data as possible, even when there are errors. For example, given this query:

```
query {
  resourceA {
    id
  }
  resourceB {
    id
  }
}
```

and provided that the response indicates that both fields are not nullable, then graphql-ruby will currently seemingly give up on resolving `resourceB` when resolving `resourceA` is somehow nil.

What seems to be going is that whenever a node being resolved results in a `nil` being put in a non-nullable field (-> no valid result possible), it marks the parent as having `received_null_child`. When continuing resolving the siblings, specifically in `continue_resolve_field`, this causes the following early-exit:

```lang=ruby
          if field_ctx.parent.invalid_null?
            return nil
          end
```

So, reusing the example, graphql-ruby evaluate `resourceA`, this results in a nil, and the parent (`query`) is marked with `received_null_child`. Next, when evaluating `resourceB`, graphql-ruby checks that the parent (again, `query`) has received `invalid_null?`. This evaluates to `true`, and causes `resourceB` to be `nil` as well.

We "solved" this problem by making all top-level types nullable. However, we still thought this behavior was a bit puzzling, so wanted to bring it up here to see if it's helpful to make a change here!

cc @mvgijssel

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -219,9 +219,6 @@ module GraphQL
           when GraphQL::Execution::Execute::PROPAGATE_NULL, nil
             @invalid_null = true
             @value = nil
-            if @type.kind.non_null?
-              @parent.received_null_child
-            end
           when GraphQL::Execution::Execute::SKIP
             @parent.skipped = true
             @parent.delete(self)


esjee:
I still have to fix the dozen or so failing specs, but I figured it'd be worth putting this out there anyway, just to see if anybody has any strong thoughts on the matter.

rmosolgo:
The current behavior is an attempt to implement this part of the spec: 

> Since Non-Null type fields cannot be null, field errors are propagated to be handled by the parent field. If the parent field may be null then it resolves to null, otherwise if it is a Non-Null type, the field error is further propagated to it’s parent field.

(http://facebook.github.io/graphql/October2016/#sec-Errors-and-Non-Nullability)

Do you think that the current behavior matches the spec? Or, given your query above, what _should_ be the response?

esjee:
What currently happens is a response like `data: null`. What I would like is something like

```
data: {
  resourceA: null,
  resourceB: {
    id: 37,
  }
}
```

rmosolgo:
What's the return type of `Query.resourceA` in that case? (Is it a nullable type, or a non-null type?)

esjee:
Both are non-null.

rmosolgo:
(If `Query.resourceA` is a _nullable_ type, then it should work the way you described. If it's a _non-null_ type, then the response you described would be a violation of the GraphQL type system, since a non-null field ought not to return `null`)

esjee:
I suppose that's fair enough, but I don't quite see how it makes sense that if resourceA is invalid, that resourceB should also be seen as invalid. Is this part of the spec, or is this an optimization?

rmosolgo:
It's part of the spec, that when a non-null field returns null, it's considered an error at the _parent_ level, so sibling fields are also removed, see: 

http://facebook.github.io/graphql/October2016/#sec-Errors-and-Non-Nullability

Here's an example of the same behavior using the JS implementation: 

https://launchpad.graphql.com/8rmlq94kq

For example, if you make a request _without_ invalid null, you get a response with data: 

```graphql 
{
  # This field returns invalid null,
  # uncomment it to see null propagation behavior
  # resourceA { id }
  resourceB { id }
}

# => 
# {
#   "data": {
#     "resourceB": {
#       "id": "5"
#     }
#   }
# }
```

But, if you include the invalid null, the parent field (root, in this case) is nulled out: 

```graphql 
{
  # This field returns invalid null,
  # uncomment it to see null propagation behavior
  resourceA { id }
  resourceB { id }
}

# => 
# {
#   "errors": [
#     {
#       "message": "Cannot return null for non-nullable field Query.resourceA.",
#       "locations": [
#         {
#           "line": 6,
#           "column": 3
#         }
#       ],
#       "path": [
#         "resourceA"
#      ]
#     }
#   ],
#   "data": null
# }
```

I think the _reason_ is so that GraphQL can support type-safe clients (eg, native apps in Objective-C or Java). GraphQL endpoints _promise_ that you won't get any unexpectedly-null values; only nullable fields can return null. So this is how that guarantee is implemented!

esjee:
Ahh, now I see. Thanks a lot for the explanation @rmosolgo, that makes sense! If you take typed languages into account, I can see how this requirement makes it easier to write libraries in those languages.

mvgijssel:
Thanks for the explanation @rmosolgo! Makes sense now 👍

PR #1004, opened 2017-10-04
Fix subscription serialization

rmosolgo:
This takes commits from #997 and #985, colocates some serialization helpers, and adds tests for ActionCable behavior. 

TODO: 

- [x] finish tests for ActionCable


rmosolgo:
:tada: nice, a running system test! 

I added a Rails 5 app to do end-to-end testing with ActionCable. 

I'd love to test `graphql-ruby-client` here too, but maybe I'll do that in a separate PR

Issue #992, opened 2017-09-27
how to return rails routes in resolver


vegetabill:
Hey there it's Bill from Goodreads. I remember you from [various issues](https://github.com/reactjs/react-rails/issues/212). This is a really awesome gem I'm trying out.

My question is I want to expose route attributes, e.g. I want `UserType` to have a `profile_url` that is the web URL for viewing a user's profile either in a browser of if they have a native app installed, it will deep link into that.

What is the best way to do this? I naively tried this:
```ruby
Types::UserType = GraphQL::ObjectType.define do
  include Rails.application.routes.url_helpers
  field :profile_url, types.String do
      description 'web show url'
      resolve -> (user, args, ctx) { user_url(user) }
  end
end
```

I tried adding `include Rails.application.routes.url_helpers` in the `define` body of the type, but that is not in scope during actual execution. Whatever the right way to do it is, having an example in the docs would be great and I'd be happy to add one as a PR as long as it's the way you think it would best be done.

robak86:
Not sure if it is the best answer, but I guess you can create reference to routes helper inside ```ctx```.

```ruby
Types::UserType = GraphQL::ObjectType.define do
  field :profile_url, types.String do
      description 'web show url'
      resolve -> (user, args, ctx) { ctx.routes.user_url(user) }
  end
end

MySchema.execute(
  params[:query],
  variables: params[:variables],
  context: { routes: Rails.application.routes.url_helpers },
)
```

rmosolgo:
:+1: to the solution above, of course, there's a "naughty" way to accomplish the same thing by reaching for the global constant inside `resolve`, eg: 

```ruby 
resolve -> (user, args, ctx) { Rails.application.routes.url_helpers.user_url(user) }
```

It's probably _not_ a better way of doing it, but it makes the relationship a bit clearer :P

rmosolgo:
> `include` ... not in scope during actual execution

this is such a good thing to have tried, I'm sorry it didn't work ... I regret using `instance_eval` 😖 

A PR to the docs would be great :)

vegetabill:
Thanks for the info. I'll make a doc PR soon!

PR #1005, opened 2017-10-04
added FAQ doc page to have tips such as the question in #992

vegetabill:
added doc in #992

wasn't sure what category it belonged in so made a new FAQ page ?

Tested locally using `bundle exec rake site:serve `


rmosolgo:
This looks great! I wonder if the new page will attract some more suggestions :) Sorry for the late review on this one!

PR #1006, opened 2017-10-05
Explore RDL 

rmosolgo:
I'm a lover of type systems, so I thought I would try [RDL](). 

This is just an experiment and could never be merged UNTIL: 

- [ ] The perf hit for development is not bad (currently a 300% slowdown) 
- [ ] We can be sure that gem users won't be affected 
- [ ] It's worth the value 

Personally, I'm not interested in annotating types _twice_, once for YARD and again for RDL. But I think it might be possible to _generate_ types for RDL from YARD. But the first step is to try RDL.


rmosolgo:
Oh I forgot, I had to patch my local copy of the gem for it to work due to this issue: 

https://github.com/plum-umd/rdl/issues/59

rmosolgo:
Fun, maybe try again later sometime!

PR #1008, opened 2017-10-06
Add section on using Rack::Auth::Basic for UI

kevinjalbert:
A quick approach for locking down the provided graphql-pro's UI is to
use basic auth. Adding a section to the `Authorizing the UI` section to
present both the Rails Routing Constraints and Rack Basic Auth.


rmosolgo:
Very cool, TIL! Thanks for improving the docs :)

PR #1010, opened 2017-10-09
test(ActionCableSubscriptions) test partial unsubscribe

rmosolgo:
I'm still getting the hang of how to do subscriptions over ActionCable. 

This adds a new step to the test case: after making two subscriptions and triggering them separately, also unsubscribe one of them. Make sure the unsubscribed channel doesn't get more updates, but the still-subscribed one does.

### DIFF
--- a/spec/dummy/app/views/pages/show.html
+++ b/spec/dummy/app/views/pages/show.html
@@ -1,6 +1,6 @@
 <h1>ActionCable Test Page</h1>
-<button onClick="sub1.trigger()" id="trigger-1">Trigger 1</button>
-<button onClick="sub2.trigger()" id="trigger-2">Trigger 2</button>
+<button onClick="sub1.trigger()">Trigger 1</button>
+<button onClick="sub2.trigger()">Trigger 2</button>
 <p>ActionCable updates:</p>
 <ul id="updates-1">
 </ul>
@@ -8,6 +8,8 @@
 <ul id="updates-2">
 </ul>
 
+<button onClick="sub1.unsubscribe()">Unsubscribe 1</button>
+<button onClick="sub2.unsubscribe()">Unsubscribe 2</button>
 <script>
 var sub1 = App.subscribe("updates-1")
 var sub2 = App.subscribe("updates-2")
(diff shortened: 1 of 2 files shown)


michaelstockton:
The problem that I referred to in #990 has to do with having multiple, active subscriptions. I think you're assuming that there's one, active subscription (or GraphqlChannel) a time. With one active subscription, there's no problem. But with multiple subscriptions -- even if they're the same field on the SubscriptionType -- you will have an issue with updater functions being called multiple times.

Checkout this code path in action cable:
https://github.com/rails/rails/blob/master/actioncable/app/assets/javascripts/action_cable/connection.coffee#L95
https://github.com/rails/rails/blob/master/actioncable/app/assets/javascripts/action_cable/subscriptions.coffee#L57

When a message event is received, action cable calls notify for "GraphqlChannel". Well if you're js client created multiple channels for graphql, it will receive the payload for each of them. As a result, any subscription created by the client will receive the payload.

When I wrote #990, I didn't spend a lot of time thinking about unsubscribing and it was an oversight on my part. I think having multiple channels makes sense server-side because unsubscribing is explicit, but it causes problems on the client with processing responses and vice-versa. 

Let me know if this isn't making sense. Happy to explain more. Thanks for all your work, too!

rmosolgo:
Thanks for those links! I really still don't _quite_ understand it. Is there any chance you could "say it with code" by adding a failing test to this project? 

The reason I don't understand is because the dummy app in this project seems to behave correctly. You can start it locally: 

```
cd spec/dummy 
BUNDLE_GEMFILE=../../gemfiles/rails_5.1.gemfile rails s
```

Then take it for a spin:
 
action-cable-test

In that example, it looks like the two subscriptions function independently. When you click "Trigger 1", the first subscription receives an update. When you click "Trigger 2", the second subscription receives an update. Does that look right to you?

I _think_ the reason it works is because the two subscriptions have different `identifier`s: 

image

So am I testing the wrong thing? Or is there another test case which would _fail_ if we add it to this project?

michaelstockton:
Ah! Ok, I see now. I thought the channel name had to be a string. Guess not.

https://github.com/rails/rails/blob/master/actioncable/app/assets/javascripts/action_cable/subscriptions.coffee#L15

Okay, this'll work then. I will rebase my work on top of yours and submit a patch to migrate to es6.

rmosolgo:
Cool, thanks again for the review :)

PR #1011, opened 2017-10-09
remove redundant 'if after'

bgentry:
This is already inside an `if after` block so this condition is redundant and always evaluates to true.

Unless there's some side effect I'm missing 😃


rmosolgo:
Probably just overlooked during review of a [previous refactor](https://github.com/rmosolgo/graphql-ruby/commit/c262ac984fe4f60bc88eadf66ff09d112fc5da77). Postfix `if` is such a foot-gun 😩 

Thanks for the fix!

PR #1014, opened 2017-10-09
WIP: add a StableRelationConnection for Sequel

bgentry:
This adds value-based cursors for stable keyset pagination over `Sequel::Dataset`s. It is a work in progress and does not yet have test coverage. Some of the code is kinda 😨

In my own app, I pair this with an encoder which dumps the value hash to JSON and then url-safe Base64 encodes it. There may be a more efficient way to do this (especially wrt resulting cursor string size). Ideally this would be paired with an encrypted encoder as well to prevent tampering of cursors.

Notably, this adds your model's primary key as a final sort criteria (if it's not already in the existing `order` list) to ensure a unique sort order. Not sure if I chose the right way to do this, as relying on the dataset's `model` means this only works with `Sequel::Model`-based datasets.

I'm sure others have ideas on how to improve this so please chime in :v:

### DIFF
--- a/lib/graphql/relay/stable_relation_connection.rb
+++ b/lib/graphql/relay/stable_relation_connection.rb
@@ -1,0 +1,177 @@
+# frozen_string_literal: true
+
+module GraphQL
+  module Relay
+    # A connection implementation to expose SQL collection objects using stable
+    # keyset pagination. All columns used in the sort order will have their
+    # values saved in the resulting cursor. This should probably be paired with
+    # an encrypted or at least signed/verified cursor encoder.
+    #
+    # It currently works only for Sequel::Dataset.
+    class StableRelationConnection < GraphQL::Relay::BaseConnection
+      def cursor_from_node(item)
+        # TODO(bgentry): in the future this should use an encrypted cursor encoder,
+        # possibly from GraphQL Pro.
+        cursor_data = order_fields.map { |f, _result| item.public_send(f) }
+        encode(cursor_data)
+      end
+
+      def has_next_page
+        if first
+          paged_nodes_length >= first && sliced_nodes_count > first
+        elsif GraphQL::Relay::ConnectionType.bidirectional_pagination && before
+          !nodes_with_pk.seek(value: cursor_selectors(before), include_exact_match: true).empty?
+        else
+          false
+        end
+      end
+
+      def has_previous_page
+        if last
+          paged_nodes_length >= last && sliced_nodes_count > last
+        elsif GraphQL::Relay::ConnectionType.bidirectional_pagination && after
+          # Check if we've already paginated through the collection a bit and there
+          # are nodes behind us.
+          !nodes_with_pk.seek(value: cursor_selectors(after), include_exact_match: true).empty?
+        else
+          false
+        end
+      end
+
+      def first
+        return @first if defined? @first
+
+        @first = get_limited_arg(:first)
+        @first = max_page_size if @first && max_page_size && @first > max_page_size
+        @first
+      end
+
+      def last
+        return @last if defined? @last
+
+        @last = get_limited_arg(:last)
+        @last = max_page_size if @last && max_page_size && @last > max_page_size
+        @last
+      end
+
+      private
+
+      # apply first / last limit results
+      def paged_nodes
+        return @paged_nodes if defined? @paged_nodes
+
+        items = sliced_nodes
+
+        if first
+          if relation_limit(items).nil? || relation_limit(items) > first
+            items = items.limit(first)
+          end
+        end
+
+        if last
+          if (relation_limit(items) && last <= relation_limit(items)) || !relation_limit(items)
+            items = items.limit(last)
+          end
+          primary_key = items.model.primary_key
+          # reverse the order again via a subselect *after* the limit is applied so that
+          # we still get the last N nodes, but in their original order:
+          items = items.unfiltered.unlimited.where(
+            primary_key => items.select(primary_key)
+          ).reverse
+        end
+
+        if max_page_size && !first && !last
+          if relation_limit(items).nil? || relation_limit(items) > max_page_size
+            items = items.limit(max_page_size)
+          end
+        end
+
+        @paged_nodes = items
+      end
+
+      def relation_offset(relation)
+        relation.opts[:offset]
+      end
+
+      def relation_limit(relation)
+        relation.opts[:limit]
+      end
+
+      # If a relation contains a `.group` clause, a `.count` will return a Hash.
+      def relation_count(relation)
+        count_or_hash = relation.count
+        count_or_hash.is_a?(Integer) ? count_or_hash : count_or_hash.length
+      end
+
+      # Apply cursors to edges
+      def sliced_nodes
+        return @sliced_nodes if defined? @sliced_nodes
+
+        @sliced_nodes = nodes_with_pk
+        @sliced_nodes = append_conditions(@sliced_nodes, cursor_selectors(after), false) if after
+        @sliced_nodes = append_conditions(@sliced_nodes, cursor_selectors(before), true) if before
+        @sliced_nodes
+      end
+
+      def nodes_with_pk
+        return @nodes_with_pk if defined? @nodes_with_pk
+
+        @nodes_with_pk = nodes
+        primary_key = nodes.opts[:model].primary_key
+        unless order_has_pk?(nodes, primary_key)
+          @nodes_with_pk = nodes_with_pk.order_append(primary_key)
+        end
+        @nodes_with_pk
+      end
+
+      def order_has_pk?(sliced_nodes, pk)
+        all_order_fields = sliced_nodes.opts[:order].map do |i|
+          i.is_a?(Sequel::SQL::OrderedExpression) ? i.expression : i
+        end
+        all_order_fields.include?(pk)
+      end
+
+      def limit_nodes(sliced_nodes, limit)
+        limit > 0 ? sliced_nodes.limit(limit) : sliced_nodes.where(false)
+      end
+
+      def sliced_nodes_count
+        return @sliced_nodes_count if defined? @sliced_nodes_count
+
+        # If a relation contains a `.group` clause, a `.count` will return a Hash.
+        @sliced_nodes_count = relation_count(sliced_nodes)
+      end
+
+      def cursor_selectors(cursor)
+        decode(cursor)
+      end
+
+      def paged_nodes_array
+        return @paged_nodes_array if defined?(@paged_nodes_array)
+        @paged_nodes_array = paged_nodes.to_a
+      end
+
+      def paged_nodes_length
+        paged_nodes_array.length
+      end
+
+      def order_fields
+        return @order_fields if defined? @order_fields
+
+        sliced_nodes.opts[:order].map do |i|
+          i.is_a?(Sequel::SQL::OrderedExpression) ? i.expression : i
+        end
+      end
+
+      # add where conditions for each cursor selector in the appropriate direction
+      def append_conditions(sliced_nodes, values, reversed)
+        if reversed
+          # reverse the order so we can select the last N nodes:
+          sliced_nodes.reverse.seek(value: values)
+        else
+          sliced_nodes.seek(value: values)
+        end
+      end
+    end
+  end
+end


bgentry:
Actually I think the way I'm chaining `where()` conditions is all broken beyond the simple case of just sorting by ID. Looks like this can be quite a bit more complicated (see [sequel-seek-pagination](https://github.com/chanks/sequel-seek-pagination))

bgentry:
ActiveRecord also has equivalent functionality in [order_query](https://github.com/glebm/order_query). Is the right answer to make this connection dependent on one of those two gems?

Both of them make forward pagination straightforward with `seek()`, but don't have anything built-in to support reverse / bidirectional pagination. Maybe that can still be handled with `reverse`?

bgentry:
I updated this to rely on sequel-seek-pagination, and afaict the behavior is actually correct now (for full bidirectional pagination).

One issue is that I'm seeing the actual pagination query execute twice on my database. The first of these is triggered where I would expect it to be, by the `paged_nodes_array` function:

<details>

```
SELECT * FROM "items" WHERE (("name", "id") > ('item fb97', '14186009-9608-4431-9cf0-3424afaecb10')) ORDER BY "name", "id" LIMIT 20
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/sequel_pg-1.7.1/lib/sequel_pg/sequel_pg.rb:77:in `each'
/Users/bgentry/my-api/app/graphql/relay_cursor_connection.rb:133:in `to_a'
/Users/bgentry/my-api/app/graphql/relay_cursor_connection.rb:133:in `paged_nodes_array'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/relay/base_connection.rb:122:in `start_cursor'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field/resolve.rb:31:in `public_send'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field/resolve.rb:31:in `call'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field.rb:230:in `resolve'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:253:in `call'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/schema/middleware_chain.rb:45:in `invoke_core'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/schema/middleware_chain.rb:38:in `invoke'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:107:in `resolve_field'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:71:in `block (2 levels) in resolve_selection'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/tracing.rb:40:in `block in trace'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/tracing.rb:71:in `call_tracers'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/tracing.rb:40:in `trace'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:70:in `block in resolve_selection'
```
</details>

However, it gets called again shortly thereafter by EdgesInstrumentation#build-edges:

<details>

```
SELECT * FROM "items" WHERE (("name", "id") > ('item fb97', '14186009-9608-4431-9cf0-3424afaecb10')) ORDER BY "name", "id" LIMIT 20
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/sequel_pg-1.7.1/lib/sequel_pg/sequel_pg.rb:77:in `each'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/sequel-5.1.0/lib/sequel/dataset/actions.rb:441:in `map'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/sequel-5.1.0/lib/sequel/dataset/actions.rb:441:in `map'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/sequel_pg-1.7.1/lib/sequel_pg/sequel_pg.rb:35:in `map'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/relay/edges_instrumentation.rb:51:in `build_edges'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/relay/edges_instrumentation.rb:37:in `resolve'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field.rb:230:in `call'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field.rb:230:in `resolve'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:253:in `call'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/schema/middleware_chain.rb:45:in `invoke_core'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/schema/middleware_chain.rb:38:in `invoke'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:107:in `resolve_field'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:71:in `block (2 levels) in resolve_selection'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/tracing.rb:40:in `block in trace'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/tracing.rb:71:in `call_tracers'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/tracing.rb:40:in `trace'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/execution/execute.rb:70:in `block in resolve_selection'
```
</details>

I'm not sure if this would somehow be caused by how I've set up this connection on my QueryType:

```ruby
  connection :items, Types::ItemType.connection_type, max_page_size: 100 do
    argument :ids, types[!Types::IDType], "The IDs of the items"
    resolve ItemsResolver
  end
```

Other than that one issue, I believe this is ready for discussion.

bgentry:
I traced that extra SQL query a bit deeper, and found that it's triggered by this part of BaseConnection:

```ruby
      # These are the nodes to render for this connection,
      # probably wrapped by {GraphQL::Relay::Edge}
      def edge_nodes
        @edge_nodes ||= paged_nodes
      end
```

<details>
```
/Users/bgentry/my-api/app/graphql/relay_cursor_connection.rb:54:in `paged_nodes'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/relay/base_connection.rb:102:in `edge_nodes'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field/resolve.rb:31:in `public_send'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field/resolve.rb:31:in `call'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/relay/edges_instrumentation.rb:33:in `resolve'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field.rb:230:in `call'
/Users/bgentry/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/graphql-1.7.3/lib/graphql/field.rb:230:in `resolve'
```
</details>

Because it accesses the underlying relation via `paged_nodes`, the query is re-resolved. AFAICT this is an issue with the existing RelationConnection too. In both cases that value should be fetched from `paged_nodes_array` when [this `EdgesInstrumentation` call to `map`](https://github.com/rmosolgo/graphql-ruby/blob/eca8a8651ee4d6d4841f4751481e28862bad40b5/lib/graphql/relay/edges_instrumentation.rb#L50-L52) is made.

Though I'm unsure where is the best place to accomplish that change.

rmosolgo:
Thanks for opening this issue! I like the idea of doing a similar implementation for ActiveRecord, drawing from my previous implementation in graphql-pro and the one I'm working with now at GitHub. 

I will need some time to review this carefully, but I have a couple of big-picture questions to start off: 

- `RelationConnection` has been difficult to maintain for _both_ Sequel and ActiveRecord. I think it would be better to maintain separate implementations for the two platforms, do you agree? 
- I want to make sure we support SQLite, MySQL and Postgres. When it comes time to test, we should make sure that our new implementation works on all of those. (In graphql-pro, I had a _nightmare_ getting cross-DB-compatible queries when ordering by joined columns 😩 )

bgentry:
> `RelationConnection` has been difficult to maintain for both Sequel and ActiveRecord. I think it would be better to maintain separate implementations for the two platforms, do you agree?

I think that will be much easier to comprehend, yes. It's pretty tricky to get this right as it is, and trying to abstract away the ORM will make that tougher.

In just this Sequel version, it was pretty tough to get backwards pagination working right, particularly the part about fetching the last N rows but returning them in the regular order (so that pages are the same in both directions). I ended up using a subquery to get this working right.

> I want to make sure we support SQLite, MySQL and Postgres. When it comes time to test, we should make sure that our new implementation works on all of those. (In graphql-pro, I had a nightmare getting cross-DB-compatible queries when ordering by joined columns 😩 )

Unfortunately I'm not sure I can help too much with SQLite and MySQL. GitHub has plenty of MySQL expertise though :wink:

It seems like sequel-seek-pagination is primarily designed for Postgres, though on the AR side order_query appears to be more compatible.

The queries end up looking something like this in the simple case:

```sql
-- forward pagination:
SELECT * FROM "items" WHERE (
  ("items"."org_id" = '68e3efe6-a42a-4e0d-a855-f5fb87a52923') AND
  (("name", "id") > ('Item 648b', 'dbb02294-ddb8-4403-a862-5b63e02fb6dc'))
) ORDER BY "name", "id" LIMIT 5;


-- backward pagination:
SELECT * FROM "items"
WHERE (
  "id" IN (
    SELECT "id" FROM "items" WHERE (
      ("items"."org_id" = '68e3efe6-a42a-4e0d-a855-f5fb87a52923') AND
      (("name", "id") < ('Item fb97', '14186009-9608-4431-9cf0-3424afaecb10'))
    )
    ORDER BY "name" DESC, "id" DESC
    LIMIT 5
  )
)
ORDER BY "name" ASC, "id" ASC;
```

bgentry:
@rmosolgo should I open a separate issue for the double-query issue I encountered above? AFAICT it exists on the current codebase and we shouldn’t need to run the same SQL query twice during a request.

rmosolgo:
Oh, sure, if you've found a bug, please open an issue!

rmosolgo:
Hi, I don't really have plans to maintain stable connections as part of this project. It inevitably becomes _really_ complicated 😅 . I think it would be a great fit for an external gem, though! Please let me know if you run into any trouble with implementing it that way.

Aryk:
@bgentry 

I've been using your StableRelationConnection as I've been developing my project. I've found a [few bugs](https://gist.github.com/sarahhenkens/bb0e29afe5b104291e03606326028c62) if you scroll down. 

I'm curious about why you force the primary key sorting. Is that in the spec somewhere? I understand you are trying to make the order deterministic, but ultimately, wouldn't it be less magical and more flexible if you made that the responsibility of the person generating the query?

For example, if they are on a uniq field, your plugin doesn't know that, and then adds a second column to sort by, which could potentially have some effect on performance for large tables.

just curious about your rational on that.

PR #1015, opened 2017-10-10
Make dummy app work

andersklenke:
I wanted to try out the Dummy app inside the `spec` folder, but I wasn't able to run `bundle exec rails server` due to the missing gems, and because Puma was required when it shouldn't.


rmosolgo:
Awesome, thanks for upstreaming this fix! I was always cheating with `BUNDLE_GEMFILE=gemfiles/rails_5.1.gemfile` 😬  But this is much better!

PR #1021, opened 2017-10-12
CI against Ruby 2.4.2

284km:
Ruby 2.4.2 has been released.

https://www.ruby-lang.org/en/news/2017/09/14/ruby-2-4-2-released/


rmosolgo:
Thanks for keeping us up-to-date 😊

PR #1022, opened 2017-10-12
Rename Backtrace::InspectResult.inspect to .inspect_result

jahfer:
### Summary

While running the Shopify codebase through [Mirrors](https://github.com/Shopify/mirrors), I hit a snag that caused it to blow up when trying to `inspect` the `GraphQL::Backtrace::InspectResult` module. This is because that module's `inspect` definition has an arity of 1, where the assumed arity of inspect is always 0.

This PR renames `Backtrace::InspectResult.inspect` to `Backtrace::InspectResult.inspect_result`, so that the default definition of `inspect` still exists on the object.

### Details

The [current implementation](https://github.com/rmosolgo/graphql-ruby/blob/b9433c3f5ce1554a2b37329d3eca467c02f066ba/lib/graphql/backtrace/inspect_result.rb#L8) of `Backtrace::InspectResult.inspect` (introduced in https://github.com/rmosolgo/graphql-ruby/pull/946) accepts an object to interpret and stringify, but calling `inspect` should return a human-readable string of the invoked object, not its arguments.

Ruby encourages the overriding of this method, but `Backtrace::InspectResult.inspect` was never intended to replace the default definition of `inspect` since it provides new functionality with a different arity.

@rmosolgo @eapache for review/comments 🙏

### DIFF
--- a/lib/graphql/backtrace/inspect_result.rb
+++ b/lib/graphql/backtrace/inspect_result.rb
@@ -5,7 +5,7 @@ module GraphQL
     module InspectResult
       module_function
 
-      def inspect(obj)
+      def inspect_result(obj)
         case obj
         when Hash
           "{" +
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for the context, that sounds like a good move. Cool project btw, I would love to learn more about it!

rmosolgo:
sorry for the slow turn-around, finally released in 1.7.5!

PR #1025, opened 2017-10-14
Improve generated controller code

derekprior:
* `ensure_hash` was used once to parse the variables for a query. This
  logic was moved into a `variables` method and simplified some through
  the use of `fetch` to default `nil` variables to a hash and by
  eliminating the explicit `raise` case from the switch statement. In
  normal operation, that case should never be hit.
* For the sake of balance, the other parts of the query were also pulled
  into their own private methods. This is particularly useful for
  `context` which is the most likely to change.
* The resulting `execute` action is reformatted to obey common long-line
  linters.

### DIFF
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -1,33 +1,37 @@
 class GraphqlController < ApplicationController
   def execute
-    variables = ensure_hash(params[:variables])
-    query = params[:query]
-    operation_name = params[:operationName]
-    context = {
-      # Query context goes here, for example:
-      # current_user: current_user,
-    }
-    result = <%= schema_name %>.execute(query, variables: variables, context: context, operation_name: operation_name)
+    result = <%= schema_name %>.execute(
+      query,
+      variables: variables,
+      context: context,
+      operation_name: operation_name,
+    )
+
     render json: result
   end
 
   private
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi, thanks for taking a look at this and offering an improvement! However, since this is a matter of taste (and doesn't happen to match _my_ taste 😬 ), I don't plan to merge it. 

Just to share my thoughts, a few things I prefer are:

- Explicit inputs (method arguments) over implicit ones (bare method calls on `self`) 
- Explicit handling of unexpected inputs (the `ArgumentError`) instead of returning an unknown value 
- Obvious code at the call site (`context = {...}`) instead of indirection (`def context`)

So, thanks again for suggesting, but sorry it's not a good fit right now! Feel free to open an issue if you think the project would benefit from more strict style guidelines. (I've left `.rubocop.yml` pretty empty so that people wouldn't be too bothered by it 😆 )

derekprior:
Understood. This is a minor change and I can respect stylistic differences.

For what it's worth, I've yet to see anyone meet `ensure_hash` without some confusion or intrigue but that's a small sample size of 4 or 5 folks. Even still, it's relatively minor to overcome so if you prefer it then :+1:

rmosolgo:
😆  Oops, not meant to create intrigue! If I've missed the mark on this one, I'm willing to yield. Does anyone else have a thought on this generated code?

PR #1031, opened 2017-10-17
Doesn't need check rails installed for GlobalID

cpunion:
GlobalID does not dependency Rails.


gemfiles/without_rails.gemfile
```diff
@@ -8,5 +8,6 @@ gem "pg", platform: :ruby
 gem "pry"
 gem "pry-stack_explorer", platform: :ruby
 gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
+gem "globalid"
```
rmosolgo: This file is generated from the config in `Appraisals`, could you also update that file?

cpunion: Updated.

rmosolgo:
Cool, thanks for this!

PR #1035, opened 2017-10-17
Better error handling documentation linking to graphql-errors gem.

dblock:
It took me a while to get this setup properly for a production app. The doc seems to recommend doing things per resolver, which is tedious and problematic for most production scenarios. 

It's strange that `RescueFrom` wants a class to rescue from, seems like an unnecessary pattern because you typically handle exceptions in that rescue block, why would you want to declare in two places what that exception is? So I removed that option and documented how you can handle just about anything there with some useful examples for ActiveRecord.

Finally, graphql-errors does a great job at wrapping it in a schema. Don't know if that's ok since this is a library on top of this library, but seems like a useful link and that's what we ended up using.


rmosolgo:
Thanks for taking a look at these docs! That looks like a nice solution 😎

PR #1037, opened 2017-10-17
Class-based schema definition API

rmosolgo:
Here's another idea for schema definition API, use classes to define GraphQL types. It has a lot of advantages: 

- No DSL required for code-sharing or customizing 
- `super` has a predictable, useful meaning 
- Maybe classes will free us from Rails reloading torment

You can see an example in the specs: https://github.com/rmosolgo/graphql-ruby/blob/class-based-api/spec/support/jazz.rb

It's inspired by [GraphQL @ Shopify](https://youtu.be/Wlu_PWCjc6Y?t=8m8s) and the result of a lot of discussion and help from my teammates at GitHub. 

It has to be incrementally-adoptable, so we can transition bit-by-bit. I don't want to leave anyone high-and-dry on the switch. But that means lots of complexity in the code, to do all this working-around. 

Other thoughts: 
- ~~What if, instead of going from class to `.define`, we went the _other_ way? Maybe easier?~~ Instead, this new API is built independent of `.define` 
- Is the transition straightforward enough that we could write an automated codemod? 
- ~~How to generalize about cases like ActiveRecord::Base, where methods aren't all available at load-time? (Leave this issue til later?)~~ Static checking was removed altogether 
- ~~Still lots of TODO for other types~~ See a TODO list further down


cpunion:
Just saw https://github.com/rmosolgo/graphql-ruby/blob/bf519a52f1d42559f8e6c3f1b32eb933ffd5d3dc/spec/support/jazz.rb, seems beautiful. Does it support many types with same name? I want to build a multi-tenancy CMS, some models has same names likes `User` or `Comment` between different accounts.

rmosolgo:
> Does it support many types with same name?

Wow, that's really interesting 😬  Do you mean, the system has many different schemas which contain types with the same _names_ but aren't necessarily the same _types_? 

(All the more reason to maintain the current API!)

cpunion:
Can you change the type declaration to make it support that?

In example https://github.com/rmosolgo/graphql-ruby/blob/bf519a52f1d42559f8e6c3f1b32eb933ffd5d3dc/spec/support/jazz.rb#L26-L27, I guess just replace `"[Ensemble]"` to `types[Ensemble]` likes current API.

eapache:
The stringification of types is something I was meaning to ask about as well. At shopify we literally just pass in the ruby class object (e.g. `field :foo, Foo`) which wins us slightly stronger typing and lets us disambiguate when we have multiple types with the same name (you can have `field :foo, Module::Foo` in one schema and `field :foo, AnotherModule::Foo` in another). I'm not sure why types are stringified in this proposal?

lib/graphql/object/field.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+# test_via: ../object.rb
+
+module GraphQL
+  class Object
+    class Field
+      def initialize(name, return_type_expr, desc = nil, null: false, deprecation_reason: nil, method: nil, &args_block)
```
eapache: I would suggest making `null:` a required argument with no default; we've found internally that otherwise people simply forget to think about it more often than not.

spec/support/jazz.rb
```diff
@@ -0,0 +1,47 @@
...
+    name "Instrument"
+    field :name, !types.String
+  end
+
+  # Another new-style definition, with method overrides
+  class Query < GraphQL::Object
+    field :ensembles, "[Ensemble]"
+    field :instruments, "[Instrument]"
+
+    def ensembles
```
eapache: In our version we have `field` take an optional `resolve: ->(obj, args, ctx) { }`. The parameters are an important missing detail (does the caller check the arity here or something?) but otherwise I think I like this approach too. I think I'd suggest defining each method with its `field` call though, rather than putting all the `fields` at the top.

eapache:
I haven't reviewed the implementation, but the schema definition format looks pretty nice. Looking forward to using this!

swalkinshaw:
One of the main discussions in the previous attempt at this was around "how to add field-specific data in this API?" (ie: `before_resolve`).

Have any new ideas come up for this new solution?

lib/graphql/object/resolvers.rb
```diff
@@ -0,0 +1,103 @@
...
+        def initialize(*)
+          super
+          @has_args = has_args?
+        end
+
+        def call(obj, args, ctx)
+          if @has_args
+            kwargs = {}
+            args.to_h.each { |k, v| kwargs[k.to_sym] = v }
+            obj.public_send(@method_name, kwargs)
```
theorygeek: It would be pretty sweet to have some nicer ways of mapping the field's arguments to the method's arguments. Not sure if that's too magical though.

Mostly, I'm thinking automatically un-camelizing argument names and splatting them.

rmosolgo: :+1:  this splats them, so you can use keyword args, but it doesn't underscore-ize them yet 😊  definitely _should_ though.

rmosolgo: it doesn't _deep_ splat, that's definitely a todo, for input objects, i mean

rmosolgo:
@swalkinshaw no, nothing new in that area 😬 

@eapache thanks for your review, yeah, I think referencing proper objects (instead of strings) will be nice, I was just having a tough time with cycles, so I skipped it at first !

lib/graphql/object/resolvers.rb
eapache: Splatting them is cooling. And the object being resolved against can be accessed in these methods just with `self`. But how is `ctx` accessible? Is there an accessor I'm missing?

rmosolgo: They're each available as `@object` and `@context`, but `self` is an instance of the `GraphQL::Object` subclass

lib/graphql/object/field.rb
Yoshiji: :+1:

rmosolgo:
Push a couple more API changes, plus switched the Star Wars example schema to be `class Schema < GraphQL::Schema` (it doesn't quite work though)

rmosolgo:
Added a few more things: 

- `GraphQL::Enum`, a class-based API for `EnumType` 7ba4809
- Make `null:` required for arguments and fields 60a7bc2
- Add `GraphQL::Interface`, a class-based API for `InterfaceType` 121ec06 . 

  This one is pretty icky, I'm not quite sure the best way to make it Ruby-ish. Maybe GraphQL interfaces are Ruby modules (not classes) and can be included in Object classes?

If anyone wants to take a look, I'd love to hear your thoughts :)

eapache:
> This one is pretty icky, I'm not quite sure the best way to make it Ruby-ish. Maybe GraphQL interfaces are Ruby modules (not classes) and can be included in Object classes?

We ended up not fighting to make this any more rubyish, it does get messy. We have simply
```ruby
class Foo < GraphQL::Interface
  field ...
end

class Bar < GraphQL::Object
  implements Foo
end
```

Worth noting though that because we don't do resolve methods the way they're done in this PR, providing default implementations of interface fields isn't affected by the actual ruby class hierarchy.

lib/graphql/object/resolvers.rb
```diff
@@ -0,0 +1,48 @@
...
+    module Resolvers
+      class Dynamic
+        def initialize(method_name:)
+          @method_name = method_name
+        end
+
+        def call(obj, args, ctx)
+          if obj.respond_to?(@method_name)
+            public_send_field(obj, @method_name, args)
+          elsif obj.object.respond_to?(@method_name)
```
swalkinshaw: At some point, should we ensure that the only public instance methods on the `object` class correspond to fields?

swalkinshaw:
I realized that the question around "how to add field-specific data in this API" is irrelevant with this solution. It was only an issue with the IDL-based approach. 👍

lib/graphql/object/resolvers.rb
rmosolgo: I love the idea, but I also want to reduce friction for people getting started. Maybe the way to do it is a `Schema#lint`-type method that checks for that.

spec/support/jazz.rb
```diff
@@ -0,0 +1,123 @@
...
+    value "DIDGERIDOO", "Makes a sound by amplifying the sound of buzzing lips", deprecation_reason: "Merged into BRASS"
+  end
+
+  # Lives side-by-side with an old-style definition
+  InstrumentType = GraphQL::ObjectType.define do
+    name "Instrument"
+    implements GloballyIdentifiable::Interface.to_graphql
+    field :id, !types.ID, "A unique identifier for this object", resolve: ->(obj, args, ctx) { GloballyIdentifiable.to_id(obj) }
+    field :name, !types.String
+    field :family, Family.to_graphql.to_non_null_type
```
swalkinshaw: For the backwards compatibility, we could automatically call `to_graphql` if it responds to it? Same could go for `implements`.

rmosolgo: ohh, that would be nice ☺️

rmosolgo: Done in df3bd07

swalkinshaw:
> This one is pretty icky, I'm not quite sure the best way to make it Ruby-ish. Maybe GraphQL interfaces are Ruby modules (not classes) and can be included in Object classes?

This sounded good in theory to me at first. But I think it would get confusing with actual Ruby modules. One of the benefits of this class-based definition approach is that modules are useful now so you'd both plan Ruby modules *and* interfaces being included.

```ruby
class Ensemble < GraphQL::Object
  include ActiveRecordCache # or something
  include GloballyIdentifiable

  description "A group of musicians playing together"

  field :name, "String", null: false
  field :musicians, "[Jazz::Musician]", null: false
end
```

Without some naming scheme, it's not obvious what's an interface or not.

```ruby
class Ensemble < GraphQL::Object
  include ActiveRecordCache # or something

  implements GloballyIdentifiable

  description "A group of musicians playing together"

  field :name, "String", null: false
  field :musicians, "[Jazz::Musician]", null: false
end
```

This is more obvious and uses the GraphQL concept/name.

rmosolgo:
>  not obvious what's an interface 

That's a _great_ point, let me give it a try with `implements` as you posted!

swalkinshaw:
So this is missing the ability to "extend the DSL" right? (putting aside there is no DSL 😛 ).

Looks like a `field` definition right now just accepts a fixed set of method arguments. Do you have any ideas for something like `accepts_definitions`?

We've definitely felt the growing pains of this at Shopify. Our main `field` method takes 7 kwargs. Beyond that, we've also have "shortcut" methods which delegate into `field`. Each of those can take another kwarg or two.

We've had debates around these shortcut methods as well. I'd personally rather keep the API as small as possible with just `field` and `connection` since it maps more directly to GraphQL spec (and Relay) concepts.

One option is changing how `field` yields and don't use pass through the `block` for arg definitions only.

rmosolgo:
> missing the ability to "extend the DSL"

Yes, I have some thoughts there, and it's a hard requirement for our work at GitHub too. Let me try a few things and get back to you :)

rmosolgo:
A few updates: 

- Use `implements` for Interfaces 74be2d6 (See the `implemented do ... end` API for inheriting methods. It's pretty tricky to pass a method defn from one class to another, is that worth the trouble?)
- Build this API _without_ `.define` under the hood. It will make it easier to keep them separate 😬  c5af23f
- Add very basic InputObject support aa39f8a

rmosolgo:
@swalkinshaw Here's an idea for extending the DSL which I haven't tried yet: allow `GraphQL::Object` subclasses (and others) to override `to_graphql`, so they can do: 

```ruby 
def self.to_graphql 
  type_defn = super 
  if my_special_thing? 
    type_defn.metadata[:my_special_thing] = :whatever
  end 
  type_defn 
end 
```

The `.redefine` API could also be used there. (I think it will break currently with the `||=` caching in `to_graphql`. TODO 😬 )

rmosolgo:
I added some examples of customization by overriding `.to_graphql` and `.field_class`, but i think it's not quite there yet, for example, you have to override `.field_class` on _both_ Interface and Object in order to get matching behavior in both places. Not great.

lib/graphql/object/field.rb
xuorig: @eapache @rmosolgo agreed x1000, especially for Public APIs (because when it's deployed it's already too late to change 🙈)

It's pretty nice to make gem users to think about `null: true` vs `null: false`

spec/graphql/object_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    it "uses the custom field class" do
+      query_str = <<-GRAPHQL
+      {
+        ensembles { upcaseName }
+      }
+      GRAPHQL
+
+      res = Jazz::Schema.execute(query_str)
+      assert_equal ["BELA FLECK AND THE FLECKTONES"], res["data"]["ensembles"].map { |e| e["upcaseName"] }
```
xuorig: 👌 awesome band, :shipit:

xuorig: And I prefer this to 🧀 :trollface:

xuorig:
I see fields are not inherited in this version (i see you have a todo in there). Just thought I'd mention an example of how this could be useful:

```ruby
class RelayPayload < GraphQL::ObjectType
  field :client_mutation_id
end

class AddIssuePayload < RelayPayload
  field :issue, Issue
  field :errors, [Error]
end

# i would expect
# AddIssuePayload.fields == [:client_mutation_id, :issue, :errors]
```

wdyt?

lib/graphql/object/field.rb
```diff
@@ -0,0 +1,111 @@
...
+# test_via: ../object.rb
+
+module GraphQL
+  class Object < GraphQL::SchemaMember
+    class Field
+      include GraphQL::SchemaMember::CachedGraphQLDefinition
+
+      attr_reader :name
+
+      def initialize(name, return_type_expr = nil, desc = nil, null: nil, field: nil, function: nil, deprecation_reason: nil, method: nil, connection: nil, max_page_size: nil, resolve: nil, &args_block)
```
xuorig: should `connection` be that closely coupled to `GraphQL::Object` or could this be a Relay specialization on this class?

lib/graphql/object/field.rb
```diff
@@ -0,0 +1,111 @@
...
+              method_name: method_name,
+              connection: connection,
+            })
+          end
+          field_defn.connection = connection
+          field_defn.connection_max_page_size = @max_page_size
+
+          # apply this first, so it can be overriden below
+          if connection
+            conn_args = FieldProxy.new(field_defn)
```
xuorig: Related to above comment: Looks like this and other connection related logic could be extracted to a Connection function or something similar

spec/support/jazz.rb
```diff
@@ -0,0 +1,193 @@
...
+      objs = Models.data["Instrument"]
+      if family
+        objs = objs.select { |i| i.family == family }
+      end
+      objs
+    end
+  end
+
+  class EnsembleInput < GraphQL::InputObject
+    argument :name, String, null: false
```
cjoudrey: I wonder if `required: true | false` is more natural for arguments?

spec/support/jazz.rb
```diff
@@ -0,0 +1,193 @@
...
+  end
+
+  class Mutation < BaseObject
+    field :addEnsemble, Ensemble, null: false do
```
cjoudrey: Would it make sense to support `field :add_ensemble, ...` which would automatically get camel-cased?

spec/support/jazz.rb
```diff
@@ -0,0 +1,193 @@
...
+      argument :input, EnsembleInput, null: false
+    end
+
+    def add_ensemble(input:)
+      # TODO, how should this object be presented here?
+      # Maybe an instance of the class above, whose methods may be called?
+      ens = Models::Ensemble.new(input["name"])
```
cjoudrey: Should we prefer the use of symbols here? i.e. `input[:name]` instead.

lib/graphql/object/field.rb
eapache: Internally, 99% of our connection logic has been extracted to filters anyway, but I think filters are still something that's internal to us? Our `connection` method is literally just two lines now:
- turn `Type` into `TypeConnection`
- apply the pagination filter

spec/support/jazz.rb
```diff
@@ -0,0 +1,193 @@
...
+  class EnsembleInput < GraphQL::InputObject
+    argument :name, String, null: false
+  end
+
+  class Mutation < BaseObject
+    field :addEnsemble, Ensemble, null: false do
+      argument :input, EnsembleInput, null: false
+    end
+
+    def add_ensemble(input:)
```
cjoudrey: Just taking an idea from [Graphene](http://docs.graphene-python.org/en/latest/types/objecttypes/), what are your thoughts on prefixing the name of the method with `resolve_`? Could this alleviate future naming conflicts?

Also, should we still keep `root_obj` and `context`? Seems like they could come in handy.

rmosolgo: they're available inside methods as `@object` and `@context`, does that seem like a good solution?

spec/support/jazz.rb
rmosolgo: 😆  yes i suppose you're right. 

Maybe I should drop Ruby 2.1 support so I can `.to_sym` willy-nilly

spec/support/jazz.rb
cjoudrey: Oops, I totally missed that. 😊 

No strong opinions there. Having them in the args feels easier to discover, but also a hurdle when getting started.

lib/graphql/object/field.rb
rmosolgo: what's the _benefit_ to extracting it? it's not any easier to maintain, because the reality is that the dependency is very close (that's my experience from the dead `graphql-relay` gem). if anything, I'd like to see the gem become _more_ opinionated on cursor-based pagination, while keeping Relay connections as _one possible_ API for cursor-based pagination. 

or, is there another benefit to separating this code into another file?

spec/support/jazz.rb
rmosolgo: Oh, and about naming conflicts, are there any clearly predictable cases of it? Otherwise, how about we find a solution when the problem arrises? (eg, `method: :resolve_X` would be a simple case-by-case work-around)

lib/graphql/object/field.rb
eapache: > I'd like to see the gem become more opinionated on cursor-based pagination, while keeping Relay connections as one possible API for cursor-based pagination.

I'm not sure how specifically baking connections into the gem accomplishes that goal though?

I don't have a strong preference on API here as long as the full implementation is easily pluggable. We have some very shopify-specific things we do to paginate in the backend.

rmosolgo:
Here are a few items that I consider big TODOs: 

- How to support user-defined customizations for `field...` calls? For example, `field :ssn, ..., private: true`. I have the idea that you can override the built-in `GraphQL::Object::Field` class, (https://github.com/rmosolgo/graphql-ruby/pull/1037/commits/957908bf7681ed934d452d52ac44d88d1e2703f5), but the current implementation is _not_ that good because it requires copy pasta between `Object`s and `Interface`s. So how can we make that API very low-friction? 
- How to support customized introspection types? This is something I want to accomplish in this PR and I think it is possible by:  
  0. Implement field inheritance (as described by @xuorig in https://github.com/rmosolgo/graphql-ruby/pull/1037#issuecomment-340487213)
  1. Turn the built-in introspection types into class-based definitions 
  2. Support user-provided subclasses of the built-in ones _somehow_. It will be tricky. Maybe you call `introspection(...)` in the schema definition. 
- Get the "long tail" of auto-camelization & string-vs-symbol issues, such as deeply-nested input objects. 
- Work out a proper OO design for `InputObjects`. It methods can be called to access input values; methods may be overridden with the same effect as `prepare` functions. Maybe in a future version, these classes can accept & apply validation rules. 
- Implement inheritance where it makes sense. Many classes which have class-level ivars (such as `@fields` or `@interfaces`) should use an "inheritance-friendly" approach instead:

  ```ruby 
  class << self 
    # Get the fields implemented on this class, then get any parent fields which _aren't_ overridden.
    def fields 
      own_f = @own_fields.dup 
      # If the superclass has fields, get those
      inherited_fields = superclass < GraphQL::Object ? superclass.fields : []
      # Ignore fields which were overridden in this class 
      not_overridden_fields = inherited_fields.select { |i| !own_f.find {|f| f.name == i.name} }
      own_f + not_overridden_fields 
    end 

    # TODO: the .field method should add entries to `@own_fields`
  end 
  ```
- Make sure Relay helpers, `function:`, and `resolve:` have good test coverage. Sorry, I hustled those in here trying to get GitHub to work 😬 
- Same thing for `lazy_resolve`, make sure the test coverage is good. Maybe just port one of the example schemas. (eg `lazy_spec.rb`)
- Hit any remaining `# TODO`s in the code

rmosolgo:
(added a TODO item about customizing types)

eapache:
> the current implementation is not that good because it requires copy pasta between Objects and Interfaces. So how can we make that API very low-friction?

Internally we have a `HasFields` module that is shared by `Object` and `Interface`.

spec/support/jazz.rb
rmosolgo: :+1:

lib/graphql/object/field.rb
rmosolgo: Yes, sure, certainly want to _support_ advanced users building their own pagination. But I also want a "pit of success" for new users to get going quickly!

rmosolgo:
Yeah, a module like `HasFields` is dry, but it's asking people to make a lot of boilerplate. Here's what I _wish_ I could do: 

```ruby 
  # This field class will somehow be used by all your objects and interfaces 
  # since it's defined inside your schema  
  class Field < GraphQL::Object::Field 
    # customize, customize, customize
  end 
end 
```

But I can't think of a good way to make that work, with dependency loading. Maybe it's the wrong play anyways, because if you start using `MySchema` as a namespace, you have to remember the `< GraphQL::Schema` everywhere which is annoying. 

But that's the gist: is there some way where people can make a class, then we can fetch it by convention? (I pondered introducing a new namespace like `Environment`, where you could put your overridden Field, Object, Context, Introspection, etc, but I haven't thought it through either ...)

rmosolgo:
I took a try at the config namespace thing here: https://github.com/rmosolgo/graphql-ruby/compare/class-based-api...class-based-api-with-config-namespace

But it wasn't so good, so I'll forget about it for now!

spec/support/jazz.rb
rmosolgo: I was torn about _where_ the renaming should happen. For example, this field is called `:addEnsemble` here, but it calls the `#add_ensemble` method below.

One thing I want to take care of is compatibility: some people _already_ have underscored APIs, and I want them to be able to migrate to this API.

Can we accomplish that somehow, providing a user-friendly auto-camelize feature, while supporting underscore names as needed? Here are some thoughts: 

- Treat strings and symbols differently: symbols are camelized, but if a string is given, assume it has already been prepared, so don't auto-camelize it. 
- Make the auto-camelize function configurable, so it can be disabled 

Anything else?

rmosolgo:
Added proper support for `InputObject`s in 669cc4ec . 

The class that you define with `< GraphQL::InputObject` is used at runtime (on top of a `GraphQL::Query::Arguments`), so you can: 

- Override argument methods for `prepare:`-like behavior 
- Define helper methods willy-nilly
- Access `@context` in those methods 

For compatibility: 

- The underlying `@arguments` instance is also accessible 
- Several methods are delegated to that `@arguments`: `#[]`, `#key?`, `#to_h`, etc

rmosolgo:
It has some impact on "classic" `args` instances too: 

- They must be initialized with `context:` and may not be initialized with `argument_definitions:` (this could break someone's test setup, we can find a way to work around it) 
- They get underscored methods to access values

spec/support/jazz.rb
theorygeek: I’m not a huge fan of the string vs symbol behavior, but I could get behind it anyway.

But, could camelization be handled through some sort of instrumentation? If you could change the field’s name that way, without it changing any of the resolution behavior (ie, it would still try to resolve against the underscore name), that seems like it’d be ideal.

Or the library could camelize by default, but you can specify `camelize: false` on the field to bypass.

I do think it’s important that camelizing field names be something that is at _least_ opt-out, and maybe the safest is to make it opt-in.

cjoudrey: We might be able to [follow along the lines of Graphene](http://docs.graphene-python.org/en/latest/types/schema/#auto-camelcase-field-names):

> By default all field and argument names (that are not explicitly set with the `name` arg) will be converted from `snake_case` to `camelCase`
> 
> [...]
> 
> To disable this behavior, set the `auto_camelcase` to `False` upon schema instantiation.
> ```python
> my_schema = Schema(
>     query=MyRootQuery,
>     auto_camelcase=False,
> )
> ```

rmosolgo:
🎉  I got CI passing at GitHub. I left a couple of "stretch goals" unchecked, but I think they can be accomplished down the road. 

I still want to do a couple more things: 

- Reorganize this code & add API docs
- Run some benchmarks with GitHub, update accordingly 
- Write some guides w/ the experimental header

rmosolgo:
I ran some benchmarks in GitHub where I converted just a couple of object types to this new API. The benchmark mostly covered _other_ parts of the schema, but the good news was, there wasn't any overhead for _that_ part of the system! 

So, maybe those particular objects have a bit of overhead, but there's basically no penalty for _trying_ the new API in an isolated way :)

spec/support/jazz.rb
rmosolgo: 🤔  I'll give it a try ...

rmosolgo:
Tests pass 🎉  now, some docs...

rmosolgo:
I started on the docs here, so far I covered: 

- Goals & roadmap 
- Migration overview 
- Schema class 
- Object classes 
- How to customize 

https://github.com/rmosolgo/graphql-ruby/blob/a95ce400883015c2bf1142119d977ad4cbda935c/guides/schema/class_based_api.md

If you get a chance to review, I'd love to hear any new feedback on topics covered there!

lib/graphql/schema/member/has_fields.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        # Override this method to customize fields in your schema.
+        #
+        # To customize fields on {GraphQL::Schema::Object} and {GraphQL::Schema::Interface},
+        # - extend each of those classes, making a custom base class for your app
+        # - extend {GraphQL::Schema::Field}, making a custom base class for your app
+        # - override {.field_class} on your Object and Interface classes to return your new field class
+        #
+        # @return [Class] The class to initialize when adding fields to this kind of schema member
+        def field_class
```
swalkinshaw: What about exposing this as a class attribute? Looking at the docs it felt a little weird to see:

```ruby
class Foo < BaseObject
  Field = CustomField

  # vs
  field_class CustomField
  # or
  self.field_class = CustomField
end
```

Also seems a little more descriptive/obvious to what's happening under the hood.

rmosolgo: :+1:  you're so right, that's how the internal API works too 😆

lib/graphql/schema/object.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module GraphQL
+  class Schema
+    class Object < GraphQL::Schema::Member
```
bessey: In a world where we'll all be writing a _lot_ of `class BlahType < GraphQL::Schema::Object` I can imagine wishing that {Object, Scalar, Interface, Union, Enum, InputObject} were not namespaced to Schema, just to save some characters.

I understand not wanting to have an explosion of classes at the `GraphQL` level, but I think its merited for those 6 to live there (as is the case in the `GraphQL::ObjectType` syntax)

Perhaps, for the sake of internal consistency, they could be aliased to these shorthands?
```ruby
module GraphQL
  Object = Schema::Object
  ...
```

bessey: Although, if you are encouraging the pattern
```ruby
class GraphQL
  class BaseObject < Schema::Object
```
then perhaps this is unnecessary

guides/schema/class_based_api.md
```diff
@@ -0,0 +1,353 @@
...
+
+The second argument to `field(...)` is the return type. This can be:
+
+- A GraphQL type object built with `.define { ... }`
+- A GraphQL type class which you defined
+- A Ruby built-in such as `Integer`, `Float` or `String`
+- A string which corresponds to a GraphQL built-in, such as `"ID"`, `"Boolean"`, or `"Int"`
+- An _array_ of any of the above, which denotes a list type. Inner list types are always made non-null.
+
+Nullability is expressed with the required `null:` keyword:
```
bessey: Hmmm, this syntax is definitely more Rubyish than the `!` overloading, but not having a default of non-nullable seems counter to the GraphQL spec to me. `String` is nullable in the spec, and IMO it should be in the Ruby DSL too. 

In other words, I think these should be valid and equivalent:
```ruby
field :name, String, null: true
field :name, String
```

guides/schema/class_based_api.md
```diff
@@ -0,0 +1,353 @@
...
+class Types::BaseObject < GraphQL::Schema::Object
+end
+
+# then...
+class Types::TodoList < BaseObject
+  field :name, String, "The unique name of this list", null: false
+  # Related Object:
+  field :owner, Types::User, "The creator of this list", null: false
+  # List field:
+  field
```
bessey: Missing example

guides/schema/class_based_api.md
```diff
@@ -0,0 +1,353 @@
...
+
+When using these resolution implementations, they will be called with the same `(obj, args, ctx)` parameters as before.
+
+### Resolution with methods
+
+If you implement a field by defining a method, you should expect some automatic transformations:
+
+- GraphQL arguments will be converted to Ruby keyword arguments.
+- If the field name is `camelCased`, the method name should be `underscore_cased`.
+- If any argument names are `camelCased`, they will be passed to the method as `underscore_cased` Ruby keyword args.
```
bessey: 🤘

bessey: Sorry if I'm jumping the gun here since you just committed these docs, but is the intent to also perform this auto underscore_casing for all input objects (recursively?)

lib/graphql/schema/object.rb
Willianvdv: @bessey Arel has some shortcuts to do this, for example, you can use `Arel.sql` instead of `Arel::Nodes::SqlLiteral`. Maybe a similar pattern can be used for GraphQL schema definitions as well?

swalkinshaw: I think the intention is to mirror what Rails has recently done with `ApplicationRecord` for example. So the docs would need to encourage that behaviour. If that's the case, the Rails generator could even generate the `BaseObject` (or whatever it's called).

guides/schema/class_based_api.md
eapache: I argued for making it required somewhere up above. What we've found internally is that if we don't make it required developers forget to think about it entirely and end up making fields nullable that should really be non-null.

guides/schema/class_based_api.md
rmosolgo: yes, the intent is to do it recursively! I'll definitely touch on that more in the "Input Objects" section in this doc.

guides/schema/class_based_api.md
rmosolgo: oops, thanks 😊

lib/graphql/schema/object.rb
rmosolgo: Yeah, that's what I'm going for, maybe I can make that clear by starting _each_ section with an example like that.

rmosolgo:
Thanks for the review, @bessey!

lib/graphql/schema/member/has_fields.rb
rmosolgo: Updated in 3125a259

rmosolgo:
Aw man, TODO: custom `Argument` classes (and a few others added to the list above.

spec/support/jazz.rb
rmosolgo: Updated to `required:` in c5283139

rmosolgo:
I filled out the rest of the docs sections: 

- Interfaces 
- Unions 
- Input Objects 
- Enums 
- Scalars

You can see the updated docs here: 

https://github.com/rmosolgo/graphql-ruby/blob/1fa7a2a4/guides/schema/class_based_api.md

rmosolgo:
I'm going to release this as `1.8.0.pre1`, but it doesn't end the conversation here! 

If you get a chance to review these changes and have any suggestions, please feel free to add them here or open a fresh issue. I'm still open to serious changes!

cpunion:
Just saw https://github.com/rmosolgo/graphql-ruby/blob/89bab2d03fb2c4247b979cd2fadb673de2f5a9c3/lib/graphql/schema/member/build_type.rb#L56, is that a typo?

cpunion:
@rmosolgo I'm using `1.8-dev` branch and trying convert a project to class based API, how to define a Relay mutation?

lib/graphql/schema/member/instrumentation.rb
```diff
@@ -0,0 +1,113 @@
...
+              proxy_to_depth(result, @list_depth, @inner_return_type, ctx)
+            end
+          end
+
+          private
+
+          def proxy_to_depth(obj, depth, type, ctx)
+            if depth > 0
+              obj.map { |inner_obj| proxy_to_depth(inner_obj, depth - 1, type, ctx) }
+            elsif obj.nil?
```
phyrog: I thin this check needs to be the first one. I'm having problems with nullable lists here (`[SomeType!]`), where it tries to `map` over `nil` because `depth` is `1`.

Willianvdv:
@cpunion a mutation is just an ObjectType (see http://graphql-ruby.org/queries/mutations.html) so you just transform them just like any other query type. The inputs have changed, check https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/guides/schema/class_based_api.md#input-object-classes for more details on those types. If it doesn't work, can you file a new task? I'm happy to help.

Regarding the typo, you think it should be `if type_expr < GraphQL::Schema::Member` instead of `if Class < GraphQL::Schema::Member` since Class can't have the GraphQL::Schema::Member as an ancestor? I need @rmosolgo's help to answer this.

lib/graphql/schema/member/instrumentation.rb
Willianvdv: @phyrog thanks for your help! Can you file a PR on 1.8-dev and try to add a failing spec?

cpunion:
The following is a Relay mutation type:

```ruby
Mutations::SigninWithEmail = GraphQL::Relay::Mutation.define do
  name "SigninWithEmail"

  return_field :token, types.String
  return_field :user, Types::UserType
  return_field :errors, !Types::ErrorsType

  input_field :email, !types.String
  input_field :password, !types.String

  resolve ->(obj, args, ctx) {
    # ..
  end
end
```

It generates 3 types `SigninWithEmail`, `SigninWithEmailInput` and `SigninWithEmailPayload`, I don't need code to define them. I think rewrite every mutation type to **THREE** new types is too heavy for me.

cpunion:
> Regarding the typo, you think it should be if type_expr < GraphQL::Schema::Member instead of if Class < GraphQL::Schema::Member since Class can't have the GraphQL::Schema::Member as an ancestor? I need @rmosolgo's help to answer this.

Yes, I tested `Class < GraphQL::Schema::Member`, it always returns `nil`. I think those two types are `static`, maybe it should be `type_expr < GraphQL::Schema::Member` for this context.

Willianvdv:
@cpunion sorry, I misunderstood your initial question. I'm not sure if there's already a solution for the Relay mutation already. I have to, again, defer this to @rmosolgo.

cpunion:
@Willianvdv Thanks.

@rmosolgo Seems we can't mix **classical** style and **class based** style, so please ensure cover this case. I used some features but can't test right now. Do you know whether they are supported in new API?

* Batch loader (https://github.com/Shopify/graphql-batch)
* Relay mutation (http://graphql-ruby.org/relay/mutations.html)
* Resolver classes (http://graphql-ruby.org/fields/resolve_wrapper.html)

lib/graphql/schema/member/instrumentation.rb
phyrog: Could you point me to the best location for that test? I'm not sure where it belongs.

lib/graphql/object/field.rb
ianks: I would ❤️ top have a solid API for pagination. At AdHawk we integrate with bunch of APIs which do pagination differently so a mechanism to build connection adapters is incredibly important.

ianks:
Do you think this justifies a 2.0 release @rmosolgo?

rmosolgo:
> 2.0 release

Not in my opinion, nothing's broken and everything that used to work should still work just as it did!

PR #1040, opened 2017-10-18
Test GraphQL::Field#resolver_proc with a class.

lunks:
Also ensure we test using respond_to?(:call) instead of its class.

If you think testing with a class is unnecessary I can create another PR with just with just the assertion change.


lunks:
Thanks for the merge! :)

PR #1041, opened 2017-10-19
Add usage example for `hash_key` option for `field` definition

Yoshiji:
No hint or doc on this magical option which can be used instead of rewriting `resolve` with a lambda or resolver class


rmosolgo:
Sorry about that, and thanks for adding it to the docs!

PR #1047, opened 2017-10-20
Allow generator directory to be configured

ianks:
This PR allows users to specify where to install generated files in Rails. At AdHawk, we setup graphql before generators were around, so it is in `app/graph`. Instead of moving all the files in git, we would like to be able to configure this option. This PR makes that happen 😄


rmosolgo:
:+1: Nice! Glad you found something that worked for you, and thanks for upstreaming it :)

rmosolgo:
sorry for the slow turn-around, finally released in 1.7.5!

PR #1049, opened 2017-10-23
Replay previously seen fields

gjtorikian:
Suppose you have a query like this:

``` graphql
query getNodes($ids: [ID!]!) {
  nodes(ids: $ids) {
    __typename
    id
  }
}
```

Suppose `$ids` is a bit large, around 40 or 50. It turns out that the recursion in `build_rows` will continue to build up the same contexts over and over.

This PR is my attempt to address that problem by caching the contexts of previously iterated fields.


gjtorikian:
Discussed this in person and decided to pursue a solution elsewhere.

PR #1053, opened 2017-10-23
Adds parsing for https urls passed into the client via commandline

andxyz:
### The issue

using the client like this:

```
graphql-ruby-client sync --path='app/javascript/packs/myapp/graphql/' --url='https://myapp.mycompany.com/graphql/sync'
```

Was not hitting the `https` endpoints. I needed to specify the port in the url like this `url='https://myapp.mycompany.com:443/graphql/sync'`

### The Solution

Parse the urlString for the protocol and default to `443` for `https`, otherwise use port `80`.

### DIFF
--- a/javascript_client/sync/__tests__/sendPayloadTest.js
+++ b/javascript_client/sync/__tests__/sendPayloadTest.js
@@ -13,6 +13,27 @@ describe("Posting GraphQL to OperationStore Endpoint", () => {
     })
   })
 
+  it("Uses HTTPS when provided", () => {
+    var mock = nock("https://example2.com")
+      .post("/stored_operations/sync")
+      .reply(200, { "ok" : "ok" })
+
+    return sendPayload("payload", { url: "https://example2.com/stored_operations/sync" }).then(function() {
+      expect(mock.isDone()).toEqual(true)
+    })
+  })
+
+  it("Uses auth, port, and query", () => {
+    var mock = nock("https://example2.com:229")
+      .post("/stored_operations/sync?q=1")
+      .basicAuth({ user: "username", pass: "pass" })
+      .reply(200, { "ok" : "ok" })
+
+    return sendPayload("payload", { url: "https://username:pass@example2.com:229/stored_operations/sync?q=1" }).then(function() {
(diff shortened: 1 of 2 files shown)


rmosolgo:
Ohhh nice find and thanks for the fix! This seems like much better logic for providing a default. 

Do you think we should continue supporting _custom_ ports? For example, maybe we could use the port from the URL if it was provided, otherwise use the logic you added here: 

```js 
var parsedPort = parsedURL.port 
// Nothing was passed explicitly, so make a guess from the protocol
if (!parsedPort) {
  parsedPort = parsedURL.protocol === "https" ? "443" : "80"
}
```

andxyz:
@rmosolgo That is much smarter, I will make the change.

rmosolgo:
Very cool, I think that looks great! Just one more thing, how about adding a unit test for this new behavior (namely, using port `443` when the protocol is HTTPS)? 

Here's an existing test which using `HTTP`:

https://github.com/rmosolgo/graphql-ruby/blob/d7bb70dd18d17cf9eae119a0ba88340b85ff9044/javascript_client/sync/__tests__/sendPayloadTest.js#L5-L14

I think you could: 

- Copy that test case into a new test case 
- Change the URL passed to `nock(...)` so that it includes `https`. (This url is used to set up an expectation. I wonder, it might _also_ need the `:443`, I'm not sure!)
- Change the URL passed to `sendPayload` so that it also includes `https` 

Then the test should pass, and it should cover the case which was _previously_ broken, and make sure I don't break it again 😆 

Do you mind giving it a try?

andxyz:
Will do. Thanks for the pointers, I will try my luck at JS testing.

rmosolgo:
Geez, turned out a bit tough. My HTTP calls weren't being mocked, and I looked in all the _wrong_ places before learning that Node.js has a _different_ library for making HTTPS calls. 😩

andxyz:
@rmosolgo Thanks for picking up my slack. Apologies about the delayed work.

rmosolgo:
No problem, thanks for leading the charge :) Released in 1.3.0!

PR #1054, opened 2017-10-24
Adding some helpers for querying graphql types

theorygeek:
Adds a couple helpers:
```ruby
type = GraphQL::EnumType.define do
  name "Hello"
  value 'WORLD'
end

type.non_null?
# => false

type.to_non_null_type.non_null?
# => true

type.to_non_null_type.to_list_type.non_null?
# => false

type.list?
# => false

type.to_list_type.list?
# => true

type.to_list_type.to_non_null_type.list?
# => true
```

### DIFF
--- a/lib/graphql/base_type.rb
+++ b/lib/graphql/base_type.rb
@@ -206,6 +206,16 @@ module GraphQL
       printer.print_type(self)
     end
 
+    # Returns true if this is a non-nullable type. A nullable list of non-nullables is considered nullable.
+    def non_null?
+      false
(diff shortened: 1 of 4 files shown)


rmosolgo:
This is great, I have some ad-hoc implementations of `list?` which I will replace!

rmosolgo:
(oh, nevermind, looks like it's something specialized: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/context.rb#L255-L265)

PR #1056, opened 2017-10-25
Adding `all_arguments` and `get_argument` to InputObjectType

Yoshiji:
Adds `get_argument(arg_name)` and `all_arguments` to `InputObjectType`, very similar to `get_field` and `all_fields` of `ObjectType`:

```ruby
input_type = GraphQL::InputObjectType.define do
  name "Hello"
  argument :word, types.String
end

input_type.all_arguments
# => [#<GraphQL::Argument @name="word" ...>] 

input_type.get_argument('world') # works with symbol too 
# => #<GraphQL::Argument @name="word" ...>

# alias *_input_field instead of *_argument
input_type.method(:get_argument) == input_type.method(:get_input_field)
# => true

input_type.method(:all_arguments) == input_type.method(:all_input_fields)
# => true
```

### DIFF
--- a/lib/graphql/input_object_type.rb
+++ b/lib/graphql/input_object_type.rb
@@ -80,6 +80,18 @@ module GraphQL
       result
     end
 
+    # @return [GraphQL::Argument] The argument definition for `argument_name`
+    def get_argument(argument_name)
+      arguments[argument_name.to_s]
+    end
+    alias :get_input_field :get_argument
+
+    # @return [Array<GraphQL::Argument>] All arguments
+    def all_arguments
+      self.arguments.values
+    end
+    alias :all_input_fields :all_arguments
+
     private
 
     def coerce_non_null_input(value, ctx)
--- a/spec/graphql/input_object_type_spec.rb
+++ b/spec/graphql/input_object_type_spec.rb
@@ -329,6 +329,35 @@ describe GraphQL::InputObjectType do
     end
   end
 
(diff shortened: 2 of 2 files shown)


rmosolgo:
Can you help me understand the advantage of `get_argument`/`all_arguments` vs accessing `arguments` hash directly? 

With `GraphQL::ObjectType`, we have to introduce some other methods in order to get fields which are "inherited" from interfaces. But is there an advantage to this other API here?

Yoshiji:
It is just a code sugar, nothing more than a shortcut which follows the same method names/logic as for `ObjectType`.

There are some places where I get all the arguments of an `InputObjectType` and do something for each. I did something similar with `ObjectType`, and I thought that having `all_fields` was neat in that context but having to do `.arguments.values` was ... not natural?

Yoshiji:
Update:
I noticed the same naming convention for `EnumType` (`values`) and understood that there is probably several other Types having this simple hash storing all the `fields`/`values`/`arguments` etc. I first thought `InputObjectType` was an exception and was missing those `get_argument` & `all_arguments` methods, but now I see that it is actually the opposite: `ObjectType` is the exception, having to deal with interfaces.

This PR can be discarded, sorry for the waste of time.

rmosolgo:
Not a waste of time at all! I appreciate your review of the code and your suggestion. If someone else is also interested in this change later on, we can reconsider it :)

PR #1039, opened 2017-10-17
Add support for Statsd tracing

esjee:
Usage:
```
MySchema = GraphQL::SChema.define do
  ...
  use GraphQL::Tracing::StatsdTracing, statsd: Statsd.new('localhost', 1337)
  ...
end
```

PR #1060, opened 2017-10-27
Add DataDog Tracing support

michalg-:
dash1
dash2
details
trace


rmosolgo:
This looks great! Thanks for sharing your work. I'm going to leave it open for a bit in case any other DataDog users want to take a look :)

lib/graphql/tracing/data_dog_tracing.rb
```diff
@@ -0,0 +1,44 @@
...
+        pin.tracer.trace(platform_key, service: pin.service) do |span|
+          if key == 'execute_multiplex'
+            span.resource = data[:multiplex].queries.map(&:selected_operation_name).join(', ')
+          end
+
+          if key == 'execute_query'
+            span.set_tag(:selected_operation_name, data[:query].selected_operation_name)
+            span.set_tag(:selected_operation_type, data[:query].selected_operation.operation_type)
+            span.set_tag(:query_string, data[:query].query_string)
+            span.set_tag(:variables, data[:query].variables.to_h)
```
esjee: While awfully useful, this could be a bit dangerous! Imagine a mutation like "log in", for example. Wouldn't this cause you to run the risk of storing the user's credentials in DataDog? Even worse, imagine a mutation that passes along a user's payment details, such as a creditcard!

lib/graphql/tracing/data_dog_tracing.rb
```diff
@@ -0,0 +1,44 @@
...
+          end
+          yield
+        end
+      end
+
+      def platform_field_key(type, field)
+        "#{type.name}.#{field.name}"
```
esjee: It might be helpful to have the query name here as well -- that makes it easier to find a specific query inside DataDog's UI.

lib/graphql/tracing/data_dog_tracing.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        pin.tracer.trace(platform_key, service: pin.service) do |span|
+          if key == 'execute_multiplex'
+            span.resource = data[:multiplex].queries.map(&:selected_operation_name).join(', ')
+          end
+
+          if key == 'execute_query'
+            span.set_tag(:selected_operation_name, data[:query].selected_operation_name)
+            span.set_tag(:selected_operation_type, data[:query].selected_operation.operation_type)
+            span.set_tag(:query_string, data[:query].query_string)
```
esjee: FWIW I've sent an email to datadog informing if we could use their resource field to store the query instead. This is similar to how their SQL tracing works. This could potentially give us some small advantages, such as seeing the resource mentioned in the list view, as well as having syntax highlighting inside DataDog. No change required, just wanted to mention it. :smile:

lib/graphql/tracing/data_dog_tracing.rb
```diff
@@ -0,0 +1,44 @@
...
+      }
+
+      def platform_trace(platform_key, key, data)
+        pin = Datadog::Pin.get_from(self)
+        unless pin
+          pin = Datadog::Pin.new('graphql-ruby', app: 'graphql-ruby')
+          pin.onto(self)
+        end
+
+        pin.tracer.trace(platform_key, service: pin.service) do |span|
```
esjee: For the service, it might be nice to conform to the same names that dd-trace-rb seems to uphold? So

"#{app-name}-#{service}". E.g. `myapp-controller`, `myapp-postgresql`.

It might be nice if this were configurable. E.g.

```
  use GraphQL::Tracing::DataDogTracing, service: "myapp-graphql"
```

lib/graphql/tracing/data_dog_tracing.rb
michalg-: @esjee what do You think about current solution? :)

lib/graphql/tracing/data_dog_tracing.rb
michalg-: There is strange aggregation strategy in datadog panel. If that field on the top level is different, datadog redirects to the most popular query name when You click on the service name :(

lib/graphql/tracing/data_dog_tracing.rb
```diff
@@ -0,0 +1,51 @@
...
+          yield
+        end
+      end
+
+      def platform_field_key(type, field)
+        "#{type.name}.#{field.name}"
+      end
+
+      def scrub_variables(variables, blacklisted_variables)
+        Scrubber::Variables.perform(variables, blacklisted_variables)
```
esjee: This scrubbing seems like a good idea!

I'm still a bit afraid of a blacklist approach versus a whitelist approach, though. For queries I think it's fine, but for mutations less so. How would you feel about an option that removes arguments from mutations, and users can toggle this when setting the tracing? So, something along the lines of:

```ruby
use DataDogTracing, send_mutation_variables: true
```

esjee:
I created a PR to address my own comments! https://github.com/Foap/graphql-ruby/pull/1

rmosolgo:
Based on suggestions and your work here, I implemented `Query#scrubbed_variables`, and used that in the tracing hook. 

Does that look ok to you, @michalg- and @esjee ?

lib/graphql/schema/scrubber.rb
```diff
@@ -0,0 +1,72 @@
...
+        @mutations = mutations
+      end
+
+      SCRUBBED = "*****"
+
+      # @return [Hash]
+      def scrubbed_variables(query)
+        var_h = query.variables.to_h
+        if @mutations == false && query.mutation?
+          filter_value(var_h, scrub_all: true)
```
xuorig: that would filter out variables of other operations in the query that might not be mutations right?

rmosolgo: Haha, I wonder how that would behave. Let me write a couple of tests and well see 😊

lib/graphql/schema/scrubber.rb
```diff
@@ -0,0 +1,72 @@
...
+      # It only filters variables, not literal values in the query string.
+      #
+      # @param whitelist [Array<String, Symbol>] Argument names to _allow_
+      # @param blacklist [Array<String, Symbol>] Argument names to _hide_
+      # @param mutations [Boolean] If false, _no_ variables to mutations will be shown
+      def initialize(whitelist: nil, blacklist: nil, mutations: nil)
+        @whitelist = whitelist ? whitelist.map(&:to_s).to_set  : nil
+        @blacklist = blacklist ? blacklist.map(&:to_s).to_set : nil
+        if @whitelist && @blacklist
+          raise ArgumentError, "#{self.class.name} supports whitelist: _or_ blacklist:, but not both."
```
xuorig: like we talked earlier should we open this up to:

```
Scrubber.new(->(variable, query, context) { query.mutation? })
```

or something like `ScrubVariables.new(scrubber).scrub`?

Maybe this can stay private for now, and I can refactor this into a public API?

rmosolgo: 🤔  I don't quite understand, what's the difference here? (Different outputs, different use cases?)

xuorig: I'm thinking of ways to allow users to scrub according to their own rules, like`context[:is_secret_query]` :D

Maybe it's not needed though 🤷‍♂️

rmosolgo:
You know, I think there's actually a big issue in this design: 

It filters based on the _variable_ name (eg `$password`) instead of the _argument_ name (eg `login(password: ...)`. So, in some cases, we will miss important sensitive data: 

```graphql 
mutation($myCrazyPassword: String!) {
  login(password: $myCrazyPassword) 
}
```

In that case, `$myCrazyPassword` would get logged, even if you put `password` on the blacklist. 

But, I think when we put `password` on the blacklist, we mean to filter _any_ variable which is used to _populate_ a `password` argument. So, let me revisit this!

xuorig:
@rmosolgo even just argument name might fall short right? `password` might be an easy case but what if I want to hide:

```
creditCard(number: "")
```

vs 

```
phone(number: )
```

I might want to hide number on credit card fields, but not for others 🤔

rmosolgo:
Yeah, the current implementation treats `Phone.number` and `CreditCard.number` as the same... not ideal, but better than nothing! 

A straightforward improvement would be: 

- If the named item has a dot, blacklist that item only (eg `PhoneInput.number`) 
- If the named item doesn't have a dot, blacklist that item on _all_ fields (eg `password`)

esjee:
How about recording per mutation whether or not its variables should be removed or not?

```
# Mutations::MyMutation
  input_field :foo, !types.String, do_not_scrub: true
```

The worry I have is that putting all the scrubbing data in schema.rb is prone to errors and easily outdated or forgotten. This is just a suggestion though, and I'm completely unsure of how it would be implemented, assuming it's even possible :sweat_smile:

esjee:
@xuorig @rmosolgo 
It feels like this PR has gotten a bit stuck. How about removing the variable scrubbing, and merging the DD tracing without that? I could get on with #1039 , and you could add a general-purpose variable scrubbing for all traces. Does that seem reasonable?

rmosolgo:
Yeah @esjee, I think you're right it's gotten stuck, removing scrubbing sounds :ok_hand: to me

rmosolgo:
Thanks for your work here @michalg- , your commit was merged in #1129

PR #1061, opened 2017-10-27
Split up the instrumentation of fields for tracing

esjee:
As mentioned in a comment in https://github.com/rmosolgo/graphql-ruby/pull/1060, this code should make it easier to do tracing for a field should you be so inclined. 

I'm unsure how this should work exactly, considering that depending on your schema and the queries you execute, this could result in a lot of additional keys being used on the service. That, in turn, may result in a much higher bill.

At least (selfishly :see_no_evil:) this allows us to play around with it internally and see if it's worth the cost.
Curious to know what others think!

### DIFF
--- a/lib/graphql/tracing/platform_tracing.rb
+++ b/lib/graphql/tracing/platform_tracing.rb
@@ -41,12 +41,20 @@ module GraphQL
         return_type = field.type.unwrap
         case return_type
         when GraphQL::ScalarType, GraphQL::EnumType
-          field
+          instrument_scalar_field(type, field)
         else
-          new_f = field.redefine
-          new_f.metadata[:platform_key] = platform_field_key(type, field)
-          new_f
+          instrument_nonscalar_field(type, field)
         end
+      end
+
+      def instrument_scalar_field(type, field)
+        field
+      end
+
+      def instrument_nonscalar_field(type, field)
+        new_f = field.redefine
+        new_f.metadata[:platform_key] = platform_field_key(type, field)
+        new_f
       end
 
       def self.use(schema_defn)


rmosolgo:
The goal here is to add instrumentation to certain scalar fields, even though it's disabled for those fields by default, right? 

If that's right, I'd like to suggest an alternate implementation, one that I hope will be easier for people to use without digging into the tracers. What do you think of it: 

- Add a flag to fields called `trace`, something like: 

  ```ruby 
  field :complicatedValue, types.Float, trace: true 
  ```
  (This can be accomplished by [extending the DSL](http://graphql-ruby.org/schema/extending_the_dsl) in `graphql/field.rb`)

- Then, in the tracing implementation, modify the logic for skipping Enum and Scalar fields. Instead of skipping them, check the type's `metadata` hash to see if the flag was set, something like: 

  ```ruby 
  if return_type.is_a?(GraphQL::ScalarType) || return_type.is_a?(GraphQL::EnumType)
    if field.metadata[:trace]
      # Usually, this would be skipped, but it was explicitly flagged
      apply_tracing(field)
    else 
      # Since the flag wasn't set, skip tracing on this field 
     field 
    end 
  end 
  ```  

What do you think, would that do the trick? If you want to take a crack at it, I'm happy to help out, otherwise, I can take a try in the next day or two!

rmosolgo:
(This is the same technique used in graphql-pro for `monitoring: true`)

esjee:
Thanks for taking a look @rmosolgo ! I like the suggestion, but just one question. I think it would be nice to able to control this from a schema level. So rather than having to say `trace: true` for every field, I'd like to be able to set the default value for the entire schema. I suppose this can be solved by adding a `trace_fields_by_default true|false` option to the schema?

rmosolgo:
Yeah, that solution sounds great!

esjee:
I'll work on an alternative solution like the one you suggested!

PR #1062, opened 2017-10-28
Add message about gem order in tracing guide

mengledowl:
I was receiving an error when trying to implement the built-in tracing feature for Scout. It turns out that this is due to this piece of code where the library gets included if it has been defined:

https://github.com/rmosolgo/graphql-ruby/blob/d7bb70dd18d17cf9eae119a0ba88340b85ff9044/lib/graphql/tracing/scout_tracing.rb#L6-L8

If your Gemfile is ordered such that `gem graphql` comes before `gem scout_apm`, the `include` never gets run because `ScoutApm` is not defined yet, causing the code to blow up when it can't find the `instrument` message. If you order your Gemfile such that `scout_apm` comes first, it works fine.

This change adds a quick note in the docs to warn developers that they might need to make sure that they are defining their monitoring gem before the `graphql` gem.

### DIFF
--- a/guides/queries/tracing.md
+++ b/guides/queries/tracing.md
@@ -66,6 +66,16 @@ Leaf fields are _not_ monitored (to avoid high cardinality in the metrics servic
 
 Implementations are based on {{ "Tracing::PlatformTracing" | api_doc }}.
 
+You may need to define the monitoring gem **before** the `graphql` gem in your Gemfile for this to work.
+ 
+For example:
+ 
+```ruby
+# Gemfile
+gem 'scout_apm'
+gem 'graphql'
+```
+
 ## Appsignal
 
 To add [AppSignal](https://appsignal.com/) instrumentation:


rmosolgo:
Ohh interesting, I remember handing this a bit differently in the graphql-pro version, here it is: 

```ruby
          def initialize(...)
            #...
            # Do it during initalize to avoid load order issues
            self.class.include(ScoutApm::Tracer::ClassMethods)
          end
```

Maybe we should just do it _that_ way to avoid the issue, whadd'ya think? (I know it would be _bad_ to do that if we were initializing that _often_, or a times other than during app boot, but I think it's fair game _here_.)

mengledowl:
I agree - this isn't something that should really ever get initialized more than once, so I don't really see it being "bad". In any case, this is definitely better than just telling people to change the order of their Gemfile.

Working on a PR now! I'll go ahead and close this one in favor of correcting the code itself 😄

PR #1064, opened 2017-10-29
Include necessary Scout module from initialize on ScoutTracing

mengledowl:
### Issue

When attempting to use the built-in tracing feature and plugging in Scout, the order that you specify your dependencies in your Gemfile can cause your code to blow up. For example, if your Gemfile specifies:

```
gem 'scout_apm'
gem 'graphql'
```

Then the code works fine. If, however, it specifies `gem 'graphql'` _first_, it will blow up because `ScoutApm` is not defined when `ScoutTracing` attempts to include the module.

### Solution

Per [this conversation](https://github.com/rmosolgo/graphql-ruby/pull/1062) with @rmosolgo, I have removed the offending lines and am now performing the include on initialize.

### DIFF
--- a/spec/graphql/tracing/scout_tracing_spec.rb
+++ b/spec/graphql/tracing/scout_tracing_spec.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe GraphQL::Tracing::ScoutTracing do
+  class ScoutApm
+    module Tracer
+    end
+  end
+
+  describe "Initializing" do
+    it "should include the module only after initilization" do
+      refute GraphQL::Tracing::ScoutTracing.included_modules.include?(ScoutApm::Tracer)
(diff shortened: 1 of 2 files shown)


spec/graphql/tracing/scout_tracing_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+require "spec_helper"
+
+describe GraphQL::Tracing::ScoutTracing do
+  class ScoutApm
+    module Tracer
+      def self.instrument(_, _, _)
+        "stubbed"
+      end
+    end
+  end
```
rmosolgo: I love the idea of stubbing the various platforms, this was how i tested them in graphql-pro, too. But is this stubbed method called anywhere? If not, can we remove it? (I guess we _need_ the modules in order to `include` them)

mengledowl: Good call - I'll pull it out!

rmosolgo:
💰 awesome, thanks for the fix!

PR #1065, opened 2017-10-30
Parameterize description

nickpoorman:
Otherwise you get an error:

```
NameError:
     wrong constant name
```


rmosolgo:
Oopsie, thanks for the fix!

PR #1070, opened 2017-11-01
Remove support for Ruby 2.1

rmosolgo:
My interest here is in [Symbol GC](https://bugs.ruby-lang.org/issues/9634), which means we can use `.to_sym` without worrying about a DOS vulnerability. This will simplify code in Argument and Variable hashes, which can use symbols under the hood, which people prefer for access anyways. 

Oh, and I should add, support for Ruby 2.1 [ended in April 2017](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
   - gem install bundler
 
 rvm:
-- 2.1.10 # Lowest version officially supported by the gem is 2.1
+- 2.2.8 # Lowest version officially supported by the gem is 2.2
 - 2.4.2
 - jruby-9.1.9.0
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
Is anyone still depending on Ruby 2.1? (I've never heard of it, but ya never know!)

theorygeek:
Do it!

mull:
Wouldn't this warrant a major version change?

rmosolgo:
No, I don't think so, you can see my documented versioning scheme here: http://graphql-ruby.org/development#versioning

rmosolgo:
Would you prefer it released in 1.8.0 instead of 1.7.5?

mull:
It at least seems to warrant a minor version bump rather than patch.

> PATCH version indicates bug fixes or small features for specific use cases. Ideally, you can upgrade patch versions with only a quick skim of the changelog.

Shouldn't have breaking changes right?

> MINOR version indicates significant additions, internal refactors, or small breaking changes.

This seems more appropriate.

FWIW I'm on 2.4 so this doesn't apply to me, I'd be a little bit pissed to upgrade from 1.7.4->1.7.5 and have to also upgrade my Ruby version though!

rmosolgo:
But to be fair, it's a Ruby version that doesn't even get security patches anymore!

mull:
Your call, it just stood out to me when I saw the PR and the milestone :)

rmosolgo:
:+1: I'll release it in 1.8.0

PR #1071, opened 2017-11-01
Fix Backtrace execution context

swalkinshaw:
`execution_context.last` was not guaranteed to be an instance of `Context` in certain cases (at least during introspection).

This applies some normalization and filtering to `execution_context` to ensure that we're passing a `Context` to `TracedError`.

Note: I don't fully understand the root cause of this, but this test replicates what we ran into internally and the fix works.

Suggestions on naming welcome, every term is overloaded 😖 

cc @eapache

### DIFF
--- a/lib/graphql/backtrace/tracer.rb
+++ b/lib/graphql/backtrace/tracer.rb
@@ -31,7 +31,7 @@ module GraphQL
             rescue StandardError => err
               # This is an unhandled error from execution,
               # Re-raise it with a GraphQL trace.
-              raise TracedError.new(err, execution_context.last)
+              raise TracedError.new(err, get_context_from_execution(execution_context))
             ensure
               execution_context.clear
             end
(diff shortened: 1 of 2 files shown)


eapache:
What is `execution_context` supposed to be... an array of `Context`s? I don't really understand this either...

eapache:
So... this rescue only runs when `key == "execute_multiplex"` which means `push_data` is `metadata[:multiplex].queries` which gets pushed onto `execution_context` so that should be the last element? But that's not a context at all.

The tracing API must be recursive, so that by the time the rescue gets run something else has been pushed onto the context inside the `yield` call.

I'm wondering if it makes more sense to just not raise a traced error if `execution_context.last` isn't a `Context`? If I'm reading properly, that means the exception wasn't raised from a field, so there is no value in tracing the error anyway, and it's probably simpler and more obvious to just re-raise it?

swalkinshaw:
> The tracing API must be recursive, so that by the time the rescue gets run something else has been pushed onto the context inside the yield call.

Correct from my testing.

If the backtrace feature is only meant for errors during resolvers (or execution), then you're right it may be more correct to just re-raise the original exception at that point.

rmosolgo:
> not raise a traced error

This seems like a good solution too, since the backtrace doesn't contribute anything useful anyways, right?

swalkinshaw:
https://github.com/rmosolgo/graphql-ruby/pull/1078 is the alternative to this approach.

rmosolgo:
Thanks, let's try that one instead and see how it goes!

PR #1074, opened 2017-11-02
Mark queries as @executed when unhandled errors are raised

### LINKED ISSUE #1072: Infinite resolution when instrumenter inspects query result
Given the following conditions, I've run into an infinite loop when resolving a query. This is used to build tracing data, similar to https://github.com/uniiverse/apollo-tracing-ruby/blob/0ac17ca17f328ca13cee2ad29308874726f44f90/lib/apollo_tracing.rb#L62

* The query instrumenter inspects the query with `query.result` in `after_query`.
* The result throws an exception not inheriting a `GraphQL::ExecutionError`

This will prevent the multiplexer from executing [`finish_query`](https://github.com/rmosolgo/graphql-ruby/blob/4eef45fee9cdda0b139fea5d96cec06b0a8448ae/lib/graphql/execution/multiplex.rb#L90)as the [resolver](https://github.com/rmosolgo/graphql-ruby/blob/4eef45fee9cdda0b139fea5d96cec06b0a8448ae/lib/graphql/execution/multiplex.rb#L85) will raise an exception.

With an instrumenter, the [`ensure` block](https://github.com/rmosolgo/graphql-ruby/blob/4eef45fee9cdda0b139fea5d96cec06b0a8448ae/lib/graphql/execution/multiplex.rb#L182) will still execute. Note that `finish_query` has not set the query to [`executed`](https://github.com/rmosolgo/graphql-ruby/blob/72cf8d6fe637c2273730b1a6a16cd6431e4b08b7/lib/graphql/query.rb#L135). This will cause `after_query` to execute in a loop, since the multiplexer will never have the opportunity to set the query to `executed`.

I've attached a repro. This is on version 1.7.4.

```ruby
require 'graphql'

class QueryInstrumentation
  def before_query(query)
  end

  def after_query(query)
    puts query.result.to_json
  end
end

  name "Query"

  field :raiseExecutionError, types.String do
    resolve ->(object, args, ctx) {
      raise GraphQL::ExecutionError, "Whoops"
    }
  end

  field :raiseError, types.String do
    resolve ->(object, args, ctx) {
      raise GraphQL::Error, "Whoops"
    }
  end
end

  query QueryType
  instrument(:query, QueryInstrumentation.new)
end

MySchema.execute("{ raiseExecutionError }")
# {"data":{"raiseExecutionError":null},"errors":[{"message":"Whoops","locations":[{"line":1,"column":3}],"path":["raiseExecutionError"]}]}

MySchema.execute("{ raiseError }")
# `after_query': stack level too deep (SystemStackError)
```

rmosolgo:
Fixes #1072

### DIFF
--- a/lib/graphql/execution/multiplex.rb
+++ b/lib/graphql/execution/multiplex.rb
@@ -91,6 +91,10 @@ module GraphQL
             # Get the Query::Result, not the Hash
             query.result
           end
+        rescue StandardError
+          # Assign values here so that the query's `@executed` becomes true
+          queries.map { |q| q.result_values ||= nil }
+          raise $!
         end
 
         # @param query [GraphQL::Query]
--- a/lib/graphql/execution/multiplex.rb
+++ b/lib/graphql/execution/multiplex.rb
@@ -94,7 +94,7 @@ module GraphQL
         rescue StandardError
           # Assign values here so that the query's `@executed` becomes true
           queries.map { |q| q.result_values ||= nil }
-          raise $!
+          raise
         end
 
         # @param query [GraphQL::Query]
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -137,6 +137,9 @@ module GraphQL
       end
     end
 
+    # @api private
+    attr_reader :result_values
+
     def fragments
       with_prepared_ast { @fragments }
     end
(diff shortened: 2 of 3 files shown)


lib/graphql/execution/multiplex.rb
```diff
@@ -91,6 +91,10 @@ def run_as_multiplex(multiplex)
             # Get the Query::Result, not the Hash
             query.result
           end
+        rescue StandardError
+          # Assign values here so that the query's `@executed` becomes true
+          queries.map { |q| q.result_values ||= nil }
+          raise $!
```
eapache: `raise` with no arguments will re-raise the rescued error automatically

PR #1075, opened 2017-11-02
Remove confusing Notify statement.

dblock:
Just like https://github.com/exAspArk/graphql-errors/pull/3. The `Notify` statement is confusing because it doesn't work in a default Rails setup, I copy-pasted it a few times by mistake.

Also is there anything in http://code.dblock.org/2017/10/30/generating-and-handling-errors-in-graphql-in-ruby.html that you would find valuable in this doc?

### DIFF
--- a/guides/queries/error_handling.md
+++ b/guides/queries/error_handling.md
@@ -120,10 +120,10 @@ class RescueFrom
     nil
   rescue ActiveRecord::RecordInvalid => err
     # return a GraphQL error with validation details
-    GraphQL::ExecutionError.new("Validation failed: " + e.record.errors.full_messages.join("\n"))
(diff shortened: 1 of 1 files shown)


rmosolgo:
Oh, sure, I was imagining it was a "your bug tracker here" kind of thing. 

That's a nice blog post, I appreciate how errors are split into three categories. I'm still getting a handle on how things can go wrong (as you've noticed 😬 ). I guess this part can be updated now: 

> graphql-ruby doc on errors gives a lot of options to help you shoot yourself in the foot

😆

dblock:
https://github.com/dblock/code.dblock.org/commit/e7607450cd9899ca5aeb99f7a7c8f1875437acd1

PR #1078, opened 2017-11-07
Only raise TracedError when context exists

swalkinshaw:
`execution_context.last` is not guaranteed to be a context object.

This can occur when an exception is raised outside of field resolution.
There's not much use in a `TracedError` in this situation, instead of
trying to normalize and find the context, let's just re-raise the
original exception.

See https://github.com/rmosolgo/graphql-ruby/pull/1071 for more background.

cc @eapache


swalkinshaw:
@rmosolgo not sure if there's any better (or more direct) ways to test this.

rmosolgo:
LGTM, those same JRuby errors are on my other branch, so i guess they're unrelated 😖

PR #1085, opened 2017-11-11
Pluggable action cable serializer

staugaard:
This allows us to use our own serializer for action cable subscriptions.

I can't find a good way to test this.

### DIFF
--- a/lib/graphql/subscriptions/action_cable_subscriptions.rb
+++ b/lib/graphql/subscriptions/action_cable_subscriptions.rb
@@ -62,6 +62,8 @@ module GraphQL
     class ActionCableSubscriptions < GraphQL::Subscriptions
       SUBSCRIPTION_PREFIX = "graphql-subscription:"
       EVENT_PREFIX = "graphql-event:"
+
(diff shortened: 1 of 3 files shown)


rmosolgo:
:+1: This is great, I might even move it up to the parent class so I can use it _outside_ of ActionCable! 

Do you mind sharing, how is your custom serializer different from the included one?

staugaard:
The project where I need this on builds ActiveModel objects from http responses, and I need to send those objects to subscriptions.

I'm currently using this serializer:

```ruby
module ActionCableModelSerializer
  def self.dump(model)
    JSON.dump(
      class: model.class.name,
      attributes: model.as_json
    )
  end

  def self.load(str)
    json = JSON.parse(str)
    model_class = Object.const_get(json.fetch('class'))
    model_class.new(json.fetch('attributes'))
  end
end

```

staugaard:
any chance of getting this merged?

rmosolgo:
Yes, I'll try to add a test today! I was just waiting to see what your customization was, in case there was another takeaway for the gem. But it sounds like not!

rmosolgo:
Thanks for the improvement!

PR #1090, opened 2017-11-14
CI against JRuby 9.1.14.0

284km:
JRuby 9.1.14.0 has been released.
http://jruby.org/2017/11/08/jruby-9-1-14-0


rmosolgo:
Very cool, thanks for the update! (And as a bonus, it passes CI 😆 )

PR #1092, opened 2017-11-15
Fix Sync client compatibility with Apollo's createBatchingNetworkInterface

etripier:
This is required to use Apollo's `createBatchingNetworkInterface` API. Fortunately, the rest of the Sync client works just fine after this change (that's a different story with Apollo Client 2.0, but one step at a time!).

### DIFF
--- a/javascript_client/sync/generateClient.js
+++ b/javascript_client/sync/generateClient.js
@@ -68,6 +68,18 @@ var OperationStoreClient = {
    * Replace the query with an operationId
   */
   apolloMiddleware: {
+    applyBatchMiddleware: function(options, next) {
+      options.requests = options.requests.map(function(req) {
+        // Fetch the persisted alias for this operation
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the improvement! I never saw the batching interface before, so I'll have to take a look. 

Just curious, do you use `multiplex(...)` on the server for those queries? 

🚢  Just released these changes as 1.2.0 to npm!

etripier:
Yep, we use your nifty `Schema.multiplex` method. Actually, just as a side note, here's how I modified the multiplex mapping to work with the `operationId` parameter:

```ruby
private def multiplex_queries(request_body)
  queries = request_body.map do |query|
    query[:context] = query_context.merge(operation_id: query.delete(:operationId))
    query[:operation_name] = query.delete(:operationName)
    query[:variables] = (query[:variables].presence || {}).with_indifferent_access
    query
  end

  Graph::Schema.multiplex(queries)
end
```

Issue #1083, opened 2017-11-09
Instrumentation query hooks bug + teardown behaviour


swalkinshaw:
We're seeing some surprising behaviour when using query level instrumentation. This might have been a naive assumption, but we assumed that the sequence of `before_query` -> `after_query` would be reliable.

I've created a failing test case to demonstrate our situation: https://github.com/swalkinshaw/graphql-ruby/blob/45351d42dded3b682b59561b2ceb845d68b6d91c/spec/graphql/instrumentation_spec.rb

Summary:
* two instrumenters
* first one in the chain causes an exception during `before_query` (or maybe even during execution)
* second one relies on state from `before_query` in its `after_query` hook which causes an error
* `RuntimeError: bad teardown` gets raised instead of `NameError`.

The last point is especially confusing. The earlier exception is swallowed even though it happens first.

Basically, any instrumentation can't rely on state that's set in `before_query` being available in `after_query` if you have multiple instrumenters.

Here's an example in the wild: https://github.com/uniiverse/apollo-tracing-ruby/blob/master/lib/apollo_tracing.rb

This one relies on the existence of a key (`apollo-tracing`) in the query context. This could easily cause problems if you had another instrumenter.

Right now we've worked around this by ensuring any state we need actually exists but this seems like something the gem should ideally handle properly.

cc @eapache

swalkinshaw:
cc @exAspArk too since this affects apollo-tracing

swalkinshaw:
Thinking about this more, I don't think an instrumenter's `after_query` should depend on `before_query` running since there would definitely be use cases for using each independently.

So I'm not sure what or even if there's a solution to this. At a minimum the instrumentation docs should be updated to warn not to rely on state.

Beyond that, should an exception in one instrumenter's `before_query` stop the others from running? Same thing goes for `after_query`.

We could just rescue when calling each `before_query` than raise the first exception after they all run?

rmosolgo:
These are great questions. I'm open to a lot of the solutions here and I agree, it's a bit murky in any case!

I will think about this more tomorrow ... Happy to take a recommendation 😅

eapache:
> there would definitely be use cases for using each independently.

I'm curious what you're thinking of here?

> We could just rescue when calling each before_query than raise the first exception after they all run?

[...]

PR #1100, opened 2017-11-16
Better query instrumentation behavior

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/1083

Implements a instrumentation policy as documented here: 

https://github.com/rmosolgo/graphql-ruby/blob/dc7bd7c7b45fd882ca255e8bf85365ba1419a36e/lib/graphql/execution/instrumentation.rb#L5-L18

Also, the instrumentation stack is extracted to `GraphQL::Execution::Instrumentation`. 

You can see the _incremental_ changes to `GraphQL::Execution::Multiplex` (without the extraction) here: 

https://github.com/rmosolgo/graphql-ruby/compare/0307b8e...ce7e317


lib/graphql/execution/instrumentation.rb
```diff
@@ -0,0 +1,60 @@
...
+        # Let them be executed
+        yield
+      ensure
+        # Finally, run teardown instrumentation for each query + the multiplex
+        # Use `reverse_each` so instrumenters are treated like a stack
+        last_complete_query_idx = completely_instrumented_queries - 1
+        partial_query_idx = last_complete_query_idx + 1
+        queries.each_with_index do |query, idx|
+          if idx <= last_complete_query_idx
+            query_instrumenters.reverse_each { |i| i.after_query(query) }
```
dylanahsmith: If an exception happens here, then the following after_query hooks won't get run

rmosolgo: I took a try in dc7bd7c. It's reasonably clear behavior, but it's odd to ignore some errors. What do you think of that solution?

rmosolgo:
(updated the policy snippet in the description)

lib/graphql/execution/instrumentation.rb
swalkinshaw: 🤔  what's the alternative here? Creating a wrapper exception?

lib/graphql/execution/instrumentation.rb
```diff
@@ -0,0 +1,93 @@
...
+        raised_teardown_error = nil
+        queries.each_with_index do |query, idx|
+          if idx <= last_complete_query_idx
+            query_instrumenters.reverse_each { |i|
+              begin
+                i.after_query(query)
+              rescue
+                # It raised an error, but we promised to call the next one.
+                raised_teardown_error ||= $!
+              end
```
dylanahsmith: This is better, but it still hides error information that could be captured through `Exception#cause`.  What was wrong with the simple solution of using nested begin-ensure blocks as I suggested in https://github.com/rmosolgo/graphql-ruby/issues/1083#issuecomment-344382708 ?

rmosolgo: Woah, nothing is wrong with that, I missed the recursion-based sample! Let me give it a try.

lib/graphql/execution/instrumentation.rb
```diff
@@ -0,0 +1,82 @@
...
+        # Call the before_ hooks of each query,
+        # Then yield if no errors.
+        # `call_hooks` takes care of appropriate cleanup.
+        def each_query_call_hooks(instrumenters, queries, i = 0)
+          if i >= queries.length
+            yield
+          else
+            query = queries[i]
+            call_hooks(instrumenters, query, :before_query, :after_query) {
+              each_query_call_hooks(instrumenters, queries, i + 1) {
```
dylanahsmith: I didn't think about the fact that you were calling the query hooks for all queries before and after all queries, instead of around each query.  I guess that is because lazy execution is meant to overlap across all the queries.  Do you expect a large number of queries would often get multiplexed together?  Since this would increase the stack size by O(Q * I) where `Q` is the number of queries and `I` is the number of query instrumenters.

If this is a concern or does become a problem, then I thought of a more complicated solution could be used to collapse the stack.

```ruby
def call_after_hooks(instrumenters, object, after_hook_name, i)
  while i >= 0
    instrumenter = instrumenters.fetch(i)
    i -= 1
    instrumenter.public_send(after_hook_name, object)
  end
ensure
  # Call remaining hooks even if an exception or `throw` is used.
  # Exception#cause can be used to access all exceptions when multiple
  # after hooks raise an exception
  if i >= 0
    call_after_hooks(instrumenters, object, after_hook_name, i)
  end
end

def call_hooks(instrumenters, object, before_hook_name, after_hook_name, i=0)
  instrumenters.each do |instrumenter|
    instrumenter.public_send(before_hook_name, object)
    i += 1
  end
  yield
ensure
  call_after_hooks(instrumenters, object, after_hook_name, i - 1)
end
```

We don't multiplex multiple queries, so what you have right now works for us.

rmosolgo: Cool, I'll tuck this away and cross that bridge when we come to it!

Issue #1016, opened 2017-10-10
Relay RelationConnection makes two identical SQL queries


bgentry:
I traced this in detail on #1014. See [here](https://github.com/rmosolgo/graphql-ruby/pull/1014#issuecomment-335295869) and [here](https://github.com/rmosolgo/graphql-ruby/pull/1014#issuecomment-335302214).

When using Relay-style pagination via the `RelationConnection`, graphql-ruby makes two identical SQL queries to fetch the result page from the database. One of these happens within the `RelationConnection` implementation in `paged_nodes_array`, while the other happens when `EdgesInstrumentation#build-edges` ultimately calls `BaseConnection#paged_nodes` (returning a relation in this case) and then runs `map` on it, which causes it to re-query.

This was encountered with Sequel, not sure if it would behave differently with AR.

PR #1101, opened 2017-11-16
fix(RelationConnection) avoid unneeded query on Sequel

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/1016


rmosolgo:
jruby stuff, whatever

bgentry:
Awesome to see this fixed, thanks @rmosolgo 🙏

PR #1103, opened 2017-11-17
Make it possible to trace scalar types both per field and globally per schema

esjee:
A second attempt at #1061!

You can either specify it per field like so:

```
  # ...
  
  field :foo, types.String, trace: true
  field :bar, types.String, trace: false
```

Or, when not set explicitly (`trace= nil`), it will defer to the schema value, which can be set like so:

```
  # ...
 
  # This will cause fields that return a ScalarType or EnumType to be traced by default
  # default: false
  trace_scalar_fields_by_default true
```

Tested on the schema of Hacker0x01 :grin: 
Still trying to figure out how to properly test this :thinking:


rmosolgo:
I have a suggestion which might simplify the implementation and testing, what do you think of this ?

- Instead of storing the schema-level "trace scalars" value on the `GraphQL::Schema`, store it on the tracer instance:
  - Extend [`GraphQL::Tracing::Platform#initialize`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/platform_tracing.rb#L15-L17) tracing to accept `trace_scalars:`; store the value as an ivar 
  - Also accept `trace_scalars` in [`.use(...)`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/platform_tracing.rb#L52), pass it thru to `.new(...)`. 
- During `#instrument`, use the same logic you added, but you can access the ivar instead of the passed-in value. (I think this will allow us to remove the `.is_a?` check in `traversal.rb`?)
- Then, for testing, I think you could 
  - Test local overrides: 
    - Modify `Dummy::Schema`: add `trace: true` to a scalar
    - Add another test case to `platform_tracer_spec` which calls the updated field; it should have more tracing output than the [current test case](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/tracing/platform_tracing_spec.rb#L40-L54)
  - Test schema-level setting: 
    - Modify `Dummy::Schema`: add `trace: false` to a scalar
    - In another test, override the existing [`let(:schema)`](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/tracing/platform_tracing_spec.rb#L30-L34) adding a `trace_scalars: true` option to the tracer hookup. 
    - Make two more test cases: 
      - One that hits the new `trace: false` field; assert that the local override took precedence over the schema-level setting 
      - One that hits an unmodified scalar field; assert that the schema-level setting was applied (and the field was traced). (You could probably copy the existing test case for that; but it should emit more tracing output.)

What do you think, would that be feasible (or worth the effort) ?

esjee:
Oh hey, that's a good suggestion. Thanks @rmosolgo, I'll give it a try.

esjee:
I tried to implement your suggestion. Please let me know what you think!

lib/graphql/schema.rb
```diff
@@ -137,6 +137,7 @@ def initialize
       @subscription_execution_strategy = self.class.default_execution_strategy
       @default_mask = GraphQL::Schema::NullMask
       @rebuilding_artifacts = false
+      @trace_scalar_fields_by_default = false
```
rmosolgo: Can this ivar be removed now?

esjee: Yup! Sorry, looks like I forgot about this one.

Issue #1097, opened 2017-11-15
Add a warning that the new name override method is graphql_name


rmosolgo:
To override the name of a class-based type, the method is `graphql_name`, not `name`. 

We can override the `name` method to provide a helpful error

esjee:
Let me take a stab at this!

rmosolgo:
Fixed by #1104

PR #1104, opened 2017-11-17
Add convenience method "name" to point out that graphql_name exists

esjee:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/1097


rmosolgo:
Cool, thanks for taking care of this!

PR #1105, opened 2017-11-17
Fix typo in GraphQL::InternalRepresentation::Node#typed_children

chrisbutcher:
Fixes typo of instance variable name. It's only used for caching the result of `typed_children`, and the typo isn't referenced anywhere else, so no difference in behaviour.


rmosolgo:
Oopsie! Thanks for the fix, and good thing it didn't come with a high price 😅

PR #1110, opened 2017-11-18
[WIP] Adds transformer to upgrade schema to class based syntax

Willianvdv:
A port from https://github.com/Willianvdv/graphql_upgrader but using Minitest instead of Rspec. Not all specs are ported yet, there's some functionality missing, and I need to clean the Regex mess before I can remove the WIP from the title.

### DIFF
--- a/lib/graphql/upgrader/member.rb
+++ b/lib/graphql/upgrader/member.rb
@@ -1,0 +1,107 @@
+module GraphQL
+  module Upgrader
+    class Member
+      def initialize(member)
+        @member = member
+      end
+      def transform
+        return unless transformable?
+        transformable = member.dup
+        transformable = transform_to_class transformable
+        transformable = transform_or_remove_name transformable
+        transformable = put_the_field_name_and_return_type_and_description_on_one_line_for_easy_processing transformable
+        transformable = move_the_type_from_the_block_to_the_field transformable
+
+        transformable.scan(/(?:field|connection) .*$/).each do |field|
+          field_regex = /(?<field_or_connection>field|connection) :(?<name>[a-zA-Z_0-9]*)?, (?<return_type>.*?)(?<thing>,|$|\})(?<remainder>.*)/
+
+          if (matches = field_regex.match(field))
+            name = matches[:name]
+            return_type = matches[:return_type]
+            remainder = matches[:remainder]
+            thing = matches[:thing]
+            field_or_connection = matches[:field_or_connection]
+
+            # Someone that knows regular expressions probably would do this in the regex, I don't
+            doable = remainder.gsub!(/\ do$/, '') || return_type.gsub!(/\ do$/, '')
+
+            # Remove the old null type indicator
+            nullable = return_type.gsub! '!', ''
+
+            # Remove the types
+            return_type.gsub! 'types.', ''
+            return_type.gsub! 'types[', '['
+
+            nullable_as_keyword = nullable ? ', null: true' : ''
+            connection_as_keyword = field_or_connection == 'connection' ? ', connection: true' : ''
+
+            transformable.sub!(field) do
+              "field :#{name}, #{return_type}#{thing}#{remainder}#{nullable_as_keyword}#{connection_as_keyword}#{doable ? ' do' : ''}"
+            end
+          end
+        end
+
+        transformable
+      end
+
+      private
+
+      def move_the_type_from_the_block_to_the_field(transformable)
+        transformable.gsub(
+          /(?<field>(?:field|connection) :(?:[a-zA-Z_0-9]*)?) do(?<block_contents>.*?)([ ]*)type (?<return_type>.*?)\n/m
+        ) do
+          field = $~[:field]
+          block_contents = $~[:block_contents]
+          return_type = $~[:return_type]
+
+          "#{field}, #{return_type} do#{block_contents}"
+        end
+      end
+
+      def put_the_field_name_and_return_type_and_description_on_one_line_for_easy_processing(transformable)
+        transformable.gsub(/(?<field>(?:field|connection).*?),\n(\s*)(?<next_line>"(.*))/) do
+          field = $~[:field].chomp
+          next_line = $~[:next_line]
+
+          "#{field}, #{next_line}"
+        end
+      end
+
+      def transform_to_class(transformable)
+        transformable.sub(
+          /([a-zA-Z_0-9]*) = GraphQL::(Object|Interface|Enum)Type\.define do/, 'class \1 < Base\2'
+        )
+      end
+
+      def transform_or_remove_name(transformable)
+        if (matches = transformable.match(/class (?<type_name>[a-zA-Z_0-9]*) < Base(Object|Interface|Enum)/))
+          type_name = matches[:type_name]
+          type_name_without_the_type_part = type_name.gsub(/Type$/, '')
+          name = transformable.match(/name ('|")(?<type_name>.*)('|")/)[:type_name]
+        end
+
+        if type_name_without_the_type_part != name
+          transformable = transformable.sub(/name (.*)/, 'graphql_name \1')
+        else
+          transformable = transformable.sub(/\s*name ('|").*('|")/, '')
+        end
+
+        transformable
+      end
+
+      def transformable?
+        return false if member.include? '< GraphQL::Schema::'
+        return false if member.include? '< BaseObject'
+        return false if member.include? '< BaseInterface'
+        return false if member.include? '< BaseEnum'
+
+        true
+      end
+
+      private
+
+      attr_reader :member
+    end
+  end
+end
+
--- a/lib/generators/graphql/upgrader.rb
+++ b/lib/generators/graphql/upgrader.rb
@@ -1,0 +1,7 @@
+module Graphql
+  module Upgrader
+    class GraphQLMember
+    end
+  end
+end
+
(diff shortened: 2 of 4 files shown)


rmosolgo:
This is awesome! Glad to see those tests ported. 

I think Rake tasks are a great idea. What do you have in mind there? I wonder if we could get really fancy and "just" 

```bash
rake graphql:update app/graphql 
# - Visit each .rb file 
# - Determine whether or not a transformation applies to it 
# - Perform a transformation, if one applies 
```

What do you think?

Willianvdv:
@rmosolgo yeah, let's do that!

esjee:
How about this?

```
rake graphql:update some/dir
# do exactly what @rmosolog posted above

rake graphql:update some/file
# Attempt to perform a transformation on the provided file
```

While only a small change, the possibility to add a file greatly increases the flexibility of the rake task. You could piece together your own little bash script to feed files to the upgrade in case your directory structure isn't supported, or if you have multiple classes defined in a single file.

rmosolgo:
That's a good point, perhaps [Dir.glob](https://ruby-doc.org/core-2.4.2/Dir.html#method-c-glob) would be a good choice under the hood.

Willianvdv:
I'll add the nullability to all the fields so:
```ruby
field :name, !String
field :maybeName, String
```

Becomes:
```ruby
field :maybeName, String, null: true
```

Willianvdv:
More planned features!
```
field :name, types.String, property: :id
```
Becomes
```
field :name, String, null: false, method: :id
```

I don't know how `upcase` works just yet, but looking at the specs I should do something like
```
field :name_name, types.String
field :nameName, types.String
```
Becomes
```
field :nameName, String, null: false, upcase:true
field :nameName, String, null: false, upcase:true
```

B

rmosolgo:
:clap: for `method:`, but `upcase:` is not really a library feature. It's used as an example of how fields can be customized.

Regarding camelized/underscored field names, the goal is to: 

- write `field :field_name` (underscore)
- expose `TypeName.fieldName` (camel)
- call the `:field_name` method or hash key (underscore) 
- but, allow this to be customizable, so you can expose underscore-case names if you want 😉 

see https://github.com/rmosolgo/graphql-ruby/pull/1126

Willianvdv:
I'd like to split up the rake task up into three smaller ones.
```bash
bin/rake graphql:upgrade[app/graphql] # Upgrades everything, accepts the base directory
bin/rake graphql:upgrade:schema[app/graphql/schema.rb] # Upgrades the schema
bin/rake graphql:upgrade:member[app/graphql/types/user_type.rb] # Upgrades the user type
bin/rake graphql:upgrade:member[app/graphql/types/] # Upgrades all the members within the given directory
```
This allows me to have one command that just upgrades everything. It's a bit easier to, for example, create the new base type/enum/etc if I know the base path of your GraphQL definitions. Still, I like to upgrade types one by one; hence the two other tasks.

rmosolgo:
I think this is looking 👌 🔥 , is there anything else on your to-do list? 

The `rake` API you implemented here looks good to me. 

I noticed the current implementation uses `Railtie` to add `rake_tasks` automatically. This is great for Rails users, but should (could?) we also support non-Rails users? 

Something like [`GraphQL::RakeTask`](http://www.rubydoc.info/gems/graphql/1.7.6/GraphQL/RakeTask) would work, where you `require ...` + initialize in order to install tasks. (It's ~~inspired by~~ ripped off from [`Rubocop::RakeTask`](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/RakeTask) and [`Rake::TestTask`](http://www.rubydoc.info/gems/rake/Rake/TestTask.) Do you think we could add something like `Upgrader::RakeTask` in case a Sinatra or Hanami app wants an upgrade?

esjee:
> I noticed the current implementation uses Railtie to add rake_tasks automatically. This is great for Rails users, but should (could?) we also support non-Rails users?
> 
> Something like GraphQL::RakeTask would work, where you require ... + initialize in order to install tasks. (It's inspired by ripped off from Rubocop::RakeTask and Rake::TestTask Do you think we could add something like Upgrader::RakeTask in case a Sinatra or Hanami app wants an upgrade?

While I'm absolutely not opposed to supporting those users, I don't think it should be something that blocks this PR from being merged. I think this tool belong in the same bucket as [graphiql](https://github.com/rmosolgo/graphiql-rails) in the sense that it's nice to have but not absolutely essential tools. 

If there are experienced Sinatra/Hanami users who would be interested in this feature for their framework, I think it's best to leave it up to them to write a PR for it, using @Willianvdv's work here as a base.

I guess what I'm saying: this thing has been in progress for a while already, and it finally appears to be in a usable state. Let's merge it, so we can expose it as early as possible, and gather feedback on edge cases, because _surely_ we missed some of those.

Does this make sense to you, @rmosolgo @Willianvdv ? :smile:

lib/graphql.rb
```diff
@@ -3,6 +3,7 @@
 require "json"
 require "set"
 require "singleton"
+require_relative "./graphql/railtie" if defined? Rails::Railtie
```
esjee: `::Rails::Railtie`? I mean, it's probably not necessary, but you never know.

lib/graphql/railtie.rb
```diff
@@ -0,0 +1,83 @@
...
+require_relative './upgrader/schema'
+
+module GraphQL
+  class Railtie < Rails::Railtie
+    rake_tasks do
+      namespace :graphql do
+        task :upgrade, [:dir] do |t, args|
+          unless (dir = args[:dir])
+            fail 'You have to give me a directory where your GraphQL schema and types live. ' \
+             'For example: `bin/rake graphql:upgrade[app/graphql/**/*]`'
```
esjee: `bin/rake` could be `bin/rails` in Rails 5, right? :sweat:

lib/graphql/upgrader/member.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      def upgrade
+        transformable = member.dup
+        transformable = transform_to_class transformable
+        transformable = transform_or_remove_name transformable
+        transformable = simplify_field_definition_for_easier_processing transformable
+        transformable = move_the_type_from_the_block_to_the_field transformable
+        transformable = rename_property_to_method transformable
+
+        transformable.scan(/(?:field|connection|argument) .*$/).each do |field|
```
esjee: I think it would improve legibility if you move this entire scan portion into a method called something along the lines of `transform_fields_and_connections_and_arguments`. Then it's on the "same level" as the methods above it.

Just a minor comment though; it's your code, I trust you.

rmosolgo:
Good point, and ok by me!

spec/graphql/upgrader/schema_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      class StarWarsSchema < GraphQL::Schema
+        def self.id_from_object(object, type_definition, ctx)
+          nil
+        end
+      end
+    }
+
+    assert_equal upgrade(old), new
+  end
+
```
esjee: Unintentional extra blank lines?

lib/graphql/railtie.rb
Willianvdv: Yeah, but we want to support Rails 4 and even 3 (looking at you, Github)

spec/graphql/upgrader/schema_spec.rb
Willianvdv: Yes, I'll remove them

lib/graphql/upgrader/member.rb
Willianvdv: I like the suggestion, let me try

Willianvdv:
I did figure out some issues I have to fix: 
- [ ] There's no task yet to upgrade all the members, just one
- [ ] `graphql:upgrade` only updates the members, I didn't add the schema migration and the creation of the base types.

rmosolgo:
I'm hoping to cut a new 1.8 beta today, would you like me to merge this with its progress so far? Then you could follow up on those other tasks before 1.8 "goes live."

Otherwise, happy to leave this PR open until you think it's totally ready!

Willianvdv:
@rmosolgo lets merge! The two open issues are nice-to-haves and not blocking, I'll file a new PR to fix them.

rmosolgo:
Another good follow-up would be to add docs in the `class_based_api.md` guide or in a separate guide, so we can point people to it

swalkinshaw:
Since 1.8 now handles/defaults to underscored field and argument names, I'm guessing the transformers should do that conversion as well?

Likely at this point for fields: https://github.com/rmosolgo/graphql-ruby/blob/c39c352e7e2e68887b02e9e0619d7c7846153b33/lib/graphql/upgrader/member.rb#L22

Willianvdv:
@swalkinshaw like #1126? definitely! Let me PR that!

swalkinshaw:
Yep 👍. Just to be confirm: if someone had camelCased names in the old DSL, these transforms should convert them to underscored.

esjee:
@Willianvdv @swalkinshaw Couldn't that possibly cause clients to break? For example, a relay client? 

I still think it's a worthwhile suggestion if so, but perhaps it should be made configurable (camelCaseNames: true|false)? :thinking:

swalkinshaw:
@esjee the names remain camelCased in the schema though. What it *would* break is anyone who actually had and wanted to keep their names underscored (or some other inconsistent casing).

You can override this behaviour but you're right that the transformers should probably have an option.

PR #1111, opened 2017-11-19
Splits spec's dummy schema into seperate files

Willianvdv:
For #1110's rake task I'd like to use some real files to test against. Unfortunately, in the specs the schema and all its type definitions are defined in one file. My thinking was, (hopefully) 
 this isn't how you write them in your app so let's refactor the spec's schema file into something more real-life like. It gives me the ability to use them in #1110 and it makes this dummy schema a bit more readable.

What do you think? This PR isn't ready yet, to finish it I'd like to first get your feedback before spending too much time on it.


cjoudrey:
I haven't fully reviewed your changes, but I think I agree that splitting the schema up into multiple files makes it a bit more readable.

Somewhat unrelated to this PR, my issue with this schema is that it is pretty big and we keep tacking more stuff onto it. I sometimes find it hard to follow, especially when working on the `static_validation/rules` specs. Many months ago, I was toying with the idea of using `GraphQL::Schema.from_definition` to have a small schema side-by-side with each test (or test file) instead of relying on this massive schema. I think it would make it easier to understand what a test is doing.

PR #1112, opened 2017-11-19
Fix call to to_query_string in Language::Generation

xuorig:
Looks like this one slipped through when refactored.

`to_query_string` does not take `indent` as an argument. Instead, call generate on an abstract node.

On another note, when would we get an abstract node here?


rmosolgo:
Hmm, this was _always_ broken, right? So it's certainly untested, and probably completely unused. Can we remove it altogether?

xuorig:
Removing in https://github.com/rmosolgo/graphql-ruby/pull/1118

PR #1114, opened 2017-11-20
AST from Schema

xuorig:
This PR implements `Language::Nodes::DocumentFromSchema` which takes a Schema definition and turns it into a GraphQL Document AST.

This is a step towards making `Schema::Printer` and `Language::Generation` use the same logic.

Exposes:

  - `Schema#idl_document`
  - `Member.to_ast_node`
  - `Type.to_ast_type_name_node`
  - `GraphQL::Language::Nodes.from_schema`
  - `GraphQL::Language::Nodes::DocumentFromSchema`


lib/graphql/argument.rb
```diff
@@ -94,6 +94,16 @@ def prepare=(prepare_proc)
       @prepare_proc = BackwardsCompatibility.wrap_arity(prepare_proc, from: 1, to: 2, name: "Argument#prepare(value, ctx)")
     end
 
+    def to_ast_node
```
xuorig: Are these dangerous to memoize? I wouldn't think so with `redefine` but maybe 🤔 ?

xuorig:
Started working on this, let me know what you think of the approach. Missing some more tests!

lib/graphql/argument.rb
rmosolgo: The issue is that the _ast nodes_ are mutable. There's really no way to work with the AST right now except to modify nodes, so the cache's contents could become invalid. Can this be a "cross that bridge when we come to it" feature? Or, is there a need to cache the output now?

xuorig:
@rmosolgo my first integration on this was exactly what you describe. I decided to keep it close to definitions because it felt really internal to those and they would never change, but with:

> , if someone wants to print an IDL with custom directives mixed in

I think you're right that the first approach, just like the one we chose for the printer, seems to make more sense.

I'll give another shot at this using this approach!

xuorig:
Closing this in favor of another approach!

PR #1117, opened 2017-11-20
[Class based API] Check for nil before checking depth

phyrog:
Fixes ``undefined method `map' for nil:NilClass`` for nullable lists of an Interface when resolving to `nil`

I'm not sure what the best location for a test for this is, since there is no `spec/graphql/schema/members/instrumentation_spec.rb` and the bug affects `Union`s and `Interface`s, but also `List`s.

Anyway, for now here's how to reproduce the bug:

```ruby
require 'graphql'

class Foo < GraphQL::Schema::Interface

  def name
    "something"
  end
end

class Bar < GraphQL::Schema::Object
  implements Foo
end

  field :foos, [Foo], null: true

  def foos
    nil
  end
end

  query Query
  orphan_types [Bar]
end

Schema.execute("query { foos { name } }")
```

### DIFF
--- a/lib/graphql/schema/member/instrumentation.rb
+++ b/lib/graphql/schema/member/instrumentation.rb
@@ -85,10 +85,10 @@ module GraphQL
           private
 
           def proxy_to_depth(obj, depth, type, ctx)
-            if depth > 0
-              obj.map { |inner_obj| proxy_to_depth(inner_obj, depth - 1, type, ctx) }
-            elsif obj.nil?
+            if obj.nil?
               obj
+            elsif depth > 0
+              obj.map { |inner_obj| proxy_to_depth(inner_obj, depth - 1, type, ctx) }
             else
               concrete_type = case type
               when GraphQL::UnionType, GraphQL::InterfaceType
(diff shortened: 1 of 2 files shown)


rmosolgo:
Wow, thanks for hunting this down! I appreciate the reproduction code. We should definitely include a test so we can keep this working in the future 😅 

Right now, this new API is mostly tested via integration tests, using this dummy schema: 

https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/spec/support/jazz.rb

So, if you don't mind adding a test, I think you could go one of at least two ways:

- Port your example code to the schema above, then add a test to one of the existing files (or add `instrumentation_spec.rb`, as you named above). 
- Create a new test file, `instrumentation_spec.rb`, and put your exact replication code above into that file. Then add a successful test case, so that we can maintain this behavior in the future. (In fact, you'd probably have to put your example code into a namespace so that it doesn't conflict with other test fixtures.) 

How do those sound, do you want  to take a stab at it?

phyrog:
I opted for the second option; I think it would have taken waay too long for me to come up with some example that would fit the musical theme :smile:

rmosolgo:
Weird, I dunno why CI didn't trigger on that last commit. Anyways, looks ok to me, I guess we'll see how it does after a merge! 

Thanks again for the fix on this one!

Issue #1106, opened 2017-11-17
Implement Query#sanitized_query_string


rmosolgo:
We want to improve logging, but don't leak private info into the logs. So, let's take a [page out of Shopify's book](https://graphql.slack.com/archives/C07MVL1AB/p1510945531000387) (again 😊 ) and add a custom printer API. 

Let's add a couple of steps: 

- Implement `class GraphQL::Language::Printer` which has `print_*` methods corresponding to each node type; this is the base class for custom printers. 
- Extend `GraphQL::Language::Nodes::AbstractNode#to_query_string(printer: GraphQL::Language::Printer)`
- Add `Schema#sanitizing_printer_class`; accept options via `sanitize(whitelist:, blacklist:, ... ?)`
- Add `Query#sanitized_query_string` which prints the document with the sanitization settings. 

(I'm not sure I've got the right split of which methods go where; I'm open to reshuffling that, for sure. But I think it'd be nice to support an _injected_ printer and a base implementation for printers.)

__TODO:__ What does it look like to integrate this with Rails' [parameter filtering](http://api.rubyonrails.org/classes/ActionDispatch/Http/FilterParameters.html)?

If the approach works, maybe we can do something similar for https://github.com/rmosolgo/graphql-ruby/issues/791

xuorig:
Working on implementing a prototype of this, will take the opportunity to refactor `GraphQL::Schema::Printer` and `Language::Generation` which implement the same logic twice in a number of places 👍

swalkinshaw:
When I was working on our query filterer, I did wonder about moving all the string generation into each nodes class. Is it still a goal to keep these separated?

Right now only `AbstractNode` has `to_query_string`, but every node type could too?

theorygeek:
Just for posterity, thought I'd add this code that @rmosolgo referenced, because the GraphQL Slack only lets you see the last 10k messages:

```ruby
module FilteredLanguageGeneration
  include GraphQL::Language::Generation
  extend self

  def generate(node, indent: "")
    case node
    when GraphQL::Language::Nodes::Argument
      filtered_values = FilterParameters.filtered_graphql_values(node.name => node.value)
      "#{node.name}: #{generate(filtered_values[node.name])}"
    when GraphQL::Language::Nodes::InputObject
      generate(FilterParameters.filtered_graphql_values(node.to_h))
    else
      super
    end
  end
end
```

xuorig:
[...]

PR #1118, opened 2017-11-20
GraphQL::Language::Printer

xuorig:
This PR implements `GraphQL::Language::Printer` as discussed in https://github.com/rmosolgo/graphql-ruby/issues/1106

  - `Language::Generation` now delegates to to `Printer`.

  - `Printer` serves as a base type for custom printer implementations.

  - `to_query_string` now takes an optional `printer` param, which defaults to `Language::Printer`

Here's an example of how the Shopify folk achieve custom query strings:

```ruby
module FilteredLanguageGeneration
  include GraphQL::Language::Generation
  extend self

  def generate(node, indent: "")
    case node
    when GraphQL::Language::Nodes::Argument
      filtered_values = FilterParameters.filtered_graphql_values(node.name => node.value)
      "#{node.name}: #{generate(filtered_values[node.name])}"
    when GraphQL::Language::Nodes::InputObject
      generate(FilterParameters.filtered_graphql_values(node.to_h))
    else
      super
    end
  end
end
```

With this PR, this can be implemented in the following way:

```ruby
class FilteredPrinter < GraphQL::Language::Printer
  def print_argument(node)
    filtered_values = FilterParameters.filtered_graphql_values(node.name => node.value)
    "#{node.name}: #{generate(filtered_values[node.name])}"
  end

  def print_input_object(node)
    print_node(FilterParameters.filtered_graphql_values(node.to_h))
  end
end
```

```ruby
query.document.to_query_string(printer: FilteredPrinter)
```


xuorig:
@rmosolgo, @swalkinshaw what do you think of this printer API?

swalkinshaw:
The new version of the customized printer is a big improvement 👍

lib/graphql/language/generation.rb
```diff
@@ -15,188 +15,8 @@ module Generation
...
-      end
-
-      def generate_field_definitions(fields)
-        out = " {\n".dup
-        fields.each.with_index do |field, i|
-          out << generate_description(field, indent: '  ', first_in_block: i == 0)
-          out << "  #{generate(field)}\n"
-        end
-        out << "}"
+      def generate(node, indent: "", printer: GraphQL::Language::Printer)
```
cjoudrey: A nitpick, but we might want to document the `printer` param above.

cjoudrey:
Any thoughts on deprecating `GraphQL::Language::Generation` in favor of `Node#to_query_string` and `Printer#print`?

spec/graphql/language/printer_spec.rb
```diff
@@ -0,0 +1,203 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe GraphQL::Language::Printer do
```
rmosolgo: Is this ported from `generation_spec.rb`? If so, do you think we can remove that file without losing any coverage? (If so, :+1: from me, so don't leave dead code around.)

xuorig: this is ported from generation_spec yes! I think coverage is the same with or without this test. I was keeping this one because I was scared of forgetting to test the printer if generation is ever deprecated / 🔥 🤔

rmosolgo: Sorry, it seems like I didn't express myself well 😬 

I think we should _keep_ `printer_spec.rb`, since a lot of logic was moved into `printer.rb`. 

I also think we should _reduce_ `generation_spec.rb`, since a lot of logic was taken _out_ of `generation.rb`. Maybe it just needs one test, that it uses the passed-in printer, but to be honest, it could probably just be removed entirely. 

(I agree with 

> scared of forgetting to test

!)

xuorig: Ah I see what you mean now! Agreed!

xuorig:
@rmosolgo bunch of docs updated! Ready to go if there's no other comments

xuorig:
Changed the API a bit after realising this isnt the best for custom printers.

Instead of passing the printer class to `to_query_string`, we instead pass an instance. This allows custom integrations to do:

```
printer = MyVariableScrubberPrinter.new(blacklist: [], whitelist: [])
document.to_query_string(printer: printer)
``

**Comment by xuorig** [comment]
Just had another thought, which may be discussed /  implemented after this ships.

Should we have a bit more information passed around in the printer? Suppose I want to custom print variables but it depends on its parent. Currently, you only get access to an `Argument` node, which has no information on its parent.

Should we maybe pass around a `printer_context` object, or an extra parent argument to all print methods?

File: lib/graphql/language/generation.rb [hunk 3]
```diff
@@ -14,189 +14,10 @@ module Generation
...
+        printer.new(node).print(indent: indent)
```
rmosolgo: Hmm, maybe this code isn't tested? 

I see that `Printer` is initialized with `node`, but `#initialize` was removed in https://github.com/rmosolgo/graphql-ruby/pull/1118/commits/0b9ade35123286354faa351d34abdb3ec420d7fe#diff-c8deb24bd9b63b4eec0de9a19914a038L26. Can this API also accept a printer _instance_ now? I think you were 👌  on that suggestion.

lib/graphql/language/nodes.rb
```diff
@@ -8,6 +8,8 @@ module Nodes
       # - `children` returns all AST nodes attached to this one. Used for tree traversal.
       # - `scalars` returns all scalar (Ruby) values attached to this one. Used for comparing nodes.
       # - `to_query_string` turns an AST node into a GraphQL string
+      DEFAULT_PRINTER = GraphQL::Language::Printer.new
```
rmosolgo: Can we get away with making a fresh `Printer` every time? I appreciate the thriftiness of reusing one, but I've also seen people do some funny stuff because they don't know that objects are long-lived. If it's not a perf concern, can we do `printer: GraphQL::Language::Printer.new` in the method signature below?

lib/graphql/language/printer.rb
```diff
@@ -0,0 +1,338 @@
+# frozen_string_literal: true
+module GraphQL
+  module Language
+    class Printer
+      # Turn an arbitrary AST node back into a string.
+      #
+      # @example Turning a document into a query string
+      #    document = GraphQL.parse(query_string)
+      #    GraphQL::Language::Printer.new(document).print
```
rmosolgo: This doc needs an update? (To `.new.print(document)` ? Also below, if so.)

xuorig:
Addressed your comments @rmosolgo.

WDYT of https://github.com/rmosolgo/graphql-ruby/pull/1118#issuecomment-347656857? I've got a feeling we're going to need that for a variable scrubber.

rmosolgo:
The "nice" thing about a class is that we can hack together ivars for that, passing the dependency _implicitly_ instead of _explicitly_, which is sad, but hackable. 

So, if we want to add a `node_stack`, I think we can hack one: 

```ruby 
class PrinterWithStace < Printer 
  def initialize 
    super 
    @node_stack = [] 
  end 

  private 

  def print_node(n) 
    @node_stack << n 
    super 
    @node_stack.pop
  end 
end 
```

Then we could reach for `node_stack` when we need to backtrack. 

If we needed to provide some other extensible API, we could probably build from there, making two-argument methods that take the stack: 

```ruby
def print_argument_with_stack(node, stack)
  print_argument(node) 
end 
```

So, I think the current implementation is a solid, workable starting point!

(And TBH, I don't quite see the connection to scrubbing, can you elaborate? (Or, elaborate in another PR, when it comes needed?))

xuorig:
(I was seeing a connection to scrubbing to differentiate a `CreditCard.number` vs `Phone.number`)

You're right that this can probably be hacked together if needed. We can YAGNI this for now.

PR is ready I believe then :)!

rmosolgo:
Ohh, I see what you mean now! 🤔  I guess this is two cases to scrub: 


```graphql
mutation { 
  # Scrub the value here:
  updateCreditCard(number: "0040 0040 0040 0040")
}
```

```graphql 
# Scrub the default value here?
mutation($newNumber: ID = "0040 0040 0040 0040") { 
  updateCreditCard(number: $newNumber)
}
```

How would that be implemented here? 

Here's another thought: at some point, we need a way to convert a `GraphQL::Query` back to a string. I think a good way to do it is to re-convert `irep_selection` into a flattened, normalized Document. At that time, we could replace some literal values with `FILTERED`, etc, so, maybe we can solve the issue that way.

rmosolgo:
IMO this is a win either way, happy to merge this as-is and sort out the details later, or happy to keep this open while we work out next steps, what do you prefer?

xuorig:
Let's merge this first :) It'll help refactoring `Schema::Printer` and we can discuss scrubbing / improving the API in the meantime?

rmosolgo:
:tada: !

Issue #1099, opened 2017-11-16
`rails destroy graphql:mutation` removes mutations folder and mutation_type.rb


cpunion:
Steps:

```console
MBP:backend lijie$ rails g graphql:mutation Foo
Running via Spring preloader in process 91374
       exist  app/graphql/mutations
      create  app/graphql/mutations/.keep
      create  app/graphql/types/mutation_type.rb
add_root_type  mutation
      create  app/graphql/mutations/foo.rb
        gsub  app/graphql/types/mutation_type.rb

MBP:backend lijie$ rails g graphql:mutation Bar
Running via Spring preloader in process 91378
       exist  app/graphql/mutations
   identical  app/graphql/mutations/.keep
        skip  app/graphql/types/mutation_type.rb
add_root_type  mutation
      create  app/graphql/mutations/bar.rb
        gsub  app/graphql/types/mutation_type.rb

MBP:backend lijie$ rails d graphql:mutation Bar
Running via Spring preloader in process 91567
      remove  app/graphql/mutations
      remove  app/graphql/mutations/.keep
      remove  app/graphql/types/mutation_type.rb
add_root_type  mutation
      remove  app/graphql/mutations/bar.rb
```

Expects: Don't remove mutations folder and mutation_type.rb (At least keep it if I have some mutation types)

Willianvdv:
@cpunion interesting! I never heard about the `rails destroy` (http://guides.rubyonrails.org/command_line.html#rails-destroy) command before.

Do you think you can send in a PR to get this fixed?

cpunion:
@Willianvdv Yay, I will try.

thefliik:
Uggg. I had to update the generator of another gem I use to properly accommodate `rails destroy`. I found the documentation surrounding `rails destroy` / generators in general to be rather poor and the whole process took me much longer than it should have. Let me know, @cpunion, if you run into issues figuring it out. I _may_ be able to help.

cpunion:
@thefliik Awesome. I feel I can't figure out how to make the generators full reversible quickly. So I'm very happy if you can fix it or help. 😃

thefliik:
I'll try to look into it this weekend. Anyone know if there's a way for me to mark this issue as "unread" / make it show up in Github's notifications panel so that I don't forget about it?

Willianvdv:
@thefliik no, but I can ping you ;)

thefliik:
[...]

PR #1119, opened 2017-11-20
fix: rails destroy graphql:mutation

thefliik:
This patch fixes issue #1099 
Now, if you run `rails g graphql:mutation TestMutation` you'll see

```
       create  app/graphql/mutations
       create  app/graphql/mutations/.keep
       create  app/graphql/types/mutation_type.rb
add_root_type  mutation
       create  app/graphql/mutations/test_mutation.rb
         gsub  app/graphql/types/mutation_type.rb
```
And if you then run `rails d graphql:mutation TestMutation` you'll see

```
        gsub  app/graphql/types/mutation_type.rb
      remove  app/graphql/mutations/test_mutation.rb
```
`app/graphql/types/mutation_type.rb` will also be updated to remove the appropriate line of code (rather than removing the file entirely).

Some comments / tips for people trying to create rails generators in the future (because generators aren't well documented)

1. When a generator is run with `destroy`, rails automatically runs the `generate` version of the generator, sees what files would be created, and instead removes those files. Oftentimes, this results in the correct destroy functionality and so devs get `rails destroy` for free. This strategy for building `rails destroy` isn't well documented however, so figuring this out the first time can be rather confusing (at least it was for me).
2. In the cases where the default `rails destroy` behavior isn't working, in order to get it working you need to update the `rails generate` code to behave differently if the code is being run with `@behavior == :revoke`. `@behavior` is an instance variable that `rails` automatically sets for you. This also wasn't very intuitive for me, because I wouldn't naturally guess that fixing `rails destroy` involved patching `rails generate` (I would have expected it to involve creating a separate `destroy` method). 
3. Most of the generator functionality is provided by the `thor` gem, which rails' generator code relies heavily upon. Thor methods are automatically added with `require 'rails/generators/named_base'` and provide a lot of functionality out of the box. In this case [`inject_into_file` is a Thor action](http://www.rubydoc.info/github/wycats/thor/Thor%2FActions:insert_into_file) and it automatically finds and removes the appropriate code when run with `@behavior == :revoke`

### DIFF
--- a/lib/generators/graphql/mutation_generator.rb
+++ b/lib/generators/graphql/mutation_generator.rb
@@ -27,7 +27,12 @@ module Graphql
       attr_reader :file_name, :mutation_name, :field_name
 
       def create_mutation_file
-        create_mutation_root_type unless @behavior == :revoke
+        unless @behavior == :revoke
+          create_mutation_root_type
+        else
+          log :gsub, "#{options[:directory]}/types/mutation_type.rb"
+        end
+        
         template "mutation.erb", "#{options[:directory]}/mutations/#{file_name}.rb"
 
         sentinel = /name "Mutation"\s*\n/m
--- a/lib/generators/graphql/mutation_generator.rb
+++ b/lib/generators/graphql/mutation_generator.rb
@@ -27,7 +27,7 @@ module Graphql
       attr_reader :file_name, :mutation_name, :field_name
 
       def create_mutation_file
-        create_mutation_root_type
+        create_mutation_root_type unless @behavior == :revoke
         template "mutation.erb", "#{options[:directory]}/mutations/#{file_name}.rb"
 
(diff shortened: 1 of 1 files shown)


cpunion:
Wow! Very clear. Thanks your great work and detailed explanation.

rmosolgo:
Wow, thanks for hunting this down _and_ documenting your findings. Rails generators are so powerful, I wish it was easier to learn all the features! 

~~Since we properly support this now, do you mind adding a test for this behavior? I think an easy way to do so would be to~~

I can't find a way to run `rails destroy` from a test 😖  I guess we should merge it without a test, and just _hope_ it doesn't break 😖

thefliik:
I have no experience testing generators myself, unfortunately.

rmosolgo:
Well, at least we're smarter from your good docs above :) We'll have to skip it for now. Thanks for the improvement here!

PR #1126, opened 2017-11-23
[1.8] Handle underscored field names

### LINKED ISSUE #1124: Class-based API: field name casing
The class-based API provides a bunch of automatic `underscore` transformations, but field names are still passed through directly preserving original case.

There's a lot of subjective arguments on both sides. Should the gem even provide a solution to this?

Simplest solution could be a few pre-defined options on the schema:

* preserve/nothing
* camelcase
* underscore

Or if there's a need it could even take a transformation proc.

However, with the class-based API, another solution is just overriding `field` in your own base `Object` and `InputObject` class and the gem doesn't have to do anything.

swalkinshaw:
Closes #1124

Here's a quick attempt at supporting underscored field names.

I think this is the easiest place to handle it? The only question is how opinionated the gem should be on this. This means that only camelCased field names will end up in the schema.

We could allow this to be configuration in the same way as `field_class`.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -47,7 +47,8 @@ module GraphQL
         else
           GraphQL::Field.new
         end
-        field_defn.name = @name
+
+        field_defn.name = camelize(name)
 
         if @return_type_expr
           return_type_name = Member::BuildType.to_type_name(@return_type_expr)
@@ -94,6 +95,16 @@ module GraphQL
         end
 
         field_defn
+      end
+
+      private
+
+      def camelize(string)
+        return string unless string.include?('_')
+
+        string.split('_').map(&:capitalize).join.tap do |camelized|
+          camelized[0] = camelized[0].downcase
+        end
       end
 
       class << self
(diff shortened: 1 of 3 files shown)


rmosolgo:
:+1: Thanks for taking a crack at it! And the testing approach seems good to me, just ... make sure it works as expected. 

I have another thought about where the transformation could be applied: what about `GraphQL::Schema::Field#to_graphql`? We could say `field_defn.name = camelize(@name)` [here](https://github.com/swalkinshaw/graphql-ruby/blob/48258898cd5ffd0f0aa68cfd5a328b5adb574279/lib/graphql/schema/field.rb#L50). That way, if people want to customize it, they can do so in the _already-documented_ extension point, instead of adding a new one (they could override `to_graphql` to re-assign `.name=`, without the transformation).  Do you think it would work, or does it seem like an improvement?

swalkinshaw:
I think that's a much better idea since it also solves the question of how to customize the naming behaviour. I forgot that `fields` is an array and not a hash keyed on the field name so I thought it mattered to transform it there too.

Updated 👍

rmosolgo:
💰 !

PR #1127, opened 2017-11-23
Increase flexibility when generating functions and loaders through the CLI

Illianthe:
### Overview

This enforces camel-casing of class names when generating functions and loaders. It also allows them to be namespaced if one wants to structure them that way.

### Examples

Running `rails g graphql:function find_records` generates the following in `app/graphql/functions/find_records.rb`:
```
class Functions::FindRecords < GraphQL::Function
  ...
end
```

Running `rails g graphql:function finders::find_records` generates the following in `app/graphql/functions/finders/find_records.rb`:
```
class Functions::Finders::FindRecords < GraphQL::Function
  ...
end
```

### DIFF
--- a/lib/generators/graphql/function_generator.rb
+++ b/lib/generators/graphql/function_generator.rb
@@ -11,7 +11,7 @@ module Graphql
       source_root File.expand_path('../templates', __FILE__)
 
       def create_function_file
-        template "function.erb", "#{options[:directory]}/functions/#{file_name}.rb"
+        template "function.erb", "#{options[:directory]}/functions/#{file_path}.rb"
       end
     end
   end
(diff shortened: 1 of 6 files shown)


rmosolgo:
This looks like a nice improvement, thanks for adding it! I'm still learning a lot about rails generators 😬 

Do you mind adding a test for this behavior? That way, we can make sure it works on CI, and we can maintain compatibility for it in the future. You can see the one existing test here: 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/generators/graphql/function_generator_spec.rb

I think you can add another test case, but change the input to `run_generator`, something like: 

```ruby
test "it ..." do 
  run_generator(["finders::find_record"])
  # then assert file contents 
end 
```

Do you mind giving it a try?

Illianthe:
Done!

rmosolgo:
That's awesome, thanks again 😄

PR #1129, opened 2017-11-24
Add Datadog Tracing Support (v2: without variable scrubbing)

esjee:
Alternative to #1060. As I suggested in that PR, let's remove the data scrubbing because it feels like we're stalling the PR on something that is not its core functionality.


rmosolgo:
Thanks for following up on this! 

Thanks also to @michalg- for starting this effort!

PR #1134, opened 2017-11-27
Document from definition

xuorig:
This PR introduces `DocumentFromSchemaDefinition` which produces an AST given a `GraphQL::Schema` object.

This is another approach than https://github.com/rmosolgo/graphql-ruby/pull/1114.

Few questions:

  - Should it be `NodeFromSchemaMember` instead? I like restricting to schema only because we can have options such as passing a mask, `include_schema_type`, `include_built_ins`, etc.

In a next PR, I'd like `Schema::Printer` to use this API and use `Language::Printer` (https://github.com/rmosolgo/graphql-ruby/pull/1118) to generate an IDL.

This will require adding all the options I mentioned above to `DocumentFromSchemaDefinition`. For now it will print the complete schema (including scalars, directives and introspection types).

@rmosolgo what do you think of this approach? Should it accept any schema member instead?

### DIFF
--- a/lib/graphql/schema/traversal.rb
+++ b/lib/graphql/schema/traversal.rb
@@ -89,7 +89,6 @@ module GraphQL
             end
           elsif !prev_type.equal?(type_defn)
             # If the previous entry in the map isn't the same object we just found, raise.
-            binding.pry
             raise("Duplicate type definition found for name '#{type_defn.name}'")
           end
         else
--- a/lib/graphql/schema/traversal.rb
+++ b/lib/graphql/schema/traversal.rb
@@ -89,6 +89,7 @@ module GraphQL
             end
           elsif !prev_type.equal?(type_defn)
             # If the previous entry in the map isn't the same object we just found, raise.
+            binding.pry
             raise("Duplicate type definition found for name '#{type_defn.name}'")
           end
         else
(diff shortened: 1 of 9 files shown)


spec/graphql/language/document_from_schema_definition_spec.rb
```diff
@@ -0,0 +1,337 @@
...
+
+      expected_document = GraphQL.parse(expected_complete_idl)
+
+      assert equivalent_node?(expected_document, document)
+    end
+  end
+
+  private
+
+  def equivalent_node?(expected, node)
```
rmosolgo: Just to help me understand, how is this different from [AbstractNode#eql?](https://github.com/rmosolgo/graphql-ruby/blob/7109e74dca7893016e326624dc1e1bef51b7d709/lib/graphql/language/nodes.rb#L35) ? 

Is it just that this one returns `true` even when children & scalars are in different _order_?

lib/graphql/schema/traversal.rb
```diff
@@ -89,6 +89,7 @@ def visit(member, context_description)
             end
           elsif !prev_type.equal?(type_defn)
             # If the previous entry in the map isn't the same object we just found, raise.
+            binding.pry
```
rmosolgo: oops!

spec/graphql/language/document_from_schema_definition_spec.rb
xuorig: Yep.. thats all 😓  but as discussed on slack,

xuorig:
@rmosolgo I've addressed your comments! Let me know if you see anything missing here :)

spec/graphql/language/document_from_schema_definition_spec.rb
rmosolgo: 👌 thanks!

PR #1135, opened 2017-11-27
Fix default value input object b from def

xuorig:
`build_from_definition` would 💥 when passed a schema with input object default values. See test for example :)


rmosolgo:
Nice find! TIL String#chop.

PR #1140, opened 2017-11-28
Replace instances of 'assert_equal nil, x' with 'assert_nil x'

chrisbutcher:
Resolves warnings when running tests of the form `Use assert_nil if expecting nil from <file name, context>. This will fail in MT6.`

@rmosolgo


rmosolgo:
Awesome, thanks for taking care of this!

PR #1141, opened 2017-11-28
Add ArgumentValue#default_used? and GraphQL::Query::Arguments#default_used?(key)

chrisbutcher:
As discussed yesterday in the GraphQL #ruby Slack channel.

I've been looking for a way, via instrumentation http://graphql-ruby.org/queries/instrumentation.html in an `after_query` method, to determine whether an `GraphQL::Argument` defined on a `GraphQL::InputObjectType` _that has a default_ actually used the default during resolution.

Argument values can be of type `GraphQL::Language::Nodes::VariableIdentifier` when the input object value is defined in a variable, or of the type `GraphQL::Language::Nodes::InputObject` when its defined inline. I could traverse through either to determine if a given argument fell back to its default that way.

But it seems like instrumentation code that seeks to know if argument defaults were used would be greatly simplified if we simply added this detail to ArgumentValue.

@rmosolgo I took your suggestion:

> makes sense to me, maybe `arguments_class.new(...)` could take a new parameter of names where the default was applied?

... and came up with a working solution, but would like some input on my approach and naming of things.

## tldr
With this change, arguments specified either inline or in variables that end up falling back to their definition's default are marked as such through the new methods `GraphQL::Query::ArgumentValue#default_used?` and `GraphQL::Query::Arguments#default_used?(key)`.

Also note that the assert_nil warning surfaced when running ArgumentsSpec is fixed in https://github.com/rmosolgo/graphql-ruby/pull/1140

cc: @eapache @swalkinshaw

### DIFF
--- a/lib/graphql/input_object_type.rb
+++ b/lib/graphql/input_object_type.rb
@@ -84,6 +84,7 @@ module GraphQL
 
     def coerce_non_null_input(value, ctx)
       input_values = {}
+      arg_default_used = {}
 
       arguments.each do |input_key, input_field_defn|
         field_value = value[input_key]
@@ -93,10 +94,11 @@ module GraphQL
           input_values[input_key] = input_field_defn.prepare(coerced_value, ctx)
         elsif input_field_defn.default_value?
           input_values[input_key] = input_field_defn.default_value
+          arg_default_used[input_key] = true
         end
       end
 
-      arguments_class.new(input_values)
+      arguments_class.new(input_values, arg_default_used)
     end
 
     # @api private
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -12,8 +12,10 @@ module GraphQL
         argument_owner.arguments_class = Class.new(self) do
           self.argument_definitions = argument_definitions
 
-          def initialize(values)
-            super(values, argument_definitions: self.class.argument_definitions)
+          def initialize(values, values_used_default = Hash.new(false))
+            super(values, argument_definitions: self.class.argument_definitions).tap do |arguments|
+              set_default_used(arguments, values_used_default)
+            end
           end
 
           argument_definitions.each do |_arg_name, arg_definition|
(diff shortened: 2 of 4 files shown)


lib/graphql/query/arguments.rb
```diff
@@ -12,8 +12,10 @@ def self.construct_arguments_class(argument_owner)
         argument_owner.arguments_class = Class.new(self) do
           self.argument_definitions = argument_definitions
 
-          def initialize(values)
-            super(values, argument_definitions: self.class.argument_definitions)
+          def initialize(values, values_used_default = Hash.new(false))
```
chrisbutcher: The main reason for this default value is this call: https://github.com/chrisbutcher/graphql-ruby/blob/b9a6401218b76b6dd0c96105816050591dd3928b/lib/graphql/query/arguments.rb#L142

I'm not 100% sure if this is the correct behaviour here.

My initial approach was to pass through the `values_used_default` hash to `super`,  and initialize ArgumentValues with a `default_used` value directly, similar to how `arg_defn` is passed along. Maybe you'd prefer this?

eapache: I'm honestly not sure what the default value has anything to do with the passing of `values_used_default` to super?

I believe I would prefer that, but I don't see why you'd change it in the first place so I think I'm missing something.

chrisbutcher: I've pushed up the alternate approach here: https://github.com/chrisbutcher/graphql-ruby/pull/1 

Let me know if my description still doesn't clarify things.

lib/graphql/input_object_type.rb
```diff
@@ -84,6 +84,7 @@ def coerce_result(value, ctx = nil)
 
     def coerce_non_null_input(value, ctx)
       input_values = {}
+      arg_default_used = {}
```
chrisbutcher: Woops, I'll need to update the naming of things between `used_default` and `default_used` for consistency.

lib/graphql/query/arguments.rb
eapache: Personally I prefer that branch

chrisbutcher: Me too, closing this one 🔥

PR #1142, opened 2017-11-28
Refactor JS client to allow new types (json), add outfileType option

kevinjalbert:
A new outfileType option is added to the graphql-ruby-client that allows
you to specify the generated file type (i.e., for apollo-js or json). It
would be possible to add new types and their generators in the future.

The outfileType and outFile itself are independent. For example, you
could have a json file output in a .js file. It is left up to the user
to make sure they match up.

The new json outfileType is useful for non JS clients (i.e., android and
iOS).

Added tests around the new changes.

Refactored the keyValuePairs generation in /sync/generateClient.js to
be valid json (i.e., no trailing ',').


kevinjalbert:
@rmosolgo in its current form the generated JS client is _react-apollo_ JS specific (https://github.com/rmosolgo/graphql-ruby/blob/31ff644e2153edabfcc131ccb84c6e5b61015ccf/javascript_client/sync/generateClient.js#L66-L92).

Apollo can follow the relay convention with edges/nodes though if the GraphQL server supports relay conventions. Is that possibly what you are thinking of?

rmosolgo:
The `getOperationId` function in the generated code is my attempt to support Relay: 

http://graphql-ruby.org/javascript_client/sync.html#use-with-relay

So from my perspective, the generated code supports _both_ Relay and Apollo, it just supports Apollo _better_! 

You can even use the `getOperationId` in plain JS: http://graphql-ruby.org/javascript_client/sync.html#use-with-plain-javascript

kevinjalbert:
@rmosolgo Ah totally missed the fact that you could simply use the `getOperationId` in those contexts. Serves me right -- I've only been working at this from an Apollo perspective.

One option I'm thinking of is to break it out like:

1. `js` -- services both `relay` and plain JS, as it'll just expose the `getOperationId` function.
2. `apollo-js` -- services just `react-apollo`, as it'll just expose the middleware.

My main reason for this is that the `js` generated code wouldn't have any dead code if you are using `relay` or plain JS.

Or... we just rename it to `js` and we call it a day ;)

rmosolgo:
Let's just rename it for now, if you (or someone else) want to split it up later, we can investigate that in a separate PR!

kevinjalbert:
@rmosolgo renamed in the latest commit (4328dda).

rmosolgo:
:bow: thanks for accommodating me on that! 😊

rmosolgo:
Just released in 1.3.0!

PR #1143, opened 2017-11-28
[1.8] Handle underscored argument names

swalkinshaw:
This is a follow-up to https://github.com/rmosolgo/graphql-ruby/pull/1126.

Completely forgot about arguments 😰 . The actual implementation of this was a little more involved than fields. I think it's correct?

Also updated the new guide to use this convention.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -48,7 +48,7 @@ module GraphQL
           GraphQL::Field.new
         end
 
-        field_defn.name = camelize(name)
+        field_defn.name = Member::BuildType.camelize(name)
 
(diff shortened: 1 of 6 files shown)


lib/graphql/schema/field.rb
```diff
@@ -48,7 +48,7 @@ def to_graphql
           GraphQL::Field.new
         end
 
-        field_defn.name = camelize(name)
```
rmosolgo: It looks like the implementation was moved from this class into `BuildType`, can we remove the code from this file?

swalkinshaw: oops removed!

PR #1144, opened 2017-11-28
[RFC] Language Server 

rmosolgo:
I've been home with a new baby, so to stay sane, I've been fooling around with something new to me, the [Language Server Protocol](https://github.com/Microsoft/language-server-protocol) (LSP). 

What do you think, would it be a good addition to this project?

## TL;DR 

It can provide a lot of features to editors: 

- Code completion 
- Error highlighting & suggestions
- Auto-formatting
- Jump-to-definition

See [atom/atom-languageclient](https://github.com/atom/atom-languageclient) for an example editor client. 

Early screenshots:

image

image

image

## Goals

In this PR I hope to: 

- [ ] Determine whether fellow graphql-rubyists see value in a hand-rolled language server 
- Build some basic features: 
  - [ ] autocomplete (completion and signatureHelp)
  - [ ] error checking
  - [ ] formatting
- Support GraphQL in several contexts: 
  - [ ] `.graphql` files
  - [ ] `GRAPHQL` Ruby heredocs
  - [ ] `<%graphql` ERB tags
  - [ ] `gql`, `graphql`, and `Relay.QL` JS template strings
- [ ] Unit-test core components (eg, `CompletionProvider`)
- [ ] Integration-test the language server 
- [ ] Build support for one editor (Atom) 
- [ ] Build a low-barrier integration with Rails + a generator 
- [ ] Document usage with Rails and without Rails 

Eventually, I'd like to: 

- Support more LSP features: 
  - Jump-to-definition, using the `self.extended` hook, plus class-based type definitions 
  - Error suggestions ("code actions" in LSP-speak) 
  - Hover data (show documentation) 
- Support more editors as requested & contributed (I'm an Atom user)
- Improve performance as needed 

## Non-Goals 

- Implement every LSP method
- Support all text editors immediately
- Focus on performance (not _yet_)

## Proposed Architecture 

- User API: 
  - Users create a _server script_, a stand-alone `.rb` file which is executed by the editor when the editor wants a language server. 
  - This stand-alone script is bootstrapped by `rails g graphql:language_server`
- Internals: 
  - Core components are LSP-agnostic, rigorously unit-tested
    - For example, `CompletionProvider` takes input of text, cursor position, and project data, and returns a list of suggestions. 
    - `CompletionProvider` can be unit-tested without any dependency on LSP 
    - LSP-specific code converts the list of suggestion to a `textDocument/completion` response 
  - The Language Server itself is relatively small, the response classes are like "skinny controllers" that delegate to business objects. 
    - Minimally test the language server by some kind of integration test which _really_ starts up a server and communicates with it via IO, see the [language_server-protocol tests](https://github.com/mtsmfm/language_server-protocol-ruby/blob/91214ec628d02daa52436ee4bd0057ff68edeb78/test/language_server/protocol_test.rb#L10) for inspiration

The initial import contains some of this structure ... but none of the tests 😊 

## Why not JavaScript? 

There's a JS project called [GraphQL Language Service](https://github.com/graphql/graphql-language-service), but I didn't try to use it because: 

- I'm not familiar with it's dependencies 
- The support for language server methods was not great anyways 
- I don't like JavaScript

Besides that, I had doubts about doing Ruby-specific features like `<%graphql` and `<<-GRAPHQL`. 

-------

Here's an example server, based on this initial import (I'm using it in a toy app) 

```ruby 
#!/usr/bin/env ruby

# app/graphql/language_server.rb
# Load your application: 
require_relative "../../config/environment"
# Load the language server class:
require "graphql/language_server"

# Extend the base language server class:
class GraphQLLanguageServer < GraphQL::LanguageServer
  # Hook up your schema 
  self.schema = MySchema 
  # Add a logger
  self.logger = Logger.new("#{Rails.root}/log/lsp.log")
end
```


gjtorikian:
I think this is pretty great!

mtsmfm:
> Here's an example server, based on this initial import (I'm using it in a toy app)

What do you think about loading config/environment first?

```ruby
#!/usr/bin/env ruby

require_relative "../config/environment"
require "graphql/language_server"

class GraphQLLanguageServer < GraphQL::LanguageServer
  self.schema = StarWarsSchema
  self.logger = Logger.new(Rails.root.join("log/lsp.log"))
  self.development_mode = true
end

GraphQLLanguageServer.new.start
```

https://github.com/mtsmfm/graphql-ruby-demo/blob/5169f4a29a9110519e1ca73ff86b8f3358148e3d/bin/graphql_language_server

This server requires our own schema.
So I think it should be loaded as same as how Rails loads.

rmosolgo:
> config/environment first? 

👌  that's great. I also hope I can implement an "autorun" feature like `minitest/autorun`.

arturtr:
What about simlify this issue to add some tools for making graphql config?
https://github.com/prismagraphql/graphql-config

This config supported by most of editiors and IDE

rmosolgo:
Thanks for the link, I'll read up on that!

arturtr:
Do you have any news in this way? I'm about `graphql-config` usage with your gem

rmosolgo:
Maybe someday 🍻

PR #737, opened 2017-05-26
Improve lexing performance by generating lexer with Ragel 7.0.0.9 and avoiding lambda in token emission

ivoanjo:
While investigating why query lexing was measurably slower in JRuby than with MRI (see https://github.com/jruby/jruby/issues/4623 ) I noticed that GraphQL's lexer seemed to be generated with Ragel 6.

I then noticed that this version was rather old and that a newer Ragel 7 seemed to have a lot of improvements on the Ruby output backend, even adding four different output options for generated lexers:

* `-T0` Binary search (default)
* `-T1` Binary search with expanded actions
* `-F0` Flat table
* `-F1` Flat table with expanded actions

Testing all four of these combinations and comparing with the current GraphQL lexer on both MRI and JRuby showed that for a lexing-only benchmark (included below), Ragel 7's `-F1` improved MRI peformance by 87% and JRuby performance by 263%:

* MRI before:

```
Running with ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
Warming up --------------------------------------
               Lexer   111.000  i/100ms
Calculating -------------------------------------
               Lexer      1.124k (± 3.9%) i/s -     33.744k in
30.069500s
```

* MRI after:

```
Running with ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
Warming up --------------------------------------
               Lexer   209.000  i/100ms
Calculating -------------------------------------
               Lexer      2.102k (± 3.0%) i/s -     63.118k in
30.052143s
```

* JRuby before:

```
Running with jruby 9.1.9.0 (2.3.3) 2017-05-15 28aa830 Java HotSpot(TM)
64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [linux-x86_64]
Warming up --------------------------------------
               Lexer    24.000  i/100ms
Calculating -------------------------------------
               Lexer    250.435  (± 4.8%) i/s -      7.512k in
30.080584s
```

* JRuby after:

```
Running with jruby 9.1.9.0 (2.3.3) 2017-05-15 28aa830 Java HotSpot(TM)
64-Bit Server VM 25.131-b11 on 1.8.0_131-b11 +jit [linux-x86_64]
Warming up --------------------------------------
               Lexer    88.000  i/100ms
Calculating -------------------------------------
               Lexer    910.242  (± 3.0%) i/s -     27.280k in
29.999337s
```

[...]

PR #1145, opened 2017-11-28
Support blank curlies

gjtorikian:
`print_schema` will happily print empty curlies for types without fields `type Foo{}`, but the parser will not accept this as a valid IDL file:

This PR changes parser.y to accept empty field lists. This is also in compliance with the IDL spec at https://github.com/facebook/graphql/pull/90/files#r132484441.

xref https://github.com/gjtorikian/graphql-docs/issues/32
cc @Mori


rmosolgo:
This looks great, thanks for taking care of it! 

But, I see that some changes were committed to `lexer.rb`, was this intentional? That file is generated by `ragel`, based on `lexer.rl`, so I don't expect any changes to it, except when the `.rl` file also changes. Do you mind reverting the changes to `lexer.rb`?

gjtorikian:
Hmm, I sample ran the rake task but sure I’ll try reverting that change!

gjtorikian:
`rake build_parser` insists on including lexer.rb:

```
graphql-ruby (support-blank-curlies): git status          
On branch support-blank-curlies
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   lib/graphql/language/parser.rb
	modified:   lib/graphql/language/parser.y

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   lib/graphql/language/lexer.rb

graphql-ruby (support-blank-curlies): git checkout lib/   
graphql-ruby (support-blank-curlies): git status       
On branch support-blank-curlies
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   lib/graphql/language/parser.rb
	modified:   lib/graphql/language/parser.y

graphql-ruby (support-blank-curlies): be rake build_parser
135 shift/reduce conflicts
45 reduce/reduce conflicts
graphql-ruby (support-blank-curlies): git status          
On branch support-blank-curlies
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   lib/graphql/language/parser.rb
	modified:   lib/graphql/language/parser.y

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   lib/graphql/language/lexer.rb

```

Have I done something wrong?

rmosolgo:
Very weird, the commands all look right! I assumed it was some kind of format-on-save plugin, but I guess not 😖 

Anyhow, I pulled the branch and re-ran `build_parser`. It cleared up the changes to `lexer.rb`, but I was surprised to also see [changes to `parser.rb`](https://github.com/rmosolgo/graphql-ruby/pull/1145/commits/619dabeb785fb20a089bee5b6ba0e3e5967b863a#diff-85a436d444a3faed35353741f5108507). Then, I realized that I don't actually _check_ the Racc & Ragel versions on people's dev environment when they build those files, so it could be that different versions output slightly different code. So, I added those checks to make sure we're all running the same versions when building those files. Maybe that was it, hopefully that will catch it in the future!

gjtorikian:
Woop woop, thank you!

rmosolgo:
:metal: !

Rakefile
```diff
@@ -40,6 +40,26 @@ end
...
+
+To fix this issue:
+
+- Update #{dep_name} to the required version
+- Update the assertion in `Rakefile` to match the current version
+ERR
+    end
+  end
+
+  assert_dependency_version("Ragel", "7.0.0.9", "ragel -v")
```
zzak: @rmosolgo @gjtorikian Is there a reason to pin to this particular version of ragel?

I'm only able to find 6.9 in my [apt repos](https://packages.ubuntu.com/artful/ragel).

rmosolgo: It was found to be significantly faster back in https://github.com/rmosolgo/graphql-ruby/pull/737, but I didn't add the lock until here. Yeah, sorry, you have to get the "Development" build from http://www.colm.net/open-source/ragel/!

PR #1148, opened 2017-11-29
Failing test: Query without selections raise validation error

chrisbutcher:
This PR adds a failing test to demonstrate that, at present, a GraphQL Ruby app will fail to report any errors if the query document contains only comments or no contents at all.

Expected: The result should populate `errors` to indicate to the client that the request was invalid
Actual: The result contains only `{}`

@rmosolgo

### DIFF
--- a/spec/graphql/static_validation/validator_spec.rb
+++ b/spec/graphql/static_validation/validator_spec.rb
@@ -137,5 +137,14 @@ describe GraphQL::StaticValidation::Validator do
         assert_equal(1, errors.length)
       end
     end
+
+    describe "queries with no selections" do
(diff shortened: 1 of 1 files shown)


rmosolgo:
It sounds like nobody objected way back in https://github.com/rmosolgo/graphql-ruby/issues/1149, do you want to make this test pass? I think it's :+1: to make this change if so.

chrisbutcher:
Sure thing, @rmosolgo! I lost track of this PR :)

rmosolgo:
Feel free to reopen this or make a new PR if you get a chance to update it!

PR #1152, opened 2017-11-30
Add ArgumentValue#default_used? and Arguments#default_used?

chrisbutcher:
As discussed the other day in the GraphQL #ruby Slack channel, I've been looking for a way, via instrumentation http://graphql-ruby.org/queries/instrumentation.html in an `after_query` method, to determine whether an `GraphQL::Argument` defined on a `GraphQL::InputObjectType` _that has a default_ actually used the default during resolution.

Argument values can be of type `GraphQL::Language::Nodes::VariableIdentifier` when the input object value is defined in a variable, or of the type `GraphQL::Language::Nodes::InputObject` when its defined inline. I could traverse through either to determine if a given argument fell back to its default that way.

But it seems like instrumentation code that seeks to know if argument defaults were used would be greatly simplified if we simply added this detail to ArgumentValue.

@rmosolgo I took your suggestion:

> makes sense to me, maybe `arguments_class.new(...)` could take a new parameter of names where the default was applied?

... and came up with a working solution, but would like some input on my approach and naming of things.

## tldr
With this change, arguments specified either inline or in variables that end up falling back to their definition's default are marked as such through the new methods `GraphQL::Query::ArgumentValue#default_used?` and `GraphQL::Query::Arguments#default_used?(key)`.

Also note that the `assert_nil` warning surfaced when running ArgumentsSpec is fixed in https://github.com/rmosolgo/graphql-ruby/pull/1140

Closed https://github.com/rmosolgo/graphql-ruby/pull/1141 in favour of this PR

cc: @eapache @swalkinshaw

### DIFF
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -41,7 +41,7 @@ module GraphQL
           arg_name = inner_key.to_s
 
           arg_defn = argument_definitions[arg_name]
-          arg_default_used = defaults_used[arg_name]
+          arg_default_used = defaults_used.include?(arg_name)
 
           arg_value = wrap_value(inner_value, arg_defn.type)
           string_key = arg_defn.expose_as
@@ -133,7 +133,7 @@ module GraphQL
             wrap_value(value, arg_defn_type.of_type)
           when GraphQL::InputObjectType
             if value.is_a?(Hash)
-              arg_defn_type.arguments_class.new(value, Hash.new(false))
+              arg_defn_type.arguments_class.new(value, Set.new)
             else
               value
             end
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -12,8 +12,8 @@ module GraphQL
         argument_owner.arguments_class = Class.new(self) do
           self.argument_definitions = argument_definitions
 
-          def initialize(values)
-            super(values, argument_definitions: self.class.argument_definitions)
+          def initialize(values, defaults_used)
+            super(values, argument_definitions: self.class.argument_definitions, defaults_used: defaults_used)
           end
 
           argument_definitions.each do |_arg_name, arg_definition|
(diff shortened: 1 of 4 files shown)


lib/graphql/query/arguments.rb
```diff
@@ -115,7 +133,7 @@ def wrap_value(value, arg_defn_type)
             wrap_value(value, arg_defn_type.of_type)
           when GraphQL::InputObjectType
             if value.is_a?(Hash)
-              arg_defn_type.arguments_class.new(value)
+              arg_defn_type.arguments_class.new(value, Hash.new(false))
```
chrisbutcher: I think this is a sensible value to pass back to the initializer if an argument default happens to be a Hash.

rmosolgo: Yeah, it makes sense to me, too. But it seems like this code is a bug waiting to happen, huh? I mean, since default values are applied elsewhere, I wonder if there's some case where a default value isn't properly applied. :S 

Anyhow, that doesn't have to do with these changes!

chrisbutcher:
Thanks @rmosolgo!

> do you think we could express that idea with an Array or Set? Personally, I find it a bit clearer, since we're not actually storing anything in the hash's values. What do you think about that change?

💯 in fact my first approach was to use a `Set` for this very reason. I'll switch back to that implementation.

lib/graphql/query/arguments.rb
```diff
@@ -85,7 +95,7 @@ def each_value
         end
       end
 
-      NO_ARGS = self.new({}, argument_definitions: [])
+      NO_ARGS = self.new({}, argument_definitions: [], defaults_used: {})
```
rmosolgo: Should be `Set.new`? I guess it's splitting hairs since `#include?` will still return false..

spec/graphql/query/arguments_spec.rb
```diff
@@ -306,7 +328,7 @@
       assert_equal nil, input_object.arguments_class
 
       GraphQL::Query::Arguments.construct_arguments_class(input_object)
-      args = input_object.arguments_class.new({foo: 3, bar: -90})
+      args = input_object.arguments_class.new({foo: 3, bar: -90}, {})
```
rmosolgo: Also `Set.new`?

chrisbutcher:
(Forgot to ping again after new commit) Updated!

rmosolgo:
Thanks for making this happen! 🎉

chrisbutcher:
Thank _you_, Robert!

PR #1154, opened 2017-11-30
Add introspection attribute to fields

swalkinshaw:
Types already had the `introspection` attribute, but without it on fields, there was no easy to check if a field is an introspection field or not.

Usually an introspection field returns an introspection type, but this breaks down for `__typename` which always returns a `String`. Instead of treating it differently, this just adds the attribute to all fields.

Now `introspection?` is available. This can be useful for analysis like computing complexity.

### DIFF
--- a/lib/graphql/field.rb
+++ b/lib/graphql/field.rb
@@ -131,6 +131,7 @@ module GraphQL
       :relay_nodes_field,
       :subscription_scope,
       :trace,
+      :introspection,
       argument: GraphQL::Define::AssignArgument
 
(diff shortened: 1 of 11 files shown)


swalkinshaw:
I think the alternative here is just a static check for `name == '__typename'` so this isn't the biggest deal. But it makes semantic sense since they *are* introspection fields?

chrisbutcher:
Seems like a reasonable change 👍

rmosolgo:
:ok_hand: nice!

PR #1151, opened 2017-11-30
Fixes GraphQL Upgrader: multi-line fields/connections, and helpful message for :create_base_objects task

esjee:
After trying out the initial version of the upgrader in https://github.com/rmosolgo/graphql-ruby/pull/1110 on the H1 codebase, I ran into some surprises. This PR is supposed to highlight those, and hopefully @Willianvdv can use his regex magic to fix some of these :innocent:

PR #1156, opened 2017-12-01
WIP: Fixes GraphQL Upgrader: description inside block 

esjee:
Depends on https://github.com/rmosolgo/graphql-ruby/pull/1151! Let's merge that one first.

Add the scenario for having a description inside the block when defining a field.


spec/graphql/upgrader/member_spec.rb
```diff
@@ -213,10 +213,54 @@ class UserInterface < BaseInterface
...
+  end
+
+  # describe 'field with description inside block' do
+  #   it 'moves the description as an argument' do
+  #     old = %{
+  #       field :_id, !types.ID, property: :id do
+  #         description 'The primary key from the database'
+  #       end
+  #     }
+  #     new = %{
```
esjee: I'm unsure if this "new" is as expected!

Willianvdv:
@esjee should we wait for the outcome of the discussion on #1158 before fixing this?

rmosolgo:
Is this change still required? If so, it'll probably need some rework since I change the Upgrader quite a bit 😬

rmosolgo:
Feel free to open a new PR against 1.8-dev if there's still something missing there!

PR #1159, opened 2017-12-01
Refactor Schema::Printer to use Language::Printer

xuorig:
This PR does a few things:

  - Adds a lot of options to `DocumentFromSchemaDefinition`:
    - context: nil
    - only: nil
     - except: nil
     - include_introspection_types: false
     - include_built_in_directives: false
     - include_built_in_scalars: false, 
     - always_include_schema: false

This allows printers to customize the AST before printing it, and removes this logic from printers which become "dumb".

  - Refactors `SchemaPrinter` to use `DocumentFromSchemaDefinition`, this 🔥 a lot of code.
    - SchemaPrinter now inherits from `Language::Printer`.
    - Makes it much easier to extend now

   - Fixes a few issues with `Language::Printer` for printing schemas.

I tried to keep the Schema::Printer API exactly the same, only the underlying logic is changing.

### DIFF
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -38,8 +38,6 @@ module GraphQL
         )
       end
 
-      protected
-
       def build_schema_node
         schema_node = GraphQL::Language::Nodes::SchemaDefinition.new
 
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -17,18 +17,13 @@ module GraphQL
       )
         @schema = schema
         @always_include_schema = always_include_schema
-
-        filter = GraphQL::Language::DocumentFromSchemaDefinition::Filter.new(
-          only,
-          except,
-          include_introspection_types: include_introspection_types,
-          include_built_ins: include_built_ins,
-        )
+        @include_introspection_types = include_introspection_types
+        @include_built_ins = include_built_ins
 
         @warden = GraphQL::Schema::Warden.new(
-          filter,
+          GraphQL::Filter.new(only: only, except: except),
           schema: @schema,
-          context: @context
+          context: context,
         )
       end
 
@@ -59,7 +54,7 @@ module GraphQL
       def build_object_type_node(object_type)
         GraphQL::Language::Nodes::ObjectTypeDefinition.new(
           name: object_type.name,
-          interfaces: warden.interfaces(object_type).map { |iface| build_type_name_node(iface) },
+          interfaces: warden.interfaces(object_type).sort_by(&:name).map { |iface| build_type_name_node(iface) },
           fields: build_field_nodes(warden.fields(object_type)),
           description: object_type.description,
         )
@@ -78,7 +73,7 @@ module GraphQL
         GraphQL::Language::Nodes::UnionTypeDefinition.new(
           name: union_type.name,
           description: union_type.description,
-          types: warden.possible_types(union_type).map { |type| build_type_name_node(type) }
+          types: warden.possible_types(union_type).sort_by(&:name).map { |type| build_type_name_node(type) }
         )
       end
 
@@ -93,7 +88,7 @@ module GraphQL
       def build_enum_type_node(enum_type)
         GraphQL::Language::Nodes::EnumTypeDefinition.new(
           name: enum_type.name,
-          values: warden.enum_values(enum_type).map do |enum_value|
+          values: warden.enum_values(enum_type).sort_by(&:name).map do |enum_value|
             build_enum_value_node(enum_value)
           end,
           description: enum_type.description,
@@ -175,11 +170,19 @@ module GraphQL
       end
 
       def build_argument_nodes(arguments)
-        arguments.map { |arg| build_argument_node(arg) }
+        arguments
+          .map { |arg| build_argument_node(arg) }
+          .sort_by(&:name)
       end
 
       def build_directive_nodes(directives)
-        directives.map { |directive| build_directive_node(directive) }
+        if !include_built_ins
+          directives = directives.reject { |directive| built_in?(directive) }
+        end
+
+        directives
+          .map { |directive| build_directive_node(directive) }
+          .sort_by(&:name)
       end
 
       def build_definition_nodes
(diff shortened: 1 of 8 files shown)


xuorig:
cc @rmosolgo this is ready for review!

(not sure what is going on with codeclimate)

rmosolgo:
Cool! I'll take a close look soon.

Code Climate was _kinda_ broken before, now it looks _very_ broken 😬

lib/graphql/language/document_from_schema_definition.rb
```diff
@@ -18,18 +36,18 @@ def document
...
+          schema_node.query = schema.query.name
+        end
 
-        if schema.mutation
+        if schema.mutation && warden.get_type(schema.mutation.name)
           schema_node.mutation = schema.mutation.name
         end
 
-        if schema.subscription
+        if schema.subscription && warden.get_type(schema.subscription.name)
```
rmosolgo: How about using [`#root_type_for_operation("subscription")`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/warden.rb#L118) here and for `mutation` and `query` ?

lib/graphql/language/document_from_schema_definition.rb
```diff
@@ -6,10 +6,28 @@ module Language
     # {GraphQL::Language::DocumentFromSchemaDefinition} is used to convert a {GraphQL::Schema} object
     # To a {GraphQL::Language::Document} AST node.
     #
+    # @param context [Hash]
+    # @param only [<#call(member, ctx)>]
+    # @param except [<#call(member, ctx)>]
+    # @param include_introspection_types [Boolean] Wether or not to include introspection types in the AST
+    # @param include_built_in_scalars [Boolean] Wether or not to include built in scalars in the AST
+    # @param include_built_in_directives [Boolean] Wether or not to include built in diirectives in the AST
```
rmosolgo: W**h**ether 😊

lib/graphql/schema.rb
```diff
@@ -564,6 +564,12 @@ def to_json(*args)
       JSON.pretty_generate(as_json(*args))
     end
 
+    def respects_root_name_conventions?
```
rmosolgo: Is this concern only for printing the schema? If so, can we keep it out of `schema.rb`? There's already a lot going on in here, and if we can get away with putting it somewhere else, I'd like to do that!

spec/graphql/schema/printer_spec.rb
```diff
@@ -390,7 +389,7 @@
 
 enum Choice {
   BAR
-  BAZ @deprecated(reason: "Use \\\"BAR\\\".")
```
rmosolgo: 🤔  Can you help me understand why these escapes are not needed anymore? (Were they wrong to begin with?)

lib/graphql/language/document_from_schema_definition.rb
xuorig: What's the difference here? It doesnt look like `root_type_for_operation` uses the warden 🤔

spec/graphql/schema/printer_spec.rb
xuorig: Apparently it has to do with the call to inspect here: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/printer.rb#L164

```
irb(main):031:0> field_or_enum_value.deprecation_reason
=> "use \"this\""
irb(main):032:0> field_or_enum_value.deprecation_reason.to_s
=> "use \"this\""
irb(main):033:0> field_or_enum_value.deprecation_reason.to_s.inspect
=> "\"use \\\"this\\\"\""
```

I don't think its needed 🤔

lib/graphql/schema.rb
xuorig: done!

xuorig:
Addressed your comments, few other things I worry about:

  - Were these printer modules used by users? 😨 (might be an acceptable breaking change..)

  - The schema printer now only has access to AST nodes and not schema members definitions. This means printing according to something in metadata for example is now much more complex. I suspect there's not that many valid use cases here. Information would usually find its way in AST, especially as we start maybe accepting custom directives. Worst case we can open up `DocumentFromSchemaDefinition` so users can have custom ASTs from schema definitions.

In the end this last point might be a good thing since you can only print valid schema ASTs, making your schema readable by any graphql implementation.

lib/graphql/language/document_from_schema_definition.rb
rmosolgo: Oh, to specify, `Warden#root_type_for_operation`, not `Schema#root_type_for_operation`: 

https://github.com/rmosolgo/graphql-ruby/blob/20d70eb374809835fedaf85bd8bffd0d71799abe/lib/graphql/schema/warden.rb#L118-L125

xuorig: ohhhhhh 💡 awesome.

Done :)

lib/graphql/language/document_from_schema_definition.rb
```diff
@@ -18,73 +36,81 @@ def document
...
+          arguments: build_argument_nodes(warden.arguments(field)),
           type: build_type_name_node(field.type),
           description: field.description,
         )
+
+        if field.deprecation_reason
+          field_node.directives << GraphQL::Language::Nodes::Directive.new(
+            name: GraphQL::Directive::DeprecatedDirective.name,
+            arguments: [GraphQL::Language::Nodes::Argument.new(name: "reason", value: field.deprecation_reason)]
+          )
```
rmosolgo: 👌

lib/graphql/schema/printer.rb
```diff
@@ -74,277 +85,37 @@ def self.print_schema(schema, **args)
...
+          print_deprecated(directive)
         else
-          true
+          super
         end
-      }
+      end
 
-      private_constant :IS_USER_DEFINED_MEMBER
+      def print_deprecated(directive)
```
rmosolgo: Can this method be private? Or, better yet, can this code be inlined into `print_directive`? (I couldn't see any usages of it, but maybe I overlooked them!)

spec/graphql/schema/build_from_definition_spec.rb
```diff
@@ -520,7 +520,7 @@ def build_schema_and_compare_output(definition)
 }
 
 type Mutation {
-  hello(str: String, int: Int, color: Color = RED, nullDefault: Int = null): String
+  hello(color: Color = RED, int: Int, nullDefault: Int = null, str: String): String
```
rmosolgo: yay, sorting

spec/graphql/schema/printer_spec.rb
rmosolgo: Cool, thanks for the details!

lib/graphql/schema/printer.rb
xuorig: good point, fixed!

rmosolgo:
🙌  🎉

xuorig:
🎉

PR #1161, opened 2017-12-02
[1.8] Support custom context classes

rmosolgo:
Custom context classes will allow you to put a bit of logic or caching in `@context`

This also points out a change in 1.8: 

Previously, `context` was field-specific, holding references to things like `ast_node`, `irep_node`, `parent`, `object`, `value`. But now, individual fields don't get called with their own contexts. Instead, the context is provided at object-level, then shared between fields on that object. 

So, we will need (?) some API for getting that info to fields who are using it.


rmosolgo:
Those objects are readable via `extras:` in https://github.com/rmosolgo/graphql-ruby/pull/1170/commits/4d71f8e076591ce943999f1e058a4913bb9d5782, if we need to write to them, I bet we can find a way (perhaps by accepting `extras: [:field_ctx]`).

PR #1164, opened 2017-12-02
[1.8] Implement around_resolve for wrapping fields on classes

rmosolgo:
Here's a take at a field-wrapping API for class-based types. It's ok, but after the first commit, there are some issues: 

- Let's say you define a generic wrapper. How can you figure out which field it's wrapping? 
- To test:
  - Test `around_resolve` with interfaces. How does that work? 
  - Unit-test `Arguments#to_kwargs`
- Build `before_resolve` and `after_resolve` on top of this? 

cc @swalkinshaw  what do you think of this direction?


swalkinshaw:
Is the idea to replace instrumentation with this new concept? (at least as a recommendation if you're using the class-based API)

With the old instrumentation you can easily instrument *every* field/type. Doesn't that become harder with this method?

You'd have to implement an `around_resolve` in at least 3 places I think? `BaseInterface`, `BaseUnion`, and `BaseObject`. I'd think it would be ideal if it could be done once in a `BaseMember` class but that's not possible right now due `around_resolve` being specific to types with `HasFields`.

I guess there's some performance implications too? With instrumenters they run once at schema definition time. `around_resolve` would happen on every resolve call. Not sure how much that matters in reality though.

rmosolgo:
Thanks for taking a look! 

> replace instrumentation?

It serves the same purpose, but I hope to make it easier for people to understand and implement (more Rails-y). It's hard to imagine _ever_ sunsetting instrumentation, since it's been in use for so long.


> every field/type

I think only `BaseObject` is required, since Interface and Union objects are never _actually_ initialized. 

One thing that's not well-supported right now is wrapping an interface field on all objects that implement that interface. You could do 

```ruby 
class MyInterface < GraphQL::Schema::Interface 
  module Implementation 
    def self.included(obj_class) 
      obj_class.around_resolve(:field_name, :method_name) 
    end 

    def method_name
      # do stuff 
    end 
  end 
end 
```

But maybe that could be made easier somehow...

Perhaps the same thing could be said about Unions. There's no way to say "wrap every field of each object that belongs to this Union." 

Do you know of use-cases like this yet?   

> performance implications

I have the same conclusion as you, I'm not sure how much it matters in practice. FWIW, after theorygeek refactored middleware to stop capturing the block, the performance overhead from there was really reduced. So my instinct was not to worry about this one!

My first goal here is to make it easy to add these kind of customizations without learning too many  GraphQL-Ruby-specific concepts, so I hope we can cross this bridge when we get there 😬

swalkinshaw:
> Do you know of use-cases like this yet?

I don't think so. If `BaseObject` is all that's required, then that becomes a much better alternative.

Thinking about this at a higher-level and how the new class-based API would replace various instrumenters:

* query instrumentation: custom Query classes
* field instrumentation: `around_resolve` in this PR
* type instrumentation: any normal Ruby way of extending classes 😛 

Does that look correct?

I guess one slight downside to the above is you lose the idea of "instrumentation" and the single entry point. You can get the same functionality, but it's done in different ways and maybe not as obvious.

That being said, I think `around_resolve` is probably an improvement 👍 . It would kind of suck to go from nice instance methods in a class to the normal field instrumentation pattern of nested procs.

rmosolgo:
> Does that look correct?

That's definitely the idea. I'm not exactly sure how query/multiplex instrumentation will play into the class-based API. 

One _nice_ thing about alternatives to "instrumentation" is that instrumentation was already a disjoint idea, how queries had `before_` and `after_` (ok, pretty normal), but fields had `instrument(type, field)` (pretty weird). So, I wouldn't miss it too much :P

lib/graphql/schema/member/has_fields.rb
```diff
@@ -44,6 +45,34 @@ def field_class(new_field_class = nil)
           end
         end
 
+        # Call `method_name` as a wrapper around each of `field_names`.
+        # If no `field_names` are passed, `method_name` will be called around _all_ fields.
+        # @param field_names [Array<Symbol>]
+        # @param method_name [Symbol]
+        # @return [void]
+        def around_resolve(*field_names, method_name)
```
jeffcarbs: This method signature seems a bit weird. I think it would make more sense as either:
```ruby
def around_resolve(method_name, *field_names)
end

# Usage:
# around all fields
around_resolve :upcase_value
# around just the listed fields
around_resolve :upcase_value, :name, :description
```
I think even more intuitive would be to borrow from rails `around_action`:
```ruby
def around_resolve(method_name, only: nil, except: nil)
end

# Usage:
# around all fields
around_resolve :upcase_value
# around just the fields listed
around_resolve :upcase_value, only: [:name, :description]
around_resolve :upcase_value, only: :name, :description
# around all except the fields listed
around_resolve :upcase_value, except: [:name, :description]
around_resolve :upcase_value, except: :name, :description
```

rmosolgo:
I'm going to let this rest and re-investigate options like this in the context of pagination (#1359)

PR #1166, opened 2017-12-02
 nil-check should be avoided.

johnunclesam:
https://github.com/troessner/reek/issues/1276#issuecomment-348709808


rmosolgo:
Personally, I prefer as few `return` expressions as possible. If you want to apply some other preferences to this project, consider opening an issue to discuss it!

PR #1169, opened 2017-12-04
Fix(GraphQL::Schema::Field) use the GraphQL name as the hash key for arguments

### LINKED ISSUE #1168: [1.8.0.pre2] Arguments with underscores not being passed through to resolve method
I've been playing around with the new class-based API in 1.8.0.pre2, and I've run into an issue.

I have the following `MutationType`:

```
class Types::MutationType < Types::BaseObject
  field :create_user, Types::User, 'Create a new user', null: false do
    argument :first_name, String, "The user's first name", required: false
    argument :last_name, String, "The user's last name", required: false
    argument :email, String, "The user's email", required: true
    argument :about, String, 'Details about the user', required: true
  end

  def create_user(**args)
    puts args
  end
end
```

From my reading of the docs, all of the arguments above that are passed in a query should be getting passed down to `create_user` in keyword args. However when I run a mutation like so:

```
mutation {
  createUser(
    firstName: "Matt", 
    lastName: "Engledowl", 
    email: "test@test.com", 
    about: "Software developer, father, graphql/ruby enthusiast"
  ) {
    id
    firstName
    lastName
    email
  }
}
```

`puts args` results in the following in the logs:

```
{:email=>"test@test.com", :about=>"Software developer, father, graphql/ruby enthusiast"}
```

Notice that `first_name` and `last_name` did not come through. It appears that arguments that have underscores in the name aren't correctly being passed down to the method. I further confirmed this by adding more underscored arguments and more non-underscored arguments and seeing the same results.

There's another strange thing though. I was curious what would happen if I didn't use snake_case in my argument name, so I changed `first_name` to be `firstName`:

```
class Types::MutationType < Types::BaseObject
  field :create_user, Types::User, 'Create a new user', null: false do
    argument :firstName, String, "The user's first name", required: false # <-- here's the line I changed
    argument :last_name, String, "The user's last name", required: false
    argument :email, String, "The user's email", required: true
    argument :about, String, 'Details about the user', required: true
  end

  def create_user(**args)
    puts args
  end
end
```

Ran the same exact mutation, and here's what I get in the logs:

```
{:first_name=>"Matt", :email=>"test@test.com", :about=>"Software developer, father, graphql/ruby enthusiast"}
```

I'm guessing this has something to do with the code that's converting between camelCase and snake_case, but I haven't actually dove in to look.

This is on a fresh repo, ruby 2.4.0, rails 5.1.4.

rmosolgo:
Fixes #1168 

I noticed the same bug while converting the introspection schema to class-based definitions. I'll push a test before merging.

cc @mengledowl


swalkinshaw:
Oops, think this was me. Thanks 👌

rmosolgo:
Blame the maintainer for poor test coverage :P

rmosolgo:
Merged as part of #1170

PR #1170, opened 2017-12-05
[1.8] custom introspection

### LINKED ISSUE #733: Can metadata information be included in the introspection query?
I admit I haven't looked into the spec or the code much, but: one can define [metadata definitions](https://github.com/rmosolgo/graphql-ruby/blob/45a5a0f0be359060897b39f127bf4c09b41b5720/lib/graphql/define.rb) for new properties on GraphQL items. Could this be exposed in the introspection query? 

I think it'd be neat for tooling _cough_ https://github.com/gjtorikian/graphql-docs _cough_ to be able to represent this information. I am not sure if the use case is so narrow that it doesn't apply, and that's fine. I just want to open a discussion to see if it makes sense.

### LINKED ISSUE #793: Only way to override introspection types is to reassign constants
I needed to add some fields to the `__Field` introspection type. The only way to do this was to reassign `GraphQL::Introspection::FieldType`, which is triggers a warning since it's a constant reassignment.

The only way to really do it in a clean way is like this:

```ruby
GraphQLIntrospectionFieldType = GraphQL::Introspection::FieldType
GraphQL::Introspection.send(:remove_const, :FieldType)
GraphQL::Introspection::FieldType = GraphQLIntrospectionFieldType.redefine do ... end
```

It'd be great if you could optionally specify the introspection types on construction of a schema.

rmosolgo:
A beast of a PR to add custom introspection to GraphQL-Ruby 

- [x] custom types for schema introspection 
- [x] custom introspection entry points (a la `__type(name:)` & `__schema`)
- [x] custom dynamic fields (a la `__typename`)

I've added docs [here](https://github.com/rmosolgo/graphql-ruby/blob/1.8-custom-introspection/guides/schema/class_based_api.md).

Fixes #793 
Fixes #733 

Also, this includes some other interesting stuff: 

- Inheriting name & description (needs test) 
- Introspection schema totally converted to class-based API 
- `GraphQL::Schema::LateBoundType`: for referencing types by name, this could be nice to make public somehow
- `extras:` API for sneaking `:irep_node` and `:ast_node` into method resolve. Should I document that, or just let it disappear and see if anyone needs it?


spec/graphql/schema/object_spec.rb
```diff
@@ -29,6 +29,8 @@
       name_field = new_object_class.fields.find { |f| f.name == "name" }
       assert_equal "The new description", name_field.description
     end
+
+    it "inherits name and description"
```
gjtorikian: this is empty 🙊

rmosolgo: 😉 

image

rmosolgo: added, thanks for the reminder!

PR #1171, opened 2017-12-05
Fix code climate?

rmosolgo:


rmosolgo:
I'll report it to code climate, but then it's a bug in `ruby_parser`, I get the same issue when I try it locally.

PR #1174, opened 2017-12-06
squash double negative

stanishev:


rmosolgo:
Oopsie,  thanks for the correction!!

Issue #1158, opened 2017-12-01
1.8 Feedback: Allow description in the block for a field


esjee:
It appears that in 1.8 it's no longer possible to provide a description using the block argument. Instead the only way to provide a description is to have it be the third argument. So:

Pre 1.8:
```
# this works
   field :triggered_pre_submission_trigger, -> { TriggerType } do
    description 'A pre-submission trigger that notified the hacker before submission. '\
                'This field is only present for reports filed after February 14, 2016.'
  end
```

1.8:
```
# this works
   field(
     :triggered_pre_submission_trigger,
     -> { TriggerType },
     'A pre-submission trigger that notified the hacker before submission. '\
       'This field is only present for reports filed after February 14, 2016.'
     null: true
```

1.8:
```
# this doesn't work, but I would like it to
   field(
     :triggered_pre_submission_trigger,
     -> { TriggerType },
     null: true do
     description 'A pre-submission trigger that notified the hacker before submission. '\
       'This field is only present for reports filed after February 14, 2016.'      
      end
```

There are two reasons why I would like the support the latter.

- It makes it easier to create an automatic upgrade (see graphql_upgrader) to 1.8
- This is subjective, but I think the second style is a bit nicer when it comes to long descriptions

rmosolgo:
It's ok with me to support it, I heard somewhere (?) that it was not helpful to have "so many ways to do things", and better to have one way. So, either I picked the _wrong way_, or we should keep having the different ways!

esjee:
In general, I think that comment makes sense. I can see how it would be helpful to have one 'official' way of doing things. It makes the library less complex. 

At the same time though, this usage can already be observed in classess. For example:

1.8
```
# This works
class Types::TodoList < GraphQL::Schema::Object # or Scalar, Enum, Union, whatever
  description "Things to do (may have already been done)"
end
```

For this reason, I feel that the added complexity is not a major concern.

Are there any other reasons I could have missed to warrant not implementing this?

rmosolgo:
🤷‍♂️  Unless someone disagrees, go for it!

swalkinshaw:
Semi-related, but it kind of surprised me that `description` wasn't a kwarg in 1.8; it's just a positional argument. It's weird since descriptions are optional but I guess since it's the *last* positional arg it matters less.

rmosolgo:
Yeah, I think a kwarg would be nice too, it makes for some weird, long lines as it is.

[...]

PR #1175, opened 2017-12-07
allow description to be defined within field block

esjee:
It seems there are no objects to https://github.com/rmosolgo/graphql-ruby/issues/1158, so let's implement it!

### DIFF
--- a/spec/graphql/schema/field_spec.rb
+++ b/spec/graphql/schema/field_spec.rb
@@ -14,5 +14,17 @@ describe GraphQL::Schema::Field do
     it "camelizes the field name" do
       assert_equal 'inspectInput', field.graphql_definition.name
     end
+
+    describe "description in block" do
+      it "will raise if description is defined both in the argument and in the block" do
+        assert_raises RuntimeError do
+          class DescriptionInBlock < Jazz::BaseObject
+            field :should_raise, Jazz::Key, "this should not raise" do
+              description "This should raise"
+            end
+          end
+        end
+      end
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


Willianvdv:
@esjee what happens if you define the description in the block _and_ as an argument? Should we refuse the double definition?

Willianvdv:
@esjee lol you already did that 💃 ! Can you spec this behaviour?

spec/graphql/schema/field_spec.rb
```diff
@@ -14,5 +14,17 @@
     it "camelizes the field name" do
       assert_equal 'inspectInput', field.graphql_definition.name
     end
+
+    describe "description in block" do
+      it "will raise if description is defined both in the argument and in the block" do
+        assert_raises RuntimeError do
+          class DescriptionInBlock < Jazz::BaseObject
+            field :should_raise, Jazz::Key, "this should not raise" do
```
esjee: This doesn't work! :thinking:

spec/support/jazz.rb
```diff
@@ -59,9 +59,9 @@ def to_graphql
   # A custom field class that supports the `upcase:` option
   class BaseField < GraphQL::Schema::Field
     argument_class BaseArgument
-    def initialize(*args, options, &block)
+    def initialize(*args, **options, &block)
       @upcase = options.delete(:upcase)
-      super(*args, options, &block)
+      super(*args, **options, &block)
```
Willianvdv: Nice catch! 🐟

esjee: :muscle:

rmosolgo:
👌 !

PR #1181, opened 2017-12-09
return false for hasNextPage when using last: qualifier

geneeblack:
I can think of no conditions where hasNextPage would ever be other than false when last:n is used

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -27,7 +27,11 @@ module GraphQL
         if first
           paged_nodes_length >= first && sliced_nodes_count > first
         elsif GraphQL::Relay::ConnectionType.bidirectional_pagination && last
-          false #sliced_nodes_count > last
+          if (before || after)
+            sliced_nodes_count > last
+          else 
+            false
+          end
         else
           false
         end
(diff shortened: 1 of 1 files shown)


rmosolgo:
I think `hasNextPage` may be true in a case like this: 

```ruby
# given these items: 
[a, b, c, d, e, f, g]
#      |--|
items(last: 2, before: e)
# => nodes [c, d]
#    hasPreviousPage: true 
#    hasNextPage: true
```

What do you think?

geneeblack:
I would think that given the last 2 before e there would be no next page as long as the before e statement was in effect since there could be no items on the next page that are before e.

geneeblack:
Ah, yah. I guess that's a little heavy handed to ensure last returns a false when not using before/after. maybe the second commit will address that edge case better.

rmosolgo:
Thanks for the update, I see the tests are passing now (sorry about the JRuby situation, JRuby is a constant struggle). Can you add a test which demonstrates the bug, or describe the case where you noticed this issue? I want to make sure we don't introduce a regression in this area in the future!

olleolleolle:
@geneeblack The JRuby tests are green on current master - could you try on top of that?

PR #1185, opened 2017-12-12
Give introspection fields a complexity of 0

nneufeld:
I think typically users wouldn't expect introspection fields such as `__typename` to contribute to a query's complexity, this PR will have those fields given a complexity of 0 rather than the default of 1. Since these fields are automatically defined by the gem, users of the gem also have no way of changing their complexity, so 0 seems like a better default for them.

### DIFF
--- a/lib/graphql/field.rb
+++ b/lib/graphql/field.rb
@@ -185,7 +185,6 @@ module GraphQL
     attr_accessor :arguments_class
 
     attr_writer :connection
-    attr_writer :introspection
 
     # @return [nil, String] Prefix for subscription names from this field
     attr_accessor :subscription_scope
@@ -231,6 +230,11 @@ module GraphQL
     # @return [Boolean] Is this field a predefined introspection field?
     def introspection?
       @introspection
+    end
+
+    def introspection=(new_introspection_value)
+      @introspection = new_introspection_value
+      @complexity = 0 if introspection?
     end
 
     # Get a value for this field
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for describing the issue and suggesting an implementation! I have some related ideas, what do you think: 

- in 1.8, graphql-ruby will support custom introspection types & fields via #1170, so these can be properly customized 
- also in 1.8, you can customize the `field_class` which is used to convert field configurations into `GraphQL::Field`s. You could put this logic in your custom field to create `complexity = 0` whenever `introspection` is true. 

Would those solutions work in this case? 

I agree that `0` would be a better value for `__typename`, but for the _rest_ of the introspection system (eg `__schema { ... }`), I think the complexity analysis should be _improved_ rather than removed. For example, `__Schema.types` currently costs a constant `1`, but it should cost `number_of_types * child_complexity`. I don't really have bandwidth to redo that right now, but what do you think of the idea?

nneufeld:
Both those solutions in 1.8 sound like they'd address this case, the 2nd one in particular as it sounds like with the first, each introspection field would have to be individually overridden.

I agree that fields like `__Schema.types` would benefit from a dynamic complexity definition, but thinking more broadly about complexity, I wonder if defining specific complexity values is something that the gem should be concerned about. How complexity is defined, what sort of scale is used, those are things that may vary significantly between different users of the gem. 

What if the gem were to default complexity to 0 rather than 1, and allows users to configure a  default based on their own complexity definition/scale? The basic case would just be setting a `default_complexity` when defining your schema, and the custom `field_class` option in 1.8 could allow more complex logic(ie. use a default complexity based on type, introspection, etc...)

PR #1186, opened 2017-12-13
[1.8] Allow resolve within field block [Don't merge]

cpunion:
Base on following reasons:

1. We don't have a good transformer to convert resolve within field block to argument
2. It's ugly when use resolve and argument likes below:

```ruby
field :project, Types::ProjectTypes, null: false,
  resolve: ->(obj, args, ctx) {
    #
    # Very
    #
    # long
    #
    # code
    #
    # block
    #
  } do
end
```

The following is better:

```ruby
field :project, Types::ProjectTypes, null: false do

  resolve ->(obj, args, ctx) {
    #
    # Very
    #
    # long
    #
    # code
    #
    # block
    #
  }
end
```

But currently missing some checking for resolve_proc, just record it as an issue.


rmosolgo:
Do you think we could also solve this by adding a transformation to `upgrader`? Here's my attempt: 

https://github.com/rmosolgo/graphql-ruby/pull/1212/commits/e7de3249256c177504395869ee0738b6d5885bd9

rmosolgo:
I've been having good success with the upgrader, which moves resolve procs into method definitions. Please open another issue if it doesn't work for you! I want to leave resolve procs behind us if I can, they are unfamiliar to newcomers.

PR #1188, opened 2017-12-13
[DRAFT] Upgrade with AST parser

cpunion:
@Willianvdv 

Just for discussion. rewrite with AST parser.

Code need refactor but I have no enough time to do, but I think the transformer based on AST can do more.

Feel free to close it.


cpunion:
Rebased on last commit of 1.8-dev for review.

rmosolgo:
I really like this approach, but I realized one issue is with rewriting: I think it will remove comments. (This came up when chatting with `@willianvdv` as well, and it's part of why he chose regexps.)

Personally, this is a stopper, since we don't want to throw out people's code comments. Is anyone interested in more exploration here? Or is there a good way to work around this? 

For my part, I'm going to keep exploring options with regexps, and maybe revisit a hybrid AST/regexp approach. For example, you might be able to _parse_ the AST but use regexp substitution in small, targeted ways.

cpunion:
For comments, maybe I did wrong for some place. This is an example that it kept comment:

https://github.com/cpunion/graphql-ruby/blob/f8c9dbf4b2c22256f83a02dae60869951e26f888/spec/graphql/upgrader/member_spec.rb#L252-L281

It kept comment in the line:
```ruby
type -> { types.String } # name's type is String
```

I will look it when I'm free.

cpunion:
Aha, it works fine, missing the comment in your case.

cpunion:
> For my part, I'm going to keep exploring options with regexps, and maybe revisit a hybrid AST/regexp approach. For example, you might be able to parse the AST but use regexp substitution in small, targeted ways.

Agree. The code of transforming AST is not readable, sometimes it's too lengthy.

rmosolgo:
🤦‍♂️  sorry!

rmosolgo:
Thanks for pushing us in this direction, I've applied a hybrid approach in `1.8-dev`, please open an issue if you find a shortcoming of that implementation!

PR #1196, opened 2017-12-19
Fix documentation on `ObjectType.mutation`

xuorig:
Very nit picky and non urgent change, just stumbled upon it 🤷‍♂️


rmosolgo:
👌

PR #1198, opened 2017-12-20
Update error_handling.md

samirph:
There was an error in the Error Handling with Wrappers code example.
Changed e.record.errors.full_messages.join("\n") to err.record.errors.full_messages.join("\n")


rmosolgo:
oops, thanks for the fix on this!

PR #1200, opened 2017-12-20
Pass along context in Query#result

### LINKED ISSUE #1121: Per-query backtrace wrapping doesn't work
From http://graphql-ruby.org/queries/backtrace_annotations.html.

Adding the backtrace module with `use GraphQL::Backtrace` at the schema level works correctly. Adding it by passing `backtrace: true` in the query's context does not.

The reason is that the backtrace module gets added as a tracer to the query object itself https://github.com/rmosolgo/graphql-ruby/blob/21fb53ced3ada7e088c22f7442bcfcedb05bdabb/lib/graphql/query.rb#L89
but the multiplex tracing event occurs on the generated multiplex object
https://github.com/rmosolgo/graphql-ruby/blob/21fb53ced3ada7e088c22f7442bcfcedb05bdabb/lib/graphql/execution/multiplex.rb#L57-L58
which does not have the backtrace tracing installed.

The backtracing module relies on the `execute_multiplex` event in order to install its rescue hook https://github.com/rmosolgo/graphql-ruby/blob/21fb53ced3ada7e088c22f7442bcfcedb05bdabb/lib/graphql/backtrace/tracer.rb#L26-L31

So 99% of the backtrace module is present and working (the context is fully populated etc.) but because the rescue is never hooked, exceptions don't actually get wrapped.

@rmosolgo @swalkinshaw

eapache:
Should fix #1121.

I looked for a place to add a test for this and couldn't figure out a good one.

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -157,7 +157,7 @@ module GraphQL
     def result
       if !@executed
         with_prepared_ast {
-          Execution::Multiplex.run_queries(@schema, [self])
+          Execution::Multiplex.run_queries(@schema, [self], context: @context)
         }
       end
       @result ||= Query::Result.new(query: self, values: @result_values)
--- a/spec/graphql/backtrace_spec.rb
+++ b/spec/graphql/backtrace_spec.rb
@@ -126,6 +126,16 @@ describe GraphQL::Backtrace do
       assert_includes err.message, "more lines"
     end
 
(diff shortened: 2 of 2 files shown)


rmosolgo:
:+1: This seems good to me, I will try to find a place to test it to avoid regressions

rmosolgo:
I added a test for the bug described in #1121, thanks again for the fix!

PR #1212, opened 2018-01-08
[1.8] Make upgrader more configurable

rmosolgo:
I want to use `GraphQL::Upgrader` to work on GitHub's codebase, but we have some ... unique ... configurations, so I'll need to open it up for extension. Here's my plan: 

- Break each part of the upgrader into a composable unit (a class :tada: 😩  ... 😆 ) 
- Make the upgrader accept a list of steps 
- Ship with a default list of steps; accept overrides of any user-defined steps 

This way, we can ship with some defaults, but also support custom cases. 

Also TODO: 

- [x] For some reason I made it write to `description: "..."` ~~which is not supported. Oops~~. Description works as a kwarg now
- [ ] Interface methods need to go in an `Implementation` module


rmosolgo:
There's an outstanding TODO here, but I'm going to push it to a later task: 

https://github.com/rmosolgo/graphql-ruby/issues/1215

PR #1213, opened 2018-01-10
[1.8] Bug: new-style interface w/method-based fields not inherited by old-style objects

rmosolgo:
Trying to upgrade an interface in GitHub, I found this incompatibility: 

- A class-based interface
- A field on that interface is implemented with a method 
- The interface is implemented by an old-style object
- The old-style object does _not_ override the interface's method-based implementation 

In that case, the field fails to resolve. The reason is because there's no object class for the `Implementation` module to be inherited into. 

For the time being, I'm going to focus on upgrading object classes :S 

Please leave a comment here if you're affected by this and we can hunt for a solution.


rmosolgo:
This bug will not be fixed :'(

PR #1226, opened 2018-01-17
WIP => Fix object type fields dup

xuorig:


rmosolgo:
This is a weird fail 😖

rmosolgo:
Ok, it's starting to become clear: in order to support custom introspection, we have to make _completely_ fresh versions of `__Type`, `__Schema`, etc. Here's the code: 

https://github.com/rmosolgo/graphql-ruby/blob/fcd9f0b1d99fbb32734959103b85c184791d0eb0/lib/graphql/schema/introspection_system.rb#L13-L19

The reason is because we _can't_ reuse any previously-defined fields, because environments with multiple schemas might break. Here's the scenario: 

- Schema A uses the default introspection types
- Schema B uses a custom `__Type` class 

Let's say Schema A is built first, and `__Schema.types` creates a reference to the built-in `__Type`. Then, if Schema B tries to re-use the cached `__Schema` type, it will be broken, because Schema B's `__type` field returns _custom_ `__Type`, while `__Schema.types` returns _built-in_ `__Type`. 

Instead of using _cached_ `__Schema`, we should rebuild it from scratch, so that it can correctly find the custom `__Type`.

I'm not sure how to work around this 🤔

rmosolgo:
Is this change still needed? if so, I think we could cherry-pick cbdb114 on to master, but I don't remember what this was for 😬

PR #1227, opened 2018-01-17
[1.8] Instance_eval the Field instead of a FieldProxy

xuorig:
This PR takes another approach to Field definition blocks like this one:

```ruby
field :a, GraphQL::STRING_TYPE, null: true do
end
```

The `FieldProxy` that powers this API is currently hard to customize, making it hard to extend with custom methods like this:

```ruby
field :a, GraphQL::STRING_TYPE, null: true do
  my_custom_thing(true)
end
```

This PR remove `FieldProxy` and instead chooses a more explicit approach. Instead of using `FieldProxy`, the Field Class is passed to the block, letting users call `Field` methods directly on the object:

```ruby
field :a, GraphQL::STRING_TYPE, null: true do |f|
  f.argument :id, ID, required: true
  f.my_custom_thing(true)
end
``

  - First commit implements the change
  - Second commit fixes fixtures and tests

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -16,7 +16,7 @@ module GraphQL
       # @return [Hash{String => GraphQL::Schema::Argument}]
       attr_reader :arguments
 
-      def initialize(name, return_type_expr = nil, desc = nil, null: nil, field: nil, function: nil, description: nil, deprecation_reason: nil, method: nil, connection: nil, max_page_size: nil, resolve: nil, introspection: false, extras: [], &args_block)
+      def initialize(name, return_type_expr = nil, desc = nil, null: nil, field: nil, function: nil, description: nil, deprecation_reason: nil, method: nil, connection: nil, max_page_size: nil, resolve: nil, introspection: false, extras: [], &definition_block)
         if !(field || function)
           if return_type_expr.nil?
             raise ArgumentError, "missing positional argument `type`"
@@ -46,8 +46,22 @@ module GraphQL
         @extras = extras
         @arguments = {}
 
-        if args_block
-          with_proxy_dsl(&args_block)
+        if definition_block
+          instance_eval(&definition_block)
+        end
+      end
+
+      # This is the `argument(...)` DSL for class-based field definitons
+      def argument(*args)
+        arg_defn = self.class.argument_class.new(*args)
+        arguments[arg_defn.name] = arg_defn
+      end
+
+      def description(text = nil)
+        if text
+          @description = text
+        else
+          @description
         end
       end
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/graphql/schema/field.rb [hunk 1]
```diff
@@ -46,11 +48,17 @@ def initialize(name, return_type_expr = nil, desc = nil, null: nil, field: nil,
         @extras = extras
         @arguments = {}
 
-        if args_block
-          with_proxy_dsl(&args_block)
+        if block
```
swalkinshaw: Isn't this just `yield self if block_given?`. wouldn't need explicit `&block` param either.

rmosolgo:
I see this also switches from `field.instance_eval(&block)` to `block.call(field)`. That's ok with me but: 

- What do you think are the benefits/downsides here?
- We should also update the docs 
- How about a compatibility check? You can see if `block.arity` is `1` or `0`, and if it's `0`, warn, then use instance-eval. (I have written a lot of code that uses the old way now 😖 )

rmosolgo:
Oh, I didn't say in my previous comment, :+1: for yielding the field instance instead, good call!

lib/graphql/schema/field.rb
xuorig: lol yes ty

xuorig:
Good point @rmosolgo , ill try to make this backward compatible 💃

guides/schema/class_based_api.md
```diff
@@ -309,12 +309,16 @@ class Types::TodoList < Types::BaseObject
   field :name, String, "The unique name of this list", null: false
   field :is_completed, String, "Completed status depending on all tasks being done.", null: false
   # Related Object:
-  field :owner, Types::User, "The creator of this list", null: false
+  field :owner, Types::User, "The creator of this list", null: false do
   # List field:
-  field :viewers, [Types::User], "Users who can see this list", null: false
+  field :viewers, [Types::User], "Users who can see this list", null: false do |f|
+    # Field definitions still accept a block but now provide the field class
```
xuorig: Added this, not sure if it deserves it's own section explaining field definition?

xuorig:
Changed this PR to something even more simple:

  - Use `field.instance_eval` instead of `field_proxy.instance_eval`

This allows us to extend `Schema::Field` with a custom DSL and for it to be usable within the definition block.

It has the advantage of not breaking any existing code since it behaves the same.

Minus point is that we need special DSL methods for every single settable item... not a fan but no better way comes to mind.

Another option but more complex: allow to set a custom proxy so it can be extended

```
class MyFieldClass < Schema::Field
  self.proxy_dsl = MyClassWithSomeSuperCoolMethods
end
```

But I think for now this is the simplest approach to get the result we want 🤷‍♂️

rmosolgo:
> simplest approach to get the result we want
👌

lib/graphql/schema/field.rb
```diff
@@ -132,36 +144,6 @@ def argument_class(new_arg_class = nil)
...
-
-        # This is the `argument(...)` DSL for class-based field definitons
-        def argument(*args)
-          arg_defn = @argument_class.new(*args)
-          @field.arguments[arg_defn.name] = arg_defn
-        end
-
-        def description(text)
-          if @field.description
-            fail "`description` was already assigned to this field (#{@field.name}). Assign it as a keyword argument or in the block, but not both."
```
rmosolgo: Makes sense to remove this, since previously, duplication was _only_ possible with `description`, but now it's possible with lots of things, and we can't really guard against all of them.

PR #1229, opened 2018-01-18
[1.8] Support nullable list items

rmosolgo:
Previously, given type `T`, you could only create `[T!]`, not `[T]`. This is usually fine, but I want to keep compatibility with an existing schema.

### DIFF
--- a/guides/schema/class_based_api.md
+++ b/guides/schema/class_based_api.md
@@ -69,7 +69,7 @@ Previously, list types were expressed with `types[T]` and non-null types were ex
 
 - List types are expressed with Ruby Arrays, `[T]`, for example, `field :owners, [Types::UserType]`
   - By default, list members are _non-null_, for example, `[Types::UserType]` becomes `[User!]`
-  - If your list members may be null, add `,nil` to the array: `[Types::UserType, nil]` becomes `[User]` (the list may include `nil`)
+  - If your list members may be null, add `, null: true` to the array: `[Types::UserType, null: true]` becomes `[User]` (the list may include `nil`)
 - Non-null types are expressed with keyword arguments `null:` or `required:`
   - `field` takes a keyword `null:`. `null: true` means the field is nullable, `null: false` means the field is non-null (equivalent to `!`)
   - `argument` takes a keyword `required:`. `required: true` means the argument is non-null (equivalent to `!`), `required: false` means that the argument is nullable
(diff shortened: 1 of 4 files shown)


guides/schema/class_based_api.md
```diff
@@ -68,6 +68,8 @@ As described below, `.to_graphql` can be overridden to customize the type system
 Previously, list types were expressed with `types[T]` and non-null types were expressed with `!T`. Now:
 
 - List types are expressed with Ruby Arrays, `[T]`, for example, `field :owners, [Types::UserType]`
+  - By default, list members are _non-null_, for example, `[Types::UserType]` becomes `[User!]`
+  - If your list members may be null, add `,nil` to the array: `[Types::UserType, nil]` becomes `[User]` (the list may include `nil`)
```
eapache: this is a clever syntax, :+1:

eapache: though maybe `[Types::UserType, null: true]` would be even more obvious?

rmosolgo: i was going to ask ... good clever or bad clever 😅 

I thought about a keyword argument, say, `null_items:`, but the problem there is with nested lists. For example, how could you express `[[[T]!]]`? It's crazy, but I think it happens sometimes 😬

rmosolgo: I like your suggestion! Actually, I didn't know it was valid Ruby, but voila: 

```ruby
# $ irb
irb(main):001:0> [1, a: true]
# => [1, {:a=>true}]
irb(main):002:0> [[1, a: true], a: false]
# => [[1, {:a=>true}], {:a=>false}]
```

eapache: Are there any other args that you might want to apply to the inner item but not the field as a whole? That would be a point in favour of `[Type, nested_arg_hash]` since you could add args to the nesting later.

xuorig: Pretty cool, other option is always `[Types::UserType.to_non_null]` or similar but it definitely looks less nice 😁

eapache: There is very little that *isn't* valid ruby sometimes :P

rmosolgo:
Pushed with `null: true` as [discussed above](https://github.com/rmosolgo/graphql-ruby/pull/1229#discussion_r162434547). I also like the open possibility of more keyword arguments, although I certainly can't _think_ of what they might be. The other good thing is that it matches the _existing_ `null: true` API.

Issue #1230, opened 2018-01-19
Enable resolvers to return an array of GraphQL::ExecutionErrors


stanishev:
When a field resolver returns a single [GraphQL::ExecutionError](http://www.rubydoc.info/gems/graphql/1.7.8/GraphQL/ExecutionError), the field is resolved as nil and the error is added to the response's `errors` key.
graphql-ruby ensures that only that one error is added to the errors key, even if the field is marked as `non_null`.

**There is no mechanism to achieve the same behavior for multiple errors for non_null fields.** 

The [add_error](http://www.rubydoc.info/gems/graphql/1.7.8/GraphQL/Query/Context/SharedMethods#add_error-instance_method) method allows adding multiple errors to the errors key, but resolvers are still required to return a value if the field is marked as non_null. 

This leaves us with only the following options for a non_null field:
- returning nil and having an additional error added in the errors key by graphql-ruby (the error indicates that a nil value is not permitted for a non_null field) 
- or to use add_error and return a bogus non nil value of the correct return type.
- or use add_error for the first `n-1` errors and return the last error 

none of which are very nice. 

**Would like to suggest that GraphQL::Execution::Execute.resolve_value returns a PropagateNull type when the result from a resolve method is an array of `GraphQL::ExecutionError`s**

PS: Can't be sure but looking through the code in [resolve_value](https://github.com/rmosolgo/graphql-ruby/blob/56e2e884519919d4d6179d624e89e95ca6b271ca/lib/graphql/execution/execute.rb#L145-L153) it seems that this was perhaps intended to begin with... 

also this PR doesn't handle the case where the return type is an Array where only *some* of the elements are GraphQL::ExecutionErrors.. Quite frankly I am not sure how that should be handled. Hopefully rmosolgo can weight in.

stanishev:
PR was merged. 💥

PR #1231, opened 2018-01-19
Allow resolvers to return multiple errors

stanishev:
This PR is to enable resolvers to return an **array** of `GraphQL::ExecutionError`s and to have the same behavior we see when a single GraphQL::ExecutionError is returned from a resolver.

More detailed description here: https://github.com/rmosolgo/graphql-ruby/issues/1230

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -178,6 +178,12 @@ module GraphQL
             else
               nil
             end
+          elsif value.is_a?(Array) && value.any? && value.all? {|v| v.is_a?(GraphQL::ExecutionError)}
+            if field_type.kind.non_null?
+              PROPAGATE_NULL
+            else
+              nil
+            end
           elsif value.is_a?(Skip)
             field_ctx.value = value
           else
--- a/spec/graphql/execution_error_spec.rb
+++ b/spec/graphql/execution_error_spec.rb
@@ -271,4 +271,22 @@ describe GraphQL::ExecutionError do
       assert_equal(expected_result, result)
     end
   end
+  
+  describe "more than one ExecutionError" do
+    let(:query_string) { %|{ multipleErrorsOnNonNullableField} |}
+    it "the errors are inserted into the errors key and the data is nil even for a NonNullable field " do
+      expected_result = {
+          "data"=>nil,
+          "errors"=>
+              [{"message"=>"This is an error message for some error.",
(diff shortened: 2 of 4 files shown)


rmosolgo:
Thanks for the detailed description in the issue and for proposing this fix, I'll take a close look soon!

rmosolgo:
This fix looks good to me, do you mind adding a test case for it? You can either extend one of the (giant) fixture schemas that are already in the test suite, or you can write a small schema for testing this behavior in isolation, then demonstrate the new behavior that you added. That way, we can be sure it will keep working in the future!

stanishev:
Will do, thanks for the suggestion (I thought about adding a test case with the original commit but felt disoriented within the test suite and wasn't sure where to start, I can go back and look at it now with more focus).

cabello:
I am curious, does this violate the spec?

> If the field returns null because of an error which has already been added to the "errors" list in the response, the "errors" list must not be further affected. That is, only one error should be added to the errors list per field.

http://facebook.github.io/graphql/October2016/#sec-Errors-and-Non-Nullability

stanishev:
I am honestly confused by the wording in that spec. The errors key is a list of errors, the only required key is `message`.  To not have multiple errors per field (as that final sentence on its own implies) would be very limiting in my opinion, but I also can't make any sense of that statement because there is nothing in the graphQL spec for an error that ties the error to a field: [errors spec](https://facebook.github.io/graphql/October2016/#sec-Errors)

My take is that neither this PR (nor the existing [add_error](http://www.rubydoc.info/gems/graphql/GraphQL/Query/Context/SharedMethods#add_error-instance_method) method) are a violation of the graphql specs, but @rmosolgo is the final arbiter in this.

cjoudrey:
Here's a related issue in the reference implementation: https://github.com/graphql/graphql-js/issues/205

rmosolgo:
Hmm, I suppose graphql-ruby _already_ violates the spec in two ways: 

> If an error is thrown while resolving a field, it should be treated as though the field returned null...

In fact, graphql-ruby allows adding errors via `add_error` and _still_ returning a response. (`add_error` was added for this very purpose a long time ago)

> That is, only one error should be added to the errors list per field. 

As @stanishev  points out, this is already possible via multiple calls to `add_error`. 

So, I don't think this PR makes the violation any _worse_ 😆  And of course, any user who doesn't wish to violate the spec in this way can simply use `raise` to ensure that: 

- only one error field; AND 
- fields with errors return null.

PR #1232, opened 2018-01-21
`Schema#find` Finding schema members using string paths

xuorig:
Allows finding schema members using string paths. Helpful for when referencing members lazily by their name for example.

Examples are provided in the class documentation: 

```ruby
# Find schema members using string paths
#
# @example Finding object types
# MySchema.find("SomeObjectType")
#
# @example Finding fields
# MySchema.find("SomeObjectType.myField")
#
# @example Finding arguments
# MySchema.find("SomeObjectType.myField.anArgument")
#
# @example Finding directives
# MySchema.find("@include")
#
```

IDL directives are intentionally left out for now until it becomes clearer / spec is updated / this gem is updated.

cc @rmosolgo


lib/graphql/schema/finder.rb
```diff
@@ -0,0 +1,153 @@
...
+    # MySchema.find("@include")
+    #
+    class Finder
+      class MemberNotFoundError < ArgumentError; end
+
+      def initialize(schema)
+        @schema = schema
+      end
+
+      def find(path)
```
olleolleolle: Perhaps add a docblock like

```ruby
# @param path [String] dot-separated path to find
# @raise [MemberNotFoundError] if path could not be found
```

My description may be missing the distinction between type/directive, but you know best how to describe that.

rmosolgo:
👍  I think this is a great addition! 

Does this return something for `__typename`? Should it? (I think `get_field` returns something there.)

xuorig:
good question @rmosolgo. I forgot to use `get_field` in this PR too 🐒 added a test too!

rmosolgo:
thanks for upstreaming this! 👏

PR #1233, opened 2018-01-22
Update link to Relay Global Object Identification spec

christopherstyles:
The current link in [Object Identification](http://graphql-ruby.org/relay/object_identification) is 404ing, as the url for Facebook’s _Relay Global Object Identification Specification_ has changed.


rmosolgo:
Oops, thanks for the update!

PR #1241, opened 2018-01-25
CodeClimate v2 configuration

olleolleolle:
- see https://docs.codeclimate.com/docs/advanced-configuration

Or, more exactly: https://docs.codeclimate.com/docs/advanced-configuration#section-analysis-configuration-versions

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,4 +2,6 @@ version: "2"
 exclude_patterns:
   - "lib/graphql/language/lexer.rb"
   - "lib/graphql/language/parser.rb"
+  # plugin - duplication - is failing on these
   - "spec/graphql/language/printer_spec.rb"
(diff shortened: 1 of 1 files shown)


olleolleolle:
@rmosolgo Hi, this PR adds the V2 variant of the CodeClimate configuration to the repo, which stops the badge in the README from being _gray_ (due to a build failure on one of the files which are now in the "exclude pattern" list.

rmosolgo:
Oh, that would be nice! I tried to debug but I never made it that far. Here are the results for this branch, but it looks like it still failed: 

https://codeclimate.com/github/rmosolgo/graphql-ruby/builds/138

Maybe we should exclude another file?

olleolleolle:
@rmosolgo Yes! It got there!

rmosolgo:
🎉  Awesome, thank you for figuring this out!

PR #1246, opened 2018-01-28
[1.8] Bug: Missing max_complexity class method in schema

noku:
Trying to upgrade the [schema](https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/guides/schema/class_based_api.md#schema-class) I encountered this error:

`ArgumentError (wrong number of arguments (1 for 0)):` on line with `max_complexity` method definition, adding class method `max_complexity` to the schema fixes the issue.


rmosolgo:
Thanks for the fix!

Issue #1205, opened 2017-12-24
NotImplementedError (schema contains Interfaces or Unions, so you must define a resolve_type -> (obj, ctx) { ... } function):


johnunclesam:
If I follow this guide: http://tech.eshaiju.in/blog/2017/06/09/dry-graphql-definitions-using-interfaces/

I have this error:
`NotImplementedError (schema contains Interfaces or Unions, so you must define a resolve_type -> (obj, ctx) { ... } function):`

I'm using a simple `player_type.rb`:

```ruby
Types::PlayerType = GraphQL::ObjectType.define do
  name 'Player'

  interfaces [Interfaces::ActiveRecordTimestamp]

  field :id, !types.ID
  field :team_id, !types.ID
end
```

and this:

```ruby
Interfaces::ActiveRecordTimestamp = GraphQL::InterfaceType.define do
  name 'ActiveRecordTimestamp'

  field :createdAt, types.String, property: :created_at
  field :updatedAt, types.String, property: :updated_at
end
```

thefliik:
Sounds like a problem with the guide you're using. Check out [graphql-ruby's website.](http://graphql-ruby.org/guides) Specifically, the [Interface and Union types](http://graphql-ruby.org/types/abstract_types.html) guide. It explains how to fix your problem.

rmosolgo:
Yep, thanks for the link to those docs above!

johnunclesam:
It doens't work. I don't undestand how to do. I receive all the time this error:

`NotImplementedError (schema contains Interfaces or Unions, so you must define a `resolve_type -> (obj, ctx) { ... }` function):`

johnunclesam:
To fix this problem I used this in my `project_schema.rb` file:

```
ProjectSchema = GraphQL::Schema.define do
  query(Types::QueryType)
  mutation(Types::MutationType)

  resolve_type lambda { |_obj, _ctx|
  }
end
```

But I TOTALLY don't understand why!

**Why?**

johnunclesam:
Now I can use interfaces in my types files.

johnunclesam:
I ended up using this in `project_schema.rb`:

```
...
resolve_type ->(type, obj, ctx) {}
```

But please explain it to me what it does. And **why it doesn't work here**:

```
Interfaces::ActiveRecordTimestamp = GraphQL::InterfaceType.define do
  name 'ActiveRecordTimestamp'

  field :createdAt, types.String, property: :created_at
[...]

PR #1247, opened 2018-01-29
[docs] Abstract Types - note about resolve_type

olleolleolle:
This PR adds wording to the UnionType `resolve_type` docs, so that new users do not remove the `resolve_type` function from their Schema.

This is linguistic nudging.

### DIFF
--- a/guides/types/abstract_types.md
+++ b/guides/types/abstract_types.md
@@ -36,7 +36,7 @@ end
 
 ## Type-Level Resolution Hooks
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
Here's an Issue reported which noted that this is "how things behave": https://github.com/rmosolgo/graphql-ruby/issues/1205

rmosolgo:
Yes, better to have correct docs while that bug goes unfixed. Thanks for the improvement!

Issue #1216, opened 2018-01-11
[1.8] Upgrader should camelize argument names in definition & method body


rmosolgo:

esjee:
Fixed by #1251?

rmosolgo:
thanks!

PR #1251, opened 2018-01-31
Upgrader: underscore & symbol args access

rmosolgo:
Oops, the upgrader didn't convert strings to symbols, which are required for Ruby kwargs. Also, it should convert the used arguments to underscore-case.


rmosolgo:
Fixes #1216

PR #1252, opened 2018-01-31
fix(Relay::Node) use consistent description for id fields

rmosolgo:
oops, these could be different, but I think they were meant to be the same.


rmosolgo:
I'm going to address this locally instead, it causes a big change to the schema representation. If someone else is interested in adding this, feel free to comment here and we can find something that works!

PR #1255, opened 2018-02-01
Fix printing object type definitions

swalkinshaw:
Directives come after the implemented interfaces.

See
https://github.com/graphql/graphql-js/blob/360fa06e43fed32b63fba99b353515e7b56eee20/src/language/printer.js#L117-L129
for the canonical printer.

This is pretty difficult to add a test for (I think?) and likely why there wasn't any to begin with.

@xuorig any suggestions for how to test this? Since right now we're building a document by parsing it but I don't think it supports directives like this?

### DIFF
--- a/lib/graphql/language/printer.rb
+++ b/lib/graphql/language/printer.rb
@@ -154,8 +154,8 @@ module GraphQL
       def print_object_type_definition(object_type)
         out = print_description(object_type)
         out << "type #{object_type.name}"
-        out << print_directives(object_type.directives)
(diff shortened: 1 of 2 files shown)


xuorig:
oh damn 😨 

AST nodes can have direcitves I think, so you could build your own using Nodes and print it.

```
Nodes::ObjectType.new(
  directives: ...,
  interfaces: ...,
)
```

would that work? You'll need to find the right objects for directives and interfaces (TypeNameNode i think there?) but it should work

swalkinshaw:
I was making it harder than it was. I just updated the existing kitchen sink schema to include this case.

rmosolgo:
👍

PR #1260, opened 2018-02-02
Static validation for deprecated fields in query

piotrze:
This validation is optional. I find it useful to check if my queries are not using deprecated fields.


rmosolgo:
Hi @piotrze , I'm glad you found something that works for you! But I don't want to include it in graphql-ruby right now. Instead, we have another feature for this use case: 

http://www.rubydoc.info/gems/graphql/1.7.9/GraphQL/Analysis/FieldUsage#used-deprecated-graphql-fields-cheesefatcontent

There's one advantage to using a query analyzer, it can find deprecated fields when the query references an Interface, and one of those interface's fields is deprecated. (That should be rare, because usually you would deprecate at interface-level, too.)

PR #1261, opened 2018-02-04
Update link to Relay Connections documentation

christopherstyles:
The documentation url for Relay Connections has changed. This update refreshes the link in the [Connections](http://graphql-ruby.org/relay/connections) section of the docs to reflect the new url.


rmosolgo:
Thanks for the fix!

PR #1268, opened 2018-02-07
Fix error whilst parsing empty inline fragments

gjtorikian:
You know how much I love parsing problems, @rmosolgo. 😆 

Although this gem will accept a query string like this:

```
query{
  viewer{
    login{
      ... on String {
        
      }
    }
  }
}
```

It is not correct, and blows up later:

```
undefined method `name' for #<GraphQL::Language::Nodes::InlineFragment:0x00005588208418d8>

vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/static_validation/rules/fields_have_appropriate_selections.rb:27:in `map'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/static_validation/rules/fields_have_appropriate_selections.rb:27:in `validate_field_selections'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/static_validation/rules/fields_have_appropriate_selections.rb:12:in `block in validate'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:65:in `block in apply_hooks'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:65:in `each'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:65:in `reduce'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:65:in `apply_hooks'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:54:in `begin_visit'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:45:in `visit_node'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:47:in `block in visit_node'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:47:in `each'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:47:in `visit_node'
vendor/gems/2.4.3/ruby/2.4.0/gems/graphql-1.8.0.pre6/lib/graphql/language/visitor.rb:47:in `block in visit_node'
```

As well, GraphiQL complains:

image

That's because an empty set here is [not spec compliant](http://facebook.github.io/graphql/October2016/#SelectionSet).

A word of caution: I bumped the parser using ragel@6.10 even though the dependency checker asks for > 7.0.0.9. 6.10 is the latest stable release, and I didn't want to mess around with a dev version. It's _probably_ fine, but you may want to re-run the `build_parser` task locally if that's important? 🤷🏻‍♂️


gjtorikian:
THE PLOT THICKENS

@d12 tells me that the error might actually be on the spread on `login` itself:

```
{
  viewer {
    login {
      ... on String {
        id
      }
    }
  }
}
```

GraphiQL again rightfully complains:

image

It's a bit late where I'm at but I'll try to push up a fix for that generally when I can.

gjtorikian:
I'm a bit out of my depth, but a `debugger` here suggests that the validation rule responsible for this is not being entered: https://github.com/rmosolgo/graphql-ruby/blob/b5af7ad66dfddfcb219497f420f9baf8d08e1a65/lib/graphql/static_validation/rules/fragments_are_on_composite_types.rb#L15

gjtorikian:
I wonder if `FragmentSpread` needs to be added to the types to check? 🤔

gjtorikian:
I see now my mistake -- the validation error was somewhere completely different. This is now ready for review.

rmosolgo:
:+1: This looks great, thanks so much for tracking it down all over the place! I'll rebuild `lexer.rb` and push it, then merge as long as it has no changes. (I'm surprised it even _let_ you build it, it's supposed to assert version `7.0.0.9` 🤔 ) 

Thanks for the parser fix, I think I misinterpreted the relevant spot of the spec: 

https://facebook.github.io/graphql/October2016/#sec-Selection-Sets

Where `Selection<sub>list</sub>` means "one or more", but I thought it meant "zero or more". Now, I see that `<sub>opt, list</sub>` is used to mean "zero or more"!

gjtorikian:
> I'm surprised it even let you build it, it's supposed to assert version 7.0.0.9 🤔

I commented the assertion out. 😅

PR #1274, opened 2018-02-12
[1.8] Docs

rmosolgo:
I want to release 1.8 with some docs about the new APIs: 

- Types: 
  - [x] Objects, Fields, Arguments  
  - [x] Scalars 
  - [x] Enums & Enum Values 
  - [x] Input Objects 
  - [x] Interfaces 
  - [x] Unions 
  - [x] Lists 
  - [x] Non-Null Types 
- [x] Extending the type system 
- [x] Introspection 
- [x] Context 
- [x] Schema class 
- [x] Add a header flag about the class-based api 
- Any others?


guides/type_definitions/objects.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Field Return Type
+
+The second argument to `field(...)` is the return type. This can be:
+
+- A built-in GraphQL type (`Integer`, `Float`, `String`, `ID`, or `Boolean`)
+- A GraphQL type from your application
+- An _array_ of any of the above, which denotes a list type.
+
+Nullability is expressed with the required `null:`/ keywords:
```
swalkinshaw: errant `/`?

guides/type_definitions/objects.md
```diff
@@ -0,0 +1,282 @@
...
+end
+```
+
+If an argument has `required: true`, then all queries to the field _must_ provide a value for that argument. `required: false` means that the argument is optional.
+
+Arguments can also accept a description and a `default_value:`, for example:
+
+```ruby
+field :transactions, [Types::Transaction], null: false do
+  # Description is added after the type name
```
swalkinshaw: Not entirely relevant to this PR, but I didn't realize this was a positional arg. Wouldn't a kwarg be more consistent with how `field` does it?

guides/type_definitions/objects.md
rmosolgo: oops, yep!

guides/type_definitions/objects.md
rmosolgo: /shrug `field` does _both_ positional and kwarg. Fine with me to make `argument` also do both, but I haven't done it yet.

rmosolgo:
🚂  makin progress

guides/type_definitions/objects.md
rmosolgo: addressed in #1280

rmosolgo:
Ok, I wrote the guides I want to write, the next step is to re-read.

guides/schema/definition.md
```diff
@@ -0,0 +1,205 @@
...
+    # or `nil`
+  end
+end
+```
+
+__`id_from_object`__ is used to implement Relay's `Node.id` field. It should return a unique ID for the given object. This ID will later be sent to `object_from_id` to refetch the object.
+
+```ruby
+class MySchema < GraphQL::Schema
+  def self.object_from_id(object, type, context)
```
ximus: `object_from_id` => `id_from_object`

rmosolgo: oops, thanks 😊

guides/schema/class_based_api.md
```diff
@@ -97,6 +96,36 @@ __A monkeypatch__, activated in global scope:
...
+
+If the type name ends in `Connection`, the field is treated as a connection field.
+
+This default may be overridden by passing a `connection: true` or `connection: false` keyword.
+
+For example:
+
+```ruby
+# This will be treated as a connection, since the type name ends in "Connection"
+field :projects, Types::ProjectType.connection_type
```
mscoutermarsh: Should the type name be `Types::ProjectConnection` here?

rmosolgo: Not quite, `#connection_type` returns an instance of `GraphQL::ObjectType` whose name ends in `"Connection"`, so that's what triggers connection behavior here!

PR #1277, opened 2018-02-13
Fix NO_TYPED_CHILDREN compatibility

rmosolgo:
My wacky object in 1.7.10 was a bit broken, this is a simpler and better solution which is just as fast. 

I used `be rake bench:validate` to confirm that performance is the same.


rmosolgo:
Ug, this was still wrong, it should be `{}.freeze` not `[].freeze`

PR #1283, opened 2018-02-15
Add readers for context and object

rmosolgo:
Previously, the way to customize field resolution in a method involved ivars: 

```ruby 
field :comments, [Types::Comment]

def comments 
  @object.comments.visible_for(@context[:viewer])
end 
```

But this adds reader methods `#context` and `#object`: 

```ruby
def comments 
  object.comments.visible_for(context[:viewer])
end 
```

Cons: 

- Magic method 

Pros: 

- Is a magic method really worse than ivar? I don't think so 
- You get `NoMethodError` in case of a typo, instead of silent nil (eg `@contxte # => nil`)
- You can override the method if you want 

## TODO 

~~update docs (wait for #1274 )~~

### DIFF
--- a/guides/type_definitions/objects.md
+++ b/guides/type_definitions/objects.md
@@ -184,14 +184,14 @@ If you don't want to delegate to the underlying object, you can define a method 
 field :total_games_played, Integer, null: false
 
 def total_games_played
-  @object.games.count
+  object.games.count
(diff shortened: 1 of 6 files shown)


rmosolgo:
`InputObject` could have similar reader methods

PR #1286, opened 2018-02-16
[1.8] interfaces as modules

rmosolgo:
Would it be any good to have GraphQL interfaces defined as Ruby modules? (Continue using `implements(I)`, not `include I`, because `include` is ambiguous) 

Pro: 

- `obj.is_a?(MyInterface)` would work 
- To share field methods, you don't need the `module Implementation ...` API 

Con: 

- There will be some serious inheritance-fu to make the API work, like ActiveSupport::Concern but different (?)


rmosolgo:
Ok, I hacked it together using ActiveSupport::Concern and it kind of works. So, we can chat next week and if it seems worth it, we can pick the parts from AS::C that we want, and take it for a spin.

rmosolgo:
It might be nice, but it's not a clear benefit now, so I'll close it.

PR #1288, opened 2018-02-17
Make Schema#as_json return an instance of Hash

ianks:
Certain JSON encoders (namely, Oj) were previously serializing `GraphQL::Query::Result` incorrectly. Instead of returning a schema definition, it would return serialize the output of `GraphQL::Query::Result#inspect` as a string. In order to have better compatibility, this patch opts to represent the schema as a Hash when calling `#as_json`.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -554,7 +554,7 @@ module GraphQL
     # @param except [<#call(member, ctx)>]
     # @return [Hash] GraphQL result
     def as_json(only: nil, except: nil, context: {})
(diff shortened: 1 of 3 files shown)


rmosolgo:
I reviewed this but forgot to write back, sorry 😖 

This looks good to me, but I think the fail on Rails 3.2 is related to this code change, can you see if there's a tweak that will get that suite passing? Then :+1: from me.

ianks:
@rmosolgo I altered the test to check the logical equivalence of the JSON rather than the exact string representation. This should be less brittle. Let me know if that works!

rmosolgo:
yes, that makes a ton of sense, thanks for taking care of it!

PR #1290, opened 2018-02-19
[Language::Visitor] use method-based visit handlers

rmosolgo:
I'm thinking of stealing a page from the `parser` gem's book. There, the AST visitor is a class with `on_*` handlers for each kind of AST node. It's really easy to get going, I'd like to do the same thing in GraphQL-Ruby. 

So, I tried the API on a couple of validation rules, and I'm going to try a few more. If it seems to work, I'll improve the support for that kind of API and deprecate the `[node_class] <<` API. I think we will need to use `yield` or `super` to support `[node_class].leave <<`. (`parser` uses `super`). Also, I'll be interested to see the performance impact.


arthurschreiber:
This is rad! ❤️I really love this change.

olleolleolle:
This is a helpful change.

rmosolgo:
I pushed a little harder and re-wrote the remaining bits to use the new API, now it's a decent bit faster, 10% give or take. 

```
~/code/graphql-ruby $ be rake bench:validate && be rake bench:validate
validate - introspection
                        32.000  i/100ms
validate - abstract fragments
                        76.000  i/100ms
                        46.000  i/100ms
validate - introspection
                        322.026  (± 3.7%) i/s -      1.632k in   5.075263s
validate - abstract fragments
                        744.823  (± 5.0%) i/s -      3.724k in   5.013195s
                        452.471  (± 6.0%) i/s -      2.254k in   5.000423s
validate - big query     40.256  (± 5.0%) i/s -    204.000  in   5.084613s
validate - introspection
                        32.000  i/100ms
validate - abstract fragments
                        76.000  i/100ms
                        45.000  i/100ms
validate - introspection
                        320.540  (± 4.4%) i/s -      1.600k in   5.001771s
validate - abstract fragments
                        756.113  (± 4.0%) i/s -      3.800k in   5.033471s
                        454.004  (± 5.3%) i/s -      2.295k in   5.069387s
validate - big query     39.463  (± 5.1%) i/s -    200.000  in   5.082648s
```

rmosolgo:
I removed `TypeStack`, but I should put it back, graphql-client depends on it.

rmosolgo:
This is almost passing. We still need one more thing out of the new visitor API, a way to modify the AST that's being iterated over. But I will add that in a later PR, this is a big step forward as-is!

Issue #1294, opened 2018-02-21
[1.8] Complexity on field


esjee:
field.rb has some documentation surrounding the complexity, and how to specify a complexity per field: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/field.rb#L86,L107 

However, it appears that the example doesn't actually work. When attempting to provide complexity either as a keyword argument, or as part of the block, an exception is raised

```
  field :complexField, Integer, complexity: 25
```

```
ArgumentError: unknown keyword: complexity
```

```
  field :complexField, Integer do
        complexity ->(_ctx, _args, _child_complexity) { 25 }
  end
```
```
NoMethodError: undefined method `complexity' for #<GraphQL::Schema::Field:0x00007f3d551849a0>
```

esjee:
I'll try to write a PR to fix this issue. :crossed_fingers:

rmosolgo:
Fixed by #1297

PR #1297, opened 2018-02-21
[1.8] Support complexity field

esjee:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/1294


lib/graphql/schema/field.rb
```diff
@@ -91,6 +92,17 @@ def description(text = nil)
         end
       end
 
+      def complexity(new_complexity_proc)
```
rmosolgo: Integers are also valid for this, right? But I think it will be rejected by this method?

spec/graphql/schema/field_spec.rb
```diff
@@ -36,5 +36,43 @@
...
+          Class.new(Jazz::BaseObject) do
+            graphql_name "complexityKeyword"
+
+            field :complexityTest, String, null: true do
+              complexity ->(one, two) { 52 }
+            end
+          end.to_graphql
+        end
+
+        assert_equal true, !!err.message.match(/^A complexity proc should always accept 3 parameters/)
```
rmosolgo: for nicer error messages, you can also use [assert_match](http://devdocs.io/minitest/minitest/assertions#method-i-assert_match)

lib/graphql/schema/field.rb
esjee: Integers are valid even in the block? Alright, adding support and a test case for that.

rmosolgo:
Awesome, thank you for fixing this!

lib/graphql/schema/field.rb
```diff
@@ -34,7 +34,8 @@ class Field
       # @param field [GraphQL::Field] **deprecated** for compatibility with <1.8.0
       # @param function [GraphQL::Function] **deprecated** for compatibility with <1.8.0
       # @param camelize [Boolean] If true, the field name will be camelized when building the schema
-      def initialize(name, return_type_expr = nil, desc = nil, null: nil, field: nil, function: nil, description: nil, deprecation_reason: nil, method: nil, connection: nil, max_page_size: nil, resolve: nil, introspection: false, hash_key: nil, camelize: true, extras: [], &definition_block)
+      # @param complexity [Numeric] When provided, set the complexity for this field. Default 1.
```
rmosolgo: (for the record, YARD will pick up default values from the source code, no need to document them here in my opinion, for example: http://www.rubydoc.info/gems/graphql/GraphQL%2FObjectType:implements)

esjee: YARD so smart! I didn't realise this, thanks!

Issue #1298, opened 2018-02-21
[1.8] global_field_id does not work for interfaces anymore


esjee:
Pre 1.8 it was acceptable to specify the `global_field_id` in an interface:

```
GlobalFieldIdTestInterface = GraphQL::InterfaceType.define do
  name 'GlobalFieldIdTest'

  global_id_field :uuid
end
```

In 1.8, I would expect the following example to equivalent to the above:

```
class GlobalFieldIdTestInterface < Types::BaseInterface
  global_id_field :uuid
end
```

However, it fails with the exception

```
NoMethodError: undefined method `global_id_field' for GlobalFieldIdTestInterface:Class
```

Inheriting from `Types::BaseObject` does work, but it seems that the `global_field_id` method should _also_ be available in `BaseInterface` (through `GraphQL::Schema::Interface`).

esjee:
I'll try to write a PR for this.

PR #1299, opened 2018-02-21
[1.8] Add support for global_field_id in interface

esjee:
https://github.com/rmosolgo/graphql-ruby/issues/1298


rmosolgo:
Thanks for the fix!

PR #1300, opened 2018-02-21
[1.8] Add camelize: false option

rmosolgo:
This is for backwards compat; if your schema is _all_ underscore-case, then you can override the base classes and reset the default value to `true`.

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -8,17 +8,25 @@ module GraphQL
 
       attr_reader :name
 
-      def initialize(arg_name, type_expr, desc = nil, required:, description: nil, default_value: NO_DEFAULT)
+      # @param arg_name [Symbol]
+      # @param type_expr
+      # @param desc [String]
+      # @param required [Boolean] if true, this argument is non-null; if false, this argument is nullable
+      # @param description [String]
+      # @param default_value [Object]
+      # @param camelize [Boolean] if true, the name will be camelized when building the schema
+      def initialize(arg_name, type_expr, desc = nil, required:, description: nil, default_value: NO_DEFAULT, camelize: true)
         @name = arg_name.to_s
         @type_expr = type_expr
         @description = desc || description
         @null = !required
         @default_value = default_value
+        @camelize = camelize
       end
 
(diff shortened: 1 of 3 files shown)


trevorturk:
Sorry, but would you mind providing an example for how to override the base class(es) and reset the default value? For a Rails app, would that be in an initializer? I'd like to set `null: false, camelize: false` for all `field`s if possible for backwards compat without having to repeat it for every `field`. Thanks!

trevorturk:
The #ruby channel on the GraphQL Slack was very helpful -- they pointed out: http://graphql-ruby.org/type_definitions/extensions.html#customizing-fields and helped me figure out that I want something like this:

```ruby
class CustomField < GraphQL::Schema::Field
  def initialize(*args, **kwargs, &block)
    kwargs.merge! camelize: false
    kwargs.merge! null: false
  end
end

  field_class CustomField
end
```

...which allows me to define fields like so:

```
class ExampleType < BaseObject
  field :count, Integer
end
```

rmosolgo:
Glad you found something that works for you! Thanks for sharing the solution, too. 

If anyone's into it, here's another approach to providing defaults that leverages Ruby's keyword support a bit more: 

```ruby 
class CustomField < GraphQL::Schema::Field
  # Add `null: false` and `camelize: false` which provide default values 
  # in case the caller doesn't pass anything for those arguments. 
  # **kwargs is a catch-all that will get everything else 
  def initialize(*args, null: false, camelize: false, **kwargs, &block)
    # Then, call super _without_ any args, where Ruby will take 
    # _all_ the args originally passed to this method and pass it to the super method.
    super 
  end
end
```

I suppose it's six of one, half dozen of the other. 

Thanks again for sharing!

trevorturk:
Oh nice -- thanks!

marcosdemelo:
Hello, very cool this option.

In the declaration of an ʻargument` I can also configure `camelize: false` for example.

Is it possible to define a `CustomArgument` as well as a` CustomField`? How would I do it?
I tried to follow the documentation for this and it makes `camelize: false` by default, but I couldn't get it to work. If you can help me.

Thank you very much in advance!

PR #1310, opened 2018-02-26
[1.8] Class based mutations

### LINKED ISSUE #1275: New Mutation API Brainstorming
`GraphQL::Relay::Mutation` is really showing its age, and I'd like to replace it with a new class-based API for structuring mutations. I want to answer a few specific questions that have come up over the years. Here are some issues and the current best practice: 

- __How do you authorize a mutation?__ Current best practice is to put auth code at the top of the resolve function. 
- __How do you convert `id`s to objects in a consistent way?__ Current best practice is to load ids as needed. 
- __How do you express errors to clients?__ No current best practice. Maybe top-level `"errors"` key, maybe return `"errors":` as part of the mutation response, maybe a dedicated `ClientError` type. 
 
So, I have some thoughts about how to solve this which I will add below as a comment, and I'd love to hear if anyone else has other issues to consider or suggestions to propose.

rmosolgo:
A new mutation API, see discussion at #1275 

Fixes #1275 

## TODO: 

- [ ] Add `RelayClassicMutation` with `input: { ... }` generation & `clientMutationId`? 
- [ ] Special support for Relay Modern? `clientMutationId`? 
- [x] Add input fetching (ID -> object) 
- [ ] Hook into batching for input fetching
- [ ] Add pre-authorization support (before loading objects)
- [x] Add raised error handling 
- [x] Add client error message support


lib/graphql/query/arguments.rb
```diff
@@ -102,6 +102,18 @@ class << self
...
+      def to_kwargs
+        ruby_kwargs = {}
+
+        keys.each do |key|
+          ruby_kwargs[Schema::Member::BuildType.underscore(key).to_sym] = self[key]
+        end
+
+        ruby_kwargs
+      end
+
```
pstare: I know some people are philosophically opposed to `tap`, but it's really great at building up hashes and then returning them automatically. May I suggest:

```ruby
def to_kwargs
  {}.tap do |kwargs|
    keys.each { |key| kwargs[Schema::Member::BuildType.underscore(key).to_sym] = self[key] }
  end
end
```

pstare: actually in this case perhaps:

```ruby
keys.each_with_object({}) do |key, kwargs|
  kwargs[Schema::Member::BuildType.underscore(key).to_sym] = self[key]
end
```

rmosolgo: Thanks, I'm satisfied with it as is.

rmosolgo:
I'm going to take a more conservative start on this project: https://github.com/rmosolgo/graphql-ruby/pull/1360

PR #1315, opened 2018-02-28
[1.8] Validation is run for wrong field

esjee:
Hey, I found this weird error that I'm not sure how to deal with. It looks like the validation for the `id` field declared in the interface is run for the `_id` in the object. The error message is:

```
TestObject failed to implement some interfaces: "id" is required by TestInterface to return ID! but TestObject.id returns String
```

I'm afraid this might be a bug!


rmosolgo:
👏  Great find, thanks!

rmosolgo:
Oh woops, I should have waited for CI to pass 😬 It looks right to me though, to preserve leading `_`s

PR #1316, opened 2018-02-28
Use urlsafe_encode64 for global ids

mmmries:
Using strict base64 encoding can result in global ids that contain `/` and `+` characters which can cause problems when using global ids as part of a path. For instance if your application looked up a user and then wanted to show an avatar for that user, it might do something like ``. But depending on the user id this might generate a path like ``.

I think it is a safer default to assume that most users will want a [urlsafe base64 encoded id](https://en.wikipedia.org/wiki/Base64#URL_applications). This variant of base64 replaces the characters which are not safe for URLs or filepaths with characters that are. See [Base64.urlsafe_encode64](http://ruby-doc.org/stdlib-2.3.1/libdoc/base64/rdoc/Base64.html#method-i-strict_encode64) for details.


rmosolgo:
I agree with your point that `urlsafe` is a good default. But if we change the default behavior of the gem, what effect will this have on current users? 

For example, if someone updates their `graphql-ruby` dependency, will it cause objects to start generating different IDs?

mmmries:
I think you are right that this is a backwards incompatible changes. If someone had bookmarked a previous page with the `strict_encode64` value,  not all of those same ids will be decodable with `urlsafe_decode64`.

I guess bookmarked links probably wouldn't be a big problem since the only differences between `strict_encode64` and `urlsafe_encode64` are the characters that would cause problems in URLs anyway, so those bookmark links would have already been broken. But if you had a mobile that cached ids those could be broken when it calls back for to the graphQL API.

So this would probably need to be released as a major version update with some release notes on that change and a fallback plan. 😭 

Users are obviously free to implement a urlsafe version in their own projects and maybe that's a good workaround for now? Leaving the default in there might still be surprising to some users down the road, but at least it avoids the backwards incompatibility issue. I tried generating a bunch of base64 data and I was only able to create `/` or other unsafe characters when I was using random binary data. Using ASCII characters and decimal digits  seemed to always generate URL-safe strings from what I could see. (I generated a few million examples)

mmmries:
BTW thank you so much for your work on `graphql-ruby` overall. This has been a really amazing tool and made it easy to see the benefits of adopting graphql because the tool onboarding was so good. 💛 💚 💙 ❤️ 💛 💚 💙 ❤️ 💛 💚 💙 ❤️ 💛 💚 💙 ❤️

rmosolgo:
> free to implement a urlsafe version in their own projects

Yes, I agree this is the best thing, for compatibility considerations. 

But, could we make it easier for people to opt-in to that? For example, we could have a URL-safe encoder in the gem source, and we could document how to add it to your schema. I'd be happy to accept that PR if you're interested. Thanks for looking into this issue!

Issue #1318, opened 2018-02-28
Paging back to "first" page does not allow forward paging afterwards


winterchord:
On master @ ee4c065, with bidirectional pagination turned on (#955).

When querying a connection with `first`, then to the next "page" with `first+after`, and then back to the previous "page" with `last+before` (which is the first page), the `hasNextPage` field is set to  `false`.

For example, if there are 10 empire bases (SW example).

Actual: 

1. bases(first: 1) => hasPreviousPage: false, hasNextPage: true, endCursor: X
1. bases(first: 1, after: X) => hasPreviousPage: true, hasNextPage: true, endCursor: Y
1. bases(last: 1, before: Y) => hasPreviousPage: false, hasNextPage: **false**

Expected: 

1. bases(first: 1) => hasPreviousPage: false, hasNextPage: true, endCursor: X
1. bases(first: 1, after: X) => hasPreviousPage: true, hasNextPage: true, endCursor: Y
1. bases(last: 1, before: Y) => hasPreviousPage: false, hasNextPage: **true**

rmosolgo:
Hi, thanks for the great write-up and this nice patch! It looks good to merge, let me take a closer look tomorrow.

PR #1319, opened 2018-02-28
Paging back to beginning sets hasNextPage to true

winterchord:
When querying a connection with first, then to the next "page" with
first+after, and then back to the previous "page" with last+before
(which is the first page), the `hasNextPage` field was set to false.

The expectation is that it would be set to true (when there are more
pages). Without this change, you could page forward then back, and get
into a situation where `hasPreviousPage` and `hasNextPage` are both
false, when `hasNextPage` should be true.

Fixes https://github.com/rmosolgo/graphql-ruby/issues/1318


rmosolgo:
Oops, sorry I forgot to merge this, thanks for the comment!

PR #1321, opened 2018-03-02
Dont double count field time

rmosolgo:
I'm taking a closer look at this and realized a funny issue: If you add up `execute_field` + `execute_field_lazy`, and compare it to `execute_query` and `execute_query_lazy`, they don't add up! The `_field` events record more time. 

I think it's because the `_field` events wrongly include some child time. So I'm trying to find the source of this inequality and remove it.

The tricky part is backtrace handling, it depends on `_field` events covering some execution internals (that's the failing `NoMethodError` spec, we want that error to be wrapped by GraphQL.)


rmosolgo:
That was nice, `GraphQL::Backtrace` is simpler after all.

PR #1323, opened 2018-03-06
Use .size instead of .count for connection counts

leonardofalk:
When you create a custom connection type using [this](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/relay/connections.md#connection-types) example in Rails it creates a N+1 query, because `.count`is inherited from ActiveRecord and not from Array, whereas `.size` calls the array without refetching the query for each edge.


rmosolgo:
Hi, sorry for the slow response, but thank you for this improvement to our docs! Internally, RelationConnection still uses `.count`. But this is to support Sequel, and to use `.count(:all)`, so I'm not sure it will be easy to change in that case.

PR #1327, opened 2018-03-07
[Language::Printer] Bug with custom object default values

theorygeek:
This fixes a bug when you have an input field, or an argument, that uses some sort of custom object for its default value. I think it was introduced 1.7.8 with #1159.

The bug is that it will raise a `TypeError`, because the [`Language::Printer`](https://github.com/rmosolgo/graphql-ruby/blob/7929ce579b77a5d57b80829d00c131326e66712b/lib/graphql/language/printer.rb#L342) doesn't handle other kinds of objects.

Previously, it looks like `Schema::Printer` would call [`to_s.inspect`](https://github.com/rmosolgo/graphql-ruby/blob/v1.7.7/lib/graphql/schema/printer.rb#L219) on your custom object types.

### DIFF
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -173,7 +173,7 @@ module GraphQL
 
         case type
         when GraphQL::ScalarType
-          default_value
+          type.coerce_isolated_result(default_value)
         when EnumType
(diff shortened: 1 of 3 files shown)


theorygeek:
cc: @xuorig (changed this code recently)

xuorig:
damn good catch @theorygeek, thanks for the additional test coverage ❤️

rmosolgo:
Sorry for the slow review, and thanks for the fix! Looks great.

PR #1303, opened 2018-02-21
Travis.ci run with Ruby 2.5

Willianvdv:

PR #1329, opened 2018-03-08
Unique hash for fields so the union on Ruby 2.5 works as expected

Willianvdv:
A different approach to #1303, this change gives every `GraphQL::Language::Nodes::Field` a hash depending on its properties and uses this hash to compare objects. As `eql?` now uses the `hash`, Ruby's union (| and |=) gives the same result for big and small arrays (this issue https://bugs.ruby-lang.org/issues/14573).

If we go this route, this PR currently only implements the hash for the field node and I'm wondering if we should do the same for all other types since we probably run into the same problems when comparing those objects. 

What do you think @rmosolgo?

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -40,6 +40,10 @@ module GraphQL
             other.children.eql?(self.children)
         end
 
+        def hash
+          line.hash ^ col.hash ^ filename.hash
+        end
+
         # @return [Array<GraphQL::Language::Nodes::AbstractNode>] all nodes in the tree below this one
         def children
           self.class.child_attributes
@@ -211,6 +215,14 @@ module GraphQL
           @arguments = arguments
           @directives = directives
           @selections = selections
+        end
+
+        def eql?(other)
+          hash == other.hash
+        end
+
+        def hash
+          super.hash ^ @name.hash ^ @arguments.hash ^ @alias.hash ^ @directives.hash ^ @selections.hash
         end
       end
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ rvm:
 - 2.1.10 # Lowest version officially supported by the gem is 2.1
 - 2.4.3
 - jruby-9.1.15.0
+- 2.5
 
 services:
   - postgresql


Willianvdv:
The failing build is due to some other (unexpected) error

Willianvdv:
**Master:**
```
➜  graphql-ruby git:(master) bundle exec rake bench:validate
validate - introspection
                        27.000  i/100ms
validate - abstract fragments
                        59.000  i/100ms
                        36.000  i/100ms
validate - big query

     3.000  i/100ms
validate - introspection
                        252.588  (±12.7%) i/s -      1.242k in   5.005712s
validate - abstract fragments
                        533.229  (± 9.6%) i/s -      2.655k in   5.029985s
                        329.961  (±10.3%) i/s -      1.656k in   5.076709s
validate - big query     31.695  (± 9.5%) i/s -    159.000  in   5.068736s
```

**This branch:**
```
➜  graphql-ruby git:(2-5-hash) bundle exec rake bench:validate
validate - introspection
                        26.000  i/100ms
validate - abstract fragments
                        56.000  i/100ms
                        32.000  i/100ms
validate - big query     3.000  i/100ms
validate - introspection
                        264.662  (± 6.4%) i/s -      1.326k in   5.035661s
validate - abstract fragments
                        557.086  (± 7.5%) i/s -      2.800k in   5.057996s
                        321.770  (± 5.6%) i/s -      1.632k in   5.088153s
validate - big query     29.361  (± 6.8%) i/s -    147.000  in   5.020585s
```

rmosolgo:
I'll be interested what they say on the Ruby bug tracker. If they keep it the same in Ruby 2.5, then we'll have to add `hash` one way or another. 

I'm _really_ not sure what to make of this _other_ error: https://travis-ci.org/rmosolgo/graphql-ruby/jobs/350762414#L1041

```
PageInfo::hasNextPage / hasPreviousPage#test_0003_has both if first and last are present:
TypeError: Cannot visit Integer
    /home/travis/build/rmosolgo/graphql-ruby/gemfiles/vendor/bundle/ruby/2.5.0/gems/arel-3.0.3/lib/arel/visitors/visitor.rb:25:in `rescue in visit'
    /home/travis/build/rmosolgo/graphql-ruby/gemfiles/vendor/bundle/ruby/2.5.0/gems/arel-3.0.3/lib/arel/visitors/visitor.rb:18:in `visit'
    /home/travis/build/rmosolgo/graphql-ruby/gemfiles/vendor/bundle/ruby/2.5.0/gems/arel-3.0.3/lib/arel/visitors/to_sql.rb:407:in `visit_Arel_Nodes_Equality'
```

Seems like an incompatibility between Ruby/Rails versions?

Regarding the implementation of `#hash`, I wonder if we can use `#==` as inspiration, something like: 

```ruby
# this node's hash is a function of: 
# - its leaf values 
# - its child nodes' hashes 
def hash 
  scalars.map(&:hash).inject(&:^) ^ children.map(&:hash).inject(&:^)
end 
```

Do you think something like that might work?

Willianvdv:
Let me try to debug the `TypeError: Cannot visit Integer` error first!

lib/graphql/language/nodes.rb
```diff
@@ -212,6 +216,14 @@ def initialize_node(name: nil, arguments: [], directives: [], selections: [], **
           @directives = directives
           @selections = selections
         end
+
+        def eql?(other)
+          hash == other.hash
+        end
```
tenderlove: I don't think this is quite right.  Two objects could hash to the same value, but not be equal (that is a hash collision).

jpserra:
What's the status on this? is it still an existing issue on the latest version of the gem?

rmosolgo:
Yes, although #1352 might be a better approach to the same issue. Personally, I'm hoping to finish my current work on authorization, then start a `1.9-dev` branch with some of the recent PRs regarding `Language::Nodes`, including a solution to this issue.

Willianvdv:
👉 #1352

PR #1332, opened 2018-03-09
Use `super` to collect nodes when accessing `children` or `scalars`

tenderlove:
We can avoid copying information between parent and child classes by
leveraging `super` and inheritance.  This means that each class only
needs to know about it's particular information, and can just ask it's
parent class for any information it has.

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -42,35 +42,27 @@ module GraphQL
 
         # @return [Array<GraphQL::Language::Nodes::AbstractNode>] all nodes in the tree below this one
         def children
-          self.class.child_attributes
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi, thanks for the patch! It seems fine, but can you help me understand the benefits of the new way?

tenderlove:
@rmosolgo this patch removes the "inherited" hook in the module, meaning that we can leverage Ruby's object system to maintain the "parent - child" relationship rather than copying attributes and implementing it ourselves.  You can see from the patch that we do not need to store instance variables on the class or copy parent instance variables to the child class on inheritance; we just let the object system deal with that for us.

tenderlove:
Closing this in favor of #1337 (nice)

PR #1337, opened 2018-03-09
Remove `scalar_attributes`, `child_attributes` etc

tenderlove:
This PR removes `scalar_attributes` and `child_attributes` inherit / copy scheme and replace it with direct definitions and method calls.  This should speed things up as we don't need to do "public_send" anymore, it should (slightly) decrease memory usage as we don't have to hold on to a list of names, and we should create fewer objects on traversal (on average).

I could have made the method definitions one-liners to keep parity with the previous declarative syntax, but I like room to breathe in the code! 😀

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -10,6 +10,14 @@ module GraphQL
       # - `to_query_string` turns an AST node into a GraphQL string
 
       class AbstractNode
+        module Scalars # :nodoc:
+          module Name
+            def scalars
+              super + [name]
+            end
+          end
+        end
+
         attr_accessor :line, :col, :filename
 
         # Initialize a node by extracting its position,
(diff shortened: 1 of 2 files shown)


tenderlove:
I think we could reduce some repetition with other scalars and children as I did in the `Scalars::Name` module, but I'm not sure it's worthwhile.

rmosolgo:
👌  Nice, this is a lot simpler! And since it's such stable code, I think a bit of repetition doesn't hurt.

rmosolgo:
Woops, I thought this was private, but maybe not: 

```
/language/nodes/deep_freeze_ext.rb:13:in `deep_freeze': undefined method `child_attributes' for GraphQL::Language::Nodes::Document:Class (NoMethodError)
```

Taking a look at https://github.com/github/graphql-client/pull/138

nbudin:
Hey @rmosolgo, it looks like this ended up in the 1.7.14 patch release somehow, which broke graphql-client with the error you mentioned.  Is there a possibility to roll this back in the 1.7 series since it's a breaking change?

rmosolgo:
Sorry! I just released graphql-client 0.12.3 which handles this issue, want to try updating to that?

nbudin:
Thanks @rmosolgo, the new version does seem to work.

Mange:
`graphql-client` 0.12.2 -> 0.12.3 prevents our app from booting, so I'd also appreciate that `graphql` 1.17 was kept backwards compatible as I cannot upgrade either at this point.

Issue #455, opened 2016-12-11
Improve AST manipulation options


rmosolgo:
I think there are some good use cases for working with strings from Ruby: 

- Programmatically building or modifying a query, as in graphql-client 
- Merging queries together as in #406 
- Handling subscriptions efficiently (eg, unioning all queries into one, resolving it once, then distrubuting the requested parts of the response to the various subscribers) 

However, this gem provides nothing for doing that. The best API would be one that doesn't reference AST node classes by name, instead, something decoupled from that, in case we have to change the nodes in the future. 

I heard good things about [estools](https://github.com/estools) a while back, there may be some inspiration there!

rmosolgo:
GraphQL-JS supports this really nicely through their visitor API: http://graphql.org/graphql-js/language/#visit

For each kind of node, you may define an `enter()` and `leave()` function, and the return value of that function may affect the AST: 

```js
var editedAST = visit(ast, {
  enter(node, key, parent, path, ancestors) {
    // @return
    //   undefined: no action
    //   false: skip visiting this node
    //   visitor.BREAK: stop visiting altogether
    //   null: delete this node
    //   any value: replace this node with the returned value
  },
  leave(node, key, parent, path, ancestors) {
    // @return
    //   undefined: no action
    //   false: no action
    //   visitor.BREAK: stop visiting altogether
    //   null: delete this node
    //   any value: replace this node with the returned value
  }
});
```

I would love to do something like that, but to support immutable AST nodes, it could be a bit tricky!

rmosolgo:
This will be available in 1.9 with #1740

PR #1338, opened 2018-03-09
Make language nodes immutable

tenderlove:
We have code in our client that "deep freezes" all AST nodes.  I think we could eliminate that if we just made AST nodes immutable by default.  I only found one place where the gem is mutating nodes, but I guess all those "writer" methods are public API, so this PR might only be good for a major version release.

I separated this PR to two commits because I think avoiding mutation internally has merit on its own.


rmosolgo:
:+1: Looks like a nice improvement to me, I want to check tomorrow to make sure it doesn't break graphql-client, but other than that,  I think it's all clear!

rmosolgo:
Somehow I thought I already merged this, but I didn't 😌 

It's going to break graphql-client: 

https://github.com/github/graphql-client/blob/master/lib/graphql/client/query_typename.rb#L21-L34

I like the idea of a more stable AST structure, but I'm going to change the target of this PR to `1.8-dev` and merge it in a bit, when I have some time to update that code as well.

tenderlove:
I'll look in to what it takes to make graphql-client *not* mutate nodes.  We may need a more functional interface for manipulating ASTs.

rmosolgo:
GraphQL-js has a really nice approach for this, I added a comment to the previous issue for improving AST manipulation: https://github.com/rmosolgo/graphql-ruby/issues/455#issuecomment-373472078

rmosolgo:
There are still a few things to do before immutable nodes are _really_ ready: 

- consider freezing the arrays that make up`.children`, too, so it doesn't get changed underfoot (then `.children` could also be frozen)
- add `Node#path` (and maybe `Node#parent`?) to simplify working with the AST, and it can be quite efficient now that they're frozen 
- improve Visitor (see #1290) to support manipulating the AST, so people have a new way of messing with hand-rolled ASTs

But this is a great first step, so merging into the dev branch!

Issue #1292, opened 2018-02-20
Validating enum and input names


KieranP:
Made some changes today, changed some enum and input name to "Enums::User::Gender" and "Inputs::User::Profile". ruby graphql parsed it fine. Then deployed, and apollo started breaking. Turned out "::" is not valid in names. ruby-graphql already validates for unique names, it should validate on name format too, so that issues can be caught early.

rmosolgo:
That would be nice! A PR would be welcome, EnumValue already has a check like this, so maybe the same code could be reapplied.

KarlHarnois:
So we're talking about making a test like pass for both `enum_type` and `input_object_type`?

```ruby
describe "invalid names" do
  # ...

  it "rejects names with colons" do
    assert_raises(GraphQL::InvalidNameError) do
      InvalidEnumTest = GraphQL::EnumType.define do
        name "Some::Invalid::Name"
      end
    end

    InvalidEnumTest.name
  end
end
```

KarlHarnois:
Aka passing the type's name through the `GraphQL::NameValidator`?

KarlHarnois:
Sorry for polluting this issue with my rebases 😮

KieranP:
@KarlHarnois Awesome, thank you :-)

PR #1340, opened 2018-03-11
Fix/validate types names

KarlHarnois:
### What's New?
This is an alternative to https://github.com/rmosolgo/graphql-ruby/pull/1339 in which we move the name validation to the `GraphQL::BaseType` since we hypothesize all type's name should be validated. Curious to know if that's a valid assessment 😛 

**Issue**
#1292


rmosolgo:
Sorry for my late reply! This looks great, thanks for taking care of it. I agree with your assessment that _all_ types need valid names. I hope it will prevent some frustrating confusion in the future 😊

PR #1342, opened 2018-03-11
[1.8] Fix class based API for union and interface type resolution

### LINKED ISSUE #1336: [1.8] GraphQL::Schema::Interface.resolve_type not invoked
`GraphQL::Schema::Interface.resolve_type` is supposed to be invoked during type resolution but this doesn't work in graphql-ruby 1.8. This reproduces using the example from the [documentation](https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/guides/type_definitions/interfaces.md#resolve-type):

```ruby
class Types::RetailItem < Types::BaseInterface
  # This is never invoked!
  def self.resolve_type(object, context)
    if object.is_a?(::Car) || object.is_a?(::Truck)
      Types::Car
    elsif object.is_a?(::Purse)
      Types::Purse
    else
      raise "Unexpected RetailItem: #{object.inspect}"
    end
  end
end

class Types::Car < Types::BaseObject
  implements Types::RetailItem
end

class Types::Purse < Types::BaseObject
  implements Types::RetailItem
end
```

The workaround is to override `GraphQL::Schema::Interface.to_graphql`:
```ruby
class Types::RetailItem < Types::BaseInterface
  def self.to_graphql
    super.tap do |type_definition|
      type_definition.resolve_type_proc = method(:resolve_type)
    end
  end

  def self.resolve_type(object, context)
    if object.is_a?(::Car) || object.is_a?(::Truck)
      Types::Car
    elsif object.is_a?(::Purse)
      Types::Purse
    else
      raise "Unexpected RetailItem: #{object.inspect}"
    end
  end
end
```

jturkel:
This fixes the class based API for [union](https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/guides/type_definitions/unions.md#defining-union-types) and [interface](https://github.com/rmosolgo/graphql-ruby/blob/1.8-dev/guides/type_definitions/interfaces.md#resolve-type) type resolution to behave like the documentation. `GraphQL::Schema::Union` was mistakenly checking for the presence of a `resolve_type` instance method rather than class method. `GraphQL::Schema::Interface` was missing the necessary code altogether.  

Fixes #1336

### DIFF
--- a/lib/graphql/schema/interface.rb
+++ b/lib/graphql/schema/interface.rb
@@ -23,6 +23,9 @@ module GraphQL
             field_defn = field_inst.graphql_definition
             type_defn.fields[field_defn.name] = field_defn
           end
+          if respond_to?(:resolve_type)
+            type_defn.resolve_type = method(:resolve_type)
+          end
           type_defn
         end
       end
--- a/lib/graphql/schema/union.rb
+++ b/lib/graphql/schema/union.rb
@@ -23,9 +23,7 @@ module GraphQL
           type_defn.name = graphql_name
           type_defn.description = description
           type_defn.possible_types = possible_types
-          # If an instance method is defined, use it as a
-          # resolve type hook, via the class method
-          if method_defined?(:resolve_type)
+          if respond_to?(:resolve_type)
             type_defn.resolve_type = method(:resolve_type)
           end
           type_defn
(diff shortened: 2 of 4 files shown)


rmosolgo:
Thanks for digging in on this! Now we've got proper test coverage, too 😊  Sorry it was a bit of a mess before 😛

Issue #1335, opened 2018-03-09
Allow interfaces to register known implementations


jturkel:
When using GraphQL interfaces I often have to register types that implement interfaces with the schema:

```ruby
class MySchema < GraphQL::Schema
  mutation Types::MutationType
  query Types::QueryType

  orphan_types Types::CircleType, Types::SquareType
end
```
This works but it breaks encapsulation since my schema class needs to know about all the classes that implement all of my interfaces. It would be great if I could instead register known interface implementations in the interface definitions e.g.

```ruby
class Types::Shape < GraphQL::Schema::Interface
  implementations Types::CircleType, Types::SquareType
end
```

rmosolgo:
:+1: I agree it would be a nice addition! Implementation-wise, I think you'd have to add the class method to `GraphQL::Schema::Interface`, then modify `GraphQL::Schema::Traversal` to iterate over those types and visit them, so that they're included in the schema. 

Here's the code for visiting interfaces: 

https://github.com/rmosolgo/graphql-ruby/blob/c60abdd52ed4c32bff7724a80e5f9b6f657438ef/lib/graphql/schema/traversal.rb#L73-L74

And it would have to be modified to _also_ visit `implementations`, similar to the way Unions visit their `possible_types`: 

https://github.com/rmosolgo/graphql-ruby/blob/c60abdd52ed4c32bff7724a80e5f9b6f657438ef/lib/graphql/schema/traversal.rb#L75-L79

PR #1346, opened 2018-03-13
[1.8] Add GraphQL::Schema::Interface.orphan_types configuration

jturkel:
As discussed in #1335 this PR implements a method for registering "orphaned" interface implementations which leads to better encapsulation than registering this information with the schema:

```ruby
# Old way (but still supported)
  mutation Types::MutationType
  query Types::QueryType

  orphan_types Types::CircleType, Types::SquareType
end

# New way
class Types::Shape < GraphQL::Schema::Interface
  orphan_types Types::CircleType, Types::SquareType
end
```

I went with the terminology "orphan types" vs. "implementations" to be consistent with the existing orphan types API on `GraphQL::Schema` and to avoid confusion about being required to register all implementations vs. just ones that are unreachable from the schema roots. This does introduce a cyclic dependency between the types and the interfaces they implement but Rails auto-loading or the appropriate require statements should handle this.

The implementation is a mix of code inspired (i.e. copied) from `GraphQL::Schema::Union.possible_types` and `GraphQL::Schema.orphan_types`. For consistency between `GraphQL::Schema::Union.possible_types`, `GraphQL::Schema::Interface.orphan_types` and `GraphQL::Schema.orphan_types` I modified `GraphQL::Schema.orphan_types` to accept either an array or splatted array.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -773,9 +773,9 @@ module GraphQL
         end
       end
 
-      def orphan_types(new_orphan_types = nil)
-        if new_orphan_types
-          @orphan_types = new_orphan_types
+      def orphan_types(*new_orphan_types)
+        if new_orphan_types.any?
+          @orphan_types = new_orphan_types.flatten
         else
           @orphan_types || []
         end
--- a/lib/graphql/schema/interface.rb
+++ b/lib/graphql/schema/interface.rb
@@ -15,10 +15,21 @@ module GraphQL
           end
         end
 
+        def orphan_types(*types)
+          if types.any?
+            @orphan_types = types
+          else
+            all_orphan_types = @orphan_types || []
+            all_orphan_types += super if defined?(super)
+            all_orphan_types.uniq
+          end
+        end
+
         def to_graphql
           type_defn = GraphQL::InterfaceType.new
           type_defn.name = graphql_name
           type_defn.description = description
+          type_defn.orphan_types = orphan_types
           fields.each do |field_name, field_inst|
             field_defn = field_inst.graphql_definition
             type_defn.fields[field_defn.name] = field_defn
--- a/guides/type_definitions/interfaces.md
+++ b/guides/type_definitions/interfaces.md
@@ -149,11 +149,20 @@ Specifically, in the example above, `CarType` would get a field named `price` an
 
 ## Orphan Types
 
-If you add an object type which implements an interface, but that object type doesn't properly appear in your schema, then you need to add that object to the schema's `orphan_types`, for example:
+If you add an object type which implements an interface, but that object type doesn't properly appear in your schema, then you need to add that object to the interfaces's `orphan_types`, for example:
+
+```ruby
+class Types::RetailItem < Types::BaseInterface
+  # ...
+  orphan_types Types::Comment
+end
+```
+
+Alternatively you can add the object types to the schema's `orphan_types`:
 
 ```ruby
 class MySchema < GraphQL::Schema
-  orphan_types [Types::Comment, ...]
+  orphan_types Types::Comment
 end
 ```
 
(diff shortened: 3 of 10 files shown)


lib/graphql/schema/interface.rb
```diff
@@ -15,10 +15,21 @@ def apply_implemented(object_class)
           end
         end
 
+        def orphan_types(*types)
+          if types.any?
+            @orphan_types = types
+          else
+            all_orphan_types = @orphan_types || []
+            all_orphan_types += super if defined?(super)
```
rmosolgo: Woah, I have never seen this before! This looks like a good fit for lots of code around here ...

lib/graphql/schema.rb
```diff
@@ -773,9 +773,9 @@ def max_depth(new_max_depth = nil)
         end
       end
 
-      def orphan_types(new_orphan_types = nil)
-        if new_orphan_types
-          @orphan_types = new_orphan_types
+      def orphan_types(*new_orphan_types)
+        if new_orphan_types.any?
+          @orphan_types = new_orphan_types.flatten
```
rmosolgo: Hmm, can you help me understand how this change fits into the feature?

rmosolgo:
:+1: I think the implementation is good, but I don't think `orphan_types` is the right name. For the schema, `orphan_types` are ones that aren't reachable by traversing fields, starting at query, mutation or subscription roots. They're orphans because they have no parents in the tree that's traversed. 

I think `implementations` would be a better name here, what do you think? Do you mind switching it?

lib/graphql/schema.rb
jturkel: Currently `GraphQL::Schema` requires an array and `GraphQL::Schema::Union` accepts varargs:

```ruby
  orphan_types [Types::CircleType, Types::SquareType]
end

class Types::ShapeUnion < GraphQL::Schema::Union
  possible_types Types::CircleType, Types::SquareType
end
```
This change was done to make `GraphQL::Schema.orphan_types`, `GraphQL::Schema::Interface.orphan_types` and `GraphQL::Schema::Union.possible_types` all accept varargs to make them consistent. I left the `GraphQL::Schema.orphan_types` support for an array argument for backwards compatibility but maybe for consistency we should make all of these APIs accept varargs or an array?

jturkel:
I'm definitely not tied to using the term `orphan_types` but do you think `implementations` would lead to confusion that all interface implementations (vs. ones that aren't reachable from a root) need to be registered?

eapache:
`orphan_implementations`?

jturkel:
`orphan_implementations` works for me. What do you think @rmosolgo? 

Looks like the git history on `1.8-dev` seems to have been rewritten which messed up the diff of this PR :(

rmosolgo:
doh, sorry! I'm prepping a release, so I rebased it on master 😖 

I think the history should clear up if you rebase these changes on `1.8-dev`, for example:

```
# Get the latest 1.8-dev: 

# Add upstream, if you didn't already: 
git remote add upstream git@github.com:rmosolgo/graphql-ruby.git
# Get the latest from rmosolgo/graphql-ruby
git pull upstream 
# Double-check that everything pulled correctly: 
git checkout 1.8-dev 
git log --oneline -n 1
# Should be: 
# a854205f (HEAD -> 1.8-dev-test-release, origin/1.8-dev-test-release, origin/1.8-dev, 1.8-dev) Fix bad rebase

# Now let's rebase your changes on 1.8-dev:

git checkout feature/interface-implementations
# I always do this in case I screw something up, make an extra branch 
git checkout -b feature/interface-implementations-backup 
# Back to the original branch 
git checkout feature/interface-implementations
git rebase 1.8-dev 
# Hopefully it doesn't have any conflicts 
git push -f origin feature/interface-implementations 
```

Does that work?

jturkel:
@rmosolgo - No worries. I rebased the change. When you have sec let me know your thoughts on using the term `orphan_implementations`.

rmosolgo:
Sorry about the bikeshedding, I suppose `orphan_types` is actually fine. If we call it `implementations`, it seems like it's _all_ implementations, which is not necessarily true. Thanks for your work and for your patience 😁

PR #1348, opened 2018-03-15
Change `eql?` to `===`

tenderlove:
If you implement `eql?`, you must also implement `hash`.  Otherwise
hashing will not work correctly as collisions won't resolve, or will
resolve in unexpected ways.

This commit changes `eql?` to `===` so that we have an easy way to do
tree comparison, but we don't mess up hashing semantics (or have to
implement our own hashing semantics).

Also enabling the Ruby 2.5 build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ before_install:
 rvm:
 - 2.1.10 # Lowest version officially supported by the gem is 2.1
 - 2.4.3
+- 2.5
 - jruby-9.1.15.0
 
 services:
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for taking a look at this Ruby 2.5 issue! I confess that I'm not _exactly_ sure about the usage of `#eql?`, it was added way back in #231. But, I think if our tests pass, we can take a quick look at graphql-client and make sure it still works as expected. 

Regarding `#===`, is there a benefit to overriding that method in particular? It has some special uses by Ruby, does our implementation conflict with those, or do we benefit by implementing `#===` in particular?

tenderlove:
I was going with `===` until I was sure that we really wanted to take the perf hit of implementing `.eql?` recursively.  However, I think changing method names at this point is just too much.  I'm going to open another PR with an implementation of `.hash` and enable the 2.5 build.

Issue #1349, opened 2018-03-15
Mutation `input_field` block called before initial setup


cjoudrey:
I'm not sure if this is intentional, but when passing a block to `input_field`, the block will be called before setting up the argument.

This means the `description` of the argument is `nil` within the block and gets set after the block is called.

Repro script:

```ruby
require "graphql"

module Deprecated
  extend self

  def call(schema_member, deprecation_reason)
    p schema_member

    p "Description within deprecated:"
    p schema_member.description

    schema_member.description = [schema_member.description, deprecation_reason].join("\n\n")

    schema_member
  end
end

module GraphQL
  Argument.class_eval do
    accepts_definitions(
      deprecated: Deprecated
    )
  end
end

Mutation = GraphQL::ObjectType.define do
  name "Mutation"

  field :addStar, AddStar.field
end

AddStar = GraphQL::Relay::Mutation.define do
  name "AddStar"
  input_field :id, types.ID, "Starrable ID" do
    deprecated("This will be removed.")
  end
end

Schema = GraphQL::Schema.define do
  mutation Mutation
end

p "Description after setup:"
p Schema.types["Mutation"].fields["addStar"].arguments["input"].type.unwrap.arguments["id"].description

# #<GraphQL::Argument:0x00007ff873adacc8 @prepare_proc=GraphQL::Argument::DefaultPrepare, @name="id", @pending_methods=nil, @pending_definition=nil>
# 
# "Description within deprecated:"
# nil
#
# "Description after setup:"
# "Starrable ID"
```

rmosolgo:
Fixed by #1350

PR #1350, opened 2018-03-15
Fix argument def

xuorig:
Fixes: https://github.com/rmosolgo/graphql-ruby/issues/1349

Changes `Argument.from_dsl` to be a bit more like `GraphQL::Field`. Definition proc takes precedence over kwargs, and kwargs are defined before the proc is called, safer for extensions.

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -58,8 +58,13 @@ module GraphQL
     end
 
     def default_value=(new_default_value)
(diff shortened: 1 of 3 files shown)


rmosolgo:
:+1: Looks great, that way the arguments will apply in a much more expected order.

rmosolgo:
🙇 thanks for hunting this down!

PR #1352, opened 2018-03-15
Implement `.hash` on Nodes, enable Ruby 2.5

tenderlove:
AST nodes were implementing `.eql?` without implementing `.hash`.
`.eql?` is supposed to return true if an object is exactly the same, or
if its contents are exactly the same.  This is used to resolve hash
collisions.  Any object whose `.hash` is the same, the hash
implementation will call `.eql?` on them to determine if those to
objects are *exactly* the same (if so, then one is discarded as it's a
dupe).  In this case, we were implementing `.eql?` without implementing
`.hash`, so nodes that are semantically equivalent would return true for
`.eql?` but would not produce the same `.hash` value.  Two objects that
return true for `eql?` must also hash to the same value, so this patch
implements the `.hash` method.

Implementing the `.hash` method causes the test to fail the same way on
Ruby 2.4 as it does on Ruby 2.5, so I updated the test to fix the
assertion.

### DIFF
--- a/spec/graphql/internal_representation/rewrite_spec.rb
+++ b/spec/graphql/internal_representation/rewrite_spec.rb
@@ -124,7 +124,7 @@ describe GraphQL::InternalRepresentation::Rewrite do
 
       nut_selections = plant_selection.typed_children[schema.types["Nut"]]
       # `... on Tree`, `... on Nut`, and `NutFields`, but not `... on Fruit { ... on Tree }`
-      assert_equal 3, nut_selections["leafType"].ast_nodes.size
+      assert_equal 1, nut_selections["leafType"].ast_nodes.size
       # Multi-level merging when including fragments:
       habitats_selections = nut_selections["habitats"].typed_children[schema.types["Habitat"]]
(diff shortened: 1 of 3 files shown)


spec/graphql/internal_representation/rewrite_spec.rb
```diff
@@ -124,7 +124,7 @@
 
       nut_selections = plant_selection.typed_children[schema.types["Nut"]]
       # `... on Tree`, `... on Nut`, and `NutFields`, but not `... on Fruit { ... on Tree }`
-      assert_equal 3, nut_selections["leafType"].ast_nodes.size
+      assert_equal 1, nut_selections["leafType"].ast_nodes.size
```
rmosolgo: On the one hand, this isn't the _intended_ result here, it's meant to keep track of all the AST nodes that were merged into this part of the query. 

On the other hand, I don't know of a real use case for it, and 1 node will probably be enough.  Hmm.

rmosolgo:
I'm going to address this here instead: #2328

PR #1354, opened 2018-03-15
Add schema.json route to Rails generator

marcotc:
>  _WIP: Looking for feedback to gauge if it's worth continuing this effort._

We've been using the result of `GraphQL::Introspection::INTROSPECTION_QUERY` to easily retrieve the complete schema.
We normally serve it from `GET /graphql/schema.json`, which allows you to fetch it from our web-browser if needed, without fiddling with `POST` requests, or having to create the non-trivial full introspection query.

There's more work to do here:
- [ ] Make this feature optional
- [ ] Add tests

If this seems like something valuable to add to the generator, I'll polish it up.

### DIFF
--- a/lib/generators/graphql/install_generator.rb
+++ b/lib/generators/graphql/install_generator.rb
@@ -100,6 +100,7 @@ RUBY
 
         template("graphql_controller.erb", "app/controllers/graphql_controller.rb")
         route('post "/graphql", to: "graphql#execute"')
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi, please feel free to reopen this (or a new PR) if you decide to go forward with it! Personally I don't use an endpoint like this, but I'd be glad to support it as an option in the generator.

PR #1356, opened 2018-03-16
Eval argument block when provided

### LINKED ISSUE #1347: [1.8] Schema::Field.argument(...) silently ignores blocks
`GraphQL::Schema::Field.argument` method appears to silently ignore blocks passed to it.

Repro code:

```ruby
require "graphql"

  field :field, String, null: false do
    p 'Field block'
    argument :arg, String, required: false do
      p 'Argument block'
    end
  end
end

  query Query
end

Schema.types
```

We should either not accept a block or accept and call it.

I would vote for the latter so that the API is consistent with `field(...) do ... end`.

cjoudrey:
Fixes #1347

`GraphQL::Schema::Field.argument` method appears to silently ignore blocks passed to it.

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -18,7 +18,7 @@ module GraphQL
       # @param description [String]
       # @param default_value [Object]
       # @param camelize [Boolean] if true, the name will be camelized when building the schema
-      def initialize(arg_name, type_expr, desc = nil, required:, description: nil, default_value: NO_DEFAULT, camelize: true, owner:)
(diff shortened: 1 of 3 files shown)


rmosolgo:
huh, I kind of forgot this is Ruby's _default_ behavior, to silently ignore unused blocks 

```ruby 
$ irb
irb(main):001:0> "abc".upcase { |x| x.whatever }
=> "ABC"
irb(main):002:0> String.instance_method(:upcase).parameters
=> [[:rest]]
irb(main):003:0> {a: 1}[:a] { stuff }
=> 1
```

rmosolgo:
Thanks for the fix!

PR #1358, opened 2018-03-20
[1.8] Allow fields to use anonymous classes for types

jturkel:
I hit this bug while using some meta-programmed classes for field types. The test case shows a contrived example of this.


rmosolgo:
What a bummer typo! Thanks for hunting it down. I'll merge it as soon as the build passes, for the sake of decorum :P

PR #1360, opened 2018-03-20
[1.8] Simple mutation classes

rmosolgo:
I'm trying to take some advice from @cjoudrey here: let's build out the _minimum_ required class-based mutations first, and start iterating on it at GitHub, then build out more features. 

There's a chance we'll set off in the wrong direction, but the base class is simple enough that I think we'll be able to recover 😆


spec/support/jazz.rb
```diff
@@ -362,11 +362,29 @@ class EnsembleInput < GraphQL::Schema::InputObject
...
+      { instrument: instrument }
+    end
+  end
+
   class Mutation < BaseObject
     field :add_ensemble, Ensemble, null: false do
       argument :input, EnsembleInput, required: true
     end
 
+    # TODO don't require `field` somehow
```
swalkinshaw: We have a `MutationObjectType` class which handles this.

So you'd have `class Mutation < MutationObject`

rmosolgo: Interesting, do you have an equivalent to `field :doStuff, field: DoStuff.field`  ?

swalkinshaw: `MutationObjectType` defines a `mutation` method:

```ruby
mutation :doStuff, DoStuff
```

However I think this confuses people who don't realize that a mutation is just a field. So I do prefer keeping `field`.

rmosolgo: > confuses people 

I was worried about the same thing, glad you feel that way too. Something like 

```
field :do_stuff, DoStuff
```

could work fine, but it doesn't give the reader a _good_ clue how `DoStuff` fits in the picture. But maybe not a big deal.

swalkinshaw: There's a middle ground of `field :doStuff, mutation_field: DoStuff`

I think the main thing is just not having to manually call `DoStuff.field`

rmosolgo:
Ok, my todos here are done and the code in GitHub is looking pretty good, so my plan is to write up some docs and then merge.

PR #1367, opened 2018-03-23
Track AST node in build from definition

cjoudrey:
When building a schema from a GraphQL schema definition, it can be useful to know where a type originated from within the schema definition. This information can be used by tools like linters to report where the violation was in the original definition file.

This PR adds an `ast_node` attribute accessor to all types and sets them when `Schema::BuildFromDefinition` is used.

Unfortunately not everything can hold an `ast_node`, for instance if a type implements an interface (i.e. `type A implements Node`) there is no spot where I can set the `ast_node` of `Node`. That being said, from the `ast_node` of the type definition `A`, it's possible to find the `ast_node` of the interface name `Node`.

The same problem happens for `@directives`, but can be solved the same.

**Example use:**

```ruby
require "graphql"

schema_definition = <<-GRAPHQL
schema {
  query: Query
}

type Query {
  field: String
}
GRAPHQL

schema = GraphQL::Schema.from_definition(schema_definition)

puts "Schema defined at: #{schema.ast_node.position}"
puts "Query.field defined at: #{schema.types["Query"].fields["field"].ast_node.position}"

# Output:
#
# Schema defined at: [1, 1]
# Query.field defined at: [6, 3]
```

When implementing #1304 I accidentally added parser tests in `parser_spec.rb` instead of `query_parser_specification.rb`. I'm not sure if this split is still used, but I've moved them over just in case. There's a few more tests that I think could be moved, but wanted to wait for your confirmation.


lib/graphql/language/parser.y
```diff
@@ -153,7 +153,7 @@ rule
     | schema_keyword
 
   enum_value_definition:
-    enum_name directives_list_opt { return make_node(:EnumValueDefinition, name: val[0], directives: val[1], description: get_description(val[0])) }
+    enum_name directives_list_opt { return make_node(:EnumValueDefinition, name: val[0], directives: val[1], description: get_description(val[0]), position_source: val[0]) }
```
rmosolgo: wow, oops, they didn't even have `line` and `col` 😬

rmosolgo:
`query_parser_specification` was added so that graphql-libgraphqlparser could reuse the tests, but graphql-libgraphqlparser was never updated since libgraphqlparser got schema parsing (https://github.com/graphql/libgraphqlparser/pull/45) 😬 

Anyhow, LGTM 😎 👏

rmosolgo:
it will be a bit tricky getting this on 1.8-pre, I've been rebasing 1.8-dev on master from time to time, I'm hoping to do another pre version this weekend, we'll see.

Issue #1368, opened 2018-03-24
Subscription Action Cable Implementation Example leads to JavaScript client Uncaught TypeError: Cannot read property 'data' of null


mlwilkerson:
The guide for [Action Cable Implementation](http://graphql-ruby.org/subscriptions/action_cable_implementation.html) of Subscriptions refers to the API doc for example code to implement `GraphqlChannel < ApplicationCable::Channel`.

Unfortunately, that example code seems to have an error (maybe code that was correct for a previous version?) that leads to this JavaScript client side error when using `ActionCableLink`:
`import ActionCableLink from 'graphql-ruby-client/subscriptions/ActionCableLink'`

Here's the error on the client:
```
ActionCableLink.js:30 Uncaught TypeError: Cannot read property 'data' of null
```

This happens when initially setting up the subscription. The server generates a `nil` result with `more: true`. And [it sets up the payload like this](https://github.com/rmosolgo/graphql-ruby/blob/3f1cb573ad42c912f41e62c92ee5bf6f34e3bd93/lib/graphql/subscriptions/action_cable_subscriptions.rb#L42):

```ruby
    payload = {
        result: result.subscription? ? nil : result.to_h,
        more: result.subscription?
    }
```

But `ActionCableLink` chokes when it tries to look at `payload.result.data` in `received`:
```javascript
        received: function(payload) {
          if (payload.result.data) {
            observer.next(payload.result)
          }

          if (!payload.more) {
            this.unsubscribe()
observer.complete()
          }
        }
```

This is because `payload.result` is `null`, because it was set to `null` by the server (above).

I couldn't seem to isolate the problem until downloading the [demo app offered here](https://github.com/SenteraLLC/graphql-subscriptions-spike) and comparing to a working example.

The key is that the [demo app from SenteraLLC wraps](https://github.com/SenteraLLC/graphql-subscriptions-spike/blob/344d953233e480674886f6585730c544380c39c7/graphql-subscriptions-rails/app/channels/graphql_channel.rb#L30) the `nil` result in `{data: nil}` like this:
```ruby
    payload = {
        result: result.subscription? ? {data: nil} : result.to_h,
        more: result.subscription?
    }
```

PR #1369, opened 2018-03-24
Fix example code to return correct nil data in result.

mlwilkerson:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/1368


rmosolgo:
👏  sorry about the bug, thanks for tracking it down! 🤔  Hmm, it looks like the system tests even use a slightly _different_ setup:

https://github.com/rmosolgo/graphql-ruby/blob/fe0656fe54d4c6d8f456c87986b4c21bd39b13b8/spec/dummy/app/channels/graphql_channel.rb#L67-L70

PR #1371, opened 2018-03-25
Fixing typo in error handling

cianmce:
`RescueFrom` is used in the sample class definition but `Rescuable` is used in the sample use case.
Updated class definition to use `Rescuable`

https://github.com/rmosolgo/graphql-ruby/blob/df28b7da1451931f7dbafc9bd17395d775f11367/guides/queries/error_handling.md#error-handling-with-wrappers


rmosolgo:
Oops, thanks for the fix!

PR #1372, opened 2018-03-26
[1.8] Try again: interfaces as Ruby modules

rmosolgo:


rmosolgo:
I got a great question on Slack so I wrote the response here for posterity: 

> rmosolgo [2:44 PM]
> Man, the big difficulty with interfaces
> i had the same thing in the gem, now in the app
> is inheritance of class methods
> with classes, it just works
> but with modules
> not good
> so the `url_field` macro above no worky

> theorygeek [2:45 PM]
> what about ActiveSupport::Concern’s `class_methods` pattern?

Regarding AS::C, I took a crack at it, even porting most of it to `GraphQL::Schema::Member::Concern`. 

But the problem is, AS::C treats modules in one of two different ways: 

- Either, the module is _another_ concern, in which case, class methods from its parent concerns are _not_ mixed in; 
- Or, the module is _not_ a concern, so class methods are mixed in, but then AS::C's inheritance magic stops there. 

Here's the source: https://github.com/rails/rails/blob/master/activesupport/lib/active_support/concern.rb#L114

The problem here is that we need a mix. For example, this doesn't work: 

```ruby
module BaseInterface 
  field_class MyCustomField 
  
  # A helper for defining fields:
  def self.field_macro(...); end 
end

module Commentable 
  include BaseInterface 

  field_macro(...)
end 
```

It doesn't work because `Commentable` doesn't get `self.field_macro` from `BaseInterface`. 
Since BaseInterface wasn't a concern, all the class methods were dumped there, but 
modules that include it don't get any special help. 

So, what if you _made_ `BaseInterface` into a Concern? 

```ruby
module BaseInterface 
  extend GraphQL::Schema::Member::Concern
  field_class MyCustomField 
  
  # A helper for defining fields:
  def self.field_macro(...); end 
end

module Commentable 
  include BaseInterface 

  field_macro(...)
end 
```

Then you get `undefined method field_class`, because, since `BaseInterface` was a concern, it didn't get class methods from its ancestors. 

So, how can you get _both_ macros from upstream and continue sending them downstream? I really fiddled with this a lot, tweaking various `included` hooks
and `ClassMethods` modules, but I couldn't find a solution.

theorygeek:
One little nit on the syntax, is `extend` a better choice?
```ruby
module Commentable 
  extend GraphQL::Schema::Interface
```
(On an `ActiveSupport::Concern`, I think you use `include` to add functionality to your _class_ and `extend` to add functionality to your _module_. I might be wrong though.)

rmosolgo:
Another good question was: how about getting rid of the `module Implementation` altogether? Something like: 

```ruby
class MyInterface < GraphQL::Interface
  field :foo
  def foo; end
end
```

The thing I couldn't figure out here was, how to get a `self` that made sense for the body of that method. I _think_ best would be if it was included into the class just like a module's method, but I couldn't find any way to do that. There are a lot of ways to grab methods, but you can't really put them on other classes: 

```$ irb
irb(main):001:0> class C1; def m1; end; end
=> :m1
irb(main):002:0> class C2; end
=> nil
irb(main):003:0> method = C1.instance_method(:m1)
=> #<UnboundMethod: C1#m1>
irb(main):004:0> method.bind(C2)
        3: from /Users/rmosolgo/.rbenv/versions/2.5.0/bin/irb:11:in `<main>'
        2: from (irb):4
        1: from (irb):4:in `bind'
TypeError (bind argument must be an instance of C1)
irb(main):005:0> method.bind(C2.new)
        3: from /Users/rmosolgo/.rbenv/versions/2.5.0/bin/irb:11:in `<main>'
        2: from (irb):5
        1: from (irb):5:in `bind'
TypeError (bind argument must be an instance of C1)
```

So, I'm not sure how to execute that method in the context of the runtime object. 

There's another option which I have not explored: perhaps interface fields could be called on an instance of the interface. That is, whenever an interface is requested, you'd initialize an instance of the Interface type, and call methods on it. 

I'm afraid it would be pretty weird, because, for example, ivars would not be shared between methods, the way they are between ruby modules. Also, it would get tricky if objects override their interface fields in some way. For example, one or more of: re-call `field`, redefine the method. 😬

lib/graphql/schema/member/concern.rb
```diff
@@ -0,0 +1,38 @@
...
+            @_dependencies.each { |dep| base.include(dep) }
+            super
+            base.extend const_get(:ClassMethods) if const_defined?(:ClassMethods)
+            base.class_eval(&@_included_block) if instance_variable_defined?(:@_included_block)
+          end
+        end
+
+        def included(base = nil, &block)
+          if base.nil?
+            raise "Multiple included { ... } blocks, merge them" if instance_variable_defined?(:@_included_block)
```
theorygeek: Does it break anything if I make an interface module, and do something like:
```ruby
module SomeInterface
  # what if I flip these two around?
  extend ActiveSupport::Concern
end
```

rmosolgo:
Regarding `extend` vs `include`, what if you want to inherit field implementations, for example: 

```ruby 
module AbstractNameable 
  extend_or_include GraphQL::Schema::Interface 

  def name 
     @object.display_name 
  end 
end 

module Actor 
  extend_or_include AbstractNameable
end 

class Person < GraphQL::Schema::ObjectType
  implements AbstractNameable 
end 
```

Somehow, we want `Person#name` to go to `AbstractNameable#name`, right? I think we'll get that for free if we use `include`.

lib/graphql/schema/member.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 require 'graphql/schema/member/accepts_definition'
+require 'graphql/schema/member/base_dsl_methods'
+require 'graphql/schema/member/cached_graphql_definition'
+require 'graphql/schema/member/graphql_type_names'
```
rmosolgo: All of these were extracted so I could keep `Schema::Member` as a base class, but easily get these methods into `Schema::Interface`.

lib/graphql/schema/member/accepts_definition.rb
```diff
@@ -37,55 +37,85 @@ class Member
...
+            child.class_eval do
+              class << self
+                prepend(ToGraphQLExtension)
+              end
             end
           end
         end
 
-        module ClassMethods
+        module AcceptsDefinitionClassMethods
```
rmosolgo: Here was a weird naming conflict with `ClassMethods`, case in point for renaming `Schema::Interface`'s inherited module to `DefinitionMethods`

lib/graphql/schema/interface.rb
```diff
@@ -1,25 +1,36 @@
...
-          # so it can contain methods to implement fields.
-          # It's always added to interfaces, but sometimes it's left empty.
-          child_cls.const_set(:Implementation, Module.new)
-        end
+        # Here's the tricky part. Make sure behavior keeps making its way down the inheritance chain.
+        def included(child_class)
+          if !child_class.is_a?(Class)
+            # In this case, it's been included into another interface.
+            # This is how interface inheritance is implemented
+            child_class.const_set(:ClassMethods, Module.new)
```
rmosolgo: TODO: this should only set it if it's not already defined

spec/graphql/schema/interface_spec.rb
```diff
@@ -39,8 +35,8 @@ def new_method
       assert_equal 2, new_object_2.fields.size
       # It got the new method
       assert new_object_2.method_defined?(:new_method)
-      # But not the old method
-      refute new_object_2.method_defined?(:id)
+      # And the inherited method
+      assert new_object_2.method_defined?(:id)
```
rmosolgo: (This was just a document-how-it-actually-works test, I think the new behavior is better anyway)

theorygeek:
It is looking pretty nice! I think I was hoping that you would do:
```ruby
module MyInterface
  extend GraphQL::Schema::Interface
end

class MyObjectType < GraphQL::Schema::Object
  include MyInterface
end
```

Because then you get these semantics, which seem nice:
```ruby
# Is this type an Interface type?
MyInterface.is_a?(GraphQL::Schema::Interface)
# => true

# Is this type an Interface type?
MyObjectType.is_a?(GraphQL::Schema::Interface)
# => false

# Test for interface implementation
MyObjectType < MyInterface
# => true

# Test if the object instance supports that interface
MyObjectType.new.is_a?(MyInterface)
# => true
```

However... it is probably overrated, and perhaps even a bit confusing. Because right now, you would test if a type is an _object_ type using:

```ruby
MyObjectType < GraphQL::Schema::Object
# => true
```

rmosolgo:
Yeah, I can't think of a way to make it _consistent_! I think I'm going to merge this branch soon, if you want to experiment with the different hooks, I'm open to switching it around a bit, but I'm not optimistic I can make it drastically better than it is now.

theorygeek:
I wrote up a little gist that had a real basic example of how it might work:
https://gist.github.com/theorygeek/9f4f68980f0c0dd89b116ad8ac0e7acc

I am not sure that they are huge improvements. I think the nicest thing is that it solves:
- `MyObjectType < GraphQL::Schema::Object` is true
- `MyObjectType < GraphQL::Schema::Interface` is falsey

rmosolgo:
Thanks for sharing those notes, I took a try at it here: 

https://github.com/rmosolgo/graphql-ruby/compare/interface-as-module...interface-as-module-extends

I'm not quite sure it's worth it because: 

- inheriting methods doesn't Just Work anymore. We could `include` them modules during the `extended` hooks, but that defeats the purpose, right? 
- Although `MyObj < GraphQL::Schema::Object` works, `MyInterface < GraphQL::Schema::Object` doesn't work (in either case, afaik) , so i don't think it's really worth investing in `<` to check object type. Maybe we can add a different class method for checking that.

rmosolgo:
also, I added `definition_methods do ... end` in 56c6a14, thanks for the tip!

PR #1378, opened 2018-03-28
Fix get field introspection

xuorig:
`schema.get_field("Query", "__schema")` would fail because we compared a string vs the query root object.


rmosolgo:
@xuorig is this change still in the works?

rmosolgo:
Feel free to reopen or comment here if this is still needed!

PR #1393, opened 2018-04-05
Improve RescueMiddleware error matching

### LINKED ISSUE #1392: Schema level error handling - Improve rescue_from logic
I believe that `rescue_from` should take a more ruby `rescue`like approach when matching errors. 

It's common for projects to define their own error classes which can quickly grow in number, requiring us to update the schema every time there's a new error to handle. 

Imagine the following scenario: 
```ruby
module Errors
  class BaseError < StandardError; end
end

module A
  class B < Errors::BaseError; end
  class C < Errors::BaseError; end
  class D < Errors::BaseError; end
end

module X
 class Z < Errors::BaseError; end
end
```

If I want to have graphql rescuing all these different errors I have to have multiple entries on the rescue_table 1 for each different error as opposed to ruby where I could just `rescue Errors::BaseError` and handle all of them at once. 
 
What do you think about such approach? IMO it's simpler and more intuitive if we follow the ruby way of handling exceptions.

jpserra:
Support error matching by superclasses in rescue_table.

RescueMiddleware, when attempting to rescue the errors was only matching
against the exact class if it was present in the `rescue_table`. This PR
introduces the change where we can `rescue_from` a superclass and
RescueMiddleware will be able to handle all its descendants by checking
if either the error class provided or any of it's superclasses is in the
`rescue_table`.

Closes #1392

### DIFF
--- a/lib/graphql/schema/rescue_middleware.rb
+++ b/lib/graphql/schema/rescue_middleware.rb
@@ -41,13 +41,14 @@ module GraphQL
       private
 
       def attempt_rescue(err)
-        handler = rescue_table[err.class]
-        if handler
-          message = handler.call(err)
-          GraphQL::ExecutionError.new(message)
-        else
-          raise(err)
-        end
+        rescue_table.each { |klass, handler|
+          if klass.is_a?(Class) && err.is_a?(klass) && handler
+            message = handler.call(err)
+            return GraphQL::ExecutionError.new(message)
+          end
+        }
+
+        raise(err)
       end
     end
   end
--- a/spec/graphql/schema/rescue_middleware_spec.rb
+++ b/spec/graphql/schema/rescue_middleware_spec.rb
@@ -25,6 +25,17 @@ describe GraphQL::Schema::RescueMiddleware do
       assert_equal(GraphQL::ExecutionError, result.class)
     end
 
+    describe "rescue_from superclass" do
+      class ChildSpecExampleError < SpecExampleError; end
+
+      let(:error_class) { ChildSpecExampleError }
+      it "handles them as execution errors" do
+        result = middleware_chain.invoke([])
+        assert_equal("there was an example error: ChildSpecExampleError", result.message)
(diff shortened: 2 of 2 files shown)


lib/graphql/schema/rescue_middleware.rb
```diff
@@ -41,13 +41,14 @@ def call(*args)
...
       def attempt_rescue(err)
-        handler = rescue_table[err.class]
-        if handler
-          message = handler.call(err)
-          GraphQL::ExecutionError.new(message)
-        else
-          raise(err)
-        end
+        rescue_table.each { |klass, handler|
+          if klass.is_a?(Class) && err.is_a?(klass) && handler
```
rmosolgo: I understand one of the conditions: `err.is_a?(klass)`, but what about `klass.is_a?(Class)` and checking for `handler`? Are those conditions required, or can we remove them? 

Or, if these checks are necessary, do you know if we have tests for that case?

jpserra: I opted to go for the safest scenario possible: 

since there is no validation when populating the `rescue_table` `klass.is_a?(Class)` ensures that we only run `err.is_a?(klass)` if in fact `klass` is a `Class`, otherwise we'd get a `TypeError` .

the `handler` presence check was for the similar reasons, nothing forbids us from calling `rescue_from(Error)` without the block, which would lead to `handler=nil` right?

jpserra: let me know if this makes sense to you or if perhaps we should take a more conservative approach and `raise` an error when `error_classes` provided to `rescue_from` are not a `class` in https://github.com/rmosolgo/graphql-ruby/pull/1393/files/651a12bdae4db9355eefbff09eafe0bd12890309#diff-b8d49293896b9144b1b73bd7906d7513R22

rmosolgo:
Thanks again for this improvement!

jpserra:
my pleasure!

PR #1394, opened 2018-04-05
Interpreter

rmosolgo:
Problem: 

- GraphQL execution takes a lot of time and memory by taking fragments on abstract types and `.dup`ing them for each possible concrete type. This is a serious problem for `Node` queries written by Relay, see #1172 . In fact, in many cases, this is exponential time complexity, bad!
- GraphQL execution can't be _completely_ prepared, since `@skip` is hardcoded in `Rewrite`, even static queries must go through that step. 
- GraphQL execution is not hackable. There are odd hooks here and there, but they're not really hackable.

Solution: 

- Stop "concrete-izing" fragments, instead, use a pure interpreter to execute queries (remove the `Rewrite` step) 
- Remove the rewrite step altogether (TODO: we probably need _some_ kind of caching during query execution, eg, for coerced arguments) 
- Call execution methods on the type classes themselves, so that they can be overridden by users (🙀 😎 )

Problems with the solution: 

- Query analyzers go out the window 
- `irep_node`s go out the window 
- So, some advanced tooling goes out the window.

Possible solutions to problems with the solution: 

- Replace query analysis with before-execution pass of the interpreter, allow analyzers to register methods which are called along the query tree, much like an interpretation pass 

Possible problems with possible solutions to problems with the solution: 

- It would probably support the same kinds of tools, but it would require a lot of work. 

#### TODO 

- [x] Lazy values 
- [x] Authorization 
- [x] GraphQL::ExecutionError 
- [x] SKIP 
- [x] Tracing 
- [x] Nil propagation
- ~~rewrite with cps instead of blocks?~~ reconsider this as a perf consideration and customization option, later
  - I think passing a block _counts_ as passing a continuation. It's callable, it's bound to the local environment, and as a bonus, it's really efficient since it's  built-in to Ruby and we don't have to capture them into Procs.
  - One problem with blocks is the long call traces. Could these be cleaned up? same for the `.with_*` methods.
- [x] Find a more elegant solution than `:__temp_running_interpreter`? (Something that doesn't clutter user's `ctx.to_h`)
- [x] Add standard handling for `LateBoundType`
- [x] Make sure that new-style definitions are passed to introspection (TypeType, FieldType, SchemaType, ArgumentType)
- ~~Profile~~ can be done after initial merge
- ~~Custom directives~~ can be done after initial merge
- ~~rescue_from~~ skipped the tests for this, it would be good to replace it with something.
- [x] Multiplex, specifically batching across multiplex context
- [x] Test tooling to run _both_ new and old executes with Dummy::Schema (and Jazz::Schema ?)
- [x] Make every test schema run the interpreter in CI
- [x] Clean integration and docs for interpreter
- [x] Find all `# TODO`s and address them
- [x] Rename and document `resolve_field_2`
- [x] Double-check all switches on TESTING_INTERPRETER in the test suite


rmosolgo:
I force-pushed this branch with a new interpreter based on an AST visitor.

xuorig:
This is really really promising 🙏, what if analyzers just became standard AST visitors, leaving to the client to navigate fragments / other things the rewrite would've previously done 🤔?

rmosolgo:
>  analyzers just became standard AST visitors 

💰  yep, i think that's great.

rmosolgo:
> navigate fragments

I'm guessing, that, when we port QueryDepth and QueryComplexity, we'll develop some helpers for this which we can make public in some way. 

The other thing is, it might make sense to cache some stuff like which fragments go where, so maybe that will help too.

xuorig:
I wonder if we can extract some of the very useful logic found in the interpreter's `Trace` and `Visitor`, namely the `on_fragment_spread` and `on_inline_fragment` logic, as well as the `Trace.with_object` helpers to provide member definitions to an AST visitor.

This will certainly come in useful for Analyzers, as well as a few custom use cases (that maybe can now be rewritten as analyzers?)

lib/graphql/schema/object.rb
```diff
@@ -56,6 +68,10 @@ def initialize(object, context)
         @context = context
       end
 
+      def __typename
+        self.class.graphql_name
+      end
```
rmosolgo: TODO I think this can be removed

spec/graphql/query_spec.rb
```diff
@@ -260,7 +260,8 @@ module ExtensionsInstrumenter
         def self.before_query(q); end;
 
         def self.after_query(q)
-          q.result["extensions"] = { "a" => 1 }
+          # This call is causing an infinite loop
+          # q.result["extensions"] = { "a" => 1 }
```
rmosolgo: TODO fix this

spec/graphql/schema/mutation_spec.rb
```diff
@@ -23,7 +23,8 @@
 
   describe "argument prepare" do
     it "calls methods on the mutation, uses `as:`" do
-      query_str = 'mutation { prepareInput(input: 4) }'
+      skip "I think I will not implement this"
+      query_str = "mutation { prepareInput(input: 4) }"
```
rmosolgo: TODO should I implement this?

lib/graphql/execution/interpreter.rb
```diff
@@ -0,0 +1,54 @@
...
+  module Execution
+    class Interpreter
+      def run_query(query)
+        @query = query
+        @schema = query.schema
+        evaluate
+      end
+
+      def self.use(schema_defn)
+        # TODO encapsulate this in `use` ?
```
rmosolgo: more like TO _DONE_

emilebosch:
Any benchmarks on the performance increase/decreased memory usage?

rmosolgo:
See https://github.com/rmosolgo/graphql-ruby/pull/1886

Issue #891, opened 2017-08-13
RFC: Missing InputValue node in parser


cjoudrey:
Currently, `InputValue` get turned into Ruby literals when parsed.

For example if we parse this GraphQL query:

```graphql
{
  field(argument: "value")
}
```

The `ArgumentNode` is:

```
#<GraphQL::Language::Nodes::Argument:0x007f8f8430bef0
  @line=3,
  @col=9,
  @filename=nil,
  @name="argument",
  @value="value">
```

The problem with this is we lose the `line` and `col` for the `InputValue` node `"value"` which is useful for validation rules.

In the reference implementation, they solve this by having an `InputValue` node and thus the `ArgumentNode` is:

```js
export type ArgumentNode = {
  kind: 'Argument';
  loc?: Location;
  name: NameNode;
  value: ValueNode;
};
```

I'd be willing to help add `ValueNode` support to the parser, but if we want to do it cleanly, it will be a breaking change for anyone relying on `GraphQL::Language::Nodes` as `Argument#value` (for example) would become a `ValueNode` instead of a Ruby literal.

Alternatively, we could leave the current `value` as a Ruby literal and add `value_node` which would be the `ValueNode`. While this is backwards compatible, I feel it might not be consistent with the rest of the parser.

@rmosolgo thoughts?

rmosolgo:
It would be great to get position info for literal values! I think your suggestion of adding `#value_node` is great. Like you said, it's not consistent, but being non-breaking is a huge win. 

Alternatively, if you _wanted_ to explore a similar change in graphql-libgraphqlparser-ruby, we could merge them together and bump their gemspecs as needed.

cjoudrey:
> Alternatively, if you _wanted_ to explore a similar change in graphql-libgraphqlparser-ruby, we could merge them together and bump their gemspecs as needed.

Just to confirm what you mean, you're asking if I could do the same change in `graphql-libgraphqlparser-ruby` so that we can ship both at the same time, correct?

It's not one or the other since, we'd still want this in the Ruby implementation.

rmosolgo:
Sorry for being vague -- what I meant was: 

- Adding value_node is :+1: 
- If you want to make the breaking change on `InputValue#value` to return a `ValueNode` instead of a `String | Integer | Float | Boolean | nil`, it's ok with me _if_ we make the corresponding change to graphql-libgraphqlparser-ruby 

[...]

PR #1398, opened 2018-04-06
Support custom JSON scalar types

### LINKED ISSUE #146: json scalar type
Is it possible to add a json scalar type  to avoid having to send raw json as a string? That way in a JS web environment it won't be necessary to JSON.parse the string data.

jturkel:
This PR adds support for building a custom JSON scalar type (i.e. actual JSON not JSON encoded in a string) that is similar to [graphql-type-json](https://github.com/taion/graphql-type-json) for [graphql.js](https://github.com/graphql/graphql-js). The custom scalar can be defined as:

```ruby
class JsonType < GraphQL::Schema::Scalar
  def self.coerce_input(value, _context)
    value
  end

  def self.coerce_result(value, _context)
    value
  end
end  
```
The parser was converting JSON arguments to a `GraphQL::Language::Nodes::InputObject` so I had to convert it back to it's more raw form before attempting scalar coercion. The changes to `GraphQL:: StaticValidation::LiteralValidator` could definitely use some extra scrutiny to ensure I didn't break anything :)

Fixes #146

### DIFF
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -117,6 +117,8 @@ module GraphQL
         value.to_h
       elsif value.is_a?(Array)
         value.map { |element| raw_coercion_input(element) }
+      elsif value.is_a?(GraphQL::Language::Nodes::Enum)
+        value.name
       else
         value
       end
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -109,7 +109,17 @@ module GraphQL
     end
 
     def coerce_non_null_input(value, ctx)
-      @coerce_input_proc.call(value, ctx)
+      @coerce_input_proc.call(raw_coercion_input(value), ctx)
+    end
+
+    def raw_coercion_input(value)
+      if value.is_a?(GraphQL::Language::Nodes::InputObject)
+        value.to_h
+      elsif value.is_a?(Array)
+        value.map { |element| raw_coercion_input(element) }
+      else
+        value
+      end
     end
 
     def validate_non_null_input(value, ctx)
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -16,7 +16,7 @@ module GraphQL
         elsif type.kind.list?
           item_type = type.of_type
           ensure_array(ast_value).all? { |val| validate(val, item_type) }
-        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::AbstractNode) && !ast_value.is_a?(Array)
+        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
           type.valid_input?(ast_value, @context)
         elsif type.kind.enum? && ast_value.is_a?(GraphQL::Language::Nodes::Enum)
           type.valid_input?(ast_value.name, @context)
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -16,23 +16,35 @@ module GraphQL
         elsif type.kind.list?
           item_type = type.of_type
           ensure_array(ast_value).all? { |val| validate(val, item_type) }
-        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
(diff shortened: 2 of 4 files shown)


lib/graphql/static_validation/literal_validator.rb
```diff
@@ -16,7 +16,7 @@ def validate(ast_value, type)
         elsif type.kind.list?
           item_type = type.of_type
           ensure_array(ast_value).all? { |val| validate(val, item_type) }
-        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::AbstractNode) && !ast_value.is_a?(Array)
+        elsif type.kind.scalar? && !ast_value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
```
jturkel: Looking at the grammar again this will mistakenly accept an enum as a scalar literal. I think I need to rework this to ensure that ast_value is recursively only composed of `GraphQL::Language::Nodes::InputObject`s and Ruby primitives.

jturkel:
It looks like the grammar in the [spec](http://facebook.github.io/graphql/October2016/#Value) does support enums and variable references nested in JSON scalars but graphql.js doesn't support variable references nested in JSON scalars so I punted on that. This should be good for review now.

lib/graphql/scalar_type.rb
```diff
@@ -109,7 +109,19 @@ def ensure_two_arg(callable, method_name)
     end
 
     def coerce_non_null_input(value, ctx)
-      @coerce_input_proc.call(value, ctx)
+      @coerce_input_proc.call(raw_coercion_input(value), ctx)
+    end
+
+    def raw_coercion_input(value)
+      if value.is_a?(GraphQL::Language::Nodes::InputObject)
```
jturkel: I'd love to avoid type switches like this by adding a `GraphQL::Language::Nodes::AbstractNode#scalar_conversion_input` method but how you feel about adding methods like this to the AST nodes.

lib/graphql/static_validation/literal_validator.rb
```diff
@@ -16,23 +16,35 @@ def validate(ast_value, type)
...
         end
       end
 
-
       private
 
+      # The GraphQL grammar supports variables embedded within scalars but graphql.js
+      # doesn't support it so we won't either for simplicity
+      def constant_scalar?(ast_value)
+        if ast_value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
```
jturkel: This is another one where I'd love to avoid the type switch by either adding a `GraphQL::Language::Nodes::AbstractNode#constant?` or perhaps better yet leveraging `Visitor` to check the subtree for the presence of a `GraphQL::Language::Nodes::VariableIdentifier` node. Let me know what you think.

rmosolgo:
Wow, this looks great! I'll take a close look soon but certainly no issues jump out at me. I'm happily surprised that it didn't require a parser overhaul 😅

jturkel:
I did a quick spike adding methods to the AST nodes but it didn't work very well since string, int, boolean, float and array "nodes" just use the corresponding Ruby classes. I'm good with the current implementation if you are :)

rmosolgo:
😖  Oh, that thing about proper AST nodes 😖  there's an open issue for it, in case you want to keep an eye out: https://github.com/rmosolgo/graphql-ruby/issues/891

Thanks again for adding this!

jturkel:
Awesome. I'll definitely keep an eye on #891. Thanks for reviewing this!

sarink:
How does one use this? I've got the custom scalar implemented, but `graphiql` keeps yelling at me `Unexpected value of type JsonType`

shaneardell:
@sarink I was struggling with the same thing, then noticed that the changes in the `spec/support/jazz.rb` file are for an example app and demonstrate how to use it. Looks like mostly this adds support for it, but you still have to define your custom json type.

I added a file in my directory where I define custom types (for me it was `/app/graph/types/raw_json.rb' and used the following from the example:

```
class RawJson < GraphQL::Schema::Scalar
    def self.coerce_input(val, ctx)
      val
    end
     def self.coerce_result(val, ctx)
      val
    end
  end
```

Then I was able to use `RawJson` as a field type when defining other types and it worked beautifully.

ZeroSoulEater:
For everyone struggling an addition to @shaneardell.
I had to add `Types::` to make this work propper 
so 

```
class Types::JsonType < GraphQL::Schema::Scalar
  def self.coerce_input(val, ctx)
    val
  end

  def self.coerce_result(val, ctx)
    val
  end
end
```

PR #1404, opened 2018-04-09
Correctly check for `null` argument when `mutation` argument used

mengledowl:
When using the `mutation` argument to set a mutation field and passing in `null: false`, it should blow up with an `ArgumentError` but doesn't because it's simply checking for presence (which evaluates to `false` on `false`).

This changes the code to check for a `null` argument that is anything other than `nil` instead when deciding to blow up.


rmosolgo:
🤦‍♂️  Oops! Good find, sorry about that, and thanks for the fix!

PR #1406, opened 2018-04-09
Add ability to set nullability of mutations

mengledowl:
When using the class-based API for mutations (inheriting off of `GraphQL::Schema::Mutation`), there is no way to set whether or not the mutation field is nullable. This change adds a `null` helper method that subclasses can use to specify whether or not they should be nullable:

```
  null true
end
```

This will be inherited by any subclasses:

```
class AnotherMutation < MyMutation
end

AnotherMutation.null # => true
```

If it hasn't been set, it will raise an error when you attempt to use the mutation.

`GraphQL::Schema::RelayClassicMutation` sets `null` to `true`, so no need to worry when inheriting from it:

```
class MyRelayMutation < GraphQL::Schema::RelayClassicMutation
end

MyRelayMutation.null # => true
```

Any thoughts/critiques on this from @rmosolgo or any of the other more regular maintainers more than welcome!

### DIFF
--- a/guides/mutations/mutation_classes.md
+++ b/guides/mutations/mutation_classes.md
@@ -30,6 +30,8 @@ GraphQL-Ruby includes two classes to help you write mutations:
 
 Besides those, you can also use the plain {% internal_link "field API", "/type_definitions/objects#fields" %} to write mutation fields.
 
+An additional `null` helper method is provided on classes inheriting from `GraphQL::Schema::Mutation` to allow setting the nullability of the mutation. This is required when inheriting off of `GraphQL::Schema::Mutation` - `GraphQL::Schema::RelayClassicMutation` sets `null` to `true` for you.
+
 ## Example mutation class
 
 You should add a base class to your application, for example:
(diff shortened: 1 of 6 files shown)


rmosolgo:
👏  I dig it! I think this is a great suggestion. The API fits right in, too. 

I think there's one way the implementation could be improved. Let's say you want to establish some conventions in your app, so you create a base class: 

```ruby 
class BaseMutation < GraphQL::Schema::Mutation 
  null true 
  # other conventions & helper methods here 
end 
```

Then extend it: 

```ruby 
class AddThing < BaseMutation 
  # mutation behavior here
end 
```

Will `AddThing` inherit `null true`? I think it would be _nice to_, but I didn't see a test for it. 

Also, regarding the default value, do you mind making it default to `null true`, instead of raising an error? I think that will help people get going without a hurdle and it's a good way to start out, since adding `null false` later is _not_ a breaking schema change.

mengledowl:
I updated the code such that `null` defaults to `true` now.

I had originally thought that the inheritance that you suggested would work, but when I took a second look, I realized that I forgot to put it in there actually 😅 So now that's been added as well, along with tests to prove it!

rmosolgo:
Looks great, thanks for taking care of this feature!

Issue #1359, opened 2018-03-20
Rethink Pagination


rmosolgo:
Currently you have a few options to paginate with GraphQL-Ruby: 

- Don't paginate at all, return list types everywhere (:-1:)
- Roll your own pagination system (:+1:, but tricky, and we could probably help each other out here)
- `GraphQL::Relay::Connection` & friends (😬 ) 

In my opinion, `GraphQL::Relay::Connection` has some problems: 

- Important parts of the system are hidden from view (namely, the handling and application of `first`/`after`/`last`/`before`)
- The automagic object-to-wrapper mapping based on class hides pagination, so: 
   - It's easy to add a regression where something that _should_ be an AR::Relation starts being an Array, but GraphQL-Ruby doesn't miss a beat. If you want to paginate something as a Relation, it would be nice to have a stricter mapping, to avoid accidentally loading the whole relation into memory. 
  - Sometimes, a class-to-wrapper mapping isn't enough: 
    - Rails 3 does weird things, I think AR::Relation _is_ an Array, or something 
    - If you want to apply a specialization in some cases (when the returned object is of the same class), you don't have a good option. You have to introduce a proxy class, and a custom wrapper, etc. 
- The system is not easy to customize, customization requires some code duplication or un-Ruby-ish factory functions. 

So, I would like to reconsider our pagination system with a few goals: 

- Make the application object -> pagination wrapper mapping _explicit_, in the resolver 
- Don't hide any arguments from the resolver; pass everything in there. 
- Use a class-based system for extending and customizing pagination, for improved hackability. (Combined with an explicit object -> wrapper mapping, this could support more case-by-case customization.)

I would like to preserve some things from the current system: 

- Backwards compat, for sure 
- Cursor-based, Relay-style pagination (although I'd be open to adding some other options for non-Relay users?) 

So, does anyone else have thoughts about how pagination could be improved in GraphQL-Ruby?

eapache:
So we've completely rolled our own pagination internally, and I'm not sure how well the patterns will apply to the gem generally but I'll share the shape for context at least.

Internally, we have these things called `Filter`s which are little classes which can be appended to fields. They may define arguments, and they get an opportunity to manipulate the result of resolving the field before it's actually returned. The canonical example internally is `FirstFilter` whose implementation (in its entirety) is:
```ruby
  class FirstFilter < Filter
    argument :first, :integer, description: "Truncate the array result to this size"

    def apply_to(field)
[...]

PR #1411, opened 2018-04-10
Field Filters

rmosolgo:
Inspired by hints about Shopify's structure in [this comment](https://github.com/rmosolgo/graphql-ruby/issues/1359#issuecomment-374954336). 

Make a new way to modify field resolution. This is roughly equivalent to instrumentation and middleware, but a bit simpler. 

- Fields accept a `filters: [...]` option 
- Classes in the list will be initialized with `new(field:)` 
- Filters' `resolve_field` method will be called like and around_filter on the field, they must yield to continue flow.

The goal is the same one described in the comment above, a pagination system with more fine-grained control over which-fields-get-what. 

Additionally, this could do a better job with `client_mutation_id` and, I think, rescuing from errors.

Any thoughts? 

I won't ship this til a stable release of 1.8 hits rubygems 

__TODO__

- Maintain compatibility with old connection API, don't remove that one test I removed 
- Add a simple opt-in for pagination, port the old connections to the new way 
- Implement a better error rescuer as a proof of concept
- Abstract away `SKIP` and `lazy?` from user code, don't make people check then in filter `field_resolve` methods


lib/graphql/schema/filter.rb
```diff
@@ -0,0 +1,77 @@
...
+        inner_args = args.dup
+        inner_args.delete(:first)
+        inner_args.delete(:after)
+        inner_args.delete(:last)
+        inner_args.delete(:before)
+        nodes = yield(obj, inner_args)
+        after_lazy(obj.context.schema, nodes) do |value|
+          if value.nil?
+            nil
+          elsif value.is_a?(GraphQL::Execution::Execute::Skip)
```
eapache: It would be great if the filter base class could handle these cases somehow (and the lazy resolution case). In our filters we ended up with two methods a la before/after instead of a single `yield`able resolve, so that the base class could do this logic in the middle. That said I do really prefer the `yield` pattern because it's easy to understand and more in line with how middleware works everywhere else...

rmosolgo:
I'm coming back to this from time to time, but having a hard time finding an implementation which: 

- Cleanly abstracts the pre-Promise and post-Promise steps of field resolution; AND 
- Doesn't add significant runtime overhead 

For example, if the Promise-related API allocates new objects, and is applied by default, then it will add a lot of overhead to scalar fields, even when they _don't_ return Promises. 

I _need_ something like this, but I don't have a good solution yet! 😖

lib/graphql/schema/filter.rb
rmosolgo: I'm needing this same thing in another branch, and trying out `Schema#after_lazy`: 

https://github.com/rmosolgo/graphql-ruby/blob/9ae8e86e5ff15de740e213288f42162b61014ba8/lib/graphql/schema.rb#L968-L981

It's a bit odd to have it as a method on `schema`, but it's because each schema maintains a map of `class => method` pairs for resolution. 

There's a crazy (🙈 ) easier-to-use solution which would involve adding `.graphql_lazy?`, `.graphql_then(&:block)` and `.graphql_sync` to `Object`, and then letting people re-implement those methods on should-be lazy objects (eg, `Promise`). So you could always use `obj.graphql_then { |resolved_obj| ... }` , and for most objects, it would just be `yield_self`. 

I think I'll try it on another branch sometime. 

(Funny, it's technically a feature where refinements could work, but I think we support some Rubies that never implemented `refine`.)

swalkinshaw:
@rmosolgo does your work on supporting lazy resolve in authorization unblock this at all? Or is the pre/post Promise steps still an issue?

dylanahsmith:
It sounds like the problem is that you want to do something like Promise#then on promises and Object#yield_self on regular objects without the overhead of allocating objects.  The [promise.rb](https://rubygems.org/gems/promise.rb) gem already has `Promise.sync` and `Promise.map_value` for cases where you want to deal with objects that may or may not be promises.  `Promise.map_value` only allocates an object if the passed in object is a `Promise`.  So I don't think object allocation should be necessary to solve this problem, you could just use conditionals and methods not defined on the object itself.

This wouldn't be as performant as monkey patching, but it is worth considering this alternative, since the performance improvements of monkey patching over conditionals might be negligible outside of micro benchmarks.

> It's such a bad API, but the thing is, you need schema to be involved because it has the registry of `lazy_class => method_name` for resolving those things.

So essentially you are trying to replace that registry with the `graphql_then` and `graphql_sync` monkey patches?  So if you still think monkey patching Object is the way forward, then I think this monkey patch of `::Promise` should be done in the [graphql-batch](https://rubygems.org/gems/graphql-batch) gem rather than the promise.rb gem, since currently the graphql-batch is doing this mapping.  `::Promise` shouldn't have to know anything about graphql, so I don't plan on adding any monkey patches to that gem.

I agree that depending on these monkey patches internally to the graphql gem would be a breaking change, so it should its major version should be bumped when making this change. That major version bump will force the graphql-batch gem to be upgraded as well, which would avoid confusion on such an upgrade.

PR #1430, opened 2018-04-19
Support setting the new relic transaction name

rmosolgo:
This adds an option to new relic tracer to set the transaction name for all queries.

### DIFF
--- a/lib/graphql/tracing/new_relic_tracing.rb
+++ b/lib/graphql/tracing/new_relic_tracing.rb
@@ -14,7 +14,29 @@ module GraphQL
         "execute_query_lazy" => "GraphQL/execute",
       }
 
+      # @param set_transaction_name [Boolean] If true, the GraphQL operation name will be used as the transaction name.
+      #   This is not advised if you run more than one query per HTTP request, for example, with `graphql-client` or multiplexing.
+      def initialize(set_transaction_name: false)
+
+        @set_transaction_name = true
+      end
+
       def platform_trace(platform_key, key, data)
+        if @set_transaction_name && platform_key == "execute_query"
+          query = data[:query]
+          # Set the transaction name based on the operation type and name
+          selected_op = query.selected_operation
+          if selected_op
+            op_type = selected_op.operation_type
+            op_name = selected_op.name || "anonymous"
+          else
+            op_type = "query"
+            op_name = "anonymous"
+          end
+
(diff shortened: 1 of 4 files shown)


lib/graphql/tracing/new_relic_tracing.rb
```diff
@@ -14,7 +14,29 @@ class NewRelicTracing < PlatformTracing
         "execute_query_lazy" => "GraphQL/execute",
       }
 
+      # @param set_transaction_name [Boolean] If true, the GraphQL operation name will be used as the transaction name.
+      #   This is not advised if you run more than one query per HTTP request, for example, with `graphql-client` or multiplexing.
+      def initialize(set_transaction_name: false)
+
+        @set_transaction_name = true
```
akshah123: looks like this might always set it to `true`.

rmosolgo: 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️

lib/graphql/tracing/new_relic_tracing.rb
```diff
@@ -14,7 +14,29 @@ class NewRelicTracing < PlatformTracing
...
+      end
+
       def platform_trace(platform_key, key, data)
+        if @set_transaction_name && platform_key == "execute_query"
```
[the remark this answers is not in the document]
rmosolgo: yeah :S I've been putting off proper unit tests for these platform tracers, but two fatal errors in a row means I can't afford to skip them anymore 😆  thanks!

rmosolgo:
Thanks again for the reviews, I've pushed some proper tests and fixed a few issues, and added a simple per-query override. @obrie , thanks for sharing your advanced use case -- I'd like to publish these small improvements as-is, and consider that more sophisticated implementation later, if more folks express interest. It sounds like nice work!

PR #1437, opened 2018-04-20
Support `super` in field methods

rmosolgo:
There's a bit of hand-waving here so that in an object class, you can call `super` to get the default field resolution behavior. 

When you extend `GraphQL::Schema::Object`, the inheritance chain would usually be: 

```
MyObject < GraphQL::Schema::Object < Object < ... 
```

So, if `MyObject` has a method like: 

```ruby 
field :username, String, null: false 

def username 
  super 
end 
```

Ruby will error out, because the parent class, `GraphQL::Schema::Object`, has no method called `username`. 

The solution, which I think is also used by ActiveRecord, is to insert a secret module where the default methods are added: 

```
MyObject < default_resolve_module < GraphQL::Schema::Object
```

And whenever a field is added, also, add a method to that secret module. 

With that technique, `super` is supported for objects, interfaces and mutations. 

As a caveat, any class which overrides `inherited` hooks must call `super` (as the added `super`s in this diff show).

### DIFF
--- a/spec/support/jazz.rb
+++ b/spec/support/jazz.rb
@@ -302,6 +302,11 @@ module Jazz
 
   class HashKeyTest < BaseObject
     field :falsey, Boolean, null: false
+
+    def falsey
+      # test that super can be used in objects
+      super
+    end
   end
 
   # Another new-style definition, with method overrides
--- a/spec/support/jazz.rb
+++ b/spec/support/jazz.rb
@@ -65,15 +65,13 @@ module Jazz
       super(*args, **options, &block)
     end
 
-    def to_graphql
-      field_defn = super
-      if @upcase
-        inner_resolve = field_defn.resolve_proc
-        field_defn.resolve = ->(obj, args, ctx) {
-          inner_resolve.call(obj, args, ctx).upcase
-        }
+    def resolve_field(*)
+      result = super
+      if @upcase && result
+        result.upcase
+      else
+        result
       end
-      field_defn
     end
   end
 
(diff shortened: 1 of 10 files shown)


theorygeek:
🎉 💰 🚀  This is really nice! I think it'll make meta-programming a lot easier!

spec/support/jazz.rb
```diff
@@ -304,6 +312,11 @@ def self.resolve_type(object, context)
 
   class HashKeyTest < BaseObject
     field :falsey, Boolean, null: false
+
+    def falsey
+      # test that super can be used in objects
+      super
+    end
```
theorygeek: I would suggest a couple test cases, mostly as sanity checks in case someone alters this code:

1. If I set up two chains of interfaces (A, B, and D are interfaces):
- `C (class object type) < B < A < GraphQL::Schema::Interface`
- `E (class object type) < D < A < GraphQL::Schema::Interface`
- `F (class object type) < GraphQL::Schema::Object`

Then I should be able to:
- Have a field on `A` that is callable from both `C` and `E`, but not `F`
- Have a field on `B` that is callable from `C` but not `E` or `F (interfaces don't cross-pollute)
- Have a field on `C` that is callable from `C`, but not `E` or `F` (objects don't cross-pollute)

2. Same thing, but make A, B, and D superclasses instead of interfaces.

theorygeek: I think that would make sure that all of the book-keeping is set up, and also future-proof it in case someone refactors it.

rmosolgo: Thanks for taking a look and making this great suggestion, I made some proper tests in 7674a590. I think the specific inheritance chains were a bit different, but I think I covered all the bases!

PR #1443, opened 2018-04-24
Update input_objects.md

pradyumna2905:
- Adding the real output of `attributes.full_text` to make it even more clear.


rmosolgo:
Thanks for the improvement!

Issue #1445, opened 2018-04-24
[1.8] Upgrader generated invalid syntax when arguments are passed to `#field` method in brackets


michal-samluk:
**Input**
```ruby
field(:my_field, types.String) do
  resolve(->(obj, _args, _ctx) { obj.my_field })
end
```
**Output**
```ruby
field :my_field, Stringdo, resolve(->(obj, _args, _ctx) { obj.my_field }) end, null: true
```

But when I remove the brackets: 
```ruby
field :my_field, types.String do
  resolve(->(obj, _args, _ctx) { obj.my_field })
end
```
the output will be correct: 
```ruby
field :my_field, String, null: true do
  resolve(->(obj, _args, _ctx) { obj.my_field })
end
```

michal-samluk:
One more thing, removing brackets around lambda also gives different result. 

**input**
```ruby
field :my_field, types.String do
  resolve ->(obj, _args, _ctx) { obj.my_field }
end
```
**output**
```ruby
field :my_field, String, null: true
def my_field
  @object.my_field
end
```

rmosolgo:
Thanks for reporting this! The last thing you posted is the expected result, I guess the upgrader should be improved to work with those parens! 

In fact, there's already a transform to remove the first set of parens, but it doesn't seem to work: 

https://github.com/rmosolgo/graphql-ruby/blob/d256dd26bc780a14884e5e910bf9e8656350b6db/lib/graphql/upgrader/member.rb#L176-L184

What do you think about adding a field like that to one of the [fixture files](https://github.com/rmosolgo/graphql-ruby/tree/master/spec/fixtures/upgrader) and pushing a failing test? Then I'd be happy to dig in and find a way to support it properly.

michal-samluk:
@rmosolgo 
```ruby
2.4.2 :007 > input_text.sub(/(field|input_field|return_field|connection|argument)\( *(.*?) *\) */, '\1 \2')
 => "field 'dfasfdas', types.Stringdo" 
2.4.2 :008 > input_text.sub(/(field|input_field|return_field|connection|argument)\( *(.*?) *\) */, '\1 \2 ')
 => "field 'dfasfdas', types.String do" 
```
I think we can improve this sub with adding one space ;) 

```
'\1 \2'
'\1 \2 '
```

michal-samluk:
I will submit fixture for this and with lambda. The above one will fix parentheses around field, but not around resolve.

PR #1448, opened 2018-04-24
Parentheses are breaking upgrader

michal-samluk:
Related to the issue: https://github.com/rmosolgo/graphql-ruby/issues/1445


rmosolgo:
Thanks for adding those fixtures! I think I've got it working, I'll merge when the tests pass.

PR #1452, opened 2018-04-25
Add Mongo relay connection support

benkimball:
This PR aims to support `Mongoid::Criteria` as a relation underlying a `RelationConnection`. I found that almost all of the existing `RelationConnection` code could be reused, with just a few tweaks to preserve its behavior when using a `Mongoid::Criteria`. As a result I subclassed `RelationConnection` and modified the following methods:

- `relation_offset` and `relation_limit` use [`Mongoid::Criteria::Queryable::Options`](https://docs.mongodb.com/mongoid/7.0/api/Mongoid/Criteria/Queryable/Options.html) to retrieve offset and limit option values

- `relation_count` has to perform the query in order to count the results; MongoDB's `count` function otherwise ignores `limit`

- `limit_nodes` treats "limit 0" as meaning "do not apply any limit", rather than "return no results". In order to reuse the existing `RelationConnection` code, I wanted this method to mimic the ActiveRecord/Sequel behavior. Therefore it checks to see if the limit is 0, and if so returns [the None criteria](https://docs.mongodb.com/mongoid/7.0/api/Mongoid/Criteria.html#none-instance_method) to ensure that the result set is empty and the database is not queried. The way I did this -- using `Mongoid::Criteria#without_options` to clone the criteria and discard any cached query results, then `#none` to get the None criteria -- does not precisely match ActiveRecord, in that if the user later calls `limit` on this relation with some positive limit, it will remain a None criteria.

Finally, I created a suite of tests for the connection by mimicking the Star Wars suite, but for a slightly different sci-fi franchise.


benkimball:
Hm, that CI failure makes me think I need to tell Appraisal about mongoid.

benkimball:
Working through a few test failures from running `appraisal rake` now.

benkimball:
Feeling kind of uneasy about that Mongoid 3.1.7 `where(id: nil)` hack.

rmosolgo:
Wow, thanks for all your work on the setup for this! That was always a scary part for me. 

Regarding the latest CI fail: 

> No server is available matching preference: #<Mongo::ServerSelector::Primary:0x37687120 tag_sets=[] max_staleness=nil> using server_selection_timeout=30 and local_threshold=0.015 (Mongo::Error::NoServerAvailable)

It sounds like it didn't find any Mongo database at all, right? Maybe we need to update our `.travis.yml` to include mongo in the build: 

https://docs.travis-ci.com/user/database-setup/#MongoDB

Want to give it a shot?

benkimball:
I'm kinda stuck on the CI failures :( tests pass locally, and I'm not sure how to resolve the `refute defined?(Rails)` test failure on Travis.

rmosolgo:
Sorry about the CI failures, I forgot that the no-rails gemfile is a bit tricky. To simplify development, `rails` is in `Gemfile`. (Otherwise you'd always have to pick a target in `gemfiles/`.) But then it gets copied into the other gemfiles, including the no-rails one. So I've always just removed it by hand, I don't think anyone else messed with it since then 🙈

lib/graphql/relay/mongo_relation_connection.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+module GraphQL
+  module Relay
+    # A connection implementation to expose MongoDB collection objects.
+    # It works for:
+    # - `Mongoid::Criteria`
+    class MongoRelationConnection < RelationConnection
```
rmosolgo: :+1: I'm happily surprised how much this is able to reuse from RelationConnection! TBH it's just _not a mess_ ... and I was always afraid it would be a mess, but it's nice!

rmosolgo:
Looks good to me: clear implementation, well tested! Thanks for putting the time into this 😊

PR #1454, opened 2018-04-26
Take field.definition_directives and put it in the SDL

rmosolgo:
This is not really ready for prime time, but it gives an example of how to get field data into the SDL. 

You could do this in your own app by: 

- extending `GraphQL::Language::DocumentFromSchemaDefinition` 
- overriding `build_field_node` 
- in your new `build_field_node`, use the `GraphQL::Field` instance to generate AST nodes 

Then, use the custom `DocumentFromSchemaDefinition` class to create a string: 

```ruby 
class CustomSchemaDocument < GraphQL::Language::DocumentFromSchemaDefinition 
  # ...
end 

# Use the class to build a string: 
CustomSchemaDocument.new(MySchema).document.to_query_string
```

It would be nice to put this feature in the library, but this PR is probably not how to do it. Instead, we should make Directive classes with some class methods to facilitate this.


rmosolgo:
This is not good enough to merge, after 1.8 I'll look at a new Directive class that supports a flow like this properly.

Issue #1345, opened 2018-03-13
[1.8] Defaulting of Field#null and Argument#required


jturkel:
The new class based API in 1.8 requires the `null` option to be specified for fields and the `required` option to be specified for arguments:

```ruby
module Types
  class User < ::GraphQL::Schema::Object
    field :id, String, null: false
    # raises ArgumentError (missing keyword argument null:)
    field :name, String
  end
end

module Types
  class UserInput < ::GraphQL::Schema::InputObject
    argument :id, String, required: true
    # raises ArgumentError (missing keyword argument required:)
    argument :name, String
  end
end
```

In 1.7 and the [GraphQL spec](http://facebook.github.io/graphql/October2016/#sec-Types.Non-Null) fields default to nullable and arguments default to optional. The community seems to [favor optional fields](https://medium.com/@calebmer/when-to-use-graphql-non-null-fields-4059337f6fc8) for easy schema evolution and better partial failure handling. Should the new 1.8 class based API also default to making fields nullable and arguments optional? I'd be happy to make the change if you think it's a good idea.

rmosolgo:
I'm pretty torn. On the one hand, making them _required_ was borrowed from Shopify's GraphQL system, and I appreciate making people consider it. 

However, I also agree that `null: true` and `required: false` are sane defaults; they're future-friendly, as you mentioned. 

Let's leave this open for a bit and see if anyone else has an opinion!

jturkel:
@rmosolgo - Looks like no one else has weighed in for or against this issue. How do you feel about moving forward with it? I'd be happy to put up a PR.

eapache:
Just wanted to weigh in since I was one of the people who originally pushed for the current 1.8 approach; we've gone back and forth internally and settled on it being more useful to force people to answer the question. With a default people just leave everything nullable even when it's not, and that leads to worse APIs. I realize that's probably a contentious point though. It's also worth noting that we don't tend to use nullability for partial failure handling, so the majority of our fields are not nullable - also a contentious decision, but one that's worked well for us so far.

KevinColemanInc:
@rmosolgo just to quip in, it would be cool if this followed [rail's decision and made the default `null: false`](https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html).

I'd rather have a default rather not no default, no matter what it is.

jturkel:
[...]

PR #1457, opened 2018-04-26
Make field nullable by default

andersklenke:
Instead of explicitly having to set a field as able to be null it should default to `null: true`. 

**Background:**
When I tried the new class-based API I thought that this was the default behaviour. And the `missing keyword argument null` ArgumentError surprises me everytime.


jturkel:
FYI: This was raised in #1345 but rejected. I'm still in favor of it though :)

andersklenke:
@jturkel Sorry, must have missed that.

PR #1458, opened 2018-04-26
Update object_identification.md

pradyumna2905:
- Use `item_id` instead of `id` when decrypting object from ID.


rmosolgo:
oops, thanks!

PR #1459, opened 2018-04-27
Raise undefined `.graphql_name` in anonymous class

### LINKED ISSUE #1455: Better error when calling `.graphql_name` on anonymous class
Currently it's 

```
/Users/rmosolgo/github/github/vendor/gems/2.4.2/ruby/2.4.0/gems/graphql-1.8.0.pre10.104.gbb7627c5/lib/graphql/schema/member/base_dsl_methods.rb:21:in `graphql_name': undefined method `split' for nil:NilClass (NoMethodError)
```

But we can do better

david50407:
close #1455

### DIFF
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -15,10 +15,15 @@ module GraphQL
         # @param new_name [String]
         # @return [String]
         def graphql_name(new_name = nil)
-          if new_name
+          case
+          when new_name
             @graphql_name = new_name
-          else
-            overridden_graphql_name || name.split("::").last.sub(/Type\Z/, "")
+          when overridden = overridden_graphql_name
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi! Thanks for taking on this issue, it will be a really nice improvement :) 

How about adding a test for the new behavior? For example, you could add a new test case to `spec/graphql/schema/object_spec.rb` that asserts that a `NotImplementedError` is raised: 

```ruby 
assert_raises NotImplementedError do 
  # An anonymous class without explicit `.graphql_name` setting
  obj_class = Class.new(GraphQL::Schema::Object) 
  obj_class.graphql_name 
end 
```

What do you think?

david50407:
I also added spec for Ruby name fallback :)

rmosolgo:
👏  awesome, thanks for your work here!

Issue #1470, opened 2018-05-03
Field classes in the class-based API


mmun:
In the class-based API, is there a way to define field classes and hook them up to fields in your schema similar to how Mutation classes work?

This would help to reduce the clutter in my top-level Types::Query by externalizing field arguments and resolver functions to a separate file. There may be some other benefits as well (parity with mutations, reusing fields on different types, etc.)

rmosolgo:
Not yet, but that's a really interesting question!

To be honest, there's _room_ for some superclass of `Mutation` which is "just" a field, expressed as a class. In this way, it's something that takes the place of `GraphQL::Mutation`. 

I'll think on it a bit and try something out, I wonder if anyone else has thought about it ...

mmun:
That was my thinking as well. I mentioned it in Slack and it didn't garner much attention so I'm not sure how big of a need there is. Nevertheless, I think it would be a nice unification to have the option to write things like

```ruby
field :comments, class: Queries::Comments

field :add_comment, class: Mutations::AddComment

field :comment_changed, class: Subscriptions::CommentsChanged
```

rmosolgo:
Personally I think there is a need, `GraphQL::Function` is not quite right but we need the ability to split out code. I took a try in #1472; it's not quite there yet but it's a start.

zspencer:
I am also in favor of this. Being able to inherit from an ApplicationQuery would be super useful to allow us to remain writing conventional Ruby instead of the more pure-functional style.

Something like the following interface would feel really natural: 

```
module Types
  class Query < GraphQL::Schema::Object
    field :search, query: Queries::Search
  end
end

module Queries
  def current_user
    ctx[:current_user]
  end
end

module Queries
  class Search < ApplicationQuery
    argument :search, Types::SearchInput
    type Types::Search
    def resolve(search)
      if(current_user)
        PrivateSearch.new(user: current_user, parameters: search)
      else
        PublicSearch.new(parameters: search)
      end
    end
end
```

mmun:
@zspencer This is actually possible now in 1.8 as of https://github.com/rmosolgo/graphql-ruby/pull/1472. Your example would look like this:

```ruby
module Types
  class Query < GraphQL::Schema::Object
    field :search, resolver: Queries::Search
  end
end

module Queries
  class ApplicationQuery < GraphQL::Schema::Resolver
    def current_user
      context[:current_user]
    end
  end
end

module Queries
  class Search < ApplicationQuery
[...]

PR #1472, opened 2018-05-03
Extract SingletonField

rmosolgo:
Along the lines of #1470, I'm not sure really about the design or naming, and it needs tests.

I want to fix a problem of `GraphQL::Function`, which was that it is a foot-gun. Since the function object was long-lived, it was easy to use instance variables in an unsafe way, resulting in data being left behind from request-to-request, or the sharing of values between calls to `.resolve`.

I think the solution is to extract some superclass of `Schema::Mutation`, a container for resolve behavior and type information _but_ it has a short lifecycle. This new API could replace GraphQL::Function and you could transition to it over time.


rmosolgo:
> pure schemas

This was basically true in the old style, that singleton instances were used and reused all around. I didn't think we got very good mileage from that, and because people are used to short-lived objects like ActiveRecord instances and controller instances, they were prone to foot-gunning with improper usage of instance variables. 

So, now I'm trying to build in short-lived objects, to prevent foot-gunning. 

But I'm interested to hear more, is there a certain feature or advantage you're interested in there?

mmun:
No, actually I prefer the simplicity of short-lived objects for the reasons you've mentioned!

Returning to this PR and #1470, naively I would expect to be able to write something like

```ruby
class Fields::Comments < GraphQL::Schema::Field
  argument :order, Types::Order, null: false

  def resolve(order:)
    object.send(field_name).order(order)
  end
end


class Types::Post < GraphQL::Schema::Object
  field :comments, field_class: Fields::Comments
end

class Types::Photo < GraphQL::Schema::Object
  field :photo_comments, field_class: Fields::Comments
end
```

where each instance of Types::Photo/Types::Post would have its own instance of Fields::Comments and they would have similar lifetimes. I do realize that GraphQL::Schema::Field is already a thing and I'm not sure exactly what its current role is and whether it conflicts with this example.

rmosolgo:
Yeah, one thing I like about the snipped that you posted is that `GraphQL::Schema::Field` _is_ `GraphQL::Schema::SingletonField`, rather than having two classes. 

It would be pretty tough since `GraphQL::Schema::Field` already has a pretty structure lifecycle, but it's something to consider...

rmosolgo:
The other consideration is how code can be shared between "normal" fields and singleton fields. That is, the way to customize fields now is to make a base class from `Schema::Field`, using its instance lifecycle. That won't really apply for singleton resolution. But it would be a pain to require porting options manually from one to the other, and possibly they would come out of sync, making a rough developer experience. 

So, how could they be unified, or built as layers, to guarantee that extensions only have to be built and maintained in one spot, and won't fall out of sync? I want to think about this a bit, too.

lib/graphql/schema/resolver.rb
```diff
@@ -0,0 +1,128 @@
...
+          end
+          @type || (superclass.respond_to?(:type) ? superclass.type : nil)
+        end
+
+        private
+
+        # Build an instance of {.field_class} which uses this class.
+        # To customize field generation, override this method.
+        def generate_field
+          # TODO support deprecation_reason
```
michal-samluk: @rmosolgo Do you think there could be also support for `connection` option?

mmun:
@rmosolgo I like the new `Resolver` name much better. That makes a field conceptually decompose down to a (name, resolver) tuple which is pretty intuitive.

lib/graphql/schema/mutation.rb
```diff
@@ -58,42 +58,10 @@ class Schema
     #  }
     #  GRAPHQL
     #
-    class Mutation < GraphQL::Schema::Member
+    class Mutation < GraphQL::Schema::Resolver
```
pyromaniac: Oh wow, it is even called `Resolver` now. Great stuff!

lib/graphql/schema/resolver.rb
rmosolgo: Yes, I really need someway to _generalize_ about how to extend fields, passing arbitrary options to the `field(...)` call ... 🤔

jacob-s-son:
@rmosolgo, any ETA on this? I’ve transitioned most of our new codebase to the new class API and I am stuck with function objects that we use as field resolvers :(

swalkinshaw:
Am I the only one who doesn't like the `Resolver` name? 🤔

It seems counter-intuitive to me since the new class has a type and can have arguments which are usually on a field.

I realize we're pretty limited on available names that make sense though.

PR #1499, opened 2018-05-14
Fix BaseInterface generator to generate a module

### LINKED ISSUE #1498: [1.8] generated BaseInterface: superclass must be a class
The generated BaseInterface is not valid, because it attempts to create a class from a module:

```rb
class Types::BaseInterface < GraphQL::Schema::Interface
end
```

```
     TypeError:
       superclass must be a Class (Module given)
     # ./app/graphql/types/base_interface.rb:4:in `<module:Types>'
     # ./app/graphql/types/base_interface.rb:3:in `<top (required)>'
```

Based on [the docs](http://graphql-ruby.org/type_definitions/interfaces#defining-interface-types), I think this is what you want instead:

```rb
module Types::BaseInterface
end
```

bgentry:
Rather than trying to extend a module as though it were a superclass. Fixes #1498.

### DIFF
--- a/lib/graphql/railtie.rb
+++ b/lib/graphql/railtie.rb
@@ -72,7 +72,7 @@ module GraphQL
             unless File.exists?(destination_file)
               FileUtils.mkdir_p(File.dirname(destination_file))
               File.open(destination_file, 'w') do |f|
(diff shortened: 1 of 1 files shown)


rmosolgo:
Oh, thanks! I guess I forgot to update that :S

PR #1512, opened 2018-05-16
Fix AST node on input field

d12:
Fixes a bug in schema parsing which was setting input field AST nodes to the input object AST node.

cc @cjoudrey @rmosolgo


rmosolgo:
😬  oopsie, thanks for the fix!

PR #1515, opened 2018-05-17
Unions with no visible possible types should be hidden

cjoudrey:
Currently if you have a `union` type who's possible types are all hidden by a mask, the `union` type will not be hidden.

This is problematic when dumping a schema as it will cause an invalid SDL to be returned:

```graphql
#...

union Thing =

# ...
```

This PR contains a fix for this bug and added test coverage.

cc @xuorig if you can 👀 to confirm I didn't break anything.
fyi @d12 because you discovered the bug 😄

### DIFF
--- a/lib/graphql/schema/warden.rb
+++ b/lib/graphql/schema/warden.rb
@@ -140,7 +140,9 @@ module GraphQL
         return true if root_type?(type_defn)
         return true if type_defn.introspection?
 
-        if type_defn.kind.union? || type_defn.kind.interface?
+        if type_defn.kind.union?
+          visible_possible_types?(type_defn) && referenced?(type_defn)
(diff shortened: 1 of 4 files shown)


rmosolgo:
I guess the difference is that, previously, `referenced?(t)` was also applied to abstract types, even when they had no visible members, and if that was true, the type was visible. 

Now, `referenced?(t)` is never applied to abstract types, so a field reference is never enough to make it visible. Instead, the only consideration is about its members. 

The test case you added here certainly seems right to me, but I wonder if we still need the `referenced?(t)` check in some cases, I'm going to poke around locally.

rmosolgo:
👌 !

xuorig:
🎉 thanks for that great fix @cjoudrey

Issue #1526, opened 2018-05-22
In class based API, which is preferable/official API `object` or `@object`?


mtsmfm:
[Document](http://graphql-ruby.org/type_definitions/objects.html) shows us to use `object` to access the original object but upgrader convert to use `@object`

PR #1529, opened 2018-05-23
Use object/context method instead of @object/@context

mtsmfm:
ref https://github.com/rmosolgo/graphql-ruby/issues/1526


rmosolgo:
:+1: Thanks!

PR #1532, opened 2018-05-23
Fix accessible_by link is documentation

ijones16:
The markdown link was broken so because the underscore is missing.


rmosolgo:
Thanks!

PR #1541, opened 2018-05-29
Track position of union member nodes

cjoudrey:
When parsing union type members, we were not tracking the position of the `TypeName` node.

This PR fixes that and adds some test coverage.


rmosolgo:
Oops, nice find! Thanks for taking care of it.

PR #1543, opened 2018-05-29
Use a default empty hash if no arguments supplied

### LINKED ISSUE #1542: New API: Mutation doesn't resolve if no arguments declared in mutation class
@rmosolgo Hey, this used to work in the old setup but not with new class based API (or may be I am doing something wrong). 

Consider this mutation, with no arguments supplied:

```rb
class FooMutation < Base::Mutation
  graphql_name 'Foo'
  description 'returns foo'

  field :foo, String, null: false
  
  def resolve 
    "bar"
  end
end
```

Failing code:

```rb
# field.rb
puts "method", @method_sym
puts "kwargs", ruby_kwargs.inspect

if ruby_kwargs.any?
  obj.public_send(@method_sym, **ruby_kwargs)
else
  # seems to work if we pass an empty hash but then it fails on query calls
  obj.public_send(@method_sym)
end
```

```rb
# relay_classic_mutation.rb
def resolve_mutation(kwargs)
  # This is handled by Relay::Mutation::Resolve, a bit hacky, but here we are.
  puts "resolve", kwargs.inspect
  kwargs.delete(:client_mutation_id)
  resolve(**kwargs)
end
```

Stacktrace:

```bash
ArgumentError (wrong number of arguments (given 0, expected 1)):
api :
api : /Users/admin/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/graphql-1.8.0/lib/graphql/schema/relay_classic_mutation.rb:30:in `resolve_mutation'
api : /Users/admin/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/graphql-1.8.0/lib/graphql/schema/field.rb:367:in `public_send'
api : /Users/admin/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/graphql-1.8.0/lib/graphql/schema/field.rb:367:in `public_send_field'
api : /Users/admin/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/graphql-1.8.0/lib/graphql/schema/field.rb:288:in `resolve_field'
api : /Users/admin/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/graphql-batch-0.3.9/lib/graphql/batch/setup.rb:19:in `call'
```

gauravtiwari:
Fixes #1542 

//cc @rmosolgo

### DIFF
--- a/lib/graphql/schema/relay_classic_mutation.rb
+++ b/lib/graphql/schema/relay_classic_mutation.rb
@@ -27,10 +27,14 @@ module GraphQL
 
       # Override {GraphQL::Schema::Mutation#resolve_mutation} to
       # delete `client_mutation_id` from the kwargs.
-      def resolve_mutation(kwargs)
+      def resolve_mutation(**kwargs)
         # This is handled by Relay::Mutation::Resolve, a bit hacky, but here we are.
         kwargs.delete(:client_mutation_id)
-        resolve(**kwargs)
+        if kwargs.any?
+          resolve(**kwargs)
+        else
+          resolve
+        end
       end
 
       resolve_method(:resolve_mutation)
--- a/lib/graphql/schema/relay_classic_mutation.rb
+++ b/lib/graphql/schema/relay_classic_mutation.rb
@@ -27,7 +27,7 @@ module GraphQL
 
       # Override {GraphQL::Schema::Mutation#resolve_mutation} to
       # delete `client_mutation_id` from the kwargs.
-      def resolve_mutation(kwargs = {})
+      def resolve_mutation(kwargs)
         # This is handled by Relay::Mutation::Resolve, a bit hacky, but here we are.
         kwargs.delete(:client_mutation_id)
         resolve(**kwargs)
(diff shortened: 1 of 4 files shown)


rmosolgo:
This looks good to me, but could you add a test to make sure there's no regression? For example, you could add a mutation with no arguments to `spec/support/jazz.rb`, and hook it up to the schema there, then run it in a test to be sure it was executed properly.

gauravtiwari:
Done, is that commit looks okay to you?

rmosolgo:
I think the test could use _one_ more step: now that an example mutation is added to the schema (:+1:), we should add a test which _calls_ the new mutation to be sure that it is properly executed. 

For example, you could find `spec/graphql/schema/relay_classic_mutation_spec.rb`, and add a test there which executes `mutation { addSitar(input: {}) { instrument { name } } }`, then check that the returned name is `Sitar` as expected. 

By the way, since the fix for this issue was added to `RelayClassicMutation`, should we use `RelayClassicMutation` as the base class for `AddSitar`?

rmosolgo:
Thanks for your work on this, I pushed a few more commits to get it over the finish line: 

- Add a test for the new mutation; it turned out to pass without changes 
- Refactor the added mutation to use `RelayClassicMutation`, then the test failed 
- Fix the test by adding a similar fix

PR #1547, opened 2018-05-30
Guides : Wrong metadata name in concrete class

DucMirack:


rmosolgo:
oops, thanks for the fix!!

PR #1548, opened 2018-05-30
Only connect to MongoDB when required

### LINKED ISSUE #1522: Separate MongoDB support from the core
From #1452 on, graphql-ruby requires mongodb to run tests, but I'm not interested in MongoDB.

Can you consider to separate MongoDB support from the core? I think [kaminari](https://github.com/kaminari) is a good example.

rmosolgo:
We have mongodb support now (:tada:) but making it required for _all_ tests can be a damper on developing the gem. This change makes it so it's: 

- Only loaded when needed
- Skipped when not found 

For example, when it's absent, you get: 

```
SSSSSSSSSSSSSSSSSSSSS

Finished tests in 0.004816s, 4360.4652 tests/s, 0.0000 assertions/s.


Skipped:
GraphQL::Relay::MongoRelationConnection::results#test_0007_handles cursors below the bounds of the array [/Users/rmosolgo/code/graphql-ruby/spec/graphql/relay/mongo_relation_connection_spec.rb:12]:
Mongo not detected
```

Fixes #1522

### DIFF
--- a/spec/graphql/relay/mongo_relation_connection_spec.rb
+++ b/spec/graphql/relay/mongo_relation_connection_spec.rb
@@ -1,7 +1,18 @@
 # frozen_string_literal: true
 require 'spec_helper'
 
-describe GraphQL::Relay::RelationConnection do
+if MONGO_DETECTED
+  require "support/star_trek/data"
+  require "support/star_trek/schema"
+end
+
+describe GraphQL::Relay::MongoRelationConnection do
+  before do
+    if !MONGO_DETECTED
+      skip("Mongo not detected")
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


dblock:
Coming from #1754, maybe a cleaner solution would have been to condition everything on `defined?(Mongoid)` vs. presence of a mongodb database and include/exclude those tests via Rakefile's spec patterns. Similar for PG. Eventually having both is going to create a bunch of unnecessary conflicts. Just thinking out loud, if you think it's better I can put that on my backburner.

rmosolgo:
:+1: definitely open to a better implementation here!

PR #1549, opened 2018-05-31
Remove redundant if branch in ensure_hash method

DamirSvrtan:
This removes a redundant branch and improves readability IMO.


rmosolgo:
Thanks for the suggestion, but I'm happy to leave it as is, so that all the `return` expressions are together, instead of spread out. If people want to outvote me here, I'm open to that, and I'll accept this with the necessary updates to the tests.

PR #1550, opened 2018-05-31
Add dedicated errors guides

rmosolgo:
There is some really old documentation out there, but nothing recent, and no decent introduction to GraphQL-Ruby errors, so here's some new documentation.


guides/mutations/mutation_errors.md
```diff
@@ -0,0 +1,129 @@
...
+
+Another way to handle rich error information is to add _error types_ to your schema, for example:
+
+```ruby
+class Types::UserError < Types::BaseObject
+  description "A user-readable error"
+
+  field :message, String, null: false,
+    description: "A description of the error"
+  field :attribute, String, null: false,
```
cjoudrey: Have you considered advising to use `[String!]` instead of `String!` here to future proof for nested arguments.

We may also want to advise using `path` as a field name instead of `attribute`. The `path` would be scoped to the current field.

We could also include a request/response example like this:

```graphql
mutation {
  createPost(title: "", body: "") {
    post {
      id
    }
    errors {
      message
      path
    }
  }
}
```

```json
{
  "data": { 
    "createPost": {
      "post": null,
      "errors": [
        { "message": "Title can't be blank", "path": ["title"] },
        { "message": "Title can't be blank", "path": ["body"] }
      ]
    }
  }
}
```

rmosolgo: :+1: sure, I'll update that

guides/errors/execution_errors.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Customizing Error JSON
+
+The default error JSON includes `"message"`, `"locations"` and `"path"`. You can customize this in two ways:
+
+- Pass `options:` when raising an error, for example:
+  ```ruby
+  raise GraphQL::ExecutionError.new("Something went wrong", options: { "code" => "BROKEN" })
+  ```
+  In this case, `"code" => "BROKEN"` will be added to the error JSON.
```
swalkinshaw: I'm pretty sure the spec says any custom properties should be nested under `extensions`. Right now `options` just  get's merged into the error itself. I was planning on doing a PR to fix/support this anyway, but it should probably be mentioned or accounted for in the docs as well.

See section below http://facebook.github.io/graphql/draft/#example-08b62

eapache: > Previous versions of this spec did not describe the extensions entry for error formatting. 
 While non‐specified entries are not violations, they are still discouraged.

I definitely had not seen that before.

rmosolgo: Thanks for the pointer, I updated it a bit in 81b22820 to make that suggestion and use `"extensions"` in the examples. Maybe a nice feature would be to add an `extensions:` keyword that merged that value into any previously-added extensions.

PR #1551, opened 2018-05-31
Properly validate interfaces without selections

swalkinshaw:
Previously this was only checking object types. This means a query containing a field which returns an interface type is valid without any selections. This is wrong according to the spec and the reference JS implementation though.

This updates the validation rule to use `fields?` instead of `object?`.

The error message has also been updated to reflect it's not just object types and a helpful "Did you mean" part was borrowed from graphql-js.

@rmosolgo let me know if you want the error message changed


rmosolgo:
Wow, that bug has been here for a _looong_ time! Looks good to me, but do you mind rounding up that one last failing test?

swalkinshaw:
Done 🎉

rmosolgo:
🙌 Thanks !!

PR #1552, opened 2018-06-01
Add extensions support for execution errors

swalkinshaw:
`ExecutionError` already supports `options` to merge in custom data. However the GraphQL spec now recommends that all additions entries to an error be under the `extensions` key to avoid conflicts with any future spec changes.

Ref: http://facebook.github.io/graphql/draft/#sec-Errors

This adds the ability to specific `extensions` which will get merged into any existing `extensions` data.

`options` has also been marked as deprecated.


rmosolgo:
👏 !

Issue #1281, opened 2018-02-14
[1.8] Argument to_h


joshforbes:
On 1.7 I am able to call args.to_h on an argument and any nested input objects will also be converted. On 1.8 this will only convert the top level object all the nested items will remain as objects. Not sure if this was an intentional change.

Specifically for me this is happening in the context of a graphql function but I suspect it would be true anywhere.

rmosolgo:
No, it wasn't intentional, I noticed it myself too and just hacked around it 😖  

The problem is that in 1.8, `args` is converted to a real Ruby Hash, so `.to_h` calls `Hash#to_h`, not `GraphQL::Query::Arguments#to_h`. I don't really want to monkey-patch `Hash#to_h`, because that could have some unexpected side effects. 

🤔  I wonder what options we have for providing the same deep `to_h` behavior ...

jturkel:
I hit this issue too and also hacked around it. Would it make sense to modify `GraphQL::Schema::InputObject#to_kwargs` to always do a "deep" `to_kwargs`?

joshforbes:
With the way mutation and resolver objects handle arguments now. This is no longer a concern for me. I don't know if it is relevant for others?

jturkel:
I'm still hitting this issue using the class based mutation API.

PR #1554, opened 2018-06-02
Improve Arguments#to_h for nested input objects

schpet:
this is an attempt at fixing #1281 and does a few things:

- changes `Arguments#to_h` to recursively hit `unwrap_values` so that nested attributes have ruby style keyword args
- adds `GraphQL::Schema::InputObject` as a thing that gets unwrapped, since the nested arguments seem to be instances of those

i also added a commented out alternative approach, which keeps nested stuff as instances of `Arguments`. not sure which makes sense, this bit in `InputObject` confuses me:

https://github.com/rmosolgo/graphql-ruby/blob/022477fdb2f29467659f2009161329d7e87bf579/lib/graphql/schema/input_object.rb#L74-L78

anyway i'm very excited to have discovered the class based api! thanks for looking at this and i'd be grateful for any feedback 🙌🏻

update: oh no, i see this breaks things. i'll try to get things working.

### DIFF
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -108,10 +108,14 @@ module GraphQL
         ruby_kwargs = {}
 
         keys.each do |key|
-          ruby_kwargs[Schema::Member::BuildType.underscore(key).to_sym] = self[key]
+          value = unwrap_value(self[key])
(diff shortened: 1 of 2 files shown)


schpet:
closing this in favor of #1555

PR #1555, opened 2018-06-03
Unwrap values in InputObject#to_h

### LINKED ISSUE #1281: [1.8] Argument to_h

schpet:
recursively unwraps values in `InputObject#to_h` in a way that's consistent with `Arguments`

fixes #1281

### DIFF
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -27,7 +27,28 @@ module GraphQL
       attr_reader :arguments
 
       # Ruby-like hash behaviors, read-only
-      def_delegators :@ruby_style_hash, :to_h, :keys, :values, :each, :any?
+      def_delegators :@ruby_style_hash, :keys, :values, :each, :any?
+
+      def to_h
+        @ruby_style_hash.inject({}) do |h, (key, value)|
+          h.merge(key => unwrap_value(value))
+        end
+      end
+
+      def unwrap_value(value)
+        case value
+        when Array
+          value.map { |item| unwrap_value(item) }
+        when Hash
+          value.inject({}) do |h, (key, value)|
+            h.merge(key => unwrap_value(value))
+          end
+        when InputObject
+          value.to_h
+        else
+          value
+        end
+      end
 
       # Lookup a key on this object, it accepts new-style underscored symbols
       # Or old-style camelized identifiers.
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hey, thanks for opening this PR. I'm really interested in improving this, but let me find a chance to consider the compatibility side of things. I would hate for someone's app to start breaking because they previously used one kind of key, and now they need to use the other kind!

schpet:
thanks for checking this out! please let me know if i can make any changes to improve backwards compatibility, e.g. using camel case keys instead.

spec/graphql/schema/input_object_spec.rb
```diff
@@ -108,4 +108,36 @@ class Schema < GraphQL::Schema
...
+        argument :a, Int, required: true
+        argument :b, Int, required: true
+        argument :c, TestInput1, as: :inputObject, required: true
+      end
+
+      TestInput1.to_graphql
+      TestInput2.to_graphql
+    end
+
+    it "returns a stringified, aliased, ruby keyword style hash" do
```
rmosolgo: Should this say "symbolized" instead of "stringified"?

rmosolgo: (I think besides that, this change is :+1:)

schpet: fixed! thank you

rmosolgo:

lib/graphql/schema/input_object.rb
```diff
@@ -27,7 +27,28 @@ def initialize(values, context:, defaults_used:)
       attr_reader :arguments
 
       # Ruby-like hash behaviors, read-only
-      def_delegators :@ruby_style_hash, :to_h, :keys, :values, :each, :any?
+      def_delegators :@ruby_style_hash, :keys, :values, :each, :any?
+
+      def to_h
+        @ruby_style_hash.inject({}) do |h, (key, value)|
+          h.merge(key => unwrap_value(value))
```
jturkel: Should this do a `merge!` to avoid allocating a new hash for each entry? Same question in `unwrap_value`.

PR #1557, opened 2018-06-05
Validate argument names are unique

cjoudrey:
[As per the spec](http://facebook.github.io/graphql/October2016/#sec-Argument-Uniqueness):

> Fields and directives treat arguments as a mapping of argument name to value. More than one argument with the same name in an argument set is ambiguous and invalid.

This PR adds extra validation to adhere to the spec and remove ambiguity.

/cc @d12 @nickvanw as they discovered this

### DIFF
--- a/lib/graphql/static_validation/all_rules.rb
+++ b/lib/graphql/static_validation/all_rules.rb
@@ -24,6 +24,7 @@ module GraphQL
       GraphQL::StaticValidation::ArgumentsAreDefined,
       GraphQL::StaticValidation::ArgumentLiteralsAreCompatible,
       GraphQL::StaticValidation::RequiredArgumentsArePresent,
+      GraphQL::StaticValidation::ArgumentNamesAreUnique,
       GraphQL::StaticValidation::VariableNamesAreUnique,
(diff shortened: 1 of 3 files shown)


lib/graphql/static_validation/all_rules.rb
```diff
@@ -24,6 +24,7 @@ module StaticValidation
       GraphQL::StaticValidation::ArgumentsAreDefined,
       GraphQL::StaticValidation::ArgumentLiteralsAreCompatible,
       GraphQL::StaticValidation::RequiredArgumentsArePresent,
+      GraphQL::StaticValidation::ArgumentNamesAreUnique,
```
cjoudrey: I wasn't exactly sure where to add this new rule, but I think here should be fine.

rmosolgo: Sadly a few of them are order-dependent 😩  but I think as long as it doesn't 💥  then we're fine.

rmosolgo:
Nice catch, and thanks for taking care of it!

gjtorikian:
Dude... I literally called @grokys out for this today. Assuming that's where the @nickvanw call came in from? 🤣

PR #1558, opened 2018-06-05
Propagate null for list type of non-null types

### LINKED ISSUE #1510: List and Non-Null Element type should propagate nulls
Tested on
* 1.7.14
* 1.8.0pre11

-----------

From [the spec](http://facebook.github.io/graphql/October2016/#sec-Combining-List-and-Non-Null)

> If the modified type of a List is Non‐Null, then that List may not contain any null items.

So for a type like **\[!** String **\]**, an array of null elements must not be returned as *data*. 

Rather than returning the entire List as null, the null elements are still returned.

```ruby
require 'graphql'

  name "Query"

  field :nonNullArrayWithNullElements, types[!types.String], "" do
    resolve ->(obj, args, ctx) {
      [nil]
    }
  end
end

  query QueryType
end

result = MySchema.execute("{ nonNullArrayWithNullElements }").to_h

print result["data"]["nonNullArrayWithNullElements"] == nil ? "✔️" : "✖︎"
puts " nonNullArrayWithNullElements: [String!] must not return array of null elements"
puts
puts result
puts
```

```log
✖︎ nonNullArrayWithNullElements: [String!] must not return array of null elements

{"data"=>{"nonNullArrayWithNullElements"=>[nil]}, "errors"=>[{"message"=>"Cannot return null for non-nullable field Query.nonNullArrayWithNullElements"}]}
```

cjoudrey:
Fixes #1510 

This is my first time playing in the executor code so I'm not sure if other changes need to be made.

I believe this fixes the issue reported by @joshjcarrier. I've tried with the repro script attached to the issue and have added similar code in our tests.

As a precaution I also confirmed this behaviour in the reference implementation:

```js
var graphql = require("graphql");

var schema = new graphql.GraphQLSchema({
  query: new graphql.GraphQLObjectType({
    name: 'RootQueryType',
    fields: {
      hello: {
        type: graphql.GraphQLList(graphql.GraphQLNonNull(graphql.GraphQLString)),
        resolve() {
          return [null];
        }
      }
    }
  })
});

var query = '{ hello }';

graphql.graphql(schema, query).then(result => {
  console.log(result);
});

// { errors:
//    [ { message: 'Cannot return null for non-nullable field RootQueryType.hello.',
//        locations: [Array],
//        path: [Array] } ],
//   data: { hello: null } }
```

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -198,8 +198,6 @@ module GraphQL
               result = []
               field_ctx.value = result
 
-              non_null_inner_type = inner_type.non_null?
-
               value.each do |inner_value|
                 inner_ctx = field_ctx.spawn_child(
                   key: i,
@@ -213,11 +211,11 @@ module GraphQL
                   inner_ctx,
                 )
 
+                return PROPAGATE_NULL if inner_result == PROPAGATE_NULL
+
                 inner_ctx.value = inner_result
                 result << inner_ctx
                 i += 1
-
-                return PROPAGATE_NULL if non_null_inner_type && inner_ctx.value.nil?
               end
 
               result
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -198,6 +198,8 @@ module GraphQL
               result = []
               field_ctx.value = result
 
+              non_null_inner_type = inner_type.non_null?
+
               value.each do |inner_value|
                 inner_ctx = field_ctx.spawn_child(
                   key: i,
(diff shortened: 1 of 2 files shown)


lib/graphql/execution/execute.rb
```diff
@@ -215,7 +215,12 @@ def resolve_value(value, field_type, field_ctx)
                 result << inner_ctx
                 i += 1
               end
-              result
+
+              if inner_type.non_null? && result.any? { |result| result.value.nil? }
```
cjoudrey: A more efficient approach could be to short-circuit the `value.each` above if we see any `inner_result.nil?`, but I wasn't sure if it was okay to do this as it might lead to unresolved lazy resolvers. Is that ok?

rmosolgo: I _think_ it's a _feature_! Right, because instead of doing work, then throwing away the result, we skip the unnecessary work altogether.

cjoudrey: You're right, short-circuiting is a much better approach as we'll stop resolving child nodes.

rmosolgo:
👏 🎊 !

PR #1559, opened 2018-06-05
Update lots of guides for class-based API

rmosolgo:


guides/relay/connections.md
```diff
@@ -10,43 +10,43 @@ index: 1
...
+field :featured_comments, CommentType.connection_type do
   # Add an argument:
-  argument :since, types.String
+  argument :since, String, required: false
+end
 
+def featured_comments(since: nil)
+  comments = post.comments.featured
+  if since
+    comments = comments.where("created_at >= ?", since)
```
Willianvdv: the missing `?`, nice catch! :D

guides/relay/connections.md
```diff
@@ -55,13 +55,13 @@ end
 You can limit the number of results with `max_page_size:`:
 
 ```ruby
-connection :featured_comments, CommentType.connection_type, max_page_size: 50
+field :featured_comments, CommentType.connection_type, null: falsemax_page_size: 50
```
michal-samluk: `falsemax_page_size` instead of: `false, max_page_size`

guides/relay/connections.md
```diff
@@ -87,9 +87,9 @@ end
 Now, you can use `PostConnectionWithTotalCountType` to define a connection with the "totalCount" field:
 
 ```ruby
-AuthorType = GraphQL::ObjectType.define do
+class AuthorType < GraphQL::Schema::Object
   # Use the custom connection type:
-  connection :posts, PostConnectionWithTotalCountType
+  field :posts, PostConnectionWithTotalCountType, null: false
```
michal-samluk: When the type name does not end with `Connection` then you need to pass `connection: true` option.

guides/relay/mutations.md
```diff
@@ -41,10 +45,9 @@ For Relay-compliant GraphQL, a mutation field must comply to a strict API. `Grap
 After defining a mutation (see below), add it to your mutation type:
 
 ```ruby
-MutationType = GraphQL::ObjectType.define do
-  name "Mutation"
+class MutationType < GraphQL::Schema::Object
   # Add the mutation's derived field to the mutation type
-  field :addComment, field: AddCommentMutation.field
+  field :add_comment, field: AddCommentMutation.field
```
michal-samluk: I think `mutation:` should be used instead of `field:`

Willianvdv: No, a mutation type is just an object type and so it defines fields. For example, check out how the mutation is used in the specs: https://github.com/rmosolgo/graphql-ruby/blob/f77dd391f3e8a46b094808820c15e858a3ef7bd1/spec/support/star_wars/schema.rb#L334

guides/relay/connections.md
Willianvdv: `PostConnectionWithTotalCountType` is being created as a connection using the `defined_connection` method. As these types will be postfixed with `Connection, I'm not sure if you need this connection argument as the lib can detect that based on that connection return type.

guides/relay/mutations.md
Willianvdv: Sorry, I misread your comment. You're right, the argument should be mutation rather than field.

guides/relay/connections.md
michal-samluk: Connection is just a type with connection specific fields. You need to configure the field so the field resolver could be wrapped with connection resolver. 
And currently it is done by connection option or by the object type name. 

That's what current documentation says (http://graphql-ruby.org/schema/class_based_api.html#connection-fields--types):
>>>
There is no connection(...) method. Instead, connection fields are inferred from the type name.
If the type name ends in Connection, the field is treated as a connection field.

rmosolgo: Oh, thanks for the correction !

guides/relay/mutations.md
rmosolgo: There's some confusion here because of `GraphQL::Relay::Mutation` (old) vs `GraphQL::Schema::Mutation` (new) 

For the _old_ one, you can do `field: OldMutation.field`. For the _new_ one, you can do `mutation: NewMutation`. 

This guide starts with a note to direct people to the new mutation docs. And in a month or so, I'll probably remove this one altogether, so it's ok with me to leave it for now.

rmosolgo:
Thanks for your careful reviews, I really appreciate it, and as you can see, I also really _need_ it 😆

PR #1563, opened 2018-06-06
Fix default_mask configuration not being compatible in the class-based API

obrie:
Prior to 1.8, I used to be able to do:

```ruby
GraphQL::Schema.define do
  # ...
  default_mask(...)
end
```

This doesn't work on the class-based API in 1.8:

```ruby
  # ...
  default_mask(...)
end
```

This PR adds support to be compatible between the old and new APIs.


rmosolgo:
Cool, thanks for the patch, and sorry for the slip on that one 😊 !

PR #1564, opened 2018-06-07
Add DirectiveLocation AST node

cjoudrey:
I tried to use `Language::Visitor` to visit a schema definition and noticed it would crash when visiting a `DirectiveDefinitionNode`:

```
graphql-1.8.0.pre10/lib/graphql/language/visitor.rb:47:in `visit_node': undefined method `children' for "SCALAR":String (NoMethodError)
```

This is because after visiting a `DirectiveDefinition` node, the visitor visits its `children` which consists of the directive's argument nodes and location nodes. The latter were Ruby strings so it would 💥.

I believe the root cause of this issue is that we don't have an AST node for directive locations. This is also problematic for anyone using GraphQL Ruby to parse a schema as they will lose position information for directive locations.

[GraphQL.js appears to have a generic `Name` node that they use in this case](https://github.com/graphql/graphql-js/blob/43c0d46ac9a279a1c3be9c8592381e2a99eec209/src/language/parser.js#L180-L191). We seem to have specific nodes for this, so I went with the `DirectiveLocation` node name.  Let me know if you prefer something else. 😄

### DIFF
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -149,8 +149,18 @@ module GraphQL
         GraphQL::Language::Nodes::DirectiveDefinition.new(
           name: directive.name,
           arguments: build_argument_nodes(warden.arguments(directive)),
-          locations: directive.locations.map(&:to_s),
+          locations: build_directive_location_nodes(directive.locations),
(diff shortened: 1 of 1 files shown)


cjoudrey:
Oh! Good to know! Thanks Robert! 😄

PR #1570, opened 2018-06-10
Fix mutation guide example code

kenny-hibino:
Fixed inconsistency between field name defined with `field` macro and what's actually returned from `resolve` method.


rmosolgo:
Thanks!!

PR #1572, opened 2018-06-11
Update arg description on before/after connection args

d12:
Previously, the description stated before/after took in a global ID, but connection before/after args take cursors, not global IDs.

cc @rmosolgo


rmosolgo:
Oops 😊  Thanks for the fix!

PR #1574, opened 2018-06-12
Update Execution errors guide for extensions

swalkinshaw:
Change the example to take advantage of new `extensions` feature from https://github.com/rmosolgo/graphql-ruby/pull/1552


rmosolgo:
🤘

Issue #1576, opened 2018-06-12
[1.8] TypeError (superclass must be a Class (Module given))


everdrone:
> user_type.rb
```ruby
class Types::UserType < Types::BaseObject
  implements GraphQL::Relay::Node.interface, Types::OwnerType
  # ...
end
```

> owner_type.rb
```ruby
class Types::OwnerType < Types::BaseInterface
  field :login, String, null: false
  # ...
end
```

I suppose the error is referring to the generated `base_interface.rb` ?
```ruby
module Types::BaseInterface
  include GraphQL::Schema::Interface
end
```

`TypeError (superclass must be a Class (Module given))`

rmosolgo:
Check out the stack trace on the error message, which line of code is it pointing to? 

Then compare your `Types::OwnerType` to the interface examples here: 

http://graphql-ruby.org/type_definitions/interfaces.html#defining-interface-types

Your interfaces should be Ruby `module`s, for example: 

```ruby 
module Types::OwnerType 
  include Types::BaseInterface 
  # ... 
end 
```

Feel free to reopen if that doesn't help sort out the issue!

PR #1577, opened 2018-06-13
Fix graphql:interface generator

everdrone:
changed generated class implementation to module

related to https://github.com/rmosolgo/graphql-ruby/issues/1576


rmosolgo:
Derp! I botched it back in #1562 . Thanks for fixing it. Do you want to take a stab at updating the tests to reflect the corrected output, too?

everdrone:
sure @rmosolgo, i'll be happy to work on the tests ASAP

rmosolgo:
Awesome, thanks again for the fix!

PR #1579, opened 2018-06-13
Update timeout_middleware.rb

phaedryx:
Reasoning: I want to create a couple of subclasses of `TimeoutMiddleware` and having them use the same namespace was causing conflicts.


rmosolgo:
Nice!

PR #1580, opened 2018-06-13
Add possible type helpers to abstract types

swalkinshaw:
This adds `get_possible_type` and `possible_type?` to `InterfaceType` and `UnionType`.

These methods differ from the static `Schema.possible_types` since they take into account the context and go through warden (for things like visibility).

`get_possible_type` returns the type if it exists and is a possible type from a type name.

`possible_type?` accepts with a type definition or a type name and returns a boolean.

### DIFF
--- a/lib/graphql/interface_type.rb
+++ b/lib/graphql/interface_type.rb
@@ -64,5 +64,23 @@ module GraphQL
     def all_fields
       fields.values
     end
+
+    # Get a possible type of this {InterfaceType} by type name
+    # @param type_name [String]
+    # @param ctx [GraphQL::Query::Context] The context for the current query
+    # @return [GraphQL::ObjectType, nil] The type named `type_name` if it exists and implements this {InterfaceType}, (else `nil`)
+    def get_possible_type(type_name, ctx)
+      type = ctx.query.get_type(type_name)
+      type if type && ctx.query.schema.possible_types(self).include?(type)
+    end
+
+    # Check if a type is a possible type of this {InterfaceType}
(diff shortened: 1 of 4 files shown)


lib/graphql/interface_type.rb
```diff
@@ -64,5 +64,23 @@ def get_field(field_name)
     def all_fields
       fields.values
     end
+
+    # Get a possible type of this {InterfaceType} by type name
+    # @param type_name [String]
+    # @param ctx [GraphQL::Query::Context] The context for the current query
+    # @return [GraphQL::ObjectType, nil] The type named `type_name` if it exists and implements this {InterfaceType}, (else `nil`)
+    def get_possible_type(type_name, ctx)
```
swalkinshaw: I just duplicated these methods for both types right now even though they are identical (other than doc strings). Would you rather these be moved into a module or something?

spec/graphql/interface_type_spec.rb
```diff
@@ -149,4 +149,50 @@
...
+
+  describe "#get_possible_type" do
+    let(:query_string) {%|
+      query fav {
+        favoriteEdible { fatContent }
+      }
+    |}
+
+    let(:query) { GraphQL::Query.new(Dummy::Schema, query_string) }
+    let(:dummy_query_context) { OpenStruct.new(schema: Dummy::Schema, query: query) }
```
swalkinshaw: Not sure if there's a preferred/easier way of getting a "real" context for tests like these.

PR #1581, opened 2018-06-13
Add Example For Overriding Default Object Field Parameters

ijones16:
As per [this comment](https://github.com/rmosolgo/graphql-ruby/pull/1300#issuecomment-396933931).


rmosolgo:
Awesome, thanks!

PR #1584, opened 2018-06-14
Include directives in InputObjectTypeDefinition's children

cjoudrey:
When using `GraphQL::Language::Visitor`, each node will be visited and then its child nodes (defined by the `children` method) will be visited. 

Currently the `children` of `InputObjectTypeDefinition` is only its fields. This means it is not possible to visit the directives of an `InputObjectTypeDefinition`.

This PR fixes this.


rmosolgo:
oopsie, thanks for the fix!

PR #1590, opened 2018-06-15
Allow Resolvers to define complexity

obrie:
Resolvers can't currently define the complexity for the field it's being applied to.  This change now allows them to do that.

I've added some tests based on how I saw it being done elsewhere, but feel free to clean them up or leave feedback on anything I should change.

Thanks!

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -99,6 +99,15 @@ module GraphQL
           end
         end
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
Looks good to me! Those test failures look like CI issues, I just rebuilt them and I'll let em pass for good measure.

rmosolgo:
Thanks for the fix!

PR #1599, opened 2018-06-19
index.html page: Link to an existing article for explaining Types

olleolleolle:
- the types/introduction text is no longer in there
  - the schema definition text is _perhaps_ where to begin, now?


rmosolgo:
Works for me, thanks for the fix!

Issue #1269, opened 2018-02-08
scalar coercion borks if argument passed in via variable


axos88:
A mutation, 

```
AddMember = GraphQL::Relay::Mutation.define do
      name 'addMemberToAgentGroup'

      input_field :group_id, !ObjectId, 'The id of the group to add the agent(s) for'
      input_field :agent_ids, !types[!ObjectId]

      return_field :agentGroup, Type, property: :data

      resolve ->(_obj, args, ctx) do
        (...)
      end
end
```

having an argument of type:

```
    ObjectId = GraphQL::ScalarType.define do
      name "ObjectId"
      description "Hexadecimal representation of a 12-byte hexadecimal id, consisting of: a 4-byte value representing the seconds since the Unix epoch, a 3-byte machine identifier, a 2-byte process id, and a 3-byte counter, starting with a random value."

      coerce_input ->(value, ctx) do
        value.to_s.downcase.tap do |v|
          raise GraphQL::CoercionError, " '#{v}' is not a 12-byte hexadecimal string" unless OBJECT_ID_REGEX.match(v)
        end
      end

      coerce_result ->(value, ctx) { value.to_s.downcase }
    end
```

will return the coercion result if passed in as a literal, although the message too is duplicated:

```
mutation {
  addMemberToAgentGroup(input: {
    group_id: "invalid",
    agent_ids: []
  }) {
    agentGroup {
      id
      members {
        data {
          id
        }
      }      
    }    
  }
}
```

```
{
  "errors": [
    {
      "message": " 'invalid' is not a 12-byte hexadecimal string",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "fields": [
        "mutation",
        "addMemberToAgentGroup",
        "input"
      ]
    },
    {
      "message": " 'invalid' is not a 12-byte hexadecimal string",
      "locations": [
        {
          "line": 2,
          "column": 32
        }
      ],
      "fields": [
        "mutation",
        "addMemberToAgentGroup",
        "input",
        "group_id"
      ]
    }
  ]
}
```

But returns an exception if the argument is sent in as a variable:

```
mutation($input: addMemberToAgentGroupInput! ) {
  addMemberToAgentGroup(input: $input) {
    agentGroup {
      id
      members {
        data {
          id
        }
      }      
    }
    
  }
}
```

```
{
  "input": {
    "group_id": "invalid",
    "agent_ids": []
  }
}
```

```
[...]

Issue #1024, opened 2017-10-13
Argument's prepare behaviour differs between literal input and input objects


swalkinshaw:
When argument values are specified inline, an `argument`'s `prepare` function works as expected:

https://github.com/rmosolgo/graphql-ruby/blob/17a497d0ed4514b7af90ee1abf5e1ce18e9b12a1/lib/graphql/query/literal_input.rb#L79-L83

Each argument's value is prepared and raises if it returns a `GraphQL::ExecutionError`.

However, if you use variables for argument values, the behaviour differs.

https://github.com/rmosolgo/graphql-ruby/blob/17a497d0ed4514b7af90ee1abf5e1ce18e9b12a1/lib/graphql/input_object_type.rb#L93

There's no error checking going on here. As `ExecutionError` gets returned from `prepare` and set as the value of the argument but it's not raised.

My naive attempt was to do a similar check and add an error to `ctx`, but we have no AST context  within the input object type.

rmosolgo:
😖  This is gonna be a toughie

swalkinshaw:
I dug into this a bit more and came away more confused than before 😄 

Here's a failing test case though: https://github.com/rmosolgo/graphql-ruby/compare/master...swalkinshaw:argument-prepare-variables

rmosolgo:
I took a try at clearing up the preparation steps this morning, but it failed: https://github.com/rmosolgo/graphql-ruby/compare/eager-query-processing

So i will catch my breath and try again tomorrow 😓

modosc:
is there any progress on this here (or another branch)? we are having major issues with error handling right now and i just traced it down to this behavior.

modosc:
as a workaround i ended up writing code to inline any incoming query variables which feels dirty but solves the problem for us.

modosc:
i'm going to look into this now, i'm having a difficult time porting my query inliner to 1.9 and it might be easier to solve this.

jturkel:
Is the behavior consistent if you raise a `GraphQL::ExecutionError` rather than return one? I added some tests as part of #2198 that seem to cover that scenario.

modosc:
@rmosolgo by any chance this was magically resolved in 1.10? or does the new parser make solving it easier?

swalkinshaw:
So I think this issue was solved at some point? Here's a script to test out all scenarios: https://gist.github.com/swalkinshaw/ede5be9e5ab8408e4fb88148bc93d768

[...]

PR #1602, opened 2018-06-19
Fix scalar coercion errors in variable values (including defaults) not being handled

obrie:
Currently, if a variable value for a scalar cannot be coerced, it results in an exception getting raised rather than being tracked in the `errors` array of the response.  This same thing happens when a default value for a variable is provided that cannot be coerced.

This fixes that problem by handling `GraphQL::CoercionError` in the same way we see here: https://github.com/rmosolgo/graphql-ruby/blob/bf29cde08e9707100703d0bef9464c92aced65e3/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb#L10-L14

Example scenarios where this problem presents itself are shown in the specs.

This should also fix https://github.com/rmosolgo/graphql-ruby/issues/1269

### DIFF
--- a/lib/graphql/query/variables.rb
+++ b/lib/graphql/query/variables.rb
@@ -33,7 +33,7 @@ module GraphQL
             begin
               validation_result = variable_type.validate_input(provided_value, ctx)
             rescue GraphQL::CoercionError => ex
-              validation_result = Query::InputValidationResult.new
+              validation_result = GraphQL::Query::InputValidationResult.new
               validation_result.add_problem(ex.message)
             end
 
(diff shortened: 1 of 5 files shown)


obrie:
Added specs + squashed

rmosolgo:
Makes sense to me! Thanks for the fix, I'll give it a close review soon.

eapache:
@obrie would this happen to fix or affect https://github.com/rmosolgo/graphql-ruby/issues/1024? They seem fairly related.

obrie:
I agree they're related, but my understanding is that #1024 happens at a different part of the query execution pipeline.  First, the schema will ensure that the incoming types match what the schema defines -- that's where this fix comes in.  The next part after that would then be preparing the properly typed values for resolver methods -- which is where #1024 is discussing.

rmosolgo:
Thanks again for the fix!

PR #1612, opened 2018-06-22
Improve ability to debug errors in development environment

ajkerr:
This pull request borrows some logic from https://github.com/howtographql/graphql-ruby/blob/master/app/controllers/graphql_controller.rb that allows for much easier debugging of errors when using tools such as graphiql.

It basically overrides the default Rails error handling for the development environment, which shows an HTML page by default, and instead returns the stack trace in a json response.

I think these are pretty good defaults to use for scaffolding when first creating a project with this gem.

Before:
21-59-yddnv-bzih1

After:
21-02-ww5b2-8bt2t


rmosolgo:
Very nice! I agree that will make development a lot easier, thanks for sharing this improvement 😊

PR #1613, opened 2018-06-22
Fix lists of abstract type with promises

rmosolgo:
Another bug discovered with `integrated-auth`.

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -123,10 +123,10 @@ module GraphQL
                   err
                 end
               }
-              continue_or_wait(inner_value, arguments, field_ctx)
+              continue_or_wait(inner_value, field_ctx.type, field_ctx)
             }
           else
-            continue_or_wait(raw_value, arguments, field_ctx)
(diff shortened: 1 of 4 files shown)


lib/graphql/execution/execute.rb
```diff
@@ -169,19 +169,22 @@ def continue_resolve_field(raw_value, field_ctx)
...
-                error.ast_node = field_ctx.ast_node
-                error.path = field_ctx.path + [index]
-                query.context.errors.push(error)
+            if !field_type.list?
+              # List type errors are handled above, this is for the case of fields returning an array of errors
+              list_errors = raw_value.each_with_index.select { |value, _| value.is_a?(GraphQL::ExecutionError) }
+              if list_errors.any?
+                list_errors.each do |error, index|
+                  error.ast_node = field_ctx.ast_node
+                  error.path = field_ctx.path + [index]
```
rmosolgo: I think adding `index` here is probably wrong, since this codepath supports returning multiple errors from a single field. But this is how it's always worked so I'm not gonna change it now.

PR #1615, opened 2018-06-23
Fix uninitialized constant GraphQL::Types::ISO8601DateTime

jpawlyn:
This error occurs when setting a field type as GraphQL::Types::ISO8601DateTime

The gem version is 1.8.4


jpawlyn:
Sorry, missed `To use it, require it in your project:.... require "graphql/types/iso_8601_date_time"`

gfx:
+1 for this pull-request because I have also spent this issue 😭 

@rmosolgo What do you think of it?

rmosolgo:
This change is fine with me, but if we require it by default, we should update the docs to remove the instruction to `require` it. I can't reopen this PR but I'll happily merge the diff if it's reopened or if the change is submitted in another PR

PR #1617, opened 2018-06-25
Resolve directive bug when using variable with default value of false

bswinnerton:
👋 Hi there,

This pull request resolves a bug that was originally reported in our site's [public Discourse](https://platform.github.community/t/include-with-default-false-behaves-like-it-received-true/6302). The reporter had mentioned that given a query like:

```graphql
query($showName: Boolean = false) { 
  viewer {
    login @include(if: $showName)
  }
}
```

Their response would include the field that was not to be included:

```json
{
  "data": {
    "viewer": {
      "login": "bswinnerton"
    }
  }
}
```

This ultimately boiled down to this `case` statement:

https://github.com/rmosolgo/graphql-ruby/blob/e0b3e728f38c73529cf68846ac9c2870526a1832/lib/graphql/query/variables.rb#L46

Which, in the query above, `default_value` would have been equal to `false`.

This pull request changes the behavior to specifically check for the presence of `nil`.

### DIFF
--- a/lib/graphql/query/variables.rb
+++ b/lib/graphql/query/variables.rb
@@ -43,7 +43,7 @@ module GraphQL
             elsif value_was_provided
               # Add the variable if a value was provided
               memo[variable_name] = variable_type.coerce_input(provided_value, ctx)
-            elsif default_value
+            elsif default_value != nil
               # Add the variable if it wasn't provided but it has a default value (including `null`)
               memo[variable_name] = GraphQL::Query::LiteralInput.coerce(variable_type, default_value, self)
             end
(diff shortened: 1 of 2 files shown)


bswinnerton:
When I had first started debugging this problem, I added failing tests in the `GraphQL::Directive` spec to help me navigate the codebase. I ultimately resolved the issue in `GraphQL::Variable`, so it'd make sense to move the tests over there. I do that first thing tomorrow 🌆.

rmosolgo:
Nice find! Feel free to move the tests or not, either place makes sense to me.

bswinnerton:
> Nice find! Feel free to move the tests or not, either place makes sense to me.

Awesome! If they make sense in their current location, I'll opt for keeping them there ☺️.

rmosolgo:
🤘 Thanks again!

PR #1618, opened 2018-06-25
Include 'filename' key in static validation error hash

pushrax:
This may not be the best way of accomplishing this, but in our use-case we'd like to be able to pull the filename back out of the validation error, since we're concatenating definitions from multiple documents before running validation.

@rmosolgo 
cc @jackychiu @jpittis

### DIFF
--- a/lib/graphql/static_validation/message.rb
+++ b/lib/graphql/static_validation/message.rb
@@ -33,7 +33,11 @@ module GraphQL
       private
 
       def locations
-        @nodes.map{|node| {"line" => node.line, "column" => node.col}}
+        @nodes.map do |node|
+          h = {"line" => node.line, "column" => node.col}
+          h["filename"] = node.filename if node.filename
+          h
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: Hey, I guess someone is _using_ the filename support 😆  

This approach seems fine to me, I'll share a couple other things that came to me while brainstorming: 

- Setting this option during `GraphQL.parse` (or somewhere else higher-up) and passing it along 
- _Always_ including `filename` when it's present on a `node`  (I'm wondering if there are cases that people don't _want_ to include it, I can't think of any, but there certainly might be.) 

Do you think there's anything to be caned from those approaches? Otherwise I'll merge as-is

rmosolgo:
> make that change

😎  let's do it!

pushrax:
Should be good to go.

rmosolgo:
💰  I'll be sure to mention it in "Breaking changes" with a snippet about how to override this new feature in case anyone doesn't mention it. Thanks!!

PR #1619, opened 2018-06-25
Deprecate 'resolves?' in favour of 'abstract?' in TypeKind

RobertWSaunders:
This is a proposal and implementation to deprecate 'resolves?' in favour of adding 'abstract?' to `TypeKind`. The reason for this proposal is because I found that that the language 'resolves?' to be a bit misleading and 'abstract?' to be more explicit in what the check is actually doing.

I understand that abstract types need to provide a resolve type, but this might not be apparent to everyone at first glance, as it wasn't for me. It can also be easily confused with whether or not a type can actually resolve at all.

### DIFF
--- a/lib/graphql/introspection/type_type.rb
+++ b/lib/graphql/introspection/type_type.rb
@@ -60,7 +60,7 @@ module GraphQL
       end
 
       def possible_types
-        if @object.kind.resolves?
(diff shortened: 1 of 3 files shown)


rmosolgo:
Sounds good to me, that's some very dusty code so thanks for the improvement 😆

Issue #1600, opened 2018-06-19
Support `extend` in Schema.from_definition? 


rmosolgo:
The GraphQL SDL has a concept of "extending" type definitions, for example: 

http://facebook.github.io/graphql/June2018/#sec-Object-Extensions

Should we support this in GraphQL-Ruby? I have heard several folks who use `Schema.from_definition` to build their schemas, would this be helpful?

cjoudrey:
As a first step I'll take a stab at adding support for `extend type` to the lexer and parser.

Tybot204:
Have there been any latest thoughts on implementing this?

I have a system of many microservices, each implementing their own GQL schema. These schemas are stitched together in a gateway service that exposes all of the queries / mutations to clients in one convenient schema. `extend` would help greatly in allowing me to define resolvers that return types from another service's schema!

Example:
```gql
# Defined in service A
type User {
  id: ID!
  name: String!
}

extend type Account {
  user: User!
}

# Defined in service B
type Account {
  id: ID!
  type: String!
}

extend type User {
  accounts: [Account]!
}
```

I'm not super familiar with the inner workings of this gem, but I'll start digging in and seeing what needs to be added to make this happen!

rmosolgo:
Hi! No, as far as I know, there hasn't been more thought in implementing it. Personally, I don't use the `Schema.from_definition` feature, so I don't work on it much 😬. (Although I know plenty of folks use it for their whole schema!)

cjoudrey:
After implementing the lexer/parser pieces of `extend`, I remember taking a look at GraphQL.js to see how `extend` works there. [Reading the implementation](https://github.com/graphql/graphql-js/blob/9a43d477fe5acc9170fb6dcbf2002fdafa903e0a/src/utilities/extendSchema.js#L83-L100) one thing that seemed important is that extending a schema should not alter the existing instance of the schema, but rather return a new schema. I probably won't have time to implement the rest of this in the near future, but I'd happily help review any PR that adds this functionality.

PR #1620, opened 2018-06-26
Add support for `extend` to parser

cjoudrey:
re: #1600

I had fun over the weekend adding support for type extensions to the parser. 😄 

I didn't include the generated lexer and parser as the version of Ragel enforced by our `Rakefile` is not yet available via `brew`. @rmosolgo would you mind committing those to my branch to confirm CI turns green?


lib/graphql/language/parser.y
```diff
@@ -293,6 +297,47 @@ rule
...
+    | interface_type_extension
+    | union_type_extension
+    | enum_type_extension
+    | input_object_type_extension
+
+  scalar_type_extension: EXTEND SCALAR name directives_list { return make_node(:ScalarTypeExtension, name: val[2], directives: val[3], position_source: val[0]) }
+
+  object_type_extension:
+      /* TODO - This first one shouldn't be necessary but parser is getting confused */
+      EXTEND TYPE name implements LCURLY field_definition_list RCURLY { return make_node(:ObjectTypeExtension, name: val[2], interfaces: val[3], directives: [], fields: val[5], position_source: val[0]) }
```
cjoudrey: Something fishy is going on in the parser. I tried debugging (without success) why this one is required. 🤔 

You would think the 2nd line covers this case, but without it, this:

```
extend type Something implements A { field: String }
```

Gets parsed as:

`ObjectTypeExtension` + `OperationDefinition`

Instead of:

`ObjectTypeExtension`

cjoudrey:
> I wonder, would it be possible to add `extend_opt` to the existing parser rules, instead of adding `_extension` rules that match the definition rules?

Here is the current definition for `object_type_definition`:

```racc
object_type_definition:
  TYPE name implements_opt directives_list_opt LCURLY field_definition_list RCURLY {
    return make_node(:ObjectTypeDefinition, name: val[1], interfaces: val[2], directives: val[3], fields: val[5], description: get_description(val[0]), position_source: val[0])
  }
```

Here is the syntax we need to support as per the spec:

screen shot 2018-06-26 at 9 45 06 am

It might be possible to add `extend_opt` to that line, but we'd only be supporting the first version of `extend type ..`. We would still need to add new rules for the other two versions.

Using `extend_opt` also means we'd have to add a conditional to the rule's block in order to return `:ObjectTypeExtension` instead of `:ObjectTypeDefinition`.

I may have misunderstood your suggestion though, let me know if that's the case. Are you mostly worried about code duplication and maintainability of all these rules? While there is some repetition, I was aiming to have a parser that is as close as possible to the spec, that way when the spec gets updated we can apply the changes to our parser quite easily.

rmosolgo:
You understood my question perfectly, and thanks for the answer, now I see the advantage to a bit of duplication!

PR #1623, opened 2018-06-27
fix prev/next page when not using first/last

### LINKED ISSUE #1109: when setting max_page_size on a connection pageInfo hasNextPage, hasPriorPage is wrong if no first: or last: are specified
If I have a max_page_size:10 (and a default_max_page_size:5) set on a connection and do not pass a first: or last: parameter to the connection the pageInfo incorrectly reports that hasNextPage:false hasPriorPage:false. 

**Example issue**
I set the max_page_size: 10 on a connection with a schema max_default_page_size: 5, there are a total of 39 items available on the connection. Without specifying a first: or last: the result is hasNextPage:false hasPriorPage:false on pageInfo. The result is limited to 10 items.

**Works correctly**
If I set the first: parameter to 50 (or any number>=10) it limits the results to 10 and reports pageInfo correctly as hasNextPage:true hasPriorPage:false.

If I set the last: parameter to 50 (or any number >= 10) it limits the results to 10 and reports pageInfo correctly as hasNextPage:false hasPriorPage:true.

If I set the first: parameter to 2 (or any number<10) it limits the results correctly and reports pageInfo correctly as hasNextPage:true hasPriorPage:false.

If I set the last: parameter to 2 (or any number < 10) it limits the results correctly and reports pageInfo correctly as hasNextPage:false hasPriorPage:true.

**Assumption**
There is no default ordering assumed on the connection (first or last), therefor if neither a first or last are sent it is unordered and no way to tell if there is a next or prior page.

**Possible fix** 
set first if max_page_size or default_max_page_size are limiting the output and !first && !last here:
https://github.com/rmosolgo/graphql-ruby/blob/c1c1d1b57a5c1d76c6089106554047b09b162a1c/lib/graphql/relay/relation_connection.rb#L94

**Issue code**
here is where it is setting the pageInfo, if !last and !first it doesn't appear to set it correctly
https://github.com/rmosolgo/graphql-ruby/blob/c1c1d1b57a5c1d76c6089106554047b09b162a1c/lib/graphql/relay/relation_connection.rb#L28

https://github.com/rmosolgo/graphql-ruby/blob/c1c1d1b57a5c1d76c6089106554047b09b162a1c/lib/graphql/relay/relation_connection.rb#L38

modosc:
fixes #1109


modosc:
not sure what's up with the tests - they pass locally, looking into it.

rmosolgo:
Hi, thanks for suggesting this change! Tomorrow I'll take a closer look.

rmosolgo:
Thanks again for pushing this diff. I'm just _not sure_ it's the best change. Here's the relevant bit from the [spec](https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo):

> hasPreviousPage is used to indicate whether more edges exist prior to the set defined by the clients arguments. If the client is paginating with last/before, then the server must return true if prior edges exist, otherwise false. If the client is paginating with first/after, then the client may return true if edges prior to after exist, if it can do so efficiently, otherwise may return false. More formally:
> 
> - `HasPreviousPage(allEdges, before, after, first, last)`:
>    - If `last` is set:
>        - Let `edges` be the result of calling `ApplyCursorsToEdges(allEdges, before, after)`.
>        - If `edges` contains more than `last` elements return `true`, otherwise `false`.
>    - If `after` is set:
>        - If the server can efficiently determine that elements exist prior to `after`, return `true`.
>    - Return `false`. 

I _think_ this case would fall down to the last branch there ("Return `false`"). We don't have `last`, or `after`, right? 

On the one hand, you might say "we should provide a default value for clients to make it 'easier' for them". But on the other hand, this doesn't fit the general ethos of GraphQL. In GraphQL, explicit is preferred over implicit, but this behavior introduces an undocumented default value. (GraphQL has argument default values, maybe that should be used instead?) A client might depend on this value, but since it's not part of the schema's formal structure, it could be changed at any time, altering the behavior of the client. 

What about encouraging clients to depend on stable, spec-defined behavior, for example: 

- Don't provide a default value; require clients to send pagination arguments if they want to paginate. It's explicit, but it's also very stable. 
- Provide a default value (by overriding the built-in argument with `argument(:first, ..., default_value: ...)`) which is part of the schema's introspection system. I think it's considered a "breaking change" to modify default values, so that gives the client some confidence in stability. 

What do you think?

modosc:
sure - i think explicitly setting the default value is the best way to go - it's basically what i'm trying to emulate here (and what the code already does w/r/t limiting the return set).

yskkin:
> I think this case would fall down to the last branch there ("Return false"). We don't have last, or after, right?

This problem also happens if only `after` is set and `max_page_size` is so small that trigger pagination.

PR #1624, opened 2018-06-27
Use the ID type for Relay connection cursors

crismali:
I came across this error today: 

```
GraphQLParser: Variable `$cursor` was defined as type `ID`, but used in a location that expects type `String`.
```

This should solve it by using the graphql ID type instead of strings for cursors on Relay connections.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -243,8 +243,8 @@ module GraphQL
           # TODO: this could be a bit weird, because these fields won't be present
           # after initialization, only in the `to_graphql` response.
           # This calculation _could_ be moved up if need be.
-          argument :after, "String", "Returns the elements in the list that come after the specified cursor.", required: false
-          argument :before, "String", "Returns the elements in the list that come before the specified cursor.", required: false
+          argument :after, "ID", "Returns the elements in the list that come after the specified cursor.", required: false
+          argument :before, "ID", "Returns the elements in the list that come before the specified cursor.", required: false
           argument :first, "Int", "Returns the first _n_ elements from the list.", required: false
           argument :last, "Int", "Returns the last _n_ elements from the list.", required: false
         end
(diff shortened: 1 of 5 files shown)


rmosolgo:
Thanks for the suggestion and the detail, but I don't think it's a good change for a couple of reasons: 

- The [Relay spec](https://facebook.github.io/relay/graphql/connections.htm#sec-Cursor) suggests using a String (although ID _might_ be legal there), and I _think_ Relay Classic _requires_ a string, although those docs don't seem to be online anymore 
- Changing this code in the gem will cause all users of the gem to have a breaking change in their schemas. Any queries written using `$after: String` will be broken because the type is wrong. We can't break people's code that way! 

But, if you want this for your app, I think you could implement it by adding to your custom `GraphQL::Schema::Field` subclass: 

```ruby
      def to_graphql 
        field_defn = super 
        # Override the connection arguments added by default
        if @connection
          # Add the arguments to this field: 
          argument :after, "ID", "Returns the elements in the list that come after the specified cursor.", required: false
          argument :before, "ID", "Returns the elements in the list that come before the specified cursor.", required: false
          argument :first, "Int", "Returns the first _n_ elements from the list.", required: false
          argument :last, "Int", "Returns the last _n_ elements from the list.", required: false
          # Since we added new arguments to this field,
          # we should also rebuild the arguments for the derived `GraphQL::Field`:
          arguments.each do |name, defn|
            arg_graphql = defn.to_graphql
            field_defn.arguments[arg_graphql.name] = arg_graphql
          end 
        end

        field_defn
      end 
```

What do you think about using that locally instead?

crismali:
Ah I see. The [example for the Relay Pagination Container](https://facebook.github.io/relay/docs/en/pagination-container.html#pagination-example) uses cursor, so it seems like there's some confusion on the Relay side of things. Maybe there should be a way to specify a "modern" connection that uses `ID`? Just spitballing

rmosolgo:
This is the first I've heard of using `ID` for cursors, and it sounds like the Relay spec allows any String-serialized type. So I'm happy to support `String` by default. If you want to make it configurable, I'd be happy to review a PR for that!

PR #1626, opened 2018-06-28
Add support for extensions in `CoercionError`s

### LINKED ISSUE #1621: CoercionError does not allow for extra options
I have been using the `options` accessor on `GraphQL::ExecutionError` to add additional information to errors so that they are consumable by humans and machines. I was also hoping to do this in the case of `GraphQL::CoercionError`s that occur during validation but the `CoercionError` does not allow for this. Is there a specific reason why? If not I would be happy to put together a PR.

aschepis:
This change adds an `extensions` attribute to `CoercionError` in order
to make it consistent with the [GraphQL
Specification](http://facebook.github.io/graphql/June2018/#example-fce18)

Fixes #1621

### DIFF
--- a/lib/graphql/coercion_error.rb
+++ b/lib/graphql/coercion_error.rb
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 module GraphQL
   class CoercionError < GraphQL::Error
+    # @return [Hash] Optional custom data for error objects which will be added
+    # under the `extensions` key.
+    attr_accessor :extensions
+
+    def initialize(message, extensions: nil)
+      @extensions = extensions
+      super(message)
+    end
   end
 end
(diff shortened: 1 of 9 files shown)


aschepis:
@rmosolgo i'm not completely confident in this change. There are a few paths where `CoercionError` is rescued and I attempted to pass the extensions on in all cases. Really hoping to get some feedback from you (and others) as to whether there is a cleaner way to do this.

aschepis:
I also opted to use `extensions` instead of `options` in order to match the GraphQL spec

rmosolgo:
Wow, yeah, there _is_ a lot more going on here than I expected 😅  I'll take a close look soon and see if I can make any other suggestions.

aschepis:
@rmosolgo did you ever have a chance to look at this?

Issue #1385, opened 2018-04-02
[1.8] Better handling of boolean attributes with "?" suffix


jturkel:
Readers for boolean attributes in Ruby usually end with the `?` suffix. Ideally the new 1.8 class based API would handle this without an explicit configuration of the `method` e.g.

```ruby
module Types
   class MyResultType < Types::BaseObject
     # Ideally we wouldn't have to specify the method here
      field :valid, Boolean, method: :valid?
    end

  class QueryType < Types::BaseObject
    field :my_result, MyResultType

    def my_result
      OpenStruct.new(valid?: true)
    end 
end
```
Currently this results in the following error:
```
RuntimeError (Failed to implement MyResult.valid, tried:

- `Types::QueryType::MyResultType#valid`, which did not exist
- `OpenStruct#valid`, which did not exist
- Looking up hash key `"valid"` on `#<Types::QueryType::MyResultType:0x00007f9ba55537a8>`, but it wasn't a Hash
)
```
I can think of two potential ways to solve this:

1) First try to resolve `<field name>?` for booleans and then fallback to `<field name>`
2) Borrowing from [active_model_serializers](https://github.com/rails-api/active_model_serializers) allow the field name to be specified with the `?` suffix (e.g. `field :valid?, Boolean`) but strip of the suffix for the JSON member name.

rmosolgo:
That's interesting, I noticed a lot of similar duplication in GitHub, but we have a slightly different convention: our Boolean fields (usually!) have a third-person verb, like `isValid` or `viewerCanUpdate`. I wonder if we should pick one way and provide first-class support for it 🤔

jturkel:
We got our conventions from this [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#bool-methods-qmark) which seems to be pretty widely used and has good enforcement via [Rubocop](https://github.com/bbatsov/rubocop). That said I haven't seen consistent naming conventions for boolean fields in GraphQL so it's possible that just dropping the `?` will prove to be insufficient for lots of scenarios.

rmosolgo:
I'm going to close this because I'm fine with the current behavior and don't have plans to change it. However, I'm open to it -- if someone wants to submit a PR which offers this behavior in a way that doesn't introduce a breaking a change (for example, opt-in), I'd be happy to take a look!

klyonrad:
[...]

PR #1633, opened 2018-06-28
#1385 - better handling of boolean attributes

eddytan:
* This changes allows boolean field name with `?` suffix.
* stripped `?` suffix for the json member name

it is using the second solution of issue #1385

### DIFF
--- a/lib/graphql/name_validator.rb
+++ b/lib/graphql/name_validator.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module GraphQL
   class NameValidator
-    VALID_NAME_REGEX = /^[_a-zA-Z][_a-zA-Z0-9]*$/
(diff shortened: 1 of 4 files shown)


lib/graphql/name_validator.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module GraphQL
   class NameValidator
-    VALID_NAME_REGEX = /^[_a-zA-Z][_a-zA-Z0-9]*$/
+    VALID_NAME_REGEX = /^[_a-zA-Z][_a-zA-Z0-9]*\??$/
```
rmosolgo: I'm surprised this was necessary,  I thought it only tested the finally-built GraphQL schema. Since the `?` is removed earlier on, I thought this could remain unchanged. Does it work without changing this Regexp? 

This pattern is used in lots of places in the gem, and I don't think we should add `?` to it. (For example, I think this will allow enum values to contain `?`, which is not valid GraphQL.)

PR #1635, opened 2018-06-29
Allow definition_methods defined in interfaces to be applied onto classes

joostverdoorn:
This PR allows methods defined in `definition_methods do ... end` blocks within interfaces to be used within classes implementing these interfaces. The change is simple enough, but I'm not entirely familiar with this project's codebase, so I may be missing some cases here. However, the changes proposed here do enable my own use case without hiccups.


rmosolgo:
Looks good to me! Thanks for taking this on 👏 !

PR #1636, opened 2018-06-29
update pro authorization docs for class-based

rmosolgo:
This is a bit tricky -- I don't think the `unauthorized_object` or `unauthorized_fields` handlers work quite right yet, but everything else should be 👌 

If you want to update to class-based and you need to migrate those hooks, you can attach them manually in your schema this way: 

```ruby 
  def self.to_graphql 
    schema_defn = super 
    # A compatibility patch for these hooks in a class-based schema:
    schema_defn.redefine {
      # Add your handlers here: 
      unauthorized_object(...) 
      unauthorized_fields(...)
    }
  end 
end 
```

### DIFF
--- a/guides/pro/authorization.md
+++ b/guides/pro/authorization.md
@@ -20,7 +20,7 @@ Fields and types can be [authorized at runtime](#runtime-authorization), [reject
 To use authorization, specify an authorization strategy in your schema:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


rmosolgo:
Instead, new docs have been added for http://graphql-ruby.org/authorization/pundit_integration.html and http://graphql-ruby.org/authorization/can_can_integration.html

PR #1640, opened 2018-06-30
Prevent zsh error

henk23:
zsh: no matches found: comments:[Comment]

### DIFF
--- a/guides/getting_started.md
+++ b/guides/getting_started.md
@@ -30,7 +30,7 @@ On Rails, you can get started with a few [GraphQL generators](https://rmosolgo.g
 # Add graphql-ruby boilerplate and mount graphiql in development
 rails g graphql:install
 # Make your first object type
-rails g graphql:object Post title:String rating:Int comments:[Comment]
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! Thanks for making this suggestion. 

I want to keep these docs as simple as possible. Is this only required for zsh? If so, how about adding it later in a parenthetical, for example: 

> (Note: for `zsh`, you might need to wrap some of it in single quotes, for example: `'comments:[Comment]'`)

That way, it won't distract from the main point of this guide, but it will still help people use the tool correctly. What do you think?

henk23:
Hey. Sounds good to me. I just [looked into it a little more](http://mikeballou.com/blog/2011/07/18/zsh-and-rake-parameters/) and it seems to be zsh-specific, because square brackets have a special meaning there, related to globbing. You can also just prefix the whole line with "noglob" like this: 

    noglob rails g graphql:object Post title:String rating:Int comments:[Comment]

and it will work in zsh. Maybe that's a more suitable workaround to put in the note.

rmosolgo:
Feel free to reopen this with one of the clarifying comments mentioned above.

Issue #1649, opened 2018-07-05
Introspection broken


joostverdoorn:
Sorry for this, but my recently merged PR (https://github.com/rmosolgo/graphql-ruby/pull/1635) has broken (at least) introspection queries, probably because methods defined on `GraphQL::Schema::Interface::DefinitionMethods` overriding some of `GraphQL::Schema::Object`'s methods. I'm looking into a fix right now, but we might have to revert my PR.

joostverdoorn:
See https://github.com/rmosolgo/graphql-ruby/pull/1650

rmosolgo:
Could you help me understand what's broken? For example, is something not behaving correctly for you? Or did you get a Ruby exception? (If so, could you share the error message and backtrace?)

joostverdoorn:
Well, one issue is that because of my change, any `GraphQL::Schema::Object` subclass implementing an interface will inherit the `kind` method from `GraphQL::Schema::Interface::DefinitionMethods`, overriding the original one, resulting in objects being recognized as interfaces. This should clearly not happen. The same might be the case for other methods defined on `GraphQL::Schema::Object`.

rmosolgo:
It sounds like you found a bug, right? How about including a test in your PR so we can avoid re-introducing this bug in the future? For example, in `spec/graphql/schema/object_spec.rb`, you could create an interface module and an object class, and add the interface to that object, the assert that `.kind` returns the correct value.

joostverdoorn:
I've added two specs to the PR, which test whether `GraphQL::Schema::Object`'s methods are overridden when implementing `GraphQL::Schema::Interface` or any of its descendants :).

rmosolgo:
That's great! Thanks!

PR #1650, opened 2018-07-05
Exclude GraphQL::Schema::Interface::DefinitionMethods from extend

joostverdoorn:
See https://github.com/rmosolgo/graphql-ruby/issues/1649. Extending object classes with  `GraphQL::Schema::Interface::DefinitionMethods` will either override methods or include methods which should not be included. This PR ensures `GraphQL::Schema::Interface` is skipped.


rmosolgo:

PR #1652, opened 2018-07-05
Update input_object.rb

phaedryx:
it would be handy for short-circuiting in some of my evaluations.


rmosolgo:
👌  Thanks for the improvement

PR #1654, opened 2018-07-06
[docs] Change one field name in an example

Winbobob:
This PR updates the documentation example for defining a mutation.
The `resolve` method uses the `errors` variable. However, the field is declared as `error_messages` above.


rmosolgo:
Thanks for the fix!

PR #1660, opened 2018-07-10
Require GraphQL::Types::ISO8601DateTime by default

jpawlyn:
I amended the original commit so couldn't reopen the original PR -https://github.com/rmosolgo/graphql-ruby/pull/1615


rmosolgo:
Thanks for taking care of it, and sorry for the confusion!

PR #1663, opened 2018-07-10
Support grouping by "endpoint" with skylight instrumentation

rmosolgo:
This will cause queries with the same name to be grouped together.

~~TODO: make this opt-in?~~, done, use `set_endpoint_name: true` at schema-level, and/or `context[:set_skylight_endpoint_name]` at query-level.

### DIFF
--- a/lib/graphql/tracing/skylight_tracing.rb
+++ b/lib/graphql/tracing/skylight_tracing.rb
@@ -14,15 +14,31 @@ module GraphQL
         "execute_query_lazy" => "graphql.execute",
       }
 
+      # @param set_endpoint_name [Boolean] If true, the GraphQL operation name will be used as the endpoint name.
+      #   This is not advised if you run more than one query per HTTP request, for example, with `graphql-client` or multiplexing.
+      #   It can also be specified per-query with `context[:set_skylight_endpoint_name]`.
+      def initialize(set_endpoint_name: false)
+        @set_endpoint_name = set_endpoint_name
+        super
+      end
+
       def platform_trace(platform_key, key, data)
-        if (query = data[:query])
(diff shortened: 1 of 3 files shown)


lib/graphql/tracing/skylight_tracing.rb
```diff
@@ -18,6 +18,12 @@ def platform_trace(platform_key, key, data)
         if (query = data[:query])
           title = query.selected_operation_name || "<anonymous>"
           category = platform_key
+          # Assign the endpoint so that queries will be grouped
+          current_instance = Skylight::Instrumenter.instance
```
timz1290: This doesn't work with the latest version of the skylight gem (2.0.2). Instead, do something like this:

```
inst = Skylight.instrumenter
current_trace = inst ? inst.current_trace : nil
if current_trace
  endpoint = "GraphQL/#{query.operation_type}.#{title}"
  current_trace.endpoint = endpoint
end
```

lib/graphql/tracing/skylight_tracing.rb
```diff
@@ -18,6 +18,12 @@ def platform_trace(platform_key, key, data)
...
+          if current_instance
+            endpoint = "GraphQL/#{query.operation_type}.#{title}"
```
timz1290: There is no #operation_type method on query

rmosolgo:
Thanks for taking a look @timz1290 ! I took a page out of `NewRelicTracing`'s book and added a proper test with stubs for `Skylight`. If it looks alright to you I'll get it in master and release it with 1.8.6!

rmosolgo:
woops, forgot to add my new test files 😊

PR #1667, opened 2018-07-13
Remove duplicated object_class method

RobertWSaunders:
Noticed this when reading through the code, opted to keep the newer version of the method but there is no difference other than the documentation. I'm not sure if there is a purpose in having them both in there, thoughts?


rmosolgo:
Huh, I wonder how that got there 😊  Probably a bad merge or something. Thanks for the fix. 

By the way, happy reading 😬

rmosolgo:
(Test failure looks like a flake)

Issue #1045, opened 2017-10-20
Subscriptions with action cable not working as expected.


lifeiscontent:
Hey there!

First off, thanks for the great graphql lib in ruby!

I'm trying to setup graphql subscriptions in my app but when I try to create a subscription I get an error.

here's my log:

```
Processing by GraphqlController#execute as */*
  Parameters: {"query"=>"subscription {\n  message {\n    body\n  }\n}", "variables"=>nil, "graphql"=>{"query"=>"subscription {\n  message {\n    body\n  }\n}", "variables"=>nil}}
GraphqlController#execute
Completed 500 Internal Server Error in 83ms (ActiveRecord: 0.0ms)



NoMethodError (undefined method `stream_from' for nil:NilClass):

app/controllers/graphql_controller.rb:13:in `execute'
```

from what I can tell, context is expecting a key called channel and having it be an instance of a GraphqlChannel. But I'm unsure how to pass that in, or even if that should be a requirement of the controller.

here's my code:

graphql_controller.rb
``` ruby
# frozen_string_literal: true

class GraphqlController < ApplicationController
  def execute
    variables = ensure_hash(params[:variables])
    query = params[:query]
    operation_name = params[:operationName]
    context = {
      # Query context goes here, for example:
      # current_user: current_user,
    }
    result = AppSchema.execute(query, variables: variables, context: context, operation_name: operation_name)
    render json: result
  end

  private

    # Handle form data, JSON body, or a blank value
    def ensure_hash(ambiguous_param)
      case ambiguous_param
      when String
        if ambiguous_param.present?
          ensure_hash(JSON.parse(ambiguous_param))
        else
          {}
        end
      when Hash, ActionController::Parameters
        ambiguous_param
      when nil
        {}
      else
        raise ArgumentError, "Unexpected parameter: #{ambiguous_param}"
      end
    end
end
```

graphql_channel.rb
``` ruby
# frozen_string_literal: true

class GraphqlChannel < ApplicationCable::Channel
  def subscribed
    puts 'GraphqlChannel#subscribed'
    @subscription_ids = []
  end

  def execute(data)
    puts 'GraphqlChannel#execute'
    query = data['query']
    variables = ensure_hash(data['variables'])
    operation_name = data['operationName']
    context = {
      # current_user: current_user,
      # Make sure the channel is in the context
      channel: self,
    }

    result = MySchema.execute(query: query,
      context: context,
      variables: variables,
      operation_name: operation_name)

[...]

PR #1675, opened 2018-07-17
improve error messages in ActionCableSubscriptions

gfx:
`MySchema.execute("subscription { ... }")` does not work on Rails console / grahpqil, as repoted in  https://github.com/rmosolgo/graphql-ruby/issues/1045; it is the spec and I have no idea to fix it.

However, the error message, `NoMethodError (undefined method `stream_from' for nil:NilClass)` can be more understandable: `KeyError (key not found: :channel)`, which suggests the `context` requires `:channel`.

In fact, use of `Hash#fetch` is always better if the key must exist.


rmosolgo:
Make sense to me, thanks for the improvement!

gfx:
Thanks! And thanks again for the great library 🎉

PR #1682, opened 2018-07-18
Remove delegate for `FieldResolutionContext#schema`

splattael:
It's already cached via an ivar in the initializer.

Bonus: A tiny improvement in performance.

Refs 9a36a6fd7a26b7c01c9c556529caa1f0b2c878cc


rmosolgo:
Good find, thanks!

PR #1685, opened 2018-07-18
Fix require in tracing.rb

thejchap:
Minor issue with an easy workaround, just makes slightly easier for the user


rmosolgo:
Thanks!

PR #1686, opened 2018-07-18
Fix Interface inheritance chain

### LINKED ISSUE #1674: [1.8] Interfaces and module inheritance
We have a bunch of modules that were included in our versions of `Interface` and `Object`. All of these modules are "concerns" (`extend ActiveSupport::Concern`). It seems that it's not possible to include (in any way) a module that's a concern and actually have the methods included.

Note: while this (potential) issue was discovered from our unusual use case of moving from a class-based Interface to the new Module-based one, this might apply to anyone in the future.

Here's a branch with some new tests to see examples: https://github.com/rmosolgo/graphql-ruby/compare/master...swalkinshaw:interface-module-includes?expand=1

Is this expected? Is it a bug? Or should this limitation be documented at least? My solution is to make the modules "plain" again (no need for the concern/`ClassMethods` since that's all that's in them).

Here's what *does* work:

```ruby
module DefaultFieldDescription
  def field(*args, **kwargs)
    kwargs[:description] += ' plus DefaultFieldDescription'
    super
  end
end

module BaseInterface

  definition_methods do
    include DefaultFieldDescription

    def field(*args, **kwargs)
      kwargs[:description] = 'from interface'
      super
    end
  end
end
```

Is including a module within the `definition_methods` block the best/intended solution?

swalkinshaw:
Currently an interface's `definition_methods` will not take precedence over another interface it includes which is the opposite of Ruby's normal object model.

This simply moves extending the interface's own `DefinitionMethods` *after* the base `Schema::Interface::DefinitionMethods` and any other interfaces it already includes.

The tests include a test which previously failed on master/1.8.5.

This solution seems slightly naive to me, so I'm not 100% confident it won't have any other side effects; however, all tests are passing.

edit: this would mostly close #1674.

### DIFF
--- a/lib/graphql/schema/interface.rb
+++ b/lib/graphql/schema/interface.rb
@@ -43,6 +43,14 @@ module GraphQL
             # In this case, it's been included into another interface.
             # This is how interface inheritance is implemented
 
+            # We need this before we can call `own_interfaces`
+            child_class.extend(Schema::Interface::DefinitionMethods)
+
+            child_class.own_interfaces << self
+            child_class.interfaces.each do |interface_defn|
+              child_class.extend(interface_defn::DefinitionMethods)
+            end
+
             # Use an instance variable to tell whether it's been included previously or not;
             # You can't use constant detection because constants are brought into scope
             # by `include`, which has already happened at this point.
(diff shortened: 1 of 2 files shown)


rmosolgo:
I agree this is the correct behavior, I think it was just an oversight during the initial development! Thanks for sharing the fix.

PR #1690, opened 2018-07-20
Conditional Analyzers

xuorig:
This PR adds an optional `analyze?` method to the analyzer API. The method gets the `query` as an argument and returns true or false to run the analyzer or not. This can be useful for example to only run analyzers (which can currently be quite costly)  only when needed.

We've been running a very similar API at GitHub and has been working pretty well.

It's a bit harder to have this condition apply to built in analyzers, maybe these can offer an additional hook? Another solution could be to allow overriding something similar to `build_analyzers` which would let users return the list of analyzers they want?

### DIFF
--- a/lib/graphql/query/validation_pipeline.rb
+++ b/lib/graphql/query/validation_pipeline.rb
@@ -16,7 +16,6 @@ module GraphQL
     class ValidationPipeline
       def initialize(query:, validate:, parse_error:, operation_name_error:, max_depth:, max_complexity:)
         @validation_errors = []
-        @analysis_errors = []
         @internal_representation = nil
         @validate = validate
         @parse_error = parse_error
@@ -33,12 +32,6 @@ module GraphQL
       def valid?
         ensure_has_validated
         @valid
-      end
-
-      # @return [Array<GraphQL::AnalysisError>] Errors for this particular query run (eg, exceeds max complexity)
-      def analysis_errors
-        ensure_has_validated
-        @analysis_errors
       end
 
       # @return [Array<GraphQL::StaticValidation::Message>] Static validation errors for the query string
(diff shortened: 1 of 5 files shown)


eapache:
🤔 can't analyzers just do this themselves by returning early?

xuorig:
Depends where @eapache? If you're talking about doing it in `#call`, it still ends up being called a lot of times which is not ideal. And returning early in `initial_value` doesn't help in that case.

It feels a bit wrong to just return early for every single node in the query, but I may have missed a simpler way to do it?

lib/graphql/query/validation_pipeline.rb
```diff
@@ -16,7 +16,6 @@ class Query
     class ValidationPipeline
       def initialize(query:, validate:, parse_error:, operation_name_error:, max_depth:, max_complexity:)
         @validation_errors = []
-        @analysis_errors = []
```
xuorig: This is actually not used anymore, since the analysis error logic is handled in `analyze_query`!

eapache:
No, I think I was just forgetting the fact that there were multiple entrypoints and calls to `#call`.

lib/graphql/query/validation_pipeline.rb
```diff
@@ -86,24 +79,23 @@ def ensure_has_validated
           end
 
           if @validation_errors.none?
-            @query_analyzers = build_analyzers(@schema, @max_depth, @max_complexity)
-            # if query_analyzers.any?
-            #   analysis_results = GraphQL::Analysis.analyze_query(@query, query_analyzers)
-            #   @analysis_errors = analysis_results
-            #     .flatten # accept n-dimensional array
-            #     .select { |r| r.is_a?(GraphQL::AnalysisError) }
-            # end
```
rmosolgo: lol, what the heck was all this commented-out code ... I guess it was replaced by the same stuff you mentioned above 😅

PR #1696, opened 2018-07-23
Validation: Don't traverse irep if no handlers are registered

xuorig:
While investigating performance on validation and analysis on queries with `validate: false`, I noticed a pretty important part was spent in: https://github.com/xuorig/graphql-ruby/blame/master/lib/graphql/internal_representation/visit.rb#L7-L21

What I realized is that even if `validate` is `false`, the way `Visit` works is that it will traverse the whole irep and call these "post-rewrite" validators. The way it was setup, even though there were no handlers to run, we would still pay the cost for traversal.

Super simple fix that I think will improve very large queries by quite a bit: don't run the visit when no handlers are registered.

### DIFF
--- a/lib/graphql/internal_representation/visit.rb
+++ b/lib/graphql/internal_representation/visit.rb
@@ -5,6 +5,7 @@ module GraphQL
     module Visit
       module_function
       def visit_each_node(operations, handlers)
+        return if handlers.empty?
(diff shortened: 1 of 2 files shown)


rmosolgo:
💰 ! You're finding some goodies in here!

Issue #1591, opened 2018-06-15
Reusing types in root


mlh758:
I'm fairly new to GraphQL so this could very well be me messing something basic up here.

I used the Rails generator to install the GraphQL gem and get the basic types set up.

Here's an example of a type set up under `graphql/types`:
```
class Types::PageType < Types::BaseObject
  field :name, String, null: false
  field :category, String, null: false
end
```

Then in `query_type.rb`
```
field :page, PageType, null: true do
  description 'A page!'
  argument :name, String, required: true
end

field :pages [PageType], null: true do
  description 'A list of configured pages'
  argument :category, String, required: true
end

...methods to retrieve those
```

However, if I try to query the API with this setup I get an error like:
`NameError (uninitialized constant Types::QueryType::PageType):`

This error goes away if I remove either of the fields above. I can also add queries for other types in the file.

Am I missing something obvious here? If I declare a `PagesType` and give it one field with an array of Pages and use that `PagesType` that also fixes the issue but that feels cludgy and articles like [this](https://hackernoon.com/advanced-querying-with-graphql-and-express-8cf2fd05f5ea) make me think I should be able to have a singular and a plural version of the query exposed.

rmosolgo:
Hi! sorry about the bug! 

> This error goes away if I remove either of the fields above. 

How interesting. So, if _both_ fields are present, you see that bug, but if one of the fields is commented out, then it works ok? 

Could you please share the full error message and backtrace? Readers can help you debug if you share the context that you have 😉

mlh758:
Unfortunately, that's where it gets weird for me:
```
LoadError (Unable to autoload constant Types::MutationType, expected<app_path>/app/graphql/types/mutation_type.rb to define it):
```

rmosolgo:
Is that the entire stack trace?

mlh758:
I updated my comment with more of the trace, our logging goes to a remote server and traces get truncated after a few lines. I'll switch my logging around and try to get a more thorough one.

mlh758:
Line 12 is where the second declaration using that type starts. `field :pages [PageType], null: true do` in my example above.

hidde-jan:
I’ve noticed error messages are rather iffy sometimes. Open a rails console (`bin/rails c`) and type `<YourAppName>Schema`. The real error should present itself. It’s got something to do with the rails development server and auto loading.

jabawack81:
I had the same exact problem and using the old way for the types fixed it for me:

```
[...]

PR #1699, opened 2018-07-24
Nest under common namespace for better autoloading availability

bary822:
To avoid confusion represented in #1591.

In Rails application, constant is resolved based on certain mechanism as
described in the [guide](https://guides.rubyonrails.org/v5.2/autoloading_and_reloading_constants.html#namespaces).
Essentially:

```
class Animal
end

class Animal::Dog
end

class Animal::Human
  Dog # => NameError
end
```
This fails to lookup constant `Dog`

Whereas:

```
module Animal
  class Dog
  end
end

module Animal
  class Human
    Dog # => Animal::Dog
  end
end
```
This works fine.

### DIFF
--- a/lib/generators/graphql/mutation_generator.rb
+++ b/lib/generators/graphql/mutation_generator.rb
@@ -38,7 +38,7 @@ module Graphql
         sentinel = /class .*MutationType\s*<\s*[^\s]+?\n/m
         in_root do
           gsub_file "#{options[:directory]}/types/mutation_type.rb", /  \# TODO\: Add Mutations as fields\s*\n/m, ""
-          inject_into_file "#{options[:directory]}/types/mutation_type.rb", "  field :#{field_name}, mutation: Mutations::#{mutation_name}\n", after: sentinel, verbose: false, force: false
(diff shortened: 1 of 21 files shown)


rmosolgo:
This change sounds good to me, thanks for the suggestion! How about updating the other generators, too? (I think Interface and Input could use an update, right?)

bary822:
@rmosolgo 
Anything I can help to merge this?

rmosolgo:
Yeah, I've been pondering about the _other_ generators. I think we should be consistent: _either_ recommend nested syntax (`module Types\n class X...`) _or_ `::` syntax (`module Types::X` ), but we shouldn't mix them.  Lots of folks will use this generated code as a guideline for custom code later on, so I want to give a clear message. How about migrating the rest of the generators? Is there a downside to it?

bary822:
> I think we should be consistent ... we shouldn't mix them.

Sounds good to me. I was actually wondering if I should do so or not.
I'll make changes on rest of generators.

bary822:
@rmosolgo 
Done for rest of the generators, thanks for the suggestion!

Let me know if there's anything I can do further.

rmosolgo:
Thanks for your work! While reviewing, I noticed a few unrelated issues to address in the generators, so I pushed a commit of my own. 🎉 !

PR #1704, opened 2018-07-25
memoize field instances in default resolve

xuorig:
Another small perf improvement. For a large query on a large schema, we spend quite a bit of time computing `#fields` since it's not memoized. We have a few options:

  - Memoize in the resolve method (This PR). We still have to compute fields once at resolve time for every object.

  - Memoize at boot time, that would be the most efficient, but it has implications, or we have to do cache invalidation on ancestor add + field add.

  - I tried to get the current field def from `context` instead, using `irep_node.definition.metadata[:type_class]`, but somehow this was called when irep_node was the query root, which didnt work 🤔 

Thoughts?

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -119,7 +119,8 @@ ERR
           end
           default_resolve_module.module_eval <<-RUBY, __FILE__, __LINE__ + 1
             def #{method_name}(**args)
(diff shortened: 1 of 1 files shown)


rmosolgo:
This is OK with me, and if I understand right, we'll see a payoff when _many fields_ are request on the _same object_, is that right?

The other possibility we mentioned was adding a field lookup method which _didn't_ merge hashes, but instead, traversed the ancestry chain looking for a match. Did you explore that approach?

rmosolgo:
I'm definitely interested in optimizing the new class-based types, but I'd like to wait until the new runtime is merged before doing it. That way, we'll have an accurate eye on the bottlenecks for the new runtime.

PR #1705, opened 2018-07-25
compute possible types/ interface implementers at boot

xuorig:
We can precompute this map during the traversal instead, saving that computation at runtime the first time.


xuorig:
@rmosolgo after re-profiling some queries I think this might be worth to do!

rmosolgo:
Here are my concerns with this: 

- In general, it'd be nice to do less caching this time. It really complicated, all the different caches at different levels. 
- When is _boot_? Since we don't have a way to respond to the `end` at the end of a class definition, we don't know when the class is "done" (of course, even the `end` doesn't mean it's done). Currently, all the cache-building is deferred until the first query, but this has its own issues, see: https://github.com/rmosolgo/graphql-ruby/issues/1505#issuecomment-408209171
- Will this play nicely with Rails autoloading? I can _imagine_ a scenario where a type definition is reloaded, but the schema caches are not busted. However, this problem might be solved because maybe Rails knows which classes loaded which others, and so it reloads the Schema when types change. 

Here's an alternative suggestion: we already have a _per-query_ cache in the Warden. Could we leverage that better? That gives us a smaller unit of caching and a clearer lifecycle for the cache (but, it's not as efficient, since every query starts over). What do you think, is there somewhere we could draw on that more?

xuorig:
Unfortunately it's already kinda being cached on a per query basis right now, but the cost of going through *every* type in the schema to find the implementers to an interface is super high everytime, even though that result can be reused.

Another approach to solve this could be to actually save implementers on the interface whenever it is implemented. That would save the loop over every item in the schema.

I still dream of a immutable schema 😆

rmosolgo:
Immutable _when_ is the thing. 

Maybe we could add an opt-in thing like `Schema.freeze_schema` that builds caches, and it can be the last call in the Schema definition.

xuorig:
> Maybe we could add an opt-in thing like Schema.freeze_schema that builds caches, and it can be the last call in the Schema definition.

That's really interesting. Would you see that kind of a transition towards this future? I guess my question is: would we want to migrate almost everything in `rebuild_artifacts` to `freeze_schema`, or just the new things to `freeze_schema`?

I'm betting a lot of users rely on the fact calling methods on schema builds and caches artifacts already, so we probably have to slowly add things to `freeze_schema`, and not move everything to there yet :/

rmosolgo:
Here's a future I can get excited about: 

Let's say a schema has some caches about the GraphQL system. For example, `Schema.types[type_name]` hits the cache and returns the definition of a type named `type_name`. 

This cache is (_these_ cache_s_ are?) a function of a few things: 

- The root types (query, mutation and subscription) 
- orphan types 

So, what if the caches were built in an additive way? For example, the call to `Schema.query` in `query MyApp::Types::QueryType` added to the cache(s)? Similar for `mutation`, `subscription`, and `orphan_types`. I think that would solve a few issues: 

- We don't need an explicit `freeze_schema` call, no API burden 
- We can still have caches, which we need to make things like introspection reasonably fast 

It will be harder to implement, since the cache is mutable, but I think it will tick all the right boxes from an end-user point of view. 

Anyhow, if you're interested in exploring caching some more, maybe that's an avenue to go down!

PR #1709, opened 2018-07-26
Fix method inheritance for interfaces

swalkinshaw:
When interfaces include other interfaces, their own interfaces are iterated over to extend each of their `DefinitionMethods`. This was happening in the reverse order than the usual Ruby object model.

Now `interfaces` is reversed when iterating to properly give precedence to interfaces higher in the chain.

This is another part of fixing behaviour described in #1674.


rmosolgo:
Thanks for the fix!

PR #1712, opened 2018-07-27
Yield field for definition blocks with arity of one

RobertWSaunders:
## Proposal

To improve backwards compatibility I want to propose that we provide the option to yield the field to the field definition block if there is a block provided. 

## How?

Utilize `instance_exec` in place of `instance_eval` if the definition block has an arity of one, otherwise we will default to `instance_eval` which does not yield the field.

## Why?

Improves compatibility with older versions.

Added a test to cover the case of yield the field.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -152,7 +152,11 @@ module GraphQL
         @subscription_scope = subscription_scope
 
         if definition_block
-          instance_eval(&definition_block)
+          if definition_block.arity == 1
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: This is good by me! GraphQL::Schema::Argument does the same thing I think, should we also update that?

RobertWSaunders:
Yeah, I can put up a separate PR up for that or if you prefer add it here?

rmosolgo:
🤷‍♂️  separate is fine. Thanks for the improvement!

PR #1714, opened 2018-07-27
Yield argument for definition block with arity of one

RobertWSaunders:
## Proposal

To mirror the addition to `Field` in #1712.

To improve flexibility I want to propose that we provide the option to yield the argument to the argument definition block if there is a block provided. 

## How?

Utilize `instance_exec` in place of `instance_eval` if the definition block has an arity of one, otherwise we will default to `instance_eval` which does not yield the argument.

## Why?

Improves flexibility!

Added a test to cover the case of yield the argument.

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -42,7 +42,11 @@ module GraphQL
         @prepare = prepare
 
         if definition_block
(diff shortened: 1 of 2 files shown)


rmosolgo:
🤘  thanks!

Issue #1684, opened 2018-07-18
[1.8] Objects inheriting Interface definition methods


swalkinshaw:
This is in reference to the feature added here: https://github.com/rmosolgo/graphql-ruby/pull/1635. Objects that implement an interface also get their definition methods now.

I'm curious about the reasoning behind this feature. I don't really agree with the idea that Interfaces definition methods should be included in objects that implement them. Although Interfaces are Ruby Modules, it seems to be conflating them. If you want shared behaviour between an Interface and an Object, why not use a plain Module and include them in both places?

Background: I'm migrating our codebase to use the new modules and this feature has caused large issues. We have common methods in both interfaces and objects and now they are getting overridden by the Interface's version. I'm running a fork right now where I've removed this, and it fixes (most) of our issues.

cc @joostverdoorn as well. Maybe you have some valid uses cases you can share.

joostverdoorn:
Hi there,

Here's a simplified use case similar to what we use on our systems:

```ruby
module ScopedInterface
  include GraphQL::Schema::Interface

  definition_methods do
    def base_scope(&resolve_block)
      ... 
    end

    def scope(name, type, &resolve_block)
      ...
    end

    def scoped_list_field(name, type)
      field(name, field: GraphQL::Schema::Field.new(...))
    end
  end

end
```

```ruby
module MessageInterface
   include ScopedInterface
   scope(:read, Boolean) { |base_scope, arg| base_scope.where(read: arg) }
end
```

```ruby
class CommentType < GraphQL::Schema::Object
  implements MessageInterface
  base_scope { Comment.all }
end
```

```ruby
class PostType < GraphQL::Schema::Object
  implements MessageInterface
  base_scope { Post.all }

  scoped_list_field :comments, CommentType
end
```


```ruby
class UserType < GraphQL::Schema::Object
  implements ScopedInterface

  scoped_list_field :posts, PostType
  scoped_list_field :comments, CommentType
end
```

```graphql
query($userId: ID!) {
  user(id: $userID) {
    posts(read: true) {
      comments(read: true) {
        ...
      }
    }
  }
}
```

IMO this is a valid use case where definition methods carrying over to implementing classes is beneficial. It may not be the only way of doing things, but it makes sense when `definition_methods` is already offered for interfaces.

swalkinshaw:
🤔 do `ScopedInterface` or `MessageInterface` actually have fields on them? Or *just*  `definition_methods`?

[...]

PR #1715, opened 2018-07-27
Only allow Objects to implement actual Interfaces

swalkinshaw:
Ref: #1684

Right now an object type can use `implements` to "include" a plain Ruby module.

This adds a check to enforce that the module being implemented is an actual GraphQL interface module.

Note: this could possibly be a breaking change for some people, but I believe it's worth it? The other option for now is a deprecation/warning.

### DIFF
--- a/spec/graphql/schema/object_spec.rb
+++ b/spec/graphql/schema/object_spec.rb
@@ -41,14 +41,32 @@ describe GraphQL::Schema::Object do
       assert_equal object_class.description, new_subclass_2.description
     end
 
-    it "does not inherit singleton methods from base interface when implementing base interface" do
-      object_type = Class.new(GraphQL::Schema::Object)
-      methods = object_type.singleton_methods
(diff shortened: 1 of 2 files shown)


spec/graphql/schema/object_spec.rb
```diff
@@ -41,14 +41,32 @@
...
+    end
+  end
+
+  describe "implementing interfaces" do
+    it "raises an error when trying to implement a non-interface module" do
       object_type = Class.new(GraphQL::Schema::Object)
-      methods = object_type.singleton_methods
-      method_defs = Hash[methods.zip(methods.map{|method| object_type.method(method.to_sym)})]
 
-      object_type.implements(GraphQL::Schema::Interface)
```
swalkinshaw: I removed this test since you wouldn't ever directly implement `GraphQL::Schema::Interface` on an object type.

rmosolgo:
> worth it

:+1:, yes, with a very straight-forward upgrade path.

PR #1716, opened 2018-07-27
Revert extending interface methods onto Objects

swalkinshaw:
This is effectively a revert of #1635.

When an Object Type would implement an Interface, the interface's `DefinitionMethods` were being extended.

However this causes unexpected ancestor chain issues when dealing with common methods (such as `to_graphql`) which are common to both interfaces and objects.

This simply removes this feature. Objects no longer extend the definition methods of the interfaces they implement.

Note: this is obviously a breaking change however I can't think of a better solution. This wasn't a core/fundamental feature of 1.8 either since it was introduced in 1.8.5 (and itself was technically a breaking change).


rmosolgo:

PR #1717, opened 2018-07-28
Fix `test:system` for Rails 5.1

hibariya:
Hello. I noticed that a system test task failed on Travis CI but the build was marked as "passed" on the CI. It does not seem to be an intended result. So I changed the Rakefile so the CI fails if the tests fail. Then, I fixed Capybara's version since Rails 5.1 depends on Capybara ~> 2.13.

* Before (an example of failing builds): https://travis-ci.org/rmosolgo/graphql-ruby/jobs/409088307
*  After: https://travis-ci.org/hibariya/graphql-ruby/jobs/409165615

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -31,7 +31,7 @@ appraise 'rails_5.1' do
   # Required for testing action cable
   gem 'puma'
   # Required for system tests
-  gem 'capybara'
+  gem 'capybara', '~> 2.13'
   gem 'selenium-webdriver'
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -96,7 +96,7 @@ namespace :test do
     success = Dir.chdir("spec/dummy") do
       system("bin/rails test:system")
     end
-    success || exit(0)
+    success || abort
   end
 
   task js: "js:test"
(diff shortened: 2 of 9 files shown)


hibariya:
When I generated gemfiles with Appraisal, [some build](https://travis-ci.org/rmosolgo/graphql-ruby/builds/409287397) failed on CI due to these reasons:

1. jekyll-algolia depends on Ruby 2.3.0 or later, but so far, Ruby 2.2.8 is used for the gemfiles on CI.
2. In gemfiles/rails_4.1.gemfile, activesupport (`~> 4.1.10`) depends on json (`>= 1.7.7, ~> 1.7`), but jekyll-algolia depends on json (`~> 2.0`).

I've removed the group manually for now.

rmosolgo:
👏  Wow, thanks for noticing this and fixing it! I wonder if meant `exit(1)` instead of `exit(0)` 😖 . Thanks for digging in on this one!

Yes, manually editing the gemfiles sounds good to me. Maybe you notice that there are already some manual edits in there. I'm not really sure the best way to achieve _both_ `appraisal generate` working properly _and_ `be rake test` working properly :S

PR #1718, opened 2018-07-28
Improve the mutation generator

### LINKED ISSUE #1629: mutation generator bug
If I generate a a new mutation `rails g graphql:mutation UserDelete`.

It will add 
```
class Types::MutationType  field :userDelete, Mutations::UserDelete.field
 < Types::BaseObject
```
to my mutation_type.rb. This leads to an error, because its placed in the wrong place it should be somewhere within the class.

could be intended to add it like this
```
class Types::MutationType < Types::BaseObject
  field :userDelete, Mutations::UserDelete.field
end
```

hibariya:
I fixed #1629. I suppose this fix is not perfect, but works correctly in most cases.

### DIFF
--- a/lib/generators/graphql/mutation_generator.rb
+++ b/lib/generators/graphql/mutation_generator.rb
@@ -35,7 +35,7 @@ module Graphql
 
         template "mutation.erb", "#{options[:directory]}/mutations/#{file_name}.rb"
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
🎉  Nice, thanks for the fix!

Issue #1671, opened 2018-07-15
Authorization: Scoping


rmosolgo:
One related feature to authorization is _scoping_ (see a  [comment on the auth PR](https://github.com/rmosolgo/graphql-ruby/pull/1494#issuecomment-394761458), [Pundit scopes](https://github.com/varvet/pundit#scopes) and [CanCan `accessible_by`](https://github.com/cancancommunity/cancancan/wiki/Fetching-Records). 

In this feature, there's a function that takes `(items, context)` as input and returns `filtered_items`, where items has been scoped to remove items that are inaccessible to `context`. For example, an ActiveRecord::Relation might receive `.where` calls or an Array might receive `.select`. 

This is _different_ than filtering a list based on a GraphQL field's arguments, because the filtering might be based on something outside the client's control (for example, the current user's permission level). Also, the same filter may apply to lists of the same kinds of objects, even if those lists are returned from different fields or different combinations of arguments. 

This feature is _complementary to_  authorization. Authorization is a last-ditch effort to avoid returning unauthorized objects to clients; scoping is _part of_ the application logic to filter return values. (An authorization failure should probably be corrected by adding filtering to a resolve method; scoping _may be_ that filtering.) 

How can it be added to GraphQL-Ruby? How about: 

- Objects, Interfaces and Unions have a `.scope_list(list, context)` method 
- Inside the runtime, after calling a field which returns a list of type `T`, it calls `T.scope_list(list, context)` with the returned values. The return of `T.scope_list` is used in place of the originally-returned list. 
- The default implementation of `.scope_list` is `return list` (no-op)

Advantages: 

- The filter method is inside the type definition whose objects are filtered 
- It can be implemented any which way 

Disadvantages: 

- It uses GraphQL type, not Ruby class, so it could be hard to cover all your bases 
- Arrays and AR::Relations get routed to the same method, is that bad? (This could also be fixed in user-land, by switching on the class of `list` and routing it to different methods, eg `filter_array` and `filter_relation`)

Any other thoughts on this?

eapache:
> Inside the runtime, after calling a field which returns a list of type T, it calls T.scope_list(list, context) with the returned values. The return of T.scope_list is used in place of the originally-returned list.

I suppose this works with fields that return AR relations because AR won't materialize the actual array until as late as possible so you can always append more `.where` calls. We have an internal pagination thing which isn't quite as lazy though, so this kind of scoping really of necessity needs to be built into the resolver for us.

[...]

PR #1723, opened 2018-07-30
Add .scope_items for filtering lists

rmosolgo:
Implements #1671 

I'd love to hear any thoughts about the in-progress docs and code!


gumchum:
Hi @rmosolgo, is it possible to use this feature with lazy execution? I'm running into an issue where I'd like to batch load some items related to those in the list in order to determine what should be filtered. I'm happy to open an issue if that's the more appropriate channel.

rmosolgo:
It _should_ work, please open an issue with the code you tried and all the details about what happened when you tried it!

PR #1724, opened 2018-07-30
Refactor wrapping types to use new Wrapper class

RobertWSaunders:
## Proposal

Introduce `GraphQL::Schema::Wrapper` to move out some common code in `GraphQL::Schema::NonNull` and `GraphQL::Schema::List`.

## Why

Reduces code and DRY. Adds a better equality operator for wrapping types.

Not sure if you welcome the new class but thought it was a nice refactor! 😄

### DIFF
--- a/lib/graphql/schema/wrapper.rb
+++ b/lib/graphql/schema/wrapper.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module GraphQL
+  class Schema
+    class Wrapper
+      include GraphQL::Schema::Member::CachedGraphQLDefinition
+      include GraphQL::Schema::Member::TypeSystemHelpers
+      extend Forwardable
+
+      # @return [Class, Module] The inner type of this list, the type of which one or more objects may be present.
+      attr_reader :of_type
+
+      def initialize(of_type)
+        @of_type = of_type
+      end
+
+      def to_graphql
+        raise NotImplementedError
+      end
+
+      def unwrap
+        @of_type.unwrap
+      end
+
+      def ==(other)
+        self.class == other.class && of_type == other.of_type
+      end
+    end
+  end
+end
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -21,6 +21,7 @@ require "graphql/schema/build_from_definition"
 
 
 require "graphql/schema/member"
+require "graphql/schema/wrapper"
 require "graphql/schema/list"
 require "graphql/schema/non_null"
 require "graphql/schema/argument"
--- a/lib/graphql/schema/list.rb
+++ b/lib/graphql/schema/list.rb
@@ -5,29 +5,17 @@ module GraphQL
     # Represents a list type in the schema.
     # Wraps a {Schema::Member} as a list type.
     # @see {Schema::Member::TypeSystemHelpers#to_list_type}
-    class List
-      include GraphQL::Schema::Member::CachedGraphQLDefinition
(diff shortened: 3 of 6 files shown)


lib/graphql/schema/wrapper.rb
```diff
@@ -0,0 +1,26 @@
...
+      def to_graphql
+        raise NotImplementedError
+      end
+
+      def unwrap
+        @of_type.unwrap
+      end
+
+      def to_type_signature
+        "#{@of_type.to_type_signature}!"
```
rmosolgo: Oops, I think this one should stay in the child classes, since `List`'s implementation is different.

RobertWSaunders: Oops, totally! Nice catch, thank!

RobertWSaunders:
@rmosolgo thoughts on adding an override for the equality operator in `Wrapper`, would check `of_type` is also the equal:

```
def ==(other)
   self.class == other.class && of_type == other.of_type
end
```

rmosolgo:
:+1: that `#==` sounds ok to me. (Funny, I thought it rang a bell so I found the [analogous method in the legacy types](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/base_type.rb#L100-L102) but I think it has a bug :P)

RobertWSaunders:
Sweet, I've added it! This should be good to go assuming you are good with it and we get 🍏!

lib/graphql/schema/wrapper.rb
```diff
@@ -0,0 +1,28 @@
...
+      def to_graphql
+        raise NotImplementedError
+      end
+
+      def unwrap
+        @of_type.unwrap
+      end
+
+      def ==(other)
+        self.class == other.class && of_type == other.of_type
```
eapache: 🤔 does this need to account for intermediate wrappers? i.e. should `[Foo] == `[Foo!]`?

eapache: Oh wait, you're comparing the `of_type`s directly not using the recursion in `unwrap`, so... the recursion happens through the repeated `==` calls instead and does the right thing. 🤦‍♂️

RobertWSaunders: Yep exactly!

RobertWSaunders:
I realized that the legacy types implement the methods that were defined in these wrapping types, so by using `delegate` there is now no need to redefine them. I've updated my code, thoughts?

rmosolgo:
>  using `def_delegators`

I'd actually prefer a bit of duplication for now, because my goal is to someday make GraphQL-Ruby work _without_ the legacy layer. It'll probably take until Christmas or something, but if we can help it, I'd rather not introduce a dependency on those legacy objects. Would you mind putting back those (duplicated) implementations?

lib/graphql/schema/member/type_system_helpers.rb
```diff
@@ -35,4 +35,4 @@ def kind
       end
     end
   end
-end
+end
```
rmosolgo: Should this whitespace change be included, or can it be reverted?

RobertWSaunders: Yep, seems like I need to make some editor config changes!

RobertWSaunders:
@rmosolgo Yep, that makes sense, I've added them back and included some tests for `GraphQL::Schema::List` and `GraphQL::Schema::NonNull`!

rmosolgo:
👏  Thanks for this improvement!

PR #1728, opened 2018-07-31
Coerce input field defaults to support nested default values

pushrax:
Given a schema like:

```graphql
input A {
  b: Int = 42
}

input B {
  a: A = {}
}
```

Prior to this change, constructing a `B` from defaults would result in a nil value for `a.b`, when instead it should have the value `42`. On this topic, [the spec says](http://facebook.github.io/graphql/June2018/#sec-Input-Objects):

> If a literal value is provided for an input object field, an entry in the coerced unordered map is given the result of coercing that value according to the input coercion rules for the type of that field.

Does it make sense to also call `input_field_defn.prepare` with default values? I haven't used that feature for anything yet.

@rmosolgo cc @jackychiu @michaelkipper

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -65,7 +65,7 @@ module GraphQL
         @default_value = nil
       else
         @has_default_value = true
-        @default_value = new_default_value
+        @default_value = GraphQL::Argument.deep_stringify(new_default_value)
       end
     end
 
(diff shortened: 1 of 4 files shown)


lib/graphql/argument.rb
```diff
@@ -65,7 +65,7 @@ def default_value=(new_default_value)
         @default_value = nil
       else
         @has_default_value = true
-        @default_value = new_default_value
+        @default_value = GraphQL::Argument.deep_stringify(new_default_value)
```
pushrax: I had to make this change to fix the tests. The [`order_by` field](https://github.com/rmosolgo/graphql-ruby/blob/c7d537df071b40e2fe9d59ac3a9d8aefb0adb284/spec/support/dummy/schema.rb#L236) in the test schema has a default using a symbol key, which wouldn't get correctly coerced.

Thoughts on the best place to put `deep_stringify`?

rmosolgo: The move makes sense to me, I don't really have a thought about where to put `deep_stringify`. Since it's a pure function, and private, I think it doesn't matter too much.

rmosolgo:
Thanks for this fix!

> Does it make sense to also call `input_field_defn.prepare` with default values?

I ... don't know 😖  I also haven't used that feature yet. I guess we'll see.

PR #1729, opened 2018-08-01
Add default_graphql_name in BaseDSL methods

RobertWSaunders:
## Proposal

Create a `default_graphql_name` in `GraphQL::Schema::Member::BaseDSLMethods` method to be used in the base implementation of `graphql_name`. This will allow people to override just how default schema member name gets created without having to override the entire method.

## Why

Improves flexibility! 😃 

## How

Add `default_graphql_name` and make use of it in `graphql_name`.

## Questions

Is that a good place for my test to live?

### DIFF
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -10,8 +10,7 @@ module GraphQL
         # Call this with a new name to override the default name for this schema member; OR
         # call it without an argument to get the name of this schema member
         #
-        # The default name is the Ruby constant name,
-        # without any namespaces and with any `-Type` suffix removed
+        # The default name is implemented in default_graphql_name
         # @param new_name [String]
         # @return [String]
         def graphql_name(new_name = nil)
@@ -20,10 +19,10 @@ module GraphQL
             @graphql_name = new_name
           when overridden = overridden_graphql_name
             overridden
-          else # Fallback to Ruby constant name
-            raise NotImplementedError, 'Anonymous class should declare an `graphql_name`' if name.nil?
+          else
+            raise NotImplementedError, 'Anonymous class should declare a `graphql_name`' if name.nil?
 
-            name.split("::").last.sub(/Type\Z/, "")
+            default_graphql_name
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/member/base_dsl_methods.rb
```diff
@@ -20,10 +19,10 @@ def graphql_name(new_name = nil)
             @graphql_name = new_name
           when overridden = overridden_graphql_name
             overridden
-          else # Fallback to Ruby constant name
-            raise NotImplementedError, 'Anonymous class should declare an `graphql_name`' if name.nil?
+          else # Fallback to default_graphql_name
```
swalkinshaw: not sure this comment is needed anymore

RobertWSaunders: Updated!

spec/support/jazz.rb
```diff
@@ -99,6 +99,14 @@ def to_graphql
...
   end
 
+  class OverrideNameObject < GraphQL::Schema::Object
+    class << self
+      def default_graphql_name
+        "Override"
+      end
+    end
+  end
+
```
rmosolgo: Since this object doesn't _have_ to be part of this big example schema, how about moving it into the test file itself? For example, it could go right before the `it "..."`. 

I suggest it because I have a love-hate relationship with big example schemas like this one: on the one hand, sometimes you _need_ a complex example to demonstrate some behavior, or the easiest way to add a new test case is to extend the big example somehow. But on the other hand, this schema represents a lot of criss-crossing test changes, so over time, it becomes harder and harder to change. 

If it's not too much trouble to keep the example classes somewhere more local (somewhere closer to the test case itself), I think it'll make us happier down the road!

RobertWSaunders: Updated! That makes sense to me, only reason I added that question about my test in the PR description was because it felt weird to me in the first place. 👍

rmosolgo:
🎊

lib/graphql/schema/member/base_dsl_methods.rb
```diff
@@ -20,10 +19,10 @@ def graphql_name(new_name = nil)
             @graphql_name = new_name
           when overridden = overridden_graphql_name
             overridden
-          else # Fallback to Ruby constant name
-            raise NotImplementedError, 'Anonymous class should declare an `graphql_name`' if name.nil?
+          else
+            raise NotImplementedError, 'Anonymous class should declare a `graphql_name`' if name.nil?
```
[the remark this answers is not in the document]
RobertWSaunders: Yep I agree, since `default_graphql_name` is reliant on `name`. I should have caught this!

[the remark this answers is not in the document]
RobertWSaunders: I have another PR to make today and will include this small change in it!

PR #1732, opened 2018-08-02
Add add_argument to GraphQL::Schema::Member::HasArguments

RobertWSaunders:
## Proposal

Add an `add_argument` method to `GraphQL::Schema::Member::HasArguments`. This functionality is similar to how `GraphQL::Schema::Member::HasFields` exposes an `add_field` method. 

## Why

Improve consistency and flexibility. (ex. Arguments can now be added to a field outside a field definition)

## How

Create the method `add_argument`, add a test and make use of the new method where applicable.

### DIFF
--- a/lib/graphql/schema/member/has_arguments.rb
+++ b/lib/graphql/schema/member/has_arguments.rb
@@ -16,6 +16,13 @@ module GraphQL
         def argument(*args, **kwargs, &block)
           kwargs[:owner] = self
           arg_defn = self.argument_class.new(*args, **kwargs, &block)
+          add_argument(arg_defn)
+        end
+
+        # Register this argument with the class.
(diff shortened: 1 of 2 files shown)


rmosolgo:
This is fine with me, but can you help me understand why it's beneficial? Is there a use case for `#add_argument`?

RobertWSaunders:
If anyone ever needed to add an argument to a field outside the field definition block this is useful, this could be particularly useful for adding arguments to a field for filtering dynamically etc. 

I also just think this is a nice addition to be consistent with `HasFields` and just in general makes the gem more flexible.

swalkinshaw:
Specifically this is used in Shopify's version of field filters.

rmosolgo:
:+1: Thanks, just wanted to double-check it had some real-world utility before adding the API surface area!

Issue #1256, opened 2018-02-01
require 'graphql' in RSpec generates many warnings


delner:
For GraphQL 1.7.9, Ruby 2.3.4 and RSpec 3.7.1

Given `graphql_spec.rb`:

```ruby
RSpec.configure do |config|
  config.warnings = true
end

require 'graphql' # This causes the warnings

RSpec.describe 'Dummy test' do
  it { expect(true).to be true }
end
```

Running `bundle exec rspec spec/graphql_spec.rb` produces the following:

```
$ bundle exec rspec spec/graphql_spec.rb
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/enum_type.rb:142: warning: method redefined; discarding old name=
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/argument.rb:60: warning: method redefined; discarding old default_value=
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/argument.rb:69: warning: method redefined; discarding old type=
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/argument.rb:75: warning: method redefined; discarding old type
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/field.rb:264: warning: method redefined; discarding old name=
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/field.rb:276: warning: method redefined; discarding old property=
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/field.rb:282: warning: method redefined; discarding old hash_key=
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/define/instance_definable.rb:147: warning: instance variable @pending_definition not initialized
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/define/instance_definable.rb:147: warning: instance variable @pending_definition not initialized
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/define/instance_definable.rb:147: warning: instance variable @pending_definition not initialized
/.rvm/gems/ruby-2.3.4/gems/graphql-1.7.9/lib/graphql/define/instance_definable.rb:147: warning: instance variable @pending_definition not initialized
[...]

PR #1735, opened 2018-08-03
Some fixes for GraphQL::Schema::Field#initialize

hibariya:
Hello. I noticed some code need to be fixed while investigating #1256. I will explain them in line comments below.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -102,7 +102,7 @@ module GraphQL
         if name.nil?
           raise ArgumentError, "missing first `name` argument or keyword `name:`"
         end
-        if !(field || function || mutation || resolver)
+        if !(field || function || resolver_class)
           if type.nil?
             raise ArgumentError, "missing second `type` argument or keyword `type:`"
           end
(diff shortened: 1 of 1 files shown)


lib/graphql/schema/field.rb
```diff
@@ -102,16 +102,16 @@ def initialize(type: nil, name: nil, owner: nil, null: nil, field: nil, function
         if name.nil?
           raise ArgumentError, "missing first `name` argument or keyword `name:`"
         end
-        if !(field || function || mutation || resolver)
+        if !(field || function || resolver_class)
```
hibariya: `mutation` and `resolver` always return nil because `@resolver_class` is not yet initialized here. I think we can use the `resolver_class` argument instead of these methods.

lib/graphql/schema/field.rb
```diff
@@ -102,16 +102,16 @@ def initialize(type: nil, name: nil, owner: nil, null: nil, field: nil, function
...
             raise ArgumentError, "missing second `type` argument or keyword `type:`"
           end
           if null.nil?
             raise ArgumentError, "missing keyword argument null:"
           end
         end
-        if (field || function || resolve || mutation) && extras.any?
-          raise ArgumentError, "keyword `extras:` may only be used with method-based resolve, please remove `field:`, `function:`, `resolve:`, or `mutation:`"
+        if (field || function || resolve) && extras.any?
+          raise ArgumentError, "keyword `extras:` may only be used with method-based resolve and class-based field such as mutation class, please remove `field:`, `function:` or `resolve:`"
```
hibariya: The keyword argument `extras` has been supported too in class-based fields since #1313. So I've removed `mutation` from here and edited the error message.

rmosolgo:
Thanks for these fixes!

PR #1736, opened 2018-08-03
Mutation auth improvements

rmosolgo:
I'm trying this out in GitHub and tweaking things a bit. So far: 

- Add a `#object_from_id(type, id, ctx)` hook so that loading objects can be customized 
- Support lazy objects from that hook (promises)
- Remove `#validate_{arg}` hooks and replace them with a single `#authorized?(**inputs)` hook. This is a breaking change, but if you want to preserve the old behavior, you can implement authorized to do the same thing: 

  ```ruby 
  def authorized?(inputs) 
    inputs.each { |k, v|  public_send("validate_#{k}", v) } 
  end 
  ```

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -164,43 +164,58 @@ module GraphQL
         end
       end
 
+      # Look up the corresponding object for a provided ID.
+      # By default, it uses Relay-style {Schema.object_from_id},
+      # override this to find objects another way.
+      #
+      # @param type [Class, Module] A GraphQL type definition
+      # @param id [String] A client-provided to look up
+      # @param context [GraphQL::Query::Context] the current context
+      def object_from_id(type, id, context)
+        context.schema.object_from_id(id, context)
+      end
+
       def load_application_object(arg_kwarg, id)
         argument = @arguments_by_keyword[arg_kwarg]
-        # See if any object can be found for this ID
-        application_object = context.schema.object_from_id(id, context)
-        if application_object.nil?
-          raise LoadApplicationObjectFailedError.new(argument: argument, id: id, object: application_object)
-        end
-        # Double-check that the located object is actually of this type
-        # (Don't want to allow arbitrary access to objects this way)
         lookup_as_type = @arguments_loads_as_type[arg_kwarg]
-        application_object_type = context.schema.resolve_type(lookup_as_type, application_object, context)
-        possible_object_types = context.schema.possible_types(lookup_as_type)
-        if !possible_object_types.include?(application_object_type)
-          raise LoadApplicationObjectFailedError.new(argument: argument, id: id, object: application_object)
-        else
-          # This object was loaded successfully
-          # and resolved to the right type,
-          # now apply the `.authorized?` class method if there is one
-          if (class_based_type = application_object_type.metadata[:type_class])
-            context.schema.after_lazy(class_based_type.authorized?(application_object, context)) do |authed|
-              if authed
-                application_object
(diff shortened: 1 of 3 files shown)


rmosolgo:
@cainlevy I think the `#object_from_id(type, id, ctx)` will support your case with IDs, does that sound right to you? Anything missing there? Errors as data is still ... `wip` ... :P

lib/graphql/schema/resolver.rb
```diff
@@ -91,6 +91,15 @@ def resolve(**args)
       def before_prepare(**args)
       end
 
+      # Called after arguments are loaded, but before resolving.
+      #
+      # Override it to check everything before calling the mutation.
+      # @param args [Hash] The input arguments
+      # @raise [GraphQL::ExecutionError] To add an error to the response
+      # @raise [GraphQL::UnauthorizedError] To signal an authorization failure
+      def authorized?(**args)
```
mikrobi: The name `authorized?` might imply that the method always returns a boolean? But the default implementation returns `nil`. Do you think it would make sense to use `!` similar to methods in Rails like [`last!`](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-last-21) or [`find_by!`](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21) that raise exceptions?

cainlevy:
> * Add a `#object_from_id(type, id, ctx)` hook so that loading objects can be customized

Would this look something like:

```ruby
class MyMutation < Mutations::BaseMutation
  argument :employeeId, ID, required: true, loads: Types::Employee

  def object_from_id(type, id, context)
    case type
    when Types::Employee
      Employee.find(id)
    end
  end
end
```

If so, then:

* I had to guess what type would be made available to me (`ID`? `Types::Employee`?). Would it make sense to pass the field name instead?
* I like that this is local to the mutation (resolver). I find `Schema.object_from_id` to be rather distant and correspondingly spooky at first contact.

> * Remove `#validate_{arg}` hooks and replace them with a single `#authorized?(**inputs)` hook.

This flexibility solves another problem for me: one of my authorizations depends on values from another field. 👍 

Will the `as:` option have been applied at this point? What name will the inputs have in this hook?

lib/graphql/schema/resolver.rb
rmosolgo: oh yeah, maybe it should be updated to handle a true/false return value, or renamed like you said

rmosolgo: Maybe accepting `false` would solve #1737  too, you could do something like 

```
# Don't continue execution, instead, use this hash as the mutation's return value
return false, { errors: validation_errors }
```

rmosolgo: updated in aa8b5f8a

rmosolgo:
> pass the field name instead

I'm hesitant about that. Since joining GitHub, I've taken both approaches: first, making authorization hooks field-aware, but then removing it. The problem is, with GraphQL, those hooks can get called from unexpected places (since the queries are very flexible and the schema is always growing), but when the authorization hooks are field-aware, it's easy to wrongly assume that we can account for every possibility. Since we can't, those hooks ended up with misinformation and bugs. So, I'd rather keep those hooks as _general_ as possible, so that they can be safely called wherever they're needed. 

> `as:` option

I'm ... not sure 😬  I don't use it.

rmosolgo:
I think this will be a nice improvement, even if we have to sort out some details about arguments down the line. Thanks for taking a look!

Issue #1734, opened 2018-08-02
errors as data from validation failures


cainlevy:
The `validate_X` methods are [recommended](http://graphql-ruby.org/mutations/mutation_authorization.html#can-this-user-modify-this-thing) to check that a specific user may mutate a specific thing in a specific way.

Sometimes this validation may fail for routine (non-malicious) reasons. For example, the resource may be locked (e.g. archived, expired). In these cases it's helpful to show a specific error to a human using [errors as data](http://graphql-ruby.org/mutations/mutation_errors#errors-as-data).

How can error data be returned from validate methods?

(Bonus: does the answer also work for prepare hooks?)

rmosolgo:
That's a great question, I don't have an answer but I agree it should be possible 😊  

I'm currently hooking up this mutation authorization up in GitHub so it has a few other changes coming down the pike, I'll keep this in mind.

Bahanix:
I do not know if my approach is good, but it works:

```ruby
class Resolvers::Posts::UpdateResolver < GraphQL::Function

  argument :post, !Inputs::Post::UpdateInput
  type !Types::PostFormType

  def call(obj, args, ctx)
    post = #...
    post.update(args[:post])

    OpenStruct.new({
      post: post,
      post_validation: OpenStruct.new(post.errors.to_h)
      # errors.to_h is for the example, usually that's not exactly what you want
    })
  end

end
```

```ruby
Types::PostFormType = GraphQL::ObjectType.define do
  name 'PostForm'

  field :post, Types::PostType
  field :post_validation, Types::PostValidationType
end

Types::PostValidationType = GraphQL::ObjectType.define do
  name 'PostValidation'

  field :content, types.String
end
```

The biggest downside is that every field of the OpenStruct is computed even if not requested by the client. There probably are solutions using OpenStruct and lambdas (eg: https://gist.github.com/paulfitz/10341879) but it seems even less documented.

I would be very glad having feedbacks on this pattern.

Bahanix:
I do not know i my approach is good, but it works:

```ruby
class Resolvers::Posts::UpdateResolver < GraphQL::Function

  argument :post, !Inputs::Post::UpdateInput
  type !Types::PostFormType

  def call(obj, args, ctx)
    post = #...
    post.update(args[:post])

    OpenStruct.new({
      post: post,
      post_validation: OpenStruct.new(post.errors.to_h)
      # errors.to_h is for the example, usually that's not exactly what you want
    })
  end

[...]

PR #1737, opened 2018-08-03
Add a flow for returning data early from Mutation#authorized?

rmosolgo:
As pointed out by @cainlevy in #1734 , the new mutation auth doesn't support errors-as-data very well, but it definitely _should_, because errors-as-data is nice. 

How to do it? I can think of a few implementations: 

- Somehow use the return value of `.authorized?` (this is how before-save callbacks used to work, if they returned `false`, then save was prevented. I think this was changed to [`throw(:abort)`](https://blog.bigbinary.com/2016/02/13/rails-5-does-not-halt-callback-chain-when-false-is-returned.html)
- Use `throw :early_return, response`. This has the advantage of being plain Ruby, but it's not a very familiar feature. Also, it halts program flow, which is convenient, since later code will be irrelevant. 
- Use a normal method, as shown here. It's abstract, which is nice, but doesn't halt flow. This is like Rails controller `render` methods.
- Use a method that calls `throw` under the hood (like sinatra's `halt`, I think). It's nice that it halts flow, but maybe too magical? 

Any thoughts? 

TODO: 

- [ ] before_prepare should get the same treatment, be renamed to a `?` method.

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -31,6 +31,8 @@ module GraphQL
         @context = context
         # Since this hash is constantly rebuilt, cache it for this call
         @arguments_by_keyword = {}
+        # This might be set by {#early_return(val)}
+        @early_return_value = nil
         self.class.arguments.each do |name, arg|
           @arguments_by_keyword[arg.keyword] = arg
         end
@@ -63,7 +65,9 @@ module GraphQL
             authorized_val = authorized?(loaded_args)
             context.schema.after_lazy(authorized_val) do
               # Finally, all the hooks have passed, so resolve it
-              if loaded_args.any?
+              if @early_return_value
+                @early_return_value
+              elsif loaded_args.any?
                 public_send(self.class.resolve_method, **loaded_args)
               else
                 public_send(self.class.resolve_method)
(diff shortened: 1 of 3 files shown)


cainlevy:
Thoughts on creating an official GraphQL::Result class? I imagine it's no small undertaking and would trickle out to many bits of the API, but it seems like the closest way to represent errors as data.

One option might be to always expect a Result back from `.authorized?`. The developer would need to indicate when it's a "continue" Result and when it's a "halt" result. Honestly this just feels like a Promise.

Another option would be to continue for any truthy value that is _not_ a `GraphQL::Result` object. In that case, `.authorized?` would need to return `GraphQL::Result.new(data)` when it wants to halt resolution with error data.

rmosolgo:
>  any truthy value that is _not_

Oh, that's a nice direction, something similar would be: 

- Continue on `true`
- Halt on:
  - `false`, with no data at all 
  - a non-`true` value, using that value as the response 

In that way, returning a hash would Just Work :tm:

cainlevy:
I'm inclined to caution in Ruby when there's a difference between `true` and truthy (or between `false` and falsey). I've had co-workers who thought it was an anti-pattern to use `!!` for casting a result to a strict boolean.

rmosolgo:
Oh yeah, I saw a rubocop against it once! 🤷‍♂️

rmosolgo:
I updated this to use multiple return for errors-as-data, eg 

```ruby
return false, { errors: [...] }
```

I also updated the guides to give a sense of it. How does that look to you?

lib/graphql/schema/resolver.rb
```diff
@@ -210,6 +222,23 @@ def load_application_object_failed(err)
...
+      # Call this during loading or authorization to
+      # skip {#resolve} and return data instead of top-level errors.
+      #
+      # It doesn't halt the flow of the code so be sure to return
+      # after calling it, or combine it with `return return_with(...)`.
+      #
+      #
+      # @param response [Object] The value to respond with
+      # @return [void]
+      def return_with(response)
```
cainlevy: was this a previous idea? if it's still on the table, i think `resolve_with` might fit the GraphQL nomenclature nicely.

rmosolgo: oops, oh yeah, it can be removed now since `return` is used instead

PR #1738, opened 2018-08-03
Change `call_hooks` to be iterative rather than recursive

tenderlove:
We're seeing very deep call stacks from `call_hooks` in production.
This commit just changes the `call_hooks` method to be iterative rather
than recursive so that we don't have to build such a large stack when
executing instrumentation hooks.

@rmosolgo if you check the flame graphs in production, you should be able to see these calls (I just didn't want to post the graph here).  The problem is that our stacks can get to `instrumenters.length * queries.length` or basically n^2.  This change reduces it to `queries.length * 2`.  I think we can change `each_query_call_hooks` to be iterative rather than recursive, but I haven't cracked that yet.

Thanks!

### DIFF
--- a/lib/graphql/execution/instrumentation.rb
+++ b/lib/graphql/execution/instrumentation.rb
@@ -51,30 +51,38 @@ module GraphQL
 
         # Call each before hook, and if they all succeed, yield.
         # If they don't all succeed, call after_ for each one that succeeded.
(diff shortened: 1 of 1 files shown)


tenderlove:
Hmmm, the build failure looks unrelated to my change

rmosolgo:
Makes sense to me, thanks for the improvement!

PR #1739, opened 2018-08-04
Remove duplicate method definitions

hibariya:
According to ruby's warning messages (run with `RUBYOPT=-w`), some method definitions were duplicate. I have removed them without changing existing behavior.

* removed some attribute names from `attr_*` or `def_delegators` since each of them has another definition
* added `undef` and `undef_method` to avoid warnings from ruby (related to #1256)
* removed `CustomPrinter` from toplevel
* moved multiple `GlobalIDUser` class definitions into one place

### DIFF
--- a/spec/graphql/language/generation_spec.rb
+++ b/spec/graphql/language/generation_spec.rb
@@ -7,11 +7,13 @@ describe GraphQL::Language::Generation do
       GraphQL.parse('type Query { a: String! }')
     }
 
-    class CustomPrinter < GraphQL::Language::Printer
-      def print_field_definition(print_field_definition)
-        "<Field Hidden>"
-      end
-    end
+    let(:custom_printer_class) {
+      Class.new(GraphQL::Language::Printer) {
+        def print_field_definition(print_field_definition)
+          "<Field Hidden>"
+        end
+      }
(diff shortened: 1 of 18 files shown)


spec/graphql/language/generation_spec.rb
```diff
@@ -7,11 +7,13 @@
       GraphQL.parse('type Query { a: String! }')
     }
 
-    class CustomPrinter < GraphQL::Language::Printer
```
hibariya: Note: This class seems defined at the top-level. In other words, the `CustomPrinter` in `spec/graphql/language/nodes_spec.rb` is identical to this class. I think it is not intended result because this class is defined under the `describe` block. For this reason I changed from the ordinary class definition to `Class.new` definition inside the `let`.

rmosolgo: > not intended 

I agree! Sometimes it's hard to forget that `describe ...` blocks are top-level. Thanks for fixing it!

rmosolgo:
Thanks for your careful attention here! I'm really happy to have fewer warnings from this library, but I'm intimidated by all the different issues. Thanks for the fix!!

PR #1742, opened 2018-08-07
Remove Forwardable from GraphQL::Schema::Wrapper

RobertWSaunders:
## Proposal

Seems like I missed removing `extend Forwardable` in #1724.

Circling back here to remove it for good, sorry about that. 😃 

## Why

I included Forwardable in #1724 because I wanted to delegate some of the methods to base classes however after the discussion in that PR it made sense to not delegate and instead just reimplement the methods. Now that it's not being used it makes sense to get rid of it. 

## How

Remove `extend Forwardable` from `GraphQL::Schema::Wrapper`.

### DIFF
--- a/lib/graphql/schema/wrapper.rb
+++ b/lib/graphql/schema/wrapper.rb
@@ -5,9 +5,8 @@ module GraphQL
     class Wrapper
       include GraphQL::Schema::Member::CachedGraphQLDefinition
       include GraphQL::Schema::Member::TypeSystemHelpers
-      extend Forwardable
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Nice! By the way, if you want to make one more bit of clean-up, I noticed that `attr_reader :of_type` seems to have a comment copied `List`, how about updating that on this little PR too?

RobertWSaunders:
Good catch, updated!

rmosolgo:
Thanks for taking a look 👌

PR #1748, opened 2018-08-08
Use Base64.urlsafe_decode64 for forward compatibility

gfx:
In order to ensure forward compatibility, or to safely migrate `urslafe_encode64`, graphql-ruby should be able to decode URL-safe base64 cursors and relay ids by default.

This is, of course, backward compatible.

cf. https://github.com/rmosolgo/graphql-ruby/pull/1698#issuecomment-411090879


rmosolgo:
Awesome, that will make compatibility easy in the future!

gfx:
Thanks!

PR #1751, opened 2018-08-08
Add graphql-metrics to related projects list

chrisbutcher:
Adds new gem (https://github.com/Shopify/graphql-metrics) to the list of related projects :)

@rmosolgo


rmosolgo:
Nice, thanks for sharing!

PR #1753, opened 2018-08-08
Explicitly define type kind predicate methods

RobertWSaunders:
## Proposal

Explicitly define type kind predicate methods instead of them defining them at execution.

## How

Remove the `define_method` loop and instead define all the methods explicitly.

## Why

Makes for easier code search! 😃 

I often forget that methods like `enum?` exist and I think its largely because those methods were created during execution, because there aren't many methods and the number will likely not change its nice to just explicitly define them.

### DIFF
--- a/lib/graphql/type_kinds.rb
+++ b/lib/graphql/type_kinds.rb
@@ -29,6 +29,38 @@ module GraphQL
       def to_s;       @name;      end
       # Is this TypeKind composed of many values?
       def composite?; @composite; end
+
+      def scalar?
+        self == TypeKinds::SCALAR
+      end
+
+      def object?
+        self == TypeKinds::OBJECT
+      end
+
+      def interface?
+        self == TypeKinds::INTERFACE
+      end
+
+      def union?
+        self == TypeKinds::UNION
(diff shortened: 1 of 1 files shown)


lib/graphql/type_kinds.rb
```diff
@@ -43,10 +43,36 @@ def composite?; @composite; end
     ]
 
     class TypeKind
-      TYPE_KINDS.map(&:name).each do |kind_name|
-        define_method("#{kind_name.downcase}?") do
-          self.name == kind_name
-        end
+      def scalar?
+        self.name == TypeKinds::SCALAR.name
+      end
```
rmosolgo: Honestly, I don't know why I did this way the first time (it was 3 years ago, after all :P), what do you think about comparing the _objects_ instead of their names, for example: 

```ruby 
def scalar? 
  self == TypeKinds::SCALAR 
end 
```

RobertWSaunders:
No worries, I've updated it as per your suggestions, I agree! 🎉

rmosolgo:
👏 !

Issue #938, opened 2017-09-04
Relay connection support for Mongoid


BranLiang:
Self applied a simple connection for mongoid. It would be really nice to see support from the community and integrate relay connection support for mongoid.
```ruby
# /initializer/mongo_relay.rb
GraphQL::Relay::BaseConnection.register_connection_implementation(Mongoid::Relations::Targets::Enumerable, Connections::MongoConnection)

```
```ruby
class Connections::MongoConnection < GraphQL::Relay::BaseConnection
  # derive a cursor from `item`
  def cursor_from_node(item)
    encode(item.id.to_str)
  end

  private

  # apply `#first` & `#last` to limit results
  def paged_nodes
    return @paged_nodes if defined? @paged_nodes
    # binding.pry
    @paged_nodes = sliced_nodes
    @paged_nodes = if first
                     @paged_nodes.limit(first)
                   elsif last
                     @paged_nodes.desc(:id).limit(last)
                   else
                     @paged_nodes.limit(10)
                   end
  end

  # apply cursor, order, filters, etc
  # to get a subset of matching objects
  def sliced_nodes
    return @sliced_nodes if defined? @sliced_nodes

    @sliced_nodes = nodes
    @sliced_nodes = @sliced_nodes.where(:id.gt => decode(after)) if after
    @sliced_nodes = @sliced_nodes.where(:id.lt => decode(before)) if before
  end
end
```

rmosolgo:
Hi! Thanks for sharing the code. I'm sure many people will find it very useful!

How about this: if you (or someone else) can add Mongoid to the projects, including tests, I'll maintain it 😊

BranLiang:
@rmosolgo I will see what I can do and hope someone could help me with that😊

dorthwein:
Is there clarity or a check list on what needs to be implemented with this to make it efficient with Mongoid?  If there was clarity on what needed to be implemented we maybe able to help some.

dblock:
Looks like support in https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/relay/mongo_relation_connection.rb works well for Mongoid::Criteria and relations in Mongoid work no different than a criteria. This seems to work for me (mongoid 7):

```ruby
GraphQL::Relay::BaseConnection.register_connection_implementation(
  Mongoid::Association::Referenced::HasMany::Targets::Enumerable, GraphQL::Relay::MongoRelationConnection
)
```

For mongoid 6 it should be the same.

[...]

PR #1754, opened 2018-08-08
Added support for Mongoid 5, 6 and 7 relations

dblock:
Actually closes https://github.com/rmosolgo/graphql-ruby/issues/938, enabling plain `has_many` and similar relations to be used as relay connection types. These are no longer `Mongoid::Criteria` since Mongoid 5.


dblock:
💚 and ready to merge @rmosolgo

gemfiles/rails_5.1.gemfile
```diff
@@ -15,9 +15,4 @@ gem "puma"
 gem "capybara", "~> 2.13"
 gem "selenium-webdriver"
 
-group :jekyll_plugins do
-  gem "jekyll-algolia"
-  gem "jekyll-redirect-from"
-end
-
```
dblock: Removed these blocks as in other gemfiles. The `jekyll-algolia` gem requires ruby 2.3.0, and `graphql-ruby` still tests with 2.2.8, so it was producing errors from the auto-generated appraisals. I suspect somone had removed these from the earlier rails version by hand, and they aren't necessary for the tests anyway.

spec/graphql/relay/mongo_relation_connection_spec.rb
```diff
@@ -482,4 +491,49 @@ def get_names(result, field_name)
     connection = GraphQL::Relay::BaseConnection.connection_for_nodes(relation)
     assert_equal GraphQL::Relay::MongoRelationConnection, connection
   end
+
+  describe "relations" do
+    let(:query_string) {%|
```
dblock: This one is quick and dirty. Happy to iterate on a smaller query if you care, but could use some help in writing one.

rmosolgo:
🎊  This looks awesome, thanks for your work here!

PR #1756, opened 2018-08-08
Move NotImplementedError out to default_graphql_name

RobertWSaunders:
## Proposal

Small tweak to move the `NotImplementedError` into `default_graphql_name` based on @eapache suggestion, missed this in #1729.

## Why

Raise the error where it should be raised!


rmosolgo:
Makes sense to me!

PR #1757, opened 2018-08-09
Update to make life easier

gatorjuice:
As a new user of graphql, it never occurred to me that I needed to use camel case to query even when the ruby field is in snake case. I was stuck on this for much time...much...time...many...times.

Not sure your preference on which exact field to use but please take this as a seed of an idea.

### DIFF
--- a/guides/getting_started.md
+++ b/guides/getting_started.md
@@ -49,6 +49,8 @@ class PostType < GraphQL::Schema::Object
   description "A blog post"
   field :id, ID, null: false
   field :title, String, null: false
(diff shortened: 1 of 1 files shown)


rmosolgo:
Sorry it made you stuck 😖  Thanks for suggesting this improvement, I bet you're right, it will make the behavior much clearer while getting started!

rmosolgo:
(I modified the example just a bit in 2e88a419)

gatorjuice:
Thanks so much!

PR #1762, opened 2018-08-09
Update limits.md

quinn:
Based on my understanding this is the correct syntax here?

### DIFF
--- a/guides/fields/limits.md
+++ b/guides/fields/limits.md
@@ -13,9 +13,9 @@ index: 4
 Always limit the number of items which can be returned from a list field. For example, use a `limit:` argument and make sure it's not too big. The `prepare:` function provides a convenient place to cap the number of items:
 
 ```ruby
-field :items, types[ItemType] do
+field :items, Types::ItemType do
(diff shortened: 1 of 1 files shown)


guides/fields/limits.md
```diff
@@ -13,9 +13,9 @@ index: 4
 Always limit the number of items which can be returned from a list field. For example, use a `limit:` argument and make sure it's not too big. The `prepare:` function provides a convenient place to cap the number of items:
 
 ```ruby
-field :items, types[ItemType] do
+field :items, Types::ItemType do
   # Cap the number of items at 30
-  argument :limit, types.Int, default_value: 20, prepare: ->(limit, ctx) {[limit, 30].min}
+  argument :limit, Int, default_value: 20, prepare: ->(limit, ctx) {[limit, 30].min}
```
rmosolgo: I think this should be `Integer` now

quinn:
@rmosolgo updated to `Integer`

rmosolgo:
Thanks for the update!

Issue #1711, opened 2018-07-26
GraphQL::Schema::NonNull#to_type_signature -> #graphql_name ? 


rmosolgo:
Most types respond to `#graphql_name`, but not the new list and non-null wrappers. Maybe their `#to_type_signature` methods should be aliased, just to give us fewer names to remember.

RobertWSaunders:
Are you saying to make `alias_method :graphql_name, :to_type_signature`? or rename `to_type_signature` to 'graphql_name` and `alias_method :to_type_signature, :graphql_name`? 

IMO I think the wrapping types should both implement `graphql_name` and `description` but return nil for both of them. By specification wrapping types have a name and description but they are nullable fields.

What are your thoughts on this? 

I have a PR that I'm working on that could easily add this change.

rmosolgo:
:+1: to rename and implement `description` like you said, I think I wrote some code using `:to_type_signature` ,but it's probably better to do a breaking change now rather than leave it around.

rmosolgo:
I'm working through a lot of this over in #2363 , if anyone finds a particular problem, please open an issue!

PR #1763, opened 2018-08-10
Alias to_type_signature to graphql_name in wrapping types

RobertWSaunders:
## Proposal

Closes: #1711

Most types respond to #graphql_name and #description, but not the new list and non-null wrappers. It is in spec that wrapping types do have a name and a description so our wrapping types should implement them. This is a good opportunity as well to consolidate naming, as mentioned in the attached issue it would be nice to alias `to_type_signature` to `graphql_name`.

## Why

Adhere to spec and reduce the number of names to remember.

## How

Remove `to_type_signature` and alias it to `graphql_name`, also implement `description` in wrapping types.

### DIFF
--- a/lib/graphql/schema/list.rb
+++ b/lib/graphql/schema/list.rb
@@ -20,9 +20,11 @@ module GraphQL
         true
       end
 
-      def to_type_signature
-        "[#{@of_type.to_type_signature}]"
+      def graphql_name
+        "[#{@of_type.graphql_name}]"
       end
+
+      alias_method :to_type_signature, :graphql_name
     end
   end
 end
--- a/lib/graphql/schema/member/type_system_helpers.rb
+++ b/lib/graphql/schema/member/type_system_helpers.rb
@@ -24,10 +24,6 @@ module GraphQL
           false
         end
 
-        def to_type_signature
-          graphql_name
-        end
-
         # @return [GraphQL::TypeKinds::TypeKind]
         def kind
           raise NotImplementedError
--- a/lib/graphql/schema/non_null.rb
+++ b/lib/graphql/schema/non_null.rb
@@ -25,9 +25,11 @@ module GraphQL
         @of_type.list?
       end
       
-      def to_type_signature
-        "#{@of_type.to_type_signature}!"
+      def graphql_name
+        "#{@of_type.graphql_name}!"
       end
+
+      alias_method :to_type_signature, :graphql_name
     end
   end
 end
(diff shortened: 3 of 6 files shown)


lib/graphql/schema/list.rb
```diff
@@ -20,9 +20,11 @@ def list?
         true
       end
 
-      def to_type_signature
-        "[#{@of_type.to_type_signature}]"
+      def graphql_name
+        "[#{@of_type.graphql_name}]"
       end
+
+      alias_method :to_type_signature, :graphql_name
```
RobertWSaunders: I tried putting this in `GraphQL::Schema::Wrapper` to remove code duplication but it would alias to the `graphql_name` in `GraphQL::Schema::Wrapper` and as a result raise a `NotImplementedError`. It looks like this is  how `alias_method` works, just kind of sucks. I might be wrong though, do you know otherwise?

lib/graphql/schema/member/type_system_helpers.rb
```diff
@@ -24,10 +24,6 @@ def list?
           false
         end
 
-        def to_type_signature
```
RobertWSaunders: Figured I could just remove this because no other places were reliant on it!

swalkinshaw:
🤔 not sure if this is less confusing or more. I think it's fine if wrapping types have `graphql_name` but I kind of assumed they'd just be `nil` since that matches the spec (`name` is a nullable field on all `__Type`s).

rmosolgo:
Yeah, that's a good question, we've been down this road before: https://github.com/rmosolgo/graphql-ruby/pull/355

Still, I want _some_ method to turn `List.new(T)` into `"[T]"` that also turns `T` into `"T"` 🤔

swalkinshaw:
What about the string representation `to_s`? I think those exist on the old legacy DSL types.

rmosolgo:
I think that would override `Class#to_s`, which could be a bit confusing. Personally I've _enjoyed_ having a `.to_s` that tells me what Ruby object it is. 

Another option is to go the _other_ way, and add `#to_type_signature` to the other types which calls `graphql_name`.

swalkinshaw:
I agree. Having every type respond to `to_type_signature` and `graphql_name` probably sounds like the best bet (with the wrapping types returning `nil` for `graphql_name`).

RobertWSaunders:
I like it, I will update this when I get a chance!

rmosolgo:
Feel free to reopen this if you get a chance to work on it again!

PR #1764, opened 2018-08-10
Refactored integration tests and dependencies

dblock:
Coming from https://github.com/rmosolgo/graphql-ruby/pull/1548.

- Do not require Rails or Mongoid by default
- Moved any integration tests that rely on Rails or Mongoid into `spec/integration/rails|mongoid`
- Updated .travis.yml to only use the `include` matrix which is much easier to maintain than excludes
- Updated .travis.yml to run system tests explicitly vs. via Rake that caused conflicts
- Extracted a gemfile for postgresql
- Rely on bundler to load dependencies everywhere via `Bundler.require` instead of if/then/else requires
- Get rid of `WITHOUT_RAILS` and `rails_should_be_installed?`, this is now just the default Gemfile
- Get rid of `MONGO_DETECTED`, this relies on having integration specs in `spec/integration/mongoid`

After this is merged (comments appreciated):

- Put back jekyll plugins. Is this really needed? It makes ruby versioning more complicated as they require > 2.3.0.  Then figure out a better way to deal with `group :jekyll_plugins` being picked up by appraisals.
- Do we actually care about running tests against PG? What code in this library is different in that scenario?
- Do we care about JRuby? If so it can be added to Travis.
- Fix https://github.com/rmosolgo/graphql-ruby/issues/1765

PS: I had my kids (7 and 9) watch Star Trek last night because of this ...


dblock:
@rmosolgo This is ready to review/merge.

dblock:
I put the Jekyll plugin dependencies into `.gemspec` with a ruby version check, I think that's OK for now. I would deprecate ruby 2.2.8 and make the minimum 2.3.0 at some point soon to make life easier. Also I think that Jekyll build should be in Travis at least for the generation part since it's so sensitive, but leaving this for later.

Leaving JRuby alone. It will be easier to fix now since there's a bunch of unrelated things out of the way.

rmosolgo:
Thanks again for putting some 💪  into making this better! 

All the changes sound good to me, I just want to make sure the CI system is equivalent before/after. Here's literally side-by-side of before and after: 

image

- [x] 3 versions of mongoid 
- [x] earliest supported Ruby version (2.2)
- [x] latest supported Ruby version (2.4 😢 )
- [x] system tests (`The command "cd spec/dummy && bundle exec rails test:system" exited with 0.`)
- [x] postgres (`ENV["DATABASE"] == "POSTGRESQL"` seems to be used 
- [x] without Rails (the test with just `Gemfile` lists no integrations and runs fewer tests)

Looks good to me!

PR #1766, opened 2018-08-10
Add .path method to schema members for debugging

rmosolgo:
I've done stuff like this locally a lot, it would be nice to have something built-in

### DIFF
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -108,6 +108,8 @@ module GraphQL
           end
         end
 
+
+
         private
 
         def find_inherited_method(method_name, default_value)
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -108,8 +108,6 @@ module GraphQL
           end
         end
 
-
-
         private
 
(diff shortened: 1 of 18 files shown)


lib/graphql/schema/member/has_path.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module GraphQL
+  class Schema
+    class Member
+      module HasPath
+        # @return [String] A description of this member's place in the GraphQL schema
+        def path
```
swalkinshaw: Can this be memoized?

lib/graphql/schema/member/base_dsl_methods.rb
```diff
@@ -108,6 +108,8 @@ def authorized?(object, context)
           end
         end
 
+
```
swalkinshaw: ✂️

lib/graphql/schema/member/has_path.rb
rmosolgo: 🤔  Maybe, probably, but I'm not that sure. Right now, we don't _force_ anything to be frozen, this would be one of the first things in the class-based API to be cached. Do you think it will be a performance bottleneck?

rmosolgo:
Thanks for taking a look, glad I'm not too much off my rocker :P

PR #1781, opened 2018-08-15
Enums can only be used for enum type arguments

### LINKED ISSUE #1772: Bare-word names are parsing as strings
Consider a query like
```
{
  node(id: garbage) {
    id
  }
}
```

GraphiQL red-underlines this as invalid syntax (since IDs are strings and should be quoted). On graphql-js similarly you get back a top-level error "Expected type ID!, found garbage".

In graphql-ruby (1.8.7) it executes the query with `id` given the value of `"garbage"`. As far as I can tell this is not spec-compliant.

cc  @bernardoamc @swalkinshaw

cjoudrey:
Fixes #1772

As reported in #1772, enum values can currently be used as a value for String-like arguments.

For example, in the following query, the enum value `garbage` was specified for a `id: ID!` argument and ends up getting coerced to `"garbage"`.

```graphql
{
  node(id: garbage) {
    id
  }
}
```

This is disallowed [as per the spec](http://facebook.github.io/graphql/June2018/#sec-Enum-Value):

> Enum values are only used in contexts where the precise enumeration type is known. Therefore it’s not necessary to supply an enumeration type name in the literal.

It appears like this issue was introduced in #1398 when adding support for `JSON` scalar types.

The fix I'm proposing breaks an untested `JSON` scalar case, but I'm not sure it was intended to work in the first place:

```ruby
    it "can be JSON with an enum" do
      query_str = <<-GRAPHQL
      {
        echoJson(input: WOODWIND)
      }
      GRAPHQL

      res = Jazz::Schema.execute(query_str)
      assert_equal("WOODWIND", res["data"]["echoJson"])
    end
```

On a related note, I'm not sure how I feel about this test case:

```ruby
    it "can be JSON with a nested enum" do
      query_str = <<-GRAPHQL
      {
        echoJson(input: [{foo: WOODWIND}])
      }
      GRAPHQL

      res = Jazz::Schema.execute(query_str)
      assert_equal([{"foo" => "WOODWIND"}], res["data"]["echoJson"])
    end
```

I'm all for supporting a `JSON` scalar, but I don't think we should be allowing enum values within them because they are ambiguous and it means adding exceptional code paths that bypass the spec. As far as I can tell, the inspiration for this feature ([`graphql-type-json`](https://github.com/taion/graphql-type-json/blob/5790abc351b3007dc9dc98ccf35e9ea72ed7ba91/src/index.js#L8)) doesn't support enum values either.

cc @jturkel as you may have opinions about this PR.

### DIFF
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -117,8 +117,6 @@ module GraphQL
         value.to_h
       elsif value.is_a?(Array)
         value.map { |element| raw_coercion_input(element) }
-      elsif value.is_a?(GraphQL::Language::Nodes::Enum)
-        value.name
       else
         value
       end
--- a/spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
+++ b/spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
@@ -69,7 +69,7 @@ describe GraphQL::StaticValidation::ArgumentLiteralsAreCompatible do
 
   describe "enum value" do
     it "can't be used for an ID argument" do
-      schema_sdl = <<~GRAPHQL
+      schema_sdl = <<-GRAPHQL
         type Node {
           id: ID!
         }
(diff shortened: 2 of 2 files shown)


cjoudrey:
@jturkel Thanks for the confirmation. :smile: I'll take a closer look later this week.

PR #1784, opened 2018-08-15
Add  a schema-level hook for lazy/batched resolution

rmosolgo:
GraphQL-Ruby's lazy-resolve system is OK, but it has a high overhead without much benefit from its generality. Most of the time, a schema has only one type of lazy object (batch-loader or graphql-batch), and the rest of the time, all the method lookup stuff is overcomplicated and a waste of time. 

So, what if we exposed some schema-level methods that could be overridden as needed? In fact, we've already done this in GitHub as a performance consideration (we overrode `#lazy_method_name`).

TODO: 

- [x] test that overriding works 

cc @exAspArk

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -146,7 +146,7 @@ module GraphQL
             field_ctx.value = Execution::Lazy.new {
               inner_value = begin
                   begin
-                    field_ctx.schema.sync_lazy(raw_value, field_ctx.query.context)
+                    field_ctx.schema.sync_lazy(raw_value)
                   rescue GraphQL::UnauthorizedError => err
                     field_ctx.schema.unauthorized_object(err)
                   end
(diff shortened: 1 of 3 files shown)


exAspArk:
👍 

Just as an idea, here is how I imagine it could look like:

```ruby
class PostType < GraphQL::Schema::Object
  field :id, ID, null: false
  field :comments, [CommentType], null: true, lazy: true
  #                                           ^
  # an explicit flag, no more matching by predefined classes to detect lazy objects
  # duck typing, yay!
end

  query QueryType

  def lazy_resolve(value) # a new API method with a default implementation, can be overridden
    case value
    when MyBatchClass
      value.sync # sort of existing behavior
    else
      value.resolve_all if value.quacks? # custom behavior if necessary
    end
  end
end
```

rmosolgo:
@exAspArk , I think the difference between the code on this branch is the addition of `lazy: true`, is that right? 

I don't think I want to add field-level opting in, because we have various execution wrappers that may or may not wrap eager values in lazy objects. So, it can be hard to know which fields return lazy values under which conditions!

PR #1787, opened 2018-08-16
support callable prepare objects in Schema::Argument (fixes #1786)

### LINKED ISSUE #1786: class-based argument prepare doesn't support callable objects
in the old api, something like this works:
```
class MyPreparer
  def call(arg, ctx)
   # ...
   end
end

# then somewhere else
argument :id, !types.ID, "Vehicle Model ID", prepare: MyPreparer.new
```

but in the class based api it seems that `prepare` must be a `Proc`:
https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/argument.rb#L99

would you accept a pr that allows anything that responds to `#call` here or was this decision intentional? 

thanks!

modosc:


rmosolgo:
Thanks for the fix here! I'm going to change it on master to use `is_a?`/`respond_to?`, for performance: 

```ruby
# bench.rb
require "benchmark/ips"

def with_proc(callable)
  case callable
  when String, Symbol
    callable
  when ->(x) { x.respond_to?(:call) }
    callable
  else
    nil
  end
end

def with_respond_to(callable)
  if callable.is_a?(String) || callable.is_a?(Symbol)
    callable
  elsif callable.respond_to?(:call)
    callable
  else
    nil
  end
end

callable_input = ->(x) { x }

Benchmark.ips do |x|
  x.report("proc condition") { with_proc(callable_input) }
  x.report("respond_to?/is_a?") { with_respond_to(callable_input) }
  x.compare!
end
```

```
$ ruby bench.rb
      proc condition    85.632k i/100ms
   respond_to?/is_a?   210.167k i/100ms
      proc condition      1.196M (± 3.5%) i/s -      5.994M in   5.020018s
   respond_to?/is_a?      5.023M (± 3.7%) i/s -     25.220M in   5.027962s

Comparison:
   respond_to?/is_a?:  5023030.1 i/s
      proc condition:  1195574.0 i/s - 4.20x  slower
```

PR #1792, opened 2018-08-20
Ensure enum values respond to graphql_name

eapache:
Everything else in the schema does, and these should too. They don't
automatically because they don't inherit from BaseType, which they
potentially should, but that's a much larger change.

I couldn't get alias_method to work (something about how `name` is
defined dynamically through method_missing) so I defined the method
manually.

@rmosolgo cc @swalkinshaw there's a weird goose chase behind this one


rmosolgo:
>  weird goose chase 

😩  i bet! thanks for the fix.

PR #1797, opened 2018-08-21
Add support for array of IDs to `loads:` argument

cjoudrey:
Currently it is possible to have GraphQL Ruby automatically load objects for `ID` arguments, for example:

```ruby
argument :label_id, ID, required: true, loads: Label
```

In some cases, the argument will be a list of `ID`. It would be great if the `loads:` option worked in this case too, for example:

```ruby
argument :label_ids, [ID], required: true, loads: Label, as: :labels
```

I managed to do this by re-using the existing logic for `loads:`. I _think_ this is fine, but I'd love an extra pair of :eyes: in case I didn't think of something.

Another note, in the singular case, the `as:` is inferred by stripping `_id` off the original argument name. In the list case, this is a bit trickier as we'd have to pluralize the remainder, which would mean having to leverage something like [`ActiveSupport::Inflector#pluralize`](https://github.com/rails/rails/blob/c03dde6117ed995d1ada57736b5b0d44dfe31ed5/activesupport/lib/active_support/inflector/methods.rb#L19-L33). I think this might be overkill, so I didn't bother trying to fix this.

**Edit:** If you're happy with this, I'll add some documentation to https://github.com/rmosolgo/graphql-ruby/blob/b1c376f2bccbf0f3148155474237dcdd616b74dd/guides/mutations/mutation_classes.md.

### DIFF
--- a/spec/graphql/schema/resolver_spec.rb
+++ b/spec/graphql/schema/resolver_spec.rb
@@ -192,6 +192,23 @@ describe GraphQL::Schema::Resolver do
       end
     end
 
+    class PrepResolver9Array < BaseResolver
+      argument :int_ids, [ID], required: true, loads: HasValue, as: :ints
+      # Make sure the lazy object is resolved properly:
+      type [HasValue], null: false
+      def object_from_id(type, id, ctx)
+        # Make sure a lazy object is handled appropriately
+        LazyBlock.new {
+          # Make sure that the right type ends up here
+          id.to_i + type.graphql_name.length
+        }
+      end
+
+      def resolve(ints:)
+        ints.map { |int| int * 3}
+      end
+    end
+
     class PrepResolver10 < BaseResolver
       argument :int1, Integer, required: true
       argument :int2, Integer, required: true, as: :integer_2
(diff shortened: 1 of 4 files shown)


spec/graphql/schema/resolver_spec.rb
```diff
@@ -509,6 +527,14 @@ def add_error_assertions(field_name, description)
         # (5 + 8) * 3
         assert_equal 39, res["data"]["prepResolver9"]["value"]
       end
+
+      it "supports loading array of ids" do
+        res = exec_query('{ prepResolver9Array(intIds: ["1", "10", "100"]) { value } }')
+        # (1 + 8) * 3
+        # (10 + 8) * 3
+        # (100 + 8) * 3
+        assert_equal [27, 54, 324], res["data"]["prepResolver9Array"].map { |v| v["value"] }
```
cjoudrey: I'm not a huge fan of the tests in this file because I think some are a bit unclear (including this one), but I added it anyway.

I trust the tests in `spec/graphql/schema/relay_classic_mutation_spec.rb` a bit more. :smile:

rmosolgo: 😅

cjoudrey:
> Implementation looks good to me, but how about supporting the default case of `label_ids` -> `labels`? Yes, it will be wrong sometimes, but people can use `as:` in that case, right?

If that's what you prefer, I'll make the change. :smile:

I was erring on the side of making it wrong all the time, because if we fix it for one case (i.e. naively appending `s`), someone is bound to open an issue for the other cases that will be broken: https://github.com/rails/rails/blob/c03dde6117ed995d1ada57736b5b0d44dfe31ed5/activesupport/lib/active_support/inflections.rb#L12

cjoudrey:
@rmosolgo should be good now. I strip `_ids` and if the argument doesn't already end with `s`, I append one.

rmosolgo:
💰 !

PR #1800, opened 2018-08-22
Use Rufo in GraphQL-Ruby

rmosolgo:
I'm interested in using https://github.com/ruby-formatter/rufo to standardize style, I wonder how it will work in this project.


lib/graphql/analysis/max_query_depth.rb
```diff
@@ -12,7 +13,7 @@ module Analysis
     #
     class MaxQueryDepth < GraphQL::Analysis::QueryDepth
       def initialize(max_depth)
-        disallow_excessive_depth = ->(query, depth) {
+        disallow_excessive_depth = -> (query, depth) {
```
rmosolgo: The no-space Procs were originially introduced because someone wanted to support Ruby 1.9. AFAIK they gave up, and don't actually use GraphQL-Ruby after all. So this will be a big change, but probably OK.

lib/graphql/backtrace/traced_error.rb
```diff
@@ -35,12 +35,12 @@ def initialize(err, current_ctx)
...
-          "... and 1 more line\n"
-        else
-          "... and #{cause_backtrace_remainder_length} more lines\n"
-        end
+                                 ""
+                               elsif cause_backtrace_remainder_length == 1
+                                 "... and 1 more line\n"
+                               else
+                                 "... and #{cause_backtrace_remainder_length} more lines\n"
+                               end
```
rmosolgo: Personally I don't like it but w/e

lib/graphql/compatibility/query_parser_specification.rb
```diff
@@ -116,40 +116,40 @@ def test_it_doesnt_parse_nonsense_variables
...
           end
 
           def test_enum_value_definitions_have_a_position
-            document = parse("""
+            document = parse("" "
               enum Enum {
                 VALUE
               }
-            """)
+            " "")
```
rmosolgo: Wow, I guess this is just a funny bug? I think the person was imagining Python or GraphQL triple-quotes, and it happened to work because of a rarely-used string concatenation feature: 

```ruby
$ irb
irb(main):001:0> "abc" "def"
# => "abcdef"
```

eapache: 🤔 ruby strings aren't multiline by default though, so how would this work?

phaedryx:
I'm not familiar with Rufo, but it looks like a more opinionated Rubocop?

Is there a reason you prefer this over Rubocop?

rmosolgo:
> prefer this 

only because `@asterite` made it 😆  

Yeah, I'm not too crazy about its indentation rules, so I think I may just lean more on Rubocop instead.

Issue #1803, opened 2018-08-23
Infinite resolution with instrumenters inspecting results when an exception not inheriting from StandardError 


tylerhou:
https://github.com/rmosolgo/graphql-ruby/issues/1072 was a previous version of this bug.

If an exception not inheriting from `StandardError` is thrown in a GraphQL mutation, then the [rescue block added to fix #1072](https://github.com/rmosolgo/graphql-ruby/blob/0b40b5b5766508180611aa14ad1f504e1fc3e00d/lib/graphql/execution/multiplex.rb#L94) will not catch that exception. If such an exception is always thrown, the query will never be marked as resolved, and the schema resolver will never terminate. This could happen in development if you make a syntax error inside a mutation, for example.

I've been able to reproduce this bug on 1.7.9. I haven't attempted to reproduce on later releases (1.8.x/master), but I'm pretty certain that this bug exists.

Proposed fix is to change [line 94](https://github.com/rmosolgo/graphql-ruby/blob/0b40b5b5766508180611aa14ad1f504e1fc3e00d/lib/graphql/execution/multiplex.rb#L94) to rescue from `Exception` instead. Although it is generally frowned upon to rescue from `Exception`, in this case I think it is okay since we reraise.

rmosolgo:
:+1: to your proposed fix!

tylerhou:
https://github.com/rmosolgo/graphql-ruby/pull/1804

tylerhou:
Closed by #1804.

PR #1804, opened 2018-08-24
Fix infinite resolution bug for Exception (#1803)

### LINKED ISSUE #1072: Infinite resolution when instrumenter inspects query result

tylerhou:
If an exception not inheriting from StandardError was thrown in a
GraphQL mutation, then the rescue block added to fix #1072 in 8bf0039
would not have caught that exception. If such an exception was always
thrown, the query would have never been marked as resolved, and so the
schema resolver would have never terminated with certain instrumenting
gems that inspect `query.result`. This could have happened in
development if you made a syntax error inside a mutation, for example.

Rescuing from Exception fixes this. Although it is generally not advised
to rescue from Exception, it is probably okay in this case because we
immediately re-raise it.

### DIFF
--- a/lib/graphql/execution/multiplex.rb
+++ b/lib/graphql/execution/multiplex.rb
@@ -91,7 +91,7 @@ module GraphQL
             # Get the Query::Result, not the Hash
             query.result
           end
-        rescue StandardError
+        rescue Exception
           # Assign values here so that the query's `@executed` becomes true
           queries.map { |q| q.result_values ||= {} }
           raise
--- a/spec/graphql/execution/multiplex_spec.rb
+++ b/spec/graphql/execution/multiplex_spec.rb
@@ -160,12 +160,27 @@ describe GraphQL::Execution::Multiplex do
           raise GraphQL::Error, "Crash"
         }
       end
+
+      field :raiseSyntaxError, types.String do
+        resolve ->(object, args, ctx) {
+          raise SyntaxError
+        }
+      end
+
+      field :raiseException, types.String do
+        resolve ->(object, args, ctx) {
+          raise Exception
+        }
+      end
+
     end
 
     InspectSchema = GraphQL::Schema.define do
       query InspectQueryType
       instrument(:query, InspectQueryInstrumentation)
     end
+
+    unhandled_err_json = '{}'
 
     it "can access the query results" do
       InspectSchema.execute("{ raiseExecutionError }")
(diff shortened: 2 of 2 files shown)


rmosolgo:
Thanks for taking care of this, and for the nice tests!

PR #1806, opened 2018-08-24
Suggest splitting off a 2.0

qrush:
Hey @rmosolgo, I was looking through the upgrade guide for the class-based API and was very surprised to see  a suggestion that the `.define` API would be outright removed in 1.x. I'd strongly suggest keeping a major removal/API change like that to a major version bump instead. I think as a downstream consumer of this gem I'd be quite surprised to see that happen in a minor version and the major version gives you a chance to celebrate the new change.

I've started you off here on the roadmap. 🙏


rmosolgo:
That sounds like a nice suggestion, thanks :)

PR #1808, opened 2018-08-25
Allow fields to pass additional parameters

phaedryx:
Before the 1.8 update you could use closures to pass information from a field to it's type. Now it's much more difficult (using `:method`).

Example from our codebase:

```ruby
class SomeType < GraphQL::Schema::Object
...
  # a dynamically generated field
  field generated_name, type, method: :resolver, extras: [:ast_node]

...
  def resolver(:ast_node)
    generated_name = parse_name(ast_node.name)
    # do stuff based on the name

```

This pulled request would allow this:

```ruby
# a dynamically generated field
field generated_name, type, method: :resolver, extras: [:name]
...
def resolver(name:)
  # do stuff based on the name
```

or even this:
```ruby
class SpecialField < GraphQL::Schema::Field
  attr_reader :klass, :attribute

  def initialize(*args, klass:, attribute:, **kwargs, &block)
    @klass = klass
    @attribute = attribute
...
end


...
class SomeType < GraphQL::Schema::Object
  field_class SpecialField

  field generated_name, type, klass: klass, attribute: attribute, method: :resolver, extras: [:klass, :attribute]
...

  def resolver(klass:, attribute:)
    # resolve based on a class and attribute

```

### DIFF
--- a/guides/type_definitions/objects.md
+++ b/guides/type_definitions/objects.md
@@ -294,19 +294,21 @@ end
 
 At runtime, the requested runtime object will be passed to the field.
 
-### Field Parameter Default Values 
+__Custom extras__ are also possible. Any method on your field class can be passed to `extras: [...]`, and the value will be injected into the method. For example, `extras: [:owner]` will inject the object type who owns the field. Any new methods on your custom field class may be used, too.
 
-The field method requires you to pass `null:` keyword argument to determine whether the field is nullable or not. Another field you may want to overrid is `camelize`, which is `true` by default. You can override this behavior by adding a custom field. 
+### Field Parameter Default Values
+
+The field method requires you to pass `null:` keyword argument to determine whether the field is nullable or not. Another field you may want to overrid is `camelize`, which is `true` by default. You can override this behavior by adding a custom field.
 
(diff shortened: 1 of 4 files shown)


rmosolgo:
I really like this idea! The notion of injecting custom values into the method makes a lot of sense to me. How about this: 

- Extract the `public_send(extra)` logic into method, `GraphQL::Schema::Field#get_extra(extra)`, which people can override in their custom fields 
- Add some tests for overriding that method in the way you've shown here 

How does that sound?

phaedryx:
That's interesting. I'll rework the code a bit.

phaedryx:
Now that I've looked into it a bit, I'm not sure if I understand how your proposal would work. My attempts just made the code more awkward than what I have here.

rmosolgo:
Thanks for taking a crack at it! I pushed some code and docs to go along with how you started.

phaedryx:
okay, I will write up some tests for this this weekend

phaedryx:
Oh, I see that you added tests. Thank you.

rmosolgo:
Woops, this broke `path`

PR #1810, opened 2018-08-27
Improve field type string-to-constant resolution

justisb:
Currently string arguments passed to a type definition (such as `field :user, "Types::UserType", null: true`) are resolved using a basic `Object.const_get` call. This is prone to autoloading issues, particularly in development with classes down several module levels (e.g., `Types::User::Foo::BarType` where `Foo` is also a type).

Rails' `ActiveSupport::Inflector.constantize` has a few workarounds built-in, which this PR grifts in `BuildType#parse_type` for more robust string-to-constant resolution, without assuming the project has `String.constantize` available.

A more complex solution could be allowing a custom string-to-constant resolver as an option on the schema, but that seemed like overkill. On the other hand a simpler solution could be checking that the type string `respond_to?(:constantize)` with a fallback to `Object.get_const`, though that assumes the constantize extension would always behave as expected.

### DIFF
--- a/lib/graphql/schema/member/build_type.rb
+++ b/lib/graphql/schema/member/build_type.rb
@@ -33,7 +33,7 @@ module GraphQL
               null = false
               parse_type(type_expr[0..-2], null: true)
             else
-              maybe_type = Object.const_get(type_expr)
+              maybe_type = constantize(type_expr)
               case maybe_type
               when GraphQL::BaseType
                 maybe_type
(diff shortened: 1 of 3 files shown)


lib/graphql/schema/member/build_type.rb
```diff
@@ -125,6 +125,39 @@ def camelize(string)
           camelized
         end
 
+        # Resolves constant from string (based on Rails `ActiveSupport::Inflector.constantize`)
+        def constantize(string)
+          camel_cased_string = camelize(string)
```
rmosolgo: I'm not sold we should auto-camelize. I think it's simpler to find a Ruby constant by name. What's the use case for auto-camelizing?

justisb:
Sure, I can add some tests to show that passing in a string value as the field type constantizes the value to the appropriate class.

lib/graphql/schema/member/build_type.rb
justisb: I suppose the idea is that it makes it more resilient to errant formatting in the class name (say they pass in `user` instead of `User`), but you're right in that it's unnecessary since that problem should be quickly discovered and fixed on its own.

justisb:
Finally was able to revisit this PR to add several parse_type tests, for both old and new functionality. Also removed the extra unnecessary camelization as requested.

FWIW I've been running this fork in production for the last couple of months without any issue.

rmosolgo:
Thanks for your work on this!!

PR #1824, opened 2018-08-30
Fields will merge to AST visitor

xuorig:
Converts the FieldsWillMerge validator to a "pure AST" validator, without using the rewrite. Most of the logic is based on the algorithm in graphql-js (https://github.com/graphql/graphql-js/blob/master/src/validation/rules/OverlappingFieldsCanBeMerged.js)

The algorithm is explained through comments in the class. 

  - [x] make the tests pass
  - [x] refactor / beautify logic
  - [x] performance optimizations
  - [x] Fix https://github.com/graphql/graphql-js/pull/1497 but here


xuorig:
Preliminary benchmarks:

Master:
```
validate - introspection
                        38.000  i/100ms
validate - abstract fragments
                        82.000  i/100ms
                        43.000  i/100ms
validate - introspection
                        394.693  (±11.7%) i/s -      1.938k in   5.004816s
validate - abstract fragments
                        721.422  (±16.5%) i/s -      3.526k in   5.053876s
                        439.557  (± 9.8%) i/s -      2.193k in   5.047115s
validate - big query     39.992  (± 7.5%) i/s -    200.000  in   5.036172s
```

This branch:
```
validate - introspection
                        38.000  i/100ms
validate - abstract fragments
                        82.000  i/100ms
                        43.000  i/100ms
validate - introspection
                        394.693  (±11.7%) i/s -      1.938k in   5.004816s
validate - abstract fragments
                        721.422  (±16.5%) i/s -      3.526k in   5.053876s
                        439.557  (± 9.8%) i/s -      2.193k in   5.047115s
validate - big query     39.992  (± 7.5%) i/s -    200.000  in   5.036172s
```

rmosolgo:
(For those keeping score at home, about 25% faster 😎 )

xuorig:
After some caching, hard to say if it's much better, but I think the benches we have don't necessarily require caches so that's probably why.

```
validate - introspection
                        26.000  i/100ms
validate - abstract fragments
                        71.000  i/100ms
                        39.000  i/100ms
validate - introspection
                        384.955  (±13.2%) i/s -      1.898k in   5.054420s
validate - abstract fragments
                        778.436  (±14.1%) i/s -      3.834k in   5.086308s
                        415.047  (±14.2%) i/s -      2.028k in   5.013408s
validate - big query     46.101  (±13.0%) i/s -    228.000  in   5.068996s
```

rmosolgo:
Yeah, I bet "big query" benefited the most, I think it has lots of fragments.

xuorig:
And another bench mark, I think this should be ready for review! 

```
validate - introspection
                        40.000  i/100ms
validate - abstract fragments
                        85.000  i/100ms
                        44.000  i/100ms
validate - introspection
                        417.224  (± 8.6%) i/s -      2.080k in   5.030352s
validate - abstract fragments
                        835.254  (± 8.7%) i/s -      4.165k in   5.034208s
                        446.764  (±10.5%) i/s -      2.244k in   5.096434s
validate - big query     43.860  (±16.0%) i/s -    216.000  in   5.086208s
```

lib/graphql/static_validation/rules/fields_will_merge.rb
```diff
@@ -2,48 +2,342 @@
...
+          end
+        end
+      end
 
-            if args.length > 1
-              msg = "Field '#{node.name}' has an argument conflict: #{args.map{ |arg| GraphQL::Language.serialize(arg) }.join(" or ")}?"
-              context.errors << GraphQL::StaticValidation::Message.new(msg, nodes: node.ast_nodes.to_a)
-            end
+      def fields_and_fragments_from_selection(node, parent_type:)
+        @fields_and_fragments_from_node[node] ||= begin
```
xuorig: Fields and Fragment Spreads are collected once per node, iterating over the list only one.

xuorig:
Great context here from graphql-js: https://github.com/graphql/graphql-js/pull/1497#issuecomment-418454489

Sounds like graphql-js doesn't handle these edge cases, and the spec doesn't seem to either 🤔

xuorig:
After trying a few things, I've just realized that:

```graphql
        {
          pet {
            ... on Dog {
              ... on Pet {
                name
              }
            }
            ... on Cat {
              name(surname: true)
            }
          }
        }
```

Used to be valid (Not sure how internal representation used to handle that case), but is not anymore. The algo probably has to maintain lists of parents instead of only the immediate parent of fields, in this case, `Pet` and `Cat` are not mutually exclusive (conflict), but looking at the grand parent `Dog` we should know that they are in fact exclusive. (no conflict)

rmosolgo:
The implementation of that was in `InternalRepresentation::Scope`, it kept track of which concrete types were allowed at a given point in the AST

rmosolgo:
(I'm surprised there wasn't a test for it!)

xuorig:
With new algorithm, minus caching for now. Not a terrible change performance wise it seems

```
validate - introspection
                        41.000  i/100ms
validate - abstract fragments
                        81.000  i/100ms
                        44.000  i/100ms
validate - big query     3.000  i/100ms
validate - introspection
                        408.896  (± 4.4%) i/s -      2.050k in   5.023692s
validate - abstract fragments
                        796.939  (± 7.7%) i/s -      3.969k in   5.014707s
                        448.745  (± 7.6%) i/s -      2.244k in   5.035781s
validate - big query     44.255  (± 4.5%) i/s -    222.000  in   5.022948s
```

rmosolgo:
what a relief, still the same huge improvement over master!

rmosolgo:
shoot, i really screwed up 1.9-dev, let me straighten it out and try to fix this branch.

rmosolgo:
Wow, I spent some time reading the code, thanks again for your amazing work here! 🎊  

An interesting take-away for me is how iterative this implementation is. I mean, the comparison really boils down to looping over two sets and comparing each member to each other member. I wouldn't have guessed that it would be so fast, but it sure is. I would have used a Hash, but as shown in any profiler output for GraphQL-Ruby, `Kernel#hash` is a big time-waster of ours! So maybe I should keep an eye out for applying this more often.  Maybe the Ruby VM is faster than the data-structure based approach. 

I have an outstanding question, but it doesn't block merging this PR. Let's say we have a query like this: 

```graphql
{
  a { x y } 
  b { x y } 
}
```

When we enter `a`, we'll deeply check it for conflicts, including checking its subselections (`{x y}`) against `b`'s subselections (`{x y}`). Is that right? 

Then, when we enter `b`, will we also compare _its_ subselections against `a`'s subselections? I see a cache for fragments in a case like this, I wonder if it's worth caching here, too. 

Also, I see that `@fields_and_fragments_from_node` is unused (hopefully not due to my bad git behavior 😩 ) so I will remove it before merging.

Thompson1985 (approved):

Issue #1826, opened 2018-08-30
Default values for Input Objects do not work with Enum types


nickblumenthal:
GraphQL v1.8.8

When you have an enum that has a different underlying value than its name, using that value as a default value for an input field in a mutation does not work.

Here's the example:

We have an enum that looks like this: 

```
AgentStateEnum = GraphQL::EnumType.define do
  name 'AgentStateEnum'
  description 'State of an Agent\'s account.'

  value 'ACTIVE', 'Represents an authorized agent in our system.', value: 'active'
  value 'DISABLED', 'Agent is disabled, web app access is denied.', value: 'disabled'
  value 'ROSTER_AGENT', 'Agent exists in the system, awaiting activation.', value: 'roster_agent'
end
```

and a mutation like this:

```
CreateAgentMutation = GraphQL::Relay::Mutation.define do
  name 'CreateAgent'
  description 'Create an agent. Automatically associated with the currently authenticated access token owner.'

  # Input fields
  input_field :email, !types.String,
    'Email address of Agent. Update with care as this also acts as the agent\'s login. Must be a properly formatted email address.'
  input_field :state, AgentStateEnum, 'State of the Agent\'s account. Defaults to "active".', default_value: 'active'

  # Return fields
  return_field :agent, AgentType
  resolve CreateAgentResolver.new
end
```

However, if we execute the mutation without a variable for `state`, that input is `nil` in the resolver (instead of `active` like you would expect).

This is due to https://github.com/rmosolgo/graphql-ruby/commit/4eaeb98472fab50dca321678cd4bd3a85b4e3472

That commit attempts to coerce the default value via the input field's type. Unfortunately, an Enum types coercion is done by name, not by value.  

I thought maybe I could just set the default_value to the name instead like so:
```
 input_field :state, AgentStateEnum, 'State of the Agent\'s account. Defaults to "active".', default_value: 'ACTIVE'
```

Unfortunately this does not pass validation because the validator ultimately calls `coerce_result` which compares the default value to the enumerator's underlying ruby value (https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/enum_type.rb#L113)

While a simple solution would be to change https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/input_object_type.rb#L96 to 
```
coerced_value = input_field_defn.type.coerce_input(input_field_defn.default_value, ctx) || input_field_defn.default_value
```

[...]

PR #1827, opened 2018-08-31
Fall back to input_field_defn default value if coerce_input returns nil

nickblumenthal:
This is a simple PR to solve: https://github.com/rmosolgo/graphql-ruby/issues/1826.  Basically it just falls back to the pre-existing behavior in case the lookup of default value in nested inputs fails as it does in the case of enum types.

### DIFF
--- a/lib/graphql/enum_type.rb
+++ b/lib/graphql/enum_type.rb
@@ -169,6 +169,10 @@ module GraphQL
     def coerce_non_null_input(value_name, ctx)
       if @values_by_name.key?(value_name)
         @values_by_name.fetch(value_name).value
+      elsif match_by_value = @values_by_name.find { |k, v| v.value == value_name }
+        # this is for matching default values, which are "inputs", but they're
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for tracking down this regression! I didn't feel totally sure that returning `nil` from coerce_input meant that the user-provided value should be used instead. So I just pushed a commit to try adding the behavior to `EnumType` specifically. Let's see if the tests pass 😬

nickblumenthal:
Sounds good to me.  I believe https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/validation.rb#L101 validates default inputs by underlying value, but this approach is conservative and scopes the behavior just to the `EnumType` which seem like a wise course! 👍

rmosolgo:
Oh, I see, so that validator would prevent anyone from _depending_ on returning `nil`, nice find. If I have trouble with this implementation, then I'll go back to your original one with full confidence 😊

PR #1829, opened 2018-09-03
schema_parser_specification.rb fixes

vbrazo:
This gem is awesome. Thanks for maintaining it.

I was reading the source code early on and noticed several duplications in the `schema_parser_specification.rb`. 

I created a few methods and started using them. I decided to open this PR to see if you believe this idea would be a good one.

### DIFF
--- a/lib/graphql/compatibility/schema_parser_specification.rb
+++ b/lib/graphql/compatibility/schema_parser_specification.rb
@@ -15,13 +15,13 @@ module GraphQL
             @@parse_fn.call(query_string)
           end
 
-          def assert_directive(extension, object, value, name)
-            assert_equal value, extension.directives.length
-            assert_equal object, extension.directives.first.class
-            assert_equal name, extension.directives.first.name
+          def assert_extension(extension_object, object, value, name)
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the suggestion, but in fact, I prefer the old way. There are two reasons for me: 

- Any reader can understand `assert_equal` without searching for the method source, which makes it easier to maintain these tests or modify them as a new contributor. Although the new method removes some duplication, it makes it harder to quickly understand the tests. 
- Minitest will report the test failure on the most-recently-called method whose name begins with `assert`, so if the new tests fail, they'll report a failure on the line of the new `assert_*` helper. As a maintainer, this is not helpful because I need to know which _sub-assertion_ failed so that I can continue to debug the failure. In this way, the new `assert_*` helper obscures the details of the failure, but those details are essential for me. This issue could be addressed by renaming the helpers _not_ to begin with `assert*` or to add some kind of detailed error message in case of a failure. But this would not address the previous point. 

So, thanks for taking a look and making a suggestion, but for now, I'd prefer to keep the simple approach, even though the downside is some duplication!

PR #1832, opened 2018-09-03
Fix middleware warning

vbrazo:
Just a quick fix that was marked in the codebase as `TODO:`

`Middleware that takes a next_middleware parameter is deprecated (GraphQL::Compatibility::ExecutionSpecification::SpecificationSchema::TestMiddleware); instead, accept a block and use yield.`


rmosolgo:
Thanks :)

PR #1841, opened 2018-09-07
Big WIP: spike at AST analyzers

xuorig:


rmosolgo:
Some quick feedback here: it's important that we keep supporting existing Analysis for people who don't opt into the new interpreter, so new-style analyzers should be implemented as a brand new code path, leaving the existing code in place.

guides/language_tools/visitor.md
```diff
@@ -0,0 +1,143 @@
...
+Inserting nodes is similar to modifying nodes. To insert a new child into `node`, call one of its `.add_` helpers. This returns a copied node with a new child added. For example, to add a selection to a field's selection set:
+
+```ruby
+def on_field(node, parent)
+  node_with_selection = node.add_selection(name: "emailAddress")
+  super(node_with_selection, parent)
+end
+```
+
+This will add `emailAddress` the fields selection on `node`.
```
RobertWSaunders: Think you are missing "to" before "the"!

rmosolgo: I think my screwing with `1.9-dev` makes this look like part of xuorig's diff but in fact it's from a while ago :S

RobertWSaunders: True, well this applies for that then 🤣

spec/graphql/analysis/query_depth_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require "spec_helper"
 
-describe GraphQL::Analysis::QueryDepth do
+describe GraphQL::Analysis::AST::QueryDepth do
```
rmosolgo: Should this have `AST::`?

lib/graphql/analysis/ast/visitor.rb
```diff
@@ -0,0 +1,228 @@
...
+        def on_operation_definition(node, parent)
+          object_type = @schema.root_type_for_operation(node.operation_type)
+          @object_types.push(object_type)
+          @path.push("#{node.operation_type}#{node.name ? " #{node.name}" : ""}")
+          call_analyzers(:on_enter_operation_definition, node, parent)
+          super
+          call_analyzers(:on_leave_operation_definition, node, parent)
+          @object_types.pop
+          @path.pop
+        end
```
rmosolgo: This reminds me of a bug in the current analyzer code which we can fix on this one, in a later PR maybe. 

Currently we visit _all_ operation definitions. But we should probably only visit `query.selected_operation`.

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -0,0 +1,116 @@
...
+    # Calculate the complexity of a query, using {Field#complexity} values.
+    module AST
+      class QueryComplexity < Analyzer
+        # State for the query complexity calcuation:
+        # - `complexities_on_type` holds complexity scores for each type in an IRep node
+        def initialize(query)
+          super
+          @complexities_on_type = [TypeComplexity.new]
+          @skip_stack = [false]
+          @in_fragment_def = false
```
rmosolgo: It looks like some of these IVars were moved to the parent class, is that right?

lib/graphql/analysis/ast/visitor.rb
```diff
@@ -0,0 +1,228 @@
...
+          else
+            object_types.last
+          end
+
+          object_types << object_type
+
+          fragment_def.selections.each do |selection|
+            visit_node(selection, fragment_def)
+          end
+        end
```
rmosolgo: cool

lib/graphql/analysis/ast/query_depth.rb
```diff
@@ -0,0 +1,66 @@
...
+          end
+        end
+
+        def on_enter_fragment_spread(node, _, visitor)
+          visitor.enter_fragment_spread_inline(node)
+        end
+
+        def on_leave_fragment_spread(node, _, visitor)
+          visitor.leave_fragment_spread_inline(node)
+        end
```
rmosolgo: It'll be interesting to see if we need this more, maybe a kind of base class that defaults to this handling of fragments will come in handy.

lib/graphql/analysis/ast.rb
```diff
@@ -0,0 +1,71 @@
...
+          end
+        end
+        nil
+      end
+
+      # @param query [GraphQL::Query]
+      # @param analyzers [Array<#call>] Objects that respond to `#call(memo, visit_type, irep_node)`
+      # @return [Array<Any>] Results from those analyzers
+      def analyze_query(query, analyzers, multiplex_analyzers: [])
+        query.trace("analyze_query", { query: query }) do
```
swalkinshaw: Could we trace each individual analyzer as well?

rmosolgo: The thing is, they're called all together, during one visit. They don't make  sequential visits. I did that originally because I thought it would be faster, but if I'm wrong, or if it's not worth the tradeoff, I'm open to reconsidering the design.

swalkinshaw: Oh I think that's fine then. If its all a single visit, then the overhead within a single analyzer would likely be more obvious (like calling a service or something which should be done after the query anyway?).

lib/graphql/analysis/ast/query_depth.rb
xuorig: 💯 Totally. I was unsure if we needed this now. The base class idea sounds good. If a more "raw" approach is needed they could use this method to do it, but the more common default definitly sounds more like visiting the fragment spreads inline.

lib/graphql/analysis/ast/visitor.rb
xuorig: 👍 fixed in this PR.

xuorig:
Worked on exposing this new API.

To keep the existing analysis working, it is possible to switch the analysis engine at the schema level. The new AST analysis engine is opt in, like this:


```ruby
  enable_ast_analysis
  query_analyzer MyASTAnalyzer 
end
```

The same hook is used to add AST analyzers or old analyzers, but the selected engine treats them how it wants. It might be good to eventually add a way to select the engine at the validation pipeline / query level. Might be useful to A/B test while migrating for example.

xuorig:
Tests mostly pass except a weird case with a `prepare` proc returning an execution error during analysis, not sure how this was handled before, will debug on monday.

lib/graphql/language/visitor.rb
```diff
@@ -96,41 +100,50 @@ def on_abstract_node(node, parent)
...
-      alias :on_scalar_type_definition :on_abstract_node
-      alias :on_scalar_type_extension :on_abstract_node
-      alias :on_schema_definition :on_abstract_node
-      alias :on_schema_extension :on_abstract_node
-      alias :on_type_name :on_abstract_node
-      alias :on_union_type_definition :on_abstract_node
-      alias :on_union_type_extension :on_abstract_node
-      alias :on_variable_definition :on_abstract_node
-      alias :on_variable_identifier :on_abstract_node
+      # Don't use make_visit_method becuase it breaks `super`
```
rmosolgo: This should probably be "Don't use _alias_ because it breaks `super`, right?

lib/graphql/schema.rb
```diff
@@ -159,6 +160,7 @@ def initialize
       @lazy_methods.set(GraphQL::Execution::Lazy, :value)
       @cursor_encoder = Base64Encoder
       # Default to the built-in execution strategy:
+      @analysis_engine = GraphQL::Analysis
```
rmosolgo: :+1: nice approach to generalizing this

.travis.yml
```diff
@@ -43,6 +43,7 @@ matrix:
   - env:
       - DISPLAY=':99.0'
       - TESTING_INTERPRETER=yes
+      - TESTING_AST_ANALYSIS=yes
```
rmosolgo: This test case only runs the ActionCable subscription tests from `spec/dummy`, did you mean to set this flag for those tests? I don't see the flag used by those tests.

spec/graphql/authorization_spec.rb
```diff
@@ -372,6 +372,9 @@ class Schema < GraphQL::Schema
       if TESTING_INTERPRETER
         use GraphQL::Execution::Interpreter
       end
+      if TESTING_AST_ANALYSIS
+        use GraphQL::Analysis::AST
+      end
```
rmosolgo: Should this be removed, since we're bailing on that analyzer? (If it should be removed, then I think we can remove TESTING_AST_ANALYSIS altogether, right?)

xuorig:
🍏 ! Ready for an other round or reviews @rmosolgo 🙏

PR #1843, opened 2018-09-07
Update how Schema::Field executes definition_block

swalkinshaw:
Previously this used `instance_exec` which sets `self` to the field instance.

This changes to `yield` which still executes within the context of the field instance (supporting the same functionality) but doesn't change `self`.

This is probably more relevant to our specific use case. We have resolve procs which call class methods on object type classes. As we're migrating more to use the gem directly, this is a blocker for us.

The irony being that we added this code in the first place 😂 

Note: this is technically a breaking change, but I (highly?) doubt it will affect anyone.

1. it's a fairly new feature
2. the common usage is still supported (since it yields the field instance)
3. no one else would have our use case of class methods on object classes?

cc @RobertWSaunders

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -186,7 +186,7 @@ module GraphQL
 
         if definition_block
           if definition_block.arity == 1
-            instance_exec(self, &definition_block)
+            yield self
           else
             instance_eval(&definition_block)
           end


RobertWSaunders:
Oops this is my bad, this probably makes some of those changes I was making easier! However, my change was based off of the existing `instance_eval` when the definition block does not have an arity of one. My only concern is that now the functionality is technically not the same as when the block does not have an arity of one, but as you mentioned it likely doesn't affect anyone yet.

rmosolgo:
Makes sense to me, I'll be sure to include migration notes in the changelog although I agree, I bet nobody will notice.

PR #1851, opened 2018-09-17
Passthrough `null` values for arguments using `loads`

### LINKED ISSUE #1840: `loads:` should return `nil` when provided an ID of `nil`
Sometimes you want to pass `thingId: nil` as a way of saying "there should be no `Thing` here". `loads:` bombs on this right now, but it should probably inject `thing: nil` into `#resolve` instead.

cjoudrey:
Fixes #1840 

If a nullable argument using `loads: ...` is provided a `null` value, we should passthrough the `nil` value to the resolver instead of trying to resolve it to an object.


rmosolgo:
👏 ! Thanks for fixing this oddity!

PR #1855, opened 2018-09-18
Raise execution error when trying to decode an invalid cursor

cheshire137:
A problem I hit recently is that I could run a query with an invalid cursor and instead of getting an error response back, I'd get an exception. This branch rescues that exception ("ArgumentError: invalid base64") and raises an ExecutionError, which results in a normal error response from GraphQL explaining that the caller provided a bad cursor.


rmosolgo:
This seems great, thanks for the fix!

PR #1857, opened 2018-09-19
[JS] ActionCableLink connection params

mcdave029:
* Added `connectionParams`

const wsLink = new ActionCableLink({
  cable,
  connectionParams: {
     // e.g. token: string_token
  }
})

Inspiration: [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws)

### DIFF
--- a/javascript_client/subscriptions/ActionCableLink.js
+++ b/javascript_client/subscriptions/ActionCableLink.js
@@ -6,7 +6,11 @@ function ActionCableLink(options) {
   var cable = options.cable
   var channelName = options.channelName || "GraphqlChannel"
   var actionName = options.actionName || "execute"
-  var connectionParams = options.connectionParams || {}
+  var connectionParams = options.connectionParams
+
+  if (typeof connectionParams !== "object") {
+    connectionParams = {}
+  }
 
   return new ApolloLink(function(operation) {
     return new Observable(function(observer) {
--- a/javascript_client/subscriptions/ActionCableLink.js
+++ b/javascript_client/subscriptions/ActionCableLink.js
@@ -6,15 +6,16 @@ function ActionCableLink(options) {
   var cable = options.cable
   var channelName = options.channelName || "GraphqlChannel"
   var actionName = options.actionName || "execute"
+  var connectionParams = options.connectionParams || {}
 
   return new ApolloLink(function(operation) {
     return new Observable(function(observer) {
       var channelId = Math.round(Date.now() + Math.random() * 100000).toString(16)
 
-      var subscription = cable.subscriptions.create({
+      var subscription = cable.subscriptions.create(Object.assign({},{
         channel: channelName,
         channelId: channelId
-      }, {
+      }, connectionParams), {
         connected: function() {
           this.perform(
             actionName,


rmosolgo:
Nice! Thanks for the improvements.

mlwilkerson:
@mcdave029 Where will this connectionParams be available on the server?

mlwilkerson:
I figured out a partial answer to my own question: I see that I have access to those `connectionParams` in the ActionCable _Channel_. So if I wanted to `reject` a Channel subscription,
I could do that like something this:

```ruby
def subscribed
  # ...
  token = params[:token]
  decoded_token = Auth::decode(token)
  reject unless User.find(decoded_token["user"])
end
```

However, by the time this channel subscription code runs, the `ActionCable::Connection::Base` has already been established. And that `Connection` seems to be the normal place for authenticating a user for ActionCable.

In fact, the `action-cable-react-jwt` npm works by [adding a JWT to the WebSocket protocols list](https://github.com/zekedran/action-cable-react-jwt/blob/f288512b4e1e65ae6eea5317a38b13681aae8611/action_cable_react_jwt.js#L244) so it shows up in the [`HTTP_SEC_WEBSOCKET_PROTOCOL` header](https://github.com/zekedran/action-cable-react-jwt/blob/master/README.md) and is thus available for authenticating the user in the Connection's `connect` method.

I'm trying to figure out whether this new `connectionsParam` feature will allow me to do authentication at the `Connection` level, like `action-cable-react-jwt`, but without hijacking that header.

mcdave029:
Nice @mlwilkerson thanks for sharing your ideas i will look into this 👍

casiodk:
I would like to do something like this https://www.youtube.com/watch?v=S51bd-dFTfs . I really miss some documentation on how to set it up with ActionCableLink

mlwilkerson:
@mcdave029 :
I've done a little more investigating here, focused on [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws), hacking on the [graphql-tutorial](https://github.com/apollographql/graphql-tutorial) that uses these bits. I'm trying to understand by analogy how or whether the `SubscriptionServer` and `SubscriptionClient` use that `connectionParams` option, and whether it contributes to some mechanism used _upstream_ for accepting or rejecting the initial WebSocket connection, or only further _downstream_ where a subscription is created on an already opened WebSocket connection.

My conclusion is that it works the same way there as what we now have here with this ActionCable version: `connectionParams` only works at the downstream level, and by default, any WebSocket connection is accepted at the upstream level.

According to the docs for the `ws` (WebSocket) package, the `connection` event is emitted [only after the initial handshake for the WebSocket is complete](https://github.com/websockets/ws/blob/master/doc/ws.md#event-connection).

The `SubscriptionServer` handles that `connection` event [here](https://github.com/apollographql/subscriptions-transport-ws/blob/e456cb50c5e3a8fa978bc65dcc9439188f5bbb86/src/server.ts#L138), which is where the `onSubscribe` handler is wired up. `onSubscribe` is the handler that ends up receiving that `connectionParams` object, so it's downstream of the initial WebSocket connection, at the point where a subscription is made. Thus, it's the same pattern there as we have here in these subscriptions over ActionCable.

In order to get upstream and possibly reject the WebSocket connection during handshake, a [`verifyClient` handler would have to be defined and passed as an option to the `WebSocket.Server` constructor](https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback). In the graphql-tutorial, the `verifyClient` handler would be defined [here](https://github.com/apollographql/graphql-tutorial/blob/480bdc651a32e4a4c58ebbcc7b98e77fe4661701/server/server.js#L42
). That handler would have access to the request headers, among other things, and could accept or reject the connection handshake (upstream) accordingly.

In the ActionCable scenario, the `ActionCable::Connection::Base` [`connect`](https://api.rubyonrails.org/classes/ActionCable/Connection/Base.html) method appears to be equivalent to that `verifyClient` handler.

The problem remains, though, how to send in anything from the client that is useful there. The `SubscriptionClient` calls the `WebSocket` constructor [here](https://github.com/apollographql/subscriptions-transport-ws/blob/e456cb50c5e3a8fa978bc65dcc9439188f5bbb86/src/client.ts#L539). You can see that it only provides the URL and the protocol, [which is just what the WebSocket API says it accepted here](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket).

The only vectors for sending in some authentication metadata (like the JWT) are that `url` or `protocol` param to the WebSocket constructor.

It's supposed to be the case that [basic auth works](https://stackoverflow.com/a/4361358/4642871), so, passing in a username and password in that `url`, like
```javascript
var ws = new WebSocket("ws://username:password@example.com")
```

An Authorization header would allegedly be built from that url, added to request headers, thus made  available to that `verifyClient` handler on the client. But that doesn't seem to be working as advertised for me: the Authorization request header is no where to be found. Not sure why. (I've tried it in Chrome and Firefox—maybe I'm overlooking something).

Process of elimination: that leaves the `protocol` param, which is what [`action-cable-react-jwt`](https://github.com/zekedran/action-cable-react-jwt/blob/f288512b4e1e65ae6eea5317a38b13681aae8611/action_cable_react_jwt.js#L244) uses.

Summary: with this `connectionParams` object on the ActionCable client, we do have equivalent functionality as the Apollo `SubscriptionServer` and `SubscriptionClient`. It's just that the upstream WebSocket handshake is normally accepted and it's the downstream subscription creation that is authenticated via `connectionParams`.

I'd still like to find a better way to lock down that upstream WebSocket handshake, but it's clear that I'll need to find a different approach than trying to make this `connectionParams` do the job.

mlwilkerson:
@mcdave029 Yes, the fact that `action-cable-react-jwt` just forks the `ActionCable` code, and thus cuts me off from updates to the original library, is a key reason I didn't want to keep using it.

What I've opted for instead is to leverage the `url` that is standard and well within the WebSocket spec. I don't want to pass the main JWT auth token on that URL since it will appear in console logs, but I've created a short-lived "access token" for that purpose that only authorizes the opening of a WebSocket connection. The client has to be able to authenticate with the server via JWT by the normal means in order to get that access token.

The challenge with this approach has been that the `ActionCable` code seems to assume that the cable url does not change. But using this method, it changes at least as often as the short-lived access token expires and a new connection is attempted.

It's easy to pass in a url with an access token, but more difficult to refresh the token on that url. In particular, the `ConnectionMonitor` logic is pretty well encapsulated: if it detects a stale connection, it will just keep trying to re-connect with the same url again and again.

So I've created my own small classes that extend `ActionCable.Consumer` and `ActionCable.Connection` which just make sure that the `url` has an updated access token on it before calling `open()` to establish a new connection. This now works well upon initial creation of the `ActionCable.Connection` and also works well with the `ConnectionMonitor` and `reopen()` logic.

Then I realized another limitation of this `connectionParams` feature. The way you've implemented on this PR, it's a one-time snapshot at the time of instantiating the `ActionCableLink` object. But what about when the JWT is refreshed? The only way to refresh the client's auth token would seem to be creating a new instance of `ActionCableLink` with a new `connectionParams` object containing the refreshed JWT. And it seems that the best way to refresh the `ApolloClient` with a new `ActionCableLink` is to just create a new `ApolloClient`. So it's a lot of tear down and re-build required just to refresh that auth token.

By contrast, the `subscriptions-transport-ws` that inspired this feature [allows `connectionParams` to be a function](https://github.com/apollographql/subscriptions-transport-ws/blob/e456cb50c5e3a8fa978bc65dcc9439188f5bbb86/src/client.ts#L291). If we did the same thing here, then each time the `ActionCableLink` processes an operation, it would be able to retrieve the auth token that is _current at that moment_ (from `localStorage`, probably), rather than always using the token given at the time of instantiation.

PR #1859, opened 2018-09-20
Regression in `extras: [:path]`

cjoudrey:
It appears like a regression was introduced in #1808.

Given the following code:

```ruby
require 'graphql'

  field :hello, String, null: true,
    extras: [:execution_errors, :path]

  def hello(execution_errors:, path:)
    p path
  end
end

  query Query
end

Schema.execute('{ alias: hello }')
```

Previously, `path` would return `["alias"]`.

As of v1.8.9, it returns `"Query.hello"`.

It seems like the issue is that `path` is obtaining its value from the field's `path` instead of `context.path`. I'm not sure if this is the most elegant way of fixing this so please let me know if you think there is a better way.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -8,6 +8,8 @@ module GraphQL
       include GraphQL::Schema::Member::HasArguments
       include GraphQL::Schema::Member::HasPath
 
+      CONTEXT_EXTRAS = [:path]
+
       # @return [String] the GraphQL name for this field, camelized unless `camelize: false` is provided
       attr_reader :name
       alias :graphql_name :name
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/field.rb
```diff
@@ -407,8 +407,9 @@ def apply_scope(value, ctx)
       end
 
       # @param ctx [GraphQL::Query::Context::FieldResolutionContext]
+      CONTEXT_EXTRAS = [:path]
```
rmosolgo: sorry to split hairs, but I think inserting this constant _between_ the method docs and the method signature will break our doc generator. Would you mind moving it up a line?

cjoudrey: Of course! Consider it done. 😄

rmosolgo:
Thanks for taking care of this, sorry for the regression 😖

PR #1861, opened 2018-09-21
read context values using dig

elfassy:
# Why
It's sometimes useful to use `dig` to fetch deeply nested context values 

https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig

# What 

delegate `dig` to `@provided_values`

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -154,7 +154,7 @@ module GraphQL
       # @api private
       attr_writer :value
 
-      def_delegators :@provided_values, :[], :[]=, :to_h, :key?, :fetch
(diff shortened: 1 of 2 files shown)


swalkinshaw:
`dig` is also missing from `InputObject` for arguments. Would be nice to have there as well.

rmosolgo:
Looks good to me! Feel free to add the same to input object in another PR. Thanks for the improvement!

modosc:
fwiw i have a branch with`#dig` added to `Schema::InputObject` and `Query::Arguments` - i'm adding tests now and i'll open a pr up today

Issue #1761, opened 2018-08-09
Improve documentation of GraphQL arguments to Ruby keyword arguments


rmosolgo:
Because Ruby kwargs are relatively new, people aren't too familiar with it. We should have some documentation to clarify: 

- GraphQL args are camelized before calling methods 
- `**inputs` and `**arguments` can always be specified as specific, individual keyword arguments
- Ruby argument default values are required when GraphQL arguments are optional 
- If a GraphQL default value is present and a query is sent without a value for that argument, the default value will be passed as the Ruby keyword argument.

davidrf:
Hi @rmosolgo I'd love to contribute to the project and put up a PR for this. Would it just be updating the [Fields > Arguments](http://graphql-ruby.org/fields/arguments.html) page? Or are there other pages which also mention kwargs?

rmosolgo:
Hey, cool!

Hmm, interesting, those "Fields" guides actually use the old API, they should be replaced altogether. 

I think a good place to start would be to add a _new_ `arguments.md` guide in the `type_definitions/` folder. It could include: 

- How to add arguments to a field 
- How arguments are injected to field methods 
- Working with keyword arguments (the bullets mentioned above)

That would give us a place to add more argument documentation for the new API if we ever need it, and we could remove that old guide. 

Are you interested in adding that new page? I'd be happy to help out along the way if you want to give it a shot.

davidrf:
Yes definitely interested in adding that new page. I'll go ahead and give it a try and we can go from there

rmosolgo:
Fixed by #1864 🎉

PR #1864, opened 2018-09-21
Create Type Definitions > Arguments guide

davidrf:
Removes `Fields > Arguments` guide and creates a new Type `Definitions >
Arguments guide` that has documentation for arguments using the new 1.8 API. Addresses #1761


guides/fields/arguments.md
```diff
@@ -1,88 +0,0 @@
...
-Only certain types are valid for arguments:
-
-- {{ "GraphQL::ScalarType" | api_doc }}, including built-in scalars (string, int, float, boolean, ID)
-- {{ "GraphQL::EnumType" | api_doc }}
-- {{ "GraphQL::InputObjectType" | api_doc }}, which allows key-value pairs as input
-- {{ "GraphQL::ListType" | api_doc }}s of a valid input type
-- {{ "GraphQL::NonNullType" | api_doc }}s of a valid input type
-
-
-The `args` parameter of a `resolve` function will always be a {{ "GraphQL::Query::Arguments" | api_doc }}. You can access specific arguments with `["arg_name"]` or `[:arg_name]`. You recursively turn it into a Ruby Hash with `to_h`. Inside `args`, scalars will be parsed into Ruby values and enums will be converted to their `value:` (if one was provided).
```
davidrf: I removed this section from the new guide (everything from line 78 down) because it seemed to specific to the old API. However, let me know if there is any information here that you'd like in the updated guide

rmosolgo:
Thanks for taking a look at this! I'll make a careful review soon 😊

guides/type_definitions/arguments.md
```diff
@@ -0,0 +1,130 @@
...
+field :search_posts, [PostType], null: false do
+  argument :category, String, required: false, default_value: "Programming"
+end
+
+def search_posts(category:)
+  Post.where(category: category).limit(10)
+end
+```
+
+Use `as: :alternateName` to use a different key from within your resolvers while
```
rmosolgo: I don't recommend using camel-case at all in Ruby code for GraphQL-Ruby, can we change this example to `as: :alternate_name`?

guides/type_definitions/arguments.md
```diff
@@ -0,0 +1,130 @@
...
+
+The corresponding GraphQL query will look like:
+
+```graphql
+{
+  posts(startYear: 2018) {
+    id
+  }
+}
+```
```
rmosolgo: Can you add a note here, something like:

> To disable auto-camelization, pass `camelize: false` to the `argument(...)` configuration.

davidrf:
@rmosolgo those suggestions make sense, so I went ahead and implemented them. Thank you for reviewing.

Also, I noticed that the search is still showing the deleted guide and is missing the new guide. I read on the development guide that I will need the `ALGOLIA_API_KEY` from you to update the search index. Do you want me to update the search index as part of this PR or will you do it later when you re-deploy the site?

rmosolgo:
I'll update it when I deploy, thanks for asking! The problem is, there's only _one_ search index, the production index, so it doesn't update to match our changes in development.

davidrf:
Ah, that makes sense. Let me know if there is anything else that needs to be done before you approve this PR.

rmosolgo:
Looks great, thanks again!

rmosolgo:
:ship: http://graphql-ruby.org/fields/arguments 👏 !

PR #1865, opened 2018-09-22
Document import of PusherLink for Apollo

ksylvest:
The current documentation did not indicate where PusherLink comes from (i.e. not clear if it is part of `apollo` or part of `graphql-ruby-client`. This documents that it is part of `graphql-ruby-client` (and the path for import).


rmosolgo:
Thanks for the fix!

PR #1868, opened 2018-09-24
add #dig to Schema::InputObject and Query::Arguments

modosc:
see also #1861


lib/graphql/dig.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+module GraphQL
+  module Dig
+    # implemented using the old activesupport #dig instead of the ruby built-in
+    # so we can use some of the magic in Schema::InputObject and Query::Arguments
+    # to handle stringified/symbolized keys.
+    #
+    # @param args [Array<[String, Symbol>] Retrieves the value object corresponding to the each key objects repeatedly
+    # @return [Object]
+    def dig(*args)
```
rmosolgo: It looks like the first item from this list is used for `self[]`, then the rest are passed on to the resulting object, is that right? Another way to specify that would be: 

```ruby 
def dig(own_key, *rest_keys) 
```

Then, you don't have to do any confusing array modifications. (For example, `self[args.shift]` is a pretty complicated line, performing a read, an array modification, and a key lookup!)

lib/graphql/dig.rb
```diff
@@ -0,0 +1,19 @@
...
+      val = self[args.shift]
+      if val.nil? || (args.respond_to?(:empty?) ? !!args.empty? : !args)
```
rmosolgo: Won't `args` always be an array here? Why should we check `args.respond_to?(:empty?)`?

spec/graphql/query/arguments_spec.rb
```diff
@@ -121,6 +121,33 @@
...
+    it "returns nil for missing keys" do
+      assert_nil arguments.dig("z")
+      assert_nil arguments.dig(7)
+    end
+
+    it "handles underscored keys" do
+      # TODO shouldn't either of these work?
+      # assert_equal 3, arguments.dig('input_object', 'd')
+      # assert_equal 3, arguments.dig(:input_object, :d)
+    end
```
rmosolgo: Did you expect these to work, but they dont? Or, what do you mean by this comment?

lib/graphql/dig.rb
modosc: you're right - i just copied the code from:

https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/blank.rb#L18

i'll update it.

lib/graphql/dig.rb
modosc: good idea, i'll update this.

spec/graphql/query/arguments_spec.rb
modosc: hah, no, there's no reason that should work, i was just copy-pasting. will remove.

modosc:
ok, changes made

modosc:
ping - curious if any other changes are needed here?

rmosolgo:
Thanks for the bump, sorry I let it slide, looks great!

modosc:
@rmosolgo any idea when this will get released?

Issue #903, opened 2017-08-21
RFC: Input Object Types v2


theorygeek:
I'm not really sure the best title for this, but here's what I'm after. I'm trying really hard to have this in my schema:
```ruby
DateRangeInputType = GraphQL::InputObjectType.define do
  name "DateRangeInput"

  input_field :min, !DateType
  input_field :max, !DateType
end
```

I have a similar `DateRangeType` that is an ordinary object type, so that there's a nice harmony between the input object types and normal object types. But what would really make this spectacular is if my code received an instance of `Range` instead of a `Query::Arguments`.

I first tried to do this little monkey-hack:
```ruby
DateRangeInputType.define_singleton_method(:coerce_non_null_input) do |value, ctx|  
  args = super(value, ctx)
  args[:min]..args[:max]
end
```

I think that this hack would work when the value is provided as a variable. But when the value is provided as a literal in the query, it turns out that [`Query::LiteralInput` has its own coercion code](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/literal_input.rb#L42) which duplicates some of the [logic in `InputObjectType`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/input_object_type.rb#L85).

Then I start thinking long and hard about whether I want to monkey-patch `LiteralInput` too, so I figured I'd just raise the question here. I can see a couple of ways of getting at this behavior, but they all require code changes.

#### Option 1 - Overriding input object coercion
This is essentially what I was aiming to do. To accommodate this well, we'd probably add some sort of a `prepare` API onto `InputObjectType`, very similar to how you can specify that option on arguments.

The next step would be to refactor both of those places in code I mentioned above, so that they respect the `prepare` option. Going this route with my example above, it would look like:
```ruby
DateRangeInputType = GraphQL::InputObjectType.define do
  name "DateRangeInput"

  input_field :min, !DateType
  input_field :max, !DateType

  # Your application code will receive the result of this proc, instead of a
  # Query::Arguments object.
  prepare -> (args, ctx) { args[:min]..args[:max] }
end
```

#### Option 2 - Adding a prepare option to the individual fields on an `InputObjectType`
The other place where this kind of work could live is inside of the `prepare` option that we already have today on field arguments. To make this really nice, we would want to extend it so that you could get this behavior when an input object is embedded inside another input object.

In this world, the usage would look a little more complicated, because the work would live on the field that is using the input type, instead of on the input type itself:
[...]

PR #1869, opened 2018-09-24
Prepare input objects

theorygeek:
This allows you to build input objects for, eg, range types, and get a `Range` passed into your app.

Example:
```ruby
class DateRangeInput < GraphQL::Schema::InputObject
  argument :min, Int, required: true
  argument :max, Int, required: true

  def prepare
    return unless min <= max
    min..max
  end
end
```

Resolves https://github.com/rmosolgo/graphql-ruby/issues/903#issuecomment-424113851


lib/graphql/query/literal_input.rb
```diff
@@ -108,7 +108,13 @@ def self.from_arguments(ast_arguments, argument_owner, variables)
           end
         end
 
-        argument_owner.arguments_class.new(values_hash, context: context, defaults_used: defaults_used)
+        result = argument_owner.arguments_class.new(values_hash, context: context, defaults_used: defaults_used)
+
+        if result.respond_to?(:prepare)
```
eapache: should the base classes for all of these just `def prepare; self; end`?

theorygeek:
A couple of challenges with this implementation:

1. I tried to also give you a nice validation error if your `prepare` returns nil, similar to the way that `coerce_non_null_input` on scalar types behaves. However, it doesn't work for literal input; I'm not entirely sure what's missing there, or the best way to rectify it. 

(Probably too much duplicated code for handling input objects?)

2. This introduces a performance regression, since as-implemented, it will instantiate each input object _twice_:
- Once during `validate_non_null_input` inside of GraphQL::InputObjectType, to check for `nil`'s
- And then again when the value is actually being passed into a resolver (the exact mechanics here is a bit of a mystery to me, though)

eapache:
I'm a bit surprised this doesn't break something when it tries to recurse into `DateRangeInput` in order to prepare/validate `min` and `max` but can't because it not longer an `InputObject`?

theorygeek:
> I'm a bit surprised this doesn't break something when it tries to recurse
I think that recursion happens inside of the `#initialize` method on GraphQL::Schema::InputObject.

spec/graphql/schema/input_object_spec.rb
```diff
@@ -89,12 +89,81 @@ class Schema < GraphQL::Schema
...
+      query_str = <<-GRAPHQL
+      query ($input: InputObj!){ inputs(input: $input) }
+      GRAPHQL
+
+      res = InputObjectPrepareTest::Schema.execute(query_str, variables: { input: { min: 5, max: 10 } })
+      expected_obj = (5..10).inspect
+      assert_equal expected_obj, res["data"]["inputs"]
+    end
+    
+    it "gives an error if prepare returns nil (literal)" do
```
theorygeek: This test isn't passing, but I'm not entirely sure what's missing.

lib/graphql/input_object_type.rb
```diff
@@ -147,6 +153,11 @@ def validate_non_null_input(input, ctx)
         end
       end
 
+      # Better way to do this, perhaps keeping the coerced value around?
+      if coerce_non_null_input(input, ctx).nil?
```
theorygeek: Introduces a performance regression: input is converted into the `InputObject` class created by the user, and the custom `prepare` method is invoked, but all of that work is thrown away.

If the user's `initialize` and `prepare` methods are expensive, it'll cause a problem. Also, they might not expect those methods to get invoked twice: once for validation, and then once for the actual value.

theorygeek:
I eliminated the use-case for dynamic, runtime validation (eg, by returning `nil` from a `prepare` method). The tests are passing, except for a seemingly unrelated one (something for `GraphQL::InternalRepresentation::Rewrite`).

I also eliminated the `respond_to?` checks, and updated the guides. Would love some feedback from @rmosolgo on this.

rmosolgo:
Sorry to leave this hanging for a year 😞 . Are you still interested in a change like this? 

I was initially turned off by it, because argument `prepare:` hooks have been nothing but a headache for me. But honestly, this approach seems really straightforward, and if you have a use for it, I'm game to merge it. 

I've tried to resolve the conflicts 🤞

theorygeek:
> Are you still interested in a change like this?

Yeah, I still think this would still be valuable.

PR #1876, opened 2018-10-02
Added Kickstarter to the list of gem users ❤️

natachaS:
The Kickstarter Engineering team loves the GraphQL-Ruby gem (and we use it in production on our site extensively!) so we added ourselves to your list of users. Let us know know if this is an issue. ❤️


rmosolgo:
Very cool, thanks for adding that! Glad it's been working out for you :)

PR #1885, opened 2018-10-05
Update 1.9-dev with changes on master

rmosolgo:


rmosolgo:
THis is not gonna do what i want

PR #1889, opened 2018-10-07
Pusher: use require and module.exports instead of import/export

jsierles:
Without this, you need to explicitly transpile the node module contents.

### DIFF
--- a/javascript_client/subscriptions/PusherLink.js
+++ b/javascript_client/subscriptions/PusherLink.js
@@ -34,7 +34,8 @@
 //     // Do something with `data` and/or `errors`
 //   }})
 //
-import {ApolloLink, Observable} from "apollo-link"
+var ApolloLink = require("apollo-link").ApolloLink
+var Observable = require("apollo-link").Observable
 
 class PusherLink extends ApolloLink {
   constructor(options) {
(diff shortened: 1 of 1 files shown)


jsierles:
A common practice for published NPM modules is to ensure they do not use modern JS that isn't supported by all javascript runtimes. For example, nodejs only supports import/export with an experimental feature flag now.

To make modern javascript work, you need to either write your modules using ES5 javascript, or use webpack (or similar) to bundle a transpiled copy of your library at publishing time. While this could be done by the user, it's a bad practice to needlessly transpile ALL the node modules you have in your app.

I solved this by explicitly transpiling your module, but I think this is an advanced technique and not good for devs who expect libraries to 'just work' in most environments.

In the case of this library, I only needed this change to get it to work on NodeJS. But you might want to also consider doing a universal build!

rmosolgo:
Ohhh, I see, so it worked for me when I was developing it because I was using webpacker, but if someone (you!) wants to run it server-side with Node.js, then it won't work, is that right? 

Makes sense, thanks for the clarification!

If anyone was previously using `import PusherLink ...`, will they have to update their code? (This isn't a blocker, but I want to update the changelog appropriately.) While I'm at it, I should probably check other usages to make sure they're using `require` and `module.exports` for this reason.

rmosolgo:
Thanks for the fix!

PR #1891, opened 2018-10-08
Add an owner-level resolve wrapper method 

rmosolgo:
Sometimes, you need to treat a field differently based on who its _owner_ is. An example is batching: mutation root fields are treated specially, they don't read batched data because they might have just updated it. 

We have something similiar for  `Query.rateLimit`, where other fields may be skipped if `rateLimit(dryRun: true)` is passed. 

I'm not sure this is the right approach though. In `1.9-dev` this is addressed by adding field extensions to root types (eg, the implementation of subscriptions)

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -338,7 +338,7 @@ module GraphQL
         ctx.schema.after_lazy(obj) do |after_obj|
           # First, apply auth ...
           query_ctx = ctx.query.context
-          inner_obj = after_obj && after_obj.object
+          inner_obj = after_obj.object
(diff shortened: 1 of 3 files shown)


rmosolgo:
The other options are: 

- Subclass your base field and apply it to the object type with `field_class`. 
- Override `def self.field` to add an extra keyword option, then, in the base field class, use the presence of that option to add extra behavior 
- 1.9-dev: Override `def self.field` to add an extra instrumenter, which will be hooked up to the created Schema::Field instance (that's how subscriptions were implemented)

I think those options will work ok, and I won't have to add _another_ step of back-and-forth between object and field

Issue #1107, opened 2017-11-17
Not handling an ExecutionError when returned from prepare on an InputObjectType


joshforbes:
When I return a GraphQL::ExecutionError from prepare on an argument for an InputObjectType I am getting some strange behavior.

![image](https://user-images.githubusercontent.com/3682518/32968851-c2b02368-cbb0-11e7-8798-fe3b96ceaec9.png)

Instead of returning to the client that the field is invalid it is passing the exception along to the resolver as the value of the argument.

If I raise the ExecutionError instead of returning, it is not caught.

rmosolgo:
😖  Like you mentioned on slack, I wonder if it's because it's on an input object instead of a field!

rmosolgo:
I made a script to check this behavior on a recent graphql-ruby version: 

```ruby 
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", "1.10.4"
end


class SomeInput < GraphQL::Schema::InputObject
  argument :int, Int, required: true, prepare: -> (val, ctx) {
    if val > 5
      return GraphQL::ExecutionError.new("Too big!")
    else
      val
    end
  }
end
class Query < GraphQL::Schema::Object
  field :int, Int, null: true do
    argument :input_object, SomeInput, required: true
  end

  def int(input_object:)
    input_object[:int]
  end
end

class Schema < GraphQL::Schema
  query Query
end

pp Schema.execute("{ int(inputObject: { int: 3 }) }").to_h

pp Schema.execute("{ int(inputObject: { int: 13 }) }").to_h
```

It outputs the expected: 

```
$ ruby test.rb
{"data"=>{"int"=>3}}
{"data"=>{"int"=>nil},
 "errors"=>
  [{"message"=>"Too big!",
    "locations"=>[{"line"=>1, "column"=>3}],
    "path"=>["int"]}]}
```

So, I think this was fixed somewhere along the way 😅 . Please open a new issue if you're still encountering this on a recent graphql-ruby version!

PR #1892, opened 2018-10-08
Clarify argument prepare documentation

cmatheson:
The current documentation suggests that returning a
GraphQL::ExecutionError from an argument's prepare function is valid,
but you must actually raise the GraphQL::ExecutionError to achieve the
desired behavior.


rvirani1:
#1107 Seems related to this issue.

rmosolgo:
Thanks for addressing this!

PR #1894, opened 2018-10-09
Add a basic lookahead object

rmosolgo:
I'm trying to migrate from `irep_node:`-based lookahead, so I want to implement something that will work with the current execution code _and_ with the future execution code, i think this will work.


rmosolgo:
Oops, this won't work in cases like 

```graphql 
{
  f1 { a } 
  f1 { b } 
}
```

because it only keeps the _first_ AST node matching a selection. So `selection("f1").selection("b").selected?` will be false, when it should be true. 

Just like execution, it should keep a _list_ of AST nodes, not just 1.

rmosolgo:
This still doesn't work great with Unions, I mean, fields that return unions won't play nice with lookahead yet.

ianks:
@rmosolgo is it possible to get all selected fields with this API?

rmosolgo:
Not yet, I put it off because I didn't need it yet. I think a nice API would be to add `#selections` which returns a list of Lookaheads for each selection, and then make each `Lookahead` respond to `#name`, so that you can tell what it represents. 

```ruby 
next_lookaheads = lookahead.selections  
# => [#<GraphQL::Execution::Lookahead ...>, ...]
next_lookaheads.map(&:name) 
# => ["next", "field", "names", "go", "here"]
```

What do you think? Want to take a crack at it?

PR #1902, opened 2018-10-15
Fix type printing in Schema printer

swalkinshaw:
Fixes a very old regression in `print_type`. Right now it only handles
object types and breaks on anything else.

This updates it to use `build_type_definition_node` which detects the
type kind and works as expected with every kind.

### DIFF
--- a/lib/graphql/schema/printer.rb
+++ b/lib/graphql/schema/printer.rb
@@ -89,7 +89,7 @@ module GraphQL
       end
 
       def print_type(type)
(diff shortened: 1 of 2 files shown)


adamhollett:
@rmosolgo, any way I could ask for a version release? It would be really amazing for my current project to be able to use this fix.

rmosolgo:
Just shipped 1.8.11!

adamhollett:
Thank you so much!

PR #1907, opened 2018-10-17
Add GraphQL::Execution::Lookahead#selections

ianks:
This PR adds the ability to get all of the selections from a Lookahead. I decided to keep the implementation simple by just re-using the `selection` method instead of refactoring, let me know if that's alright :smile:

Continuation of https://github.com/rmosolgo/graphql-ruby/pull/1894

### DIFF
--- a/lib/graphql/execution/lookahead.rb
+++ b/lib/graphql/execution/lookahead.rb
@@ -94,6 +94,29 @@ module GraphQL
         end
       end
 
+      # Like {#selection}, but for all nodes.
+      # It returns a list of Lookaheads for all Selections
+      #
+      # If `arguments:` is provided, each provided key/value will be matched
+      # against the arguments in each selection. This method will filter the selections
+      # if any of the given `arguments:` do not match the given selection.
+      # @param arguments [Hash] Arguments which must match in the selection
+      # @return [Array<GraphQL::Execution::Lookahead>]
+      def selections(arguments: nil)
+        @ast_nodes
+          .flat_map(&:selections)
+          .map(&:name)
+          .map { |name| selection(name, arguments: arguments) }
+          .select(&:selected?)
+      end
+
+      # The name of the field.
+      # It returns the same of the Lookahead's field.
+      # @return [Symbol]
+      def name
+        @field&.method_sym
+      end
+
       # This is returned for {Lookahead#selection} when a non-existent field is passed
       class NullLookahead < Lookahead
         # No inputs required here.
--- a/lib/graphql/execution/lookahead.rb
+++ b/lib/graphql/execution/lookahead.rb
@@ -110,11 +110,15 @@ module GraphQL
       # @param arguments [Hash] Arguments which must match in the selection
       # @return [Array<GraphQL::Execution::Lookahead>]
       def selections(arguments: nil)
-        @ast_nodes
-          .flat_map(&:selections)
-          .map(&:name)
-          .map { |name| selection(name, arguments: arguments) }
-          .select(&:selected?)
+        subselections_by_name = {}
+        @ast_nodes.each do |node|
+          node.selections.each do |subselection|
+            subselections_by_name[subselection.name] ||= selection(subselection.name, arguments: arguments)
+          end
+        end
+
+        # Items may be filtered out if `arguments` doesn't match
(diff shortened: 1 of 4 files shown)


lib/graphql/execution/lookahead.rb
```diff
@@ -94,6 +94,43 @@ def selection(field_name, arguments: nil)
...
+      #
+      # @example getting the name of a selection
+      #   def articles(lookahead:)
+      #     article.selection(:full_content).name # => :full_content
+      #     # ...
+      #   end
+      #
+      # @return [Symbol]
+      def name
+        @field&.method_sym
```
rmosolgo: A couple thoughts here: 

- As you noticed on the CI failures, we support some rubies that don't have `&.`,  so this will have to be checked "manually", eg `@field && @field.something` 
- I don't think `.method_sym` is really what we want, because if the field is created with `field :name, ..., method: :some_other_method`, this will return `:some_other_method` instead of `:name`. As you probably noticed, there's _no way_ to get the originally-configured name 🤦‍♂️ . Maybe the thing to do is improve `schema/field.rb` here: 

  https://github.com/rmosolgo/graphql-ruby/blob/5527179a7b2f100ce9ee093764c924e87d155f1b/lib/graphql/schema/field.rb#L145

  To instead be like 
   ``` 
    @original_name = name 
    @name = # ... 
    ```

    And add an `original_name` reader method. Do you think that would do the trick?

lib/graphql/execution/lookahead.rb
```diff
@@ -94,6 +94,43 @@ def selection(field_name, arguments: nil)
...
+      #   end
+      #
+      # @param arguments [Hash] Arguments which must match in the selection
+      # @return [Array<GraphQL::Execution::Lookahead>]
+      def selections(arguments: nil)
+        @ast_nodes
+          .flat_map(&:selections)
+          .map(&:name)
+          .map { |name| selection(name, arguments: arguments) }
+          .select(&:selected?)
```
rmosolgo: I _think_ this will have a bug if two of the `@ast_nodes` have two of the same selections, for example: 

```
a { b }
a { b c }
```

In that case, `.selections` will return three items, one for the first `b`, another for the second `b`, and one for `c`. It should only return 2 in that case, right? 

~~Also, I noticed you added `.select(&:selected?`. Did you find that unselected selections were return there? I can't imagine~~, Oh, nevermind, I guess it's for when `arguments` doesn't match, is that right? 

Anyhow, how about adding a test case for that multiple-selection issue to make sure it's handled properly?

lib/graphql/execution/lookahead.rb
ianks: Good catch, did not realize that is what method_sym was. I will make that enhancement :smile:

ianks:
@rmosolgo  addressed your comments, should be good now once CI passes. Thanks for the help!

rmosolgo:
Cool, thanks for taking a look! What do you think of my suggestion in 8bb135c ? Effectively, it _merges_ selections, so that 

```graphql 
a { b } 
a { c } 
```

Is treated the same as 

```graphql
a { b c } 
```

The GraphQL runtime does this, too -- `a` is only executed once. This causes `selections.map(&:name)` to return only `[:a]`, not `[:a, :a]`, does that seem ok to you?

ianks:
Ahhh ok. What is the reasoning for only returning it once? I would think that they are different selections because they have different sub-selections, This could potentially cause issues when attempting to dig deeped into the sub-lookaheads.

ie.

```
query_ one = a { b } 
query_two  = a { c } 

query_one.selections.find('a').selections # => ['b']
query_one.selections.find('a').selections # => ['c']
```

ianks:
This makes sense now, and is a much better way of dealing with the problem. Essentially, we can consider 'name' to be unique?

rmosolgo:
👌 Bingo! `.selections` will never return two lookaheads with the same name. 

Ok, caveat: there are still unhandled edge cases here with aliases. For example, I'm not really sure what this will return: 

```graphql
query {
  users { 
    totalCount 
    name 
    alsoName: name 
  }
}
```

Because `Lookahead` pays no mind to `alias`. Honestly, I'm not to worried about it, because it works ok for my own case. It suits me to cross that bridge when we get to it!

Issue #1353, opened 2018-03-15
Unable to sort while using relay connection pagination.


lynndylanhurley:
The following error occurs when trying to paginate with an `order` applied:

~~~
     ActiveRecord::StatementInvalid:
       PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
       LINE 1: ...nator_id = 1 OR assignments.user_id = 1) ORDER BY title OFFS...
                                                                    ^
       : SELECT COUNT(DISTINCT count_column) FROM (SELECT DISTINCT "matters"."id" AS count_column FROM "matters" LEFT OUTER JOIN "assignments"
ON "assignments"."matter_id" = "matters"."id" AND "assignments"."deleted_at" IS NULL WHERE (matters.originator_id = 1 OR assignments.user_id =
1) ORDER BY title OFFSET $1) subquery_for_count
~~~

The issue can be resolved by unscoping the `order` before performing the `count` on [this line](https://github.com/rmosolgo/graphql-ruby/blame/f6aab69afe53505f0bb4e0bf8fbae9e6152e4fd2/lib/graphql/relay/relation_connection.rb#L127).

This works:
~~~ruby
      # If a relation contains a `.group` clause, a `.count` will return a Hash.
      def relation_count(relation)
        count_or_hash = if(defined?(ActiveRecord::Relation) && relation.is_a?(ActiveRecord::Relation))
          # unscope the :order before performing the `count`
          relation.unscope(:order).count(:all)
        else # eg, Sequel::Dataset, don't mess up others
          relation.count
        end
        count_or_hash.is_a?(Integer) ? count_or_hash : count_or_hash.length
      end
~~~

Would you accept a PR with this change?

sideshowdave7:
I second this for performance reasons - sending an ordered relation, especially when ordering by fields from joined tables can be really expensive in MySQL.

sideshowdave7:
I did some more testing and ran into the same error as @lynndylanhurley actually.  I've put in a PR that addresses the issue (supported by Rails 4/5)

rmosolgo:
Fixed by #1911

PR #1911, opened 2018-10-19
Add unscope(:order) to relation count on relay

sideshowdave7:
Passing in ordered relations to Graphql Relay can give undesirable results as seen in issue #1353 

Doing this, or adding an option to unorder the relation before counting would be extremely useful.

My use case for this is ordering on joined tables makes counts in MySQL really slow.


rmosolgo:
Thanks for the fix, it looks good! Should we do anything to test it?

sideshowdave7:
I'd be happy to do that, any pointers on where to write a test for this?

mazikwyry:
@rmosolgo ping. I have the same issue after migrating an app from MySQL to Postgres and would be great to have that fix released.

The easiest way to test that would be to check if `.order("table_name.attrubute DESC")` works when asking for `total_count` in the query. (note it works for the first page and the last page, you have to have at least 3 pages).

rmosolgo:
Hey, sorry for leaving this hanging and not recommending a place to test 😞  I just pushed a test myself and I'll get this rolled into the next release!

sideshowdave7:
@rmosolgo Thanks for doing that!

PR #1912, opened 2018-10-19
Add m gem for easier test running

ianks:
Having had some RSI issues in the past, I like to try to keep my typing as minimial as possible while developing, and the current way to run specs is a bit verbose for my liking. Would it be OK to add the m gem to make this process easier?


rmosolgo:
Sure, thanks for the improvement and for including it in the docs! I'm interested to give it a try myself.

PR #1918, opened 2018-10-23
Use new analysis engine when using Interpreter

xuorig:
Automatically opt in into new analysis engine when using the interpreter runtime!

### DIFF
--- a/lib/graphql/analysis/ast.rb
+++ b/lib/graphql/analysis/ast.rb
@@ -13,11 +13,6 @@ module GraphQL
     module AST
       module_function
 
-      def use(schema_defn)
-        schema = schema_defn.target
-        schema.analysis_engine = GraphQL::Analysis::AST
-      end
-
       # Analyze a multiplex, and all queries within.
       # Multiplex analyzers are ran for all queries, keeping state.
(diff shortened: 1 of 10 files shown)


spec/graphql/static_validation/rules/variables_are_used_and_defined_spec.rb
```diff
@@ -71,7 +71,6 @@
       GRAPHQL
     }
 
-    focus
```
xuorig: 😨

lib/graphql/analysis/ast.rb
```diff
@@ -13,11 +13,6 @@ module Analysis
     module AST
       module_function
 
-      def use(schema_defn)
```
xuorig: Moved to `use Interpreter` instead.

xuorig:
Weird tracing test failing with `TESTING_INTERPRETER`, somehow can't reproduce locally 🤔

rmosolgo:
It turns out a bit convenient that these are separate, for migrating github, I did `use GraphQL::Exection::Interpreter` _first_, then migrated the analyzers afterward. So, let's keep them separate to support a more gradual adoption.

PR #1928, opened 2018-10-29
Guard against recursive calls for cached methods

theorygeek:
This PR raises an exception when it detects a recursive invocation of a cached method, like `graphql_definition` or `connection_type`. This helps you avoid the dreaded, "Duplicate type definition found for SomeType" error message, by attempting to detect the scenario early.

The solution usually involves wrapping the type definition in a proc inside of the legacy-style type (eg, `-> { SomeType.graphql_definition }`).

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -7,6 +7,53 @@ require "forwardable"
 require_relative "./graphql/railtie" if defined? Rails::Railtie
 
 module GraphQL
+  class RecursionGuard
+    attr_accessor :trace, :guard_name, :object
+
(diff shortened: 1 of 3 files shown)


rmosolgo:
This makes sense to me, I've also worked around it by using a string instead of a constant reference. It will be great to give people an actual, proper error message! Do you want to add a little test for it?

theorygeek:
Yeah, I can definitely do that. Also looks like older versions of Ruby don't like this style of heredocs, I'll try to get that fixed also.

Issue #1910, opened 2018-10-18
Question: Executing and Returning Initial Query with Subscription


ksylvest:
For a GraphQL subscription is it possible to execute the initial query (to support pseudo live-queries)?

```ruby
module Types
  class SampleType < GraphQL::Schema::Object
    field :greeting, String, null: false

    def greeting
     Time.current < Time.current.noon ? 'Morning!' : 'Evening'
    end
  end
end

class SampleSchema < GraphQL::Schema
  use GraphQL::Subscriptions::ActionCableSubscriptions
  subscription Types::SampleType
end

SampleSchema.execute('subscription greeting { greeting }')['data'] # nil -- can this be configured to be { greeting: '...' } ?
```

Per [FB RFC on subscriptions](https://github.com/facebook/graphql/blob/master/rfcs/Subscriptions.md) - I'm a bit unclear if this type of thing is meant to be supported:

> The server may send an initial publish from executing the subscription operation.

Realize this can likely be done via a `Query` + `Subscription`.

Also - is the logic on controlling execution wrapped up here?: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions/instrumentation.rb#L59

ksylvest:
Thanks for the response Robert! Appreciate the work you put into graphql-ruby (super useful library). Closing.

rmosolgo:
Btw I was looking back on the spec and didn't see any mention one way or another about initial responses, do you? 

https://facebook.github.io/graphql/June2018/#sec-Subscription

I see your link above is to the RFC, but I wonder if/how that made it to the spec. I'm going to add support for initial responses one way or another but I'll include a bit from the spec if I can find it!

ksylvest:
Hmm, I can not find anything within the spec linked. It seems like it'd be challenging to avoid race conditions without (via the query / subscribe combo.

rmosolgo:
yeah, i [agree](https://github.com/facebook/graphql/issues/283#issuecomment-284553218) 😉

rmosolgo:
Trying to add proper support for this in #1930

Issue #1799, opened 2018-08-22
what to define in a "Subscription Type" method


github0013:
# env

- Darwin mbp.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 21 20:07:39 PDT 2018; root:xnu-3789.73.14~1/RELEASE_X86_64 x86_64
- rails (5.2.1)
- graphql (1.8.7)

NPM
- "graphql-ruby-client": "^1.4.0"
- "actioncable": "^5.2.1"
- "@rails/webpacker": "3.5"

I am scratching my head trying to understand how to implement a subscription.

http://graphql-ruby.org/subscriptions/subscription_type.html
![graphql_-_subscription_type](https://user-images.githubusercontent.com/809378/44440038-52da0c80-a601-11e8-88c9-c1dee2b8ad1c.png)
It says define a field like above, but the field actually needs a corresponding method like this.  (otherwise it shows errors)

```rb
def post_was_published
  # ...
end
```
and just the below the code, it says `This will be called on the initial request`
![graphql_-_subscription_type-2](https://user-images.githubusercontent.com/809378/44440219-22df3900-a602-11e8-9425-2381c0076870.png)

But as far as I checked, this method will be called 

- when client initially loads its page
- when Schema.subscriptions.trigger is called

Plus, even if I have a return value written in the method, it will not be sent to the client, but the triggered object value will be sent.

```rb
def post_was_published
  Post.first # this will never be sent to the client on either initial or trigger
end 
```
```rb
Schema.subscriptions.trigger("postWasPublished", {}, Post.last) # this Post.last will be sent
```

Here are what I am trying to figure out.

1. The SubscriptionType method is being called on initial and trigger call, shouldn't it be just on initial?
1. The return value in the SubscriptionType method won't be sent to the client, I shouldn't return any values?
1. Should I just have a job call (like sidekiq perform_async which calls `Schema.subscriptions.trigger`) in the method?, but the method is called on trigger as well, that will make a loop

ex:
```rb
def post_was_published
  SidekiqJob.perform_async # this will do the trigger call
end
```

phyllisstein:
I think I'm struggling with something very similar. I've been wracking my brain and combing the source trying to figure out what I'm missing, but I can't figure out why the `object` argument to `Schema.subscriptions.trigger` is always the result of the reloaded query.

My current setup involves a simple subscription for tracking the progress of an ActiveJob-powered upload task:

```ruby
[...]

Issue #1528, opened 2018-05-22
How would you authorize a subscription?


Conturbo:
**How would you authorize in Rails?** _(Assuming you set `context: current_user` properly in `app/channels/graphql_channel.rb`)_ I'm trying to update users when a friend request has been received. 

<-- Edit: Can [scope](http://graphql-ruby.org/api-doc/1.8.0/GraphQL/Subscriptions.html#trigger-instance_method) help me?  -- >

<-- Edit: It's similar to  #1057  -->

So far it works. Only drawback: authorization. Currently, the client puts in a user_id to specify which user's content to subscribe to, but any user can pass in any other user's id. Instead, I want each subscription to subscribe using `ctx[:current_user].id`. But idk how to access the **context**. How would I do this? I tried adding a `call(obj, args, ctx)` or `resolve` method to my subscription, but it didn't work. Here's what I've got for reference:

**A subscription for newly received friend requests:**
```ruby
# app/graphql/subscriptions/friend_request_received.rb
class Subscriptions::FriendRequestReceived < GraphQL::Function
  type !Types::FriendRequestType
  argument :id, !types.ID
  description 'A friend request was received by the user.'
end

```

**A job that triggers all subscribers:**
```ruby
# app/jobs/create_friend_request_job.rb
class CreateFriendRequestJob < ApplicationJob
  queue_as :friend_requests

  def perform(friend_request)
    MySchema.subscriptions.trigger('friendRequestReceived', { id: friend_request.receiver.id }, friend_request)
  end

end
```

**A method creates a job after_create:**
```ruby
# app/models/friend_request.rb
class FriendRequest < ApplicationRecord
  after_create :update_subscribers

...

  def update_subscribers
    CreateFriendRequestJob.new(self).enqueue
  end

...

end
```

Let me know if anyone can help in any way! Also, does this make sense? I'm doing all of my work for the subscription OUTSIDE of the subscription file itself, which kind of feels weird. Is this the proper way to do things? Thanks!

flyfy1:
Hi @Conturbo ,  my recent PR just got merged: https://github.com/rmosolgo/graphql-ruby/pull/1502

It's to support authorization. So I think it would be supported when the next release comes out, and @rmosolgo also added in the docs for it: https://github.com/rmosolgo/graphql-ruby/commit/a329ff3cb84cc83da15b6283cf80dcdd68f49286#diff-d206d5f8891deaca408fcccd141e6193R63

[...]

PR #1930, opened 2018-10-29
Class-based subscription fields

rmosolgo:
The current subscription system leaves a lot to be desired, see for example #1528 #1799 #1910 

I'm proposing a `GraphQL::Schema::Subscription` like `GraphQL::Schema::Mutation` with authorization, initialization, and update methods.

You can see the rendered doc here: https://github.com/rmosolgo/graphql-ruby/blob/good-subscriptions/guides/subscriptions/subscription_classes.md

__TODO__: 

- [x] What if you don't want to update a certain client of an event? (For example, that client _triggered_ the event)


theorygeek:
Nice! I like it. How do you hook up the transport? I'm assuming there needs to be some method that GraphQL can invoke when an update needs to be sent to a client?

> Then, `def update` will be called for each client's subscription
It is going to send the update to _every_ client by default, and your `update` should return `skip` if that particular client doesn't need the update?

rmosolgo:
Thanks for taking a look! 

> How do you hook up the transport?

That consideration is separated a bit. The changes here revolve around how subscriptions are _evaluated_ by GraphQL, but not how the result is transmitted to clients. 

In terms of transport, the subscription implementation should define [`#deliver` for sending payloads](http://graphql-ruby.org/api-doc/1.8.11/GraphQL/Subscriptions.html#deliver-instance_method) to clients. See [ActionCable's `#deliver`](http://graphql-ruby.org/api-doc/1.8.11/GraphQL/Subscriptions/ActionCableSubscriptions.html#deliver-instance_method) for example.

> _every_ client by default? 

Yes, that's right. I'm glad you mentioned it: it should really have a way to halt, and send _nothing_ to a client if desired, but there's not a way to do that yet. I added a TODO above.

modosc:
> What if you don't want to update a certain client of an event? 

we've hit this a few different times - mostly because of my understanding on how scoping works it's hard to have multidimensional filters (ie, company A + role B). we sometimes end up sending out notifications on a model which is not visible for some clients. if we allow the subscription field to be null, then we send out empty messages to a bunch of clients. my hacky solution was to make the field non-nullable and hide the `Cannot return null for non-nullable field Subscription.foo` error but this isn't great. ideally there'd be some option for "not null for the client but if the server returns null just skip delivery".

rmosolgo:
It's basically implemented as written, feel free to take another look if you're interested! Otherwise I'll merge it tomorrow or so.

phyllisstein:
I'm so excited about this @rmosolgo, thank you for all this amazing work. ✨

PR #1937, opened 2018-10-31
Fix lexer for multiple block strings

pushrax:
Parsing any document with more than a single block string would result in all text between them being combined into the string. `^'"""'` is not correct; `^` is a character-level negation.

I first tried just switching `^` to `!` (machine-level negation) but this doesn't work. `!'"""'` matches `"`, which can then be repeated 3 times to match `"""`. Instead, to match anything that's not `"""`, it works to match any non-quote character, or one or two quotes surrounded by other characters.

To be completely spec-compliant, as far as I can tell it's also necessary to allow `""""" foo """""`, which is why `QUOTE{0,2}` is included in the `BLOCK_STRING` definition.

r @rmosolgo 
cc @Azhng @jpittis


spec/graphql/language/lexer_spec.rb
```diff
@@ -26,12 +26,18 @@
       assert_equal tokens[0], tokens[1].prev_token
     end
 
+    it "allows escaped quotes in strings" do
+      tokens = subject.tokenize('"a\\"b""c"')
+      assert_equal 'a"b', tokens[0].value
+      assert_equal 'c', tokens[1].value
+    end
```
pushrax: This test isn't related to the PR, but coverage was missing and I wrote it as part of figuring out what to do.

lib/graphql/language/lexer.rl
```diff
@@ -37,9 +37,9 @@
   QUOTE =         '"';
   BLOCK_QUOTE =   '"""';
   ESCAPED_BLOCK_QUOTE = '\\"""';
-  BLOCK_STRING_CHAR = (ESCAPED_BLOCK_QUOTE | ^'"""');
+  BLOCK_STRING_CHAR = (ESCAPED_BLOCK_QUOTE | ^QUOTE | ^QUOTE QUOTE{1,2} ^QUOTE);
```
pushrax: Arguably this whole definition should just go in `BLOCK_STRING` now, it's not really a character class or constant sequence.

lib/graphql/language/lexer.rl
```diff
@@ -49,7 +49,7 @@
   AMP =           '&';
 
   QUOTED_STRING = QUOTE STRING_CHAR* QUOTE;
-  BLOCK_STRING = BLOCK_QUOTE BLOCK_STRING_CHAR* BLOCK_QUOTE;
+  BLOCK_STRING = BLOCK_QUOTE QUOTE{0,2} BLOCK_STRING_CHAR* QUOTE{0,2} BLOCK_QUOTE;
```
pushrax: Hmm this allows `””””””””””` due to *, that’s out of spec. Will need to do something like `(QUOTE{1,2} ^QUOTE){,1}` to fix.

rmosolgo:
Wow, tricky! Thanks for taking care of it. Feel free to inline `BLOCK_STRING_CHAR` too, if you'd like.

PR #1941, opened 2018-11-02
Add generateClient function for creating JS code without HTTP operation

rmosolgo:
This adds a dedicated `generateClient` function which may be used to create a JS/JSON file based on the project's `.graphql` code. You can use it to integrate with webpack or make sure your generated files are up to date.


daemonsy:
Hey @rmosolgo will there be an export point for the client generation? 

I'm using sync in API mode directly from https://github.com/rmosolgo/graphql-ruby/blob/343be3b0aa3677146ea074324a9253139fadbee3/javascript_client/index.js#L4

daemonsy:
Thanks for the update, looks good to me!

javascript_client/sync/generateClient.js
```diff
@@ -9,18 +9,82 @@ var generators = {
...
+ * @param {String} options.mode - If `"file"`, treat each file separately. If `"project"`, concatenate all files and extract each operation. If `"relay"`, treat it as relay-compiler output
+ * @param {Boolean} options.addTypename - Indicates if the "__typename" field are automatically added to your queries
+ * @param {String} options.clientType - The type of the generated code (i.e., json, js)
+ * @param {String} options.client - the Client ID that these operations belong to
+ * @param {Function} options.hash - A custom hash function for query strings with the signature `options.hash(string) => digest` (Default is `md5(string) => digest`)
+ * @return {Array} Array of operations with name and alias
+*/
+function gatherOperations(options) {
+  var graphqlGlob = options.path || "./"
+  var hashFunc = options.hash || md5
```
rmosolgo: I think this needs to import md5, why does this work 🤔

rmosolgo:
🚢  in 1.6.0! Let me know if you get to give it a try.

PR #1943, opened 2018-11-06
Remove unused requires

daemonsy:
## What's up? 

Spotted some unused `require`s while reading the code in the JS client. Thought removing it will make it easier for future readers


rmosolgo:
Oops, thanks for the cleanup!

PR #1944, opened 2018-11-06
Remove confusing non-relevant example

chevinbrown:
The usage of non-library code in examples is confusing to beginners and should be avoided.


rmosolgo:
🤷‍♂️ Ok, I wanted to give an example of what you might do with the data that was provided.

chevinbrown:
Sorry for the terse description!

I've found that it's a better rule-of-thumb for docs to contain only relevant code. 

I'm not necessarily a beginner, and I was wondering where that other code came from for a minute! 

1% better is all we can shoot for! 😸

Issue #1948, opened 2018-11-08
Apollo Client doesn't recognize batched multiplex result


Mando75:
We (SimpleNexus) have been exploring using batched queries with Apollo Client/batch-http and GraphQL Ruby. We ran into an issue with GraphQL Ruby rendering a batched query JSON result that Apollo Client was unable to understand. When using the `render json: result` as shown in the [multiplex documentation](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/queries/multiplex.md#apollo-query-batching), the rails controller would return back the following JSON object: 
```
{
   "graphql": [{"data": {array}},{"data": {of}},{"data": {query}},{"data": {results}}]
}
```
From what we can tell, Apollo Client only recognizes responses that contain only an array of query results like so:

```
 [{"data": {array}},{"data": {of}},{"data": {query}},{"data": {results}}]
```

Attaching that array to a "graphql" key made it unrecognizable to Apollo. We were able to fix the issue by using `render body: result.to_json, content_type: 'application/json'` instead of the example `render` call.

We are not sure if this is an underlying issue with graphql-ruby, Apollo-link-batch-http, or something with our Rails configuration, but we wanted to bring this to your attention. I have opened a PR that adds a brief blurb to the multiplex documentation that suggests our solution if the provided example doesn't work. 

`Rails version: 5.1.4`
`GraphQL Ruby version: 1.8.4 -> unscoped order patch (dhurst's branch) with GraphQL Pro`

Relevant GraphQL Controller Code: 
```
    if params["_json"]
      # We have a batched query
      queries = params["_json"].map do |param|
        {
          query: param["query"],
          operation_name: param["operationName"],
          variables: ensure_hash(param["variables"]),
          context: context
        }
      end
      result = SimplenexusComSchema.multiplex(queries)
      render body: result.to_json, content_type: 'application/json'
    else
      # Single Query
      result = SimplenexusComSchema.execute(
          query: query,
          variables: variables,
          context: context
      )
      render json: result
    end
```

PR #1949, opened 2018-11-08
add tip to documentation

Mando75:
See issue #1948 

Adds a tip to the documentation that could help resolve confusion with query batching and Apollo Client.


rmosolgo:
Thanks for sharing this tip! Sorry for the slow response here 😅

PR #1954, opened 2018-11-09
Fix scalar tracing in NewRelic and Skylight instrumentation

jturkel:
The New Relic and Skylight subclasses of `GraphQL::Tracing::PlatformTracing` weren't plumbing through superclass options.


rmosolgo:
Derp, thanks for the fix here!!

PR #1955, opened 2018-11-12
Add GraphQL::Query::Context#to_hash to allow splatting

### LINKED ISSUE #1821: Splatting Query::Arguments and Query::Context
I recently discovered you cannot simply `**args` in resolvers, because `args` is a duck-typed hash, but is really an instance of `GraphQL::Query::Arguments`

Splatting it breaks the "duck-typing" because Ruby looks for `to_hash`, which is unimplemented

I thought "all this class needs is an alias `to_hash` to `to_h` and it will be a perfectly transparent hash duck". Then I remembered that double-splat needs symbol keys to work and `to_h` returns a string-keyed Hash...

Do the maintainers have an opinion on how to have double-splat work with `Arguments`?

The options could be:
1. Don't
2. Change `to_h` to return a symbol-keyed Hash
3. Add a custom `to_hash` instead of an alias that returns a symbol-keyed Hash

ojab:
Right now extending of fields via something like

```ruby
module ExtendedField
  def initialize(*args, custom_key:, **kwargs, &block)
    …
  end
end
```

because sometimes context is passed as the second argument and it
ends up in args instead of kwargs.

Fixes #1821

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -160,7 +160,7 @@ module GraphQL
       # @api private
       attr_writer :value
 
-      def_delegators :@provided_values, :[], :[]=, :to_h, :key?, :fetch, :dig
+      def_delegators :@provided_values, :[], :[]=, :to_h, :to_hash, :key?, :fetch, :dig
       def_delegators :@query, :trace
 
       # @!method [](key)
--- a/spec/graphql/query/context_spec.rb
+++ b/spec/graphql/query/context_spec.rb
@@ -235,6 +235,16 @@ TABLE
     end
   end
 
+  describe "splatting" do
+    let(:context) { GraphQL::Query::Context.new(query: OpenStruct.new(schema: schema), values: {a: {b: 1}}, object: nil) }
+
+    let(:splat) { ->(**context) { context } }
+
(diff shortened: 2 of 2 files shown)


rmosolgo:
Hi! It's ok with me to implement Context#to_hash, but could you please add: 

- Some rationale for your change to the PR description 
- Some documentation to the new method 
- A test demonstrating the new behavior 

Also, could the method be implemented as `def to_hash; to_h; end`? Why do we need the extra array and `.to_h` call?

ojab:
Done (method is replaced by delegator, so no documentation).

ojab:
BTW, what should be done to have it backported to 1.8?

rmosolgo:
Thanks! Master is still the 1.8 branch, so it will get released in the next 1.8 release. To try it out early, you can use 

```ruby 
gem "graphql", github: "rmosolgo/graphql-ruby", ref: "6a30ba3fef56029665c1b7a7ea59d34fd09b7833"
```

to bundle master right after this merge. Otherwise, keep an eye out for 1.8.12!

ojab:
Ah, I've thought that it's already 1.9. Thanks!

PR #1961, opened 2018-11-16
Add Field#resolver_method, remove `super` on field methods

swalkinshaw:
Adds a new `resolver_method` option to fields which is a breaking change.

Previously `method` was an overloaded option which did two things:

1. specified the method to call on the *underlying* object
2. also specified the method to call on the object type instance

This new option disambiguates this functionality and makes it explicit which one is doing what.

Specifying the method on the underlying object being resolved is assumed to be the most common case, so `method` is kept for that functionality, while the new `resolver_method` specifies the method to look up on the type class instance.

This also makes `method` consistent with `hash_key` as they both now only deal with the underlying resolve object.

Because of the above, `resolver_method` cannot be used in combination with either `method` or `hash_key`.

This also refactors the implementation of field resolution. Before a field *always* called a method on the type instance even just to end up calling a method on the underlying resolve object.

`HasFields` ensured a default "super" method was always defined via metaprogramming.

However, now that a `Field` instance can call `method` directly on the underlying resolve object, the default super method is no longer needed.

cc @chrisbutcher @eapache


rmosolgo:
:+1: Definitely need to disambiguate this somehow, and the name is fine with me.

rmosolgo:
I think it's ok to break in 1.9, I'll be sure to note it in the announcement and changelog.

swalkinshaw:
This is all updated:

* works with new interpreter `resolve` method
* tests all updated
* docs added

One thing of note: I removed the `HasFields` tests entirely since they were all testing the `super` functionality. That's one more minor breaking change within this. I don't think it's a big loss though? No other tests were using it and I doubt many people were in real apps (please correct me if you think that's wrong).

swalkinshaw:
I think it also makes sense to update `Resolver#resolve_method` to `resolver_method` to be consistent with this name? They are the same concept which is nice.

rmosolgo:
Sorry I haven't had time to make a proper review. 

`super` was added by request, so that you could wrap existing resolvers without knowing how they're implemented. Is it possible to maintain the supermethod behavior along with this change?

swalkinshaw:
🤔 maybe, I'll have to dig into it again. It wouldn't work right now since `resolver_method` has a default. It would need to remain `nil` if it's not set.

Considering this would be a breaking change anyway, is `super` worthwhile enough to keep all the extra code and metaprogramming to achieve it?

cmatheson:
I'm using `super` in my resolvers a fair bit (usually to pass extra arguments like this):

```ruby
    implements Interfaces::AssignmentsConnectionInterface
    def assignments_connection(filter: {})
      super(filter: filter, course: course)
    end
```
I guess there's no reason I couldn't call a method defined on `Interfaces::AssignmentsConnectionInterface` instead though.

theorygeek:
@rmosolgo I don’t think we ended up using it much. I’m OK with removing it.

rmosolgo:
Thanks again for your work on this! I also updated the MongoDB tests to use the new option.

theorygeek:
> @rmosolgo I don’t think we ended up using it much. I’m OK with removing it.

I think I misspoke. It turns out we weren't using `super`. However, when we define a field, we check if the return type implements the `Node` interface, and if so, we automatically build a corresponding ID field.

The resolve method for our ID field basically looks like:
```ruby
define_method(id_field_name) do |*args|
  node = public_send(regular_field_name, *args)
  node.try(:node_id)
end
```

Any suggestions on a workaround? Maybe:
```ruby
define_method(id_field_name) do |**kwargs|
  field_inst = self.class.get_field(regular_field_name)
  node = field_inst.resolve_field_method(self, **kwargs, context)
  node.try(:node_id)
end
```

lsanwick:
Is there a way to opt into the old behavior? I'm trying to upgrade to >1.9.0 for subscriptions, but a lot of testing is currently dependent on the old behavior of default methods existing for each field when defined.

We'd have test expectations like:
```ruby
expect(object_type_instance.key).to eq(object.key)
```

that no longer work unless we define a method, even if it's just:
```ruby
def key
  object.key
end
```

Basically it means that testing for our naive/simple fields would differ from more complex ones.  The `resolver_method` need that this issue fixed is one that hasn't come up in our experience so far. (I guess because we've never wanted a `context` or `object` field on a type?)

rmosolgo:
There's no way to opt back into it, sorry. You could monkey patch `def self.field` to define the method if no resolver or `method:` is given.

PR #1973, opened 2018-11-26
Fix fragment are used with invalid frags

xuorig:


lib/graphql/static_validation/rules/fragments_are_used.rb
```diff
@@ -12,7 +12,7 @@ def on_document(node, parent)
         end
 
         dependency_map.unused_dependencies.each do |fragment|
-          if !fragment.name.nil?
+          if fragment && !fragment.name.nil?
```
xuorig: Fragment is `nil` here because it fails the fragment validation and doesnt reach the point where it's added to `unused_dependencies`.

PR #1977, opened 2018-11-29
add relay-modern subscription handler for ably (fixes #1975)

### LINKED ISSUE #1975: [Pro] relay client integration with ably?
i noticed there's an apollo integration but no relay integration. we really want to try out ably but we use relay, was this intentionally left out?

modosc:
this works for me locally


rmosolgo:
Awesome, thanks for your work here! We don't have tests for these yet afaik, so if it works for you ... :shipit: .

rmosolgo:
Shipped in 1.6.1 on npm 🎉

PR #1979, opened 2018-11-30
Support field hash_key: values that are not valid ruby method names

dpkp:
ruby-2.5.1 / graphql-1.8.9

The schema default_resolver crashes if a field uses a non-standard hash_key, like so:

```
field :foo, String, hash_key: :"foo/bar"
field :bar, String, hash_key: "2nd test case"
```

```
SyntaxError (/usr/local/rvm/gems/ruby-2.5.1/gems/graphql-1.8.9/lib/graphql/schema/member/has_fields.rb:121: syntax error, unexpected '/', expecting ';' or '\n'
            def foo/bar(**args)
                   ^
/usr/local/rvm/gems/ruby-2.5.1/gems/graphql-1.8.9/lib/graphql/schema/member/has_fields.rb:124: syntax error, unexpected keyword_end, expecting end-of-input
            end
            ^~~):
```
Putting aside the wisdom of such hash_keys, this PR is intended to allow the user to use any valid string or symbol as a hash_key. I've updated the spec tests to cover at least one such hash_key.

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -127,7 +127,10 @@ Check `inherited` on #{self}'s superclasses.
 ERR
           end
           default_resolve_module.module_eval <<-RUBY, __FILE__, __LINE__ + 1
-            def #{method_name}(**args)
+            # method_name can be a hash_key (see lib/graphql/schema/field.rb)
+            # so we need to make a raw define_method :sym call to avoid crashing
+            # on hash keys like :'some key with/weird-chars'
+            define_method #{method_name.inspect} do |**args|
               field_inst = self.class.fields[#{field_key}] || raise(%|Failed to find field #{field_key} for \#{self.class} among \#{self.class.fields.keys}|)
               field_inst.resolve_field_method(self, args, context)
             end
--- a/spec/graphql/schema/member/has_fields_spec.rb
+++ b/spec/graphql/schema/member/has_fields_spec.rb
@@ -75,6 +75,7 @@ describe GraphQL::Schema::Member::HasFields do
       field :string3, SubSubObjectWithStringField, null: false, method: :object
       field :float1, ObjectWithFloatField, null: false, method: :object
       field :float2, ObjectWithSubFloatField, null: false, method: :object
+      field :hash_key1, String, null: false, hash_key: :'foo bar/fizz-buzz'
     end
 
     class Schema < GraphQL::Schema


rmosolgo:
Hi! Thanks for your work here. I admit that I'm unprepared for hash keys like that 😆 but, you don't always get to choose these things, do ya... 

Anyway, I have a couple of suggestions: 

- Could we add a test that _calls_ the field with the difficult hash key? (I wasn't sure that `public_send(...)` would work with such a method name. I tried it locally and it sure does, but I think it would be good to cover this in the tests too, in case the implementation changes sometime.) 
- Bigger picture: you probably noticed that `Schema::Field` conflates `@method_name` and `@hash_key`, but it probably shouldn't. Do you think we could address this case by fixing that ambiguity? For example, maybe `@method_name` could be `name || method`, while `@hash_key` is only `hash_key` (from kwargs). Then, we could use the field name as the Ruby method name and keep using `def`. (I want to use `def` because it puts the method name in the backtrace, which really helps with debugging.) What do you think of trying that out?

dpkp:
Hi, thanks for the feedback! I bring you strange keys from the world of clojure! Re your suggestions:
 
I can definitely take a stab at your first suggestion.

I will try to look at the second, but I admit that I don't have a deep understanding of this project. So I am hesitant to make any deeper changes related to system design. That said, I share your view that there are conflated bits here that could be separated.

swalkinshaw:
@dpkp if you aren't up to doing the 2nd suggestion, I'd be willing to since I've been doing some related work recently (see https://github.com/rmosolgo/graphql-ruby/pull/1961).

Should note that if #1961 gets accepted, the original bug here wouldn't exist anymore, but I still think it's good to disambiguate these two options as well.

@rmosolgo would you want this for 1.9 as well since it's also a breaking change?

rmosolgo:
> a breaking change?

Do you mean, properly disambiguating `hash_key` and `method`? I'd call it a _breaking fix_, so either way is fine with me :P

rmosolgo:
Hey, thanks again for your work here @dpkp . I cherry-picked your work onto the 1.9-dev branch to show that it was fixed by #1961 : #2016 

So, this behavior will be fixed by 1.9! I'll have a new new prerelease version cut soon. And in the meantime, you can always use `gem "graphql", github: "rmosolgo/graphql-ruby", ref: "1.9-dev"` (or pick a certain sha for `ref:` 😉 ).

dpkp:
Sorry I got distracted by snowy winter vacations! I'm so excited for this -- thanks!!

PR #1991, opened 2018-12-05
Allow multiple errors as args to Schema.rescue_from.

tiegz:
This allows you to pass multiple error types into `rescue_from`'s in your schema, instead of just one:

``` ruby 
  rescue_from(ActiveRecord::RecordInvalid, ActiveModel::ValidationError) { |e| "Validation failed." }
end
```

### Because:

* the underlying `RescueMiddleware#rescue_from` already accepts multiple error classes
* would be nice to follow the same signature as Rails `rescue_from`, which accepts multiple errors

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -855,9 +855,11 @@ module GraphQL
         end
       end
 
-      def rescue_from(err_class, &handler_block)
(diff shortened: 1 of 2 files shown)


rmosolgo:
Awesome, thanks for the improvement!

PR #1998, opened 2018-12-10
Clarify error message on build from definition

emilebosch:
The current error message made me feel like there was no way that I could make an executable schema that handles interfaces.  I forgot that I implemented this myself partially. Hopefully this error message clarifies it a little.

- Make more clear error message


rmosolgo:
Awesome, thanks for improving this!

PR #2000, opened 2018-12-10
Revert "Remove needless nil check"

maletor:
This reverts commit 843a7bc3ad62b92e37a729222473b50ab423c0b3.


rmosolgo:
Thanks for rolling this back! Too bad not to find a test for it, but I understand it's a pretty tricky situation. I added a comment so at least I _might_ not break it next time around!

PR #2003, opened 2018-12-11
Work better with identified Ably client

jscheid:
This avoids warnings when using an Ably "Realtime" client instance
that uses "identified" authentication.


rmosolgo:
Thanks for the addition, released to npm in 1.6.2 🚢 !

PR #2005, opened 2018-12-12
Skip rewrite when interpreter is enabled

rmosolgo:
We have this alternative path that doesn't require `irep_node`s, so let's see if we can skip building them altogether (which is the goal). 

This is a bit different take than #1918 , because you have to explicitly opt into _both_ (but this is kind of good, since you can adopt them separately).


rmosolgo:
fyi @xuorig who worked on Analysis::AST

PR #2013, opened 2018-12-15
Error Bubbling

### LINKED ISSUE #1020: Custom scalar CoercionError clobbers parent messages
When a non-top level field has a custom scalar type that raises a coercion error. The coercion error message overwrites the parent's error message.

This gist shows the difference between a coercion returning nil, raising a `CoercionError`, and what the expected output would actually be.

https://gist.github.com/dnd/c9d6cbf18880f8fd916f962a7e10ca9e

This is from a block of code that looks like this:
```ruby
class SigninUser < GraphQL::Function
  CredentialsInputObject = GraphQL::InputObjectType.define do
    name 'CredentialsInputObject'

    argument :num, Scalars::Date
  end

  argument :credentials, !CredentialsInputObject
 
  ...
end
```
Please ignore the non-sensical use of the name `num`. :)

It would also be nice if instead of only displaying the coercion error message, that the error message be appended to the default string describing which field actually failed. Because as far as I can tell there is no way to determine inside of the coercion what the specific field, or parent is such as `num`, or `InputObject 'CredentialsInputObject'`.

```ruby
"Argument 'num' on InputObject 'CredentialsInputObject' has an invalid value. #{coercion_error.message}"
```

### LINKED ISSUE #1967: better error handling for required fields in an input object
consider something like:
```
input FooInput {
  bar: BarInput!
}
input BarInput {
  baz: BazInput!
}
input BazInput {
  bat: String!
}
```
if i leave out `bat` from my input i get an error like:
```
{      "message": "Argument 'baz' on InputObject 'BarInput' has an invalid value. Expected type 'BazInput!'.",
      "locations": [
        {
          "line": null,
          "column": null
        }
      ],
      "fields": [
        "mutation FooMutation",
        "fooMutation",
        "input",
        "bar",
        "baz"
      ]
    },
```
(sorry this may not be exact, i'm transposing from a work project)

i'd expect the missing required field (`bat`) to be the field referenced in the error message and the path, but currently that's not the case. i think the issue is here:

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/literal_validator.rb#L26

since there's no actual ast node for the missing field, the error message that's generated comes from the parent input object.

modosc:
(this fixes #1020 and #1967) see also #160)

this pr adds a new schema option `error_bubbling`. with this set `true` (which is the default) the behavior is exactly the same as the current behavior, ie field errors bubble up to all the `InputObject`'s that contain them. 

with `error_bubbling false` this doesn't happen - we don't log static validation errors if the ast node that's invalid isn't the node we're visiting in `ArgumentLiteralsAreCompatible`. 

update: i merged in my branch that fixes #1967 as well.

### DIFF
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -10,11 +10,11 @@ module GraphQL
 
       def validate(ast_value, type)
         if ast_value.is_a?(GraphQL::Language::Nodes::NullValue)
-          maybe_raise(ast_value) do
+          maybe_raise_if_invalid(ast_value) do
             !type.kind.non_null?
           end
         elsif type.kind.non_null?
-          maybe_raise(ast_value) do
+          maybe_raise_if_invalid(ast_value) do
             (!ast_value.nil?)
           end && validate(ast_value, type.of_type)
         elsif type.kind.list?
@@ -23,11 +23,11 @@ module GraphQL
         elsif ast_value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
           true
         elsif type.kind.scalar? && constant_scalar?(ast_value)
-          maybe_raise(ast_value) do
+          maybe_raise_if_invalid(ast_value) do
             type.valid_input?(ast_value, @context)
           end
         elsif type.kind.enum?
-          maybe_raise(ast_value) do
+          maybe_raise_if_invalid(ast_value) do
             if ast_value.is_a?(GraphQL::Language::Nodes::Enum)
               type.valid_input?(ast_value.name, @context)
             else
@@ -36,11 +36,11 @@ module GraphQL
             end
           end
         elsif type.kind.input_object? && ast_value.is_a?(GraphQL::Language::Nodes::InputObject)
-          maybe_raise(ast_value) do
+          maybe_raise_if_invalid(ast_value) do
             required_input_fields_are_present(type, ast_value)
           end && present_input_field_values_are_valid(type, ast_value)
         else
-          maybe_raise(ast_value) do
+          maybe_raise_if_invalid(ast_value) do
             false
           end
         end
@@ -49,9 +49,9 @@ module GraphQL
       private
 
 
-      def maybe_raise(ast_value)
+      def maybe_raise_if_invalid(ast_value)
         ret = yield if block_given?
-        if ENV['NO_BUBBLING'] && !ret
+        if @context.schema.disable_error_bubbling && !ret
           e = LiteralValidationError.new
           e.ast_value = ast_value
           raise e
@@ -90,11 +90,6 @@ module GraphQL
         ast_node.arguments.all? do |value|
           field = field_map[value.name]
           field && validate(value.value, field.type)
-          # field && begin
-          #   validate(value.value, field.type)
-          # rescue GraphQL::CoercionError, GraphQL::LiteralValidationError => err
-          #   false
-          # end
         end
       end
 
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -91,8 +91,7 @@ module GraphQL
 
       def present_input_field_values_are_valid(type, ast_node)
         field_map = @warden.arguments(type).reduce({}) { |m, f| m[f.name] = f; m}
-        args = ast_node.arguments
-        args.all? do |value|
+        ast_node.arguments.all? do |value|
           field = field_map[value.name]
           # we want to call validate on an argument even if it's an invalid one
           # so that our raise exception is on it instead of the entire InputObject
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -51,7 +51,7 @@ module GraphQL
 
       def maybe_raise_if_invalid(ast_value)
         ret = yield if block_given?
-        if @context.schema.disable_error_bubbling && !ret
+        if !@context.schema.error_bubbling && !ret
           e = LiteralValidationError.new
           e.ast_value = ast_value
           raise e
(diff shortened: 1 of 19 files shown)


lib/graphql/static_validation/literal_validator.rb
```diff
@@ -10,28 +10,56 @@ def initialize(context:)
...
+            false
+          end
         end
       end
 
       private
 
+
+      def maybe_raise_if_invalid(ast_value)
+        ret = yield if block_given?
```
rmosolgo: What's the use case for this method when there's no `block_given?` If it's never used that way, can we remove this check to make its intention a bit clearer?

lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
```diff
@@ -6,22 +6,52 @@ def validate_node(parent, node, defn, context)
         return if node.value.is_a?(GraphQL::Language::Nodes::VariableIdentifier)
         arg_defn = defn.arguments[node.name]
         return unless arg_defn
+        if !context.schema.error_bubbling
```
rmosolgo: I see the branches below have the same _happy path_, but different error handling, is that right? It looks like `LiteralValidationError` is only raised when `context.schema.error_bubbling` (because of the check in `maybe_raise_if_invalid`, is that right? 

If those assumptions are true, could this `if ... else ... end` be refactored to 

```
begin 
  # happy path as before
rescue GraphQL::CoercionError => err 
  # as before 
rescue GraphQL::LiteralValidationError => 
  # new error handling code 
end 
```

? If so, that would really help me keep track of what different behaviors we expect in different cases, and reduce some duplication!

lib/graphql/static_validation/rules/variable_default_values_are_correctly_typed.rb
```diff
@@ -23,8 +23,8 @@ def validate_default_value(node, context)
           else
             begin
               valid = context.valid_literal?(value, type)
-            rescue GraphQL::CoercionError => err
-              error_message = err.message
+            rescue GraphQL::CoercionError, GraphQL::LiteralValidationError => err
+              error_message = err.message if err.is_a? GraphQL::CoercionError
```
rmosolgo: It looks like one kind of error should be handled one way, while another kind of error should be handled another way, is that right? Why not use different rescue clauses for this, for example 

```
begin 
  # ...
rescue GraphQL::CoercionError => err 
  # as before 
rescue GraphQL::LiteralValidationError 
  # do nothing, but don't let it propagate. 
end 
```

?

modosc:
> The behavior that you've added here is much better than the previous behavior, what do you think about making it the new default ?

i'm ok with it - seems like it would make sense to do that in 1.9 but 1.8 probably should keep the current behavior? i'm also in the unenviable position of having to support the old behavior on a per-user basis for the time being so maybe that's where my conservatism comes from.

lib/graphql/static_validation/literal_validator.rb
modosc: i'll remove it - i guess it's just habit to add that.

rmosolgo: Sure, I guess another option would be to _handle_ the no-block case, for example, by raising a descriptive error. That way, a future user would understand the intention there. I just wasn't sure if it was _supposed_ to work `if !block_given?` in this case, be cause it would _return_!

rmosolgo:
>  unenviable position 

😅 

Sure, we can switch the default in 1.9.

lib/graphql/static_validation/rules/variable_default_values_are_correctly_typed.rb
modosc: fixed.

lib/graphql/static_validation/literal_validator.rb
modosc: i agree that it shouldn't ever be called without a block so i just removed the ✔️

lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
modosc: i think i tried this at one point and ended up introducing some other bugs but it certainly seems like it should work. i'll try again now.

modosc:
also out of curiosity - would you be open to migrating to rspec? test::unit / minitest make it so hard to do basic things like "just run this single spec"

rmosolgo:
Did you try [`focus`](http://graphql-ruby.org/development#unit-tests) for running a single example? Personally I've been happy with minitest and I'm not excited about the operational shakeup of migrating test frameworks.

lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
modosc: ok - cleaned it up and found/fixed a bug that was somehow getting handled properly before.

modosc:
> Did you try focus for running a single example? Personally I've been happy with minitest and I'm not excited about the operational shakeup of migrating test frameworks.

that solved my problem, thanks.

modosc:
also, i started on #1967 and the fix is really just an extension of this (including gating off of `error_bubbling`. might make sense to put that in this branch too? right now i've just forked and i'll push the pr up later today, just a heads up.

modosc:
i just tested this out with my company code and it looks like there's something not quite right. will update when i've sorted it out.

modosc:
just figured it out -  bubbling was never tested or fixed for the `GraphQL::CoercionError` case. this should be a simple fix, working on it now.

modosc:
ok, looks good on my end now.

rmosolgo:
👌 😎  Thanks again for your work here! It's such a long-needed improvement and it really required a big effort. Looks great.

modosc:
> 👌 😎 Thanks again for your work here! It's such a long-needed improvement and it really required a big effort. Looks great.

thanks! next time i'll squash first, i'm horrified to see a `wip` commit checked in.

PR #2021, opened 2018-12-21
add jekyll-redirect-from to Gemfile for fixing 'bundle exec rake site:serve'

technuma:
I got below error message.
```
Configuration file: /Users/technuma/gems/graphql-ruby/guides/_config.yml                                                                                                                                Dependency Error: Yikes! It looks like you don't have jekyll-redirect-from or one of its dependencies installed. In order to use Jekyll as currently configured, you'll need to install this gem. The full
 error message from Ruby is: 'cannot load such file -- jekyll-redirect-from' If you run into trouble, you can find helpful resources at https://jekyllrb.com/help/! 
rake aborted! 
```


rmosolgo:
Woops, thanks!

PR #2022, opened 2018-12-21
update sample code at getting started

technuma:
I updated Getting Started like following image


rmosolgo:
Thanks, that's a nice improvement!

PR #2023, opened 2018-12-24
Simplify `Relay::RelationConnection#sliced_nodes`

v-kolesnikov:
It doesn't change any behavior but just simplifies the code in order to
make it less complicated and more readable.


v-kolesnikov:
Hm, some builds were failed. Will explore it a bit later and reopen again.

PR #2024, opened 2018-12-24
Handle multiple top-level errors

fedorkk:
Good day!
I'm applying gpraphql in my project and got stuck with validation errors. As described in specification: https://facebook.github.io/graphql/draft/#sec-Data

> If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

But the gem suggests us to store validation errors in data entry.
I've discussed it with my frontend developer and he asserts that all errors should be stored in the top-level error entry and all frontend libraries are expected to find it there.
I've added a special class, which can store several errors and then put them in errors entry. It's just a draft because I'm not sure that it would be accepted.
It looks like:
```
GraphQL::ExecutionErrorsMapper.new(
  model_name.errors.full_messages.map { |message| GraphQL::ExecutionError.new(message) }
)
```
```
 "errors": [
    {
      "message": "error1",
      "locations": [
        {
          "line": 4,
          "column": 7
        }
      ],
    },
    {
      "message": "error2",
      "locations": [
        {
          "line": 4,
          "column": 7
        }
      ],
    }
  ]
}
```

Should I continue to work with it?

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -42,6 +42,7 @@ end
 # Order matters for these:
 
 require "graphql/execution_error"
+require "graphql/execution_errors_mapper"
 require "graphql/define"
 require "graphql/base_type"
 require "graphql/object_type"
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -168,6 +168,12 @@ module GraphQL
           query = field_ctx.query
 
           case raw_value
+          when GraphQL::ExecutionErrorsMapper
+            raw_value.errors.each do |error|
+              error.ast_node ||= field_ctx.ast_node
+              error.path = field_ctx.path
+              query.context.errors.push(error)
+            end
           when GraphQL::ExecutionError
             raw_value.ast_node ||= field_ctx.ast_node
             raw_value.path = field_ctx.path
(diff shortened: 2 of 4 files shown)


rmosolgo:
Hi! Thanks for sharing your thoughts and your work here. GraphQL-Ruby used to support returning multiple errors from one field, but not so much anymore. Here's why: 

I think the key part is _here_: 

> If an error is thrown while resolving a field, it should be treated as though the field returned __null__, and an error must be added to the `"errors"` list in the response... [O]nly one error should be added to the errors list per field.

(https://facebook.github.io/graphql/June2018/#sec-Errors-and-Non-Nullability)

When I read the part you posted above, and I see this: 

> prevented a valid response

I think "valid response" means _valid for the GraphQL runtime_. For example, GraphQL has restrictions on what kind of a data a field may return (maybe it can return `nil`, or a list, but maybe not). If an error occurs which would _break_ those restrictions, then we add an error. 

Application logic like data validation or authorization _doesn't_ (or at least _shouldn't_) break GraphQL constraints. For example, if you try to post a comment but leave the body _blank_, that's an invalid _comment_ in the application, but it's valid _GraphQL_. 

I hope I've clarified the difference between: 

- Structural or system-level errors that result in invalid GraphQL responses 
- Application-level considerations which prevent users from taking certain actions 

I think data validation falls in the second category there, and should be represented using normal GraphQL fields and objects. 

Also, if I read the GraphQL specification correctly, it suggests that only one error should be added per field. For that reason, I don't think we should support multiple errors per field!

PR #2028, opened 2018-12-29
Add support for generating a Dash docset

nbudin:
This adds a Rake task for generating a docset compatible with [Dash](https://kapeli.com/dash).  It uses a tool called [Dashing](https://github.com/technosophos/dashing) to make this easier.  The end result looks something like this:

screenshot 2018-12-29 10 17 38

There are quite a few areas for possible improvement here:

* I don't really know Jekyll very well and I think what I did to enable this in the Rake task and the default layout was probably a bit hackish
* The section naming isn't great and it could probably be improved if I understood Dash's indexing a bit better
* For some reason, Dashing isn't generating TOC entries for sections within a page

Ideally I'd like to contribute this to Dash using [their contribution process](https://github.com/Kapeli/Dash-User-Contributions#contribute-a-new-docset).  I'd be happy to take responsibility for keeping this up to date as you make changes to the documentation.


guides/_layouts/dash.html
```diff
@@ -0,0 +1,16 @@
...
+    <link href="https://fonts.googleapis.com/css?family=Rubik:300,400" rel="stylesheet" />
+    <link rel="stylesheet" href="{{ "/css/main.css" | relativize_url }}">
+    <link rel="icon" href="{{ "/graphql-ruby-icon.png" | relativize_url }}">
+  </head>
+  <body>
+    <div class="container">
+      {{ content }}
+    </div>
+  </body>
+</html>
```
rmosolgo: Is this layout still used, or can we remove it?

guides/_layouts/default.html
```diff
@@ -3,45 +3,55 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    {% if page.section contains "GraphQL" %}
-      <title>{{ page.section }} - {{ page.title }}</title>
+    {% if page.target == "dash" %}
+      <title>{{ page.title }}</title>
+      <link rel="stylesheet" href="{{ "/css/main.css" | relativize_url }}">
```
rmosolgo: If I understood right, this outputs `css/main.css` on the homepage, and `../css/main.css` on child pages, is that right? And I guess it's needed because we don't control the path where Dash renders the HTML?

nbudin:
@modosc Yup!  I think these are actually complementary - what you're getting there is the yard docs (which we'd talked about adding above).  What this generates is a Dash version of what's on http://graphql-ruby.org under Guides.

So, possibly yard docs in this docset aren't necessary right off the bat, since it's possible to get them in the way @modosc describes.  It might still be convenient to include them, just to avoid people having to download two different docsets, but possibly that could be added later.  What do you think @rmosolgo?

rmosolgo:
If we could get the website's guides in on RubyDoc.info, do you think that would be sufficient to solve this issue? That would be a win since we wouldn't have to add a new publishing workflow. It looks like yardoc supports "extra files": https://github.com/lsegal/yard/blob/master/docs/GettingStarted.md#documenting-extra-files. 

The only catch is that the guides use some special liquid helpers, so we'd have to either strip those, pre-render the html, or something else 🤔

rmosolgo:
Hi, I'm sorry that this has sat for so long. I'm wary of increasing the complexity of publish docs updates (it's already two commands too many). I'm willing to improve dash support if it can be accomplished with the existing tasks or by automating graphql-ruby's docs & website publishing. 

Thanks for exploring this, sorry not to merge it!

nbudin:
No problem @rmosolgo!  I agree and I think leaving it unmerged is the right call.

PR #2032, opened 2019-01-04
Fix edge case in block string lexer

pushrax:
When there are 4 or more quote characters within a block string, and there is a single non-quote character between an even and odd numbered quote, the lexer failed to match a block quote.

Fixing this also results in a simpler machine. The incorrectness of the previous machine resulted in an extra case that can now be removed.

cc @swalkinshaw @jpittis


rmosolgo:
Sigh, let me see if I can find what's up with the build.

rmosolgo:
Thanks, shipped in 1.8.13!!

PR #2035, opened 2019-01-09
Fix release year in CHANGELOG [ci skip]

y-yagi:


rmosolgo:
what year is it

Thanks for the fix!

PR #2037, opened 2019-01-10
Allow code to be accessed outside of the errors

krider2010:
This allows better reporting of codes without the need to use `to_h`. It's also better for subclasses as they can modify if they need to. Which for fine grained reporting they would want to.

I'm sure I had good reasons to make this private originally, but use cases now show this would be a much better solution.


rmosolgo:
Makes sense! Bummer about the broken build, I fixed this on master but haven't ported the changes to 1.9-dev yet. I'll take a look at that separately. Thanks for the improvement!

PR #2050, opened 2019-01-15
allow using false as enum values :value argument

vala:
Hi there,

I don't know if it's a feature or a bug, but `false` used to be an accepted `:value` argument in enum values, and didn't work when I switched to the latest 1.9.0.pre1 build and the class based schema.

This PR just allows `false` as an enum custom ruby value :

```ruby
class Family < BaseEnum
  description "Groups of musical instruments"
  # support string and symbol
  # ...
  value "SILENCE", "Makes no sound", value: false
end
```

I added an assertion in the test that seemed right to me.
I'm available to update the PR code if needed !

Thanks

### DIFF
--- a/lib/graphql/schema/enum_value.rb
+++ b/lib/graphql/schema/enum_value.rb
@@ -40,7 +40,7 @@ module GraphQL
       def initialize(graphql_name, desc = nil, owner:, description: nil, value: nil, deprecation_reason: nil, &block)
         @graphql_name = graphql_name.to_s
         @description = desc || description
-        @value = value || @graphql_name
+        @value = value.nil? ? @graphql_name : value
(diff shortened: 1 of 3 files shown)


vala:
The failed checks are from ruby 2.3.8 related to gem dependencies. It doesn't seem related to the PR but please tell me if I can do anything about that.

rmosolgo:
This looks great, thanks for the improvement! 

Weird about the build, maybe I have to fiddle with the bundler setup (again 😩  #2033 ).

rmosolgo:
Seems like the merge commit passed! 🤷‍♂️

vala:
Thanks for the merge and glad the build passed !

PR #2056, opened 2019-01-16
Fix block string line counting

### LINKED ISSUE #2054: Lexer returning wrong line number when block strings are used
I was debugging an issue in which a schema that was built from a schema dump (`Schema.from_definition`) had seemingly wrong values for `ast_node.line`.

I managed to pinpoint this to a bug in the Lexer when parsing block strings.

Repro script:

```ruby
require "graphql"

p GraphQL::VERSION

puts

sdl = <<-GRAPHQL
"""
Query
"""

type Query {
  hello: String
}

"""
Node
"""

interface Node {
  id: ID
}
GRAPHQL

sdl.split("\n").each_with_index do |line, i|
  puts "#{(i+1).to_s.ljust(2, " ")} | #{line}"
end

puts

GraphQL::Language::Lexer.tokenize(sdl).each do |token|
  puts token.inspect
end
```

Output:

```
"1.9.0.pre1"

1  | """
2  | Query
3  | """
4  |
5  | type Query {
6  |   hello: String
7  | }
8  |
9  | """
10 | Node
11 | """
12 |
13 | interface Node {
14 |   id: ID
15 | }

(STRING "Query" [1:1])
(TYPE "type" [3:1])
(IDENTIFIER "Query" [3:6])
(LCURLY "{" [3:12])
(IDENTIFIER "hello" [4:3])
(COLON ":" [4:8])
(IDENTIFIER "String" [4:10])
(RCURLY "}" [5:1])
(STRING "Node" [7:1])
(INTERFACE "interface" [9:1])
(IDENTIFIER "Node" [9:11])
(LCURLY "{" [9:16])
(IDENTIFIER "id" [10:3])
(COLON ":" [10:5])
(IDENTIFIER "ID" [10:7])
(RCURLY "}" [11:1])
```

I'd expect the second token (`(TYPE "type" [3:1])`) to be `(TYPE "type" [5:1])`.

I dug into `lexer.rl` for a few minutes and I suspect the problem is that we aren't incrementing `meta[:line]` [when emitting a `:STRING` token](https://github.com/rmosolgo/graphql-ruby/blob/799782d52288f79268300e3d6c6f08afd5340dd3/lib/graphql/language/lexer.rl#L217). Typically, `meta[:line]` gets incremented [whenever the lexer sees a `NEWLINE`](https://github.com/rmosolgo/graphql-ruby/blob/799782d52288f79268300e3d6c6f08afd5340dd3/lib/graphql/language/lexer.rl#L96-L99), but this won't account for new lines that are within the block string.

Once this is fixed, my second concern is, given the code below, what line should we be reporting:

```ruby
require "graphql"

sdl = <<-GRAPHQL
"""
Query
"""

type Query {
  hello: String
}
GRAPHQL

schema = GraphQL::Schema.from_definition(sdl)
p schema.find("Query").ast_node.line
```

Before, when descriptions were comments, the line returned by `ast_node.line` would be the `type` token's `line`. [Since we introduced block strings](https://github.com/rmosolgo/graphql-ruby/blob/799782d52288f79268300e3d6c6f08afd5340dd3/lib/graphql/language/parser.y#L352-L355), the line reported would be the one where the description starts. While this might be correct, it is somewhat of a breaking change for tools that rely on these line numbers (like linters).

rmosolgo:
Fixes #2054 , cc @cjoudrey

### DIFF
--- a/spec/graphql/language/lexer_spec.rb
+++ b/spec/graphql/language/lexer_spec.rb
@@ -95,9 +95,8 @@ describe GraphQL::Language::Lexer do
       GRAPHQL
 
       tokens = subject.tokenize(str)
-      pp tokens
 
-      string_tok, type_keyword_tok, query_name_tok, *rest = tokens
+      string_tok, type_keyword_tok, query_name_tok = tokens
 
       assert_equal 1, string_tok.line
       assert_equal 5, type_keyword_tok.line
--- a/spec/graphql/language/lexer_spec.rb
+++ b/spec/graphql/language/lexer_spec.rb
@@ -82,5 +82,26 @@ describe GraphQL::Language::Lexer do
       rparen_token = tokens[6]
       assert_equal '(RPAREN ")" [1:10])', rparen_token.inspect
     end
+
+    it "counts block string line properly" do
+      str = <<-GRAPHQL
+      """
+      Here is a
+      multiline description
+      """
+      type Query {
+        a: B
+      }
+      GRAPHQL
+
+      tokens = subject.tokenize(str)
+      pp tokens
+
+      string_tok, type_keyword_tok, query_name_tok, *rest = tokens
+
+      assert_equal 1, string_tok.line
+      assert_equal 5, type_keyword_tok.line
+      assert_equal 5, query_name_tok.line
+    end
   end
 end
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -1417,9 +1417,13 @@ def self.emit_string(ts, te, meta, block:)
 quotes_length = block ? 3 : 1
 ts += quotes_length
 value = meta[:data][ts...te - quotes_length].pack(PACK_DIRECTIVE).force_encoding(UTF_8_ENCODING)
+line_incr = 0
 if block
+line_incr = value.count("\n")
 value = GraphQL::Language::BlockString.trim_whitespace(value)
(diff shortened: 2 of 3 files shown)


spec/graphql/language/lexer_spec.rb
```diff
@@ -82,5 +82,25 @@
...
       assert_equal '(RPAREN ")" [1:10])', rparen_token.inspect
     end
+
+    it "counts block string line properly" do
+      str = <<-GRAPHQL
+      """
+      Here is a
+      multiline description
+      """
+      type Query {
```
cjoudrey: I'm a bit paranoid. I think this works, but could you include another type definition after this one that includes a line break between the `type ..` and the description?

i.e.

```graphql
"""
Test
"""

interface Node {
  id: ID!
}
```

Also maybe include one where the description uses a string without line breaks, i.e.:

```graphql
"Test"

interface Node2 {
  id: ID!
}
```

rmosolgo:
Oops! Sorry, I merged this in a stale window and didn't see your comments. Good call on those tests, I pushed them here: 4f0a4e390

daemonsy:
FYI we also fixed this in the `1.10-dev` branch where there's an issue with tokenizing values. In case you have a merge conflict, the other part of the work there is to handle empty block strings. 

https://github.com/rmosolgo/graphql-ruby/pull/2381

PR #2066, opened 2019-01-20
[WIP] Transpile and bundle JavaScript library

hootlex:
Todos:
- [x] set up babel / rollup
- [x] publish 2 packages with lerna 
- [x] bundle subscriptions
- [ ] bundle sync
- [ ] clean up dependencies in packages

This PR introduces module bundling with rollup and trasplilation with babel (env preset) to ensure browser compatibility. Since the JavaScript application contains multiple libraries (`subscriptions`, `sync`, `cli`) that target different environments, they should also be bundled as multiple libraries. Here I have set up the flow for the `subscriptions`.

### The code structure:
- `/src` - the source code has been moved to `/src`
- `/dist` - contains the built bundles

### The flow
The `yarn build` command will process each file under `src/subscriptions` with babel and will create an `es` and a `umd` bundle under the `dist/subcriptions` directory. I've also set up an index in this directory so developers can import a library without specifying the filename.

ex: `import {PusherLink} from 'graphql-ruby-client/subscriptions'`

### Tests
For the refactor I have been relying and fixing the existing tests. No new test cases have been added.

If you agree with this pattern, I will implement the same for `sync` and the `cli`. 

---

BTW I'd recommend organising each library in an individual repository and npm package to make everything more clear. This way, devs will be able to pull in only the `graphql-ruby-client-subscriptions` module for example and the usage will be slightly more pleasant. Though, that might also increase the maintenance overhead and requires quite some work.


rmosolgo:
Thanks again for your work on this! I've been taking a closer look this morning. 

__Lerna__ seems like a great choice, thanks for adding that. I was reading: 

>  ...develop all of their packages within a single repository.

👌 That's just what I want to do!

I have a couple questions: 

- For packaging and versioning, I'd like to use `lerna publish` and lerna's "fixed" versioning mode. That way, we can start both of the new packages as `1.7.0` and keep them both in sync. Does that sound reasonable? If so, I'm happy to hack on the publish flow for the two packages a bit. 
- In the past, I've maintained projects where the "compiled" (`dist/`) code was pushed to NPM but _not_ versioned in the repository, since it often got out of date or added a lot of noise to diffs. Would it make sense to ignore `dist/` in this project, too?

Here's some changes I made to work on this, trying to reorganize tests a little bit and update the various `package.json`s: https://github.com/hootlex/graphql-ruby/compare/babel-rollup-subscriptions...rmosolgo:babel-rollup-subscriptions-rmosolgo?expand=1 Do those changes look good to you? If so, feel free to merge them.

hootlex:
You're welcome and happy to hear you like the approach. 🙂
Did you notice any issues while testing the built libraries? Also, were you able to take a look on the `cli`?

Could you publish a draft version to npm to test that the workflow actually works?

---

> For packaging and versioning, I'd like to use `lerna publish` and lerna's "fixed" versioning mode. That way, we can start both of the new packages as `1.7.0` and keep them both in sync. Does that sound reasonable? If so, I'm happy to hack on the publish flow for the two packages a bit.

Sure, that's reasonable! It should work like that without any changes since `fixed` is the default lerna mode. Though, maybe consider bumping the version to `2.0.0` since this is a major upgrade?

> In the past, I've maintained projects where the "compiled" (`dist/`) code was pushed to NPM but _not_ versioned in the repository, since it often got out of date or added a lot of noise to diffs. Would it make sense to ignore `dist/` in this project, too?

Sounds great! I also prefer ignoring the output from version control.

The suggested changes look good, it's cleaner to have the tests under each lib's directory. 👍 Merged!

bbugh:
Will this modularize the imports for single sources? 

Currently, (with TypeScript) I have to do this:

```
import * as ActionCableLink from 'graphql-ruby-client/subscriptions/ActionCableLink'
```

I'd rather do this:

```
import { ActionCableLink } from 'graphql-ruby-client-subscriptions'
```

It's harder to correctly type when we're importing from subdirectories, and (if I understand correctly) we'd have to provide types for the entire `graphql-ruby-client` if submitting to DefinitelyTyped, when I want to type just the subscriptions (which is all I use).

daemonsy:
We use the Javascript sync functions (as API) with our webpack build process to sync persisted queries. This is awesome and I'll be happy to help test this out. 

I see conflicts against master, do this matter at this point?

rmosolgo:
Sorry to let this branch get cold. I've started rewriting the client in TypeScript, instead: #2745

PR #2071, opened 2019-01-23
Fix grammar error

billfienberg:
Before: `GraphQL::Pro` integrates has out-of-the-box
After: `GraphQL::Pro` integrates out-of-the-box


rmosolgo:
Thanks for the fix!

PR #2078, opened 2019-01-30
Handle visitor when no operation is selected

xuorig:
The AST visitor is a special case visitor that only visits the selected operation of a query (what will actually be ran), which emulates was the Internal Representation used to do. The irep also took care of queries with no selected operations, which was missing here.

This PR returns early when no @document is present for the visitor (when there was no selected operation on the query)

### DIFF
--- a/lib/graphql/analysis/ast/visitor.rb
+++ b/lib/graphql/analysis/ast/visitor.rb
@@ -32,6 +32,11 @@ module GraphQL
         # @return [Array<GraphQL::ObjectType>] Types whose scope we've entered
         attr_reader :object_types
 
+        def visit
+          return unless @document
+          super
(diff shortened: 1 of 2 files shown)


xuorig:
@rmosolgo hmm I wasn't sure because the base class is never meant to be initialized with `nil`, this class initializes with `query.selected_operation` which could be `nil` with certain queries. WDYT?

rmosolgo:
Ohhh I see, the key bit is: 

https://github.com/rmosolgo/graphql-ruby/blob/29c5edca32922ce25bad07a17651e0f30195f754/lib/graphql/analysis/ast/visitor.rb#L26

I didn't have that on my radar, thanks for clarifying!

Issue #1838, opened 2018-09-07
Unknown loads in input object


mcdave029:
Hi!, is `loads:` in argument are only available from [mutation classes](http://graphql-ruby.org/mutations/mutation_classes.html)?. because right now I'm using it in [input objects](http://graphql-ruby.org/type_definitions/input_objects.html) and I'm getting an error of `unknown keyword: loads`. Thanks for the help. 😄

lemuelbarango:
Hey, @rmosolgo any work/progress on this? I don't mind working on a PR for this. I have just encountered a similar situation at work, trying to use the `loads` argument in the input object type.

I assume that the work involved here would be to reimplement the `loads` logic in the resolvers class for input objects - 

https://github.com/rmosolgo/graphql-ruby/blob/3642ee92b722bccf49562fef9608eaab4e339711/lib/graphql/schema/resolver.rb#L329-L370

Let me know if you had better ideas. 😄 

### Thanks for the efforts put into building this gem!! 🙌

lemuelbarango:
Sweet! I create a PR https://github.com/rmosolgo/graphql-ruby/pull/2080

I welcome any early feedback on the initial refactor... Thanks

duderamos:
Hey @lemuelbarango, what happened to close the PR?

lemuelbarango:
@duderamos I've reopened the PR. was a mistake

duderamos:
No worries @lemuelbarango. I don't know if it helps, but I tested a scenario on my end and using `prepare` param seems enough:

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/argument.rb#L31

In my case, I have something like this:

```ruby
module Foo
    class UpdateBarAttributes < Types::BaseInputObject
      argument :my_object_id, ID, required: true, prepare: :load_object
  end
end
```

```ruby
module Types
  class BaseInputObject < GraphQL::Schema::InputObject
    def load_object(encoded_id)
      MySchema.object_from_id(encoded_id)
    end
  end
end
```

So I just need to append `prepare: :load_object` to my `ID` arguments.

lemuelbarango:
@duderamos Yeah, I didn't know there was a `prepare` param on the `InputObject`. That's good to know.

swalkinshaw:
Fixed by https://github.com/rmosolgo/graphql-ruby/pull/2080

PR #2080, opened 2019-01-30
Allow autoloading of relay ids in input objects, move `GraphQL::Schema::Resolver::LoadApplicationObjectFailedError` to `GraphQL::LoadApplicationObjectFailedError`.

lemuelbarango:
Fix for this issue: https://github.com/rmosolgo/graphql-ruby/issues/1838

Initial idea was to make `GraphQL::Schema::InputObject` inherit from `GraphQL::Schema::Resolver`...

Decided to go for a simpler PR and move loads functionality from `GraphQL::Schema::Resolver` class to `GraphQL::Schema::Member::HasArguments` module. This allows the functionality to be reused in the `GraphQL::Schema::InputObject` class

**Open for feedback or different ideas.**

TODO

- [x] Add more specs to test `loads:` arguments on input object


lemuelbarango:
I should note the `loads` argument in input objects works only in conjunction with  `MutationType`

For this to work with `QueryType` I think the `InputObject` resolve would have to be refactored

chrisseto:
@lemuelbarango, how difficult do you think it would be to get `loads:` working with `QueryType`?

It would be extremely useful to have it work in both cases.

@rmosolgo, do you think the loading logic could be refactored onto/into the `Context` class?

lemuelbarango:
@chrisseto added should work with `QueryTypes` now

chrisseto:
@lemuelbarango 🎉 you da real MVP

rmosolgo:
Thanks for your work here, I'll take a close look soon!

rmosolgo:
Thanks again for taking this on, looks like it turned out really nice! I just made one change, moving the error class to the top-level `GraphQL::` namespace. This might be a breaking change for people who reference that error in their code, but the new namespace is simpler and we can be sure to make it stable from now on 💫

PR #2088, opened 2019-02-05
Update analyzer migration docs

rmosolgo:
Update the doc with the migration path we settled on, cc @swalkinshaw who had asked about this earlier.


swalkinshaw:
Looks great thanks 🎉

PR #2090, opened 2019-02-05
performance: Replace use of Enumerable#none? without a block with empty?

dylanahsmith:
Enumerable#none? was being used without a block to check if enumerables were empty, even though its behaviour is documented as

> If the block is not given, none? will return true only if none of the collection members is true.

so `Enumerable#none?` should only be used if this is actually the behaviour we want.  A comment in the code (`if path.size == 0 # faster than #none?`) both shows that this is significant to performance and that `none?` was being used when `empty?` was more appropriate.

Let me know if any of the replaced `none?` calls were actually needed to handle falsey values.  If so, then the tests should reflect this.

### DIFF
--- a/lib/graphql/execution/interpreter/hash_response.rb
+++ b/lib/graphql/execution/interpreter/hash_response.rb
@@ -21,7 +21,7 @@ module GraphQL
         # Add `value` at `path`.
         # @return [void]
         def write(path, value)
-          if path.size == 0 # faster than #none?
+          if path.empty?
             @result = value
           elsif (write_target = @result)
             i = 0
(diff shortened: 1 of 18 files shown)


rmosolgo:
Wow, TIL! Thanks for the fix.

rmosolgo:
I added a cop to catch this in the future, it failed a lot when I ran it before merging this change, for example:

```
lib/graphql/query/validation_pipeline.rb:92:18: C: Cop/NoneWithoutBlockCop: Instead of .none? without a block:

- Use .empty? to check for an empty collection (faster)
- Add a block to explicitly check for false (more clear)

Run -a to replace this with .empty?.

        @valid = @validation_errors.none?
                 ^^^^^^^^^^^^^^^^^^^^^^^^
```

And had no failures after merging your changes 💰

PR #2093, opened 2019-02-06
Add field class-level extensions

rmosolgo:
Let's say you want to affect field resolution in the _whole schema_, this is a way to do it. 

TODO:

- [x] could the built-in Scope and Connection extensions be pushed to this system?


rmosolgo:
I'm not sure this is really necessary yet.

PR #2095, opened 2019-02-08
Fix #2085 - Ensure lists are arrays

### LINKED ISSUE #2085: input is not coerced into a list of one if there is a type error in the InputObject
**Gem Version:** v1.8.13

**Issue:** When an InputObject expects a List, but only one item is provided, it is unneccecasry to wrap the object in square brackets because it is coerced into a List of one. If there is an error in that single list item, the coercion of the object to a List of one never happens which raises an error.

**Example**

```graphql
mutation {
  productCreate(
      input: {
          id:  "gid://shopify/Product/id",
          title: "A Title", 
          variants: {
              price: 17, 
              weight: 1, 
              weightUnit: "POUNDS"
            }
        }
    ) {
    product {
      id
    }
  }
}
```

The `variants` field expects a List, but given we only have a single variant here, it should be coerced into a List of one automatically. The `weightUnit` expects an Enum value, not a String. `weightUnit: "POUNDS"` is invalid and the line should read `weightUnit: POUNDS` instead.  

The invalid input correctly raises `GraphQL::LiteralValidationError` and is rescued, but fails in the rescue block here in [lib/graphql/static_validation/rules/argument_literals_are_compatible.rb#L25](
https://github.com/rmosolgo/graphql-ruby/blob/v1.8.13/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb#L25) with `undefined method 'include?' for <GraphQL::Language::Nodes::InputObject:0x00005567ad947f40>`

Because `"POUNDS"` is not a valid enum, something goes wrong and `variants` is never coerced into a List, even though `arg_defn.type.kind.list?` evaluates to true (arg_defn.type is `[ProductVariantInput!]` at the type of the error)

CC: @rmosolgo @eapache @swalkinshaw

swalkinshaw:
Lists in GraphQL can omit the square brackets if there's only one item but the validation code was assuming it would be an array.

This coerces the node's `value` to ensure it is an array.

Note: I figured there might be a more proper fix for this in a different place but it can't be in the parser since there's no knowledge of the argument's type (correct me if I'm wrong). Is there another place I'm not thinking of?

This is also branched off of `v1.8.13`. If this is acceptable, could we backport and do a new 1.8.14 release? (since master is on 1.9 now)

### DIFF
--- a/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
+++ b/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
@@ -22,7 +22,7 @@ module GraphQL
           # the same as the node we were checking here.
           matched = if arg_defn.type.kind.list?
             # for a list we claim an error if the node is contained in our list
-            node.value.include?(err.ast_value)
+            Array(node.value).include?(err.ast_value)
           elsif arg_defn.type.kind.input_object? && node.value.is_a?(GraphQL::Language::Nodes::InputObject)
             # for an input object we check the arguments
             node.value.arguments.include?(err.ast_value)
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: Thanks for the fix, shipped in 1.8.14 and merged in to master: 0cabb4538

PR #2101, opened 2019-02-11
Bound integer scalars per the spec

eapache:
@rmosolgo @swalkinshaw I'm not 100% sure that `ExecutionError` is correct here but my dev env is acting up. I'll validate and write tests tomorrow.

Also would appreciate eyes on the logic, I'm not sure if I've made an off-by-one error or not.


alunbestor:
Am I correct that we've decided against silently clamping, in order to force APIs to clamp their `Int`s before they reach the GraphQL layer? Are we going to be able to audit the need for that, short of just logging runtime failures?

eapache:
The spec suggests that out-of-bound integers should raise a field error (https://facebook.github.io/graphql/June2018/#sec-Int). While clamping probably makes sense in some cases, it certainly doesn't make sense for all of them so silently clamping seems like a dangerous default for the gem.

Unfortunately, because most of these values are dynamic I don't know of a way to audit them other than just waiting for the runtime errors to show up.

eapache:
As written this PR *almost* does the right thing, excepting the fact that `ExecutionError` details end up in the response payload rather than being turned into a generic "Something went wrong".

I don't know if the gem has the concept of a "private" error right now, where the client just gets "Something went wrong" but the server can log the details to be fixed?

rmosolgo:
>  concept of a "private" error

It doesn't 😖  (We usually return the backend error message to the client, for example, with non-null errors)

rmosolgo:
Maybe you could take a page out of `String`'s book, how it handles non-UTF8 strings by passing them to `schema.type_error`: 

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/string.rb#L12-L13

That gives the server something to handle properly.

eapache:
Oh perfect, that's what I was looking for.

eapache:
OK, this should be ready for proper 👀 now.

rmosolgo:
I added a little refactor so that `.type_error`'s return value is used, similar to `String`. That gives you the option of silently clamping, or logging while returning the value, or something else. I'll take it for a spin in GitHub tomorrow morning and if it looks alright to you, cut one ... "last" `.pre` version 😆

eapache:
LGTM. Definitely surprised by how much slower `between?` is.

swalkinshaw:
I had checked `cover?` before and it's even slower since you need to create a `Range` first.

rmosolgo:
for the record, to maintain previous behavior in our app, I added roughly: 

```ruby 
class Schema << GraphQL::Schema 
  def self.type_error(err, ctx)
    if err.is_a?(GraphQL::IntegerEncodingError)
      # Since we started returning large integers _before_ graphql-ruby started (correctly) checking integer bounds, 
      # continue returning out-of-bounds values, for compatibility.
      err.integer_value 
    else 
      super 
    end 
  end 
end 
```

TSMMark:
if anyone else ends up here, I'll post this for the record. @rmosolgo there's a typo in your comment. The class definition should only have one angle bracket, like:

```ruby
```

The correct example code for `type_error` method is in the changelog here: https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#breaking-changes-2

```ruby
  def self.type_error(err, ctx)
    if err.is_a?(GraphQL::IntegerEncodingError)
      # Preserve the previous unbounded behavior
      # by returning the out-of-bounds value
      err.integer_value
    else
      super
    end
  end
end
```

rmosolgo:
Derp, thanks for pointing that out. I've corrected my snippet above!

TSMMark:
No problem, thanks for the fantastic library!

Issue #1882, opened 2018-10-04
Subscription trigger field resolution is fixed to "Subscription" class name


wealth:
Hi. When subscription type classname is different from Subscription (ex. AuctionsSubscription), field resolution is returning "No subscription matching trigger" error. I see that this is mostly depends on a project code organization preference, but probably, there should be a way to get_field only from supscription types, without a subscription classname.

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions.rb

```
      field = @schema.get_field("Subscription", event_name)

      if field.nil?
        # And if it wasn't found, normalize it:
        normalized_event_name = normalize_name(event_name)
        field = @schema.get_field("Subscription", normalized_event_name)
```

rmosolgo:
Oops! You're right, it shouldn't be hardcoded that way, it should probably be: 

```
field = @schema.get_field(@schema.subscription, normalized_event_name)
```

PR #2102, opened 2019-02-12
Fix bug in subscriptions#trigger

cschiewek:
See: #1882 

There's a bug when attempting to assign the field, it's using the hard-coded `Subscription` string instead of pulling the subscription from the schema.

This fixes it.


rmosolgo:
Oops! Should the lookup on line 46 use the same approach?

cschiewek:
D'oh!  You're correct sir!  I've fixed the other call.  Tx!

rmosolgo:
Thanks!

PR #2106, opened 2019-02-13
Use urlsafe_encode64 for object Id's

jgodson:
Uses `Base64Bp.urlsafe_encode64` encoding for Node id's so that they are safe for use within url's.

We have been using a generated id returned from graphql for unique identification in a route (eg: `/app/SW1wb3J0LTY3`) and in order to safely do this, we needed to make this change. I thought I'd try to contribute it back as it seems like a nice default to use the url safe version if anyone else has done/wants to do the same.


lib/graphql/schema/unique_within_type.rb
```diff
@@ -21,7 +21,7 @@ def encode(type_name, object_value, separator: self.default_id_separator)
           raise "encode(#{type_name}, #{object_value_str}) contains reserved characters `#{separator}` in the type name"
         end
 
-        Base64.strict_encode64([type_name, object_value_str].join(separator))
+        Base64.urlsafe_encode64([type_name, object_value_str].join(separator))
```
swalkinshaw: I think this should use `Base64Bp`? like the `decode` method does below

jgodson: I did see that, but I wasn't sure what the difference was or the reason for doing it. I can change it though 👍

rmosolgo:
Hi! Thanks for upstreaming this. I _really_ wish I had had the foresight to put this default from the very start, but alas, I didn't 🤦‍♂️ . 

I totally agree that URLsafe base64 is better, but here's my hangup: if we change the default in GraphQL-Ruby, it might _break_ people's IDs who are using this encoder. I'd hate for someone to bump the gem but accidentally break IDs across their system. 

So, how can we improve the default without giving that unpleasant surprise? 

- Wait until graphql-ruby 2.0. I think it would be fair to make this change there, including a compatibility snippet in the changelog for people who want to keep the previous behavior 
- Instead of modifying the existing encoder, we could add a new, URL-safe encoder that people could choose. We could update the Rails generators to recommend the new encoder for new apps and mark the existing one as deprecated. (And even make the new one default in 2.0, as described above.) 

What do you think, would one of those approaches do the job?

jgodson:
Yes, absolutely, wouldn't want this to break anything. Either of your suggestions are great! 

My thoughts on the second is that I wonder if it's worth the time to do vs just waiting until 2.0 and making it the default? I don't know if there's a good reason, beyond backwards compatibility, to use the existing encoder? 

If there is, then I think that's a nice approach. Otherwise, I think the first makes the most sense.

lifeiscontent:
@jgodson it'd probably be a good idea to let people opt-in to this behavior before 2.0 then make it default?

Issue #2100, opened 2019-02-11
Release 1.10.0


rmosolgo:
Here are some goals I have for 1.10.0. Please feel free to contribute to the conversation by leaving a comment. 

- Redo pagination, see #1359
  - [x] Implement new system #2143 
  - [x] ~~Turn it on by default~~ Add it to generator
  - [x] Improve docs (conceptual & how-to)
- [x] Remove the schema build phase, instead, construct the type caches _during_ configuration methods. 
  - In the works at #2363 
  - The goal is to fix the thread-safety issues that result from a lazily-constructed schema 
  - Also, it will move `GraphQL::Schema` _away_ from an internal dependence on `.define`
- ~~Consider a built-in batching implementation~~
  - ~~Include features from #1981~~
  - ~~In the works at #2483~~
  - I'm going to cut this because it's time to get these other changes released!
- [x] #2178 Update `Schema.from_definition` to build a class-based schema 
  - Mainly to remove the internal dependence on `.define`.
- [x] Update some specific docs, which were added a long time ago  (this can happen during the 1.10 cycle)
  - [x] #2177 Testing: add a whole section with tips and examples
  - [x] https://github.com/rmosolgo/graphql-ruby/pull/2179 Dynamic definition: just remove it?
  - Backtrace annotations (todo, port this to interpreter?)
  - [x] #2174 Instrumentation: remove it
  - [x] #2175 Remove the blue notice about class-based API being new
- [x] Mark legacy `.define`-based API docs as `@api deprecated`
- [x] update the schema generator to default to the interpreter etc.
- [x] Figure out when interpreter is required? (Is it required for printing class-based schemas?) 
- [x] Run Ruby 2.7 with no warnings

panthomakos:
Regarding pagination, the goals you have laid out in #1359 to support a clear and more explicit mapping for cursor based pagination is :+1:. It took us a long while to decode and extend the existing relay pagination for our own purposes. The method names, purposes, and the integration with GraphQL were not very clear.

:+1: to batching implementation. In particular, if (:crossed_fingers:) the concurrency abstraction is included in the next release, having a thread-safe batch implementation will be increasingly important. One more thing I would add here, existing implementations of loaders (like `graphql-batch`) effectively require an entirely new class to be created for each query or network call. This can become very cumbersome, so it would be great to provide a more lightweight way to have concurrent and batch loading lazy objects.

[...]

PR #2115, opened 2019-02-14
Support extensions with resolve procs

### LINKED ISSUE #2114: error: Extensions are not supported with resolve procs or functions
Upgrading from 1.8.13 to 1.9.1 introduces this error:

```
/home/circleci/backend_app/vendor/bundle/ruby/2.6.0/gems/graphql-1.9.1/lib/graphql/schema/field.rb:281:in `extensions': Extensions are not supported with resolve procs or functions, (ArgumentError)
but Types::ClassroomType.availabilities has: #<Proc:0x0000560b18e4f5d8@/home/circleci/backend_app/app/graphql/types/base_object.rb:14 (lambda)>
So, it can't have extensions: [].
Use a method or a Schema::Resolver instead.
```

Not really sure what the error is about. Is it saying that I can't have any extensions and that `[]` is an extension? Or is it saying that I can't have any extensions and that my list of extensions is empty?

The failing field appears to be:

```
module Types
  class ClassroomType < BaseObject
    field :availabilities, [ClassroomAvailabilityType], null: false
  end

  class ClassroomAvailabilityType < BaseObject
  end
end
```

rmosolgo:
Fixes #2114 

Interestingly, support for `extensions:` with legacy resolve procs was already _halfway_ there. It just had to be ported to the other kind of resolution.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -277,15 +277,6 @@ module GraphQL
           # Read the value
           @extensions
         else
-          if @resolve || @function
-            raise ArgumentError, <<-MSG
-Extensions are not supported with resolve procs or functions,
-but #{owner.name}.#{name} has: #{@resolve || @function}
-So, it can't have extensions: #{extensions}.
-Use a method or a Schema::Resolver instead.
-MSG
-          end
-
           # Normalize to a Hash of {name => options}
           extensions_with_options = if new_extensions.last.is_a?(Hash)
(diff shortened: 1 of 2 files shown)


cainlevy:
I presume that I should look at refactoring these resolve procs to local methods on the type?

rmosolgo:
Yep, I plan to officially deprecate it [in 1.10.0](https://github.com/rmosolgo/graphql-ruby/issues/2100)

PR #2119, opened 2019-02-17
Builtin default resolver for Boolean fields

ashkan18:
# Change
Ruby prefers using `?` for class attributes/methods returning boolean. Currently when we define a field as `Boolean` by default we'll call `<field_name>` method or `hash` key on the object.

If the underlying field is using `?` we need to manually define that by something like
```ruby
field :boolean_property, Boolean, method: :boolean_property?
```

This PR updates the code to check for existence of `<field_name>?` method for Boolean fields and if they exist it would call that, otherwise fallbacks to existing logic.

### DIFF
--- a/guides/fields/introduction.md
+++ b/guides/fields/introduction.md
@@ -96,6 +96,14 @@ field :top_score, Integer, null: false
 
 The default behavior is to look for a `#top_score` method, or lookup a `Hash` key, `:top_score` (symbol) or `"top_score"` (string).
 
+In case field type is `Boolean` and object responds to `<field_name>?` (symbol) we will use that otherwise we will call `<field_name>` In following example:
+
+```ruby
+field :is_winner, Boolean, null: false
+```
+
+We first attempt calling `#is_winner?` if doesn't exist we attempt `#is_winner` and `Hash` key.
+
 You can override the method name with the `method:` keyword, or override the hash key with the `hash_key:` keyword, for example:
 
 ```ruby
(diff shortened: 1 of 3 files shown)


rmosolgo:
Hi, thanks for making this recommendation, but I don't want to accept it right now because it doesn't match my opinion of how this should be done. 

In my opinion, it's useful for method names to indicate what kind of values they return. (Maybe I should just program in properly-typed language. Too late I guess 😆 ). In Ruby, we indicate method returns boolean appending `?` to the method name, for example `item.valid?`. In languages _without_ `?` in method names, they often prepend `is`/`has`/`can` or another helper verb, for example `obj.hasOwnProperty(...)` (JavaScript) or `django_form.is_valid()` (Python). I prefer the same convention in GraphQL, using a helper verb (`is`/`has`/`can`, etc) to indicate boolean fields. So, I'd recommend something like: 

```ruby 
  field :is_administrator, Boolean, null: false, method: :administrator?
```

So, since I prefer that approach, I don't want to include support for a different approach. 

However, I'm very happy to support your own convention in your work. If you're using class-based graphql, it's fairly easy to implement in your base field class: 

```ruby 
  def initialize(name:, type:, method: nil, **rest, &block)
    if type.unwrap.graphql_name == "Boolean" && method.nil?
      method = "#{name}?".to_sym 
    end 
    super(name: name, type: type, method: method, **rest, &block)
  end 
end 
```

I think that would do the trick. 

If you're not using class-based GraphQL, it will be a bit trickier, because the `.define` APIs are harder to customize. (In fact, I don't think the patch here will work as documented, since it's implemented for `.define`-based fields but documented for class-based fields. Sorry about the confusing structure here ... I'll remove `.define` altogether in v2.0!)

PR #2121, opened 2019-02-18
Relay 2.0.0 support

HermantNET:
Relay compilers generated files exports changed from
```
module.exports = { ..., name }
```
to
```
module.exports = { ..., params: { name } }
````


rmosolgo:
Cool, thanks for the fix here!

Issue #2109, opened 2019-02-13
Nested InputObject required flag not being checked


chrisseto:
Given the following input objects and assuming they're hooked up to a relay-ish mutation called `makeRoot`

```ruby
class Child < GraphQL::Schema::InputObject
  argument :name, String, required: true
end

class Root < GraphQL::Schema::InputObject
    argument :children, [Child], required: true
end
```

The following mutation passes validation

```graphql
mutation {
  makeRoot(input: {
    children: [{
      name: "foo"
    }, {
      # No name attribute
    }]
  }) { }
}
```

I tracked the issue down to the [static validator](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/rules/required_input_object_attributes_are_present.rb#L14).
`context.field_definition` is the definition of `makeRoot` but `get_parent_type` is searching for an argument named `children`, which doesn't exist.

I was able to fix my specific test case by tossing this into the bottom of the method but I can imagine a few issues with it.
```ruby
return parent_type.type.unwrap unless parent_type == nil

path = context.path[context.path.rindex(defn.name) + 1..-1]
path[1..-1].reduce(defn.arguments[path[0]].type.unwrap) do |type, field|
  type.arguments[field].type.unwrap
end
```

What do you think is the safest way to get the "path" from the root input object to the argument at hand?

I'll see if I can have a PR up with a failing test case and/or fix soonish but wanted to get your thoughts first.

rmosolgo:
Hey, thanks for reporting this bug! 

A failing test case would be awesome. 

It looks like the validation context has a method that might help here:

https://github.com/rmosolgo/graphql-ruby/blob/8eaef47cf0498f0012a3132a7ea9402c46b89208/lib/graphql/static_validation/base_visitor.rb#L168-L173

It seems like we want: 

- The argument that this argument belongs to, if there is one
- Or, if there is no parent argument, then we want the field this argument belongs to. 

If that's right, maybe we could implement it like: 

```diff 
- defn = context.field_definition
+ defn = if (arg_defn = context.argument_definition)
+   arg_defn.type.unwrap # Get the input object type, which has `.arguments`
+ else 
+   context.field_definition 
+ end 
```

Do you think that might do the job?

chrisseto:
@rmosolgo, that worked with a little extra checking.
PR is linked, thanks for help!

PR #2123, opened 2019-02-18
Correct static validation of nested InputObjects

chrisseto:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/2109


rmosolgo:
Thanks you for taking care of this! 🎉

PR #2127, opened 2019-02-19
Fix bug in Relay::ArrayConnection#hasNextPage (bi)

LemonAndroid:
bi = bidirectional


rmosolgo:
Hi! Thanks for sharing this patch. Could you tell me more about what scenario this fixes? I'd love to get a test in the test suite so that we can make sure to maintain the proper behavior.

LemonAndroid:
Given the following context:

Page size is 10
Total items are 21
Current page is 3

Flow:

The user paginates backward to page 2 and won't be able to paginate back to page 3 since `hasNextPage` is `false`.

rmosolgo:
I think this was covered by a recent PR which also tested the change: #2349 Thanks for taking a look at this!

PR #2138, opened 2019-02-21
fix duplicate method in G::A::A::QueryComplexity

gfx:
AST::QueryComplexity has two `#result` methods so this PR fixes it.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,9 @@ Layout/EndAlignment:
 Lint/UselessAssignment:
   Enabled: true
 
+Lint/DuplicateMethods:
+  Enabled: true
+
 Metrics/ParameterLists:
   Max: 7
   CountKeywordArgs: false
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -26,6 +26,9 @@ Layout/EndAlignment:
 Lint/UselessAssignment:
   Enabled: true
 
+Lint/DuplicateMethods:
+  Enabled: true
```
gfx: This should be helpful anyway.

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -57,7 +57,8 @@ def on_leave_fragment_spread(node, _, visitor)
           visitor.leave_fragment_spread_inline(node)
         end
 
-        def result
+        # @return [Integer]
+        def max_possible_complexity
```
gfx: This is the second, duplicate `#result`. I believe the renaming is needed, anyway.

.rubocop.yml
rmosolgo: 👏 !

rmosolgo:
Thanks for improving this!

gfx:
✌️

PR #2140, opened 2019-02-21
extend GraphQL::Schema::RescueMiddleware#attempt_rescue

masakazutakewaka:
# Summary
Hi there. 
First of all this gem is AWESOME. Really appreciate the hard work you guys put in.

So let's get right into the topic. We are using `rescue_from` in our schema class like this↓ as a safety net to make sure responses are always in the same form.

```ruby
class SomeSchema < GraphQL::Schema
  rescue_from(StandardError) { 'INTERNAL_SERVER_ERROR' }
  ...
end
```

But we are using `extensions` entry to expand error responses in other fields.
So in order to get every error forms together we started to think that we need to be able to expand an error form with `extensions` entry in `rescue_from` method too like this↓

```ruby
class SomeSchema < GraphQL::Schema
  rescue_from(StandardError) { GraphQL::ExecutionError.new('internal server error', extensions: { code: 'INTERNAL_SERVER_ERROR' }) }
  ...
end
```

This PR enables this behavior.

### DIFF
--- a/lib/graphql/schema/rescue_middleware.rb
+++ b/lib/graphql/schema/rescue_middleware.rb
@@ -43,8 +43,13 @@ module GraphQL
       def attempt_rescue(err)
         rescue_table.each { |klass, handler|
           if klass.is_a?(Class) && err.is_a?(klass) && handler
-            message = handler.call(err)
-            return GraphQL::ExecutionError.new(message)
+            result = handler.call(err)
+            case result
+            when String
+              return GraphQL::ExecutionError.new(result)
+            when GraphQL::ExecutionError
+              return result
+            end
           end
         }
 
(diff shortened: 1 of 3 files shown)


rmosolgo:
Hi! Sorry for the slow review on this. I think the change that you posted looks good , but could you add a test for the new behavior? I don't want to accidentally break it in the future 😬

masakazutakewaka:
Hi @rmosolgo. I'm sorry for the lack of tests🙏 I just added tests. Both unit test and e2e test.

rzane:
@rmosolgo, can this PR be merged? It's pretty valuable functionality for dealing with validation errors.

```ruby
  rescue_from ActiveRecord::RecordInvalid do |error|
    raise GraphQL::ExecutionError.new(
      error.message,
      extensions: {
        code: 'VALIDATION_ERROR',
        problems: error.record.errors
      }
    )
  end
```

rmosolgo:
Yes, thanks for the bump here! I must have missed the previous comment 😰 !

PR #2143, opened 2019-02-21
Add simpler, more flexible connections

### LINKED ISSUE #1359: Rethink Pagination

rmosolgo:
Fixes #1359 

__TODO:__

- [x] Make it low-overhead to the user. Somehow tack on arguments after the fact? 
- [x] Add an auto-wrapping system with override
- [x] ArrayConnection & tests 
- [x] ActiveRecordRelationConnection & tests 
- [x] SequelDatasetConnection & tests 
- [x] Mongoid ?? & tests 
- [ ] docs 
- [ ] merge into 1.10-dev


eapache:
> Somehow tack on arguments after the fact?

This is why we built and leveraged `Filter` internally (which got integrated upstream here as `FieldExtension` I believe). By making `Connection < FieldExtension` you get the argument application and a bunch of other useful hooks for free.

spec/graphql/pagination/array_connection_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      # TODO don't require `connection: false`
+      field :items, Item.connection_type, null: false, connection: false do
+        argument :first, Integer, required: false
+        argument :last, Integer, required: false
+        argument :before, String, required: false
+        argument :after, String, required: false
+      end
+
+      def items(**args)
+        GraphQL::Pagination::ArrayConnection.new(ITEMS, context, max_page_size: 6, **args)
```
eapache: by making it an extension applied to the field definition, the resolver method here just becomes `ITEMS` and the extension can do the post-processing in its regular hook instead.

rmosolgo:
Thanks for taking a look @eapache! I took a crack at implementing it the way you mentioned, what do you think?

eapache:
I'm curious why you split `Connection` and `ConnectionExtension`? Also I don't even see where `ConnectionExtension` is getting used in the test schema, but clearly it's being added somewhere...

rmosolgo:
The extension is hooked up here: 

https://github.com/rmosolgo/graphql-ruby/blob/44d0548330cf1c030d81f9be90540110ba0d57ab/lib/graphql/schema/field.rb#L247-L249

What would it look like to put that code _together_? Would the connection object also implement the extension API?

eapache:
Ya, our internal system has a single base `BidirectionalConnection < Filter` (where `Filter` is our old version of `FieldExtension`) which is responsible for:
- adding the various `first` etc. arguments in `apply`
- validating the arguments in `resolve` (e.g. `raise ... unless args[:first] || args[:last]` and others)
- deserializing the cursors in `resolve`
- wrapping the value returned from the resolver into nodes/edges/connection structs and serializing the cursors in `after_resolve`

Of course with subclasses for arrays, AR, etc.

This means that the actual resolver method for a connection doesn't usually need to know it's a connection, it can just return the array or ActiveRecord association directly and the extension can take that and do the right thing. It ends up looking like:
```ruby
field :foo, SomeType.connection, null: false, extension: ArrayConnection

def foo
  [1, 2, 3]
end
```

And there's no more magic in `Field` or elsewhere to treat it specially because it's a connection - the entire thing is handled by the extension.

rmosolgo:
That's awesome, thank you for sharing!  Just one more question at the moment: 

> into nodes/edges/connection structs

IIUC, an interesting thing is that those objects are pretty "dumb", right? The heavy lifting of handling cursors and paginating the list is done _by the extension_, and then those structs are just there to satisfy GraphQL. Is that basically right?

eapache:
Correct. Those classes have a tiny bit of logic so that they inherit certain configuration around visibility, authorization, caching, etc. from the actual underlying type. And the extension lets you override the class that it inflates, so you can provide a class to satisfy ConnectionWithCount or whatever, but all of the heavy lifting in terms of managing the actual cursors and list manipulation is still done by the extension.

rmosolgo:
Ok, next question 😅  ... 

What about unneeded work? For example, I'd like this one: 

```graphql 
items(first: 10) {
  __typename 
}
```

to require fewer database queries than: 

```graphql 
items(first: 10) {
  edges { node { name } } 
  pageInfo { hasNextPage hasPreviousPage }
  totalCount 
}
```

Did you worry about that in your implementation, for example, not calculating `hasPreviousPage` unless it was requested? Or, did you calculate all the values and pass them along, regardless of whether they're queried or not? 

To implement that "laziness", I can imagine a few approaches: 

- Using some kind of lookahead to calculate _only_ the needed values when inflating the various structs 
- Deferring work to calculate some of those values until their corresponding methods are called (this is how it's done in graphql-ruby's current implementation, so that work isn't done unless the methods are called).

Did you find something that worked nicely for you there?

eapache:
We calculate all the values and pass them along, but we've kind of just avoided this problem:
- `hasNextPage` uses the "request one more record than needed from the DB" trick
- `hasPreviousPage` we cheat a little and just check if there was a cursor; so we're sometimes over-eager, but it's a case our clients have to handle anyway because most of the lists we paginate are mutable
- we don't have any implementations of count on our connections; even in places where we do provide counts we general don't support counting with all the crazy filter/search params that our connections support, so we just make it a separate field outside the connection

rmosolgo:
> cheat a little and just check if there was a cursor

I think I'm gonna adopt this as default. It's _good enough_ and like you said, since the list is mutable, there's no guarantee it's gonna be correct _anyways_. Thanks for sharing that again. 

The other catch for me, the reason I want support for returning Connection instances manually, is because we have some lists that need special-case pagination (usually for cursor generation), so I want to distinguish between a "normal" AR::Relation and a "special" one. 

So far, we've created proxy classes to signal that custom connection wrappers should be used, but it seems like one more step than necessary.

bleonard:
@rmosolgo We're about to add our first set of paginated fields so I am investigating the options. When is the target for this (1.10?) to be released? I'm not impatient, I swear! Just trying to optimize the approach.

rmosolgo:
Sorry for the late reply, yes, this is targeted for 1.10 (#2100)

rmosolgo:
I need this to fix a bug in production. I haven't written the docs yet, but I've added them as an explicit TODO in #2100

PR #2145, opened 2019-02-22
Return resolved obj from subscription resolver

mengledowl:
For the sake of public documentation of why we're doing this:

We’ve got some logic where we are wrapping resolvers (using a base field class & `resolve`/`resolve_field`) in order to apply presenter logic to the results of our resolvers.

This has been working well for us, but we’ve been having some issues with it in subscriptions where somehow, even though this gets hit correctly, it eventually tries to call the method names on a non-wrapped version of the object causing things to blow up.We discovered that the solution to this was to return the result of `@inner_proc.call` in `Subscriptions::Instrumentation::SubscriptionRegistrationResolve`.

I'm not sure what kind of repercussions this might have. From what I can tell, it seems like the current functionality is pretty well tested by `SubscriptionSpec` and `SubscriptionsSpec`. For what it's worth, all tests are passing, and when I point our gemfile towards my changes to the gem locally, it works how we would expect (and our test suite passes as well).

Any other thoughts and reactions as to potential consequences are more than welcome.

### DIFF
--- a/lib/graphql/subscriptions/instrumentation.rb
+++ b/lib/graphql/subscriptions/instrumentation.rb
@@ -44,7 +44,9 @@ module GraphQL
 
         # Wrap the proc with subscription registration logic
         def call(obj, args, ctx)
-          @inner_proc.call(obj, args, ctx) if @inner_proc && !@inner_proc.is_a?(GraphQL::Field::Resolve::BuiltInResolve)
+          resolved_obj = if @inner_proc && !@inner_proc.is_a?(GraphQL::Field::Resolve::BuiltInResolve)
+            @inner_proc.call(obj, args, ctx)
+          end
 
           events = ctx.namespace(:subscriptions)[:events]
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hey, thanks for taking a look at this. I agree it'd be nice to make them all work the same. I didn't do this at first because it's a breaking change. If anyone wrote subscription fields without worry about the (unused) return value, they might be surprised by it. But I've heard a few people come looking for this (eg #2294), so we should probably do it! 

I'll take a close look and try to add tests before releasing 1.10.0

lib/graphql/subscriptions/instrumentation.rb
```diff
@@ -59,6 +61,8 @@ def call(obj, args, ctx)
             ctx.skip
```
rmosolgo: I wonder if this should be `resolved_obj` too, see #2294

rmosolgo:
This will be in 1.10.0.pre1!

PR #2154, opened 2019-02-26
Add analytics tag to Datadog tracing

delner:
Datadog offers a trace search and analytics feature, which can make it easy to find and analyze measurements made by the tracing in GraphQL.

This pull request adds an `event_sample_rate` tag, to the span, if configured & available, so the trace will be sampled and indexed for use in search and analytics.

GraphQL users implementing Datadog's `ddtrace` gem version >= 0.20.0 will see the benefit of this new feature, while others using older versions of `ddtrace` will remain backwards compatible.

### DIFF
--- a/guides/queries/tracing.md
+++ b/guides/queries/tracing.md
@@ -132,9 +132,20 @@ To add [Datadog](https://www.datadoghq.com) instrumentation:
 
 ```ruby
 class MySchema < GraphQL::Schema
-  use(GraphQL::Tracing::DataDogTracing)
+  use(GraphQL::Tracing::DataDogTracing, options)
 end
 ```
+
+You may provide `options` as a `Hash` with the following values:
+
+| Key | Description | Default |
+| --- | ----------- | ------- |
+| `analytics_enabled` | Enable analytics for spans. `true` for on, `nil` to defer to Datadog global setting, `false` for off. | `false` |
+| `analytics_sample_rate` | Rate which tracing data should be sampled for Datadog analytics. Must be a float between `0` and `1.0`. | `1.0` |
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hey, thanks for upgrading this integration! 

It seems like this accepts a new option, `event_sample_rate: ...`, is that right? If so, could you update the documentation with a usage example & rationale? The docs are here: 

https://github.com/rmosolgo/graphql-ruby/blob/master/guides/queries/tracing.md#datadog

(I think that would help _me_ understand this feature too!)

delner:
Sure thing, we could also give a little more detail for the other flags, too.

The `event_sample_rate` is very specific to the Datadog product, and specifies at what rate these span measurements should be sampled for analytics, which factors in towards usage calculations. By allowing users to set this rate specifically for GraphQL, they can tune exactly how much of the GraphQL instrumentation data they want to factor into their analytics.

I'll give a basic example of how to configure these settings with GraphQL, and probably link out to the relevant Datadog documentation for the details if users want to know more about how its used with the platform.

delner:
@rmosolgo Updating the documentation for this, let me know if that looks okay.

If it does, I think I'd like to hold off on merging this for just a little bit while I confirm some implementation details with my team; I expect after that, this should be all ready on my end.

rmosolgo:
Yes, looks great, thanks! Let me know when you're ready to 🚢 .

delner:
Okay, I think we need to make some changes (mostly around the naming); I will update this accordingly, then resubmit for review.

delner:
@rmosolgo Okay, I've updated the names accordingly. Should be ready for another review.

rmosolgo:
Cool, thanks for the improvement here and for updating the docs!

PR #2161, opened 2019-02-28
Handle non-node arrays

### LINKED ISSUE #2148: AST Visitor should return nil for all on_<type> hooks
This error only appears while traversing queries that contain a list of something while attempting to inspect the [`parent` parameter](https://github.com/rmosolgo/graphql-ruby/blob/44d0548330cf1c030d81f9be90540110ba0d57ab/lib/graphql/analysis/ast/visitor.rb#L167).

As the `AST::Visitor` doesn't explicitly return nil, it ends up returning a the list of analyzers being run.
https://github.com/rmosolgo/graphql-ruby/blob/44d0548330cf1c030d81f9be90540110ba0d57ab/lib/graphql/analysis/ast/visitor.rb#L236

This causes the `parent` argument to become whatever the second analyzer is, or `nil` in the case of one analyzer, as returning anything besides `nil` is seen as an attempt to rewrite the AST.
https://github.com/rmosolgo/graphql-ruby/blob/df0e6b37bbd9510b73d11ff7fcbe7c3b6a1f7107/lib/graphql/language/visitor.rb#L104

I'll put up a PR in the next few days if no one beats me to it!

chrisseto:
Fixes #2148 

This one took a while to reproduce properly 🤔 

Let me know if you have any questions, comments, or concerns!

### DIFF
--- a/lib/graphql/language/visitor.rb
+++ b/lib/graphql/language/visitor.rb
@@ -168,6 +168,8 @@ module GraphQL
       # Run the hooks for `node`, and if the hooks return a copy of `node`,
       # copy `parent` so that it contains the copy of that node as a child,
       # then return the copies
+      # If a non-array value is returned, consuming functions should ignore
+      # said value
       def on_node_with_modifications(node, parent)
         new_node_and_new_parent = visit_node(node, parent)
         if new_node_and_new_parent.is_a?(Array)
@@ -181,6 +183,10 @@ module GraphQL
             # The user-provided hook requested to remove this node
             new_parent = new_parent && new_parent.delete_child(node)
             return nil, new_parent
+          elsif new_node_and_new_parent.none? { |n| n == nil || n.class < Nodes::AbstractNode }
+            # The user-provided hook returned an array of who-knows-what
+            # return nil here to signify that no changes should be made
+            nil
           else
             new_node_and_new_parent
           end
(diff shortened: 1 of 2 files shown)


lib/graphql/language/visitor.rb
```diff
@@ -181,14 +181,18 @@ def on_node_with_modifications(node, parent)
...
+            return [node, parent]
           else
             new_node_and_new_parent
           end
         else
           # The user-provided hook didn't make any modifications.
           # In fact, the hook might have returned who-knows-what, so
           # ignore the return value and use the original values.
-          new_node_and_new_parent
+          [node, parent]
```
rmosolgo: Can we keep returning original values here? This is a hot path, so I'd like to avoid making the extra array if we can. (Or, do we need it to make this work?)

```suggestion
          new_node_and_new_parent
```

chrisseto: Ah! yes, I didn't realize that everything that called `on_node_with_modifications` was checking if the return value was an array or not.

Would it be more appropriate/performant to return `nil` on line 187 as well?

rmosolgo: Yes, if it works on line 187, that would be great! I know in _some_ spots we have to make sure to return the right thing, but I haven't dug in for a while. The test suite should cover it!

chrisseto:
@rmosolgo Updated! Tests are still passing (locally at least)

rmosolgo:
Thanks again for the fix here!

PR #2162, opened 2019-02-28
Include array indices in error paths

### LINKED ISSUE #2160: Array index excluded in error path
I noticed that when validation fails on arrays, the index of the offending object is not included in the error's `path` key. 
I got halfway through adding it in, realized that it would more or less be a breaking change, and wanted to check if this was intentional or not.

If this would be a desirable feature to add in, should the index be a string or an integer?

Examples pulled from `spec/graphql/static_validation/rules/required_input_object_attributes_are_present_spec.rb`
```rb
offending_query = "missingNestedRequiredInputObjectAttribute: searchDairy(product: [{fatContent: 1.2, order_by: {}}]) { __typename }"

missing_order_by_direction_error = {
  "message"=>"Argument 'direction' on InputObject 'ResourceOrderType' is required. Expected type String!",
  "locations"=>[{"line"=>8, "column"=>100}],
  # I would expect path to actually be:
  # "path"=>["query getCheese", "missingNestedRequiredInputObjectAttribute", "product", "0", "order_by", "direction"],
  # Rather than:
  "path"=>["query getCheese", "missingNestedRequiredInputObjectAttribute", "product", "order_by", "direction"],
  "extensions"=>{
    "code"=>"missingRequiredInputObjectAttribute",
    "argumentName"=>"direction",
    "argumentType"=>"String!",
    "inputObjectType"=>"ResourceOrderType"
  }
}
```

chrisseto:
Fixes #2160 

`is_list = @argument_definitions.last&.type&.list? == true` feels a bit smelly, is there a more idiomatic way to achieve the same result?

### DIFF
--- a/lib/graphql/static_validation/base_visitor.rb
+++ b/lib/graphql/static_validation/base_visitor.rb
@@ -145,6 +145,14 @@ module GraphQL
           @path.pop
         end
 
+        def on_input_object(node, parent)
+          is_list = @argument_definitions.last&.type&.list? == true
+
+          @path.push(parent.children.index(node)) if is_list
+          super
+          @path.pop if is_list
+        end
+
         # @return [GraphQL::BaseType] The current object type
         def type_definition
           @object_types.last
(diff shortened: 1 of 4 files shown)


rmosolgo:
Thanks for the fix here! That's a really nice improvement. I think the way you did it is fine, although I think it would be equivalent to omit `== true`: 

```ruby
is_list = @argument_definitions.last&.type&.list?
```

The values of `false` or `nil` will be treated as `false` anyways.

chrisseto:
Thanks for the tip!

Issue #2157, opened 2019-02-27
Extend authorized? method


g-gaston:
Sometimes a boolean is too limited, I would like to return some extra info when `authorized?` is not successful in a resolver. In my case, at least different codes for missing authentication or forbidden field for the authenticated user (so I can add the appropriate error code in the response error).

If I try to raise a `UnauthorizedFieldError`, that's forwarded to `schema.unauthorized_object` which is not ideal since my handling (for example messages) for unauthorized objects and fields differ. 

I thought about adding another rescue block to the field.resolve method:

```ruby
rescue GraphQL::UnauthorizedFieldError => err
   ctx.schema.unauthorized_field(err)
rescue GraphQL::UnauthorizedError => err
   ctx.schema.unauthorized_object(err)
```

And then extend `UnauthorizedFieldError` to add my error codes. The problem with that approach is I don't have the field in the resolver (or at least I don't find a way to access it) so building a `UnauthorizedFieldError` becomes problematic.

I appreciate any ideas/suggestions. Thanks!

rmosolgo:
Hey, thanks for this suggestion! I think adding another rescue block like that would be a good addition, do you want to try a PR for it? 

As for adding the field, you could add it during the rescue block, for example: 

```ruby
rescue GraphQL::UnauthorizedFieldError => err
   err.field ||= self # Add the field here in case this was raised without it 
   ctx.schema.unauthorized_field(err)
```

Do you think that would work for it?

g-gaston:
Thanks for the quick response. Of course, happy to put it together.

I've been taking a look to see if I can do the same from types. Looks like the code I need is commented out in Schema::Object:

```ruby
def authorized_new(object, context)
  auth_val = begin
    authorized?(object, context)
    # rescue GraphQL::UnauthorizedError => err
    #   context.schema.unauthorized_object(err)
  end
```
Are you ok if I uncomment it?

rmosolgo:
If it still works, it's ok with me 😬 !

PR #2165, opened 2019-03-01
Forwarding field and object unauthorized errors

g-gaston:
This allows to extend the use of `authorized?` in resolvers and objects by forwarding `UnauthorizedFieldError` and `UnauthorizedError` to `unauthorized_field` and `unauthorized_object` respectively.

Coming from #2157


rmosolgo:
Hi! Thanks for making this change. Could you add a test that covers the new behavior? I want to make sure I don't break it in the future 😅

g-gaston:
Yeah, I can do that

g-gaston:
@rmosolgo tests added

rmosolgo:
Looks great, thanks for taking care of this improvement!!

PR #2171, opened 2019-03-06
Prevent argument names from overwriting Arguments methods

gabrielmdeal:
Prevent argument names like the ones below from overwriting existing methods on `GraphQL::Query::Arguments`.
```ruby
field :field, String, null: false do
  argument :to_h, String
  argument :keys, String
end
```

This prevents errors like this:
> NoMethodError: undefined method `any?' for nil:NilClass
>    /home/gabrielx/.rvm/rubies/ruby-2.5.3/lib/ruby/2.5.0/forwardable.rb:227:in `any?'
>    /home/gabrielx/projects/oss/graphql-ruby/lib/graphql/schema/field.rb:573:in `to_ruby_args'

And this:
> NoMethodError: undefined method `key?' for nil:NilClass
>    /home/gabrielx/projects/oss/graphql-ruby/lib/graphql/schema/field.rb:579:in `block in to_ruby_args'
>    /home/gabrielx/projects/oss/graphql-ruby/lib/graphql/schema/field.rb:577:in `each'

### DIFF
--- a/lib/graphql/query/arguments.rb
+++ b/lib/graphql/query/arguments.rb
@@ -19,7 +19,7 @@ module GraphQL
             method_names = [expose_as, expose_as_underscored].uniq
             method_names.each do |method_name|
               # Don't define a helper method if it would override something.
-              if instance_methods.include?(method_name)
+              if instance_methods.include?(method_name.to_sym)
                 warn(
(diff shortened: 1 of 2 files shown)


rmosolgo:
Awesome, thanks for the writeup and the fix here, that looks great!

PR #2174, opened 2019-03-08
Removed documentation for field instrumentation.

abhaynikam:
Ref: https://github.com/rmosolgo/graphql-ruby/issues/2100#issuecomment-471007190


guides/fields/instrumentation.md
```diff
@@ -1,50 +0,0 @@
----
-layout: guide
-doc_stub: false
-search: true
-section: Fields
-title: Instrumentation
-desc: Wrap and modify resolution behavior at schema build-time
-index: 3
```
abhaynikam: @rmosolgo : Do we need to update index of other documentation after removing this?

rmosolgo: No, we don't need to update it. It just `.sort {...}`s by this, so it's ok to have a gap!

rmosolgo:
Thanks! This change will go live next time i push the site.

PR #2175, opened 2019-03-08
Removed blue notice bar which prompts about New class based API

abhaynikam:
Ref: #2100

### DIFF
--- a/guides/_layouts/guide.html
+++ b/guides/_layouts/guide.html
@@ -18,17 +18,6 @@ layout: default
     </p>
   </div>
 {% endif %}
-{% if page.class_based_api %}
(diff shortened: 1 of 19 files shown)


rmosolgo:
Thanks, this is a great start! While you're at it, would you mind removing `class_based_api: true` from the frontmatter yaml, too, since it's now unused?

abhaynikam:
@rmosolgo : Thanks for the comment. I forgot to check if it was used or not. I have fixed it here. Please have a look.

rmosolgo:
🎉 Awesome, thanks for taking care of this! It will go live next time I publish the docs.

PR #2176, opened 2019-03-09
Fix ActionCableLink when there are errors but no data

staugaard:
We ran into an issue where Apollo is not notified when sending bad queries that return errors but no data.


rmosolgo:
Awesome, thanks for the fix here!

PR #2177, opened 2019-03-09
Update testing docs

rmosolgo:
I want to split testing into its own section and add some detail about different approaches. 

- [x] Add conceptual overview
- [x] Add structural testing section 
- [x] Add runtime testing section 
- [x] Remove old doc


abhaynikam:
Woah. This looks great @rmosolgo 🥇 . I was thinking about starting to write about testing docs but I would not have written it so well. 

Let me know if I can work on adding something here. Mostly examples for testing in rspec and minitest.

> Remove old doc
This I can definitely complete this one 😄

rmosolgo:
Seems like travis is having problems.

Issue #2182, opened 2019-03-12
nil option object for Field::Extension(s) instead of empty hash


ngouy:
Hello

Wanted to know if it's on purpose, and if yes why :

When creating and using a FieldExtension (GraphQL::Schema::FieldExtension) and passing no options, in the `after_resolve` method, `option` object is not an empty hash but is at `nil` instead

So basically if you want an optional option, you have to check the non nillability of option, then check if the option is present then take the value or use a default one, instead of just make something like 
```rb
    def after_resolve(value:, **_)
       value.round(options.fetch(:ndigits) { 0 }) # but have to do value.round(options&.fetch(:ndigits) { 0 } || 0)
    end
```

rmosolgo:
Using an empty hash as default sounds good to me! Do you want to try making that change?

ngouy:
👍 on it
I will try do work this out before next monday

ngouy:
does an edit on the initializer seems the better way to achieve this ?
```
def initialize(field:, options:)
        @field = field
        @options = options || {}
        apply
        freeze
      end
```

rmosolgo:
Any implementation is ok with me, as long as it comes with tests for the new behavior 😉

ngouy:
👌 of course

rmosolgo:
Fixed by #2189

PR #2189, opened 2019-03-14
provide empty hash as default option for field resolvers

ngouy:
with updated tests, though I am not sure it's the good way to test it

related ticket : https://github.com/rmosolgo/graphql-ruby/issues/2182


rmosolgo:
That test looks great to me, thanks for the improvement here!

PR #2192, opened 2019-03-18
Doc: fix ActionCableSubscriptions channel example

such:
A simple typo that leads to memory leaks!


rmosolgo:
Thanks for the fix!

PR #2193, opened 2019-03-18
Improve wording for override camelize

ka8725:
It's not clear what current docs tell.


rmosolgo:
Thanks for the improvement!

PR #2198, opened 2019-03-20
Interpreter handling of exceptions raised during argument preparation

jturkel:
This PR makes the handling of GraphQL::ExecutionErrors thrown by argument `prepare` callbacks consistent between the old runtime and the interpreter runtime. The previous behavior was:

| Error | Old Runtime | Interpreter Runtime |
| --- | --- | --- |
| Field argument prepare raises exception | partial query failure | complete query failure |
| Literal input object argument prepare raises exception | partial query failure | complete query failure |
| Variable input object argument prepare raises exception | complete query failure | `GraphQL::Schema#execute` raises an exception |

This behavior can be seen in this [gist](https://gist.github.com/jturkel/d95364ffd9d2a56827168bb155dfdad1) which passes when using the old runtime but not when using the interpreter runtime.


holic:
I just ran into this, so I'm happy to see a patch is already up!

rmosolgo:
Awesome, thanks for this fix!

Issue #2167, opened 2019-03-05
Interpreter Support for Execution Timeouts


jturkel:
The 1.9 `GraphQL::Execution::Interpreter` runtime doesn't support middleware which means the `GraphQL::Schema::TimeoutMiddleware` won't work. Is there a recommendation migration path? The best I could come up with was using a field extension resolve hook but that's a bit painful to use throughout my schema.

jturkel:
I was able to get this working with the following [tracer](https://graphql-ruby.org/queries/tracing.html) (word of warning - I haven't yet tested this with multiplexing and lazy field execution so that might not be implemented properly):

```ruby
class TimeoutTracer
  TIMEOUT_STATE_KEY = :graphql_execution_timeout_state

  attr_reader :timeout, :timeout_callback

  def self.use(schema, **options, &block)
    tracer = new(**options, &block)
    schema.tracer(tracer)
  end

  def initialize(timeout:, &timeout_callback)
    @timeout = timeout
    @timeout_callback = timeout_callback || Proc.new {}
  end

  def trace(key, data)
    case key
    when 'execute_query'
      with_timeout_context([data.fetch(:query)]) { yield }
    when 'execute_multiplex'
      with_timeout_context(data.fetch(:multiplex).queries) { yield }
    when 'execute_field', 'execute_field_lazy'
      if Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond) > Thread.current[TIMEOUT_STATE_KEY][:at]
        timeout_callback.call(queries: Thread.current[TIMEOUT_STATE_KEY][:queries])
        GraphQL::ExecutionError.new('Query timeout')
      else
        yield
      end
    else
      yield
    end
  end

  def with_timeout_context(queries)
    # The query context isn't available in execute_field callbacks when using the interpreter so we have to
    # store the timeout in a thread local
    if Thread.current[TIMEOUT_STATE_KEY].nil?
      Thread.current[TIMEOUT_STATE_KEY] = {
        at: Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond) + timeout * 1000,
        queries: queries
      }
      begin
        yield
      ensure
        Thread.current[TIMEOUT_STATE_KEY] = nil
      end
    else
      yield
    end
  end
end
```
[...]

PR #2205, opened 2019-03-27
Include query in interpreter tracing events for execute_field and execute_field_lazy

jturkel:
This adds the current `GraphQL::Query` to tracing events for `execute_field` and `execute_field_lazy` when running with the interpreter which is useful for accessing the query, query context, etc. (see #2167 for one example use case). With the old runtime this information is available via the `context` key which has a `GraphQL::Query::Context::FieldResolutionContext` instance.


rmosolgo:
Awesome, thanks for making this improvement! I bet it will come in handy for other things, too.

PR #2209, opened 2019-03-29
Update Lookahead docs: remove fragment note

swalkinshaw:
Support was added in https://github.com/rmosolgo/graphql-ruby/pull/2068


rmosolgo:
🍻 Oops, nothing like stale docs.

PR #2216, opened 2019-04-03
Autogenerate base_field.rb

iamwhil:
In the documentation for Mutation Classes in 1.9.3 it specifies that you should build a Base Mutation.

https://graphql-ruby.org/mutations/mutation_classes.html#example-mutation-class

In the example mutation class it specifies:

field_class Types::BaseField

My install of graphql-ruby did not create a Types::BaseField at types/base_field.rb.

Therefore it throws an error when attempting to load graphiql:
```
uninitialized constant Types::BaseField
Did you mean?  Types::BaseInterface
```
The answer is no, no I did not.

The fix is to add the base_field.rb in the types/ directory.  This should be auto generated.  This patch makes it so.
```
# types/base_field.rb
module Types
    def resolve_field(obj, args, ctx)
      resolve(obj, args, ctx)
    end
  end
end
```

### DIFF
--- a/lib/graphql/version.rb
+++ b/lib/graphql/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module GraphQL
-  VERSION = "1.9.3"
+  VERSION = "1.9.3.1"
 end
--- a/lib/graphql/version.rb
+++ b/lib/graphql/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module GraphQL
-  VERSION = "1.9.3.1"
+  VERSION = "1.9.3"
 end
--- a/lib/generators/graphql/install_generator.rb
+++ b/lib/generators/graphql/install_generator.rb
@@ -13,6 +13,7 @@ module Graphql
     #   - graphql/
     #     - resolvers/
     #     - types/
+    #       - base_field.rb
     #       - base_enum.rb
     #       - base_input_object.rb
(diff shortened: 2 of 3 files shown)


rmosolgo:
Hey, thanks for taking a look at this! I love the idea of generating BaseField. Maybe we could generate BaseArgument too. 

This diff seems a bit off -- did you mean to open this PR against `master`? (It's currently open against `1.9-dev`, which is a dead-end branch.)

iamwhil:
Hey Robert!

  Sorry I posted the issue in issues and then was like "I can just fix this
:)".  I didn't look at your PR rules but just made an assumption that 1.9
jazz was going into 1.9-dev. I was in a bit of a rush trying to hit a
deadline (today).  That's my bad sorry about that!

  I'm not sure what base argument should look like, but yea!  If you have
what that should look like that I'll add it and re-aim that PR at master :).

Thanks,
-Whli

lib/graphql/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module GraphQL
-  VERSION = "1.9.3"
+  VERSION = "1.9.3.1"
```
rmosolgo:
```suggestion
  VERSION = "1.9.3"
```

rmosolgo:
I've cleaned up the diff a bit, I'll merge it when the tests pass!

PR #2219, opened 2019-04-03
Fix FieldUsage analyzer method args

swalkinshaw:
Minor thing since the arguments weren't used but still 🎉


rmosolgo:
Oops, nice find!

PR #2220, opened 2019-04-04
Interpreter support for timeouts

### LINKED ISSUE #2167: Interpreter Support for Execution Timeouts

jturkel:
This PR introduces a `GraphQL::Schema::Timeout` plugin that uses tracing callbacks to implement timeouts in a way that works with both the old runtime and the new interpreter runtime. The plugin is configured with a timeout and an optional timeout callback:

```ruby
  use GraphQL::Schema::Timeout, max_seconds: 2, timeout_callback: ->(err, query) do
    Rails.logger.info("GraphQL Timeout: #{query.query_string}")
  end
end
```
The timeout callback is only invoked for the first field in a query that exceeds the timeout to prevent timeouts from being reported multiple times to services like Bugsnag (which is a deviation from the behavior of the timeout middleware). 

The old timeout middleware is retained for backwards compatibility but there's no reason for new clients to use it.

A few additional notes:
* This PR includes the changes from #2205 to avoid using thread locals for storing query execution timeout state.
* The timeout plugin tests were shamelessly copied from the timeout middleware tests but updated to use the new class based API.

Fixes #2167


rmosolgo:
This looks great! Thanks again for sharing your solution on this. 

I have one suggestion: I'm trying to work _away_ from Procs, since they're unfamiliar to so many Ruby developers. I have an alternative suggestion for the Timeout plugin: how about making it a _class_ which can be customized by overriding an instance method. For example: 

```ruby 
class CustomTimeout < GraphQL::Schema::Timeout 
  def handle_timeout(err, query) 
    # Do something here 
  end 
end 

  use CustomTimeout, timeout: 2 
end 
```

What do you think of that? If you want to take a try at it, go ahead, otherwise I'll take a look when I get a chance.

jturkel:
I like it. It makes timeout handlers easier to unit test and supports any code reuse through standard Ruby inheritance mechanism. Pushed the change. Let me know what you think.

jturkel:
@rmosolgo - Any additional feedback on this PR?

rmosolgo:
Awesome! Sorry for the slow turnaround on this. It looks great.

PR #2224, opened 2019-04-05
Use `default_graphql_name` for BaseEdge edges

### LINKED ISSUE #2221: `graphql_name` on `...Relay::BaseEdge` causes duplicates
I was creating a custom edge with `GraphQL::Types::Relay::BaseEdge` and I specified `node_type(UserType)` and got the error `Duplicate type definition found for name 'UserEdge'`

I think maybe this line isn't doing the right thing: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/relay/base_edge.rb#L36

Maybe the `node_type`'s `graphql_name` could be used sometimes, but I would think that the class name of the edge class would be better.  Would that just be `self.name` in this case?

cheerfulstoic:
Not 100% sure if this is the right solution, but the tests seem to pass locally.  Fixes #2221 (tested fork against my app)

### DIFF
--- a/lib/graphql/types/relay/base_edge.rb
+++ b/lib/graphql/types/relay/base_edge.rb
@@ -31,9 +31,6 @@ module GraphQL
           def node_type(node_type = nil)
             if node_type
               @node_type = node_type
-              wrapped_type_name = node_type.graphql_name
-              # Set this to be named like the node type, but suffixed with `Edge`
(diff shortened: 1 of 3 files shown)


rmosolgo:
Makes sense to me, thanks! It _seems_ like it should just work. Now we'll see 😬

sss545 (commented):
dsf

PR #2226, opened 2019-04-05
Update AST visitor argument definition methods

### LINKED ISSUE #2217: AST analyzer bug with argument_definition
Testing out `on_leave_argument` and `on_enter_argument` I assumed that `visitor.argument_definition` should work and be the argument definition of the argument being visited however it was always `nil`. Looking at the source:  https://github.com/rmosolgo/graphql-ruby/blob/3e2c57b28cafd2d9555bb9e131399526002e43a7/lib/graphql/analysis/ast/visitor.rb#L221-L226

I tested this out and `@argument_definitions[-1]` seemed to give me what I'd expected. I didn't look into this enough yet to figure out if that's actually the proper solution. Just wanted to raise the issue for now at least.

cc @xuorig

swalkinshaw:
Closes #2217 

Changes `argument_definition` to return the current argument definition and adds `previous_argument_definition`.

@rmosolgo I'm guessing this is what you had in mind?

cc @xuorig

### DIFF
--- a/lib/graphql/analysis/ast/visitor.rb
+++ b/lib/graphql/analysis/ast/visitor.rb
@@ -220,8 +220,11 @@ module GraphQL
 
         # @return [GraphQL::Argument, nil] The most-recently-entered GraphQL::Argument, if currently inside one
         def argument_definition
-          # Don't get the _last_ one because that's the current one.
-          # Get the second-to-last one, which is the parent of the current one.
+          @argument_definitions.last
+        end
+
+        # @return [GraphQL::Argument, nil] The previous GraphQL argument
+        def previous_argument_definition
           @argument_definitions[-2]
         end
 
--- a/spec/graphql/analysis/ast_spec.rb
+++ b/spec/graphql/analysis/ast_spec.rb
@@ -71,6 +71,17 @@ describe GraphQL::Analysis::AST do
     end
   end
 
+  class AstArguments < GraphQL::Analysis::AST::Analyzer
+    def on_enter_argument(node, parent, visitor)
+      @argument = visitor.argument_definition
+      @previous_argument = visitor.previous_argument_definition
+    end
+
+    def result
+      [@argument, @previous_argument]
+    end
+  end
+
   describe "using the AST analysis engine" do
     let(:schema) do
       query_type = Class.new(GraphQL::Schema::Object) do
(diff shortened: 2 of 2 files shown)


swalkinshaw:
This is a breaking change technically too

spec/graphql/analysis/ast_spec.rb
```diff
@@ -71,6 +71,17 @@ def result
     end
   end
 
+  class AstArguments < GraphQL::Analysis::AST::Analyzer
+    def on_enter_argument(node, parent, visitor)
+      @argument = visitor.argument_definition
```
rmosolgo: Is `@argument` here the same as `node`? 

If so, that's basically why it was previously implemented this way. `visitor.argument_definition` was not the _current_ arg defn, but rather, the argument definition which _owns_ the value we're looking at right now. 

If `@argument` _is_ `node`, then it's not clear to me why we need the new method. 

But maybe it's something else?

swalkinshaw: Do you mean same argument in the schema context? If so, yes it's the same argument. Of course `node` is a `GraphQL::Language::Nodes::Argument` and `@argument` is a `GraphQL::Argument`.

It's equivalent to a field with `node` and `visitor.field_definition` now.

I'm trying to refresh my memory on this... but basically without this addition `visitor.argument_definition` wouldn't be the same argument as `node` but its "owner" (if it exists) and you couldn't even easily get the current argument definition.

rmosolgo: ohhh right, `node` is a Node and `argument_definition` is a definition. Also refreshing my memory apparently 😆 . Thanks for clarifying.

lib/graphql/analysis/ast/visitor.rb
```diff
@@ -222,6 +222,11 @@ def directive_definition
         def argument_definition
           # Don't get the _last_ one because that's the current one.
           # Get the second-to-last one, which is the parent of the current one.
```
xuorig: Does this comment need to change / be removed?

swalkinshaw: Yep 👀 removed

rmosolgo:
I think I'll call this a "breaking fix." Thanks for this improvement!

PR #2227, opened 2019-04-05
JSON scalar type

### LINKED ISSUE #146: json scalar type

jturkel:
Questions about JSON scalars come up frequently in the graphql-ruby Slack channel so this PR adds native support for them rather than requiring custom scalars. The real work was already done by #1398. Really fixes #146 this time.


ajsharp:
Successfully using this in my codebase now. Thanks @jturkel. Agree this would be great to have in the core library.

rmosolgo:
Thanks, that will make it a lot easier for folks who use this!

PR #2233, opened 2019-04-08
[Documentation] Naming Fix - DataDog Tracing Docs

MacroMackie:
In `GraphQL::Tracing::DataDogTracing`, we fetch the option `:service`
https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/data_dog_tracing.rb#L42

However, the docs currently [incorrectly] specify that the param is named `:service_name`.
I've updated the docs to be accurate!

Before:
https://github.com/rmosolgo/graphql-ruby/blob/master/guides/queries/tracing.md#datadog

After:
https://github.com/rmosolgo/graphql-ruby/blob/564b4f51046118a13571b4f3c9dbf0f2241b7486/guides/queries/tracing.md#datadog

(note that the example in the getting started guide is correct. https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#graphql)


rmosolgo:
Oops, thanks for the fix!

PR #2236, opened 2019-04-09
Interpreter shouldn't recursively unwrap RelayClassicMutation inputs

jturkel:
The interpreter runtime is recursively unwrapping inputs to `GraphQL::Schema::RelayClassicMutation` which is inconsistent with the old runtime. I'm assuming this was unintentional (at least it wasn't documented in the list of incompatibilities 😄) so this PR updates`GraphQL::Schema::RelayClassicMutation`to only unwrap the the top level of the input when using the interpreter.


rmosolgo:
Oops! Thanks for the fix here, yes, probably just an oversight on my part.

PR #2238, opened 2019-04-10
raise ParseError for GraphQL.parse(nil)

gfx:
I use `GraphQL.parse()` to prepare compiled GraphQL documents, and found that it raises NoMethoderror for `nil` by `query_string.unpack("c*")`.

I think it should raise `GraphQL::ParseError` for such cases. What do you think of it?


rmosolgo:
Good, thanks! I agree that's a much more appropriate error.

PR #2242, opened 2019-04-11
Add more context to input validation error message

srt32:
When a GraphQL mutation gets invalid input, we raise an error with a message like

```
Variable input of type <InputType>! was provided invalid value
```

Reproducing these issues is not always obvious because which field was
invalid is not clear.

This change updates the error message to include the path to the invalid
field when it's present. The message now looks like:

```
Variable input of type <InputType>! was provided invalid value for field foo
```

Or

```
Variable input of type <InputType>! was provided invalid value for field 0.foo
```

if the field was in an array field.

There's likely some refactorings to be done here so I'm very open to
feedback. Thank you!

### DIFF
--- a/lib/graphql/query/variable_validation_error.rb
+++ b/lib/graphql/query/variable_validation_error.rb
@@ -9,6 +9,11 @@ module GraphQL
         @validation_result = validation_result
 
         msg = "Variable #{variable_ast.name} of type #{type} was provided invalid value"
+
+        if problem_fields.any?
+          msg += " for #{problem_fields.join(", ")}"
+        end
+
         super(msg)
         self.ast_node = variable_ast
       end
@@ -24,6 +29,15 @@ module GraphQL
             newValue
           end
         end
+      end
+
+      private
+
+      def problem_fields
+        @problem_fields ||= @validation_result
+          .problems
+          .reject { |problem| problem["path"].empty? }
+          .map { |problem| problem["path"].join(".") }
       end
     end
   end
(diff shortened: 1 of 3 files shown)


srt32:
cc/ @rmosolgo your input would be ✨

lib/graphql/query/variable_validation_error.rb
```diff
@@ -25,6 +30,15 @@ def to_h
...
         end
       end
+
+      private
+
+      def problem_fields
+        @problem_fields ||= @validation_result
+          .problems
+          .reject { |problem| problem["path"].empty? }
+          .map { |problem| problem["path"].join(".") }
```
krider2010: I think it'd be super useful to include the `explanation` here. So for the modified test cases below, you'd end up with messages of the form:

```"message" => "Variable input of type ReplaceValuesInput! was provided invalid value for values (Expected value to not be null)"```
and
```"message" => "Variable input of type [DairyProductInput] was provided invalid value for 0.foo (Field is not defined on DairyProductInput), 0.source (Expected value to not be null)",```

I know this is duping information from the problems hash, but I feel just the field name alone, especially in an array case of multiple issues, not knowing why it was invalid could still hinder debugging.

What do you both think to this?

Comment (srt32) [reply lib/graphql/query/variable_validation_error.rb]: My only reservation is the messages can be quite long now but I'm into it. Updated in 8a8e189be.

**Comment by srt32** [comment]
Looks like the build is failing because of recent google chrome cert expirations

```
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install google-chrome-stable" failed and exited with 100 during .
```

rmosolgo:
This is awesome! Thanks for taking care of this improvement 🍻

PR #2244, opened 2019-04-11
Reduce memory and allocations

i-tsvetkov:
- reduce the hash allocations by using positional arguments instead of keyword arguments

The object allocations were measured with the allocation_tracer gem and ruby 2.5.0.

This reduces the memory allocated by our application by 100 MB when the response contains a large number of objects (16K in our case).

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -44,11 +44,11 @@ module GraphQL
         # @api private
         def spawn_child(key:, irep_node:, object:)
           FieldResolutionContext.new(
-            context: @context,
-            parent: self,
-            object: object,
-            key: key,
-            irep_node: irep_node,
+            @context,
+            key,
+            irep_node,
+            self,
+            object
           )
         end
 
@@ -203,7 +203,7 @@ module GraphQL
         attr_reader :irep_node, :field, :parent_type, :query, :schema, :parent, :key, :type
         alias :selection :irep_node
 
-        def initialize(context:, key:, irep_node:, parent:, object:)
+        def initialize(context, key, irep_node, parent, object)
           @context = context
           @key = key
           @parent = parent


rmosolgo:
Nice! I always attributed the memory issues to the Context object itself, and never considered those  hashes. 

Actually, I thought some later Ruby version _elimited_ the allocated hashes for kwargs 🤔  

Could you share the script you used to identify this issue? (I don't doubt your findings, but I'd like to learn how to find this myself!) 

I wrote a small script but didn't see the allocated hashes: 

```ruby 
# allocation_test.rb
require "allocation_tracer"
require 'pp'

def method_1(a:, b:)
  a + b
end

def method_2(a, b)
  a + b
end

pp ObjectSpace::AllocationTracer.trace{
  method_1(a: "a", b: "b")
}

pp ObjectSpace::AllocationTracer.trace{
  method_2("a", "b")
}
```

```
~/code/graphql-ruby $ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin16]
~/code/graphql-ruby $ ruby allocation_test.rb
{["allocation_test.rb", 13]=>[2, 0, 0, 0, 0, 0],
 ["allocation_test.rb", 5]=>[1, 0, 0, 0, 0, 0]}
{["allocation_test.rb", 17]=>[2, 0, 0, 0, 0, 0],
 ["allocation_test.rb", 9]=>[1, 0, 0, 0, 0, 0]}
```

i-tsvetkov:
Hi and thanks for the quick response!

The way I measured the memory usage was something like this:
https://gist.github.com/i-tsvetkov/78e0af07be27d9791fd54be3e6fc6c45
I also attached a verbose output of `time` measured with:
```
 $ /usr/bin/time -v ruby allocation_test.rb
```

The file and line with the most allocations:
```
  [
    [
      "/home/ito/.rvm/gems/ruby-2.5.1/gems/graphql-1.9.4/lib/graphql/query/context.rb",
      46,
      "T_HASH"
    ],
    [
      2426836,
      11,
      2412228,
      0,
      10,
      690365376
    ]
  ],
```
I have sorted the trace by the number of allocated objects.
It can be experimented with the sort order there is explanation of the numbers in the allocation_tracer readme.
Another great read is https://gist.github.com/ko1/40110a3d951c19ed6979

Ruby verison:
```
 $ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
```

With your example I also could not see the hash allocations whitch is _strange_ (I asume there is some special cases where the hash allocation is not nessesary, but I don't know for sure).

rmosolgo:
Thanks for this improvement!

Issue #1217, opened 2018-01-12
connection_type and nulls


sporto:
When I use `connection_type` it creates `edges` and `node` as nullable fields. 

```
          {
            "name": "edges",
            "description": "A list of edges.",
            "args": [

            ],
            "type": {
              "kind": "LIST",
              "name": null,
              "ofType": {
                "kind": "OBJECT",
                "name": "AwsAccountOrOnboardingEdge",
                "ofType": null
              }
            },
            ...
          },
```

Does this make sense? I don't see why these should be null. Maybe the edges does, not sure.
At least shouldn't `node` be `NOT_NULLABLE`?

rmosolgo:
It looks like `edges` is nullable according to the spec: https://facebook.github.io/relay/graphql/connections.htm#sec-Edges

But `node` may be a non-null type: https://facebook.github.io/relay/graphql/connections.htm#sec-Node

I'm open to supporting non-null `node` fields, but I don't think we should change the default behavior.

eapache:
FWIW the vast majority of our connections use non-null nodes. In fact any field we create that is a list defaults to the inner wrapped element being non-null unless otherwise specified.

thefliik:
While it certainly makes sense that a connection object, in general, could have nullable edges/nodes, for the `connection_type` helper method, specifically, does it make sense? My understanding of the `connection_type` method, is that you provide it with a list of nodes and then it creates the edges (and connection object) around those nodes. In this case, an edge will always have a node and the `edges` field itself will always return a list (which may be empty).

sporto:
For a bit more context. We are using Elm for our front end. We started using a library that generates types from our graphql schema. So at the moment for a connection we get something like:

```
List (Maybe { node = Maybe Thing} )
        ^ edge                 ^ node
```
Having to deal with these nullables in the FE is harder than it should be. We would like to be able to specify a `connect_type` where nothing is null. So we can end up with:

```
List { node = Thing }
```

Thanks

petrbela:
This is not an issue just with Elm but also TypeScript/Flow or any other type system.

There's a similar discussion for [graphql-relay-js](https://github.com/graphql/graphql-relay-js/issues/185). Specifically, it reiterates that [edges are always nullable while nodes can be non-null](https://github.com/graphql/graphql-relay-js/issues/185#issuecomment-387923487), and that [nullability is required in case of an error](https://github.com/graphql/graphql-relay-js/issues/185#issuecomment-387925025).

[...]

Issue #1683, opened 2018-07-18
[1.8] GraphQL::Types::Relay::BaseConnection adds nodes field by default


y310:
I made own custom connection class by following this instruction. http://graphql-ruby.org/relay/connections.html

It looks like

```ruby
class Types::PostConnectionType < GraphQL::Types::Relay::BaseConnection
  edge_type(Types::PostEdge)

  field :total_count, Int, null: false

  def total_count
     # calculate count
  end
end
```

Then , I realized that this custom connection has `nodes` field. It seems that it's added by default by https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/relay/base_connection.rb#L64-L66

But,  I think the original connection type made that optional https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/relay/connection_type.rb#L29-L31

Is this an intentional behavior or bug? If it's intentional, is there a chance to disable that by default or add an option to disable?

y310:
Fix for this issue #1693 is merged now. Closing, thanks 👍

PR #2249, opened 2019-04-19
Add null option to node_type of Relay::BaseEdge

tilfin:
As discussed in the issue #1217, nullable requires extra code in typed languages, so it is important to have a way to avoid it. I submitted this PR because our private project ran into that problem.

### DIFF
--- a/spec/graphql/schema/introspection_system_spec.rb
+++ b/spec/graphql/schema/introspection_system_spec.rb
@@ -52,5 +52,14 @@ describe GraphQL::Schema::IntrospectionSystem do
       ensembles_field = query_type["fields"].find { |f| f["name"] == "ensembles" }
       assert_equal [], ensembles_field["args"]
     end
+
+    it "runs the introspection query and the result contains a edge field that has non-nullable node" do
+      res = NonNullableDummy::Schema.execute(GraphQL::Introspection::INTROSPECTION_QUERY)
+      assert res
+      edge_type = res["data"]["__schema"]["types"].find { |t| t["name"] == "NonNullableNodeEdge" }
+      node_field = edge_type["fields"].find { |f| f["name"] == "node" }
+      assert_equal "NON_NULL", node_field["type"]["kind"]
+      assert_equal "NonNullableNode", node_field["type"]["ofType"]["name"]
+    end
   end
 end
--- a/lib/graphql/types/relay/base_edge.rb
+++ b/lib/graphql/types/relay/base_edge.rb
@@ -28,14 +28,15 @@ module GraphQL
           # Get or set the Object type that this edge wraps.
           #
           # @param node_type [Class] A `Schema::Object` subclass
-          def node_type(node_type = nil)
+          # @param null [Boolean]
+          def node_type(node_type = nil, null: true)
             if node_type
               @node_type = node_type
               wrapped_type_name = node_type.graphql_name
               # Set this to be named like the node type, but suffixed with `Edge`
               graphql_name("#{wrapped_type_name}Edge")
               # Add a default `node` field
-              field :node, node_type, null: true, description: "The item at the end of the edge."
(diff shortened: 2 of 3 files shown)


spec/graphql/schema/introspection_system_spec.rb
```diff
@@ -52,5 +52,14 @@
...
     end
+
+    it "runs the introspection query and the result contains a edge field that has non-nullable node" do
+      res = NonNullableDummy::Schema.execute(GraphQL::Introspection::INTROSPECTION_QUERY)
+      assert res
+      edge_type = res["data"]["__schema"]["types"].find { |t| t["name"] == "NonNullableNodeEdge" }
+      node_field = edge_type["fields"].find { |f| f["name"] == "node" }
+      assert_equal "NON_NULL", node_field["type"]["kind"]
+      assert_equal "NonNullableNode", node_field["type"]["ofType"]["name"]
+    end
```
tilfin: I thought writing tests here would not be the best, but I did this because it is unprecedented.

rmosolgo:
Looks good, thanks for this improvement! I agree the test is in a funny spot, I'll shuffle it around a bit after merging it to master.

OzzieOrca:
Any chance we could get a release for this? It's really annoying to deal with in a TypeScript client. Thanks for your work in fixing it 😃 

Currently the type generated by the Apollo codegen is:
```ts
export interface MyQuery_MyType {
  __typename: "MyTypeConnection";
  /**
   * A list of nodes.
   */
  nodes: (MyQuery_MyType_nodes | null)[] | null;
  /**
   * Information to aid in pagination.
   */
  pageInfo: MyQuery_MyType_pageInfo;
}
```
both the `nodes` array and items in the `nodes` array require null checks when accessing data there and I can't think of a reason for them to be ever be null.

OzzieOrca:
Does this support non-null nodes when `nodes_field: true`? I noticed the test example had it set to `false`. We can't seem to get it to work.

We can get the `node` inside `MyTypeEdge` to be `MyType!` but can't seem to change the `nodes` array out in the connection.

Ideally we'd like the schema to look like:
```graphql
type MyTypeConnection {
  edges: [MyTypeEdge!]!
  nodes: [MyType!]!
  pageInfo: PageInfo!
}
```
but even just
```graphql
type MyTypeConnection {
  edges: [MyTypeEdge]
  nodes: [MyType!]
  pageInfo: PageInfo!
}
```
would be better. I'm not sure where edges become useful (pageInfo seems to have all the cursors I need when using the Apollo Client) so at this point I'm mostly concerned with making `nodes` non-null.

tilfin:
@OzzieOrca This does not support `nodes` of `Connection` because `nodes` is not adaptive for the Relay specification. I think that `edges -> node` should be abolished. https://github.com/rmosolgo/graphql-ruby/issues/1683

OzzieOrca:
Thanks for the reply :) So you always go through edges to get your data? I'm still learning here. Seems a little verbose but I guess it's more flexible. Is there a way to make the edges array and edges themselves non-nullable then? Just seems like every way of accessing this currently requires a lot of extra null checks in TypeScript. I think we went with Relay connections for the easy pagination info and `endCursor` but maybe it's too much for our use case. Do you have any suggestions?

https://stackoverflow.com/questions/33323894/why-are-edges-required-in-a-relay-graphql-connection#comment74956395_34193320 is a good argument for edges if that functionality is needed.

I can move this conversation somewhere else if it's more appropriate.

tilfin:
This PR is just one step to reduce type null checking. I think it is better to discuss on #1217 but It should be discussed about the Relay specification(https://facebook.github.io/relay/graphql/connections.htm#sec-Edges) whether edges should be non-null.

OzzieOrca:
Thanks :) Seems like a good discussion there.

Issue #2223, opened 2019-04-05
Bug when using functions and .connection_type


mull:
Possibly, but I don't think, a duplicate of #2045 

I've set up a repo to reproduce the bug: https://github.com/mull/graphql-bug-reproduction

```
bundle install
bundle exec ruby test.rb
```

I've been getting rid of all of our GraphQL::Function definitions in favor of resolvers. I was hoping to keep the two generic functions we have (find 1 record, find many records) though, since they seem to work pretty well.

It seems like the library cannot figure out the return type when I pass `SomeType.connection_type`, resulting in:

```
/Users/mull/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/graphql-1.9.3/lib/graphql/schema/member/build_type.rb:80:in `parse_type': Failed to build return type for RootQuery.currencyList from nil: Unexpected type input:  (NilClass) (ArgumentError)
```

Am I perhaps doing something wrong or is this a known bug that's not set to be fixed? I upgraded from 1.8.8 to 1.9.3 and that's when this broke, if that helps.

Happy to provide further assistance if I can!

rmosolgo:
Sorry for the late response here. I don't have the bandwidth to continue supporting GraphQL::Function. If you want to open a new issue with your code, I'd be happy to advise on a migration to GraphQL::Schema::Resolver.

PR #2255, opened 2019-04-24
GraphQL::Schema::Field#scoped? checking presence of return type

mengqing:
This PR fixes an issue where `GraphQL::Function` stopped working for connections after v1.9.0, also added additional tests to cover `GraphQL::Function`

reference https://github.com/rmosolgo/graphql-ruby/issues/2223


rmosolgo:
Thanks for the fix on this one!

PR #2256, opened 2019-04-26
Add the ability to pass in context on Schema.validate

### LINKED ISSUE #2231: Incorrect authorization documentation for mutations
Documentation is incorrect. Moreover, documentation about **authorization** is incorrect.

Here is the [link](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/authorization/authorization.md) that says:

> Type and mutation classes have .authorized?(value, context) class methods

"mutation classes have... class methods"?

Here is example from the [website](https://graphql-ruby.org/mutations/mutation_authorization.html#can-this-user-perform-this-action) about mutations, and there is no class methods:

```ruby
def authorized?(employee:)
  context[:current_user].manager_of?(employee)
end
```

It's confusing. What if I follow the first link and define `authorized?` class method? (`self.authorized?`) for mutation. Will this gets ever called for mutation or it will be skipped?

I don't know if it introduces a security flaw, but it should be fixed.

### LINKED ISSUE #2251: Query validation with schema visibility rules
Hello, I'm having an issue with using `GraphQL::Schema.validate` filtering out fields by visibility, hence marking queries with errors like "Field x doesn't exist on type User".  

## Background

We have some stored queries. I made a test to make sure that some of these queries are valid as a sanity spec. Roughly

```ruby
queries.each do |graphql_query_as_string|
  MySchema.validate(graphql_query_as_string)
end
```

This is invoking the visibility method in my field class. 

```
class QueryField 
  def visible?(context)
    super && allowed_to_view?(context[:current_user]) # no user, #allowed_to_view? is returning false
  end
end
```

## What I expected 
- The ability to pass in the context to validation if schema visibility hooks are running. In this case, I might solve the visibility issue in specs by introducing some form of test user policy that grants full visibility. 

OR 

- Schema visibility hooks not to run for validation. In that case, I'll consider mentally that this validation is "static", doesn't take account of the visibility limitation feature. 

## What seems to be happening

When `MySchema.valdidate` is called, field `#visible?` hooks are invoked, but there is no way to pass an option to prevent that OR pass in a user context to satisfy visibility. 

Thanks for the help in advance!

daemonsy:
## What's up

This resolves #2251. Discussion on the issue is over there. 

## "Bonus"?
- Documentation improvements to resolve #2231 

## What this does 
Allows `.validate` to accept a `context:` hash argument. 

This is passed to Query#initialize and in the function and the rest of the machinery should work as expected.

## What is missing 
- Documentation on the context args?
- Test is a little indirect 
- Not sure if I ran all the tests correctly, assuming there's a PR CI

### DIFF
--- a/guides/authorization/authorization.md
+++ b/guides/authorization/authorization.md
@@ -13,7 +13,7 @@ While a query is running, you can check each object to see whether the current u
 
 Schema members have `.authorized?(value, context)` methods which will be called during execution:
 
-- Type and mutation classes have `.authorized?(value, context)` class methods
+- Type classes have `.authorized?(value, context)` class methods
 - Fields and arguments have `#authorized?(value, context)` instance methods
 
 These methods are called with:
--- a/guides/mutations/mutation_authorization.md
+++ b/guides/mutations/mutation_authorization.md
@@ -25,6 +25,8 @@ This check can be implemented using the `#ready?` method in a mutation:
 ```ruby
 class Mutations::PromoteEmployee < Mutations::BaseMutation
   def ready?(**args)
+    # Called with mutation args.
+    # Use keyword args such as employee_id: or **args to collect them
     if !context[:current_user].admin?
       raise GraphQL::ExecutionError, "Only admins can run this mutation"
     else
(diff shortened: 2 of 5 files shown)


daemonsy:
Hi @rmosolgo submitted a PR for the context in validation. Appreciate your feedback, thank you.

rmosolgo:
Awesome, thanks for taking care of this improvement! (And doc fix :P)

PR #2257, opened 2019-04-26
Reduce allocations in Schema::Validation.validate()

casperisfine:
While memory profiling our application boot I noticed an impressive amount of allocations coming from the `Validation.validate` method:

```
allocated objects by location
...
    713390  /tmp/bundle/ruby/2.5.0/gems/graphql-1.9.4/lib/graphql/schema/validation.rb:16
```

700k allocations is quite a lot for a method that simply iterate over a constant hash. The cause is the many array concatenations inside the first `reduce`

By rewriting the method in a more imperative way it no longer allocated anything.

@rmosolgo

### DIFF
--- a/lib/graphql/schema/validation.rb
+++ b/lib/graphql/schema/validation.rb
@@ -12,11 +12,16 @@ module GraphQL
       # @param object [Object] something to be validated
       # @return [String, Nil] error message, if there was one
       def self.validate(object)
-        rules = RULES.reduce([]) do |memo, (parent_class, validations)|
-          memo + (object.is_a?(parent_class) ? validations : [])
+        RULES.each do |parent_class, validations|
+          if object.is_a?(parent_class)
+            validations.each do |rule|
+              if error = rule.call(object)
+                return error
+              end
+            end
+          end
         end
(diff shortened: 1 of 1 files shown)


casperisfine:
Actually I improperly read the profile, I was only considering one line instead of two:

```
allocated objects by location
    713390  /tmp/bundle/ruby/2.5.0/gems/graphql-1.9.4/lib/graphql/schema/validation.rb:16
...
    431928  /tmp/bundle/ruby/2.5.0/gems/graphql-1.9.4/lib/graphql/schema/validation.rb:15
```

So it's not 700k but 1.1M allocations.

rmosolgo:
Wow, awesome, thanks for hunting this down!

xuorig:
❤️

casperisfine:
Very much welcome. Also I think I told you in a previous PR, but `lib/graphql/define/instance_definable.rb` is also a huge memory hotspot, unfortunately I have a hard time to grasp why it's defining methods lazily, so I haven't managed to improve it yet.

@xuorig 👋 nice to see you here.

PR #2263, opened 2019-05-01
Fix list enum default values in introspection

xuorig:
Fixes the introspection payload for default values where the type is a list of enums.

Example of issue:

```
{
  "name": "affiliations",
  "description": "Array of viewer's affiliation options for repositories returned from the connection. For example, OWNER will include only repositories that the current viewer owns.",
  "type": {
    "kind": "LIST",
    "name": null,
    "ofType": {
      "kind": "ENUM",
      "name": "RepositoryAffiliation",
      "ofType": null
    }
  },
  "defaultValue": "[\"OWNER\", \"COLLABORATOR\"]"
},
```

However:

> defaultValue may return a String encoding (using the GraphQL language) of the default value.

(https://graphql.github.io/graphql-spec/June2018/#sec-The-__InputValue-Type)

This was fixed a while ago for enum values, but the list edge case remained.

### DIFF
--- a/lib/graphql/introspection/input_value_type.rb
+++ b/lib/graphql/introspection/input_value_type.rb
@@ -19,7 +19,11 @@ module GraphQL
           else
             coerced_default_value = @object.type.coerce_result(value, @context)
             if @object.type.unwrap.is_a?(GraphQL::EnumType)
-              coerced_default_value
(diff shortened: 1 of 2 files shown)


rmosolgo:
Oops, thanks!

PR #2266, opened 2019-05-02
Avoid merging distinct aliased selections in Lookahead

dylanahsmith:
## Problem

`Lookahead#selections` wasn't very suitable for iterating over selections that had arguments or type conditional selections, since it neither exposed these on the returned lookahead objects nor even consider these aspects when trying to merge duplicate selections.

## Solution

This PR changes `Lookahead#selections` to no longer be based on `Lookahead#selection`, since that was searching for selections with the same name without considering aliases or whether selections type conditional.  It also seemed like a performance concern, since `Lookahead#selections` was calling `Lookahead#selection` for each selection, which would then do a nested iteration over all the selections, which unnecessarily increased the time complexity of `Lookahead#selections`.  Instead, I extracted out the argument matching into a shared method, so just that part would be shared.

`Lookahead#selections` groups selection ast nodes in a nested hash by the type they are selected on and by their response key.  It then iterates over that nested hash to create the lookahead objects from the ast nodes and to add them to the array to return.


dylanahsmith:
Made the rename to owner_type

rmosolgo:
🤘

PR #2268, opened 2019-05-03
Fixes link to Enums on scalars.md

MatheusRich:
Just a small fix

### DIFF
--- a/guides/type_definitions/scalars.md
+++ b/guides/type_definitions/scalars.md
@@ -8,7 +8,7 @@ desc: Scalars are "simple" data types like integers and strings
 index: 1
 ---
 
-Scalars are "leaf" values in GraphQL. There are several built-in scalars, and you can define custom scalars, too. ({% internal_link "Enums", "/type_definitions/enums" %} are also leaf values.) The built-in scalars are:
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! I found that the link (on https://graphql-ruby.org/type_definitions/scalars.html) works fine for me. Was it broken for you? Could you share how it was broken?

MatheusRich:
@rmosolgo I didn't know this was used on the graphql-ruby website :sweat_smile:
I was reading the docs directly on this repository and saw this:

image

So I thought It had something wrong.

The link to Schema Definition Language is on markdown style and it works both on the website and 
on the repository.

See bellow:
`In the [Schema Definition Language](https://graphql.org/learn/schema/#type-language) (SDL), scalars are simply named:`

Maybe all links should be on markdown style. Any thoughts on this?

rmosolgo:
No, I want _some_ way to make sure that all internal links stay correct. For example, if I remove a page from the website, and some other page links to it, I want to make sure I don't deploy with the dead link. Plain ol' markdown won't cut it. 

For a while I used https://github.com/gjtorikian/html-proofer for this, but it broke when I was adding _new_ pages to the site. html-proofer looked for _already-deployed_ versions of those pages, and reported false positives for dead links. 

So I rigged up `internal_link` to give build errors in this case. 

If you have another suggestion to prevent broken internal links, I'm all ears. But in the meantime, I'd like to keep this link as-is!

PR #2271, opened 2019-05-08
Add link to blog post on parallel resolvers

jonniesweb:
I wrote a blog post a while back which goes over how to parallelize the execution of graphql-ruby resolvers by utilizing the concurrent-ruby library's Future class. It took a few tries to finally figure out how to get this to work, so I think it would be valuable for others in the same situation looking for a speed boost.

Let me know what you think! 😄


rmosolgo:
Awesome, thanks for sharing!

PR #2273, opened 2019-05-08
Connection docs: remove reference to old-style name-based connections

cmatheson:


cmatheson:
I was totally reading the documentation wrong (missed the part about "return type" ending in Connection).

PR #2287, opened 2019-05-15
[Docs] Add note about default complexity.

lsanwick:
This recently bit me as I added a `max_complexity` restriction to my schema and my introspection queries began failing.
My assumption that fields are all scored as complexity "1" by default is based on the code here: https://github.com/rmosolgo/graphql-ruby/blob/a9f334b171a6caaa64bac683f54cedea76d91b28/lib/graphql/schema/resolver.rb#L233

### DIFF
--- a/guides/queries/complexity_and_depth.md
+++ b/guides/queries/complexity_and_depth.md
@@ -10,7 +10,7 @@ index: 4
 
 ## Prevent complex queries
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Oops, sorry that bit you 😖 ! Thanks for improving the docs.

PR #2292, opened 2019-05-17
Typo in referencing an option

daemonsy:
Hey @rmosolgo Happy Friday. I believe this is a typo. 

I finally got to using the `generateClient` function you exposed in https://github.com/rmosolgo/graphql-ruby/pull/1941 and noticed that in one place, there's a reference to `options.clientName` but it seems to be passed as `options.client` and documented as such. 

Btw, the `generateClient` function works brilliantly. 

Just additional context for posterity:

- The sync generates `operation-store-client.js` which has to be consumed by the Apollo client. 
- I'd like to make a webpack plugin to sync queries whenever a `.graphql` file changes in development. So the dev keeps making changes and keeps using persisted queries even in dev. We have this working with a custom webpack script.  
- The hard part is that since the sync task writes to disk during the sync, it generates a file change that causes webpack to recompile, ad infinitum. 

The current idea is to write virtual modules (no need to check in) and a json manifest somewhere (to be checked in) so it's easy to reference all the query aliases. 

Might submit a PR to break down the sync function :)

### DIFF
--- a/javascript_client/sync/generateClient.js
+++ b/javascript_client/sync/generateClient.js
@@ -24,7 +24,7 @@ var generators = {
 */
 function generateClient(options) {
   var payload = gatherOperations(options)
-  var generatedCode = generateClientCode(options.clientName, payload.operations, options.clientType)
+  var generatedCode = generateClientCode(options.client, payload.operations, options.clientType)
   return generatedCode
 }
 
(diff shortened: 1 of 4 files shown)


daemonsy:
Looking into specs failure

daemonsy:
I updated the spec that is passing in `clientName`, but I don't really understand what `syncTest.js` is doing...

javascript_client/sync/generateClient.js
```diff
@@ -45,8 +45,6 @@ function gatherOperations(options) {
   var graphqlGlob = options.path || "./"
   var hashFunc = options.hash || md5
   var filesMode = options.mode || (graphqlGlob.indexOf("__generated__") > -1 ? "relay" : "project")
-  var clientName = options.client
```
daemonsy: These are not used it seems

rmosolgo:
Sorry, apparently I broke master a few days ago 🤦‍♂ . Thanks for the fix here!

Issue #2261, opened 2019-04-30
subscription doesn't return the initial result if `use GraphQL::Subscriptions::ActionCableSubscriptions` is added


github0013:
## environments

```
Darwin abc.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Dec 20 21:53:35 PST 2018; root:xnu-3789.73.31~1/RELEASE_X86_64 x86_64
ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-darwin16]
Rails 5.2.3
graphql (1.9.4)

```

### GraphqlChannel.rb

https://graphql-ruby.org/api-doc/1.9.3/GraphQL/Subscriptions/ActionCableSubscriptions

```rb
payload = {
      result: result.subscription? ? { data: nil } : result.to_h,
      more: result.subscription?,
    }
```

This has to be changed to below.  Otherwise the initial subscription result will be ignored.

```rb
    payload = { result: result.to_h, more: result.subscription? }

```

## Expected

```rb
# https://graphql-ruby.org/subscriptions/subscription_classes.html#adding-an-initial-response
class Subscriptions::MessageWasPosted < Subscriptions::BaseSubscription
  # ...
  field :room, Types::RoomType, null: true

  def subscribe(room:)
    # authorize, etc ...
    # Return the room in the initial response
    {
      room: room
    }
  end
end
```

This subscription initial call should always return the result in the `def`

## Actual

The `def` result comes back **only when** `use GraphQL::Subscriptions::ActionCableSubscriptions` is **not** added in the schema.

```rb
class YourSchema < GraphQL::Schema
  # this will return the initial result, 
  # but can't use YourSchema.subscriptions.trigger because YourSchema.subscriptions is nil
  # use GraphQL::Subscriptions::ActionCableSubscriptions

  # this will NOT return initial result (only returns { data: nil } ) , but .trigger works
  use GraphQL::Subscriptions::ActionCableSubscriptions

  mutation(Types::MutationType)
  query(Types::QueryType)
  subscription(Types::SubscriptionType)
end

```

rmosolgo:
Hi, thanks for this detailed bug report, and sorry for the hangup 😖 !

github0013:
`@inner_proc.call(obj, args, ctx)` in this line
https://github.com/rmosolgo/graphql-ruby/blob/v1.9.4/lib/graphql/subscriptions/instrumentation.rb#L47
returns the actual result, but it's never used.

[...]

PR #2294, opened 2019-05-18
return result of initial subscription

tkamenick:
@github0013 submitted a great bug report and dug in to find a solution to the bug in #2261. I recently hit this same bug and would like to see it fixed, so I created a PR out from their work.

I've tested that this works for a subscription as described the same way in the issue.


rmosolgo:
Thanks, I want to look at this and #2145 soon and try to get these changes in 1.10.0!

ncphillips:
Could this be released in a bug fix update? i.e. release a `1.9.7`

rmosolgo:
This will be in 1.10.0.pre1!

PR #2297, opened 2019-05-21
Add `subscription_scope` method to `Subscription` class

joeytepp:
As of now, if you want to make a scoped subscription using the new `Subscription` class, the only way to do so is the following:

```ruby
  # my_subscription.rb
  class MySubscription < BaseSubscription
    ...
  end
```

```ruby
  # subscription_type.rb
  class SubscriptionType < GraphQL::Schema::BaseObject
    field :my_subscription, subscription: MySubscription, subscription_scope: :current_user_id
  end
```

My proposal is to add a `subscription_scope` method to the `Subscription` class so that all the options for the subscription can be defined within the class itself (more like the `Query` or `Mutation` class definitions). Using this change, the same example would become

```ruby
  # my_subscription.rb
  class MySubscription < BaseSubscription
    ...
    subscription_scope :current_user_id
  end
```

```ruby
  # subscription_type.rb
  class SubscriptionType < GraphQL::Schema::BaseObject
    field :my_subscription, subscription: MySubscription
  end
```

### DIFF
--- a/lib/graphql/schema/subscription.rb
+++ b/lib/graphql/schema/subscription.rb
@@ -92,6 +92,25 @@ module GraphQL
       def unsubscribe
         raise UnsubscribedError
       end
+
+      # Call this method to provide a new subscription_scope; OR
+      # call it without an argument to get the subscription_scope
+      # @param new_scope [Symbol]
+      # @return [Symbol]
+      def self.subscription_scope(new_scope = nil)
+        if new_scope
+          @subscription_scope = new_scope
+        else
+          @subscription_scope || find_inherited_method(:subscription_scope, nil)
+        end
+      end
+
+      # Overriding Resolver#field_options to include subscription_scope
+      def self.field_options
+        super.merge(
+          subscription_scope: subscription_scope
+        )
+      end
     end
   end
 end
--- a/lib/graphql/schema/subscription.rb
+++ b/lib/graphql/schema/subscription.rb
@@ -97,11 +97,14 @@ module GraphQL
       # call it without an argument to get the subscription_scope
       # @param new_scope [Symbol]
       # @return [Symbol]
-      def self.subscription_scope(new_scope = nil)
-        if new_scope
+      READING_SCOPE = ::Object.new
+      def self.subscription_scope(new_scope = READING_SCOPE)
+        if new_scope != READING_SCOPE
           @subscription_scope = new_scope
+        elsif defined?(@subscription_scope)
+          @subscription_scope
         else
-          @subscription_scope || find_inherited_method(:subscription_scope, nil)
+          find_inherited_method(:subscription_scope, nil)
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/subscription.rb
```diff
@@ -92,6 +92,25 @@ def load_application_object_failed(err)
...
+      # call it without an argument to get the subscription_scope
+      # @param new_scope [Symbol]
+      # @return [Symbol]
+      def self.subscription_scope(new_scope = nil)
+        if new_scope
+          @subscription_scope = new_scope
+        else
+          @subscription_scope || find_inherited_method(:subscription_scope, nil)
+        end
+      end
```
rmosolgo: This won't support _overriding_ an inherited scope with `nil`. Do you want to? Here's an implementation that would support overriding with `nil`: 

```ruby 
READING_SCOPE = Object.new 
def self.subscription_scope(new_scope = READING_SCOPE) 
  if new_scope != READING_SCOPE 
    @subscription_scope = new_scope 
  elsif defined?(@subscription_scope) 
    @subscription_scope # This might return `nil` if it was explicitly overridden 
  else 
    find_inherited_method(:subscription_scope, nil) 
  end 
end 
```

joeytepp: I didn't consider that, makes a lot of sense. Based it on the implementation of similar fields in the `Resolver class` but I like this implementation better. Will add this in along with some more tests 👌

joeytepp:
@rmosolgo made changes based on your feedback, I'll keep an eye out for your response 👍

spec/graphql/schema/subscription_spec.rb
```diff
@@ -437,4 +444,59 @@ def in_memory_subscription_count
...
+            }
+          }
+        }
+      }
+
+      assert_equal expected_response, mailbox.first
+    end
+
+    it "allows for proper inheritance of the class's configuration through `field_options`" do
+      direct_toot_options = SubscriptionFieldSchema::DirectTootWasTooted.field_options
```
rmosolgo: Is the `subscription_scope` being inherited here? It seems like we're checking the value from the _same class_ where we manually configured it. 

I'd like to check that the value is inherited by subclasses, for example: 

```suggestion
      # Make a subclass without an explicit configuration 
      direct_toot_subclass = Class.new(SubscriptionFieldSchema::DirectTootWasTooted)
      # Then check if the field options got the inherited value
      direct_toot_options = direct_toot_subclass.field_options
```

How does that sound?

joeytepp: Ah I see, slight confusion on my part but will update with these changes along with a small change to one of the other tests.

joeytepp:
@rmosolgo fixed the test, should be good for another review!

rmosolgo:
Looks great! Thanks again for taking this on. I agree it's a lot better this way.

noahifergan:
This sucks!

PR #2305, opened 2019-05-23
Backport String#-@ for Ruby 2.2

rmosolgo:
Oops, I intend to support Ruby 2.2, but this was breaking it. (The real problem is, I can't figure out how to run Ruby 2.2 on Travis CI.) I ran Ruby 2.2 locally and it passed after these changes. 

```
~/code/graphql-ruby $ ruby -v
ruby 2.2.8p477 (2017-09-14 revision 59906) [x86_64-darwin16]
~/code/graphql-ruby $ be rake test
# ...
1381 tests, 5871 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,10 @@ matrix:
     gemfile: spec/dummy/Gemfile
     script:
       - cd spec/dummy && bundle exec rails test:system
+  - rvm: 2.2.10
+    before_install:
+      - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
+      - gem install bundler -v '< 2'
   - env:
       - TESTING_INTERPRETER=yes
     rvm: 2.4.3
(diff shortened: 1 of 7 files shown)


JanStevens:
TBH 🍻 for supporting ruby 2.2! 

I'm just not sure if for a gem this big it makes any sense to support an EOL ruby version. It's adding more cruft to an already huge amount of code.

ref: https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

rmosolgo:
Yeah, but people have all kinds of ✨ reasons ✨ that they have to use old versions of things. I'd rather accommodate that reality and support people using the gem rather than cut them off. (And in this case, there's at least one graphql-pro customer who's using 2.2 -- that's what they pay me for 😅 !)

Issue #2214, opened 2019-04-03
SystemStackError: stack level too deep with to_json on graphql-ruby objects


bbugh:
I reported recently in Slack about a `SystemStackError: stack level too deep` error. I discovered the cause of the issue and will document it here. This was hard to track down because of the sequencing.

**Current Behavior**

We're using `graphql-errors` since the built-in `rescue_from` only returns an error message without `extensions`. It turns out that calling `to_json` on some of the `graphql-ruby` objects that are passed in will create a `SystemStackError`:

```ruby
GraphQL::Errors.configure(GraphQLSchema) do
  rescue_from StandardError do |exception, object, arguments, context|
    # any of these will raise SystemStackError: stack level too deep
    object.to_json
    arguments.to_json
    context.to_json
  end
end
```

Our problem was that when an otherwise unhandled exception occurred, we would create a Rollbar error, which internally called `to_json` on these objects:

```ruby
Rollbar.error("Unhandled StandardError: #{exception.message}", 
              exception, 
              arguments: arguments,  # to_json gets called on this
              context: context)      # to_json gets called on this
```

When this method call failed with `SystemStackError`, Rollbars catch-all error handler captured *that* exception and reported it instead, which did not have a useful stack trace.

**Desired Behavior**

This issue is happening with `graphql-errors`, but the actual objects with the `to_json` problem are `graphql-ruby` objects. 

It turns out that `context` overflowing `to_json` is my own fault, but `object` and `arguments` are internally provided by `graphql-ruby`. `arguments` seems like it should be convertible to JSON since it's basically a hash. `object` is `Types::QueryType` and I'm not sure if there's a reasonable way to define `to_json`, but maybe it could not create a stack overflow.

```
[1] pry(#<GraphQL::Errors>)> arguments.to_json
SystemStackError: stack level too deep
from /Users/bbugh/.rvm/gems/ruby-2.6.2/gems/activesupport-5.2.2.1/lib/active_support/core_ext/object/json.rb:171:in `dup'

[2] pry(#<GraphQL::Errors>)> arguments.to_h
=> {"id"=>"1"}
```

We're 100% on the class-based syntax, and not using any custom field definitions like #1957.

cc: @theorygeek in case this is relevant for your issue

rmosolgo:
Thanks for the detailed bug report! I agree `arguments.to_json` should Just Work :tm:. Perhaps it should be aliased to `#to_h`.

theorygeek:
[...]

PR #2313, opened 2019-05-30
Add InputObject#as_json for Rails via Railtie

daemonsy:
## What's up? 

This is an attempt to resolve https://github.com/rmosolgo/graphql-ruby/issues/2214. 

I've bumped into it a few times myself calling `input_object.to_json`. I guess it's a fairly common use case to serialize this for error reporting etc or when you try to serialize all the args for a query. 

## The Issue 

The last resort of Rails to fulfill `to_json` serialization of PORO Ruby objects is to use instance variables. 

https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/core_ext/object/json.rb#L57

https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/core_ext/object/instance_variables.rb#L15

I think in the case of `InputObject`, the instance variables references the object itself, causing a `CircularReferenceError` or a `SystemStackError` error.  Weird, I've seen both. 

Given that there's a bunch of machinery (e.g. `#unwrap_value`) in the class, it doesn't seem worth it to try to make `InputObject` a serializable PORO.

## The potential fixes

1. Define `#as_json(*)` and that calls `#to_h` under the hood. This PR does that via Railtie. 
2. Implement the implicit conversion method `#to_hash` which will be used by ActiveSupport in `#to_json` [Link](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/core_ext/object/json.rb#L55). Calling `#to_json` on `GraphQL::Query::Context` actually works because of this, `to_hash` is defined there. 

I personally think (2) is cleaner, but it also makes this implicit assumption that `InputObject` is implicitly a hash. Would like @rmosolgo 's opinion on that. Example [blog post](https://zverok.github.io/blog/2016-01-18-implicit-vs-expicit.html) about this.

### DIFF
--- a/lib/graphql/rails_integration/serialize_as_json.rb
+++ b/lib/graphql/rails_integration/serialize_as_json.rb
@@ -1,0 +1,9 @@
+module GraphQL
+  module RailsIntegration
+    module SerializeAsJSON
+      def as_json(*)
+        to_h
+      end
+    end
+  end
+end
--- a/lib/graphql/railtie.rb
+++ b/lib/graphql/railtie.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
-
+require 'graphql/rails_integration/serialize_as_json'
 
 module GraphQL
   class Railtie < Rails::Railtie
+    initializer 'graphql.active_support extensions' do
+      GraphQL::Schema::InputObject.include RailsIntegration::SerializeAsJSON
+    end
+
     rake_tasks do
       # Defer this so that you only need the `parser` gem when you _run_ the upgrader
       def load_upgraders


daemonsy:
Hi hi @rmosolgo if you have a chance to take a look, I'd appreciate feedback on where to take this. Thanks!

rmosolgo:
Hi, thanks for the ping! I'm getting back to my day job this week so I hope I can catch up on this PR in a day or two!

daemonsy:
Sure thing! Thank you :smile:

daemonsy:
Hey @rmosolgo just a bump on this. I've seen other places where `to_hash` is implemented. 

Happy to take it down either path to complete this PR. 

Just got reminded because someone asked me advice for a problem and it was the `to_json `issue

rmosolgo:
Hey, thanks for your patience while I review here. I admit I was scared by the title ... but the diff is pretty straightforward 😆 

Thanks for the links to explanations of this. I'm open to either approach (`#as_json` or `#to_hash`), since I think a case can be made either way. 

My question is, why should this be done in a Railtie? Is there a downside to adding this to the class directly?

daemonsy:
Hey @rmosolgo no real downside to me except that since graphql-ruby is not Rails specific, I didn't want to add code that's really only used for Rails.

rmosolgo:
Ok, for simplicity's sake, let's put it in the class definition instead :+1: Your call as to which method IMO.

daemonsy:
Superseded by #2343

PR #2319, opened 2019-06-11
Mention bug with Subscription#subscribe

ncphillips:
Added a note about the bug preventing the use of the `subscribe` method when using ActionCable.

Bug in question is #2261


rmosolgo:
This isn't exactly a bug: a new feature was added in the `Interpreter` but it's not supported in the old runtime. If you try the new interpreter runtime, do you get the initial value?

ncphillips:
We do have the new `Interpreter` runtime included. When we setup the subscription as a class the `subscribe` method was called but the value it returns was not returned. The `update` method does work as expected.

rmosolgo:
This should be fixed in 1.10.0.pre1!

PR #2320, opened 2019-06-11
Weird infinity bug

xuorig:


rmosolgo:
Fixed on 1.10-dev in https://github.com/rmosolgo/graphql-ruby/pull/2363/files#diff-c060ea1e0aeae884e69e99cc25158c97R47

PR #2323, opened 2019-06-12
Input object loads bug

### LINKED ISSUE #2280: Loads on InputObject is not working with mutations
New version of graphql introduced load on input object. I think they might not be working properly for mutations.

My schema looks like this:
```ruby
class MyMutation < BaseMutation
  argument :parent_id, ID, required: true, loads: Types::ParentType
  argument :my_object, [Types::Inputs::MyInput], required: true
end

module Types
  module Inputs
    class MyInput < BaseInputObject
      argument(
        :child_id,
        ID,
        loads: Types::ChildType,
        required: true
      )
    end
  end
end

```
I am getting some weird behaviour for this. So my ParentType is loaded properly.
Then `resolve` method in mutation is called. Argument `my_object` looks like this:
```
[{
  child: "EncodedIdHere"
}]
``` 
So the key name was transformed (no _id suffix), but ID was not decoded. After the resolution `GraphqlSchema.object_from_id` is called with first argument being `nil`.

rmosolgo:
Fixes #2280 

The `loads:` implementation should apply to key lookup on input objects, too!

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -21,6 +21,9 @@ module GraphQL
       # @return [Symbol] This argument's name in Ruby keyword arguments
       attr_reader :keyword
 
+      # @return [Class, Module, nil] If this argument should load an application object, this is the type of object to load
+      attr_reader :loads
+
       # @param arg_name [Symbol]
       # @param type_expr
       # @param desc [String]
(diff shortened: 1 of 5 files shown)


tjwallace:
I don't think this works when `Schema.object_from_id` returns a lazy value. Should I open a new bug for this?

PR #2327, opened 2019-06-13
Add method for disabling introspection entrypoints

DmitryTsepelev:
That might be helpful to turn off introspection entry points, especially in the production environments. The shortest way to do it I found (given the fact that we don't have any custom entry points):

```ruby
  if Rails.env.production?
    introspection(Module.new do
      class EntryPoints < GraphQL::Introspection::EntryPoints
        def self.authorized?(*)
          false
        end
      end
    end)
  end
end
```

It looks a little bit verbose and does not hide entry points from the schema, just restricts the access. I've added a shorthand method `#disable_introspection_entry_points` which allows to skip generating entry points:

```ruby
  disable_introspection_entry_points Rails.env.production?
end
```

### DIFF
--- a/guides/schema/introspection.md
+++ b/guides/schema/introspection.md
@@ -187,3 +187,13 @@ end
 ```
 
 Any fields defined there will be available in any selection, but ignored in introspection (just like `__typename`).
+
+## Disabling Introspection
+
+In case you want to turn off introspection entry points `__schema` and `__type` (for instance in the production environment) you can use a `#disable_introspection_entry_points` shorthand method:
+
+```ruby
+class MySchema < GraphQL::Schema
+  disable_introspection_entry_points if Rails.env.production?
+end
+```
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -90,7 +90,7 @@ module GraphQL
       :object_from_id, :id_from_object,
       :default_mask,
       :cursor_encoder,
-      :disable_introspection_entry_points,
+      disable_introspection_entry_points: ->(schema) { schema.disable_introspection_entry_points = true },
       directives: ->(schema, directives) { schema.directives = directives.reduce({}) { |m, d| m[d.name] = d; m } },
       directive: ->(schema, directive) { schema.directives[directive.graphql_name] = directive },
       instrument: ->(schema, type, instrumenter, after_built_ins: false) {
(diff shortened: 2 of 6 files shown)


eapache:
We accomplished something similar, but we used the existing `except` argument for schema visibility and passed it a class that just does `schema_member.graphql_name == '__schema' || schema_member.graphql_name == '__type'`.

DmitryTsepelev:
@eapache  ...or it's possible to define resolving method raising an error for each entry point.

I'm not sure if this use case is common enough (but your comment makes me think I'm not alone), but if it is - I guess it would be great to have an easy way to get this behavior

eapache:
Our approach was simple enough (4 lines of code) that I'm not sure it's worth actually packaging into the gem. Does seem worth to document a recommended way though.

DmitryTsepelev:
Let's wait for Robert to take a look - I'll be happy to transform this PR to the documentation amendment

rmosolgo:
Hi, thanks for the contribution! I agree that this is a common enough request that it would  be a good addition to the gem. 

I just have one request about the API of the new method. Could we please change it to: 

```ruby 
def disable_introspection_entry_points(if: true) 
  # ... 
```

I think that will yield more obvious code in the schema class, for example: 

```ruby 
if Rails.env.production? 
  disable_introspection_entry_points 
end 

# or 
disable_introspection_entry_points(if: Rails.env.production?)
```

Thanks for including tests also. If you don't mind that small change, looks good to me!

DmitryTsepelev:
Sure, I've changed the API, but may be it worth change keyword kwarg `if` to something else to avoid that dirty hack around `binding`? Maybe we should stay with
```ruby
if Rails.env.production? 
  disable_introspection_entry_points 
end 
```
and allow to pass the optional argument?

rmosolgo:
> avoid that dirty hack

🤦‍♂ Yes, you're totally right about that, sorry I forgot! Better to have the optional argument like you said.

DmitryTsepelev:
Done! Could you please review once again?

celsworth:
Is there any value to the optional `if:` argument? Could you not accomplish the same with normal Ruby flow control?

By that I mean to say, what's better about doing:

```ruby
disable_introspection_entry_points(if: Rails.env.production?)
```

Instead of:

```ruby
disable_introspection_entry_points if Rails.env.production?
```

rmosolgo:
I'd be happy to drop the `if` argument, but at the same time, I'm happy to accommodate @DmitryTsepelev 's preference. He's doing the _work_ 😆

DmitryTsepelev:
@celsworth that's a fair point 🙂 I've updated my commit @rmosolgo

rmosolgo:
Awesome, thanks again for this! I pushed a commit to make it play a bit nicer with the `.define` system.

DmitryTsepelev:
Thank you! ✨

PR #2331, opened 2019-06-13
Keep a second copy of arguments under their snake name if different

### LINKED ISSUE #2329: Args in complexity are camel cased instead of snake cased
The names of args passed to `complexity` `Proc`s in the `args` hash are camel cased instead of snake cased.

In the attached test `snake case` fails and `camel case` passes.

Test:
```ruby
  describe "case of complexity arguments" do
    let(:query) { GraphQL::Query.new(complexity_schema, query_string) }
    let(:complexity_schema) {
      query_type = Class.new(GraphQL::Schema::Object) do
        graphql_name "Query"
        field :snakes, [String], null: false do
          argument :snake_case, [Integer], required: true
          complexity ->(_ctx, args, _child_complexity) do
            args[:snake_case]&.size || 0
          end
        end

        def snakes(**_args)
          []
        end

        field :camels, [String], null: false do
          argument :camel_case, [Integer], required: true
          complexity ->(_ctx, args, _child_complexity) do
            args[:camelCase]&.size || 0
          end
        end

        def camels(**_args)
          []
        end
      end

      Class.new(GraphQL::Schema) do
        query query_type
      end
    }

    describe "snake case" do
      let(:query_string) {%|
        {
          snakes(snakeCase: [1, 2, 3])
        }
      |}

      it "sums the complexity" do
        reduce_result

        assert_equal complexities, [query, 3]
      end
    end

    describe "camel case" do
      let(:query_string) {%|
        {
          camels(camelCase: [1, 2, 3])
        }
      |}

      it "sums the complexity" do
        reduce_result

        assert_equal complexities, [query, 3]
      end
    end
  end
```

Version 1.9.6

ecopoesis:
Arguments passed to the `complexity` `Proc` are camel cased instead of the expected snake case. This change adds a second copy of the argument in its snake case form if the snake case version is different.

Fixes #2329


rmosolgo:
Hi, thanks for taking a stab at this! I'm a bit hesitant to mess with `Query::Arguments` because this will touch _everywhere_ that arguments are used (as shown by the required change to `arguments_spec.rb`). In the rare case that someone is _distinguishing between_ camel-cased and snake-cased arguments, this could break their schema.

I have another suggestion. How about, when running complexity checks, _detect_ whether the current field is class-based, and if it is, then use a plain, snake-cased Hash for `args` instead of a `Query::Arguments`? 

That approach makes complexity's `args` just like the arguments passed to field methods. 

You could detect whether a field is class-based by looking for `!!field.metadata[:type_class]`. (In this case, `metadata[:type_class]` doesn't actually return a _class_, it returns a GraphQL::Schema::Field instance if the field was created from one.) 

To convert the `Query::Arguments` instance into a plain, snake-cased Hash, you can use `#to_kwargs`: 

https://github.com/rmosolgo/graphql-ruby/blob/2bf7dcd2e206a5602909fd2e6ae17436bff78ae6/lib/graphql/query/arguments.rb#L108-L118

For example, this is even how the runtime does it: 

https://github.com/rmosolgo/graphql-ruby/blob/85554e8f1c9bfa499fe14a912b92de8c83d99932/lib/graphql/schema/field.rb#L581-L582

What do you think about this approach? Want to give it a shot?

ecopoesis:
I've switched to the approach you've outlined. The downside of course is that it will break compatibility with existing code that relies on `camelCase` in complexity args.

spec/graphql/analysis/query_complexity_spec.rb
```diff
@@ -298,4 +298,67 @@
...
+    end
+
+    describe "camel case" do
+      let(:query_string) {%|
+        {
+          camels(camelCase: [1, 2, 3])
+        }
+      |}
+
+      it "sums the complexity" do
```
rmosolgo: Does this sum the complexity, or does it _miss_, since there's no camel-case argument?

ecopoesis: Good point. I've updated the rest name.

rmosolgo:
Sorry I didn't follow up on this before :( I just added a test to make sure this works as expected with the new `Analysis::AST` module and it does, so we should be good to go from here on out. Thanks so much for looking into this and suggesting a fix!

PR #2334, opened 2019-06-14
Add FilteredArgumentsPrinter for redacting argument values

daemonsy:


rmosolgo:
Seems like a good start to me. I guess it's a bit simpler because it replaces _all_ scalars with `<FILTERED>`, which is probably a good tradeoff. 

For those following along at home, here's a similar solution extracted from GitHub: https://gist.github.com/rmosolgo/3452d2bf98837381df25949498e78e79

Issue #1153, opened 2017-11-30
1.8 upgrade doc notes


eapache:
Sorry for the brain-dumpy format, I just helped somebody through an experimental migration to the 1.8 format using https://github.com/rmosolgo/graphql-ruby/blob/c39c352e7e2e68887b02e9e0619d7c7846153b33/guides/schema/class_based_api.md and came out with a list of notes:

- If you were using the field block syntax before most of those will now throw `NoMethodError` (e.g. `description` or `resolve`) and there's no hint anywhere that they need to be converted to kwargs on the `field` call.
- docs still say arguments take `null`, but it's now `required` just like inputs
- "Resolution with methods" section seems to be missing part of the explanation? At the least, it's in dire need of an example.

rmosolgo:
👌  Thanks for sharing! I'll take a crack at making the right updates to that page.

rmosolgo:
> field block syntax

Curious about your opinion on this one, should we keep the flexibility to put these in the block? Open for discussion at #1158

rmosolgo:
Took a stab at the method docs and arg docs in #1160 , thanks gain for pointing those out!

eapache:
Thank you!

Issue #1133, opened 2017-11-27
Arguments on scalar types?


lsanwick:
Looking at the docs, it appears that if I define my own scalar types, there is no way to define arguments for them to take.  The GraphQL documentation gives examples of using arguments with scalar types (like here: http://graphql.org/learn/queries/#arguments), is there any way to do this in graphql-ruby?

rmosolgo:
Yes, those are _field_ arguments, for example: 

```ruby
field :phoneNumber, types.String do 
  argument :includeCountryCode, types.Boolean, default_value: true 
  resolve ->(obj, args, ctx) { ... } 
end 
```

could be used as: 

```graphql 
phoneNumber(includeCountryCode: false) 
# => "555-1234"
```

For custom scalars, the only change is to switch `types.String` for your custom `Types::PhoneNumberType`: 

```ruby
field :phoneNumber, Types::PhoneNumberType do 
  argument :includeCountryCode, types.Boolean, default_value: true 
  resolve ->(obj, args, ctx) { ... } 
end 
```

Usage is the same as above:

```graphql 
phoneNumber(includeCountryCode: false) 
# => "555-1234"
```

But your custom `PhoneNumberType` will be used to format the value.

Hope that helps, feel free to follow up here if you have any more trouble!

lsanwick:
Thanks for the reply @rmosolgo, that's helpful! In my case I'd like to have a TimeType that takes a `format` argument to determine how to render the time, but I would prefer to not have that format switching logic occurring in multiple places around my schema. Is there any recommendation on the proper way to encapsulate that logic?

rmosolgo:
It sounds like there would be many different fields (`Event.startsAt`, `Room.reservedUntil`, `Notification.dismissedAt`) that would each take a `format:` argument, right? And they should all perform the same string formatting based on the `format:` value. 

It seems like a [resolve wrapper](http://graphql-ruby.org/fields/resolve_wrapper.html) might do the job. The inner `resolve` function could return the proper Ruby `DateTime`, then the wrapper could perform formatting into a String, based on the argument value.

00dav00:
Hi @rmosolgo, after updating to `1.10` a project I work with started raising the exception described in this issue: https://github.com/jetruby/apollo_upload_server-ruby/issues/24

```
Failures:

  1) ActivityLog userActivityLog returns activity for the current user
     Failure/Error: mutation(Types::MutationType)
     
     ArgumentError:
[...]

PR #2335, opened 2019-06-15
upgrade js packages

birge:
Have been getting warnings after upgrading graphql.

`warning "graphql-ruby-client > apollo-link@1.0.5" has incorrect peer dependency "graphql@^0.11.3".`

Of the client side changes only the graphql package has breaking changes and none of them should effect this.

Associated change logs:
graphql: https://github.com/graphql/graphql-js/releases
apollo-link: https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link/CHANGELOG.md
glob: https://github.com/isaacs/node-glob/blob/master/changelog.md (though they dont seem to document patch updates)

dev dependencies:
jest: https://github.com/facebook/jest/blob/master/CHANGELOG.md
nock: https://github.com/nock/nock/releases?after=v10.0.5

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -8,16 +8,16 @@
   "license": "LGPL-3.0",
   "bin": "./cli.js",
   "devDependencies": {
-    "jest": "^20.0.4",
-    "nock": "^9.0.14"
+    "jest": "^24.8.0",
+    "nock": "^10.0.6"
   },
   "scripts": {
     "test": "jest"
   },
   "dependencies": {
-    "apollo-link": "^1.0.3",
-    "glob": "^7.1.2",
-    "graphql": "^0.10.5",
+    "apollo-link": "^1.2.12",
+    "glob": "^7.1.4",
+    "graphql": "^14.3.1",
     "minimist": "^1.2.0"
   }
 }
--- a/javascript_client/yarn.lock
+++ b/javascript_client/yarn.lock
@@ -2,134 +2,494 @@
 # yarn lockfile v1
 
 
-"@types/zen-observable@0.5.3":
-  version "0.5.3"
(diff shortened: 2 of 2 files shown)


daemonsy:
Hey @birge thanks for the PR and tracking the changes in the other libraries. I've been doing some unrelated work with `graphql-js`. There's been some changes (generally improvements) on query parsing etc. 

I'm going to take the branch out on a spin, sanity check it against some apps I have using persisted queries and then leaving a comment.

daemonsy:
I did a sanity check on some internal apps that do persisted queries syncing. They all worked alright. 

Looking at the code in `javascript_client`, it seems like the use of `graphql-js` is around subscriptions, `parse` and `visit`. My codebases don't have subscriptions, so did zero testing there. 

@rmosolgo I spotted a few items in the changelog look alright, but have the slightest relation to query validation/parsing or subscriptions. Other than that, this looks good during the test drive. 

⚠️  https://github.com/graphql/graphql-js/releases/tag/v0.11.0
> The subscribe() function now returns a Promise of either an AsyncIterator or ExecutionResult to better distinguish between initialization errors and publish errors; previously it just returned an AsyncIterator (#918).

https://github.com/graphql/graphql-js/releases/tag/v0.12.0
> Validation errors for value literals (and variable values) are greatly improved, doing so results in a change to the standard set of validation rules. (#1126, #1133, #1144, #1153)

https://github.com/graphql/graphql-js/releases/tag/v14.0.0
> Removes VariablesDefaultValueAllowed validation rule, and ProvidedNonNullArguments became ProvidedRequiredArguments (#1274)

rmosolgo:
Thanks for checking into this and making the upgrades! I _think_ we'll be clear with GraphQL-JS's changes with subscriptions, since we're usually interacting with it via Relay or Apollo.

birge:
@rmosolgo can you publish this to npm?

rmosolgo:
Yep! Just shipped https://www.npmjs.com/package/graphql-ruby-client/v/1.6.5 . Thanks again for the upgrades here.

PR #2343, opened 2019-06-21
Implement #to_hash so input objects can be serialized/splatted

daemonsy:
## What's up? 

This is a continuation of https://github.com/rmosolgo/graphql-ruby/pull/2313. I accidentally deleted the fork so a new PR is needed :smiley_cat:. 

I've used the `#to_hash` approach here to be consistent with an accepted PR https://github.com/rmosolgo/graphql-ruby/pull/1955 that implements `#to_hash` in context. 

## What this solves 
- You can call `input_object.to_json` now without a stack too deep error. 
- You can also splat input objects to get their hashes

### Some concerns before merging
- [x] Am I initializing the dummy input object correctly in the test? 
- [x] A little confused about the different between `InputObjectType` and `GraphQL::Schema::InputObject`

### DIFF
--- a/spec/graphql/schema/input_object_spec.rb
+++ b/spec/graphql/schema/input_object_spec.rb
@@ -277,7 +277,7 @@ describe GraphQL::Schema::InputObject do
     end
   end
 
-  describe "#to_h" do
+  describe 'hash conversion behavior' do
     module InputObjectToHTest
       class TestInput1 < GraphQL::Schema::InputObject
         graphql_name "TestInput1"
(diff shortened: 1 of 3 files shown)


daemonsy:
Is `GraphQL::InputObjectType` a legacy class? If it is, maybe we can mark the class as deprecated. I spent some time adding tests cases to `input_object_type_spec.rb`, then realize it's different :P

spec/graphql/schema/input_object_spec.rb
```diff
@@ -297,19 +297,30 @@ class TestInput2 < GraphQL::Schema::InputObject
...
 
-      assert_equal({ a: 1, b: 2, input_object: { d: 3, e: 4, instrument: Jazz::Models::Instrument.new("Drum Kit", "PERCUSSION") } }, input_object.to_h)
+    describe "#to_hash" do
+      it "returns the same results as #to_h (aliased)" do
+        assert_equal(@input_object.to_h, @input_object.to_hash)
+      end
     end
   end
 
+
```
rmosolgo:
```suggestion
```

rmosolgo: (I tried to commit it myself but it wouldn't let me 🤷‍♂ )

daemonsy: I'll try to get Rubocop working to reduce the amount of inconsistencies, thanks!

Issue #2243, opened 2019-04-11
Parser too flexible for "empty" documents


swalkinshaw:
Compared to graphql-js, `GraphQL.parse` seems to be much more flexible and allowing for "empty"-ish documents. Two examples are a query string with comments *only* and just `{}`.

```ruby
[1] pry(main)> GraphQL.parse("#comment")
=> #<GraphQL::Language::Nodes::Document:0x00007ff58dc8ead0 @definitions=[], @filename=nil>
[3] pry(main)> GraphQL.parse("{}")
=> #<GraphQL::Language::Nodes::Document:0x00007ff5813e8f90 @definitions=[#<GraphQL::Language::Nodes::OperationDefinition:0x00007ff5813e91c0 @col=1, @directives=[], @filename=nil, @line=1, @name=nil, @operation_type="query", @selections=[], @variables=[]>], @filename=nil>
```

In graphql-js, both raise an error.

* comments only: `GraphQLError: Syntax Error: Unexpected <EOF>`
* `{}`: `GraphQLError: Syntax Error: Expected Name, found }`

rmosolgo:
Happy to consider a change here, especially if it's in the spec. (And also if it's _not_!)

rmosolgo:
Fixed by #2344

PR #2344, opened 2019-06-21
Raise ParseError for some queries that are considered invalid by the reference implementation

daemonsy:
**Update: Refer to new write up below**

## What's up?

This is an attempt to resolve https://github.com/rmosolgo/graphql-ruby/issues/2243. 

Both `libgraphqlparser` and `graphql-js` raises EOF errors when the only token in a query is a comment. 

a.k.a documents with no selection sets are not possible, as far as the reference implementations tell. 

## What this does 
- Raise `ParseError` on empty sets 
- Add documentation about ragel/colm

### Update 26 June 2019: 
- Also raising errors on `node() { id }` where the `node()` is not considered legal form by `libgraphqlparser` and `graphql-js`.


graphql.gemspec
```diff
@@ -40,7 +40,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "minitest", "~> 5.9.0"
   s.add_development_dependency "minitest-focus", "~> 1.1"
   s.add_development_dependency "minitest-reporters", "~>1.0"
-  s.add_development_dependency "racc", "~> 1.4"
+  s.add_development_dependency "racc", "1.4.14"
```
daemonsy: because we pin exact versions of `racc` in the rake task.

lib/graphql/language/parser.rb
```diff
@@ -1,6 +1,6 @@
 #
 # DO NOT MODIFY!!!!
-# This file is automatically generated by Racc 1.4.15
```
daemonsy: The rake task pinning might be outdated. Happy to change it to 1.4.15 everywhere (including docs).

rmosolgo: Oops, yeah, I'm assuming that newer is better, let's go for it!

rmosolgo:
Thanks for taking this on and making the case for it. I'm game to make the change, but I want to make sure, will this raise a Ruby error when people call `Schema.execute(...)` in their controllers? I want to make sure that these _client_ errors don't become noise in our bug trackers. 

On a related note, I see that a lot of tests about empty query strings and comments were removed altogether. Could we _keep_ those tests but update the assertion to describe this improved behavior? That would also help me answer my question above.

daemonsy:
> Thanks for taking this on and making the case for it. 
😅 actually I haven't explained it well. In general, my thought is that it's best to stay close to behavior of reference implementation / spec to avoid future problems where someone might be relying on non spec compliant behavior. 

Hence taking on this issue and the one where `{ node(id:) { id } }` is considered a valid query. Actually on their own, they are kind of a "trivial breaking change", so I don't mind it being deferred to a major release. 

> will this raise a Ruby error when people call Schema.execute(...) in their controllers? I want to make sure that these client errors don't become noise in our bug trackers.
Not so sure what you mean here, it raises an error if `GraphQL.parse` is called and becomes part of the errors key if an empty comment string is passed directly into `execute(...)`. 

> Could we keep those tests but update the assertion to describe this improved behavior? 
Good idea :+1: will do. I was wondering the context behind those tests in the first place, do you have recollection?

lib/graphql/language/parser.y
```diff
@@ -165,7 +165,6 @@ rule
 
   arguments_opt:
       /* none */                    { return EMPTY_ARRAY }
-    | LPAREN RPAREN                 { return EMPTY_ARRAY }
```
daemonsy: I believe this changes removes matching `field()` if I got the understanding of racc correctly.

spec/graphql/static_validation/rules/required_arguments_are_present_spec.rb
```diff
@@ -10,7 +10,7 @@
     }
 
     fragment cheeseFields on Cheese {
-      similarCheese() { __typename }
+      similarCheese { __typename }
```
daemonsy: I think static validation assumes that the query can be parsed, this test was failing because `context.visitor.visit` is failing. But if you took the whole query in this test and either parsed it or tried executing it, it raises the `ParseError` as expected.

rmosolgo:
> context behind those tests in the first place

I don't remember specifically, but in general, they serve as a _specification_ for behavior in those cases. If they weren't tested, then they'd basically be _undefined behavior_ and it would be very easy to break GraphQL-Ruby's behavior in those cases. By testing them, we can be sure to maintain the existing behavior, or if we _mean_ to change the behavior, we can keep a close eye on _how_ we change it. There are enough apps using this gem that the slightest unexpected change will surprise _someone_ 😆

daemonsy:
## Update: PR Writeup 2.0 

Please ignore the original write up. 

This is an attempt to resolve https://github.com/rmosolgo/graphql-ruby/issues/2243, https://github.com/rmosolgo/graphql-ruby/issues/1772, namely `GraphQL.parse` treating the queries below as valid 

All the "invalid" queries are tested with `graphql-js` and `libgraphqlparser` to verify that they are considered invalid. 

### Only comment in a query (a.k.a empty query)

```graphql 
# this is a comment 
```

### Empty argument set 
```graphql
{
  node() { name }
}
```
### Malformed argument 

```graphql 
{
  node(id:) { name } 
} 
```

## The changes
- raise `ParseError` with `EOF` as the other graphql parsers when the query is empty 
- Remove `()` grammar from argument sets 
- Remove implicit `/* none */`  with `|` in front of `literal_value` 👓

spec/graphql/analysis/ast_spec.rb
```diff
@@ -138,7 +138,7 @@ def foobar
     let(:analyzers) { [AstTypeCollector, AstNodeCounter] }
     let(:reduce_result) { GraphQL::Analysis::AST.analyze_query(query, analyzers) }
     let(:variables) { {} }
-    let(:query) { GraphQL::Query.new(Dummy::Schema, query_string, variables: variables) }
+    let(:query) { GraphQL::Query.new(Dummy::Schema.graphql_definition, query_string, variables: variables) }
```
daemonsy: :eyes:

spec/graphql/query_spec.rb
```diff
@@ -32,7 +32,7 @@
   let(:operation_name) { nil }
   let(:max_depth) { nil }
   let(:query_variables) { {"cheeseId" => 2} }
-  let(:schema) { Dummy::Schema }
+  let(:schema) { Dummy::Schema.graphql_definition }
```
daemonsy: :eyes:

daemonsy:
@rmosolgo I think this is ready for a second review. I'll also appreciate if you can build the parser locally and see if there's any diff, not sure if I got racc/ragel/colm working correctly. 

When it's all ready, maybe this can be squash merged because of a lot of small noisy intermediate commits.

spec/graphql/analysis/ast_spec.rb
```diff
@@ -148,10 +148,10 @@ def foobar
       }
     |}
 
-    describe "without a selected operation" do
+    describe "without a valid operation" do
       let(:query_string) {%|
         # A comment
-        # And nothing else
+        # is an invalid operation
         # Should not break
```
rmosolgo:
```suggestion
        # Should break
```

guides/development.md
```diff
@@ -176,12 +176,35 @@ The lexer and parser use a multistep build process:
...
 
 You will need Ragel to build the lexer (see above).
 
 If you start __guard__ (`bundle exec guard`), the `.rb` files will be rebuilt whenever the definition files are modified.
 
+#### Install Ragel and Colm on a Mac
+
+GraphQL Ruby requires Ragel 7.0.0.9 which is not available on Homebrew. To install it, you might have to download it from source.
+
+This is not meant to be a step by step guide and will likely not work as the documentation age.
```
rmosolgo:
```suggestion
This is not meant to be a step by step guide and will likely not work as the documentation ages.
```

rmosolgo:
Oops, I updated it to go into 1.10-dev but it seems confused because of the commits from master. Sorry about the noise, but I think the merge will be fine. (I tried merging master into `1.10-dev` but it seems like that didn't help.)

daemonsy:
I merged `1.10-dev` in and also pulled the remote changes. Only see my own changes, so 🤞 

If the build goes green, should be ready to merge

rmosolgo:
🤘 !

daemonsy:
:pray: thank you, I learned quite a bit (but only skimped the surface) about lexing and parsing in this PR. :smile:

PR #2345, opened 2019-06-22
Correct link to Getting Started from readme.md

robacarp:
Diving into the graphql-ruby library was a bit frustrating because the "Getting Started" link in the readme didn't go to the Getting Started page on the website, but to the home page. 

Eventually I found it, but maybe this will save others the trouble.


daemonsy:
Hey @robacarp thanks for the first contribution! :smiley:

Issue #1877, opened 2018-10-02
[pro] bidirectional_pagination has an off by one bug?


modosc:
(sorry i don't have code available, there's no specs that i could see for the pro stuff so i didn't know where to start with a reproduction.)

we've noticed an issue where paginating by N items at a time through N+1 items results in a weird error. consider we have 11 entries. 

this works as expected (
```
query { 
  entries(first: 10) { 
    pageInfo {
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
  }
}
// gives us 10 entries and the following pageInfo:
    "entries": {
      "pageInfo": {
        "hasPreviousPage": false,
        "startCursor": "startCursor",
        "hasNextPage": true,
        "endCursor": "endCursor"
      },

```
and this works as expected:
```
query { 
  entries(first: 10, after: "endCursor") { 
    pageInfo {
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
  }
}
// gives us 1 entry and the following pageInfo:
    "entries": {
      "pageInfo": {
        "hasPreviousPage": true,
        "startCursor": "lastCursor",
        "hasNextPage": true,
        "endCursor": "lastCursor"
      },

```
but when we go _backwards_ from here:
```
query { 
  entries(last: 10, before: "lastCursor") { 
    pageInfo {
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
  }
}
// gives us our original 10 entries but `hasNextPage` is incorrect
    "entries": {
      "pageInfo": {
        "hasPreviousPage": false,
        "startCursor": "startCursor",
        "hasNextPage": false, <-- this is wrong
        "endCursor": "endCursor"
      },
```

if we do the same thing with 12 entries `hasNextPage` works as expected, hence my off-by-one suspicion.

rmosolgo:
Ohh, interesting, so it's just `hasNextPage` showing symptoms of an off-by-one bug?  (Probably `hasPreviousPage` has the same issue if so)

modosc:
afaict `hasPreviousPage` is working correctly, it seems to only be `hasNextPage` and only when paginating backwards.

fwiw i only found this because i added a pagination integration spec where i created N+1 items and it broke - i spent a long time debugging my own code before realizing the spec worked if i created N+2 items

modosc:
i got excited about #2349 but it (unfortunately) doesn't solve this issue for us :/

rmosolgo:
[...]

PR #2349, opened 2019-06-25
Fix incorrect hasNextPage value when the last cursor belongs to the last element in the array

yask123:
### Background

The cursors for the items for ArrayConnection are generated by a 1 based index. (Index of origin is 1, not zero). 

However, when we compute the value for `has_next_page`, we were considering that it's a zero-based index. 

So it would return incorrect `hasNextPage` when the `before` cursor belongs to the last element int the array. 

I wrote a spec which would fail on the master branch because of this error. 

https://github.com/rmosolgo/graphql-ruby/blob/bb70bd88988d5cdf66dc7aea7427eca91274de04/spec/integration/rails/graphql/relay/array_connection_spec.rb#L293-L306

### Solution
When computing `has_next_page` value, consider the array to be based 1.

### DIFF
--- a/lib/graphql/relay/array_connection.rb
+++ b/lib/graphql/relay/array_connection.rb
@@ -13,7 +13,7 @@ module GraphQL
           sliced_nodes.count > first
         elsif GraphQL::Relay::ConnectionType.bidirectional_pagination && before
           # The original array is longer than the `before` index
-          index_from_cursor(before) < nodes.length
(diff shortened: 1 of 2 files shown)


yask123:
@rmosolgo for 👀

modosc:
ooh, i think this fixes #1877 ?

rmosolgo:
Thanks for the fix and the test!

PR #2353, opened 2019-06-26
Add `modules` option to install generator for modular file structure

joeytepp:
Add an option to the install generator so that a modular file structure is generated. Usage is as follows:

```bash
$ rails g graphql:install --modules
```

Which will create a file structure as follows:

```
     - app/
       - graphql/
         - resolvers/
         - types/
           - fields/
             - base_field.rb
           - enums/
             - base_enum.rb
           - inputs/
             - base_input_object.rb
           - interfaces/
             - base_interface.rb
           - objects/
             - base_object.rb
           - scalars/
             - base_scalar.rb
           - unions/
             - base_union.rb
           - query_type.rb
         - loaders/
         - mutations/
         - {app_name}_schema.rb
```

I think it would be nice to provide this options to users, and to my knowledge this shouldn't affect any of the other generators. Interested to hear others' thoughts!

### DIFF
--- a/guides/schema/generators.md
+++ b/guides/schema/generators.md
@@ -42,6 +42,7 @@ After installing you can see your new schema by:
 
 - `--relay` will add [Relay](https://facebook.github.io/relay/)-specific code to your schema
 - `--batch` will add [GraphQL::Batch](https://github.com/Shopify/graphql-batch) to your gemfile and include the setup in your schema
+- `--modules` will set up a modular folder structure for the new files
 - `--no-graphiql` will exclude `graphiql-rails` from the setup
 - `--schema=MySchemaName` will be used for naming the schema (default is `#{app_name}Schema`)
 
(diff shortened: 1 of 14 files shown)


rmosolgo:
Hi! Thanks for the suggestion. I'd love to hear what other folks think of this.

rmosolgo:
Thanks for the suggestion here, but I'm a bit torn. Folks are already upset that graphql-ruby doesn't offer a prescriptive folder structure for their schemas. I'm afraid it won't help to offer two _different_ folder structures -- won't it add to the confusion? 

If you're interested in taking ownership of this kind of schema design, I think that's great -- we could really use some clarity here. The problem for me is, I'm usually working on one big, old schema, so I don't have the opportunity to experiment with different setups. 

How about shipping this generator in a separate gem? You could document how it's meant to be used and why it's a good idea, and if folks like it, I'd gladly consider making it graphql-ruby's default.

rmosolgo:
> separate gem

Please do share if you get a chance to extract it!

PR #2362, opened 2019-07-03
Doc: use object to perform query with connections arguments

jean-francois-labbe:
I wanted to explicitly describe where to use the connections arguments


rmosolgo:
Thanks for the correction here!

PR #2369, opened 2019-07-08
Add method_access: false to suppress warning about argument name

### LINKED ISSUE #2358: Warning on upgrade to v1.9.7
The message:

```
Unable to define a helper for argument with name 'method' as this is a reserved name. If you're using an argument such as `argument method`, consider renaming this argument.
```

I'm not sure why this message appears.

rmosolgo:
Oops, we added a warning but no way to quiet it when it wasn't necessary. Fixes #2358

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -35,13 +35,14 @@ module GraphQL
 
   class Argument
     include GraphQL::Define::InstanceDefinable
-    accepts_definitions :name, :type, :description, :default_value, :as, :prepare
+    accepts_definitions :name, :type, :description, :default_value, :as, :prepare, :method_access
     attr_reader :default_value
     attr_accessor :description, :name, :as
     attr_accessor :ast_node
+    attr_accessor :method_access
     alias :graphql_name :name
 
-    ensure_defined(:name, :description, :default_value, :type=, :type, :as, :expose_as, :prepare)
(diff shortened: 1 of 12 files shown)


swalkinshaw:
I guess this would also be a first step into solving this issue for GraphQL clients (like graphql-client) which builds a schema from SDL?

The schema loader could take an option to set this property on arguments. Or probably even better just hardcode it to `false` in that context.

swalkinshaw:
Would you be opposed to doing that in this PR as well? It would be here I assume:

https://github.com/rmosolgo/graphql-ruby/blob/e9c6a480a135d5403b7b5b2b6e57eac3c3e338c7/lib/graphql/schema/loader.rb#L147-L152

rmosolgo:
Thanks for the suggestion! How does 3e7553201 look?

swalkinshaw:
🎉 thanks

jaredbeck:
It looks like the `method_access` feature was removed in 2.0.0. What is the recommended replacement?

rmosolgo:
👋  It doesn't look like there is one ... I'm not sure how it got dropped exactly. Could you please open a new issue with the warning you're seeing? Depending on the exact warning, my _guess_ is that it's coming from here: 

https://github.com/rmosolgo/graphql-ruby/blob/db11d0cc88f1d4ffafcbf421d62365176c4f2db0/lib/graphql/schema/input_object.rb#L154-L156

PR #2370, opened 2019-07-09
Add parameter_filter to arguments in backtrace

hiphapis:
Currently, parameter is print without `Rails.application.config.filter_parameters`.
So i did  modify this part

AS IS
```
Loc  | Field                | Object | Arguments                                    | Result
21:9 | Mutations.signInUser | nil    | {"email"="test@test.com", "password"="1234"} | ~~~
```
TO BE
```
Loc  | Field                | Object | Arguments                                          | Result
21:9 | Mutations.signInUser | nil    | {"email"="test@test.com", "password"="[FILTERED]"} | ~~~
```

### DIFF
--- a/lib/graphql/backtrace/table.rb
+++ b/lib/graphql/backtrace/table.rb
@@ -36,6 +36,13 @@ module GraphQL
 
       private
 
+      def parameter_filter(params)
+        return params if Rails.application.config.filter_parameters.blank?
+
+        parameter_filter = ActionDispatch::Http::ParameterFilter.new(Rails.application.config.filter_parameters)
+        parameter_filter.filter(params)
+      end
+
       def rows
         @rows ||= build_rows(@context, rows: [HEADERS], top: true)
       end
@@ -88,7 +95,7 @@ module GraphQL
             "#{position}",
             "#{field_name}#{field_alias ? " as #{field_alias}" : ""}",
             "#{ctx.object.inspect}",
-            ctx.irep_node.arguments.to_h.inspect,
+            parameter_filter(ctx.irep_node.arguments.to_h).inspect,
             Backtrace::InspectResult.inspect_result(top && @override_value ? @override_value : ctx.value),
           ]
 
@@ -108,7 +115,7 @@ module GraphQL
             "#{position}",
             "#{op_type}#{op_name ? " #{op_name}" : ""}",
             "#{query.root_value.inspect}",
-            query.variables.to_h.inspect,
+            parameter_filter(query.variables.to_h).inspect,
             Backtrace::InspectResult.inspect_result(query.context.value),
           ]
         else


lib/graphql/backtrace/table.rb
```diff
@@ -36,6 +36,13 @@ def to_backtrace
 
       private
 
+      def parameter_filter(params)
+        return params if Rails.application.config.filter_parameters.blank?
```
swalkinshaw: graphql-ruby doesn't have a dependence on Rails so `Rails` and `ActionDispatch` can't be referenced like this

hiphapis: Ah. I see
So you mean, need to check exists Rails or not?
Hm... I guess add some configuration is better. but `graphql-ruby` doesn't have a configuration. isn't it?

Which way is better?

swalkinshaw: It's not going to be easy for the reason you mentioned... You'd also want this behaviour to be optional as well since not everyone would want it filtered by default.

swalkinshaw:
I'm guessing this is because you use the backtrace feature in production? I'm assuming a lot of people only have it enabled in development (Shopify included).

So before you put in more work on this feature maybe @rmosolgo can chime in if its a feature he's interested in.

hiphapis:
@swalkinshaw Production mode also still using this backtrace
Production used when meet Fatal Error log, like this image
image

lib/graphql/backtrace/table.rb
hiphapis: Yes. right. so i suggest "put to configuration". then user can set filters with env.  but unfortunately, don't have a configuration

hiphapis:
I concerned about this. and I guess have 2 ways
First thing is using parameter of `GraphQL::Backtrace`
For example,
```ruby
filter_parameters = [:password]
GraphQL::Backtrace(filter_parameters: filter_parameters)
```
like this

Second thing is make and using `filter_parameters`
For example,
```ruby
TestSchema = GraphQL::Schema.define do
filter_parameters([:password])
```
like this.

How about this?

swalkinshaw:
The 2nd option isn't really specific to the backtrace feature. It could be renamed to be more specific at least (`backtrace_filter_parameters`) but it's still a global schema option for something that's designed as a plugin.

To me the first option makes more sense since it's a backtrace-only setting. However, wouldn't you still need to specify *what* is going to the filter the parameters? `GraphQL::Backtrace(filter_parameters: [:password])` still assumes it's using `ActionDispatch::Http::ParameterFilter` under the hood.

You could do:

```ruby
GraphQL::Backtrace(filter_parameters: [:password], filter_class: ActionDispatch::Http::ParameterFilter)
```

Basically anything that responds to `filter` could be used.

I should note the backtrace feature is still marked as "Experimental" as well so I'm not sure how @rmosolgo feels about adding features to it.

hiphapis:
@swalkinshaw I see. and I thought, this gem does not depend on `Rails` so need to make filter function manually as like' ActionDispatch::Http::ParameterFilter'.
If concept is decide then i will try to like that

rmosolgo:
It looks like this has stalled, so I'll close it. 

As @swalkinshaw mentioned, the backtrace helper is really only meant for development. IIRC it adds some nontrivial overhead to execution because of its bookkeeping.

PR #2372, opened 2019-07-09
Fix handling of invalid UTF-8 bytes

eapache:
If the string isn't even validly encoded bytes (let alone invalid
escapes, which is already being checked) then the regex itself raises an
ArgumentError which bubbles all the way up and leaves everything in a
bad state.

I had to put the string in a separate file because Ruby completely
refuses to run the test file if it contains a string literal with
encoding this messed up.

I manually updated the generated lexer file because homebrew doesn't have ragel 7 yet and I'm lazy, but I'll jump through those hoops if it turns out to be important.

@rmosolgo @swalkinshaw


rmosolgo:
Awesome, thanks for the fix here! Yeah, Ragel 7 is a bit tough to get going 😒

PR #2375, opened 2019-07-10
Make items type a list

bcardiff:
I think the field type should be a list. Am I missing something?


rmosolgo:
🤦‍♂ A typo on my part! Thanks for the fix.

rmosolgo:
BTW, I love your work on Crystal, thanks for it! 😊

PR #2376, opened 2019-07-10
Warn for field names that conflict; include suggestion for name override

### LINKED ISSUE #1320: Field named `context` is a bit weird in the class-based API
Honestly not 100% sure if this counts as a bug, but if you have a field named `context` in the new class-based API it resolves by default to the graphql context because of the `attr_reader` on `GraphQL::Schema::Object`.

Not too terrible to work around with `field :context, String, null: true, resolve: ->(obj, _args, _ctx) { obj.context }`.

Also I suppose this would apply to fields named `object`, `fields`, etc. Because the resolver prioritizes methods on the definition over methods on the underlying object. I wonder if that priority should be reversed for reserved words? Or it should look for `def __context` or something?

fyi @rmosolgo @swalkinshaw

### LINKED ISSUE #1648: [ Feature Request ] - Warn users if field is a reserved Ruby or Rails keyword
Hello,

I write this issue with one part pride for finally figuring out what was causing my mutation to fail in its return type, one part shame at how long it took to figure it out :smile:

Part of the domain of my app relies on displaying (or not displaying) something in the UI. I added a `display` column to the database and everything seemed to work okay testing via the console, but I kept getting a `cannot return null` error because the `display` property was marked as non-null, even though I was returning `{ display: true }` from the mutation to try and get the simplest test case possible.

Eventually, I looked up a list of reserved words in Rails and, lo and behold, `display` is one of those keywords. I'm not sure what it does or how it works, but somehow it was causing a conflict - as soon as I changed the property to `display_two` as a sanity check, everything worked just fine.

My proposal is that for any `field` name, we offer a warning if the field name is a reserved word in Ruby or Rails. We could use an opt-in mechanism for Rails because I know the library can be used outside of the Rails ecosystem (though I imagine a large portion of the user base is Rails-based). We wouldn't stop the user (as there may be legitimate use cases) but a warning along the lines of `Please be aware that field "display" is a reserved word in Rails. It may cause issues with field resolution. Please rename the field or whitelist this field name in the graphql initializers file.`

Feedback welcome, just trying to save someone else from a 3 hour debugging session for what comes down to poor naming choices :smile:

rmosolgo:
Help people do the right thing when a `field` conflicts with an important method name. 

Also, make it possible to dismiss the warning somehow :P 

Fixes #1648 
Fixes #1320

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -54,7 +54,7 @@ module GraphQL
         # @return [void]
         def add_field(field_defn)
           if CONFLICT_FIELD_NAMES.include?(field_defn.method_sym)
-            warn "#{self.graphql_name}'s `field :#{field_defn.name}` conflicts with a built-in method, use `method:` to pick a different resolver method for this field (for example, `method: :resolve_#{field_defn.method_sym}` and `def resolve_#{field_defn.method_sym}`)"
+            warn "#{self.graphql_name}'s `field :#{field_defn.name}` conflicts with a built-in method, use `method:` to pick a different resolver method for this field (for example, `resolver_method: :resolve_#{field_defn.method_sym}` and `def resolve_#{field_defn.method_sym}`)"
           end
           own_fields[field_defn.name] = field_defn
           nil
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -38,10 +38,24 @@ module GraphQL
           end
         end
 
+        # A list of field names that we should advise users to pick a different
+        # resolve method name.
+        #
+        # @api private
+        CONFLICT_FIELD_NAMES = Set.new([
+          # GraphQL-Ruby conflicts
+          :context, :object,
+          # Ruby built-ins conflicts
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/member/has_fields.rb
```diff
@@ -38,10 +38,24 @@ def get_field(field_name)
...
+          # Ruby built-ins conflicts
+          :method, :class
+        ])
+
         # Register this field with the class, overriding a previous one if needed.
         # @param field_defn [GraphQL::Schema::Field]
         # @return [void]
         def add_field(field_defn)
+          if CONFLICT_FIELD_NAMES.include?(field_defn.method_sym)
+            warn "#{self.graphql_name}'s `field :#{field_defn.name}` conflicts with a built-in method, use `method:` to pick a different resolver method for this field (for example, `method: :resolve_#{field_defn.method_sym}` and `def resolve_#{field_defn.method_sym}`)"
```
swalkinshaw: This should recommend using `resolver_method:` instead of `method:` shouldn't it?

rmosolgo: yes, thanks 😊

PR #2377, opened 2019-07-10
Fix another mishandling of unicode in the lexer

eapache:
@swalkinshaw @rmosolgo Found another one, though I'm a little less confident that my solution here is correct.

A string constant that is validly encoded initially, and also passes the `VALID_STRING` regex, can end up invalidly encoded after passing through `replace_escaped_characters_in_place`.

This PR just does a second check for `valid_encoding?` after that call, but potentially the real fix belongs in the `VALID_STRING` regex or the crazy mess that is `replace_escaped_characters_in_place`?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -62,7 +62,7 @@ ERR
   end
 
   assert_dependency_version("Ragel", "7.0.0.9", "ragel -v")
-  assert_dependency_version("Racc", "1.4.14", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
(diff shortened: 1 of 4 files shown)


rmosolgo:
Wow, thanks for adding the failing test here. I agree the replacement method might be a good place to look here, I'll try to take a closer look soon.

rmosolgo:
I looked for some better option to do the unicode replacement, but I couldn't find one. I thought `String#undump` was close, but it ends up giving a _Ruby_ expression, not a decoded string.

PR #2378, opened 2019-07-11
When using a resolver, don't re-authorize arguments

### LINKED ISSUE #2367: Pro version/Pundit integration/Mutation: How to get the object, not the mutation?
Hi,

I bought the pro version and set up pundit integration according to [this guide](https://graphql-ruby.org/authorization/pundit_integration#mutation-level-roles), but I'm having trouble getting mutations to work. Consider this mutation class:

```ruby
module Mutations
  class AcceptAttendee < Mutations::BaseMutation
    pundit_role :accept

    argument :id, ID, required: true # , loads: Types::AttendeeType
    field :attendee, Types::AttendeeType, null: true
    field :errors, [String], null: false

    def resolve(id:)
      attendee = Attendee.find(id)
      if attendee.accept
        {
          attendee: attendee,
          errors: []
        }
      else
        {
          attendee: nil,
          errors: attendee.errors.full_messages
        }
      end
    end

    def self.policy_class
      self.const_get(:Policy)
    end

    class Policy < ::ApplicationPolicy
      def accept?
        debugger
        true
      end
    end
  end
end
```

When I run this mutation, and the ruby process stops at the debugger line, I can inspect the record that is passed to the policy class. And much to my surprise, it's an instance of the `AcceptAttendee` mutation class:

```
(byebug) record
mutation AcceptAttendee
```

The mutation works as expected when I remove the debugger, but instead of passing `true`, I want to implement a real check whether the `current_user` has access to the `Attendee` record. So in this position, instead of an instance of the `AcceptAttendee` class, I need an instance of the `Attendee` class. Is that understandable? Am I doing something wrong here?

I also tried adding the `load` option to the `argument` macro, but this throws the following error:

```
NotImplementedError (MySchema.object_from_id(node_id, ctx) must be implemented to load by ID (tried to load from id `26713`)):
```

I'm not sure how to implement this, also I expected this to work out of the box given I have specified an `AttendeeType` that works well in queries (and in the mutation once I return `true` from the `accept?` method.

Anyway, I'm stuck and appreciate any help. Thank you in advance.

rmosolgo:
This code was moved from GraphQL-Pro to here because it's generally relevant, and besides, we can simplify `Field#authorized?` if we know that resolver classes do their own argument authorization. 

There's some chance this could be disruptive _if_: 

- Someone is using `loads:`
- Someone _doesn't want_ their loaded argument objects to be authorized automatically. (Previously they weren't, now they are.) 

I think this is a good improvement in any case, but maybe it could be pushed to 1.10 if it seems too disruptive. 

Fixes #2367 (which was worked around by the author)

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -420,7 +420,7 @@ module GraphQL
       def authorized?(object, context)
         if @resolver_class
           # The resolver will check itself during `resolve()`
-          true
+          @resolver_class.authorized?(object, context)
         else
           # Faster than `.any?`
           arguments.each_value do |arg|
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -418,13 +418,10 @@ module GraphQL
       end
 
       def authorized?(object, context)
-        self_auth = if @resolver_class
-          @resolver_class.authorized?(object, context)
-        else
+        if @resolver_class
+          # The resolver will check itself during `resolve()`
           true
-        end
-
-        if self_auth
+        else
           # Faster than `.any?`
           arguments.each_value do |arg|
             if !arg.authorized?(object, context)
@@ -432,8 +429,6 @@ module GraphQL
             end
           end
           true
-        else
-          false
         end
       end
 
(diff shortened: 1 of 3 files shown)


ralph:
For me, it's a good improvement. 👍

rmosolgo:
Worked in GitHub, so I'm going to put it in 1.9.x

Issue #2373, opened 2019-07-09
Parser chokes with lots of escaped quotes


rmosolgo:
There's an error in the parser when strings have too many quotes. Here's an example: 

```ruby 
GraphQL.parse('{ t(x: """"""X"""""")}')
Traceback (most recent call last):
       16: from /Users/rmosolgo/.rbenv/versions/2.6.2/bin/irb:23:in `<main>'
       15: from /Users/rmosolgo/.rbenv/versions/2.6.2/bin/irb:23:in `load'
       14: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
       13: from (irb):4
       12: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql.rb:17:in `parse'
       11: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql.rb:29:in `parse_with_racc'
       10: from parser.y:463:in `parse'
        9: from parser.y:448:in `parse_document'
        8: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/tracing.rb:111:in `trace'
        7: from parser.y:449:in `block in parse_document'
        6: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql.rb:34:in `scan'
        5: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql.rb:38:in `scan_with_ragel'
        4: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/language/lexer.rb:5:in `tokenize'
        3: from /Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/language/lexer.rb:592:in `run_lexer'
[...]

PR #2381, opened 2019-07-14
Parse empty block strings correctly

daemonsy:
## What's up?

This is an attempt to resolve https://github.com/rmosolgo/graphql-ruby/issues/2373. 

```ruby
GraphQL.parse('{ t(x: """"""X"""""")}') # NoMethodError (undefined method `empty?' for nil:NilClass)
```

## The Problem 

During the lexing phase, `""""""x""""""` will actually be read as: 

```ruby
GraphQL::Language::Lexer.tokenize('""""""x""""""')
# Because a set of 6 quotes is an empty block string. So we get two strings and an identifier. This is invalid grammar and will be rejected by the parser later.
=> [(STRING "" [1:1]), (IDENTIFIER "x" [1:7]), (STRING "" [1:8])] 
```

That brings us to the underlying problem.

```ruby
GraphQL::Language::Lexer.tokenize('"""""") # empty block string
NoMethodError: undefined method `empty?' for nil:NilClass
```
The fix is not to do extra processing when dealing with an empty block string.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -62,7 +62,7 @@ ERR
   end
 
   assert_dependency_version("Ragel", "7.0.0.9", "ragel -v")
-  assert_dependency_version("Racc", "1.4.14", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
+  assert_dependency_version("Racc", "1.4.15", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
 
   `rm -f lib/graphql/language/parser.rb lib/graphql/language/lexer.rb `
   `racc lib/graphql/language/parser.y -o lib/graphql/language/parser.rb`
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -1415,10 +1415,10 @@ UTF_8_ENCODING = "UTF-8"
 
 def self.emit_string(ts, te, meta, block:)
 quotes_length = block ? 3 : 1
-ts += quotes_length
-value = meta[:data][ts...te - quotes_length].pack(PACK_DIRECTIVE).force_encoding(UTF_8_ENCODING)
+content_range = (ts + quotes_length)...(te - quotes_length)
+value = meta[:data][content_range].pack(PACK_DIRECTIVE).force_encoding(UTF_8_ENCODING) || ''
 line_incr = 0
-if block
+if block && !value.length.zero?
 line_incr = value.count("\n")
 value = GraphQL::Language::BlockString.trim_whitespace(value)
 end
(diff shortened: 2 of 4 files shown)


Rakefile
```diff
@@ -62,7 +62,7 @@ ERR
   end
 
   assert_dependency_version("Ragel", "7.0.0.9", "ragel -v")
-  assert_dependency_version("Racc", "1.4.14", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
+  assert_dependency_version("Racc", "1.4.15", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
```
daemonsy: Hey @rmosolgo since this is on `1.10-dev`, I thought we could bring this over, don't have to keep switching versions.

lib/graphql/language/lexer.rl
```diff
@@ -198,10 +198,10 @@ module GraphQL
 
       def self.emit_string(ts, te, meta, block:)
         quotes_length = block ? 3 : 1
-        ts += quotes_length
```
daemonsy: This was changing `ts`, which seems to be used below for character positioning. In the character positioning case, I think we want it to be the original pointer?

spec/graphql/language/lexer_spec.rb
```diff
@@ -85,7 +92,7 @@
       assert_equal 8, str_token.col
       assert_equal '(STRING "c" [1:8])', str_token.inspect
       rparen_token = tokens[6]
-      assert_equal '(RPAREN ")" [1:10])', rparen_token.inspect
+      assert_equal '(RPAREN ")" [1:11])', rparen_token.inspect
```
daemonsy: Comment above: I think we were off by 1 all this time?

lib/graphql/language/lexer.rb
```diff
@@ -1415,10 +1415,10 @@ def self.emit(token_name, ts, te, meta)
 
 def self.emit_string(ts, te, meta, block:)
 quotes_length = block ? 3 : 1
-ts += quotes_length
-value = meta[:data][ts...te - quotes_length].pack(PACK_DIRECTIVE).force_encoding(UTF_8_ENCODING)
+content_range = (ts + quotes_length)...(te - quotes_length)
+value = meta[:data][content_range].pack(PACK_DIRECTIVE).force_encoding(UTF_8_ENCODING) || ''
```
daemonsy: This is assuming that having no value means we have an empty string. Not sure if it is the best approach.

spec/graphql/language/lexer_spec.rb
rmosolgo: Yes, I believe it 😆 Even here in the "calculates string position correctly" test.  I think we also used to miscalculate the _beginning_ position, but obviously I didn't think to consider the end position too.

rmosolgo:
Thanks for fixing this!

Issue #1051, opened 2017-10-23
Repackage GraphiQL for Ruby 


rmosolgo:
A long time ago, I made rmosolgo/graphiql-rails to use GraphiQL with Rails apps. I would like to repackage it as `GraphQL::GraphiQL` for a couple of reasons: 

- One less repo to maintain and document 
- Make it Rails-independent; package it as a Rack app instead (no Sprockets etc)
- Support subscriptions (eg https://github.com/rmosolgo/graphql-ruby/issues/1038#issuecomment-337508203)
- Consider [graphiql-workspace](https://github.com/OlegIlyenko/graphiql-workspace) instead of GraphiQL

jvanbaarsen:
@rmosolgo Is this something you want outside contributors to work on?

rmosolgo:
That would be great! Happy to help out in some way if I can, too. I just opened the issue to see if anyone was interested ... and if anyone was interested enough to _do it_ :P

jvanbaarsen:
@rmosolgo nice! I think I'll start with a very rough concept PR. From there we can look at what tests we want to include, and where I'm forgetting things. Does that sound ok?

rmosolgo:
Sounds great!

casiodk:
Support subscriptions 👍

daemonsy:
Hi, dredging up a very old issue 😸 

I made a PoC attempt to load GraphQL Playground directly as a very simple rack app. 

The hope is to get it to a point that most people starting with graphql-ruby can easily mount a dev only playground that has some default auth setup. 

It's one of the toughest hurdle in GraphQL adoption among devs. 

"Now I got the rails server running, what do I do next to explore this new API?"

connorshea:
This'd be great. GraphiQL has been receiving pretty regular updates lately. I've been considering trying to turn graphiql-rails into a Rails Engine that uses Webpacker, but this'd be much better.

I'd rather not swap GraphiQL out for GraphQL Playground if we don't have to.

ro-savage:
@rmosolgo - Any update on this? We recently upgraded to rails 6.0.2 and now GraphiQL has broken before of #2550

butsjoh:
We managed to get a graphiql running with actioncable for subscriptions compatibility using https://github.com/prisma-labs/graphql-playground. We will eventually publish a package for this but we currently do not have a lot of time to do the grunt work for it now but here are the main outlines of what we did to get it to work:

[...]

PR #2382, opened 2019-07-14
Add a PoC rack endpoint that mounts GraphQL playground

daemonsy:
## Use GraphQL Playground as a rack app [PoC] 

This is an attempt to tackle https://github.com/rmosolgo/graphql-ruby/issues/1051. 

### Why?

One of the hurdle of getting GraphQL adopted is getting devs to try out the new API in their development lifecycle. I hope to evolve this work to satisfy this use case: 

- Gets graphql-ruby working on a Rails / Ruby web app 
- Adds the endpoint to a dev only route that exposes some default auth headers etc 
- Other devs just start the server, visit `GET /graphql` and the client is loaded 
- Devs can contribute back to onboarding by adding better initialization defaults to the Playground (e.g. a common initial query) 

### PoC done the laziest way 
- Make a mountable class that has a single `.call` method 
- Return a HTML page that loads GraphQL Playground via the CDN 
- Options passed in on the route are passed to the JS initializer
- 🍷 

### Example on the dummy app 

```ruby
# In routes.rb
# frozen_string_literal: true

require 'graphql/endpoints/playground'

Rails.application.routes.draw do
  root to: "pages#show"

  mount GraphQL::Endpoints::Playground, at: '/graphql', playground: {
    endpoint: 'https://api.graph.cool/simple/v1/swapi'
  }
end
```

Screen Shot 2019-07-14 at 12 08 29 PM

### Choice of GraphQL Client
Have a decent experience using [GraphQL Playground](https://github.com/prisma/graphql-playground) from Prisma. Some things I consider quite useful: 

- Ability to configure headers (most of the time, needed for some auth mechanism). With this, some dev endpoints can be mounted with default users 
- Default query, variables can be configured
- Workspaces 
- I think it supports subscriptions
- Actively supported by a company, offers prebuilt version (it was a bit of a pain trying to build some clients with webpack)

###  Notes
- Maybe should lock down the version of Playground or allow it to be specified as an option on the route
- This approach requires internet all the time since it uses the CDN


rmosolgo:
👌 I think this is a great start! A few thoughts: 

- Personally, the CDN-based approach is fine to start with. If someone wants better offline support, it's easy to add later. 
- The `graphql:install` generator currently adds `graphiql-rails` to the project and mounts it. How about mounting this, instead? (Any reason that this can't take graphiql-rails's place?)
- What's all the styling in the layout? Is that part of graphql-playground? It looks ... fancy ... but also hard-to-maintain 😆

daemonsy:
> The graphql:install generator currently adds graphiql-rails to the project and mounts it. How about mounting this, instead? (Any reason that this can't take graphiql-rails's place?)

Yea I can this can take its place. I'd like to flesh out some basic doc examples of doing authentication to help people just started integrating GraphQL, then we'll definitely be in a better place. 

Mental note: I'll go review all the options available in GraphiQL rails 

> What's all the styling in the layout? Is that part of graphql-playground? It looks ... fancy ... but also hard-to-maintain 😆

That's given by the repo readme, I think most of it is just the loading animation. We could take it out.

JanStevens:
It seems that request includes the "fancy" loading animation which makes it look like there is a bunch of css in there. You can also go for a minimal style without loading animation. This is what we currently use

```ruby
<!DOCTYPE html>
<html>

<head>
  <meta charset="utf-8" />
  <meta name="viewport" content="user-scalable=no, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, minimal-ui">
  <title>GraphQL Playground</title>
  <link href="//fonts.googleapis.com/css?family=Open+Sans:300,400,600,700|Source+Code+Pro:400,700" rel="stylesheet">
  <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/css/index.css" />
  <link rel="shortcut icon" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/favicon.png" />
  <script src="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/js/middleware.js"></script>
</head>

<body>
<div id="root">
  <style>
    body {
      background-color: rgb(23, 42, 58);
      font-family: Open Sans, sans-serif;
      height: 90vh;
    }

    #root {
      height: 100%;
      width: 100%;
      display: flex;
      align-items: center;
      justify-content: center;
    }

    .loading {
      font-size: 32px;
      font-weight: 200;
      color: rgba(255, 255, 255, .6);
      margin-left: 20px;
    }

    img {
      width: 78px;
      height: 78px;
    }

    .title {
      font-weight: 400;
    }
  </style>
  <img src='//cdn.jsdelivr.net/npm/graphql-playground-react/build/logo.png' alt=''>
  <div class="loading"> Loading
    <span class="title">GraphQL Playground</span>
  </div>
</div>
<script type='text/javascript'>
  window.addEventListener('load', function (event) {
    GraphQLPlayground.init(document.getElementById('root'), {
      endpoint: '<%= polymorphic_path([:api, :graphql]) %>',
      settings: {
        'editor.theme': 'light',
        'request.credentials': 'include'
      }
    })
  })</script>
</body>
</html>
```

swalkinshaw:
After digging into graphql-playground (mostly via @RobertWSaunders) I'd be hesitant to use it by default instead of GraphiQL. While it has some nicer features, it has many issues. And GraphiQL is getting some renewed focus: https://github.com/graphql/graphql-wg/issues/136

RobertWSaunders:
Yep, I spent sometime digging around with graphql-playground, its a great tool and provides some nice features on top of the default graphiql however I found the code to be a bit messy and the dependencies haven't been updated for quite sometime. I created a fork of it to try and clean it up and add some additional features, whether or not I revisit it is to be decided but it wasn't a great developer experience working with it is all I can say.

daemonsy:
Hi @swalkinshaw and @RobertWSaunders the hope for this is to provide a simple client that works with basic dev needs, I imagine them to be: 

- Set headers (for authentication and any custom stuff)
- Explore the schema, docs 
- Execute queries 
- Remember last ran query in session (survives a browser restart)
- (nice to have) History of queries

Also, not expecting consumers to be able to extend the client at this point.

I picked GraphQL Playground because it seems modern and worked for me without much problem. If there are issues with it, I rather pick something simpler like Graphiql. 

Keen to know more details about some of the problems you have using it.

swalkinshaw:
Our use case was a bit different because we were doing customizations. If you're just embedding the stock graphql-playground then it's likely fine. And easy enough to switch back to GraphiQL at a future point if need be.

rmosolgo:
GraphQL playground is now available by the generator, which covers this case IMO: https://github.com/rmosolgo/graphql-ruby/pull/2839

Issue #1627, opened 2018-06-28
Schema class inheritance support


swalkinshaw:
Right now most of the new 1.8 classes support inheritance (and it's recommended), but `Schema` does not. This has come up once or twice on Slack and we also have a use case for it.

We have two different schemas which share some base functionality. These are mostly plugins and query analyzers.

One solution is just ensuring we've define them manually on both schemas, but we also end up creating schemas in unit tests where it would be nice to always have that guarantee of consistency.

This is on our radar but nothing immediate so wanted to open an issue.

rmosolgo:
:+1: definitely an oversight on my part, it'd be great to implement the necessary ancestor merging to make that work (joining arrays, merging hashes, etc)

rmosolgo:
Thanks for chiming in! I'm working hard to wrap up the 1.9.0 stable release, and I'll circle back to this in the next version.

jturkel:
@rmosolgo - We're increasingly needing this one too. I'd be happy to take a crack at it if you haven't had a chance to take a look yet.

rmosolgo:
👌 !

rmosolgo:
We may have _missed_ some spots, but the expected behavior is that inheriting schema classes Just Works :tm: starting in 1.9.8

PR #2384, opened 2019-07-15
Schema class inheritance support

jturkel:
This is a first stab at support for inheriting `GraphQL::Schema` settings (i.e. fixing #1627) e.g.

```ruby
class BaseSchema < GraphQL::Schema
  use GraphQL::Tracing::NewRelicTracing
  use GraphQL::Analysis::AST
  max_complexity 5000
end

class ServiceSchema < BaseSchema
  # ...
end

class AdminSchema < BaseSchema
  # ...
end
```

I took a pretty naive approach to the implementation and just leveraged the existing `find_inherited_method` method from`GraphQL::Schema::Member::BaseDSLMethods` to walk up the superclass chain looking for non-null setting values. It's a bit tedious and it's error prone as new settings are added but it works. There appears to be more generic frameworks for defining inheritable settings in `GraphQL::Schema::Member::AcceptsDefinition` and `GraphQL::Define::InstanceDefinable` and I briefly considered a simplified port of ActiveSupport's [class_attribute](https://api.rubyonrails.org/classes/Class.html#method-i-class_attribute) but I'm assuming these weren't leveraged here for a good reason? I still need to finish the tests but I was hoping to get some high level feedback on the approach before I spend too much more time on it. Thanks!

### DIFF
--- a/spec/graphql/analysis/analyze_query_spec.rb
+++ b/spec/graphql/analysis/analyze_query_spec.rb
@@ -31,6 +31,8 @@ describe GraphQL::Analysis do
     end
   end
 
+  let(:schema) { Class.new(Dummy::Schema) }
+
   describe ".analyze_query" do
     let(:node_counter) {
       ->(memo, visit_type, irep_node) {
@@ -43,7 +45,7 @@ describe GraphQL::Analysis do
     let(:analyzers) { [type_collector, node_counter] }
     let(:reduce_result) { GraphQL::Analysis.analyze_query(query, analyzers) }
     let(:variables) { {} }
-    let(:query) { GraphQL::Query.new(Dummy::Schema, query_string, variables: variables) }
+    let(:query) { GraphQL::Query.new(schema, query_string, variables: variables) }
     let(:query_string) {%|
       {
         cheese(id: 1) {
@@ -58,7 +60,7 @@ describe GraphQL::Analysis do
       let(:analyzers) { [type_collector, conditional_analyzer] }
 
       describe "when analyze? returns false" do
-        let(:query) { GraphQL::Query.new(Dummy::Schema, query_string, variables: variables, context: { analyze: false }) }
+        let(:query) { GraphQL::Query.new(schema, query_string, variables: variables, context: { analyze: false }) }
 
         it "does not run the analyzer" do
           # Only type_collector ran
@@ -67,7 +69,7 @@ describe GraphQL::Analysis do
       end
 
       describe "when analyze? returns true" do
-        let(:query) { GraphQL::Query.new(Dummy::Schema, query_string, variables: variables, context: { analyze: true }) }
+        let(:query) { GraphQL::Query.new(schema, query_string, variables: variables, context: { analyze: true }) }
 
         it "it runs the analyzer" do
           # Both analyzers ran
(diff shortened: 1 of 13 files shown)


rmosolgo:
:+1: This is looking great, thanks for taking a stab at it! 

If you want to add some DRY-er options for looking up inherited values, I'm totally open to it. I didn't work too hard on it because I was focusing on other things, but I agree it could be better. 

I'd rather stay away from the `.define`/`AcceptsDefinitions` stuff because that's all for backwards-compatibility, and I'd like to make a clean deletion of it in GraphQL-Ruby 2.0. We shouldn't add a dependency on ActiveSupport, but if you want to add some methods inspired by that, I'm open to it! 

Anything else you need from me to make this PR not a "draft"?

jturkel:
That's great context and the answer I was hoping for! Thanks. I'll work on getting this ready for a real review.

jturkel:
This should be ready for a few real review now. The biggest change since the draft was I changed the behavior of collection settings to lazily merge inherited values to make it consistent with other places like [GraphQL::Schema::Object#interfaces](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/object.rb#L90) e.g. 

```ruby
class BaseSchema < GraphQL::Schema
  
end

class ServiceSchema < BaseSchema
  rescue_from ServiceException, ->(err) { } 
end

BaseSchema.rescue_from StandardError, ->(err) { } 

ServiceSchema.rescues
# { StandardError => ..., ServiceException => ... }
``` 
As part of this I changed the `orphan_types` and `directives` methods to append values rather than replace them to make the behavior of calling them multiple times more consistent e.g. 

```ruby
class BaseSchema < GraphQL::Schema
  orphan_types Foo
end

class ServiceSchema < BaseSchema
  orphan_types Bar
end
```
now results in the same values for `ServiceSchema.orphan_types` as:

```
class ServiceSchema < GraphQL::Schema
  orphan_types Foo
  orphan_types Bar
end
```

A few other changes worth calling out since the WIP:

* I renamed `find_inherited_method` to `find_inherited_value` to make it clearer that these methods are returning values rather than a [Method](https://ruby-doc.org/core-2.5.3/Method.html). I also made `nil` the default for `default_value` since that is the most common behavior and hopefully a reasonable default. Neither of these changes is essential so I'm happy to roll them back if you're not a fan.
* I updated several of the tests to work on subclasses of `GraphQL::Schema` to avoid mutating global settings that had to be rolled back in `after` blocks.

I punted on trying to factor out some of the common patterns in dealing with these inheritable settings. I think it probably makes sense to wait to see how things shake out in 2.0 after the legacy classes are removed.

spec/graphql/analysis/analyze_query_spec.rb
```diff
@@ -119,14 +121,7 @@ def call(memo, visit_type, irep_node)
...
       before do
-        @previous_query_analyzers = Dummy::Schema.query_analyzers.dup
-        Dummy::Schema.query_analyzers.clear
-        Dummy::Schema.query_analyzers << variable_accessor
-      end
-
-      after do
-        Dummy::Schema.query_analyzers.clear
-        Dummy::Schema.query_analyzers.push(*@previous_query_analyzers)
+        schema.query_analyzer(variable_accessor)
```
rmosolgo: 👌

spec/graphql/analysis/ast/max_query_complexity_spec.rb
```diff
@@ -2,14 +2,7 @@
...
 describe GraphQL::Analysis::AST::MaxQueryComplexity do
-  before do
-    @prev_max_complexity = Dummy::Schema.max_complexity
-  end
-
-  after do
-    Dummy::Schema.max_complexity = @prev_max_complexity
-  end
-
+  let(:schema) { Class.new(Dummy::Schema) }
```
rmosolgo: 👌

PR #2386, opened 2019-07-17
Limit to 100 input validation errors

eapache:
Otherwise some very malformed variable sets and others oddities can cause
the validator to spend a long time generating a massive number of
errors, with a relatively small input.

@rmosolgo @swalkinshaw not super-happy with the implementation here, but it seems to do the trick. Happy to take another crack at this if a better approach can be recommended.


rmosolgo:
:+1: I'm game for the approach, but I'd rather make it configurable, at least at schema-level, in case someone actually _wants_ to return 101 errors! (Maybe they're expecting lots of input.) What do you think?

eapache:
OK, I let this percolate a bit. Things I don't like about this approach:
- using exceptions for what is basically flow-control
- the doubled logic between InputValidationResult and ValidationContext
- I'm pretty sure there are flows that use InputValidationResult which still don't rescue the new exception
- the delegate add_error method in BaseVisitor is kinda weird

eapache:
OK, pushed a new version which is generally much better. Only three real problems here, most of which are minor:
- the config code for `max_validation_errors` is super-duper-gross, I think I probably copy-pasta-ed the wrong things, but it seems to work as-is, so...?
- a single static validator in the pipeline can still exceed the max, since we only check between validators not after each error
- harder to test because of the previous point

lib/graphql/schema.rb
```diff
@@ -887,6 +889,14 @@ def max_complexity(max_complexity = nil)
         end
       end
 
+      def max_validation_errors(max_validation_errors = nil)
```
swalkinshaw: The usual pattern in this file is to call this `new_max_validation_errors`

swalkinshaw:
> the config code for max_validation_errors is super-duper-gross, I think I probably copy-pasta-ed the wrong things, but it seems to work as-is, so...?

That's all standard 🙈 and looks correct

This approach looks much better even given the caveats

lib/graphql/schema.rb
eapache: Hmm, I was copying this off of `max_complexity` which doesn't?

swalkinshaw: 🤷‍♂ the majority do but I guess there's inconsistencies already

lib/graphql/schema.rb
```diff
@@ -887,6 +889,14 @@ def max_complexity(max_complexity = nil)
...
+        if max_validation_errors
+          @max_validation_errors = max_validation_errors
+        else
+          @max_validation_errors || find_inherited_value(:max_validation_errors) || 100
```
jturkel: Minor but `find_inherited_value` takes a default value as an optional second argument.

```suggestion
          @max_validation_errors || find_inherited_value(:max_validation_errors, 100)
```

eapache: Hmm, didn't know that, thanks. I'm still trying to find a way to avoid duplicating the default value between here and the `initialize` method if you know how to do that.

jturkel: It's not very elegant but `Schema.to_graphql` could skip assigning `max_validation_errors` if it's null to avoid clobbering the defaulting done by the constructor i.e.

```ruby
schema_defn.max_validation_errors = max_validation_errors if max_validation_errors
```
Hopefully all of this gets cleaned up in 2.0 when the legacy config objects have been removed and there's a simple way to declare config options + their default values.

eapache: I ended up just extracting it into a constant as the more obvious thing. And did move it into `find_inherited_value` too.

lib/graphql/static_validation/validator.rb
```diff
@@ -36,6 +36,7 @@ def validate(query, validate: true)
             rules_to_use.each do |rule_class_or_module|
               if rule_class_or_module.method_defined?(:validate)
                 rule_class_or_module.new.validate(context)
+                break if context.errors.size >= @schema.max_validation_errors
```
rmosolgo: I _think_ calling `.validate(context)` _attaches_ some visitor hooks, but the hooks themselves don't run until `.visit` below. Are there any contexts where this condition is triggered?

lib/graphql/query/input_validation_result.rb
```diff
@@ -4,12 +4,20 @@ class Query
...
+        @problems = []
+      end
+
       def valid?
-        @problems.nil?
+        @problems.empty?
+      end
+
+      def too_many_errors?
+        @problems.size >= @schema.max_validation_errors
```
rmosolgo: IIRC, this object is initialized _for each input_ that's validated. Many validations don't use this at all. For example, `fieldsExistOnType` runs a lot of checks, but never uses this class. 

Is this change intended to cover validations like that, or only validations on provided inputs?

lib/graphql/static_validation/validator.rb
eapache: I dug into this a bit and I think this code might actually just be dead? As far as I can tell, none of the `GraphQL::StaticValidation::ALL_RULES` actually define a `validate` method at all anymore?

lib/graphql/query/input_validation_result.rb
eapache: I was intending the change to `lib/graphql/static_validation/validator.rb` to cover all the other validators, but I'm not sure it did.

eapache:
Rebased, removed the incorrect attempt at handling non-input validations, and clarified that this PR only address the input-validation side of the issue.

I'll think about approaches to cover non-input validations in a follow-up PR, but off the top of my head I don't see many options. There isn't a good way to break from the visitor when the limit is reached without raising an exception, but the visitor is called from so many different paths that I'm not sure it's correct or possible to rescue a new exception in all the right places.

rmosolgo:
Sorry I haven't written back here. We merged a similar kind of fix to our app. It made us think of other kinds of malicious input that could be provided, and I'd like to think a little bit more about it before I go ahead with this approach.

eapache:
For sure. Hit me up in a DM if you want to discuss more.

eapache:
Did anything ever come of this Robert?

rmosolgo:
No, sorry :(

PR #2389, opened 2019-07-19
Optimize complexity analysis memory usage

jturkel:
When profiling some common queries in our application I noticed a large amount of memory was being allocated by the analysis process. The graphql gem allocated 713,654 bytes during the execution of the query with the top 5 allocations by file being:

```
    257032  /Users/jturkel/.rvm/gems/ruby-2.6.3@dandelion/gems/graphql-1.9.6/lib/graphql/language/lexer.rb
    110094  /Users/jturkel/.rvm/gems/ruby-2.6.3@dandelion/gems/graphql-1.9.6/lib/graphql/analysis/ast/query_complexity.rb
     85112  /Users/jturkel/.rvm/gems/ruby-2.6.3@dandelion/gems/graphql-1.9.6/lib/graphql/language/nodes.rb
     79456  parser.y
     48400  /Users/jturkel/.rvm/gems/ruby-2.6.3@dandelion/gems/graphql-1.9.6/lib/graphql/static_validation/rules/fields_will_merge.rb
 ```

The worst offenders in lib/graphql/analysis/ast/query_complexity.rb were:
* [query_complexity.rb:96](https://github.com/rmosolgo/graphql-ruby/blob/v1.9.7/lib/graphql/analysis/ast/query_complexity.rb#L96) allocating 48,432 bytes via hashes
* [query_complexity.rb:71](https://github.com/rmosolgo/graphql-ruby/blob/v1.9.7/lib/graphql/analysis/ast/query_complexity.rb#L71) allocating 25,584 bytes via strings
* [query_complexity.rb:109](https://github.com/rmosolgo/graphql-ruby/blob/v1.9.7/lib/graphql/analysis/ast/query_complexity.rb#L109) allocating 12,295 bytes in hashes via the hashes defaulting function

The optimizations in this patch reduce the graphql gem allocations for this scenario down to 619,704 bytes (roughly 13%). The biggest win came from avoiding hash allocations altogether when a given node in the analysis can only have a single type. The other optimizations were just avoiding some array allocations.

An iterations/sec benchmark of my changes was faster but not statistically significant:

Before: 122.269  (±13.9%) i/s -    600.000  in   5.021482s
After: 127.342  (± 7.9%) i/s -    638.000  in   5.045393s

### DIFF
--- a/lib/graphql/analysis/ast/visitor.rb
+++ b/lib/graphql/analysis/ast/visitor.rb
@@ -60,6 +60,11 @@ module GraphQL
           @response_path.dup
         end
 
+        # @return String The a stringified path to the response key for the current field
+        def response_path_str(separator = '.')
+          @response_path.join(separator)
+        end
+
         # Visitor Hooks
 
         def on_operation_definition(node, parent)
--- a/lib/graphql/analysis/ast/visitor.rb
+++ b/lib/graphql/analysis/ast/visitor.rb
@@ -60,11 +60,6 @@ module GraphQL
           @response_path.dup
         end
 
-        # @return String The a stringified path to the response key for the current field
-        def response_path_str(separator = '.')
-          @response_path.join(separator)
-        end
-
         # Visitor Hooks
 
         def on_operation_definition(node, parent)
--- a/guides/development.md
+++ b/guides/development.md
@@ -176,7 +176,7 @@ The lexer and parser use a multistep build process:
 - Run the generator (Ragel or Racc) to create `.rb` files (`lexer.rb` or `parser.rb`)
 - `require` those `.rb` files in GraphQL-Ruby
 
-To update the lexer or parser, you should update their corresponding _definitions_ (`lexer.rl` or `parser.y`). Then, you can run `bundle exec build_parser` to re-generate the `.rb` files.
+To update the lexer or parser, you should update their corresponding _definitions_ (`lexer.rl` or `parser.y`). Then, you can run `bundle exec rake build_parser` to re-generate the `.rb` files.
 
 You will need Ragel to build the lexer (see above).
 
--- a/lib/graphql/analysis/ast/query_complexity.rb
+++ b/lib/graphql/analysis/ast/query_complexity.rb
@@ -4,11 +4,11 @@ module GraphQL
     # Calculate the complexity of a query, using {Field#complexity} values.
     module AST
       class QueryComplexity < Analyzer
-        # State for the query complexity calcuation:
+        # State for the query complexity calculation:
         # - `complexities_on_type` holds complexity scores for each type in an IRep node
         def initialize(query)
           super
-          @complexities_on_type = [TypeComplexity.new]
+          @complexities_on_type = [ConcreteTypeComplexity.new]
         end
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by jturkel** [comment]
Oops. My test wasn't setting the authentication context properly so evaluation short circuited and returned a nil for the root field. This makes the 93,950 bytes in savings from this optimization a much smaller percentage of the overall 2,587,856 bytes allocated during query execution by the gem (but still around 3.6%).

File: lib/graphql/analysis/ast/visitor.rb [hunk 1]
```diff
@@ -60,6 +60,11 @@ def response_path
           @response_path.dup
         end
 
+        # @return String The a stringified path to the response key for the current field
+        def response_path_str(separator = '.')
+          @response_path.join(separator)
+        end
```
rmosolgo: What's the rationale for this addition? Should it be added to the child class (`AST::QueryComplexity`) instead, since it's the only one that needs it?

jturkel: The `dup` in `Visitor#response_path` definitely makes sense from a safety perspective but results in an unfortunate array allocation when a client is just going to call `join('.')` on the result. This part of the optimization was a lot more impactful before I isolated the calling of `QueryComplexity#selection_key` to just abstract types. For my test case this part of the change now only saves 1,096 bytes so I don't think it's justified. I'll go ahead and roll this part back.

jturkel:
The top lexer allocations are from:
```
    177888  /Users/jturkel/.rvm/gems/ruby-2.6.3@dandelion/gems/graphql-1.9.6/lib/graphql/language/lexer.rb:1384
     78792  /Users/jturkel/.rvm/gems/ruby-2.6.3@dandelion/gems/graphql-1.9.6/lib/graphql/language/lexer.rb:1386
```
I haven't dug much but it looks like they're just the tokens from the query string. I'm not sure if there are any easy optimizations there. On a somewhat related note, have you ever looked into adding caching of parsed query ASTs?

rmosolgo:
Interesting: 

https://github.com/rmosolgo/graphql-ruby/blob/85554e8f1c9bfa499fe14a912b92de8c83d99932/lib/graphql/language/lexer.rb#L1384

https://github.com/rmosolgo/graphql-ruby/blob/85554e8f1c9bfa499fe14a912b92de8c83d99932/lib/graphql/language/lexer.rb#L1386

For that second one, depending _what's_ allocated, we might be able to improve it. I bet the `...` is making a range, and maybe there's a way we could do the same array slicing _without_ allocating a Range. 

> caching of parsed query ASTs

No, I haven't. I haven't really worked with a system where _parsing_ was a significant chunk of query execution, so it didn't seem worth it to me.

jturkel:
Good call on the ranges in the lexer! That's a quick win that eliminates another 12.8KB of memory allocation. I also updated the lexer to use frozen string literals while I was in there.

re: caching of parsed query ASTs - Parsing is only about 2.5% of query execution time so I don't think it would help latency much but the lexer + parser are allocating about 316KB of memory which might put pressure on GC that affects throughput. I'll leave that experiment for another day though :)

Issue #2390, opened 2019-07-19
Interface/Union PossibleTypes visibility


grcooper:
# Problem

I would like a way to change the visibility on types that Unions and Interfaces can be. 

```GraphQL
interface A {
 VALUE
}

type B implements A {
  VALUE
  FOO_VALUE
}

type C implements A {
  VALUE
  BAR_VALUE
}
```

There may be cases where we want to hide that type C implements A in some contexts of an API call, but still allow B to implement A for a different context. Under my current understanding, we can only control visibility of A, so either it is visible through the entire schema or not at all.

# Investigation 
This brought me to investigate how Warden handles Union and Interface visibilities. The common ground seemed to be https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/warden.rb#L173 Which goes into the PossibleTypes class.

Following along with how visibility is currently implemented and to have the correct knowledge of which types to show, we need context injected somewhere here. We have it in the initializer to get the visible interfaces and could store it on `PossibleTypes` to return the correct types.

## Questions
1. How do we know which types to allow in (where do we expose this)
2. Are there other places that possible types might be used
3. Does Union's definition of possible types need to change?

## Possible Solutions
1. This can follow along with how visibility is currently implemented by providing a method like `visible_types(context)` which returns a list of the available types based on the context passed in. By default as to not mess with the current implementation of interfaces, it would return all possible types, but could be overridden on other interfaces.

2. This I am not 100% sure of, but it seems that possible types are used in a few different places. The main place that I believe it will cause issues is in `get_possible_types` methods on interfaces and unions which checks the possible types. For those of you with more knowledge of the codebase know if there are cases where we do not have the context? Or if we need all of the known possible types and not just the visible ones?

3. Question 2 leads into this solution, I believe that the Union's possible_types need to change, and we can then filter out visible types there, which is surfaced in `PossibleTypes`: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/possible_types.rb#L28. We already seem to have the context when using `get_possible_type` so this might be a better location for this visibility check?

# Summary
I believe we can provide an interface similar to `visible?` but instead provide `visible_types` which only returns the types visible in the current context. This would likely live right on the InterfaceType and UnionType to be used by the `PossibleTypes` class.

Does this seem like a reasonable approach to this problem? Does this conflict with any current thoughts on how to change up how Warden works?

rmosolgo:
[...]

PR #2391, opened 2019-07-22
Adding possible_type visibility

grcooper:
From: https://github.com/rmosolgo/graphql-ruby/issues/2390

TL;DR there isn't a great way right now to add visibility to union/interface possible types.

Mainly posting this PR to get feedback on the approach. I would also like to add in Interface visibility to the same PR as I feel like they fit under the same category.

The main change is that I needed to introduce context throughout the codebase such that `possible_types` could eventually have it. There are some cases, such as in instrumentation (which comes from [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/traversal.rb#L204) ) which cannot get the context easily. 

My main question about this approach is if there are some parts of the code which need to be context agnostic and if I am breaking that assumption.

The other is if having all fields visible if no context is provided dangerous? ie, this could lead there to be holes in the code which cause types to leak through.


grcooper:
Seems like I implemented this for the old version of object types instead of thew new way as explained in the GraphQL slack. Going to take another approach to this using the new graphql schema types.

rmosolgo:
Sorry about that! I bet a lot of the changes will still apply, since generally speaking, the new-style Classes are used to generate the old-style objects.

spec/support/jazz.rb
```diff
@@ -319,6 +319,14 @@ def self.resolve_type(object, context)
         end
       end
     end
+
+    def self.filter_possible_types(types, ctx)
+      types = types.map { |type| GraphQL::BaseType.resolve_related_type(type) }
```
grcooper: This `GraphQL::BaseType.resolve_related_type(type)` is a bit of a pain, I think this is a code smell and should be fixed higher up so the client doesn't have to deal with this themselves. But leaving it here now as a proof of concept.

grcooper:
@rmosolgo Found some time to fix this up and get a better proof of concept with both the schema types and old types by passing some procs around like we do with `resolve_types`.

Let me know if you think this interface of `filter_possible_types` makes sense to you or if you had imagined another approach.

eapache:
I came into this review expecting to argue that more of the filtering logic should live in the warden, but I actually kinda like this approach? Curious for opinions from Scott and Robert.

lib/graphql/schema/union.rb
```diff
@@ -30,6 +31,14 @@ def to_graphql
         def kind
           GraphQL::TypeKinds::UNION
         end
+
+        # Filter possible type based on the current context, no-op by default
+        # @param types [Array<GraphQL::ObjectType>] Types to be filtered
+        # @param ctx [GraphQL::Query::Context] The context for the current query
+        # @param [Array<GraphQL::ObjectType>] the types remaining after the filter is applied
```
swalkinshaw: `@return`

lib/graphql/schema.rb
```diff
@@ -448,10 +448,11 @@ def type_from_ast(ast_node)
 
     # @see [GraphQL::Schema::Warden] Restricted access to members of a schema
     # @param type_defn [GraphQL::InterfaceType, GraphQL::UnionType] the type whose members you want to retrieve
+    # @param context [GraphQL::Query::Context] The context for the current query
     # @return [Array<GraphQL::ObjectType>] types which belong to `type_defn` in this schema
-    def possible_types(type_defn)
+    def possible_types(type_defn, context=nil)
```
swalkinshaw: spaces around `=`!

lib/graphql/schema/union.rb
```diff
@@ -5,13 +5,13 @@ class Union < GraphQL::Schema::Member
...
       class << self
-        def possible_types(*types)
+        def possible_types(*types, ctx: nil)
           if types.any?
             @possible_types = types
           else
             all_possible_types = @possible_types || []
             all_possible_types += super if defined?(super)
-            all_possible_types.uniq
+            ctx ? filter_possible_types(all_possible_types.uniq, ctx) : all_possible_types.uniq
```
swalkinshaw: 🤔 it might be better to borrow an existing pattern in the gem to create a `GraphQL::Query::NullContext` when theres no ctx and make the old method signature deprecated.

Especially here `ctx` isn't even used; so `filter_possible_types` can be called in either case with the same output.

swalkinshaw: This would apply in a few different places I guess.

grcooper: Yah I need a better spot to check for nil context, I didn't want to leave it up to the user. `NullContext` looks like it is what I want though, thanks!

lib/graphql/schema/warden.rb
```diff
@@ -170,7 +171,7 @@ def visible_abstract_type?(type_defn)
       end
 
       def visible_possible_types?(type_defn)
-        @schema.possible_types(type_defn).any? { |t| visible_type?(t) }
+        @schema.possible_types(type_defn, @context).any? { |t| visible_type?(t) }
```
rmosolgo: 🤔 unrelated, but I wonder if this method should call through to `#possible_types`, eg 

```suggestion
        possible_types(type_defn).any?
```

grcooper: Yah it seems like possible types already does what this function does, but memoized? I think we could just about remove `visible_possible_types`. The only question I have is iof there is a case where the context can change but the results still be stored in memory, so we get stale data?

The other option is to just make this a bit cleaner by calling `visible_possible_types?` from inside of `possible_types` so we don't have the duplicate code.

grcooper:
@rmosolgo I think I have everything I need to keep going, I would also like to look at doing interface visibility in a similar way, would you prefer these two things be in separate PR's or together?

lib/graphql/schema/union.rb
```diff
@@ -30,6 +31,14 @@ def to_graphql
         def kind
           GraphQL::TypeKinds::UNION
         end
+
+        # Filter possible type based on the current context, no-op by default
+        # @param types [Array<GraphQL::ObjectType>] Types to be filtered
+        # @param ctx [GraphQL::Query::Context] The context for the current query
+        # @param [Array<GraphQL::ObjectType>] the types remaining after the filter is applied
+        def filter_possible_types(types, ctx)
```
chrisbutcher: This default implementation receives but doesn't use `ctx`. Should it be `_ctx` in this case?

lib/graphql/union_type.rb
```diff
@@ -57,14 +57,16 @@ def possible_types=(new_possible_types)
...
+    def possible_types(ctx=nil)
       @clean_possible_types ||= begin
         if @dirty_possible_types.respond_to?(:map)
           @dirty_possible_types.map { |type| GraphQL::BaseType.resolve_related_type(type) }
         else
           @dirty_possible_types
         end
       end
+
+      ctx ? filter_possible_types(@clean_possible_types, ctx) : @clean_possible_types
```
chrisbutcher: `filter_possible_types` already has a similar conditional that defaults to returning `@clean_possible_types`. Not sure if it's preferred to keep this conditional here and make it more explicit what happens when there's no `ctx`, or just return the result of `filter_possible_types(@clean_possible_types, ctx)` here.

lib/graphql/schema/union.rb
grcooper: you are right it should be `_ctx` I originally had these as named args which I couldn't do, but will fix.

lib/graphql/union_type.rb
grcooper: Yah I removed the possiblity of a nil context in the update I am pushing. I am changing it to `NullContext` so it can still be operated on like a hash without erroring.

lib/graphql/query/null_context.rb
```diff
@@ -21,9 +21,13 @@ def initialize
         )
       end
 
+      def [](key); end
```
grcooper: Is this a reasonable thing to do? I feel like a null context should be able to be accessible here and just return nil? Same with the instance method?

spec/graphql/union_type_spec.rb
```diff
@@ -153,7 +179,8 @@
     it "copies possible types without affecting the original" do
       union.possible_types # load the internal cache
       union_2 = union.dup
-      union_2.possible_types << type_3
```
grcooper: This changed some functionality, we no longer return the instance method from `possible_types` and instead return a copy of an array. Is this ok? Being able to mutate the possible types in the first place seems a bit concerning regardless.

To fix this properly it requires me storing a hash of context to the resulting possible types, OR when possible types is called without context then it returns the instance variable as before, but if called with a context returns the copy.

The latter is probably more reasonable and still keeps backwards compatibility if needed.

lib/graphql/query/null_context.rb
grcooper: This follows along with the GraphQL Context interface so I think it is reasonable to give it similar functionality. `https://www.rubydoc.info/github/rmosolgo/graphql-ruby/GraphQL/Query/Context`

spec/graphql/union_type_spec.rb
rmosolgo: Yes, I agree that the original usage in this test is a hack!

grcooper:
Getting back to this, I remembered why I didn't go with that idea originally, it was because removing specific types gets complicated based on the namespacing and is why I had to do: https://github.com/rmosolgo/graphql-ruby/pull/2391/files#diff-c6259e983a70d8c2b588c156c4b7dd5fR43
This can be fixed however by making some `removeType(types, type)` function to deal with this.

The idea of a different method makes more sense to me of `possible_types_for(ctx)` but the issue is I want the rest of the files to use it as well: https://github.com/rmosolgo/graphql-ruby/pull/2391/files#diff-295a5500dcc6036944278a4283e3cc76R28
This means that whatever I build, needs to be able to be used by `possible_types` and so you get a cyclic dependancy if you use `possible_types` in `possible_types_for(ctx)`.

That cyclic dependency is the part that I don't know how to fix and why I went with the option to just create a list of types to remove.

I think that is enough reason to continue with the current approach? @rmosolgo

grcooper:
@rmosolgo BTW This is how I envision interfaces to be done similarly:

https://github.com/grcooper/graphql-ruby/pull/1

(Wasn't sure how to have a PR off a branch that hasn't been merged yet so I put it in my own repo)

I can include it in this PR if you think it's worthwhile to have both in one place.

rmosolgo:
Thanks for sharing your thoughts about the other design and how this one might work out with interfaces. It seems like it would play out well.

I'm still hung up a bit on the schema definition API. I really want something that's _familiar_, extensible, and easy enough to customize. Here's another possibility, which might even be possible _along with_ the API you've developed so far. 

What if we modeled union memberships and interface implementations with a first-class "join object", for example, `GraphQL::Schema::TypeMembership`. This membership could join the object type to the abstract type it's a part of. Then, users could extend that base class in their app (like with `BaseObject`) and define `def visible?` (like with arguments). Here's an imagined implementation: 

```ruby 
class Types::BaseMembership < GraphQL::Schema::TypeMembership 
  def initialize(private: false, **rest) 
    @private = private 
    super 
  end 

  # If a membership is marked `private`, then only reveal it to superusers 
  def visible?(ctx) 
    @private == false || ctx[:current_user]&.superuser?
  end 
end 

class Types::BaseUnion < GraphQL::Schema::Union 
  membership_class Types::BaseMembership 
end 

class Types::Searchable 
  possible_types Types::Post, Types::Comment 
  # Keep this secret while our image classifier is under development: 
  possible_types Types::Image, private: true
end 
```

In that example, options passed to `possible_types` (or `implements`) would be passed along to `initialize()` of the membership object.

Then we'd have an implementation like this: 

```ruby 
def possible_types(ctx = Query::NullContext)
  # Filter out the invisible memberships
  # then return the associated object types 
  type_memberships
    .select { |tm| tm.visible?(ctx) }
    .map(&:object_type) 
end 
```

(Which is why I think it's not totally separate from this implementation). 

grcooper:
I like the idea of some kind of included module. I'll play around with this idea and see if I can get something working. I think we need some kind of stronger link between the type and what is being applied to it. I believe currently `possible_types` is overwritten each time it's called. So we need to do almost some kind of append, that also stores the visibility data linked to each type.

One other thing for interfaces that was a bit weird when trying my approach on them is that the interface is implemented on the actual type, but the `filtered_possible_types` method is on the interface itself which ends up making them fairly highly coupled and I would like to try to get this implemented in such a way that you only need to remove the interface code in one place.

As you can tell I'm just intermittently working on this, but I appreciate the consistent feedback :)

grcooper:
@rmosolgo Fixed up as per your suggestions, I left it as a separate commit if you want me to go back to the old way, but I quite like this and think it provides a nice interface. However this required some big reworks to the `union_type` class. However, I have heard we are soon getting rid of that old way of doing things so I think its reasonable to take some liberties there to get it to work.

The new interface looks like:
```ruby
class MyTypeVisibility < GraphQL::Schema::TypeMembership
  def visible?(ctx)
    ctx[visibility]
  end
end

class MyFilteredUnion < GraphQL::Schema::Union
  type_visibility_class MyTypeVisibility

  possible_types FooType, BarType
  possible_types BazType, visibility: :private
end
```

Then we make a request with `ctx[:private] == true` which will allow the `BazType` to be visible.

Thoughts?

spec/graphql/union_type_spec.rb
```diff
@@ -14,11 +14,28 @@
...
+      return true if visibility.nil?
+
+      visibility[:private] && ctx[:private]
+    end
+  end
+
+  class FilteredUnion < GraphQL::UnionType 
+    type_visibility_class FilterType2
+
+    possible_types OpenStruct.new(kind: GraphQL::TypeKinds::OBJECT), visibility: { private: true }
```
grcooper: This was not letting me use `type_1` or `type_2` within the class, if there are thoughts on how to fix this let me know :D

eapache:
I'm not sure I understand the concern around removing types and the cyclic dependency? This join type class feels pretty gross to me.

grcooper:
> This join type class feels pretty gross to me.

Is there a reason it doesn't feel good @eapache ? It follows a similar way to define possible types, the only difference is adding a class to handle those possible type's visibility. That could get a bit messy as it is kind of magic though.


For cyclic dependancies, it's been a while since I looked at the old way I was doing things and might be able to take another stab at this suggestion: https://github.com/rmosolgo/graphql-ruby/pull/2391#pullrequestreview-288733342

It would be good for us to decide on an interface now that we have seen a couple of options.

eapache:
Yeah, just the use of a class for this feels like object overkill. I much prefer something akin to Robert's suggestion that you linked.

grcooper:
👍 I'll build a prototype for that version and see how it works.

grcooper:
https://github.com/rmosolgo/graphql-ruby/compare/master...grcooper:testing-possible-types-cyclic

Here I did the other option of overwriting possible types. The interface looks like:
```ruby
class MyUnion < GraphQL::Schema::Union
  possible_types MyType1 MyType2

  def self.possible_types(*types, ctx: GraphQL::Query::NullContext)
    types = super
    if ctx[:private]
      types = self.remove_possible_type(types, Type2)
    end
    types
  end
end
```

chrisbutcher:
I like the declarative style of option 2, and the extensibility that a dedicated class would offer, over an overridable/optional method.

swalkinshaw:
My opinion: I'm assuming in most use cases there's only a single implementation of "type visibility" (such as checking for a `visibility` property or some flag). If that's the case, I don't think it warrants its own feature like `type_visibility_class`. Overriding or defining a single method on the `BaseUnion` class would be enough? So I'd favour overriding `possibly_types` to keep it simple.

If it was more common to implement different type visibility implementations on many unions, then I'd favour  `type_visibility_class`.

eapache:
Generally with Scott on this one.

```ruby
def self.possible_types(*types, ctx: GraphQL::Query::NullContext)
    types = super
```
why is it an arg that gets ignored in favour of the return from `super`?

grcooper:
This is a weird behaviour because of how possible_types behaves in `GraphQL::Schema::Union` 

`def possible_types(*types, ctx: NullContext)` is the interface, and is used as both a setter and a getter, it behaves as a setter when types is not nil.

I should probably name them differently though.

rmosolgo:
> only a single implementation of "type visibility"

Unfortunately, this is not true for us at GitHub 😅 . We have a single schema with several modes: 

- external (`POST /graphql`) and internal (in-process GraphQL-Client) 
- Except, some _external_ traffic gets access to the internal schema (first-party clients) 
- And, some parts of the would-be-internal schema are accessible publicly when "preview headers" are present. 

So, using arbitrary code to make visibility checks pretty much required for us at this point. (It's what we do with "masking", anyhow.)

spec/graphql/union_type_spec.rb
rmosolgo: (I don't think it's possible -- those `let(...)` declarations create methods that are accessible inside the `it do ... end` blocks, but the methods aren't available anyplace else.)

rmosolgo:
Thanks again for all your work here, @grcooper. It seems like we're :+1: on test coverage, so that this point we're just bikeshedding 🍻  

I pushed a commit with a few touchups that I'd love y'all's thoughts on. Here are the notes from the commit message with a bit more commentary: 

- Make the type membership join more general-purpose:
  - Allow any options to `Schema::TypeMembership` (eg, this might be used to add directives to schema print-outs)
  - Rename `type_visibility_class` to `type_membership_class` to express that the class may hold any metadata
  - Also rename `type_visibilites` to `type_memberships`
  - Make a separate instance of `TypeMembership` for each possible type
  - Also pass the abstract type into `TypeMembership` instances; rename readers to :abstract_type and :object_type
- Performance tweaks:
 - Avoid creating multiple arrays when calculating `possible_types`
 -  Swap cache on ctx to only cache when the context is not provided, to avoid memory leak
    - Since type objects are long-lived, this cache would grow for each new `Query::Context` passed into that method, and it would grow quite large over time. I _expect_ that the warden's `.possible_types` methods would do this caching "for us", but we can still cache the default case. 
- Bug fixes 
  - Fix union_type_spec.rb where `type_1` and `type_2` were considered identical for `.uniq`
  - (This bug existed in the spec before, but adding `.uniq` exposed it) 
- `.define` "support" -- this is for the deprecated API. I don't think anyone will use it, but I just wanted to make sure it worked basically the same as other type configurations. 
  - Move type visibility methods on `GraphQL::UnionType` to support the deprecated .define API, update tests
- Interface consistency:
  - Use `context:` in public API instead of `ctx:`
  - (better to use the full name in my opinion)

What do you think of those changes? 

I still have a couple outstanding questions: 

- The use of `.uniq` suggests that we're _ok_ with types being added multiple times, is that right, or should we make sure that there are no duplicates in the configuration? (eg, we could check when types are added that they aren't already present.) It will work just fine the way it is, and either way suits me, but I wanted to confirm your intention here.
- In `Schema::Union`, `type_visibilities` are not inherited from the parent class. It's ok with me to leave this unimplemented and cross that bridge when we get there, but it seemed worth mentioning.

lib/graphql/schema/union.rb
```diff
@@ -5,31 +5,52 @@ class Union < GraphQL::Schema::Member
...
+            types.each do |t|
+              type_memberships << type_membership_class.new(self, t, options)
+            end
           else
-            all_possible_types = @possible_types || []
-            all_possible_types += super if defined?(super)
-            all_possible_types.uniq
+            visible_types = []
+            type_memberships.each do |tv|
+              if tv.visible?(context)
```
grcooper: This should be `tm` for `type_membership`?

lib/graphql/schema/union.rb
```diff
@@ -5,31 +5,52 @@ class Union < GraphQL::Schema::Member
...
-          type_defn.possible_types = possible_types
+          type_defn.type_memberships = type_memberships
           if respond_to?(:resolve_type)
             type_defn.resolve_type = method(:resolve_type)
           end
           type_defn.metadata[:type_class] = self
           type_defn
         end
 
+        def type_membership_class(visibility_class = nil)
```
grcooper: The input should be renamed to `membership_class`

lib/graphql/union_type.rb
```diff
@@ -24,47 +24,70 @@ module GraphQL
...
 
-    # @return [Array<GraphQL::ObjectType>] Types which may be found in this union
-    def possible_types
-      @clean_possible_types ||= begin
-        if @dirty_possible_types.respond_to?(:map)
-          @dirty_possible_types.map { |type| GraphQL::BaseType.resolve_related_type(type) }
-        else
-          @dirty_possible_types
-        end
+    def possible_types=(types_and_maybe_options)
```
grcooper: This feels a bit off to me, I really prefer specifying the params through named variables.

Otherwise, this means that the client needs to look at the method definition in order to know what is going on, rather than just looking at the input/return values.

I would prefer `possible_types=(types, options: {})`

lib/graphql/schema/union.rb
rmosolgo: Derp, yes, failed rename 😅

lib/graphql/union_type.rb
rmosolgo: Ahh my bad, this is left from my fiddling with `.define` support. I'll switch it to just `types`. I don't think you can pass keyword arguments to setter methods (can you?) and anyways, this is a rarely used methods, so I think defaulting to `{}` for options is fine.

rmosolgo:
Thanks for taking a look, @grcooper! I tried to take care of those issues you noticed. 

Sounds good about those `.uniq`s, I'll go ahead and remove them.

As for interfaces, either way is fine with me. If you'd like to get this merged and work from there, it's OK with me, or if you'd rather keep it pending until you're sure it will all work out (I _think_ it will!), that's also fine. Just let me know what you prefer.

grcooper:
I think it's worth shipping as-is I can add interfaces at a later time on top of this, usually better to do things in chunks.

lib/graphql/schema/union.rb
```diff
@@ -5,31 +5,51 @@ class Union < GraphQL::Schema::Member
       extend GraphQL::Schema::Member::AcceptsDefinition
 
       class << self
-        def possible_types(*types)
+        def possible_types(*types, context: GraphQL::Query::NullContext, **options)
```
eapache: I find this method confusing now, it seems to be doing two very different things in the two if/else branches?

spec/support/jazz.rb
```diff
@@ -307,8 +307,23 @@ class InspectableKey < BaseObject
     field :is_flat, Boolean, null: false, method: :flat
   end
 
+  class HideEnsemble < GraphQL::Schema::TypeMembership
```
eapache: `HideEnsemble` is a weird name for this class since it controls both hiding and showing? And because the `hide_ensemble` symbol used below doesn't actually map to this class name, which is confusing.

lib/graphql/schema/union.rb
swalkinshaw: Isn't it doing the same thing at a high-level? The two branches are:

1. setting the possible types
2. getting the possible types

That's still happening but with more logic now:

1. setting the possible types via type memberships
2. getting the possible types taking into account type membership visiblity

I do think it's overloaded but it's still consistent with how it always worked I'd argue?

eapache: 🤔 👍

spec/support/jazz.rb
rmosolgo: Oops, thanks for pointing that out. I renamed it to `PerformingActVisibility`

rmosolgo:
🎉 Thanks again for your work on this, @grcooper!

Issue #2388, opened 2019-07-18
When Schema with max_complexity set, multiplex queries with an abstract type get `NoMethodError` using AST analyzers


patvice:
## Background

When a schema has `max_complexity` set and to a multiplex query is called, `MaxQueryComplexity` analyzers is added to the multiplex_analyzers for a given execution. When the analyzing phase happens in `QueryComplexity#on_leave_field` checks to see if the parent node is an abstract type. When it is an abstract it calls `possible_types` on the query that was passed to the analyzer.

The problem is multiplex_analyzers are initialized with `GraphQL::Execution::Multiplex`, which doesn't have `possible_types` defined.

## Code Example
backtrace and code example are in this [gist](https://gist.github.com/patvice/29ac2c2af5615695bfd3650ecf8a1cb4)

## Purposed Solutions

It would be nice if there some way to check wether or not your analyzer is being used at a multiplex analyzer or just being used as a plan query analyzer. Would it make sense to define `multiplex?` method on `GraphQL::Analysis::AST::Analyzer` that checks the class of the `query` variable?

If `GraphQL::Analysis::AST::Analyzer` is not the right place for this fix, we could just add something to `QueryComplexity` that would run a different code path when working with `GraphQL::Execution::Multiplex`. Or `GraphQL::Execution::Multiplex` could implement `possible_types`.

Let me know what you think the best approach is, I would be happy to put up a patch.

rmosolgo:
Thanks for such a detailed bug report! That's really helpful. What a tricky issue. Here are some thoughts: 

- I _like_ having a single Analyzer class that works on either a multiplex or a query. (That is, it runs on multiplexes of size `1` _or greater_.) 
- But, we have to keep the lines clear about which objects go where -- for example, clear lifecycle of initialize, visit each query, check result. 

I'm not exactly sure the best solution, but I want to prevent _confusion_ over whether `query` is a Query or a Multiplex -- that could come back to bite us!

patvice:
I also like that analyzer can be used in both normal Queries and Multiplex queries > 1. Based on the comments in the class around multiplex analyzers maintaining state across all queries visited, I think that changing the analyzer to take in an array of `GraphQL::Query` objects instead of a single query object but the query instance variable in the class will represent current queries being visited. We can change the query variable to the new currently being visited query per iteration in the multiplex analyzers.

[...]

PR #2394, opened 2019-07-23
Updating multiplex analyzer to better match the functionality of query analyzer

patvice:
This PR is related to https://github.com/rmosolgo/graphql-ruby/issues/2388. 

In this PR the signature of the analyzer class will now take in the `GraphQL::Query` object along side the `GraphQL::Execution::Multiplex` object. Before this change when multiplex analyzers were initialized, the `GraphQL::Execution::Multiplex` was passed as the query variable. Now, when analyzing a multiplex query the current query being analyzed is updated in all multiplex analyzers, and the multiplex object is passed into the initialization of the class. This should more closely aline the functionality query & multiplex analyzers, as well as have a clear understand of what objects are being initialized.  

A change was also made to respect the`.analyze?` method in multiplex analyzers in the same way they currently work when being used as a query analyzer. Any multiplex analyzer will not run if `.analyze?` returns false to match the query analyzer behaviour. `GraphQL::Analysis::AST#analyze_multiplex` spec were also buffed up, as they seemed to sparse.

Look forward to your feedback, let me know if you would like any other changes made based on this one. 😄

### DIFF
--- a/lib/graphql/analysis/ast.rb
+++ b/lib/graphql/analysis/ast.rb
@@ -7,7 +7,6 @@ require "graphql/analysis/ast/max_query_complexity"
 require "graphql/analysis/ast/query_depth"
 require "graphql/analysis/ast/max_query_depth"
 
-# frozen_string_literal: true
 module GraphQL
   module Analysis
     module AST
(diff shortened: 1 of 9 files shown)


rmosolgo:
Thanks so much for digging into this fix! I'm sorry I haven't written back until now, but I have often reviewed the changes here and hemmed and hawed about the solution. 

The thing that bugged me was introducing mutable state inside the analyzers (`set_query`). I can't really provide much defense for my feeling except that mutable state gives me the heebie jeebies, and I try to avoid it when I can. 

It struck me that there's an alternative to getting the "current" query, namely, using `visitor.query`. That also allows us to leave the `initialize` signature in place. 

So, I've taken your excellent unit test improvements (🙇 !) and rolled them into #2613 which uses a different approach to solving this problem. I also included some doc updates in that PR to describe the intended behavior. 

Please feel free to take a look there and tell me if you see any shortcomings or have any questions. Thanks again for your work here, and sorry for the late reply! I'll plan to merge that alternative approach in a couple of days unless any issues pop up with it.

PR #2400, opened 2019-07-25
Memoize generated GraphQL query strings

epk:
Diff between the benchmarks with and without caching:

Query & validation:
```
--- a/no_cache
+++ b/cache
@@ -1,31 +1,31 @@
                query     2.000  i/100ms
-               query     22.086  (± 4.5%) i/s -    110.000  in   5.003548s
+               query     22.323  (± 4.5%) i/s -    112.000  in   5.036334s
 validate - introspection
-                        53.000  i/100ms
+                        56.000  i/100ms
 validate - abstract fragments
-                       108.000  i/100ms
+                       112.000  i/100ms
-                        59.000  i/100ms
+                        60.000  i/100ms
 validate - big query     6.000  i/100ms
 validate - introspection
-                        561.041  (± 6.8%) i/s -      2.809k in   5.030818s
+                        569.097  (± 6.7%) i/s -      2.856k in   5.041044s
 validate - abstract fragments
-                          1.097k (± 8.7%) i/s -      5.508k in   5.058356s
+                          1.137k (± 7.5%) i/s -      5.712k in   5.051612s
-                        560.746  (±10.9%) i/s -      2.773k in   5.007613s
-validate - big query     52.351  (±17.2%) i/s -    258.000  in   5.097926s
+                        610.578  (± 8.0%) i/s -      3.060k in   5.045672s
+validate - big query     63.452  (± 6.3%) i/s -    318.000  in   5.036684s
```

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -73,7 +73,11 @@ module GraphQL
         end
 
         def to_query_string(printer: GraphQL::Language::Printer.new)
-          @query_string ||= printer.print(self)
+          if printer.is_a?(GraphQL::Language::Printer)
+            @query_string ||= printer.print(self)
+          else
+            printer.print(self)
+          end
         end
 
         # This creates a copy of `self`, with `new_options` applied.
@@ -83,7 +87,9 @@ module GraphQL
           copied_self = dup
           new_options.each do |key, value|
             copied_self.instance_variable_set(:"@#{key}", value)
-            copied_self.instance_variable_set(:@query_string, nil)
+            if copied_self.instance_variable_defined?(:@query_string)
+              copied_self.instance_variable_set(:@query_string, nil)
+            end
           end
           copied_self
         end
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -73,7 +73,7 @@ module GraphQL
         end
 
         def to_query_string(printer: GraphQL::Language::Printer.new)
-          printer.print(self)
+          @query_string ||= printer.print(self)
         end
 
         # This creates a copy of `self`, with `new_options` applied.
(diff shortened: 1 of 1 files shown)


lib/graphql/language/nodes.rb
```diff
@@ -73,7 +73,7 @@ def position
         end
 
         def to_query_string(printer: GraphQL::Language::Printer.new)
-          printer.print(self)
+          @query_string ||= printer.print(self)
```
rmosolgo: what if a different `printer:` is passed in here? 

I think we could get the lion's share of the benefit with something like 

```ruby
if printer.class = GraphQL::Language::Printer 
  @query_string ||= ... 
else 
  ... 
end 
```

does that seem appropriate?

epk: Yes, that makes sense!
But it's also hard to reason why a different printer would be returning anything different if an AST node has not changed.

epk: Oh, now I realize what the problem is:
Some could pass different instances of printer to this method

ericyd:
I believe this broke the GraphQL Client library that depends on this. I opened an issue here: https://github.com/github/graphql-client/issues/216

bai:
> I believe this broke the GraphQL Client library that depends on this.

Can confirm, github graphql client is borked.

rmosolgo:
D'oh! Thanks for sharing that. I think it's because of graphql-client's `deep_freeze` behavior. That was intended to prevent AST manipulation after parsing. Since 1.9, graphql-ruby's ASTs have been read-only, so I think we can remove that upstream. I'll take a look!

rmosolgo:
Should be fixed in graphql-client 0.15.0, please let me know if you have trouble with that version!

PR #2401, opened 2019-07-25
Avoid recomputation of field keys during tracing

jturkel:
Profiling of some of our queries showed we were spending a decent amount of time and allocating a fair amount of memory generating field trace keys. Caching this information within the scope of a query execution shaved 5.8% off of our query latency and saved 1MB in memory allocations.


rmosolgo:
Nice! I never considered caching _on the query_ that way, but I dig it!

PR #2404, opened 2019-07-29
Fix Int/Float typecast issue.

### LINKED ISSUE #2259: Type coercion truncates Float to Int instead of raising an error
Hi guys; I'm pretty new to graphQL; I noticed it automatically coerces my type-float mutation arguments into type-int by truncation without raising any error. That seems to contradict what it says in the specs:

From https://graphql.github.io/graphql-spec/June2018/#sec-Int
> Input Coercion
> When expected as an input type, only integer input values are accepted. All other input values, including strings with numeric content, must raise a query error indicating an incorrect type.

in https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/int.rb#L13
>      def self.coerce_input(value, _ctx)
>        value.is_a?(Numeric) ? value.to_i : nil
>      end

should it not be
> value.is_a?(Integer) ? value.to_i : nil

instead?

cherbst-2112:
Closes #2259

Raise an error if an Integer argument receives a Float.

### DIFF
--- a/lib/graphql/types/int.rb
+++ b/lib/graphql/types/int.rb
@@ -10,7 +10,7 @@ module GraphQL
       MAX = (2**31) - 1
 
       def self.coerce_input(value, _ctx)
-        value.is_a?(Numeric) ? value.to_i : nil
+        value.is_a?(Integer) ? value : nil
       end
 
       def self.coerce_result(value, ctx)
--- a/spec/graphql/int_type_spec.rb
+++ b/spec/graphql/int_type_spec.rb
@@ -3,14 +3,14 @@ require "spec_helper"
 
 describe GraphQL::INT_TYPE do
   describe "coerce_input" do
-    it "accepts ints and floats" do
+    it "accepts ints" do
       assert_equal 1, GraphQL::INT_TYPE.coerce_isolated_input(1)
-      assert_equal 6, GraphQL::INT_TYPE.coerce_isolated_input(6.1)
     end
 
     it "rejects other types" do
       assert_nil GraphQL::INT_TYPE.coerce_isolated_input("55")
       assert_nil GraphQL::INT_TYPE.coerce_isolated_input(true)
+      assert_nil GraphQL::INT_TYPE.coerce_isolated_input(6.1)
     end
 
     describe "handling boundaries" do
--- a/spec/graphql/query_spec.rb
+++ b/spec/graphql/query_spec.rb
@@ -418,14 +418,6 @@ describe GraphQL::Query do
       }
     |}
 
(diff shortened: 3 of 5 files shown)


cherbst-2112:
n/m, actually looks like that may be by design:

```

    describe "scalar inputs" do
      let(:query_string) {%|
        {
          cheese(id: 1.4) {
            flavor
          }
        }
      |}

      it "converts them to the correct type" do
        cheese_name = result["data"]["cheese"]["flavor"]
        assert_equal("Brie", cheese_name)
      end
    end
  end
```

rmosolgo:
Personally, I think youre right about that. Here's the spec: 

> _Input Coercion_
>
> When expected as an input type, only integer input values are accepted. All other input values, including strings with numeric content, must raise a query error indicating an incorrect type.

https://graphql.github.io/graphql-spec/June2018/#sec-Int

Doesn't it sound like your change is correct?

cherbst-2112:
Huh, OK, noticed the affected tests were mostly typecasting IDs as floats which *really* doesn't make sense anyway, they should be ID/String types.

I updated the affected tests, let me know what you think!

lib/graphql/types/int.rb
```diff
@@ -10,7 +10,7 @@ class Int < GraphQL::Schema::Scalar
       MAX = (2**31) - 1
 
       def self.coerce_input(value, _ctx)
-        value.is_a?(Numeric) ? value.to_i : nil
+        value.is_a?(Integer) ? value.to_i : nil
```
rmosolgo: Do we still need to call `.to_i` in the case that it's an Integer? Or could it be: 

```suggestion
        value.is_a?(Integer) ? value : nil
```

cherbst-2112:
Thanks! I've made that change, but I'm getting some unrelated network errors in travis

```
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/apollo-link: ETIMEDOUT".
info If you think this is a bug, please open a bug report with the information provided in "/home/travis/build/rmosolgo/graphql-ruby/javascript_client/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
yarn run v1.15.2
$ jest
/bin/sh: 1: jest: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'll try again later or tomorrow

rmosolgo:
I've been seeing that too -- It seems like there's some network-level (caching?) issue. I re-ran the failed runs and they passed. 

thanks for your work on this!

Issue #2301, opened 2019-05-22
NoMethodError: undefined method `ast_node' when trying to print schema


mdesantis:
Hello :)

I'm running into the error below when I try to print the following schema: https://github.com/solidusio-contrib/solidus_graphql_api/blob/55d1469c1a692ae2d9973b08f5755f31c458d5da/lib/spree/graphql/schema.rb

```
Traceback (most recent call last):
	18: from bin/rails:16:in `<main>'
	17: from bin/rails:16:in `require'
	16: from /gems/railties-5.2.3/lib/rails/engine/commands.rb:9:in `<top (required)>'
	15: from /gems/railties-5.2.3/lib/rails/engine/commands.rb:9:in `require'
	14: from /gems/railties-5.2.3/lib/rails/commands.rb:18:in `<top (required)>'
	13: from /gems/railties-5.2.3/lib/rails/command.rb:46:in `invoke'
	12: from /gems/railties-5.2.3/lib/rails/command/base.rb:65:in `perform'
	11: from /gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
	10: from /gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
	 9: from /gems/thor-0.20.3/lib/thor/command.rb:27:in `run'
	 8: from /gems/railties-5.2.3/lib/rails/commands/runner/runner_command.rb:38:in `perform'
	 7: from /gems/railties-5.2.3/lib/rails/commands/runner/runner_command.rb:38:in `load'
	 6: from print_schema.rb:1:in `<top (required)>'
	 5: from /gems/graphql-1.9.4/lib/graphql/schema/printer.rb:82:in `print_schema'
	 4: from /gems/graphql-1.9.4/lib/graphql/schema/printer.rb:82:in `new'
	 3: from /gems/graphql-1.9.4/lib/graphql/schema/printer.rb:56:in `initialize'
	 2: from /gems/graphql-1.9.4/lib/graphql/language/document_from_schema_definition.rb:35:in `document'
	 1: from /gems/graphql-1.9.4/lib/graphql/language/document_from_schema_definition.rb:251:in `build_definition_nodes'
/gems/graphql-1.9.4/lib/graphql/language/document_from_schema_definition.rb:46:in `build_schema_node': undefined method `ast_node' for Spree::GraphQL::Schema:Class (NoMethodError)
```

**Reproduction steps**

[...]

PR #2406, opened 2019-07-30
fix printing schema classes

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/2301

This will be fixed better in 1.10 when the class has its own AST node since it's built directly.


rmosolgo:
Oops, the failing test is a broken build for the integration tests.

PR #2409, opened 2019-07-31
Fix disabling of max_complexity and max_depth during query execution

jturkel:
The [doc](https://graphql-ruby.org/queries/complexity_and_depth) states that passing an explicit `max_complexity` or `max_depth` of nil will disable any schema level settings for `max_complexity` or `max_depth` but that wasn't working. This should fix that :)

I also fixed a problem with my changes in #2384 invoking `GraphQL::Schema.max_depth=(new_value)` rather than `GraphQL::Schema.max_depth(new_value = nil)` and a similar problem for `max_complexity`.

### DIFF
--- a/spec/graphql/analysis/max_query_depth_spec.rb
+++ b/spec/graphql/analysis/max_query_depth_spec.rb
@@ -36,9 +36,17 @@ describe GraphQL::Analysis::MaxQueryDepth do
     end
   end
 
+  describe "when the query specifies a nil max_depth" do
+    let(:result) { schema.execute(query_string, max_depth: nil) }
+
+    it "obeys that max_depth" do
+      assert_nil result["errors"]
+    end
+  end
+
   describe "When the query is not deeper than max_depth" do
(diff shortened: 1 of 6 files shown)


spec/graphql/analysis/max_query_depth_spec.rb
```diff
@@ -48,7 +56,8 @@
 
   describe "when the max depth isn't set" do
     before do
-      schema.max_depth = nil
+      # Yuck - Can't override GraphQL::Schema.max_depth to return nil if it has already been set
+      schema.define_singleton_method(:max_depth) { |*| nil }
```
rmosolgo: Oops! If we need this for something other than test setup one day, we can do something like 

```
def max_depth(new_max_depth = :unset) 
  if new_max_depth == :unset # Nothing was given 
    # read `@max_depth` with inheritance 
  else 
    @max_depth = new_max_depth 
  end 
end 
```

PR #2410, opened 2019-07-31
Just a couple clean-ups wrt ivars

tenderlove:
I like it better when objects don't set internal state on other objects, so in these patches I've moved around the instance variable manipulation so that it's done inside the object rather than being done from the outside.


rmosolgo:
Sounds good, thanks!

PR #2414, opened 2019-08-05
Fix incomplete docs in comments

ghiculescu:
The comments should match the instructions at https://graphql-ruby.org/queries/ast_analysis.html


rmosolgo:
Oops, thanks for correcting that!

PR #2415, opened 2019-08-06
Support --persist-output in sync cli

### LINKED ISSUE #2412: Persisted queries with Relay and sync
It seems that the implementation for deriving persisted queries from Relay graphql.js files may now be broken, as Relay queries no longer embed a relayHash in their artefacts.

The default export of *.graphql.js files does have a `node` property on it which I think is analogous to relayHash, but there's one wrinkle: `relay-compiler` supports language plugins which can change its output (the most commonly used one is Typescript artefacts with https://github.com/relay-tools/relay-compiler-language-typescript). But theoretically, they could be any format, so graphql-ruby shouldn't be able to make any kind of assumption about what they are at all.

Fortunately there does seem to be a more direct way: when relay-compiler is passed the `--persist-output` flag, it also generates a JSON file mapping IDs to queries. I think it would be much better to make `sync` use this JSON, though this may require changing the names of some argument to make it make sense!

For convenience, here's the link to Relay's current documentation regarding persisted queries: https://relay.dev/docs/en/persisted-queries

rmosolgo:
Earlier this year, Relay added first-class support for persisted query servers 🎉  So let's update `sync` to consume the artifact it generates. 

cc @sgwilym , how does this look to you? 

Fixes #2412


sgwilym:
This looks great! Thank you for the rapid response to this, you're a 🌟

rmosolgo:
🚢  in 1.6.6! Please let me know how it goes if you get to try it out 🍻

PR #2416, opened 2019-08-06
Support required arguments with default values [fixes #2356]

### LINKED ISSUE #2356: Support required arguments with default values
The GraphQL spec has [for some time supported required arguments with default values](https://github.com/graphql/graphql-js/pull/1274), however graphql-ruby (1.9) appears not to support this

```ruby
argument :id, ID, required: true, default_value: "123"
```
results in the schema compilation error:
`Query is invalid: field "account" argument "id" Variable id of type "ID!" is required and will not use the default value. Perhaps you meant to use type "ID".`

This seems like a very legitimate use case to me, a less contrived example:
```ruby
argument :hideDisabledAccounts, Boolean, required: true, default_value: false
# vs
argument :hideDisabledAccounts, Boolean, required: false
```
As Rubyists we understand that nil is false-y, so the behaviour here is roughly the same, but from the outside (and the inside) the 1st definition's behaviour seems to me much clearer than the 2nd, in the case where no value is provided.

References:
[This behaviour being requested in GraphQL reference implementation, and delivered](https://github.com/graphql/graphql-js/issues/1405)
[Merged spec change which adds "Non-null arguments and input object fields are no longer required to be provided a value/variable if that argument also supplies a default value."](https://github.com/graphql/graphql-spec/pull/418)

modosc:


modosc:
it looks like the failing specs are just order changes in the serialized hash:
```
-"{:required_with_default_arg=>1, :renamed=>\"x\"}"
+"{:renamed=>\"x\", :required_with_default_arg=>1}"
```
and they only happen on one version of ruby (although maybe they happen in others, given enough test runs). should i change those specs to deserialize so i can compare hash to hash or just not worry about it?

spec/graphql/schema/argument_spec.rb
```diff
@@ -144,9 +144,39 @@ class Schema < GraphQL::Schema
...
+    it 'uses provided input value' do
+      query_str = <<-GRAPHQL
+      { field(requiredWithDefaultArg: 2) }
+      GRAPHQL
+
+      res = SchemaArgumentTest::Schema.execute(query_str)
+      assert_equal '{:required_with_default_arg=>2}', res["data"]["field"]
+    end
+
+    it 'respects non-null type' do
```
rmosolgo: Ok, I think I _finally_ get why this is _good_!

modosc:
> Regarding the failing test, I think it's because that run is with TESTING_INTERPRETER=1. With that env var, it uses the new interpreter runtime, which probably builds argument hashes in a different order.

ok, specs fixed and passing

rmosolgo:
🤘

modosc:
thanks @rmosolgo , any idea when 1.9.10 will land?

Issue #2034, opened 2019-01-09
Schema initialization happens on execution after migration to class-based API


philipkocanda:
After converting our Schema to the class-based API, the initialization (loading of instrumentation, etc.) happens on the first call to `Schema.execute`, whereas it used to happen on application initialization. Since HackerOne has a big schema with a lot of instrumentation, this can take some time; and it caused whichever spec happened to be the first to run to be be slowed down by ~15 seconds, causing build failures in the process.

To work around this we're currently calling `Schema.instrumenters` manually while loading the app, which shaves ~7 seconds off of the first call to `Schema.execute`.

We're open to any suggestions on how to improve the situation!

rmosolgo:
Hey! You've basically found the best way to address it, by forcing the schema loading some way. Another way is calling `MySchema.graphql_definition` at the bottom of the file, which will build out all the caches, etc. 

I'm hoping to eliminate the all-at-once initialization in a future release (1.10?) by building those caches incrementally, but I still have some things to finish up for 1.9 😅

philipkocanda:
Thanks Robert! That's super helpful, looking forward to >1.9 😄

fgrehm:
@rmosolgo I've brought our site down because of this 😅The initial requests were timing out because of this so I'm wondering if there should be a big note about this somewhere for others in case there isn't one already

PR #2363, opened 2019-07-03
Simplify schema boot

rmosolgo:
In the past, building out a GraphQL schema has had two steps: 

- Add root types
- Then, later, traverse those types and build out a full cache of types-by-name 

This is complicated because the traversal can cause files to be autoloaded by Rails, and it can happen at odd times. See #1505 

I want to simplify app boot so that the traversal happens _during_ the attachment of a root type. That will cause the schema to be booted all at once. I'm hoping it will clarify the error messages, for example, when a file has invalid syntax. 

However, it will reduce the "laziness" of a schema, which could make it harder to handle cyclical dependencies. We'll have to cross those bridges as we get to them.

__Breaking changes:__

- For class-based schemas, methods that return types now return class-based types instead of legacy type instances.
- Class-based schemas may not contain `.define`-based types
- Lots of test cases were migrated from `.define`-based scenarios to class-based ones
- The interpreter _requires_ `Analysis::AST`
- `Schema.types` has become expensive because it merges inherited types together and it responds to dynamically-added types. Use `Schema.get_type(name)` for runtime type lookup, instead. 

__TODO__ 

- [x] backport coerce_isolated_input 
- [x] Don't pass `nil` to coerce_* methods
- [x] rename find_type -> get_type (like get_field) 

Fixes #2320

PR #2418, opened 2019-08-07
Add heads up about deferred schema initialization

fgrehm:
Related to https://github.com/rmosolgo/graphql-ruby/issues/2034

We might want to add a note about this somewhere else but I think it is super important for people that are converting to the class based API so that they are not surprised with request timeouts like me 😅


rmosolgo:
Thanks for adding that! I hope it will _stop_ working like that in 1.10 🤞  (working on #2363 )

PR #2421, opened 2019-08-09
Isolate analyzer visits of fragment spreads

### LINKED ISSUE #2419: Wrong AST max complexity computation
To reproduce an error
```ruby
# spec/support/dummy/schema.rb

class MaxSchema < GraphQL::Schema
  query DairyAppQuery

  max_depth 3
  max_complexity 1

  use GraphQL::Analysis::AST
end
```
```ruby
# spec/graphql/analysis/ast/max_query_complexity_spec.rb

describe "should not affected by other plugins" do
  let(:schema) { Dummy::MaxSchema }
  let(:query_string) {%|
    {
      a: cheese(id: 1) { ...cheeseFields }
      b: cheese(id: 1) { ...cheeseFields }
      c: cheese(id: 1) { ...cheeseFields }
      d: cheese(id: 1) { ...cheeseFields }
      e: cheese(id: 1) { ...cheeseFields }
    }

    fragment cheeseFields on Cheese { id }
  |}
  let(:result) { schema.execute(query_string) }

  it "returns an error" do
    assert_equal "Query has complexity of 10, which exceeds max complexity of 1", result["errors"].first["message"]
  end
end
```

**Expected**

`Query has complexity of 10, which exceeds max complexity of 1`

**Actual**

`Query has complexity of 15, which exceeds max complexity of 1`

Problem appeared after https://github.com/rmosolgo/graphql-ruby/pull/2389, but probably it's result of a more complex problem. In quick look, I noticed more frequent call `AST::QueryComplexity#on_leave_field` after adding `max_depth` in schema.

jturkel:
This fixes #2419 by only allowing analyzers to perform isolated visits of fragment spreads which should prevent any unintended interactions between analyzers. I replaced the `enter_fragment_spread_inline` and `leave_fragment_spread_inline` methods on `GraphQL::Analysis::AST::Visitor` with a single `visit_fragment_spread_inline` method that only calls back into the specified `GraphQL::Analysis::AST::Analyzer` instance.  By default `GraphQL::Analysis::AST::Analyzer`s will automatically recurse into fragment spreads but this behavior can be changed by overriding `on_enter_fragment_spread`. This does add a performance hit because each analyzer is independently iterating over fragment spreads but I doubt that will be significant.

This also fixes an issue with the field usage analyzer not recursing into fragment spreads.

### DIFF
--- a/lib/graphql/analysis/ast/analyzer.rb
+++ b/lib/graphql/analysis/ast/analyzer.rb
@@ -41,7 +41,6 @@ module GraphQL
         build_visitor_hooks :document
         build_visitor_hooks :enum
         build_visitor_hooks :field
-        build_visitor_hooks :fragment_spread
         build_visitor_hooks :inline_fragment
         build_visitor_hooks :input_object
         build_visitor_hooks :list_type
@@ -52,6 +51,13 @@ module GraphQL
         build_visitor_hooks :variable_definition
         build_visitor_hooks :variable_identifier
         build_visitor_hooks :abstract_node
+
+        def on_enter_fragment_spread(node, _parent, visitor)
+          visitor.visit_fragment_spread_inline(node, self)
+        end
+
+        def on_leave_fragment_spread(_node, _parent, _visitor)
+        end
 
         protected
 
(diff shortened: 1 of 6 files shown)


rmosolgo:
Sorry for the delay on this. I really want to think about a better _default_ before we introduce duplicate visits!

rmosolgo:
Fixed by #2463 , thanks again for taking this on!!

PR #2431, opened 2019-08-16
Add extensions in CoercionError

ogom:
The idea of Custom Scalars is great. 

In `raise GraphQL::CoercionError` used when the data is not correct, it seems better to be able to return optional keys and values to the client with `extensions`.

PR with reference to https://github.com/rmosolgo/graphql-ruby/issues/1621.


rmosolgo:
This looks great, thanks for taking care of it! Fixes #1621

PR #2432, opened 2019-08-19
Enable multiple execution errors for Fields defined to return a list …

stanishev:
Fields defined to return a **List** of some type should be able to return an Array of `GraphQL::ExecutionError`s instead.

### DIFF
--- a/spec/graphql/execution_error_spec.rb
+++ b/spec/graphql/execution_error_spec.rb
@@ -296,18 +296,6 @@ describe GraphQL::ExecutionError do
   describe "more than one ExecutionError" do
     let(:query_string) { %|{ multipleErrorsOnNonNullableField} |}
     it "the errors are inserted into the errors key and the data is nil even for a NonNullable field " do
-
-      # I Think the path here is _wrong_, since this is not an array field:
-      # expected_result = {
-      #     "data"=>nil,
-      #     "errors"=>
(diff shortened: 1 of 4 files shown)


spec/graphql/execution_error_spec.rb
```diff
@@ -296,18 +296,6 @@
   describe "more than one ExecutionError" do
     let(:query_string) { %|{ multipleErrorsOnNonNullableField} |}
     it "the errors are inserted into the errors key and the data is nil even for a NonNullable field " do
-
```
stanishev: This comment makes sense to me because I submitted the PR that introduced this test. The comment is correct (and thank you sincerely for pointing this out).

I've taken the liberty to remove the comment because I see that you have already corrected the spec (and the gem to actually return the expected result below) and I think that at this point the comment is no longer necessary.

spec/graphql/execution_error_spec.rb
```diff
@@ -320,6 +308,22 @@
       }
       assert_equal(expected_result, result)
     end
-  end
 
+    describe "more than one ExecutionError on a field defined to return a list" do
```
stanishev: One more test, similar to the earlier one, but this time the return type *is* a list type and so the path includes the list indices as well.

stanishev:
not yet ready for review

PR #2433, opened 2019-08-19
Enable multiple execution errors for Fields defined to return a list …

stanishev:
Fields defined to return a List of some type should be able to return an Array of GraphQL::ExecutionErrors instead.

### DIFF
--- a/lib/graphql/execution/execute.rb
+++ b/lib/graphql/execution/execute.rb
@@ -197,7 +197,7 @@ module GraphQL
             raw_value.path = field_ctx.path
             query.context.errors.push(raw_value)
           when Array
-            if !field_type.list?
+            if field_type.non_null?
               # List type errors are handled above, this is for the case of fields returning an array of errors
               list_errors = raw_value.each_with_index.select { |value, _| value.is_a?(GraphQL::ExecutionError) }
               if list_errors.any?
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -243,9 +243,9 @@ module GraphQL
             write_execution_errors_in_response(path, [value])
             HALT
           elsif value.is_a?(Array) && value.any? && value.all? { |v| v.is_a?(GraphQL::ExecutionError) }
-            value.each do |v|
-              v.path ||= path
-              v.ast_node ||= ast_node
+            value.each_with_index do |error, index|
+              error.ast_node ||= ast_node
+              error.path ||= path + (field.type.list? ? [index] : [])
             end
             write_execution_errors_in_response(path, value)
             HALT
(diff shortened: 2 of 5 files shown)


spec/graphql/execution_error_spec.rb
```diff
@@ -296,18 +296,6 @@
   describe "more than one ExecutionError" do
     let(:query_string) { %|{ multipleErrorsOnNonNullableField} |}
     it "the errors are inserted into the errors key and the data is nil even for a NonNullable field " do
-
-      # I Think the path here is _wrong_, since this is not an array field:
```
stanishev: This comment makes sense to me because I submitted the PR that introduced this test. The comment is correct (and thank you sincerely for pointing this out).

I've taken the liberty to remove the comment because I see that you have already corrected the spec (and the gem to actually return the expected result below) and I think that at this point the comment is no longer necessary. To someone else reading this without the context we have, I think it may just be confusing without add much.

spec/graphql/execution_error_spec.rb
```diff
@@ -320,6 +308,22 @@
       }
       assert_equal(expected_result, result)
     end
-  end
 
+    describe "more than one ExecutionError on a field defined to return a list" do
```
stanishev: One more test, similar to the earlier one, but this time the return type is a list type and so the path includes the list indices as well.

lib/graphql/execution/execute.rb
```diff
@@ -197,7 +197,7 @@ def continue_resolve_field(raw_value, field_type, field_ctx)
             raw_value.path = field_ctx.path
             query.context.errors.push(raw_value)
           when Array
-            if !field_type.list?
+            if field_type.non_null?
```
stanishev: I *think* this is the correct logic for including (or not) the array of errors here. 

This covers all scenarios: 
- fields defined with `null: true` will **not** have these errors included here since as the comment below mentions they were already handled somewhere else. 
- fields defined with `null: false` **will** have the list of errors included, regardless of whether they are defined to return a single type or a list of some type.

stanishev:
Currently the [spec](https://github.com/rmosolgo/graphql-ruby/pull/2433/files#diff-68e27cbe1ad3427b52db16a3938e0745R312) I added fails for the ruby 2.4.3 build in TravisCI. The actual results **do not** include the indices in the path. I am unable to replicate this failure locally.

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -243,9 +243,9 @@ def continue_value(path, value, field, is_non_null, ast_node)
             write_execution_errors_in_response(path, [value])
             HALT
           elsif value.is_a?(Array) && value.any? && value.all? { |v| v.is_a?(GraphQL::ExecutionError) }
-            value.each do |v|
-              v.path ||= path
-              v.ast_node ||= ast_node
+            value.each_with_index do |error, index|
```
stanishev: adding the index to the path for list types.

rmosolgo:
Awesome, thanks so much for taking care of this 🎉 !

PR #2435, opened 2019-08-21
Correct the example for using 'a class method to generate fields'

recurser:
I have tried using the example included, and get the error `undefined method 'argument' for Types::QueryType:Class`. After some investigation, the correct way seems to be to pass `self` to the block as a parameter.


rmosolgo:
Oops, thanks for the fix! I guess, as written, it should have been `instance_eval`.

PR #2437, opened 2019-08-21
Include introspective fields in query depth calculations

cablett:
According to a Slack conversation with @rmosolgo , the exclusion of introspective fields from query depth calculation may be a bug. 

This PR removes that exclusion so all fields are included in the query depth calculation. 

I am happy for it also to be conditional -ie. passing something like `include_all_fields: true` invokes this behaviour and the current behaviour (filtering out introspective fields) remains the default to ensure it doesn't unreasonably break functionality.

Please feel free to edit/comment.

### DIFF
--- a/lib/graphql/analysis/ast/query_depth.rb
+++ b/lib/graphql/analysis/ast/query_depth.rb
@@ -29,35 +29,22 @@ module GraphQL
         def initialize(query)
           @max_depth = 0
           @current_depth = 0
-          @skip_depth = 0
           super
         end
 
         def on_enter_field(node, parent, visitor)
           return if visitor.skipping? || visitor.visiting_fragment_definition?
 
-          # Don't validate introspection fields or skipped nodes
-          if GraphQL::Schema::DYNAMIC_FIELDS.include?(visitor.field_definition.name)
-            @skip_depth += 1
-          elsif @skip_depth > 0
-            # we're inside an introspection query or skipped node
-          else
-            @current_depth += 1
-          end
+          @current_depth += 1
         end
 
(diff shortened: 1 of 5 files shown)


rmosolgo:
Thanks, this looks great! I agree it's the right change. 

If anyone wants to exclude introspection, they can copy the analyzer from a previous version. Or if there turn out to be lots of folks who want it, we can look for an option like you mentioned. 🎉

knapo:
@rmosolgo @cablett 

This PR introduces an issue with schema dump. After upgrading to 1.9.11, my `schema.json` (in 2 small apps) looks like: 

```
{
  "errors": [
    {
      "message": "Query has depth of 13, which exceeds max depth of 5"
    }
  ]
}
```

rmosolgo:
Sorry that wasn't in the changelog! I've just updated the changelog in 6fc8bea4b. 

You should either increase the depth allowed by your schema, or pass `max_depth: 13` to `MySchema.execute(...)` when you run the introspection query, for example: 

```ruby 
MySchema.execute(GraphQL::Introspection::INTROSPECTION_QUERY, max_depth: 13)
```

jayqui-omada:
It appears that this breaks the `GraphQL::Schema#to_json` method on schemas with introspection. It might make sense to add an option to `GraphQL::Schema#to_json` and/or `GraphQL::Schema#as_json` to specify a `max_depth` for the `execute` call that those wrap.

rmosolgo:
@jayqui-omada  if you're interested in looking into it, please open an issue including the _actual_ result you got from `#to_json`!

PR #2447, opened 2019-08-27
Fix ArgumentError

mathieujobin:
This method does not take any argument. At least on ruby 2.4
Version 1.8.5 was calling this method without argument and we started getting into issues after upgrading the gem.

### DIFF
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -32,7 +32,7 @@ module GraphQL
       # @param value [DateTime]
       # @return [String]
       def self.coerce_result(value, _ctx)
-        value.iso8601(time_precision)
+        value.iso8601
       end
 
       # @param str_value [String]


rmosolgo:
Hi, it looks like [Time](https://ruby-doc.org/stdlib-2.4.0/libdoc/time/rdoc/Time.html#method-i-iso8601) and [DateTime](https://ruby-doc.org/stdlib-2.4.0/libdoc/date/rdoc/DateTime.html#method-i-iso8601) both accept an argument here, but [Date](https://ruby-doc.org/stdlib-2.4.1/libdoc/date/rdoc/Date.html#method-i-iso8601) does not. If you want to use the `DateTime` type for a Ruby `Date` object, how about converting your date to a datetime first? Or, you could use a `Date` type for `Date` objects?

rmosolgo:
Let me know if one of those solutions doesn't work for you!

Issue #2366, opened 2019-07-04
Nesting resolvers of the same type crashes the gem


tilod:
- Ruby 2.6.1
- Rails 5.2.3
- graphql-ruby 1.9.7

When using a resolver for the field definition of a type which the resolver returns, the app crashes on any query (including requesting the schema).

The error is `Failed to build return type for Task.tasks from nil: Unexpected type input:  (NilClass)` (full stacktrace included below). The stacktrace shows that the nested resolver is not loaded and therefore its `type` definition is not read resulting in the `nil`.

## Example 1: Direct nesting

Scenario: A project management app with tasks which can have subtasks. The tasks can be nested indefinitely. So you may request:

```graphql
query {
  tasks {
    title
    tasks {
      title
    }
  }
}
```

<details>
<summary>
Ruby files
</summary>

```ruby
# app/graphql/types/query_type.rb

module Types
  class QueryType < Types::BaseObject
    field :tasks, resolver: Resolvers::TasksResolver
  end
end

# app/graphql/types/task_type.rb

module Types
  class TaskType < Types::BaseObject
    field :title, String, null: false
    field :tasks, resolver: Resolvers::TasksResolver
  end
end

# app/graphql/resolvers/tasks_resolver.rb

module Resolvers
  class TasksResolver < GraphQL::Schema::Resolver
    type [Types::TaskType], null: false

    def resolve
      []
    end
  end
end
```

</details>

<details>
<summary>
Stacktrace
</summary>

```
Failed to build return type for Task.tasks from nil: Unexpected type input:  (NilClass)
/Users/tilo/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/schema/member/build_type.rb:80:in `parse_type'
/Users/tilo/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/schema/field.rb:399:in `type'
/Users/tilo/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/schema/field.rb:123:in `connection?'
/Users/tilo/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/graphql-1.9.7/lib/graphql/schema/field.rb:251:in `initialize'
[...]

PR #2457, opened 2019-09-01
Update guide around nesting resolvers of same type

cursande:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/2366

I encountered basically the same issue as above, and also solved it by expressing the type as a string.

To avoid having to hunt through issues or tickets, it would be nice if this was in the guide for resolvers.


rmosolgo:
Awesome, thanks for improving the docs on that!

PR #2458, opened 2019-09-02
Add error handling for the interpreter

### LINKED ISSUE #2139: GraphQL::Execution::Interpreter and rescue_from compatibility
Hello,

As documented [here](https://graphql-ruby.org/queries/interpreter.html#compatibility) the new `GraphQL::Execution::Interpreter` does not work with the `rescue_from` middleware. I think the rescue_from add valuable error handeling that allows your endpoint to always return a 200 and indicate the errors. The following examples are based of the `graphql-errors` gem which we currently use https://github.com/exAspArk/graphql-errors

The most common example is catching `ActiveRecord::RecordNotFound` and replace them with `nil` so whenever you try to fetch something that does not exist you return a 200 with nothing in it.

```ruby
  rescue_from ActiveRecord::RecordNotFound do |_exception|
    nil
  end
```

Another example is to error standardisation, for example in our GraphQL api we always call `update!` or `save!` this raises a `ActiveRecord::RecordInvalid` which we capture with rescue_from and standarize the error response:

```ruby
rescue_from ActiveRecord::RecordInvalid do |exception|
  RecordInvalidError.from_exception(exception)
end

class RecordInvalidError < GraphQL::ExecutionError
  def to_h
    super.merge(problems: extensions[:errors])
  end

  def self.from_exception(exception)
    record = exception.record
    errors = record.errors.map do |attr|
      {
        path: [attr.to_s.camelize(:lower)],
        explanation: record.errors.full_messages_for(attr)
      }
    end
    new(record.errors.full_messages.join("\n"), extensions: { errors: errors })
  end
end
```

Ofcourse one can say that every mutation / resolver should deal with the errors themself, but this means that you always have to go for a `if ...save else do_with_errors end` kind of thing which isn't that pretty nor ensures consistency.

Finally sometimes (even when you know its bad) you have some deeply nested code where you want to brake out of so your only resort is to raise an exception, this does not mean you want to return back a 500 but rather a customised message.

If anyone else has other use cases then please do share,

Regards

rmosolgo:
Based on https://github.com/exAspArk/graphql-errors and `@thiago-sydow`'s comment over at https://github.com/rmosolgo/graphql-ruby/issues/2139#issuecomment-524913594

closes #2139


rmosolgo:
If you want to give this a try before the next release, you can bundle the gem's master branch:

```ruby 
gem "graphql", github: "rmosolgo/graphql-ruby"
```

And add the new plugin as described [in the new docs](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/errors/error_handling.md)

thiago-sydow:
Nice to see that the initial implementation resulted in this. Thanks !

exAspArk:
Awesome job! I'm going to deprecate [graphql-errors](https://github.com/exAspArk/graphql-errors) now.
Thank you, one less project to support and worry about 😂

Issue #2461, opened 2019-09-03
Defined but empty schema.query type is not considered invalid


chrisbutcher:
# Expected behaviour

```ruby
schema_doc = "type Query { }"
GraphQL::Schema.from_definition(schema_doc) # I would expect this to 💥 
```

In accordance with the js reference implementation:

```javascript
import { buildSchema } from 'graphql';
const typeDefs = `type Query { }`;
buildSchema(typeDefs); // 💥 
```
```
/.../node_modules/graphql/language/parser.js:1433
  throw (0, _error.syntaxError)(lexer.source, token.start, "Expected ".concat(kind, ", found ").concat((0, _lexer.getTokenDesc)(token)));
  ^
GraphQLError: Syntax Error: Expected Name, found }
    at syntaxError (/.../graphql/error/syntaxError.js:24:10)
    at expectToken (/.../graphql/language/parser.js:1433:32)
    at parseName (/.../graphql/language/parser.js:98:15)
    at parseFieldDefinition (/.../graphql/language/parser.js:870:14)
    at many (/.../graphql/language/parser.js:1521:16)
    at parseFieldsDefinition (/.../graphql/language/parser.js:859:50)
    at parseObjectTypeDefinition (/.../graphql/language/parser.js:813:16)
    at parseTypeSystemDefinition (/.../graphql/language/parser.js:711:16)
    at parseDefinition (/.../graphql/language/parser.js:144:16)
    at many (/.../graphql/language/parser.js:1521:16)
```

# Actual behaviour

```ruby
schema_doc = "type Query { }"
=> "type Query { }"
[11] pry(main)> GraphQL::Schema.from_definition(schema_doc)
=> #<GraphQL::Schema ...> # ❌
```

Parsing the schema works, since the schema.query is defined and that's all the gem expects.

dteoh:
This change introduced in 1.9.14 is wrong, it goes against the spec. Issue opened here: https://github.com/rmosolgo/graphql-ruby/issues/2549

PR #2462, opened 2019-09-03
Treat defined but empty schema.query type as invalid

chrisbutcher:
Closes https://github.com/rmosolgo/graphql-ruby/issues/2461

See the above issue for background, expected vs. actual, and repro steps.

# The fix

This PR. I think we should not only ensure schema.query is defined, but that it has at least one fields defined on it.

```ruby
schema_doc = "type Query { }"
GraphQL::Schema.from_definition(schema_doc) # 💥"Schema definition Query type must define at least one field."
```

# TODO (edit: Done)

~This change, however, conflicts with an existing test case that now fails: https://github.com/chrisbutcher/graphql-ruby/blob/error-when-query-defined-but-empty/spec/graphql/schema/build_from_definition_spec.rb#L723-L730~

~I may be missing some important context on why `it 'supports empty types'`, or maybe this is just outdated and a case that this gem shouldn't support any longer.~

edit: The most recent commit, to add a new Rule entry now breaks a lot more tests (45 errors, 2 failures)

edit 2: Removed the no longer valid test case, and fixed all existing instances where object types were defined without at least one field defined.

### DIFF
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -87,8 +87,12 @@ module GraphQL
             subscription_root_type = types['Subscription']
           end
 
-          unless query_root_type && query_root_type.fields.any?
+          unless query_root_type
             raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.')
+          end
+
+          unless query_root_type.fields.any?
+            raise InvalidDocumentError.new('Schema definition Query type must define at least one field.')
           end
 
           schema = Schema.define do
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -87,13 +87,7 @@ module GraphQL
             subscription_root_type = types['Subscription']
           end
 
-          unless query_root_type
-            raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.')
-          end
-
-          unless query_root_type.fields.any?
-            raise InvalidDocumentError.new('Schema definition Query type must define at least one field.')
-          end
+          raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.') unless query_root_type
 
           schema = Schema.define do
             raise_definition_error true
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -87,7 +87,9 @@ module GraphQL
             subscription_root_type = types['Subscription']
           end
 
-          raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.') unless query_root_type
+          unless query_root_type && query_root_type.fields.any?
+            raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.')
+          end
 
           schema = Schema.define do
             raise_definition_error true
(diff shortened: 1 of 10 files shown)


chrisbutcher:
cc: @rmosolgo @swalkinshaw

lib/graphql/schema/build_from_definition.rb
```diff
@@ -87,7 +87,13 @@ def build(document, default_resolve: DefaultResolve)
             subscription_root_type = types['Subscription']
           end
 
-          raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.') unless query_root_type
+          unless query_root_type
+            raise InvalidDocumentError.new('Must provide schema definition with query type or a type named Query.')
+          end
+
+          unless query_root_type.fields.any?
```
swalkinshaw: This just handles here but wouldn't catch defining the schema normally (via class). You could add a new schema validation to handle both cases: https://github.com/rmosolgo/graphql-ruby/blob/40c838eb4a9f95b9517e7c98077e3cd294ab4677/lib/graphql/schema/validation.rb

chrisbutcher:
Updated w/ an approach that uses @swalkinshaw's suggestion

chrisbutcher:
👋 Re-ping @rmosolgo - just checking: do you think the gem should implement the reference implementation's behaviour here?

rmosolgo:
Hey, sorry for the late review. I think you're spot on here. Looking back at why that test was added, it was to support a user of graphql-docs who later decided that they could just remove those empty definitions (https://github.com/gjtorikian/graphql-docs/issues/32). 

So, I think it's ok to remove that failing tests. It looks like the schema in `authorization_test` will need some dummy fields added, too 🙈 !

chrisbutcher:
Thanks for confirming. Will return to this PR soon.

lib/graphql/schema/printer.rb
```diff
@@ -60,7 +60,9 @@ def initialize(schema, context: nil, only: nil, except: nil, introspection: fals
 
       # Return the GraphQL schema string for the introspection type system
       def self.print_introspection_schema
-        query_root = ObjectType.define(name: "Root")
+        query_root = ObjectType.define(name: "Root") do
+          field :throwaway_field, types.String
```
chrisbutcher: Adding this was needed in order to satisfy the new validation, but note that the Root type itself gets excluded from the `introspection_schema_ast` before it is stringified: https://github.com/chrisbutcher/graphql-ruby/blob/error-when-query-defined-but-empty/lib/graphql/schema/printer.rb#L70

spec/graphql/authorization_spec.rb
```diff
@@ -104,6 +104,8 @@ class HiddenObject < BaseObject
       def self.visible?(ctx)
         super && !ctx[:hide]
       end
+
+      field :some_field, String, null: true
```
chrisbutcher: Feel free to suggest a better name!

spec/graphql/schema_spec.rb
```diff
@@ -11,11 +11,23 @@ class DummyFeature2 < GraphQL::Schema::Directive::Feature
...
+      field :some_field, String, null: true
+    end
+
+    class Mutation < GraphQL::Schema::Object
+      field :some_field, String, null: true
+    end
+
+    class Subscription < GraphQL::Schema::Object
+      field :some_field, String, null: true
+    end
```
chrisbutcher: Not sure if this is all necessary. Without defining and passing these in below, I got an error:
`GraphQL::Schema::InvalidTypeError: Unexpected traversal member: Mutation (Class)`

rmosolgo: Weird, in any case, seems like a reasonable change.

PR #2463, opened 2019-09-03
Automatically inline fragment spreads during AST visits

### LINKED ISSUE #2419: Wrong AST max complexity computation

jturkel:
This changes `GraphQL::Analysis::AST::Visitor` to automatically visit fragment spreads inline to fix #2419. It's an alternative to #2421 that avoids forking visitors for each analyzer when visiting fragment spreads. As part of this I made `GraphQL::Analysis::AST::Visitor#enter_fragment_spread_inline` and `GraphQL::Analysis::AST::Visitor#leave_fragment_spread_inline` private since they can cause an analyzer to screw up the iteration for other analyzers. I left the `fragment_spread` callbacks in `GraphQL::Analysis::AST::Analyzer` to be consistent with the `inline_fragment` callbacks but I'm not sure when these would actually be useful. This is hopefully a minimal change to fix #2419 which is a pretty unfortunate regression.

### DIFF
--- a/lib/graphql/analysis/ast/query_complexity.rb
+++ b/lib/graphql/analysis/ast/query_complexity.rb
@@ -50,14 +50,6 @@ module GraphQL
           end
         end
 
-        def on_enter_fragment_spread(node, _, visitor)
-          visitor.enter_fragment_spread_inline(node)
-        end
-
-        def on_leave_fragment_spread(node, _, visitor)
-          visitor.leave_fragment_spread_inline(node)
-        end
-
         # @return [Integer]
         def max_possible_complexity
           @complexities_on_type.last.max_possible_complexity
--- a/lib/graphql/analysis/ast/query_depth.rb
+++ b/lib/graphql/analysis/ast/query_depth.rb
@@ -47,14 +47,6 @@ module GraphQL
           @current_depth -= 1
         end
 
-        def on_enter_fragment_spread(node, _, visitor)
-          visitor.enter_fragment_spread_inline(node)
-        end
-
-        def on_leave_fragment_spread(node, _, visitor)
-          visitor.leave_fragment_spread_inline(node)
-        end
-
         def result
           @max_depth
         end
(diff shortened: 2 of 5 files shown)


swalkinshaw:
This is simpler than I originally had envisioned since it's just forcing the inline visit by default. Will this cause `on_enter_fragment_definition` and `on_leave_fragment_definition` to be called as well?

jturkel:
The visit will invoke:

1. `on_enter_fragment_spread`
1. `on_field` for each selection in the fragment
1. `on_leave_fragment_spread`

It doesn't invoke any of the `fragment_definition` callbacks but clients can access the fragment definition in the `fragment_spread` callbacks if they need it.

swalkinshaw:
🤔 not 100% sure what would be the ideal solution if this was designed from scratch, but I think that makes sense 👍

rmosolgo:
This is great, thanks for taking a look down this path. I agree it seems like a cleaner solution for now, especially since it was _already happening_ for anyone using max depth or max complexity!

PR #2469, opened 2019-09-09
Fix AST analysis documentation

RobertWSaunders:
# Proposal

Since https://github.com/rmosolgo/graphql-ruby/pull/2463/ we inline fragment spreads automatically in AST visits. As such, the example in the docs is no longer correct and would cause an error because `enter_fragment_spread_inline` and `leave_fragment_spread_inline` are now private methods on the `GraphQL::Analysis::AST::Visitor` class. So this updates that example in the docs to be correct.

In addition to that, I noticed in the [docs](https://graphql-ruby.org/api-doc/1.9.12/GraphQL/Analysis/AST/Analyzer) for `GraphQL::Analysis::AST::Analyzer` that `build_visitor_hooks` had a message to not use it, I think we should make it a private method and update the docs to be more clear.

### DIFF
--- a/lib/graphql/analysis/ast/analyzer.rb
+++ b/lib/graphql/analysis/ast/analyzer.rb
@@ -25,15 +25,18 @@ module GraphQL
           raise NotImplementedError
         end
 
-        # Don't use make_visit_method because it breaks `super`
-        def self.build_visitor_hooks(member_name)
-          class_eval(<<-EOS, __FILE__, __LINE__ + 1)
-            def on_enter_#{member_name}(node, parent, visitor)
-            end
+        class << self
+          private
 
-            def on_leave_#{member_name}(node, parent, visitor)
-            end
-          EOS
+          def build_visitor_hooks(member_name)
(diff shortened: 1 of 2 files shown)


lib/graphql/analysis/ast/analyzer.rb
```diff
@@ -25,7 +25,9 @@ def result
           raise NotImplementedError
         end
 
-        # Don't use make_visit_method because it breaks `super`
+        private
```
swalkinshaw: This won't actually make it private since the method is defined as `self.build_visitor_hooks`. You have two options:

```ruby
def self.build_visitor_hooks(member_name)
  # etc
end

private_class_method :build_visitor_hooks

# or even weirder

private_class_method def self.build_visitor_hooks(member_name)
  # etc
end
```

```ruby
class << self
  private

  def build_visitor_hooks(member_name)
    # etc
  end
end
```

RobertWSaunders: Whoops, thanks for this! 🔥Updated!

rmosolgo:
Thanks for making this update!

PR #2470, opened 2019-09-10
Use Types::BaseField in appropriate rails base templates

bfad:
When running `rails g graphql:install` to setup Rails integration, make sure that `Types::BaseObject` and `Types::BaseInterface` set their field class to the generated `Types::BaseField`. (Similarly to how various templates call `argument_class`.)


rmosolgo:
That's great, thanks for making this fix!

PR #2471, opened 2019-09-10
Make ISO8601DateTime work with Date

DmitryTsepelev:
I've figured out that `ISO8601DateTime` handles `Date` objects in a funny way – it fails with `ArgumentError` (because `Date#iso8601` does not accept any arguments), but from the controller:

```ruby
ArgumentError (wrong number of arguments (given 1, expected 0)):

app/controllers/graphql_controller.rb:10:in `execute'
...
```

When I figured out what's going on, I've written my own implementation for `Date` and decided to open up this PR – the idea is that now `ISO8601DateTime` can accept `Date` too. There are another options of what we can do here: I can write a separate scalar type for Dates or just fix the docs to make it clear that the current type doesn't work with dates. What do you think of it?

### DIFF
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -33,6 +33,8 @@ module GraphQL
       # @return [String]
       def self.coerce_result(value, _ctx)
         value.iso8601(time_precision)
+      rescue ArgumentError
+        raise GraphQL::Error, "An incompatible object (#{value.class}) was given to #{self}. Make sure that only DateTimes are used with this type."
       end
 
       # @param str_value [String]
--- a/guides/type_definitions/scalars.md
+++ b/guides/type_definitions/scalars.md
@@ -16,6 +16,7 @@ Scalars are "leaf" values in GraphQL. There are several built-in scalars, and yo
 - `Boolean`, like a JSON or Ruby boolean (`true` or `false`)
 - `ID`, which a specialized `String` for representing unique object identifiers
 - `ISO8601DateTime`, an ISO 8601-encoded datetime
+- `ISO8601Date`, an ISO 8601-encoded date
 - `JSON`, ⚠ This returns arbitrary JSON (Ruby hashes, arrays, strings, integers, floats, booleans and nils). Take care: by using this type, you completely lose all GraphQL type safety. Consider building object types for your data instead.
 
 Fields can return built-in scalars by referencing them by name:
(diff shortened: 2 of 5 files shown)


lib/graphql/types/iso_8601_date_time.rb
```diff
@@ -32,7 +32,11 @@ def self.time_precision=(value)
       # @param value [DateTime]
       # @return [String]
       def self.coerce_result(value, _ctx)
-        value.iso8601(time_precision)
+        if value.is_a?(DateTime)
```
rmosolgo: How will this handle an `ActiveSupport::TimeWithZone`? I don't think it's tested, but I can imagine it would stop working here. Would you mind checking whether it will be affected by this? 

Another approach would be to use `if value.is_a?(Date)` to identify the special case, instead. (🤔 but maybe DateTime _extends_ Date?)

DmitryTsepelev: That's a valid point! What if we check method _arity_ instead of sticking with exact classes? It should cover all the cases we might face in future (e.g., new Date-ish classes)

DmitryTsepelev: @rmosolgo Could you please take another look? 🙂

michal-samluk:
Do you think we should have same object for Date and DateTime? In my project there is a separate scalar for Date values. I think there is value in differentiating those 2 types. 

Here is a sample implementation.

```ruby
module Types
  module Scalars
    class ISO8601Date < BaseScalar
      description 'An ISO 8601-encoded date'

      def self.coerce_input(input_value, _context)
        Date.iso8601(input_value)
      rescue ArgumentError
        nil
      end

      def self.coerce_result(ruby_value, _context)
        ruby_value.iso8601
      end
    end
  end
end
```

DmitryTsepelev:
@michal-samluk That's definitely an option, in my project I do have a separate type too (what else can I do 😅). From one point of view, two separate types give us better performance and control, but from another – it's a bit more error prone, cause someone can break both types with the inaccurate line like `user.created_at || Date.today`. I'll be happy to implement any option we choose.

rmosolgo:
Sorry for my indecision, but I agree with @michal-samluk . I think best practice would be: 

- Use a `Types::DateTime` type to expose DateTimes (if you want to expose a Date with this, convert it to a DateTime first)
- Use a `Types::Date` type to expose Dates 

> someone can break both types with the inaccurate line like `user.created_at || Date.today`

That code is a bug, and if we expose it, that's _good_, right? Shouldn't that application code be corrected to `user.created_at || DateTime.now` (or maybe `Time.zone.now`) ? 

Maybe a way we could improve the developer experience is to raise a more descriptive error when the `DateTime` type receives a `Date`. For example could we rescue the ArgumentError, then reraise, something like: 

```ruby 
rescue ArgumentError 
  raise GraphQL::Error, "An incompatible object (#{value.class}) was given to #{self}. Make sure that only DateTimes are used with this type."
```

?

DmitryTsepelev:
Completely agree! I'll extract the code for `Date` to the separate type, make the error more explicit and come back soon

DmitryTsepelev:
@rmosolgo @michal-samluk Could you please review again?

michal-samluk:
@DmitryTsepelev It looks good :)

PR #2481, opened 2019-09-16
Simplify default scalar setter method.

peterb:
Just making this a bit simpler.


rmosolgo:
Thanks for sharing your style preference, but as maintainer, I prefer code where the left-to-right, top-t-bottom reading matches its execution. So, I'd rather keep the current style. 

If you'd like to contribute a feature or bug fix, you can use whatever style you like, so long as the few cops in our rubocop config pass 😊 !

PR #2482, opened 2019-09-17
Update node_field.rb

Kaytal:
`schema.object_from_id` takes two arguments `id` and `context` in that order.  This is visible on line 34.


rmosolgo:
Oops, thanks for the fix!!

PR #2483, opened 2019-09-18
GraphQL::Dataloader, built-in batching system

rmosolgo:
Projects like https://github.com/Shopify/graphql-batch and https://github.com/exaspark/batch-loader have proven the value of batch loading in GraphQL. In fact, you really _can't_ run a production GraphQL system _without_ batching. 

For this reason, I want to include a batching system _in_ GraphQL-Ruby (without breaking compat with existing systems, of course!) Here are some goals for this system: 

- Feature parity with GraphQL-Batch 
- First-class support for pushing IO to a background thread 
- Traceable -- include graphql context with loads, so that a developer can see what GraphQL fields used which loaders. 
- Good built-in defaults (eg, ActiveRecord, Redis, HTTP resources ... others?) 
- Well-documented, low-friction custom loaders 

If anyone has other suggestions for a built-in dataloader, please share them! 

------

__TODO__ 

- [x] Test caching in queries vs. no cache in mutation root 
- [x] Add a good API
- [x] Test Concurrent::Future-powered background loaders 
- [x] Make sure @panthomakos's considerations regarding thread safety are addressed 
- [x] Make sure traceability matches graphql-metrics 
- [x] Add some built-in loaders & test them
- [x] Sort out `Execution::Lazy` vs `Dataloader::PendingLoad` vs `::Promise` 
- [ ] Make sure graphql-batch's tests pass with GraphQL::Dataloader https://github.com/rmosolgo/graphql-batch/pull/1
- [ ] Performance audit (GitHub uses a branch of Promise.rb for lower memory use) 
- [ ] Any reason not to install `Dataloader` by default? (dependency?)
- [ ] Update docs for concurrent-ruby dependency in Dataloader main (not only background thread) 
- [ ] Support Lazy.sync as a public api? Or support instrumentation (see graphql-batch tests)
- Docs: 
  - [x] Concepts
  - [x] Installation 
  - [x] Built-in loaders 
  - [x] Custom loaders 
  - [ ] Standardize language of batch keys / fetch parameters: what does graphql-batch do here?


rmosolgo:
@panthomakos, I'd love to get your feedback on the goals discussed here. This is where I want to take inspiration your work from #1981 😊 , so please let me know if I've overlooked any of your goals and accomplishments from that branch.

eapache:
One of our requirements at Shopify is that we make very heavy use of https://github.com/Shopify/graphql-metrics/blob/master/lib/graphql_metrics/timed_batch_executor.rb, so we'll need some sort of similar hook where we can collect performance data.

lib/graphql/dataloader/background_loader.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+module GraphQL
+  class Dataloader
+    # A loader whose `#perform` will run on a background thread.
+    #
+    # It uses `Concurrent::Future` which uses a global thread pool.
+    class BackgroundLoader < GraphQL::Dataloader::Loader
+      def sync
+        future = Concurrent::Future.execute do
```
rmosolgo:
```suggestion
        Concurrent::Future.execute do
```

rmosolgo:
👌 Thanks for the reference there, @eapache 👀  I'll keep it in mind!

lib/graphql/dataloader/loader.rb
```diff
@@ -0,0 +1,46 @@
...
+        resolved_values = perform(keys_to_load)
+        keys_to_load.each_with_index do |k, i|
+          resolved_value = resolved_values[i]
+          @loaded_values[k] = resolved_value
+        end
+        nil
+      end
+
+      def perform(values)
+        raise NotImplementedError, "`#{self.class}#perform` should load `values` for `@key` and return an item for each of `values`"
```
eapache: graphql-batch has a separate method for fulfilling rather than relying on an array return value; not sure which one I prefer "from scratch", but not really looking forward to migrating :P

chrisbutcher:
> Traceable -- include graphql context with loads, so that a developer can see what GraphQL fields used which loaders.

This would be amazing. With the existing graphql-ruby + graphql-batch, I couldn't find an obvious or clean way to, for example, attributing time spent batch loading a given field with a given `ast_node`.

With the existing executor / lazy loading implementation, I suppose this is true of any field resolvers that return promises, but I wonder if batch loader perform start / end hook methods that can read/write context would help?

rmosolgo:
> read/write context

Yeah, I think that's part of the ticket: right now, loaders (ours, anyway) throw out graphql context (including current field, path, etc) and operate independently of it. I'd love to find a way that _keeps_ the low-overhead api of loaders but _adds_ context-awareness. Maybe we could tack on the context info _after_ the application's resolver (eg, if it returns a promise, tack context onto the promise) to make it _easy_.

lib/graphql/dataloader/loader.rb
rmosolgo: :+1: I think our base loader implements this API on top of that one. I'll make sure to retain both, or something.

eapache: A return array is probably easier to reason about in simple cases, but seems like you'd run into some hairiness when `values` (should that be `keys`?) contains duplicates, or there are a lot of keys that don't have a record in the DB, etc.

sfcgeorge:
I found the current ones a bit verbose so implemented a nice little DSL that looks like this:

```ruby
def self.authorized?(timeslot, context)
  return false unless super
  return false unless context.current_instructor

  chain_load(timeslot).offer.outlet.instructors.then do |_offer, _outlet, instructors|
    instructors.include?(context.current_instructor)
  end
end
```

So you can load a chain of one-to-one relationships and optionally a one-to-many on the end, and access any of the models along the way in the block.

https://gist.github.com/sfcgeorge/e067822f174d42175fec0f2264fe399e

rmosolgo:
Nice, thanks for sharing! We have some similar shortcuts in github/github. I like the approach of adding methods like `chain_load`, and that might be an option for sneaking in `context` without making the user-facing API to burdensome.

lib/graphql/dataloader/loader.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def initialize(context, key)
+        @context = context
+        @key = key
+        @promises = {}
+        @loaded_values = {}
+      end
+
+      def load(value)
+        @promises[value] ||= GraphQL::Execution::Lazy.new do
```
panthomakos: Is it possible that you will end up with two lazy executions for the same value based on how threads are scheduled? You might consider using a https://ruby-concurrency.github.io/concurrent-ruby/1.1.4/Concurrent/Map.html.

lib/graphql/dataloader/loader.rb
panthomakos: I may not fully understand this implementation, but this does not seem to allow for multiple background loaders to be running IO at the same time. Do you not need to wrap the `perform` call in some sort of future/thread as well?

lib/graphql/dataloader/background_loader.rb
```diff
@@ -0,0 +1,32 @@
...
+      def sync
+        Concurrent::Future.execute do
+          setup_thread
+          super
+          teardown_thread
+        end
+      end
+
+      # Implement this method to prepare any thread state that your
+      # application depends on, for example, `Thread.current[:user] = @context[:current_user]`
```
panthomakos: I do not follow how this is helpful. If this method is running in a separate thread anyways, why would the thread not access `@context[:current_user]` directly?

lib/graphql/dataloader/loader.rb
```diff
@@ -0,0 +1,46 @@
...
+          if !@loaded_values.key?(value)
+            sync
```
panthomakos: In the concurrent example, `sync` will run in a separate thread. Correct?

If so, then how do you guarantee that `@loaded_values[value]` will be present on line 26 below?

lib/graphql/dataloader/loader.rb
rmosolgo: I switched to a fulfill-based API like graphql-batch has, thanks again for taking a look. 

I think you're right that the Future is in the wrong place. I'll take a closer look at that soon.

lib/graphql/dataloader/background_loader.rb
rmosolgo: I've worked on applications which depend on context like 

```ruby
class User 
  def self.current 
    Thread.current[:user]
  end 

  def self.current=(u)
    Thread.current[:user] = u 
  end 
end 
```

And I wanted to support a case like that. (Just for legacy code purposes.) But in fact, it should probably go in `#perform` anyways, so I've removed those methods!

lib/graphql/dataloader/loader.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module GraphQL
+  class Dataloader
+    class Loader
+      class StateError < GraphQL::Error; end
+
+      def self.load(context, key, value)
+        dl = context[:dataloader]
+        loader = dl.loaders[self][key]
```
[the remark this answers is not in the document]
rmosolgo: Yes, i think there's got to be some better API for this. But I'd really like to keep dataloading `context`-aware so that we can trace it as part of the GraphQL request.

lib/graphql/dataloader/loader.rb
```diff
@@ -0,0 +1,46 @@
...
+        loader.load(value)
+      end
+
+      def initialize(context, key)
+        @context = context
+        @key = key
```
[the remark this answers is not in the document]
rmosolgo: Yes, I suppose it could be more readable like:

```ruby 
class MyLoader < GraphQL::Dataloader::Loader
  def initialize(context, model)
    @model = model 
    super 
  end 

  def perform(ids)
    @model.where(id: ids)
  end
end
```

This was _required_ for graphql-batch (IIRC) because it didn't store state otherwise. It also didn't require the `super` call. I wonder how I can remove that boilerplate in this implementation 🤔

lib/graphql/dataloader/loader.rb
```diff
@@ -0,0 +1,142 @@
+# frozen_string_literal: true
+
+module GraphQL
+  class Dataloader
+    class Loader
+      # TODO: this is basically a reimplementation of Promise.rb
+      # Should I just take on that dependency, or is there a value in a
+      # custom implementation?
```
swalkinshaw: @dylanahsmith curious about your opinion on this

dylanahsmith: I don't think it matters that much to Shopify, since we mostly just use promises and graphql-batch as an implementation concern for GraphQL batch loading.  A custom implementation might allow it to be implemented it a more performant way for this specific use case.

The downside would be if anyone wants to build something that only depends on a promise gem so that it doesn't force a dependency on the graphql gem where it isn't needed, while still being able to have it integrate well with the graphql gem where needed.  It seemed like github was using the promise library more broadly and also had an interest in decoupling batch loaders from graphql.

It would be convenient to be able to share code for batch loading between a graphql implementation and other areas of an application.  We just haven't committed to promises as a way of doing that.

swalkinshaw: And thankfully promise-rb is a lightweight gem without any dependencies of its own. I'd be most interested in performance optimizations as you mentioned. Anecdotally I remember hearing someone from GitHub had some some non-upstreamed improvements to promise-rb. @rmosolgo do you know if that's true? There might be some learnings for this implementation as well.

rmosolgo: Yes, that's right. It's been a _while_ since I thought about it, but we're on this commit: https://github.com/lgierth/promise.rb/commits/4fd5539

rmosolgo:
I'm going to release 1.10 without this in the interest of time. That branch has other big changes on it already, and there's still a lot of work to do here, and I haven't gotten to it. And it isn't _essential_ -- graphql-batch works great and you can build backgrounded IO on top of it AFAIK.

rmosolgo:
> https://github.com/Shopify/graphql-metrics/blob/master/lib/graphql_metrics/timed_batch_executor.rb

It looks like, in the ... 😓 _year_ ... since I started this, this file has gone away. What's y'all's current executor situation, do you still use those hooks?

swalkinshaw:
> What's y'all's current executor situation, do you still use those hooks?

Nope! That should make things easier I hope? That library uses a combination of tracing and instrumentation now

lib/graphql/execution/interpreter/resolve.rb
```diff
@@ -25,8 +25,10 @@ def self.resolve_all(results)
         # or return {Hash}/{Array} if the query should be continued.
         #
         # @param results [Array]
-        # @return [Array] Same size, filled with finished values
+        # @return [Array] The next round of lazies to resolve
         def self.resolve(results)
+          # First, kick off any loaders that will resolve in background threads
+          Dataloader.current && Dataloader.current.process_async_loader_queue
```
rmosolgo, line 31: This isn't exactly subtle, but after a lot of attempts, I couldn't find a better way to work in a "kick off" step into the existing execution flow. This is very similar to the [original suggestion](https://github.com/rmosolgo/graphql-ruby/pull/1981/files#r238053370), but at the loader level instead of the promise level.

lib/graphql/dataloader.rb
```diff
@@ -0,0 +1,170 @@
...
+        end
+      end
+
+      def current
+        Thread.current[:graphql_dataloader]
+      end
+
+      def current=(dataloader)
+        Thread.current[:graphql_dataloader] = dataloader
+      end
```
rmosolgo, line 119: I think this _adds_ the requirement that GraphQL queries be executed within a single thread.

lib/graphql/dataloader/loader.rb
rmosolgo: (I ended up merging this into GraphQL::Execution::Lazy, but I'll have to do a round of performance assessment before I this is _done_.)

lib/graphql/dataloader.rb
rmosolgo: (The alternative would be to use `context[:dataloader]`, which earlier iterations used. But then you're stuck with, how to get that `dataloader`  into each loader, so that the loader can register itself with the dataloader's cache.)

swalkinshaw:
One small issue we've had with batch loaders is dealing with keys (or fetch parameters in your terminology?) that don't have data to fulfill. I think right now it leads to opaque `Promise::BrokenError` exceptions. Here's an example:

```ruby
class ProductImageLoader < GraphQL::Batch::Loader
  def initialize(shop_id)
    @shop_id = shop_id
  end

  def perform(product_ids)
    Images
      .where(shop_id: @shop_id, product_id: product_ids)
      .group_by(&:product_id)
      .each { |product_id, images| fulfill(product_id, images) }
  end
end
```

If no images exist for a product id, it won't be fulfilled leading to that error. We have two common solutions:

1. manually fulfill all unfulfilled keys
```ruby
class ProductImageLoader < GraphQL::Batch::Loader
  def initialize(shop_id)
    @shop_id = shop_id
  end

  def perform(product_ids)
    Images
      .where(shop_id: @shop_id, product_id: product_ids)
      .group_by(&:product_id)
      .each { |product_id, images| fulfill(product_id, images) }

    product_ids.each { |id| fulfill(id, nil) unless fulfilled?(id) } # nil here, but any "default" value works
  end
end
```

2. iterate over the keys/fetch parameters instead of the data
```ruby
class ProductImageLoader < GraphQL::Batch::Loader
  def initialize(shop_id)
    @shop_id = shop_id
  end

  def perform(product_ids)
    images = Images
      .where(shop_id: @shop_id, product_id: product_ids)
      .group_by(&:product_id)

    product_ids.each do |id|
      fulfill(id, images[id])
    end
  end
end
```

I've had the idea before that we could have a better interface to enforce/prevent this situation. A dataloader *could* declare a default value explicitly. If set, the loader could automatically fulfill all missing keys with it?

But thinking more about this, `perform` has a fairly strict requirement that `fulfill` gets called for each of its fetch parameters, yet we have to manually do that work which is error prone (as seen above). I wonder if that's a better interface we could give people instead 🤔 I'll give it more thought

rmosolgo:
> in your terminology?

I started updating those docs and realize I didn't have a good word for those different kinds of _keys_. Now I see that Batch::Loader uses `group_args` and `keys`,  seems good too. I just want to pick something that makes their usage clear. 

Also, I'm torn between doubling down on the terms from graphql-batch and batch-loader, or picking new words to make it more googleable. Oh, and avoiding "-er" classes (http://wiki.c2.com/?DontCreateVerbClasses). 

> declare a default value explicitly

Yeah, I could see that, something like 

```ruby 
unfulfilled_default nil 
``` 

Then the library could basically do 

```ruby
if self.class.set_unfulfilled_default?
  keys_to_load.each { |key| fulfill(key, self.class.unfulfilled_default) unless fulfilled?(key) }
end 
```

But probably only if `unfulfilled_default` _was_ explicitly set, otherwise we'd raise an error of some kind. (Because I don't think we want it to silently ignore unfulfilled keys, that's important feedback to the application behavior.)

gaffney:
Hey @rmosolgo thanks for the awesome library and hard work; we are currently using this in production without any issues.

> If anyone has other suggestions for a built-in dataloader, please share them!

Our team decided to go with [exAspArk/batch-loader](https://github.com/exaspark/batch-loader) due to its generic nature / the fact that it is not tied to GraphQL. We have plenty of external REST API calls we wanted to batch in addition to GraphQL fields.

`batch-loader` works great for us, but it is a little painful integrating with `graphql-ruby`, as illustrated by the [`graphql-ruby` example in the README](https://github.com/exaspark/batch-loader#graphql-example):

> To avoid this problem, all we have to do is to change the resolver to return BatchLoader::GraphQL (#32 explains why not just BatchLoader):

I found that there were several issues around this and very long threads in 2018, but unfortunately [this is where we landed](https://github.com/exAspArk/batch-loader/pull/32):

> I [suggested](https://github.com/rmosolgo/graphql-ruby/pull/1784#issuecomment-417736649) a few other potentially more flexible solutions for graphql-ruby to detect lazy objects such as duck typing or using explicit arguments. But it looks like it won't be implemented. To fix the issue BatchLoader started wrapping BatchLoader objects with PORO (plain old ruby objects) by using graphql-ruby [instrumentation](https://github.com/rmosolgo/graphql-ruby/blob/75c3529177474c8e65fd175f0c36fdc2fac185b1/guides/fields/instrumentation.md).

Since you are hard at work at a major refactor I was wondering if you could consider revisiting some of the solutions proposed by [exAspArk](https://github.com/exaspark) to make for a more seamless integration... or any alternative to avoid `BatchLoader::GraphQL.for`. The current batching workaround makes testing painful and is inherently inextensible.

rmosolgo:
Thanks for sharing that discussion, @gaffney. Unfortunately, this refactor doesn't touch the underlying behavior where GraphQL-Ruby detects lazy values by calling `value.class`. Last I checked, Batch-Loader objects implement `#class` by delegating to the batch-loaded object (instead of returning `BatchLoader`), so it just _doesn't work_. 

Interestingly, the original suggestion in those issues was to add a `lazy: true` configuration to `field(...)`. That would be possible now, something like: 

```ruby 
  # When `lazy: true` is given, add a field extension to wrap the returned value of this field
  def initialize(*args, **kwargs, lazy: false, &block)
    if lazy 
      extensions = kwargs[:extensions] ||= [] 
      extensions << BatchLoaderExtension 
    end 
    super
  end 
end 

# When this extension is added, the field was configured with `lazy: true`, so apply a wrapper so 
# GraphQL-Ruby can identify the lazy object. 
class BatchLoaderExtension 
  def resolve(object:, arguments:, **_rest)
    # call normal field execution 
    return_value = yield(object, arguments)
    # apply a wrapper and return it, TODO is `.wrap` the correct method here? Not exactly sure. 
    BatchLoader::GraphQL.wrap(return_value)
  end 
end 
```

Anyways, just a thought after reviewing that code for the first time in a while. Interestingly, a recent Ruby version added [`Object#then`](https://ruby-doc.org/core-2.7.0/Object.html#method-i-then) which I could imagine using as the basis for a duck-typing approach to batching and lazy evaluation. But that's another matter than what's in the works here 🍻 .

rmosolgo:
I ended up going a very different direction on this: #3264

PR #2487, opened 2019-09-20
Change misleading field name [ci skip]

vfonic:
I just thought that `created_at`, which is probably in 100% of the apps a DateTime rather than Date, would be slightly misleading, so I changed it to something else.


rmosolgo:
Thanks for the clarification!

PR #2491, opened 2019-09-24
Remove errant frozen_string_literal directive.

nirvdrum:
The directive already appears higher in the file, so it is unnecessary here. Moreover, it is ignored if any tokens have been seen in the file before the directive is encountered.


rmosolgo:
Oops! Thanks for the fix. Maybe those requires were added in the wrong place, then `rubocop -a` was a little _too_ helpful :P

nirvdrum:
No problem. I only noticed because I kept getting a warning message about the second directive being ignored.

PR #2493, opened 2019-09-25
Fix URL in docs

daido1976:
Since the current URL is old and redirected to the top page, the path and anchor does not work, so fix the URL.


rmosolgo:
Thanks for the correction! It will be published soon.

PR #2503, opened 2019-09-30
Fix case statement

tenderlove:
The `build` method is only ever called with Arrays, so `onlies` can never be
equal to `0`, and if `onlies` == `1`, then `1[0]` wouldn't work, so I
think we can safely remove this code.

Was this supposed to be `onlies.length`?

### DIFF
--- a/lib/graphql/filter.rb
+++ b/lib/graphql/filter.rb
@@ -33,14 +33,7 @@ module GraphQL
       end
 
       def self.build(onlies)
-        case onlies
-        when 0
-          nil
-        when 1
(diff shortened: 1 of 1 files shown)


rmosolgo:
> supposed to be `onlies.length`

🤦‍♂  yes, I'm sure it was. Is it worth implementing that way?

rmosolgo:
Fixed on 1.10-dev in https://github.com/rmosolgo/graphql-ruby/commit/ae6068e5af61874db59565e17c20f4067f9ddd84

Issue #2451, opened 2019-08-29
All GraphQL queries fail with JRuby 9.2.8.0 


jessicaspeir:
I'm trying to upgrade my app from JRuby 9.2.6.0 to 9.2.8.0 (we had to skip 9.2.7.0 because of several bugs). Everything works great on 9.2.6.0 but on 9.2.8.0 every query (including the queries to get the schema fail with this error:

```
yield called out of block :
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/field.rb:623:in `with_extensions' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/field.rb:598:in `public_send_field' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/field.rb:456:in `block in resolve_field' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema.rb:1206:in `after_lazy' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/field.rb:440:in `resolve_field' &
org/jruby/RubyMethod.java:132:in `call' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/member/instrumentation.rb:80:in `call' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/field.rb:248:in `resolve' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/execution/execute.rb:321:in `call' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/middleware_chain.rb:49:in `invoke_core' &
C:/Tools/JRuby/jruby-9.2.8.0/lib/ruby/gems/shared/gems/graphql-1.9.9/lib/graphql/schema/middleware_chain.rb:38:in `invoke' &
[...]

PR #2504, opened 2019-09-30
Skip visitor hooks entirely if possible

jhawthorn:
Similar to #2451

The visitor code is run a lot. Previously we could call `begin_visit`/`end_visit` even if there was no visitor configured for the class, this would end up allocating an empty NodeVisitor (3 objects) and inserting it into the hash.

Skipping all this should reduce allocations and avoiding the hash insertions and extra method calls should be faster.

Versus #2451, this reduces another 1.5 million allocations from our app boot (about 4%).

### DIFF
--- a/lib/graphql/language/visitor.rb
+++ b/lib/graphql/language/visitor.rb
@@ -96,8 +96,8 @@ module GraphQL
         else
           # Run hooks if there are any
           new_node = node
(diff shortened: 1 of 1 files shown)


rmosolgo:
Awesome! 🎉

PR #2505, opened 2019-10-01
Disallow casting of Enums into scalar values

### LINKED ISSUE #2371: Enum values should not be accepted as IDs
Hi!

Using the [example types of the documentation](https://graphql-ruby.org/getting_started):
```ruby
  field :post, PostType, null: true do
  end

  def post(id:)
    "Received request with id: #{id}"
  end
end
```
When using an argument of type ID, a query needs no quotation marks around the ID-value:
```ruby
{ post(id: 1) }

=> Received request with id: 1
```
This also holds true when characters are used as the ID:
```ruby
{ post(id: abc) }

=> Received request with id: abc
```
However, if the ID starts with a number followed by characters, we get an error:
```ruby
{ post(id: 1abc) }

=> #<GraphQL::Query::Result @query=... @to_h={"errors"=>[{"message"=>"Parse error on \"'\" (error) at [1, 15]", "locations"=>[{"line"=>1, "column"=>15}]}]}>
```
We can fix this, if we wrap quotation marks around the value:
```ruby
{ post(id: "1abc") }

=> Received request with id: 1abc
```

Is this behavior expected? If yes, is this documented somewhere?

daemonsy:
## What's up?

Happy Hacktoberfest 🎃 ! This resolves #2371. 

In general, it seems that we allow raw enum values to be coerced into strings, code below. 

```ruby
# in lib/graphql/scalar_type.rb
def raw_coercion_input(value)
  if value.is_a?(GraphQL::Language::Nodes::InputObject)
    value.to_h
  elsif value.is_a?(Array)
    value.map { |element| raw_coercion_input(element) }
  elsif value.is_a?(GraphQL::Language::Nodes::Enum)
    value.name
  else
    value
  end
end
```

This means that enum's value (#name) will be taken as a `string` to go through input coercion. But it seems that enums are not scalars, neither string or number from the spec. 

Hence I attempted a solution to reject enums as scalars. 

During validation, this returns errors such as the below: 

```ruby
"message"=>"Argument 'id' on Field 'milk' has an invalid value. Expected type 'ID!'.", 
"locations"=>[{"line"=>1, "column"=>17}], "path"=>["query getMilk", "milk", "id"], 
"extensions"=>{"code"=>"argumentLiteralsIncompatible", "typeName"=>"Field", "argumentName"=>"id"}}
```

### DIFF
--- a/lib/graphql/scalar_type.rb
+++ b/lib/graphql/scalar_type.rb
@@ -117,8 +117,6 @@ module GraphQL
         value.to_h
       elsif value.is_a?(Array)
         value.map { |element| raw_coercion_input(element) }
-      elsif value.is_a?(GraphQL::Language::Nodes::Enum)
-        value.name
       else
         value
       end
@@ -126,7 +124,7 @@ module GraphQL
 
     def validate_non_null_input(value, ctx)
       result = Query::InputValidationResult.new
-      if coerce_non_null_input(value, ctx).nil?
+      if value.is_a?(GraphQL::Language::Nodes::Enum) || coerce_non_null_input(value, ctx).nil?
         result.add_problem("Could not coerce value #{GraphQL::Language.serialize(value)} to #{name}")
       end
       result
--- a/spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
+++ b/spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
@@ -138,7 +138,7 @@ describe GraphQL::StaticValidation::ArgumentLiteralsAreCompatible do
   end
 
   describe "using enums for scalar arguments it adds an error" do
-    let(:query_string) { <<~GRAPHQL
+    let(:query_string) { <<-GRAPHQL
       {
         cheese(id: I_AM_ENUM_VALUE) {
           source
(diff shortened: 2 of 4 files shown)


spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
```diff
@@ -137,6 +137,40 @@
...
+      {
+        "message" => "Argument 'id' on Field 'cheese' has an invalid value. Expected type 'Int!'.",
+        "locations" => [{ "line" => 2, "column" => 3 }],
+        "path"=> ["query", "cheese", "id"],
+        "extensions"=> { "code" => "argumentLiteralsIncompatible", "typeName" => "Field", "argumentName" => "id" }
+      }
+    end
+
+    it "works with error bubbling disabled" do
+      without_error_bubbling(schema) do
```
daemonsy: I'm not sure what error bubbling does, just monkey see monkey do :)

lib/graphql/scalar_type.rb
```diff
@@ -117,16 +117,14 @@ def raw_coercion_input(value)
         value.to_h
       elsif value.is_a?(Array)
         value.map { |element| raw_coercion_input(element) }
-      elsif value.is_a?(GraphQL::Language::Nodes::Enum)
```
daemonsy: With this, enums will never be treated as a scalar anymore. Practically, only ID and String types are affected as the casting for numeric types should reject this

rmosolgo:
Thanks for digging in on this! I was afraid how complicated it might turn out to be, but I'm glad you found a straightforward solution.

daemonsy:
Thanks @rmosolgo! This might affect some people who are relying on parsing enum into strings 🤞

PR #2508, opened 2019-10-02
Stringify error extensions (fixes #2246)

### LINKED ISSUE #2246: ExecutionError returns a Hash that can have string and symbols keys
A `GraphQL::ExecutionError` can be created with options and/or extensions which may contain a `Hash` that uses symbolized keys. Calling `to_h` will then return a `Hash` in which some keys are strings and some are symbols. Granted, the choice of using symbols vs strings is up to the caller.  However, I don't believe that the caller should have to know that the implementation of `ExecutionError.to_h` happens to prefer strings.  So either:

`ExecutionError.to_h` should stringify or symbolize all keys and document which one was used. OR
`ExecutionError.to_h` should return a `HashWithIndifferentAccess `

Some additional thoughts:

- I realize this issue is a bit of a nit.
- This is not an issue when the GraphQL query is executed remotely and the result is serialized/deserialized during transmission over the wire.  However, it is an issue when the query is executed locally.
- The string-vs-symbol-keys-in-hashes issue is almost as old as Ruby
- There are probably other graphql-ruby classes that can be converted to Hashes that suffer from the same problem.  I just happened to run into the `ExecutionError` issue while writing a few specs that tested the errors result.

cmatheson:


rmosolgo:
Awesome, thanks for the fix!

PR #2510, opened 2019-10-03
Fix parsing of escaped backslashes

### LINKED ISSUE #2506: Parse error when non-last argument ends with two backslashes (\\)
graphql generates parse error response when:
* field has multiple arguments
* non-last string argument has double backslashes in the end (example: "abc\\\\")

Everything works fine if backslashes are  in last argument or backslashes are not at the end of string (this will not generate parse error: "ab\\\\c"). Also it works fine if I pass values via variables

## The proof

I have a new project and I have QueryType which looks like this:

```ruby
module Types
  class Item < Types::BaseObject
    description "A list of items which may be completed"

    field :text, String, null: true
    field :other_text, String, null: true
  end

    field :item, Item, "test item", null: false do
      argument :text, String, "Restrict items to this status", required: false
      argument :other_text, String, "Restrict items to this statuss", required: false
    end

    def item(**kwargs)
      kwargs
    end
  end
end
```

if I make query like this:
```graphql
query {
  item(text: "a", otherText: "b\\") {
    text
    otherText
  }
}
```

I will get response:
```json
{
  "data": {
    "item": {
      "text": "a",
      "otherText": "b\\"
    }
  }
}
```
So far everything works just fine.

But if I change the order or arguments and do query like this:
```graphql
query {
  item(otherText: "b\\", text: "a") {
    text
    otherText
  }
}
```

I will get:
```json
{
  "errors": [
    {
      "message": "Parse error on \"\\\"\" (error) at [31, 34]",
      "locations": [
        {
          "line": 31,
          "column": 34
        }
      ]
    }
  ]
}
```

Reproduced with graphql v1.9.12 and also with master code.
Example project where you can check this behavior: https://github.com/povilasjurcys/graphql_failure_proof

andrewdotn:
Fixes #2506

The given repro case did not seem to trigger for the block-quote case, so I
left the definition of that alone.

(Hacktoberfest submission)


andrewdotn:
Before:

before

After:

after

daemonsy:
Just wanted to say :heart: that this came in as a Hacktoberfest submission

povilasjurcys:
amazing :+1:

rmosolgo:
Awesome, thanks for the fix here!!

PR #2524, opened 2019-10-08
Add nullify session CSRF protection

jamonholmgren:
This change allows for outside API access (say, from a mobile app) while continuing to have protections against CSRF attacks.

Essentially, if the request comes from outside of the current domain, it'll nullify the session. This means that you'll have to get the current_user or other session data some other way (usually a token on the request), which is safer than just checking your current browsing session.

See more here: https://marcgg.com/blog/2016/08/22/csrf-rails/

### DIFF
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -3,7 +3,7 @@ class GraphqlController < ApplicationController
   # This allows for outside API access while preventing CSRF attacks,
   # but you'll have to authenticate your user separately
   protect_from_forgery with: :null_session
-                                               
+
   def execute
     variables = ensure_hash(params[:variables])
     query = params[:query]
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -1,4 +1,9 @@
 class GraphqlController < ApplicationController
+  # If accessing from outside this domain, nullify the session
+  # This allows for outside API access while preventing CSRF attacks,
+  # but you'll have to authenticate your user separately
+  protect_from_forgery with: :null_session
+                                               
   def execute
     variables = ensure_hash(params[:variables])
     query = params[:query]
--- a/spec/integration/rails/generators/graphql/install_generator_spec.rb
+++ b/spec/integration/rails/generators/graphql/install_generator_spec.rb
@@ -176,6 +176,11 @@ RUBY
 
   EXPECTED_GRAPHQLS_CONTROLLER = <<-'RUBY'
 class GraphqlController < ApplicationController
+  # If accessing from outside this domain, nullify the session
+  # This allows for outside API access while preventing CSRF attacks,
+  # but you'll have to authenticate your user separately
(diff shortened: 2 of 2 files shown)


rmosolgo:
Awesome, thanks 🍻 ! It looks like the only thing is to update this test which checks the output of the generator: 

https://github.com/rmosolgo/graphql-ruby/blob/c3fa411cc376ff39046db83ef80ffd923624b099/spec/integration/rails/generators/graphql/install_generator_spec.rb#L178-L179

Do you mind updating that too?

daemonsy:
`null_session` is a good TIL for me. 

I was wondering from another point of view. Since this controller is a template that inherits from `ApplicationController` where the CSRF handling strategy is usually set, might be better not to introduce an override by mistake over here?

jamonholmgren:
@rmosolgo Heh, your tests were more robust than I expected. :laughing: 

@daemonsy I think it belongs here because nullifying the session is mostly useful for API endpoints and less so for the other controllers.

daemonsy:
@jamonholmgren :+1: agree on the use case you mentioned. 

My point was more about not making this a default choice for the user in the generator template. Thanks for linking the post, it explains the different strategies well. 

> The null_session strategy is the default one. The gotcha here is that by default Rails 5 generates the ApplicationController file with the exception strategy, but the default strategy inside ActionPack is actually the null_session one, which can be confusing.

> It is very important to keep in mind that fraudulent requests will go through with the null_session and reset_session strategies. The only action taken by these strategies is to make sure the session is empty during the request.

Since it's already a default but made confusing by the Rails generator, I feel that there's no need to generate one more default in the generator. In the case that the user doesn't notice it, this might end up also end up changing the CSRF behavior. 

What is the basis of us inheriting from the standard `ApplicationController` and also taking a stance that the CSRF handling can be a nulling the session?

Personally after using the generator a few times across apps (and 2 jobs),  I find that there will be a need to customize the controller semantics when using session-less authentication anyway. I imagine if an existing app already supports header and cookie auth, that will be already done in `ApplicationController` using `null session`?

Anyway all in all, this is still just my personal opinion, feel that we shouldn't override the CSRF handling in a generated controller.

jamonholmgren:
@daemonsy Good points. I think overall this is a "failsafe" approach -- we are overriding back to the default, and essentially the worst that can happen is that a query will fail. The reason to add this is it's a gotcha for people connecting to it from a mobile app or separate domain. It took a bit of digging to figure out what was the recommended approach.

I think compelling cases can be made both ways by reasonable people. I'll leave it to Robert to decide which direction makes the most sense. ;-)

daemonsy:
Happy Friday! Agree and I got no skin in the game, was just an opinion. So which ever way it goes :smile:

rmosolgo:
Thanks for sharing your thoughts on this, folks. I'm interested in improving getting-started time, and for someone running an unauthenticated query in GraphiQL, this will make it work out of the box. At the same time, for someone doing "proper" authentication, they can include the CSRF token from Rails's head tag, and it will _also_ work fine.  So, let's give it a try and see what people say 😈

rmosolgo:
Thanks @jamonholmgren !

PR #2529, opened 2019-10-09
Guides: Repurpose class API guide 

daemonsy:
## What's up? 

Was going through the Guides with a colleague and the Class-based API guide made it sound like it's a new feature where it's actually the new norm now. 

Suggesting to rename it as the migration guide and put it at the bottom of the page. 

The thought is that it remains available for people who are still using the define DSL but does not confuse people who are new to graphql-ruby. 

### Reordered Schema Section
Screen Shot 2019-10-09 at 10 58 24 AM

### Class migration guide with deprecation message
Screen Shot 2019-10-09 at 11 06 47 AM

### DIFF
--- a/guides/schema/lazy_execution.md
+++ b/guides/schema/lazy_execution.md
@@ -4,7 +4,8 @@ doc_stub: false
 search: true
 title: Lazy Execution
 section: Schema
-desc: Resolve functions can return "unfinished" results. GraphQL will defer finishing them until other fields have been resolved.
+desc: Resolve functions can return "unfinished" results that are deferred for batch resolution.
+index: 4
 ---
 
 With lazy execution, you can optimize access to external services (such as databases) by making batched calls. Building a lazy loader has three steps:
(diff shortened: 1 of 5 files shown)


guides/schema/root_types.md
```diff
@@ -26,16 +27,16 @@ The types are `GraphQL::Schema::Object` classes, for example:
 ```ruby
 # app/graphql/types/query_type.rb
 class Types::QueryType < GraphQL::Schema::Object
-  # ...
+  field :all_posts, [PostType], 'Returns all blog posts', null: false
```
daemonsy: Small tactical changes to make the root type document a little more useful. 

Personally, I feel like this page doesn't add enough. It might be good to expand it or push the examples into the "Schema Definition" guide.

guides/schema/lazy_execution.md
```diff
@@ -4,7 +4,8 @@ doc_stub: false
 search: true
 title: Lazy Execution
 section: Schema
-desc: Resolve functions can return "unfinished" results. GraphQL will defer finishing them until other fields have been resolved.
+desc: Resolve functions can return "unfinished" results that are deferred for batch resolution.
```
daemonsy: Just a little re-write to make the description fit into one line on the menu :smiley_cat:

rmosolgo:
Awesome, I agree this makes it a lot clearer! Thanks for improving this for the next reader 🍻

Issue #2283, opened 2019-05-13
Better control of pagination arguments


cmatheson:
From the [Relay Connections Spec](https://facebook.github.io/relay/graphql/connections.htm#sec-Arguments):

> A field that returns a Connection Type must **include forward pagination arguments, backward pagination arguments, or both**. These pagination arguments allow the client to slice the set of edges before it is returned.

Currently graphql-ruby always provides arguments for **both** directions of pagination (which makes a lot of sense for the pagination backends provided by the gem).  I've added a dynamodb backend for a project I'm working on which only supports forward pagination.

I'm interested in adding support for opting out of unwanted pagination arguments but I don't know where it should live (it would be easiest on `field` (which already has the `connection` (true/false) arg), but maybe that is too clunky?)

PR #2534, opened 2019-10-10
wip: support pagination in a single direction only

cmatheson:
Hello,

I'm not super confident that this is the best approach, but I thought I would at least try to get some feedback.  Issue #2283 explains the basic rationale for this PR.

The actual use case I have for this is that I'm hitting a data store (dynamodb) that doesn't work with backwards pagination, so in an ideal world I would have some way to disable a pagination direction based on what class is handling the connection (but I realize we don't have that info at schema build time).

Anyway, let me know your thoughts and I will change this accordingly.


rmosolgo:
Hi, sorry for the slow review here. I think the solution you posted here should work fine for the issue you posted in #2283! However, it's pretty context-specific, so instead of building it into graphql-ruby, I've made it so you can customize `connection_extension` in your base field class over in #2561. 

That way, you could write your own subclass of `ConnectionExtension`, with the improvements you added here, and attach it to your schema. And if you want some other improvements in the future, you don't have to wait for the lazy graphql-ruby maintainer to finally review your PR!

PR #2535, opened 2019-10-10
Add some links to libs and articles by @evilmartians

DmitryTsepelev:
Added some links to articles and libs created by @evilmartians team


rmosolgo:
Awesome, thanks for adding these! I hadn't seen the one about handling non-null errors; we do the same thing at GitHub. I wonder if that's a sign that graphql-ruby should handle them differently.

PR #2538, opened 2019-10-11
Docs: add link to `graphql_rails` in related projects

povilasjurcys:
`graphql_rails` allows to build graphql types using already known patters from Rails: models, controllers, routers and etc.
Project is quite stable: it is maintained for more then a year and had 18 releases so far. 

I hope that's enough to treat this project as `related` and to mention it in "related projects" :)


rmosolgo:
Thanks!

PR #2543, opened 2019-10-13
Issue#2067 Replace NotImplementedError with RequiredImplementationMissingError

### LINKED ISSUE #2067: Change NotImplementedError to something else
Perhaps I'm misunderstanding something, but I'd like to suggest that this gem move away from using `NotImplementedError`.

Take a look at the docs: http://ruby-doc.org/core-2.6/NotImplementedError.html. At issue is the OS and particular ruby implementation. You're supposed to raise that error if the current ruby or OS can't do something you need it to do. NOT if a subclass hasn't implemented an expected method.

Why is this a problem? Because `NotImplementedError` does not subclass `StandardError`, which is surprising. It subclasses `ScriptError`. Thus, normal rescuing (either explicitly or implicitly rescuing `StandardError`) won't work, and we end up rescuing things we honestly shouldn't.

Thus, I'd propose that all usage of `NotImplementedError` be changed to something else, or perhaps simply removed, if that would cause a semantically meaningful error to be raised (like a `NoMethodError`). At the very least, use something subclassed under a module, so it can be rescued specifically. And subclass that module under `StandardError`, so that rescuing all `StandardError` will work as expected, or even better, `NoMethodError`, since this is really a particular type of `NoMethodError`.

See discussion at:
http://chrisstump.online/2016/03/23/stop-abusing-notimplementederror/
https://www.reddit.com/r/ruby/comments/4cal7v/please_stop_abusing_notimplementederror/
https://airbrake.io/blog/ruby-exception-handling/notimplementederror
https://github.com/rubocop-hq/ruby-style-guide/issues/458
https://stackoverflow.com/questions/13668068/how-to-signal-not-implemented-yet

vaihtovirta:
Fixes #2067

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -66,7 +66,7 @@ module GraphQL
 
         # @return [Symbol] the method to call on {Language::Visitor} for this node
         def visit_method
-          raise NotImplementedError, "#{self.class.name}#visit_method shold return a symbol"
+          raise GraphQL::RequiredImplementationMissingError, "#{self.class.name}#visit_method shold return a symbol"
         end
 
         def position
--- a/lib/generators/graphql/templates/schema.erb
+++ b/lib/generators/graphql/templates/schema.erb
@@ -24,7 +24,7 @@ class <%= schema_name %> < GraphQL::Schema
   def self.resolve_type(type, obj, ctx)
     # TODO: Implement this function
     # to return the correct type for `obj`
-    raise(NotImplementedError)
+    raise(GraphQL::RequiredImplementationMissingError)
   end
 <% end %><% if options[:batch] %>
   # GraphQL::Batch setup:
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -10,6 +10,9 @@ module GraphQL
   class Error < StandardError
   end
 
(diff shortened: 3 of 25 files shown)


lib/graphql/language/nodes.rb
```diff
@@ -66,7 +66,7 @@ def initialize_copy(other)
 
         # @return [Symbol] the method to call on {Language::Visitor} for this node
         def visit_method
-          raise NotImplementedError, "#{self.class.name}#visit_method shold return a symbol"
+          raise GraphQL::RequiredImplementationMissingError, "#{self.class.name}#visit_method shold return a symbol"
```
rmosolgo:
```suggestion
          raise GraphQL::RequiredImplementationMissingError, "#{self.class.name}#visit_method should return a symbol"
```

Unrelated to your change, but I couldn't help noticing it! 🙈

rmosolgo: Nevermind, I thought I could commit changes from the comment, but it's not working 🤷‍♂

vaihtovirta: Sorry, I didn't notice this typo 🤦‍♂ 
I can prepare a fix for that as well 😄

Issue #2332, opened 2019-06-14
RelationConnection.has_next_page could use EXISTS


jscheid:
The query to determine whether there is more data (for `RelationConnection.has_next_page`) could be sped up by using an [exists?](https://apidock.com/rails/ActiveRecord/FinderMethods/exists%3F) subquery instead of counting all rows, especially for result pages near the front (that are arguably more important than later pages).

Below some measurements on a table with ~800K records (PostgreSQL 9.6).

```
psql# explain analyze select count(*) from xxxxxxx;
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                 QUERY PLAN                                                                                  │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Aggregate  (cost=108026.08..108026.13 rows=1 width=8) (actual time=4149.045..4149.045 rows=1 loops=1)                                                                       │
│   ->  Index Only Scan using index_xxxxxxx_on_xxxxxxx_xxxxxxx_xxxxx on xxxxxxx  (cost=0.42..106127.91 rows=759267 width=0) (actual time=0.086..4092.122 rows=759267 loops=1) │
│         Heap Fetches: 611238                                                                                                                                                │
│ Planning time: 0.227 ms                                                                                                                                                     │
│ Execution time: 532.649 ms                                                                                                                                                  │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(5 rows)

psql# explain analyze select exists(select 1 from xxxxxxx offset 10000);
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                   QUERY PLAN                                                                                    │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Result  (cost=1398.33..1398.38 rows=1 width=1) (actual time=7.427..7.427 rows=1 loops=1)                                                                                        │
│   InitPlan 1 (returns $0)                                                                                                                                                       │
│     ->  Limit  (cost=1398.19..106127.91 rows=749267 width=4) (actual time=7.425..7.425 rows=1 loops=1)                                                                          │
[...]

PR #2547, opened 2019-10-14
Use exists? optimize RelationConnection.has_next_page for ActiveRecord relations

newtonry:
This is a followup from:
https://github.com/rmosolgo/graphql-ruby/pull/2513

Resolve:
https://github.com/rmosolgo/graphql-ruby/issues/2332

In this PR:
I use ```exists?``` if the nodes are a ```ActiveRecord::Relation```. Keep the original logic for ```Sequel::Dataset```.

### DIFF
--- a/lib/graphql/relay/relation_connection.rb
+++ b/lib/graphql/relay/relation_connection.rb
@@ -23,9 +23,21 @@ module GraphQL
         end
       end
 
+      def record_exists?(record)
+        if(defined?(ActiveRecord::Relation) && record.is_a?(ActiveRecord::Relation))
+          record.exists?
+        else
+          !record.empty?
+        end
+      end
+
       def has_next_page
         if first
-          paged_nodes.length >= first && nodes.offset(first).exists?
+          if after
+            paged_nodes.length >= first && record_exists?(nodes.offset(first + offset_from_cursor(after)))
+          else
+            paged_nodes.length >= first && record_exists?(nodes.offset(first))
+          end
         elsif GraphQL::Relay::ConnectionType.bidirectional_pagination && last
           sliced_nodes_count >= last
         else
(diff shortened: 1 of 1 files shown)


daemonsy:
This is a continuation of #2513 which includes the original commit.

lib/graphql/relay/relation_connection.rb
```diff
@@ -23,9 +23,21 @@ def cursor_from_node(item)
         end
       end
 
+      def record_exists?(record)
```
rmosolgo: why is this parameter called `record`? Should it be called `relation`, since we expect it to be a Relation? (or Dataset).

lib/graphql/relay/relation_connection.rb
```diff
@@ -23,9 +23,21 @@ def cursor_from_node(item)
...
+        if(defined?(ActiveRecord::Relation) && record.is_a?(ActiveRecord::Relation))
+          record.exists?
+        else
+          !record.empty?
```
rmosolgo: How does this change affect the Sequel::Dataset implementation? Does this run a different query than it used to?

daemonsy: @newtonry is a colleague and we're looking at this together.

`empty?` is probably not what we're going for. Sidenote: I think `Sequel::Dataset`'s method for `EXISTS` is `exists` without `?` :( and we're looking to see if there's a small behavioral difference WRT `ActiveRecord::Relation`.

rmosolgo: (IMO it's fine to leave the non-AR code as it was)

newtonry:
@rmosolgo feel free to take another look at this. I've done as you've suggested and left the non-AR code the same for now.

rmosolgo:
Looks great, thanks for taking care of this!

PR #2552, opened 2019-10-16
Fix  camelize arguments for subscriptions event IDs

durcak:


durcak:
TO fix issue [https://github.com/rmosolgo/graphql-ruby/issues/2553](url)

lib/graphql/subscriptions/event.rb
```diff
@@ -37,7 +37,7 @@ def self.serialize(name, arguments, field, scope:)
           arguments
         when Hash
           if field.is_a?(GraphQL::Schema::Field)
-            stringify_args(arguments)
+            stringify_args(arguments, field)
```
rmosolgo: The definition below has `arg_owner, args` for the parameters, should these be switched?

```suggestion
            stringify_args(field, arguments)
```

durcak: Ou, yes it should.

durcak: So, I have updated the tests and add small fix for stringify_args method with auto-loaded arguments, I am not sure, if it is the best solution, so I am open for a next discusion and suggestions.

durcak:
So, I have updated the tests and add small fix for auto-loaded arguments, I am not sure, if it is the best solution, so I am open for a next discusion or suggestions.

rmosolgo:
It looks good to me! I don't know of a better way of finding a match. Thanks for taking care of this!

PR #2557, opened 2019-10-17
Comment out protect from forgery code, let the consumer opt in

daemonsy:
Follow up to #2556


daemonsy:
Just a FYI ping @rmosolgo in case this gets missed because the underlying issue's closed already

rmosolgo:
Thanks for the cleanup here!

PR #2559, opened 2019-10-19
Warn user when field name is a ruby keyword

### LINKED ISSUE #2199: Field not properly resolved when using restricted keyword
Hi,

I have a mutation with a "class" field that looks something like the following:

```rb
    graphql_name("SaveClass")

    arguments: attributes, Types::ClassAttributes, required: true

    field "class", Types::ClassType, null: false
    field "errors", [Types::UserError], null :false

    def resolve(args)
        @class = Class.where(args).first
        {
             "class" => @class,
             "errors" => [],
        }
    end
end
```

However, when I try to perform the mutation and request the "class" field in the response, my resolver doesn't get the passed `@class` object as expected. Instead it gets passed SaveClassPayload's class. I believe the culprit is the following line:

https://github.com/rmosolgo/graphql-ruby/blob/d31d378c56d3c4a652ef42a2b0ae0d2d4736622d/lib/graphql/schema/field.rb#L599

Are we not supposed to be able to use reserved keywords like this as field names? Please let me know if there's a suggested workaround or if there is in face a bug. Happy to provide more information if helpful.

DmitryTsepelev:
Fixes #2199 

I've bumped into [this](https://stackoverflow.com/questions/27771439/programmatic-way-to-obtain-ruby-keywords) hacky way to get the list of Ruby keywords. On the one hand – it should properly handle different ruby versions, on the other – it's a private API and might stop working in future. Please let me know if you prefer a hard-coded list of keywords or if I should change the wording in the warning.

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'irb/ruby-token'
+
 module GraphQL
   class Schema
     class Member
@@ -38,16 +40,23 @@ module GraphQL
           end
         end
 
+        # A list of Ruby keywords.
+        #
+        # @api private
+        RUBY_KEYWORDS = RubyToken::TokenDefinitions.select { |definition| definition[1] == RubyToken::TkId }
+                                                   .map { |definition| definition[2] }
+                                                   .compact
+
+        # A list of GraphQL-Ruby keywords.
+        #
+        # @api private
+        GRAPHQL_RUBY_KEYWORDS = [:context, :object, :method]
+
         # A list of field names that we should advise users to pick a different
         # resolve method name.
         #
         # @api private
-        CONFLICT_FIELD_NAMES = Set.new([
-          # GraphQL-Ruby conflicts
-          :context, :object,
-          # Ruby built-ins conflicts
-          :method, :class
-        ])
+        CONFLICT_FIELD_NAMES = Set.new(GRAPHQL_RUBY_KEYWORDS + RUBY_KEYWORDS)
 
         # Register this field with the class, overriding a previous one if needed.
         # @param field_defn [GraphQL::Schema::Field]


rmosolgo:
Hi, thanks for kicking this off! I think the approach you've taken is fine. However, I think there are a couple of ways it could be improved: 

- We have some [similar code](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/member/has_fields.rb#L56-L58) for checking field names. How about putting these checks near each other? That will make it easier for me to maintain over time 😅  
- What if someone _wants_ (or needs) to make a field that conflicts with a Ruby built-in? We should have an option for someone to suppress this message. 
- We shouldn't emit warnings for schemas from parsed `.graphql` schema definitions 

In fact, the existing checks have those options. How about adding these built-in tokens to the existing list of conflicting names? 

-----

For posterity, this list added here is: 

```ruby 
RubyToken::TokenDefinitions.select { |definition| definition[1] == RubyToken::TkId }.map { |t| t[2] }.compact
# => ["class", "module", "def", "undef", "begin", "rescue", "ensure", "end", "if", "unless", "then", "elsif", "else", "case", "when", "while", "until", "for", "break", "next", "redo", "retry", "in", "do", "return", "yield", "super", "self", "nil", "true", "false", "and", "or", "not", "alias", "defined?", "BEGIN", "END", "__LINE__", "__FILE__"]
```

DmitryTsepelev:
Thanks for the review! Keeping these checks in a single place makes a lot of sense – I've updated my PR.

rmosolgo:
Looks great, thanks!!

ahlucas:
in ruby 2.7.0-preview3 (and i assume in every newer version) is this hacky way with accessing `irb/ruby-token` straight like this broken. just saying.

ahlucas:
@DmitryTsepelev ^^ 😉

lib/graphql/schema/member/has_fields.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'irb/ruby-token'
```
ahlucas: 😭

PR #2562, opened 2019-10-25
Name ActiveSupport::Notification events in event.library format

jatoben:
The [ActiveSupport::Notifications documentation](https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events) recommends naming events in `event.library` format. GraphQL-Ruby currently inverts that to `library.event`, which makes using [ActiveSupport::Subscriber](https://api.rubyonrails.org/classes/ActiveSupport/Subscriber.html) a bit awkward.

For example, `#execute_query` will never be called in this class:

```ruby
class GraphqlSubscriber < ActiveSupport::Subscriber
  attach_to :graphql

  def execute_query(event)
    Rails.logger.info "Executed query"
  end
end
```

A workaround is to `attach_to :execute_query`, then define a `#graphql` method, but doing so requires a new ActiveSupport::Subscriber subclass for each event an app might be interested in.

In this PR, I've reversed the event names, allowing ActiveSupport::Subscriber to work in the normal fashion.

The format change will break any existing subscribers, but I noticed that [the docs describe the AS::N tracer as experimental](https://graphql-ruby.org/queries/tracing#activesupportnotifications). If this change is acceptable for a minor release, would a warning in the changelog suffice?


rmosolgo:
Thanks for the improvement! I must have missed that suggestion before. I agree it's appropriate for a minor version change, so I've updated this PR to point at the `1.10-dev` branch.

rmosolgo:
🤦‍♂ Aaaand I introduced a lint error because conflicts. trying again...

PR #2570, opened 2019-10-29
Allow resolvers to define field extensions

DmitryTsepelev:
I think it might be useful to have the ability to define field extensions inside resolvers for cases when some logic is shared between resolvers and types:

```ruby
class GreetingExtension < GraphQL::Schema::FieldExtension
  def resolve(object:, arguments:, **rest)
    name = yield(object, arguments)
    "#{options[:greeting]}, #{name}!"
  end
end

class ResolverWithExtension < BaseResolver
  type String, null: false

  extension GreetingExtension, greeting: "Hi"

  def resolve
    "Robert"
  end
end
```

Please let me know if it sounds worthless 🙂

### DIFF
--- a/guides/fields/resolvers.md
+++ b/guides/fields/resolvers.md
@@ -200,3 +200,26 @@ end
 ```
 
 In doing so, you can defer the loading of the type class until the nested resolver has already been loaded.
+
+# Extensions
+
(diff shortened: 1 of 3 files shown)


rmosolgo:
That's great, thanks so much! I agree it'll be really useful. Thanks for the contribution.

PR #2580, opened 2019-10-31
Generate a base mutation class referencing the base type classes

louim:
I noticed that a few places in the guides reference a `BaseMutation`, including the [Mutation Classes guide](https://graphql-ruby.org/mutations/mutation_classes.html#example-mutation-class). I though it would be a good idea to have the generator create that BaseMutation, referencing the type classes that are generated.

Let me know what you think! Happy to tweak things if this is an appropriate addition to the project.


rmosolgo:
This is great! Thanks so much for making this improvement 🎉

PR #2583, opened 2019-11-01
Update guides with Skylight GraphQL probe

vaidehijoshi:
Hi there! 👋

We work at Skylight, and recently released a new version of our gem, which now offers support for GraphQL instrumentation through `graphql-ruby`! We'd love to update your guides with how to use our new probe and what it looks like in action.

We will open a separate PR to deprecate the `GraphQL::Tracing::SkylightTracing` module.


guides/queries/tracing.md
```diff
@@ -113,19 +113,19 @@ end
...
-class MySchema < GraphQL::Schema
-  use(GraphQL::Tracing::SkylightTracing)
-end
+# config/application.rb
+config.skylight.probes << "graphql"
 ```
 
-
 <div class="monitoring-img-group">
   {{ "/queries/skylight_example.png" | link_to_img:"skylight monitoring" }}
```
rmosolgo: Should we update this example screenshot of monitoring, too, to reflect the new GraphQL probe, or is it basically the same?

vaidehijoshi: We just [replaced the image](https://github.com/rmosolgo/graphql-ruby/blob/06d7c8bec07a7bdd8aaf62b52d1a0368187cb116/guides/queries/skylight_example.png) (`skylight_example.png`) with an updated screenshot of what the GraphQL probe looks like in action in the Skylight UI.

Let me know if you think that's ok!

rmosolgo:
Thanks for the update here! I hope I get a chance to try it out sometime 🍻

PR #2586, opened 2019-11-05
Update 'Executing queries' documentation with info on how to make batch updates

nadiyaka:


rmosolgo:
Thanks for sharing your solution here, but I don't think I want to include this kind of content in the documentation. That is, it _looks_ like the project recommends duplicating mutations like this. If it were _me_, I'd probably create a `createPosts` mutation _instead_, one that takes an array of inputs and saves them in a single transaction. 

Although I think there _is_ a need for schema design advice, I'd rather keep the graphql-ruby docs as focused as possible on how the gem works. I'd love to see this kind of content on a personal blog, though! Thanks again for sharing.

PR #2590, opened 2019-11-07
fix(website) new mongo install instructions

kivikakk:
:wave: Noticed this while running through the setup. Mongo has its own tap now -- see https://github.com/Homebrew/homebrew-core/pull/43770.


rmosolgo:
Thanks!

PR #2594, opened 2019-11-11
[Docs] Fixing small mistake in table for "List" docs

MacroMackie:
There was a small issue with the valid/invalid table for "Non-null lists with nullable items". 
An example that should have been _invalid_ was showing up as _valid_.

Screen Shot 2019-11-11 at 12 48 35 PM

### DIFF
--- a/guides/type_definitions/lists.md
+++ b/guides/type_definitions/lists.md
@@ -109,7 +109,7 @@ Valid | Invalid
 ------|------
 `[]`  | `null`
 `[1, 2, ...]`| `[null]`
- | `[1, null, 2, ...]`
+| | `[1, null, 2, ...]`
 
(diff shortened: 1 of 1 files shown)


guides/type_definitions/lists.md
```diff
@@ -109,7 +109,7 @@ Valid | Invalid
 ------|------
 `[]`  | `null`
 `[1, 2, ...]`| `[null]`
- | `[1, null, 2, ...]`
+| | `[1, null, 2, ...]`
```
MacroMackie: Note the original intention here - you can see that this is meant to be in the second column (invalid)

Adding a second `|` helps markdown realize that this is meant to be in column two.

rmosolgo:
Awesome, thanks for the fix here!

Issue #1333, opened 2018-03-09
[Warden] Types referenced from invisible fields still show up.


eapache:
@rmosolgo @xuorig 

Given a simple schema like this:
```graphql
type A {
  b: B
}

enum B {
  VALUE
}
```

If the type A is hidden then we would expect B to be hidden as well, but this is not the case. This is because when checking whether B is referenced, we only look to the immediate visibility setting of the field `b` rather than doing the full check as to whether `b` itself is implicitly hidden (which it is, because it's on `A` which is hidden).

Instead of checking `visible?` in https://github.com/rmosolgo/graphql-ruby/blob/eff0a6c952734225bc899ec712d20c5a4e1b9be9/lib/graphql/schema/warden.rb#L152-L155 I believe we can just check `visible_field?` instead which does the right thing?

Caveats to that suggestion:
- I understand `references_to` can return arguments as well, so we'll need something like `if field; visible_field?(x); elseif argument; visible_field?(x.field)`
- This potentially results in an infinite loop if a type references itself? I guess we'll need to keep a set of all of the "actively visited" types so we can avoid that.

rmosolgo:
Yeah, I agree it should work the way you describe! Thanks for those details.

eapache:
I took a crack at this and I think I missed something key in my initial analysis: `visible_field?(b)` doesn't check the visibility of `A`, it checks the visibility of `B`. In fact, I don't see a way to get from the `GraphQL::Field` for `b` to the type `A` in the first place; is this even possible?

eapache:
`GraphQL::Schema::Field` has an `owner` which would work, is updating the warden to use the class-based classes a possibility?

eapache:
This problem effectively devolves into https://en.wikipedia.org/wiki/Reachability from the root for a cyclic digraph. Eagerly walking from the root would work but would require preprocessing the entire schema on every new warden (aka every request). Even lazy path-climbing back to the root (assuming we can get from a field to its owner) isn't going to be very efficient depending on the schema.

I appreciate the current warden implementation as in fact a very good sort of "best-effort" to do as much reference-checking as possible without actually doing any recursion, however I'm starting to wonder if it's more effort than it's worth. Now that I've grokked the problem I can point out a few other potential surprises here that aren't fixable without a full reachability analysis.

[...]

PR #2596, opened 2019-11-12
Introspection shouldn't return types that are unreachable due to visibility checks

jturkel:
This PR takes a crack at fixing the heart of issue #1333 by hiding any types in introspection queries that become unreachable due to visibility checks. The algorithm does a depth first traversal of the schema's "graph" starting at the roots and only recursing into nodes that pass visibility checks. It punts on pushing this logic into Warden since that's a much trickier problem to solve and doesn't appear to provide much value.

I ran the introspection query benchmark and it doesn't seem to add any noticeable performance impact although this could be different for large schemas:

Branch:
```
               query     2.000  i/100ms
               query     23.533  (± 4.2%) i/s -    118.000  in   5.020621s
```

Master:
```
               query     2.000  i/100ms
               query     23.128  (± 4.3%) i/s -    116.000  in   5.019993s
```

### DIFF
--- a/lib/graphql/introspection/entry_points.rb
+++ b/lib/graphql/introspection/entry_points.rb
@@ -15,8 +15,9 @@ module GraphQL
       end
 
       def __type(name:)
-        type = context.warden.get_type(name)
+        return unless context.warden.reachable_type?(name)
 
+        type = context.warden.get_type(name)
         if type && context.interpreter?
           type = type.metadata[:type_class] || raise("Invariant: interpreter requires class-based type for #{name}")
         end
--- a/lib/graphql/introspection/schema_type.rb
+++ b/lib/graphql/introspection/schema_type.rb
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
+
+require 'set'
+
 module GraphQL
   module Introspection
     class SchemaType < Introspection::BaseObject
@@ -14,7 +17,7 @@ module GraphQL
       field :directives, [GraphQL::Schema::LateBoundType.new("__Directive")], "A list of all directives supported by this server.", null: false
 
       def types
-        types = @context.warden.types
+        types = reachable_types
         if context.interpreter?
           types.map { |t| t.metadata[:type_class] || raise("Invariant: can't introspect non-class-based type: #{t}") }
         else
@@ -23,15 +26,15 @@ module GraphQL
       end
 
       def query_type
-        permitted_root_type("query")
+        @query_type ||= permitted_root_type("query")
       end
 
       def mutation_type
-        permitted_root_type("mutation")
+        @mutation_type ||= permitted_root_type("mutation")
       end
 
       def subscription_type
-        permitted_root_type("subscription")
+        @subscription_type ||= permitted_root_type("subscription")
       end
 
       def directives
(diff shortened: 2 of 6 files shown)


eapache:
I'll have to benchmark but I imagine this will be impractically slow for Shopify's schema. Also there are a bunch of secondary cases (e.g. https://github.com/rmosolgo/graphql-ruby/issues/1333#issuecomment-397740253) which I believe it is still broken for.

eapache:
Oh I'm sorry, I read this as applying to the warden, but it's actually specific to introspection?

jturkel:
Yup. I was just targeting introspection. This PR really just attempts to make the introspection behavior consistent with the non-recursive behavior of types that are hidden by warden (or types that are unreachable for other reasons). The general warden problem is much harder as you described in #1333 :)  

Wow. The Shopify admin schema is big! I ran some benchmarks against the schemas I could find online and the change doesn't impact time or memory allocations very much:

[Shopify Storefront API](https://help.shopify.com/en/api/storefront-api)
```
      original admin     1.000  i/100ms
     reachable admin     1.000  i/100ms
      original admin      1.321  (± 0.0%) i/s -     40.000  in  30.562708s
     reachable admin      1.382  (± 0.0%) i/s -     42.000  in  30.415880s

Original Memory:
Total allocated: 11649664 bytes (102822 objects)
Total retained:  4875440 bytes (32752 objects)

Reachable Memory:
Total allocated: 11674232 bytes (103016 objects)
Total retained:  4886312 bytes (32905 objects)
```

[Shopify Admin API](https://help.shopify.com/en/api/graphql-admin-api)
```
      original admin     1.000  i/100ms
     reachable admin     1.000  i/100ms
      original admin      1.321  (± 0.0%) i/s -     40.000  in  30.562708s
     reachable admin      1.382  (± 0.0%) i/s -     42.000  in  30.415880s

Original Memory
Total allocated: 33119728 bytes (290781 objects)
Total retained:  13768824 bytes (91857 objects)

Reachable Memory
Total allocated: 33211736 bytes (291408 objects)
Total retained:  13812440 bytes (92364 objects)
```

eapache:
Hmm, I ran some local end-to-end experiments on the following query with/without this change:
```
{
  __schema {
    types {
      name
    }
  }
}
```

The total times were too noisy to draw statistical conclusions, but eyeballing them this looks sightly slower. Interestingly I threw in a `Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)` calls and the `reachable_types` method here was consistently 200ms+ and the bulk of our response time for that query... the saving grace is that it seems `@context.warden.types` is just as slow!

eapache:
OK... this still isn't 100% correct since it doesn't prune empty types, but it does seem to be an improvement from the old behaviour and not obviously slower. Curious what the github folks make of it.

rmosolgo:
I'm open to fixing this issue and I have the same concerns about performance. But I haven't tried it yet 😅 . 

By implementing this in the introspection class, does this mean that some queries will _succeed_ even though their types/fields don't appear in the introspection result? If so, I'm hesitant about the approach, because schema visibility _is_ billed as a security feature. That is, if you code it so that a type or field is hidden, then queries on that thing should _fail_. How does that play out here?

jturkel:
Testing this out locally with the Github schema the latency seems to be within the margin of error and the increase in memory usage is only 0.18%:

```
Github (Original)
Total allocated: 54477338 bytes (477315 objects)
0.794  (± 0.0%) i/s -     48.000  in  60.501011s

Github (Reachable)
Total allocated: 54574946 bytes (478044 objects)
0.788  (± 0.0%) i/s -     48.000  in  61.126936s
```
I could be missing something but I don't think this change introduces any security holes. It just filters out types from introspection results that are only reachable via hidden types/fields. Thus any query attempting to access these filtered types would have to go through a hidden type/field to get to them and would thus be flagged as invalid.

eapache:
Yeah, I think https://github.com/rmosolgo/graphql-ruby/issues/1333#issuecomment-397817135 proves that we can't have any security holes in actual execution as long as the query is well-formed and we validate the visibility function of every type/field/etc *referenced in the query*. Introspection is the hard case.

rmosolgo:
Yes, I'm sorry it didn't cross my mind that _unreachable_ objects aren't going to be leaked in queries 🤦‍♂ .

Thanks for linking that old discussion of the topic, @eapache. I forgot how good that is! I admit, I think I'd need a whole day to dwell on it to really comprehend it all 😬 

> queries with no introspection

This leads to my last question: do we have to consider `__type(name: "SomeUnreachableType")` in this case? Should we return nil there, too? 

Thanks for sharing the benchmark, @jturkel. Under the hood, we have a "private" schema which is ~3x bigger, so I'll have to test that one too 😅

jturkel:
Great point. I think `__schema { types }` and `__type(name: String!)` should behave consistently so I plumbed the logic through over there too. For the implementation, I wanted to share the set of reachable types between `GraphQL::Introspection::SchemaType#types` and `GraphQL::Introspection::EntryPoints#__type` so I put the logic in `GraphQL::Schema::Warden`. This might not be ideal since it mixes security concerns with introspection concerns so I'm definitely open to suggestions on a better place to put this logic.

jturkel:
@rmosolgo - I'm sure you're busy but any thoughts on this PR?

rmosolgo:
Sorry for the delay and thanks for the ping! I dig it as is. 

Looking at the implementation again, maybe it makes sense why the performance hit isn't too big. For clients running the _whole_ introspection query, the new `types` implementation warms a lot of caches on the Warden: arguments, fields, `visible?`, etc. They _would have_ been hit eventually, but now we're frontloading it. Theoretically, it might be a perf hit for introspection _less than_ the whole query, but that's probably an insignificant consideration. 

Thanks for again for taking this on and for your patience with me 🙇 !

PR #2598, opened 2019-11-14
Block Strings: Fix bug calling empty? on nil

theorygeek:
This fixes a bug when a block string contains only newlines.


rmosolgo:
thanks for the fix!!

PR #2601, opened 2019-11-15
Deprecate GraphQL::Tracing::SkylightTracing

vaidehijoshi:
This is a follow-up to #2583.

Ensures that we warn whenever the `GraphQL::Tracing::SkylightTracing` class is initialized*, and suggest enabling Skylight's GraphQL probe instead.

*I guessed that warning in the `initialize` made the most sense, but definitely let me know if this warning should live somewhere else! :)


rmosolgo:
:+1: I think that's a great spot for this warning, thanks for adding it. 

One thought I have is, when someone sees this warning, how will they know what to do next? Can leave a URL or a code snippet to inform them in how to make the warning go away?

vaidehijoshi:
@rmosolgo That's a good point! I've updated to include a link to the Skylight docs to point folks to how to enable the probe and get rid of that warning.

rmosolgo:
😎 Thanks!

PR #2604, opened 2019-11-18
Update Relay object indentification doc

kakas:
Update Relay Node Field from legacy-style to new-style
https://github.com/rmosolgo/graphql-ruby/blob/d31d378c56d3c4a652ef42a2b0ae0d2d4736622d/lib/graphql/types/relay/node_field.rb#L9
https://github.com/rmosolgo/graphql-ruby/blob/d31d378c56d3c4a652ef42a2b0ae0d2d4736622d/lib/graphql/types/relay/nodes_field.rb#L9


rmosolgo:
Thanks for the update!

PR #2608, opened 2019-11-19
Fixes for new pagination 

### LINKED ISSUE #1109: when setting max_page_size on a connection pageInfo hasNextPage, hasPriorPage is wrong if no first: or last: are specified

rmosolgo:
Improve capping by max_page_size, now subclasses can "just" use `first` and `last`. Max_page_size will be applied as `first` if no `first` or `last` is given. 

The advantage there is we get `has_next_page`/`has_previous_page` correctly implemented for free, and it's harder to forget to use max_page_size. Fixes #1109

### DIFF
--- a/lib/graphql/pagination/array_connection.rb
+++ b/lib/graphql/pagination/array_connection.rb
@@ -68,7 +68,6 @@ module GraphQL
 
           limited_nodes = limited_nodes.first(first) if first
           limited_nodes = limited_nodes.last(last) if last
-          limited_nodes = limited_nodes.first(max_page_size) if max_page_size && !first && !last
 
           limited_nodes
         end
--- a/lib/graphql/pagination/connection.rb
+++ b/lib/graphql/pagination/connection.rb
@@ -26,7 +26,18 @@ module GraphQL
       # @return [GraphQL::Query::Context]
       attr_accessor :context
 
-      attr_accessor :before, :after
+      # Raw access to client-provided values. (`max_page_size` not applied to first or last.)
+      attr_accessor :before_value, :after_value, :first_value, :last_value
+
+      # @return [String, nil] the client-provided cursor
+      def before
+        @before_value
+      end
+
+      # @return [String, nil] the client-provided cursor
+      def after
+        @after_value
+      end
 
       # @param items [Object] some unpaginated collection item, like an `Array` or `ActiveRecord::Relation`
       # @param context [Query::Context]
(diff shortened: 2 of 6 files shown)


raysuelzer:
This seems to break implementations which implemented a totalCount property when first is used.  

```
# frozen_string_literal: true

module Types
  # Base Connection Class
  # https://github.com/rmosolgo/graphql-ruby/blob/master/guides/relay/connections.md
  class BaseConnection < GraphQL::Types::Relay::BaseConnection
    field :total_count, Integer, "Total # of objects returned from this Plural Query", null: false

    def total_count
      # Now performs a count using the limit
     
      object.nodes&.size
    end
  end
end
```

rmosolgo:
Some things were renamed, try `object.items` instead: 

https://github.com/rmosolgo/graphql-ruby/blob/1c6acb8c8280525b1bac4dae437af654515ac514/lib/graphql/pagination/connection.rb#L23-L24

(Personally, I found it confusing that `object.nodes` was _not_ the same value as `things { nodes { ... } }`. That is, in _Ruby_, `nodes` was unpaginated, but in GraphQL, it was paginated. Now, `nodes` is always the paginated set of items.)

Does `.items` work better for you?

raysuelzer:
Items worked. I opped another issue here: https://github.com/rmosolgo/graphql-ruby/issues/2695

It seems to be executing a query to check for next page / previous page, but that can be extremely slow if the table is large.  It does this even if we do not ask for the data. My queries are even timing out in some cases.

PR #2612, opened 2019-11-20
Update conversion from ActiveRecord error

louim:
ActiveRecord errors keys are symbols, so they would need to be converted to string first. The [`camelize`](https://api.rubyonrails.org/classes/String.html#method-i-camelize) method will also default to `PascalCase` unless the `:lower` argument is passed.

I though that it would help people implementing that pattern of errors.


rmosolgo:
Thanks, I see we have the same code in our codebase, I must have just copied it wrong 🤦‍♂ Thanks for the fix!

PR #2613, opened 2019-11-20
Fix multiplex max depth

### LINKED ISSUE #2388: When Schema with max_complexity set, multiplex queries with an abstract type get `NoMethodError` using AST analyzers

rmosolgo:
Fixes #2388 

Alternative to #2394 

We were already using "query" analyzers to analyze multiplexes, but in some cases it didn't actually work. 

I want to update it to, yknow, _actually_ work, and also: 

- Update the analyzer API to clarify what kind of thing it's analyzing (while maintaining a use-agnostic access to `subject`) 
- Reflect this dual usage in the docs 

Thanks especially to @patvice who wrote up this issue in detail, provided the appropriate tests, and explored an alternative solution.

### DIFF
--- a/guides/queries/ast_analysis.md
+++ b/guides/queries/ast_analysis.md
@@ -124,3 +124,11 @@ end
 ```
 
 **Make sure you pass the class and not an instance of your analyzer. The new analysis engine will take care of instantiating your analyzers with the query**.
+
+## Analyzing Multiplexes
+
+Analyzers are initialized with the _unit of analysis_, available as `subject`.
+
+When analyzers are hooked up to multiplexes, `query` is `nil`, but `multiplex` returns the subject of analysis. You can use `visitor.query` inside visit methods to reference the query that owns the current AST node.
+
+Note that some built-in analyzers (eg `AST::MaxQueryDepth`) support multiplexes even though `Query` is in their name.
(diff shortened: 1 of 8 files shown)


chrisbutcher:
+1 this looks great.

I just did an exploration on this topic over here: https://github.com/Shopify/graphql-metrics/pull/10#discussion_r347466683 (see longer comment linked to at the bottom of this thread too).

PR #2614, opened 2019-11-20
WIP: Plugins can attach tracers, analyzers

chrisbutcher:
This PR contains a failing test to demonstrate a bug/inconsistency uncovered in https://github.com/Shopify/graphql-metrics/pull/10#discussion_r336174539

Attaching instrumentation via a plugin with `def self.use(schema_defn_proxy)` works just fine, but attaching a tracer or query analyzer this way doesn't work.

See https://github.com/rmosolgo/graphql-ruby/issues/2615 for full context.

Looking into the details of how these things are setup on schema definitions, I'm a bit unclear on how to proceed, with the flow of `def_delegators`, instance variables, accessor/reader methods involved in GraphQL::Schema.

Can anyone point me in the right direction?


rmosolgo:
This will be a little clearer in 1.10 when the _class_ instead of the proxy is passed in here. To support that, I've implemented these methods on the schema class on `1.10-dev`:

https://github.com/rmosolgo/graphql-ruby/blob/f24b1cbcde511c16a25c6e0ef045131409c18cdf/lib/graphql/schema.rb#L1375-L1392

On master, I suppose we could move those methods to the class. But alternatively, I think it might work if we _don't_ unwrap the schema class. Does it work in that case? Something like:

```diff
-        def self.use(schema_defn_proxy)
-          schema_defn = schema_defn_proxy.target
+        def self.use(schema_defn)
           schema_defn.instrument :query, NoOpInstrumentation
```

Then, it will find the configuration helpers for those methods. 🤞

chrisbutcher:
Thanks @rmosolgo! I'll add my tests to a branch off of `1.10-dev` to confirm my tests pass there.

chrisbutcher:
> On master, I suppose we could move those methods to the class. But alternatively, I think it might work if we don't unwrap the schema class. Does it work in that case? Something like:

In the meantime, re: master's current behaviour, I pushed up another commit that I think demonstrates a few of my assumptions about how plugins can add instrumentation, tracing, analyzers to a schema definition / re-definition.

As an aside, one thing I'm hoping to be able to do is attach tracing, analyzers to a schema class, without having to use the return value of Schema.redefine (which returns an instance of the schema, not the modified class).

rmosolgo:
I've updated the tests to use latest master, and also to _subclass_ the initial schema instead of using `.redefine`. Do those changes look good to you?

spec/spec_helper.rb
```diff
@@ -123,3 +123,5 @@ def trace(key, data)
     end
   end
 end
+
```
chrisbutcher: ✂️

PR #2617, opened 2019-11-22
updates mutation_generator to use underscored field names

newx:
Currently the mutation generator are generating camel case field names.

I think since this https://github.com/rmosolgo/graphql-ruby/pull/1126
and the class based API were introduced, the standard seems to have changed to ruby underscore
syntax names. See https://github.com/rmosolgo/graphql-ruby/pull/1126/files#diff-161608206c9ba9bc29a59e613e2bdbfdR307  

Before

```
field :updateName, mutation: Mutations::UpdateName
```

After

```
field :update_name, mutation: Mutations::UpdateName
```

Let me know if this is the expected behaviour

cc @rmosolgo

### DIFF
--- a/lib/generators/graphql/mutation_generator.rb
+++ b/lib/generators/graphql/mutation_generator.rb
@@ -45,7 +45,7 @@ module Graphql
       private
 
       def assign_names!(name)
-        @field_name = name.camelize(:lower)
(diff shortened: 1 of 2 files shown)


rmosolgo:
Yes, this is correct, thank you!

PR #2620, opened 2019-11-25
Recommend the right thing up front for multiplex

eapache:
I'm pretty sure this is the right way to configure the JSON rendering to not include a
root node (which breaks the multiplex spec and confuses clients).

@rmosolgo I'm doing this a tad blind since I don't know the impetus behind the original multiple recommendations, but this works for us and AFAIK the pure `render json: result` should never work, so...

cc @rrangith

### DIFF
--- a/guides/queries/multiplex.md
+++ b/guides/queries/multiplex.md
@@ -78,11 +78,9 @@ def execute
     )
   end
 
-  render json: result
+  render json: result, root: false
 end
 ```
-
-If Apollo Client has issues recognizing the result of `render json: result`, replace it with `render body: result.to_json, content_type: 'application/json'`.
 
 ## Validation and Error Handling
 


guides/queries/multiplex.md
```diff
@@ -78,12 +78,10 @@ def execute
     )
   end
 
-  render json: result
+  render json: result, root: false
 end
 ```
 
-If Apollo Client has issues recognizing the result of `render json: result`, replace it with `render body: result.to_json, content_type: 'application/json'`.
```
rrangith: Did some looking around and found https://github.com/rmosolgo/graphql-ruby/issues/1948 and https://github.com/rmosolgo/graphql-ruby/pull/1949 for the reasoning behind this additional option. They were having the same issue of `render json: result` returning { 'graphql': result } 😃

rmosolgo:
Thanks for clarifying! Looking back, I see what you mean about this being the only way it'll really _work_.

PR #2622, opened 2019-11-26
TimeoutMiddleware: switch to monotonic time

sdemjanenko:
The call to `Time.now` can be affected by changes to system time such as
NTP updating the OS' time. This can be avoided by using a monotonically
increasing time source.

For more information see:
https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/


rmosolgo:
Thanks for the fix!

PR #2623, opened 2019-11-29
Add clarification to Multiplex about ordering of responses

rrangith:
@rmosolgo 
We were worried that since `Multiplex` processes requests concurrently, that it wouldn't be able to maintain order in our responses. 

This PR is to confirm to users that results will always be kept in the same order that their respective requests were sent in. 

This way users can do something like the following without worrying about whether or not the order of the responses was maintained:
```
payload = [
  ComponentA.required_graphql_query,
  ComponentB.required_graphql_query
]

responses = GraphQLClient.batch_request(payload: payload)

ComponentA.handle_response(responses[0])
ComponentB.handle_response(response[1])
```

I could optionally remove the test case I made, but the doc change should be made so people don't get confused in the future

### DIFF
--- a/guides/queries/multiplex.md
+++ b/guides/queries/multiplex.md
@@ -47,7 +47,7 @@ Then, pass them to `Schema#multiplex`:
 results = MySchema.multiplex(queries)
 ```
 
-`results` will contain the result for each query in `queries`.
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for adding that clarification!

rrangith:
cc: @tomgatzgates 😃

Issue #2597, opened 2019-11-14
[1.9.15] prepare methods not called when using 1.9+ interpreter


arempe93:
When using `GraphQL::Execution::Interpreter`, `prepare` methods are called properly in mutations but not on query fields.

```ruby
MyPoint = Struct.new(:latitude, :longitude)

class PointInput < GraphQL::Schema::InputObject
  argument :latitude, Float, required: true
  argument :longitude, Float, required: true

  def prepare
    MyPoint.new(latitude, longitude)
  end
end
```

```ruby
class TestMutation < GraphQL::Schema::Mutation
  argument :point, PointInput, required: true

  def resolve(point:)
    point #=> #<struct MyPoint latitude=xx, longitude=xx>
  end
end
```

```ruby
class QueryType < GraphQL::Schema::Object
  field :test, Int, null: true do
    argument :point, PointInput, required: true
  end

  def test(point:)
    # if `use GraphQL::Execution::Interpreter`:
      point #=> #<PointInput:0x0000xxxxx>
    # else:
      point #=> #<struct MyPoint latitude=xx, longitude=xx>
  end
end
```

rmosolgo:
Doh, thanks for reporting this. We revived and old branch for this feature, and I guess it wasn't implemented in the Interpreter yet 😖

newx:
@rmosolgo I was able to reproduce this. Any tips on how to implement these changes on the new interpreter? I would like to try to do them. Thanks

newx:
I think this can be closed now @arempe93

PR #2624, opened 2019-11-30
Issue 2597 input_object prepare

newx:
This is an attempt to fix the issue below
https://github.com/rmosolgo/graphql-ruby/issues/2597

## TODO

- [x] handle #prepare for entire input object on queries
- [x] update `input_object_spec`

## Questions

1. Should an `InputObject` instance `#prepare` method be called and then later the `obj#prepare` (`GraphQL::Schema::Object`) be called upon the InputObject prepared value? Or should it be one OR the another? (See `GraphQL::Schema::Argument`)
2. Is this how this should be handled? Or am I missing more context?

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -121,7 +121,10 @@ module GraphQL
       # Used by the runtime.
       # @api private
       def prepare_value(obj, value)
-        binding.pry        
+        # NOTE: This fixes the issue on queries.
+        # QUESTION: Should an InputObject #prepare method be called and then
+        # later the obj#prepare (GraphQL::Schema::Object) be called upon the
+        # InputObject prepared value? Or should it be one OR the another?
         if value.respond_to?(:prepare)
           value = value.public_send(:prepare)
         end
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -121,6 +121,11 @@ module GraphQL
       # Used by the runtime.
       # @api private
       def prepare_value(obj, value)
+        binding.pry        
+        if value.respond_to?(:prepare)
+          value = value.public_send(:prepare)
+        end
+
         if @prepare.nil?
           value
         elsif @prepare.is_a?(String) || @prepare.is_a?(Symbol)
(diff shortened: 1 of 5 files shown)


newx:
@rmosolgo please let me know what you think

lib/graphql/schema/argument.rb
```diff
@@ -121,6 +121,14 @@ def type
       # Used by the runtime.
       # @api private
       def prepare_value(obj, value)
+        # NOTE: This fixes the issue on queries.
+        # QUESTION: Should an InputObject #prepare method be called and then
+        # later the obj#prepare (GraphQL::Schema::Object) be called upon the
+        # InputObject prepared value? Or should it be one OR the another?
+        if value.respond_to?(:prepare)
+          value = value.public_send(:prepare)
```
rmosolgo: Since we're always calling the same method here, why not hardcode the method call? 

```suggestion
          value = value.prepare
```

lib/graphql/schema/argument.rb
```diff
@@ -121,6 +121,14 @@ def type
       # Used by the runtime.
       # @api private
       def prepare_value(obj, value)
+        # NOTE: This fixes the issue on queries.
+        # QUESTION: Should an InputObject #prepare method be called and then
+        # later the obj#prepare (GraphQL::Schema::Object) be called upon the
+        # InputObject prepared value? Or should it be one OR the another?
+        if value.respond_to?(:prepare)
```
rmosolgo: I think this a fine way to do it. It seems like, if we ignored one configuration or the other, it could be pretty confusing. And, it doesn't seem any more complicated to do both, is that right? 


What do you think about making an explicit type check? I'm worried about calling someone's `String#prepare` monkey patch or something. We could do this: 

```suggestion
        if value.is_a?(GraphQL::Schema::InputObject)
```

lib/graphql/schema/argument.rb
newx: a2912dac627eeb61bd239b4e0a0e410900026829

lib/graphql/schema/argument.rb
newx: 342b3db882aa516608f46ef221f6083f5e3bdd6a

newx:
@rmosolgo thanks for the feedback. They were all fixed. Let me know if we need anyting else.

rmosolgo:
Looks great, thanks again!

PR #2628, opened 2019-12-03
add graphql-rails-generators to related projects

ajsharp:


rmosolgo:
Thanks for the suggestion!

PR #2630, opened 2019-12-03
Perf improvements

rmosolgo:
Here are some improvements after inspecting GitHub's GraphQL system with StackProf.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1030,7 +1030,15 @@ module GraphQL
             own_refs + inherited_refs
           end
         else
-          find_inherited_value(:references_to, EMPTY_HASH).merge(@own_references_to)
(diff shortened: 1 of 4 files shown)


lib/graphql/schema.rb
```diff
@@ -1030,13 +1030,21 @@ def references_to(to_type = nil, from: nil)
...
+          if inherited_value.any?
+            inherited_value.merge(@own_references_to)
+          else
+            @own_references_to
+          end
         end
       end
 
       def type_from_ast(ast_node, context: nil)
-        type_map = context ? context.warden.types : self.types
```
rmosolgo: Calling `.types` can be slow because it merges some big hashes. This was the _only_ use at runtime, so I found a way to replace it!

Issue #2602, opened 2019-11-15
Middleware via hooks in the "round-trip" of a field


lancelafontaine:
This issue is part-feature request and part-inquiry as to how this pattern can best be achieved:

When an argument is passed to a field, have the argument's value be accessible to only the "sub-fields" in the subsequent tree of fields that are queried from that point.

Here's an example of having a configuration value like a session ID be accessible for every field under the one in which the session is provided.

```graphql
query problemDomain {
    a(configurationValue: "1"){ # before hook to set configuration value
        aa { # is aware of configurationValue "1"
            ab # is aware of configurationValue "1"
        }
        ac # is aware of configurationValue "1"
    } # after hook to unset configuration value
    b { # nothing to set
        ba { # has no configurationValue
            bb # has no configurationValue
        }
        bc # has no configurationValue
    } # nothing to unset
    c(configurationValue: "2") { # before hook to set configuration value
        ca { # is aware of configurationValue "2"
            cb # is aware of configurationValue "2"
        }
        cc # is aware of configurationValue "2"
    } # after hook to unset configuration value
}
```
This example relies on this propagation occurring and having a trigger or hook to stop the propagation at the end of a field's "round-trip" and unset the configuration value.

I've found this requirement conceptually similar to _middleware_, a useful pattern often seen in the Node.js and Elixir worlds to avoid manipulating global mutable states from any arbitrary point within the request. I can imagine a similar concept being applied here, but instead of at the request level, at the field/resolver level. Some of the use cases off the top of my head:
- Session management (see above)
- Logging
- Benchmarking
- Prefetching and caching dependent objects
- Parsing inputs and context
- Authentication

`graphql-ruby` already offers field extensions as a concept, which are similar in nature to what this is attempting to solve, but the `after_resolve` hook only triggers immediately after a field's resolver, not on the field's "round-trip".

From what I can tell, without a primitive, the only option for enabling this sort of pattern is to change the backing model of all GraphQL objects to be a decorator which can manually propagate the argument value through every resolver. Even then, this approach only solves the session management / configuration value use case, and doesn't address any case which relies on an explicit `after_round_trip` hook.

Hoping to get some feedback on the general pattern, whether there are any other viable options for enabling such use cases, and whether such a primitive would make sense alongside constructs like field extensions. Thanks!

eapache:
Another approach to solving this might be some sort of mutable `context` namespaces scoped to parts of the AST?

rmosolgo:
[...]

PR #2634, opened 2019-12-05
Adds scoped context for propagating values specific to a field and its children

lancelafontaine:
Resolves https://github.com/rmosolgo/graphql-ruby/issues/2602.

This PR adds a "scoped context" to the existing context. This is useful for propagating values which are specific to a field and its children, as opposed to for the entirety of the query.

This is achieved with very little overhead on the interpreter runtime by simply reinitializing the scoped context at various points where new resolvers would run within a possibly new context as the query is being executed.

### DIFF
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -215,6 +215,11 @@ module GraphQL
         @scoped_context = @scoped_context.merge(hash)
       end
 
+      def scoped_set!(key, value)
+        scoped_merge!(key => value)
+        nil
+      end
+
       class FieldResolutionContext
         include SharedMethods
         include Tracing::Traceable
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -167,15 +167,27 @@ module GraphQL
       # @api private
       attr_accessor :scoped_context
 
-      def_delegators :@provided_values, :[], :[]=, :to_h, :to_hash, :key?, :fetch, :dig
+      def_delegators :@provided_values, :[]=
+      def_delegators :to_h, :fetch, :dig
       def_delegators :@query, :trace, :interpreter?
-
-      # @!method [](key)
-      #   Lookup `key` from the hash passed to {Schema#execute} as `context:`
 
       # @!method []=(key, value)
       #   Reassign `key` to the hash passed to {Schema#execute} as `context:`
 
+      # Lookup `key` from the hash passed to {Schema#execute} as `context:`
+      def [](key)
+        return @scoped_context[key] if @scoped_context.key?(key)
+        @provided_values[key]
+      end
+
+      def to_h
+        @provided_values.merge(@scoped_context)
+      end
+      alias :to_hash :to_h
+
+      def key?(key)
+        @scoped_context.key?(key) || @provided_values.key?(key)
+      end
 
       # @return [GraphQL::Schema::Warden]
       def warden
@@ -201,10 +213,6 @@ module GraphQL
 
       def scoped_merge!(hash)
         @scoped_context = @scoped_context.merge(hash)
-      end
-
-      def scoped_get(key)
-        @scoped_context[key]
       end
 
       class FieldResolutionContext
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -155,6 +155,7 @@ module GraphQL
         @path = []
         @value = nil
         @context = self # for SharedMethods
+        @scoped_context = {}
       end
 
       # @api private
@@ -162,6 +163,9 @@ module GraphQL
 
       # @api private
       attr_writer :value
+
+      # @api private
+      attr_accessor :scoped_context
 
       def_delegators :@provided_values, :[], :[]=, :to_h, :to_hash, :key?, :fetch, :dig
       def_delegators :@query, :trace, :interpreter?
@@ -193,6 +197,14 @@ module GraphQL
       def received_null_child
         @invalid_null = true
         @value = nil
+      end
+
+      def scoped_merge!(hash)
+        @scoped_context = @scoped_context.merge(hash)
+      end
+
+      def scoped_get(key)
+        @scoped_context[key]
       end
 
       class FieldResolutionContext
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -58,7 +58,7 @@ module GraphQL
             write_in_response(path, nil)
             nil
           else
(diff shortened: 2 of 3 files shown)


lib/graphql/query/context.rb
```diff
@@ -155,6 +155,7 @@ def initialize(query:, values: , object:)
         @path = []
         @value = nil
         @context = self # for SharedMethods
+        @scoped_context = {}
```
lancelafontaine: My initial implemented relied on `Context@scoped_context` being an `ImmutableHash` instead of an `Hash`. I ended up opting out of that decision since it provides little benefit unless we can sure that _all_ values, no matter how deeply nested, are immutable. That being said, with `@scoped_context` being a `Hash`, there's nothing stopping an arbitrary resolver from keeping a reference to an object which is used in a separate scoped context, and modifying it from there. I think that's okay though, since doing so would invalidate the goal of keeping this reference only for a field and its children.

rmosolgo:
:+1: Thanks for taking a crack at this! Nice to see how little overhead is required. Thanks for the comprehensive tests, too. 

One general question I have about the design is, should we require uses to distinguish between _scoped_ context values (using `scoped_get`) and unscoped ones (using `ctx[key]`)? Alternatively, we could overload `ctx[key]` to check scoped _and_ unscoped values. Similarly, should `ctx.to_h` and `ctx.key` include key-value pairs which are scoped? (I _think_ they should, what do you think?) 

I guess what I'm getting at is, what if `.scoped_merge!` added values to `ctx` which were _just like other values_ except that they're only visible to descendant fields. Do you think it would be to magical? Or do you see a specific advantage to distinguishing between scoped values as "normal" values?

lancelafontaine:
Having scoped values be fetcheable from the existing methods Hash-like methods could work! I'm assuming that if a scoped value is set, it should always take precendence over the unscoped value. In that case, we'd have to be okay with always showing the scoped value if it is set, even if the unscoped value was set more recently. Eg.

```ruby
context.scoped_merge!(value: 'old')
context[:value] = 'new'
assert_equal('old', context[:value])
assert_equal({ value: 'old' }, context.to_h)

# later, in sibling resolver ...
assert_equal('new', context[:value])
assert_equal({ value: 'new' }, context.to_h)
```

Otherwise, the implementation would need to keep track of the order in which scoped vs. unscoped values were inserted, or at least which one was most recently inserted, which also works.

rmosolgo:
> always showing the scoped value if it is set 

:+1:

lancelafontaine:
With the new API for getters, this is ready for a round of review 😄

lib/graphql/query/context.rb
```diff
@@ -163,15 +164,30 @@ def initialize(query:, values: , object:)
...
       attr_writer :value
 
-      def_delegators :@provided_values, :[], :[]=, :to_h, :to_hash, :key?, :fetch, :dig
-      def_delegators :@query, :trace, :interpreter?
+      # @api private
+      attr_accessor :scoped_context
 
-      # @!method [](key)
-      #   Lookup `key` from the hash passed to {Schema#execute} as `context:`
+      def_delegators :@provided_values, :[]=
```
swalkinshaw: 🤔 is it inconsistent that the getter `[]` will magically return scoped values but the setter `[]=` won't?

lib/graphql/query/context.rb
```diff
@@ -163,15 +164,30 @@ def initialize(query:, values: , object:)
...
       #   Reassign `key` to the hash passed to {Schema#execute} as `context:`
 
+      # Lookup `key` from the hash passed to {Schema#execute} as `context:`
+      def [](key)
+        return @scoped_context[key] if @scoped_context.key?(key)
+        @provided_values[key]
+      end
+
+      def to_h
+        @provided_values.merge(@scoped_context)
```
swalkinshaw: Should this mutate with `merge!`?

lib/graphql/query/context.rb
```diff
@@ -195,6 +211,10 @@ def received_null_child
         @value = nil
       end
 
+      def scoped_merge!(hash)
+        @scoped_context = @scoped_context.merge(hash)
```
swalkinshaw: Should this mutate with `merge!`?

lib/graphql/query/context.rb
lancelafontaine: The API could be a little confusing if the developer isn't aware of the existence of `#scoped_merge!`. Using a combination of `#[]=` and `#scoped_merge` can also lead to misleading behaviour with `#[]` because it always retrieves the value from `@scoped_context` if it exists regardless of the existing value in `@provided_values`, as described in https://github.com/rmosolgo/graphql-ruby/pull/2634#issuecomment-563289122

```ruby
context.scoped_merge!(value: 'old')
context[:value] = 'new'
assert_equal('old', context[:value])
```

I can definitely see how this usage of the API can be misleading. I could also see this more as an edge case though. This only arises when values are set in ☝️ configuration, and using `scoped_merge!(special_key: 1)` in combination with `[:special_key] = 2` is probably not what that developer _means_ to do, because by doing `[:special_key] = 2`, they are voiding the benefit of having `:special_key` be scoped.

lib/graphql/query/context.rb
lancelafontaine: Nope! This needs to create a new object, because lazy resolvers being executed out-of-order still need an untainted reference to the `scoped_context` they had originally. If this proves to be allocating too many objects or consuming too much memory, we should consider using an immutable data structure with structural sharing instead of `Hash` here.

lib/graphql/query/context.rb
lancelafontaine: See https://github.com/rmosolgo/graphql-ruby/pull/2634#discussion_r356284461 🙃

lib/graphql/query/context.rb
swalkinshaw: Yeah I think what you have makes sense 👍 

Also wondering if `set_scoped(key, value)` should exist in addition to `scoped_merge!` as a convenience. If a common use case is setting a single key then it might match people's normal usage of setting a hash key?

Meaning most people would use `[]=` instead of `merge`

lancelafontaine: Yep! I like `set_scoped(key, value)` as a convenience. I'll push a commit shortly.

lancelafontaine: Added `Context#scoped_set(key, value)` in commit https://github.com/rmosolgo/graphql-ruby/pull/2634/commits/afc292af7ca85762eea59b7693b1c0c74dad2a6b.

rmosolgo:
🎉 Thanks for all your work exploring and implementing this! I'm going to follow up on some old issues asking for this feature.

tjwallace:
Would it be possible to add/update docs with how to use this? Looks like a useful feature!

PR #2637, opened 2019-12-06
Apply visibility checks when generating SDL

jturkel:
This change updates SDL generation to apply visibility checks including the reachability pruning from #2596.


rmosolgo:
🎉 Thanks for taking care of this! 

Regarding `GraphQL::Schema::Context`, can you help me understand why that's necessary? Is it so that user code can continue to use `ctx.schema`? 

As an alternative implementation, what about using `GraphQL::Query::Context`, but passing `query: nil`?

jturkel:
Yup. The context passed to `visible?` checks needs access to the `schema` to support code like [this](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/interface.rb#L25). Passing `query: nil` to the `GraphQL::Query::Context` won't work OOTB because the `@schema` instance variable is initialized to `query.schema`. I added `schema` as an optional `GraphQL::Query::Context` constructor argument with a default value of `query.schema` to fix this. That still leaves the `trace`, `interpreter?` and `warden` methods that will fail in potentially non-obvious ways if invoked but I guess that's a small price to pay for keeping the applicable parts of `GraphQL::Query::Context` and `GraphQL::Schema::Context` in sync.

lib/graphql/query/context.rb
```diff
@@ -309,6 +309,3 @@ def list_of_non_null_items?(type)
     end
   end
 end
-
-
-GraphQL::Schema::Context = GraphQL::Query::Context
```
rmosolgo: :+1: Looks like orphaned code from my first try at custom context classes, thanks for the cleanup.

lib/graphql/language/document_from_schema_definition.rb
```diff
@@ -250,7 +256,7 @@ def build_definition_nodes
         definitions = []
         definitions << build_schema_node if include_schema_node?
         definitions += build_directive_nodes(warden.directives)
-        definitions += build_type_definition_nodes(warden.types)
+        definitions += build_type_definition_nodes(warden.reachable_types)
```
rmosolgo: 🤘

PR #2640, opened 2019-12-09
Explicitly list RUBY_KEYWORDS in GraphQL::Schema::Member::HasFields

DmitryTsepelev:
Some time ago I've [added](https://github.com/rmosolgo/graphql-ruby/pull/2559) a check to make sure user not defines a field with a name that matches one of Ruby keywords. In order to do that I've used a `RubyToken` module, which was [removed](https://github.com/ruby/ruby/commit/45bb6f28db04a1f267f5b8d79392cb35087510d8) from Ruby a couple of weeks ago, so that hack doesn't work with ruby 2.7.0-preview3 (thanks @ahlucas for pointing that out!).

After a short research I found out that `irb/ruby-token` was not used for a long time (so removing it was a matter of time). Also, I could not find a new way to list all the Ruby keywords in the runtime (I guess because neither ruby nor [parser](https://github.com/whitequark/parser) need to check _if_ a token is a keyword, they need to know _the name_ of the keyword).

Taking into the consideration the fact that keywords are not added to often, we could switch back to listing them explicitly. What do you think?

By the way, what do you think of adding RUBY HEAD to the test matrix?

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require 'irb/ruby-token'
 
 module GraphQL
   class Schema
@@ -43,9 +42,7 @@ module GraphQL
         # A list of Ruby keywords.
         #
         # @api private
-        RUBY_KEYWORDS = RubyToken::TokenDefinitions.select { |definition| definition[1] == RubyToken::TkId }
-                                                   .map { |definition| definition[2] }
-                                                   .compact
+        RUBY_KEYWORDS = [:class, :module, :def, :undef, :begin, :rescue, :ensure, :end, :if, :unless, :then, :elsif, :else, :case, :when, :while, :until, :for, :break, :next, :redo, :retry, :in, :do, :return, :yield, :super, :self, :nil, :true, :false, :and, :or, :not, :alias, :defined?, :BEGIN, :END, :__LINE__, :__FILE__]
 
         # A list of GraphQL-Ruby keywords.
         #


rmosolgo:
Thanks for the fix here! I agree that a literal list is fine, since it's likely to be very stable. 

As for RUBY_HEAD, yes, that would be fine with me!

connorshea:
I was just testing Ruby 2.7 and ran into this issue, could we get a new pre-release of 1.10 with this fix? No rush, just wanted to make sure to ask for it. Thanks :)

rmosolgo:
Thanks for the bump on that. I just shipped 1.9.17 with this patch, and I'm currently trying to get master merged into 1.10-dev. Lots going on, not an easy merge 😅

connorshea:
Yeah, I tried doing the merge myself last night so I could use the gem from my fork to test 2.7, and saw how many conflicts there were 😰

Good luck! 🙇

rmosolgo:
🚢 1.10.0.pre3 ! As always, please let me know if you run into any trouble with it.

connorshea:
My test suite passes on Ruby 2.6 and 2.7 with 1.10.0.pre3, LGTM 👍 

Should I open an issue for the 2.7 deprecation warnings that graphql-ruby causes?

swalkinshaw:
We started getting warnings from this for fields *without* defined resolver methods. Right now this is unconditionally warning just based on field name but that's actually completely safe to do as long as an explicit resolver is not defined.

This is safe because the default resolver implementation is called and it could be looking up a hash key named `end`:
```ruby
field :end, DateTime, null: false
```

This is also safe because `method` is used:
```ruby
field :end, DateTime, null: false, method: :ends_at
```

This is not safe:
```ruby
field :end, DateTime, null: false

def end
  some_custom_logic
end
```

swalkinshaw:
Actually this was already an issue with the logic for dealing with the existing `CONFLICT_FIELD_NAMES`. We just didn't happen to have any but now that a lot of names have been added it triggered some.

The main issue is that `resolver_method` defaults to the field name and the logic uses `field_defn.original_name == field_defn.resolver_method` which by default is true.

rmosolgo:
Yeah, the tough part is, when `field(...)` is evaluated, `def end` _hasn't been_ evaluated yet, so I'm not really sure how you could distinguish between the OK and NOT OK cases above 😖 

I started to work on a fix here when I ran into that issue: https://github.com/rmosolgo/graphql-ruby/pull/new/fix-method-warning

swalkinshaw:
Ah true. There's one other conditional that's easy to fix at least and it's checking for `method != original_name`. I'll try and fix that one at least!

DmitryTsepelev:
What if we move these checks to custom rubocop cops? If I'm not mistaken, cops can observe the whole class

rmosolgo:
That seems promising!

PR #2643, opened 2019-12-10
Backport original Base64.urlsafe_encode64 performance improvement

pocke:
We backport Base64.urlsafe_encode64 method for old Rubies. And the method has been updated in Ruby master branch with a performance patch.
So this pull request applies the patch to graphql-ruby also.

ref
https://github.com/ruby/ruby/commit/1bdabaa6b13344c195698ca5b0ced323cb93e2e1
https://github.com/ruby/ruby/pull/1676

Thanks!


rmosolgo:
Great, thanks for the improvement!

PR #2644, opened 2019-12-10
Add Rails master and ruby-head to travis matrix

DmitryTsepelev:
This will probably help to identify breaking changes at the early stage


rmosolgo:
And it _passes_ 🎉 😌 ! 

Thanks for taking care of this. It looks like there are just a _few_ keyword splats to clean up on Ruby 2.7 😅

PR #2648, opened 2019-12-13
Use count directly because mongoid was fixed

### LINKED ISSUE #2646: Mongo hasNextPage counts all records
It looks like https://jira.mongodb.org/browse/MONGOID-2325 is now closed in Mongoid version 3.0.6 so https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/relay/mongo_relation_connection.rb#L20 could check Mongo version if it was so inclined. I patched it in my implementation and it seems to work locally. I didn't see a clear test of this in the specs and wasn't sure how you would like it tested if I implemented it.

rmosolgo:
We had this bit in there ever since this class was added (#1452), but it seems like it was fixed long ago. 

We test on mongoid 5, 6, and 7, so let's see if they're all in the clear. 

Fixes #2646

### DIFF
--- a/lib/graphql/relay/mongo_relation_connection.rb
+++ b/lib/graphql/relay/mongo_relation_connection.rb
@@ -16,8 +16,7 @@ module GraphQL
       end
 
       def relation_count(relation)
-        # Must perform query (hence #to_a) to count results https://jira.mongodb.org/browse/MONGOID-2325
-        relation.to_a.count
+        relation.count
       end
 
       def limit_nodes(sliced_nodes, limit)


rmosolgo:
To me, it seems like something is _actually wrong_ here, still. I put some debugging code to check the different ways of counting: 

```ruby 
p ["sliced", sliced_nodes.count, sliced_nodes.to_a.count]
# ["sliced", 3, 1]
```

And it seems like `.to_a.count` was still returning a different value than plain `.count`. Here's the mongoid version: 

```
~/code/graphql-ruby $ BUNDLE_GEMFILE=gemfiles/mongoid_7.gemfile bundle show mongoid
/Users/rmosolgo/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0/gems/mongoid-7.0.4
```

PR #2651, opened 2019-12-18
New interpreter not reporting certain type of errors

kamal:
This PR illustrates an issue we encountered with `GraphQL::Execution::Interpreter` where it did not return the error when it saw a non-nullable array field returning nulls. The expected behavior is shown with a simple schema that does not use `GraphQL::Execution::Interpreter`.

I'd be happy to work on a fix if I knew where to look.

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -339,8 +339,7 @@ module GraphQL
               set_type_at_path(next_path, inner_type)
               # This will update `response_list` with the lazy
               after_lazy(inner_value, owner: inner_type, path: next_path, scoped_context: scoped_context, field: field, owner_object: owner_object, arguments: arguments) do |inner_inner_value|
-                # reset `is_non_null` here and below, because the inner type will have its own nullability constraint
-                continue_value = continue_value(next_path, inner_inner_value, field, false, ast_node)
+                continue_value = continue_value(next_path, inner_inner_value, field, inner_type.non_null?, ast_node)
                 if HALT != continue_value
                   continue_field(next_path, continue_value, field, inner_type, ast_node, next_selections, false, owner_object, arguments)
                 end
--- a/spec/graphql/execution/errors_spec.rb
+++ b/spec/graphql/execution/errors_spec.rb
@@ -105,10 +105,26 @@ describe "GraphQL::Execution::Errors" do
       field :input_field, Int, null: true do
         argument :values, ValuesInput, required: true, method_access: false
       end
+
+      field :non_nullable_array, [String], null: false
+      def non_nullable_array
+        [nil]
+      end
     end
 
     query(Query)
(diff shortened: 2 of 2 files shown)


kamal:
@jturkel the documentation says if you don’t customize this, you get the default behavior

> Unexpected nils add an error the response’s "errors" key

which is not the case here

rmosolgo:
Thanks for the detailed write-up and the failing case! I definitely want to get to the bottom of it. If it was me, here are some places I would look: 

- When handling an value of List type, we go here:
  
  https://github.com/rmosolgo/graphql-ruby/blob/695960f5c7b35be3a0f87b9ee6cf7c8af39a8b27/lib/graphql/execution/interpreter/runtime.rb#L334-L349

- Then, I'd expect the `continue_value` call to go here: 

  https://github.com/rmosolgo/graphql-ruby/blob/695960f5c7b35be3a0f87b9ee6cf7c8af39a8b27/lib/graphql/execution/interpreter/runtime.rb#L246-L253

- Which calls `write_invalid_null_at_path`, which is apparently _not doing the right thing_: 

  https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/interpreter/runtime.rb#L551-L557

- It seems to call `Schema.type_error`, which is defined here for class-based schemas:

  https://github.com/rmosolgo/graphql-ruby/blob/695960f5c7b35be3a0f87b9ee6cf7c8af39a8b27/lib/graphql/schema.rb#L1029-L1031

- and implemented:

  https://github.com/rmosolgo/graphql-ruby/blob/695960f5c7b35be3a0f87b9ee6cf7c8af39a8b27/lib/graphql/schema/default_type_error.rb#L4-L13

  (It _seems_ like it should add the error there!)

- When the class-based schema is converted to a legacy `.define`-based schema singleton, the handler is _supposed_ to be passed along here:
  
  https://github.com/rmosolgo/graphql-ruby/blob/695960f5c7b35be3a0f87b9ee6cf7c8af39a8b27/lib/graphql/schema.rb#L769

  But that may not apply for the interpreter, I bet the interpreter would use the schema class directly, not the legacy object.

I hope some of those links provide a clue! Let me know if you get to take a look, and what you find 🍻

kamal:
@rmosolgo I ran byebug through the steps you outlined, but found that `write_invalid_null_in_response` is never evaluated for the inner type. Despite the comments, I applied my change of passing through the inner type's `non_null?` value instead of resetting it to `false` and all the specs pass. (current failure is against `ruby-head` which is the same as `master`)

I'm unsure of this change as it goes against the comment but nothing seems to break.

rmosolgo:
Thanks for the fix here! Sorry I didn't make it back around to check until now. 

Too bad about that code comment -- I suppose it might the ultimate fate of all code comments: stale and unhelpful. In any case, the specs you added look right to me, so I must have hacked it wrong at the start.

Issue #2650, opened 2019-12-18
Ruby 2.7 deprecations


connorshea:
This is on `1.10.0.pre3`, so the line numbers might be a bit different from master. All of these are from the positional args/kwargs change described in [this blog post](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

The deprecation warnings are in pairs because one is the callsite and one is the method being called.

```
/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/schema/mutation.rb:71: warning: The last argument is used as the keyword parameter
/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/schema/member/has_fields.rb:11: warning: for `field' defined here

/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/execution/multiplex.rb:48: warning: The last argument is used as the keyword parameter
/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/query.rb:81: warning: for `initialize' defined here

/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/execution/multiplex.rb:49: warning: The last argument is used as the keyword parameter
/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/execution/multiplex.rb:57: warning: for `run_queries' defined here

/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/language/nodes.rb:37: warning: The last argument is used as the keyword parameter
/.rbenv/versions/2.7.0-preview3/lib/ruby/gems/2.7.0/gems/graphql-1.10.0.pre3/lib/graphql/language/nodes.rb:259: warning: for `initialize_node' defined here
```

Links:
[...]

PR #2653, opened 2019-12-21
Fix some Ruby 2.7 deprecation warnings.

connorshea:
Fixes part of #2650. This fixes all the deprecations that I had when running graphql-ruby with my Rails app, and removes almost 1000 deprecation warnings from my test suite. Looking at the travis logs for ruby-head there are some deprecation warnings my changes don't seem to fix. A lot of the deprecations from ruby-head in travis are external libraries that can be fixed by just updating them, e.g. rake, rails, and rubocop.

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -134,9 +134,9 @@ module GraphQL
       end
 
       if type_or_argument.is_a?(GraphQL::Argument)
-        type_or_argument.redefine(kwargs, &block)
+        type_or_argument.redefine(**kwargs, &block)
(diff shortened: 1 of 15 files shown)


connorshea:
I fixed a few more and that reduced the output for the graphql-ruby CI job by ~200 lines. The remaining ones are pretty weird, so I'm tapping out :P

@rmosolgo Hopefully this is helpful! And no rush, please enjoy your Christmas :)

rmosolgo:
Thanks for knocking out so many of these!

Issue #2611, opened 2019-11-20
Deserialization of arguments with numbers appears not to work


TomasBarry:
Given the following input type:

```ruby
module Inputs
  class MyInput < Types::BaseInputObject
    argument :some_argument, String, required: true
    argument :some_argument_1, String, required: true
  end
end
```

The following mutation does not have the arguments deserialized as expected:

```ruby
module Mutations
  class MyMutation < BaseMutation
    argument :my_input, MyInput, required: true

    type Types::SomeType

    def resolve(my_input:)
      # my_input.some_argument is present, as expected
      # my_input.some_argument_1 is not present
      # instead, it takes the form of my_input.some_argument1
    end
end
```

The query being used is:

```graphql
mutation MyMutation(
  $myInput: MyInput!
) {
  myMutation(
    myInput: $myInput
  ) {
    ...
  }
}
```

Where the variables are:

```es6
const body = JSON.stringify({
  variables: {
    myInput: {
      someArgument: 'some string',
      someArgument1: 'some other string'
    }
  }
})
```

Is this a bug, or is it intended behaviour?

TomasBarry:
Data was not actually being sent as expected. No bug present.

TomasBarry:
@rmosolgo, any ideas on this one?

rmosolgo:
Sorry, it's a bug! It should provide access using the same name as you give to `argument ...`. I don't really have the bandwidth to look into it right now, but I'd be happy to help review a PR if you want to give it go!

TomasBarry:
Thanks for confirming @rmosolgo, could you point me in the right direction on where to start an investigation? Would love to have a look at it.

rmosolgo:
Sure, the first step would be to add a failing test. Then the challenge is to find _where_ the hash is being built with the wrong keys. This is my first guess, if you're using the new interpreter runtime: 

https://github.com/rmosolgo/graphql-ruby/blob/ac2d2354607d39c2c18711e15489ff9d8314eca3/lib/graphql/execution/interpreter/runtime.rb#L458

Otherwise, it might be happening in this code: 

https://github.com/rmosolgo/graphql-ruby/blob/ac2d2354607d39c2c18711e15489ff9d8314eca3/lib/graphql/schema/input_object.rb#L15-L17

Which is a call to this method: 

[...]

PR #2654, opened 2019-12-22
Fix argument deserialization

TomasBarry:
As outlined in https://github.com/rmosolgo/graphql-ruby/issues/2611
there appears to be a bug in argument deserialization if the argument
contains a number. This bug was confirmed by @rmosolgo [here](https://github.com/rmosolgo/graphql-ruby/issues/2611#issuecomment-558790828)

This commit contains a patch for the bug. The culprit was
`GraphQL::Schema::Member::BuildType#underscore`. Upon investigation,
this method used `gsub` to include underscores but just split the string
based on letters. This was confirmed by adding a test for the
`#underscore` method in `spec/graphql/schema/member/build_type_spec.rb`
which was not initially present.

After identifying the issue, the following two `gsub` calls were chained
to the `underscore` method:

```ruby
.gsub(/([A-Za-z]+)([0-9]+)/,'\1_\2')  # some1 -> some_1
.gsub(/([0-9]+)([A-Za-z]+)/,'\1_\2')  # 1thing -> 1_thing
```

Having added the tests that confirm the bug has been fixed, a number of
other tests failed due to the test code using the incorrectly
deserialized values. This commit also patches these tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@
 
 ### Bug fixes
 
+## 1.9.15 (22 Dec 2019)
+
+## Bug fixes
+
+- Fix argument underscoring #2611
+
+### Breaking changes
+
+- Arguments with integers were being incorrectly deserialized #2611
+
 ## 1.9.14 (14 Oct 2019)
 
 ## New features
--- a/lib/graphql/schema/member/build_type.rb
+++ b/lib/graphql/schema/member/build_type.rb
@@ -161,6 +161,8 @@ module GraphQL
           string
             .gsub(/([A-Z]+)([A-Z][a-z])/,'\1_\2') # URLDecoder -> URL_Decoder
             .gsub(/([a-z\d])([A-Z])/,'\1_\2')     # someThing -> some_Thing
+            .gsub(/([A-Za-z]+)([0-9]+)/,'\1_\2')  # some1 -> some_1
+            .gsub(/([0-9]+)([A-Za-z]+)/,'\1_\2')  # 1thing -> 1_thing
             .downcase
         end
       end
--- a/lib/graphql/version.rb
+++ b/lib/graphql/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module GraphQL
-  VERSION = "1.9.14"
+  VERSION = "1.9.15"
 end
--- a/spec/graphql/authorization_spec.rb
+++ b/spec/graphql/authorization_spec.rb
@@ -273,14 +273,14 @@ describe GraphQL::Authorization do
 
       field :hidden, Integer, null: false
       field :unauthorized, Integer, null: true, method: :itself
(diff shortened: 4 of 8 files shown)


TomasBarry:
@rmosolgo, if this PR is fit for purpose, would you rather the extra files be submitted as patches in a set of other commits so that this commit just deals with the bug itself (I mean submitting the changes to `spec/graphql/authorization_spec.rb`, `spec/graphql/execution/errors_spec.rb`, `spec/graphql/schema/input_object_spec.rb`, and `spec/graphql/schema/resolver_spec.rb`)?

rmosolgo:
Sorry for the slow response here. I'm a little bit wary of this change because it looks like it will change behavior for anyone using names like `field1`, is that right? I'm worried we'll trade _one_ unexpected behavior for another. 

Do you think it's possible to respect the case _as written_ in the field name? That is, if someone writes `field :f_1`, then we use `f_1`, but if they write `f2`, we use `f2`?

TomasBarry:
> I'm a little bit wary of this change because it looks like it will change behavior for anyone using names like `field1`, is that right? I'm worried we'll trade _one_ unexpected behavior for another.

Yeah, that is the case which is why I'm not too keen on it either.

> Do you think it's possible to respect the case _as written_ in the field name? That is, if someone writes `field :f_1`, then we use `f_1`, but if they write `f2`, we use `f2`?

That's something that I'll have to look into. Any pointers on where to start looking for that?

rmosolgo:
> where to start

I think it basically requires a paradigm shift on how this works. Currently, the library _assumes_ that: 

- Ruby fields use underscored names;
- AND GraphQL fields use camelized name;
- THEREFORE a `camelize(...)` or `underscore(...)` transformation can be used to translate between Ruby and GraphQL. 

However, the issue you outlined over in #2611 demonstrates a flaw in this approach: we have differing expectations about how the transform should work -- or at least, the way it _previously_ worked was wrong. 

So, I think a better approach to this feature would be: 

- Ruby fields are given names using `field(name, ...)` configurations; 
- AND GraphQL names are generated during that configuration (often with `camelize: true`, but not necessarily so)
- THEREFORE the configured `GraphQL::Schema::Field` instance should be used to map GraphQL and Ruby names for different fields 

In an approach like that, `camelize(...)` is used only to generate a GraphQL name for the field. After that, no transform is used. Instead, the field objects themselves are used to match a GraphQL name to a Ruby name. 

I'm not _exactly_ sure what will be required to implement that, but I think you could basically start by: 

- Looking for places where `underscore(...)` is used to transform field names, then 
- Finding a way to replace that transformation with a lookup of the name used in the configuration

But, I'm not exactly sure how that will play out 🙈 !

rmosolgo:
I took a different approach in #2792 which maintained the current behavior, too. Please give it a try on master or 1.10.4 which will be released soon, and let me know if you run into any trouble!

PR #2656, opened 2020-01-03
fix: Use schema context class for SDL generation

cmschuetz:
This change ensures that visibility checks during SDL generation receive an instance of the context class set on the schema.  Related change: https://github.com/rmosolgo/graphql-ruby/pull/2637


rmosolgo:
Thanks for the fix! I'll have to look into that rubocop failure.

Issue #2546, opened 2019-10-14
Disable only one introspection entry point


TomasBarry:
Following the work done [here](https://github.com/rmosolgo/graphql-ruby/pull/2327), is it possible to only disable the `__schema` introspection entry point?

If not, would a PR similar to [this](https://github.com/rmosolgo/graphql-ruby/pull/2327) (which introduced the `disable_introspection_entry_points` method) be accepted?

The PR would introduce two more methods:

1) `enable_schema_introspection_entry_point`
2) `enable_type_introspection_entry_point`

A case has arisen where we would like to disable only the `__schema` entry point, or in other words, only expose the `__type` entry point (should any more entry points be added in future).

The usage would look something like:

```ruby
class MySchema < GraphQL::Schema
  # disable all introspection entry points by default
  disable_introspection_entry_points

  # enable the __type introspection entry point
  enable_type_introspection_entry_point

  mutation(Types::MutationType)
  query(Types::QueryType)
end
```

Alternatively, we could just introduce `disable_X_introspection_entry_point` methods for the two existing introspection entry point methods:

1) `disable_schema_introspection_entry_point`
2) `disable_type_introspection_entry_point`

Therefore, the usage would be like:

```ruby
class MySchema < GraphQL::Schema
  # disable the __schema introspection entry point while leaving others (__type) available
  disable_schema_introspection_entry_point

  mutation(Types::MutationType)
  query(Types::QueryType)
end
```

ryanfreddy:
I think this would be a very useful feature.

rmosolgo:
You could implement this dynamically by customizing the introspection system: 

https://graphql-ruby.org/schema/introspection.html#customizing-introspection

For example, you could raise an error if someone accesses an entry point you don't want to support. Alternatively, you could try defining an `EntryPoints` class that _doesn't_ have that field. (You'd riff off the built-in `EntryPoints` class, but I don't think you could use it as a base class, since it already includes `__schema`.) 

Finally, yes, if you want to make the change, I'd accept it! 

Hope that helps and please let me know if you try the custom introspection approach and run into trouble.

TomasBarry:
Which would be preferable:

1) Add methods to disable specific entry points (`disable_schema_introspection_entry_point`, `disable_type_introspection_entry_point`, etc)
[...]

PR #2657, opened 2020-01-04
Implement `__type`/`__schema` introspection disabling

TomasBarry:
A question was asked in issue GH-2546 asking whether it was possible to
disable a single introspection entry point considering that #2327 had
implemented then functionality to disable introspection as a whole using
`disable_introspection_entry_points`. In the issue, an option was laid
out to implement the functionality for individual
`disable_schema_introspection_entry_point` and
`disable_type_introspection_entry_point` options. This commit implements
these methods.

These new methods are used in an identical manner to
`disable_introspection_entry_points`:

 # Disabling `__type` introspection

```ruby
  disable_type_introspection_entry_point
end
```

With the above, attempting to execute a query with `__type`
introspection will return the following error message:

```
Field '__type' doesn't exist on type 'Query'
```

 # Disabling `__schema` introspection
```ruby
  disable_schema_introspection_entry_point
end
```

With the above, attempting to execute a query with `__schema`
introspection will return the following error message:

```
Field '__schema' doesn't exist on type 'Query'
```

Note, having both `disable_type_introspection_entry_point` and
`disable_schema_introspection_entry_point` is identical to having just
`disable_introspection_entry_points`. Having
`disable_introspection_entry_points` and
`disable_type_introspection_entry_point` or
`disable_schema_introspection_entry_point` is no different to just
having `disable_introspection_entry_points`.

### DIFF
--- a/guides/schema/introspection.md
+++ b/guides/schema/introspection.md
@@ -197,3 +197,12 @@ class MySchema < GraphQL::Schema
   disable_introspection_entry_points if Rails.env.production?
 end
 ```
+
+Where `disable_introspection_entry_points` will disable both the `__schema` and `__type` introspection entry points, you can also individually disable the introspection entry points using the `disable_schema_introspection_entry_point` and `disable_type_introspection_entry_point` shorthand methods:
+
+```ruby
+class MySchema < GraphQL::Schema
+  disable_schema_introspection_entry_point
+  disable_type_introspection_entry_point
+end
+```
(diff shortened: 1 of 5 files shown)


rmosolgo:
Thanks for your work here! It looks great. I just pushed some changes to address the lint errors, then I'll merge it.

TomasBarry:
For some reason I did not spot those errors. Thank you.

PR #2665, opened 2020-01-11
Add Rails 6 and Ruby 2.7 to CI

connorshea:
And allow ruby-head/rails-master to fail.


rmosolgo:
Cool, it looks like sequel will work again when this commit is released to rubygems, too: https://github.com/jeremyevans/sequel/commit/fb6675e82f14477f21bf4cfb4e36a91335e55f24

PR #2673, opened 2020-01-15
Improve sync response handling

rmosolgo:
Fix a few issues with the previous code: 

- It only supported one chunk. (I haven't heard of this being an issue, but it _could_ be!)
- It didn't log the response with `verbose`


sgwilym:
We just got hit by this today 😅

rmosolgo:
Sorry for the hangup! Just shipped in 1.7.4

PR #2675, opened 2020-01-16
Add AppopticsAPM tracing

prettycoder:
add `GraphQL::Tracing::AppOpticsTracing` including documentation and tests


guides/queries/tracing.md
```diff
@@ -66,6 +66,22 @@ Leaf fields are _not_ monitored (to avoid high cardinality in the metrics servic
 
 Implementations are based on {{ "Tracing::PlatformTracing" | api_doc }}.
 
+## AppOptics
+[AppOptics](https://appoptics.com/) instrumentation will be automatic starting 
+with appoptics_apm-4.11.0.gem. For earlier gem versions please add appoptics_apm 
+tracing as follows:
```
rmosolgo: That's awesome that a future version of appoptics_apm will install a GraphQL tracer automatically. 

Is it reasonable to suggest that users of older versions update to 4.11 in order to get GraphQL tracing? It would save us the trouble of maintaining a tracer here, and besides, it would encourage folks to get on the latest version! But -- it might not work if there are reasons people can't upgrade. 

What do you think, would it be a reasonable solution to recommend an upgrade _instead_ of shipping a plugin?

prettycoder: Thanks, yes, we thought about this too. Three reasons to add appoptics in graphql:

- unfortunately, we have no influence over when and what people upgrade.
- if there are breaking changes to PlatformTracing in graphql, it would be easier to stay in sync with code within the graphql gem.
- AppOptics is less known than others, so we thought it would be nice to join the lineup in the graphql gem. 

In case you're interested, here is the code for auto-instrumentation: https://github.com/appoptics/appoptics-apm-ruby/blob/e772e0b1d8af8ea009e575e7d6e30bc9ceb096f0/lib/appoptics_apm/inst/graphql.rb#L185

Let me talk to my project manager one more time. I think it would be nice to have a paragraph in the tracing.md doc, but ultimately maintaining the code in 2 places is a bit of a hassle.

rmosolgo: Sure, that's a good point about keeping code in sync. I don't _plan_ to break anything but, then again, who _does_ 😅 

In terms of visibility, we could also do something like [Skylight's paragraph](https://graphql-ruby.org/queries/tracing.html#skylight). Early on, I wrote a tracer for Skylight, but recently, they added built-in support. So one of their devs updated the docs to describe their built-in support and deprecate the one that I built a while back. 

As far as upgrading, even if the code is part of graphql-ruby, folks will need to upgrade to use it -- so I think an upgrade will be required either way. 

In any case, I'm happy to include the code if that's what you prefer, I just thought I'd mention it since I read that built-in support was on the way! Just let me know.

prettycoder: Ok, we decided that we would like our plugin in graphql. Can you please go ahead with this PR? Thanks. 

If ever something goes wrong with this code or tests, it will be best to create an issue here and we will respond very quickly: https://github.com/appoptics/appoptics-apm-ruby/issues

PS: There is one test suite not passing, but it is not related to the code in this PR.

rmosolgo:
Thanks again for this contribution! It looks like the git situation got a bit tricky here, so I cherry-picked your commits and merged them in #2789

PR #2676, opened 2020-01-17
Add ActionCableRelaySubscriptions class for using Facebook's Relay Front-end Library

Jakenberg:
Thank you for making this library!


rmosolgo:
Thanks very much for contributing this! I'd like to learn more about how it works so that I can support it in the future. 

Can you help me understand why this different topic format supports Relay better?

rmosolgo:
I'd be happy to merge this, but as I mentioned, I'd like to learn about it so that I can properly maintain it. Please let me know if you have time to share what you know about it and I'll reopen this PR.

PR #2678, opened 2020-01-17
[WIP] Interface type visibility

grcooper:
This is a WIP PR currently. There are a few checks that I put in place where there is some odd behaviour around some methods not existing.

However, the idea behind the implementation is there and I would like to make sure I am on the right track and also get some assistance in fixing these problems as I have gotten a bit stuck.

I also need to implement more tests to check to make sure inheritance is working correctly between interfaces, as well as querying for fields on hidden interfaces, rather than just the introspection query.

I have decided to go about this in a similar way to https://github.com/rmosolgo/graphql-ruby/pull/2391 which sets a `type_membership` on an interface, and any types that implement that interface can pass in some params for when it should be visible. For example:

```ruby
class PrivateMembership < GraphQL::Schema::TypeMembership
  def initialize(*args, visibility: nil, **kwargs)
    @visiblity = visibility
    super(*args, **kwargs)
  end

  def visible?(ctx)
    return true if @visibility.nil?

    @visibility[:private] && @context[:private]
  end
end

class MyInterface

  type_membership_class PrivateMembership

  field :secret_field, String, null: false
end

class MyObject < GraphQL::Schema::Object
  implements MyInterface, visibility: { private: true }

  field :public_field, String, null: false
end
```

In this case, if `context[:private]` is true, then the `secret_field` will show up when Querying for `MyObject`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ yarn.lock
 OperationStoreClient.js
 spec/integration/tmp
 .vscode/
+*.swp
+*.swo
--- a/lib/graphql/object_type.rb
+++ b/lib/graphql/object_type.rb
@@ -38,16 +38,14 @@ module GraphQL
       super
       @fields = {}
       @interface_fields = {}
-      @dirty_interfaces = []
-      @dirty_inherited_interfaces = []
+      @interface_type_memberships = []
+      @inherited_interface_type_memberships = []
     end
 
     def initialize_copy(other)
       super
-      @clean_interfaces = nil
-      @clean_inherited_interfaces = nil
-      @dirty_interfaces = other.dirty_interfaces.dup
-      @dirty_inherited_interfaces = other.dirty_inherited_interfaces.dup
+      @interface_type_memberships = other.interface_type_memberships.dup
+      @inherited_interface_type_memberships = other.inherited_interface_type_memberships.dup
       @fields = other.fields.dup
     end
 
@@ -55,18 +53,14 @@ module GraphQL
     # @param new_interfaces [Array<GraphQL::Interface>] interfaces that this type implements
     # @deprecated Use `implements` instead of `interfaces`.
     def interfaces=(new_interfaces)
-      @clean_interfaces = nil
-      @clean_inherited_interfaces = nil
-      @clean_inherited_fields = nil
-
-      @dirty_inherited_interfaces = []
+      @interface_type_memberships = []
+      @inherited_interface_type_memberships = []
       @dirty_inherited_fields = {}
       implements(new_interfaces, inherit: true)
     end
 
-    def interfaces
-      load_interfaces
-      @clean_interfaces
+    def interfaces(ctx = GraphQL::Query::NullContext)
+      load_interfaces(ctx)
     end
 
     def kind
@@ -91,15 +85,16 @@ module GraphQL
     # This declaration will be validated when the schema is defined.
     # @param interfaces [Array<GraphQL::Interface>] add a new interface that this type implements
     # @param inherits [Boolean] If true, copy the interfaces' field definitions to this type
-    def implements(interfaces, inherit: false)
+    def implements(interfaces, inherit: false, **options)
       if !interfaces.is_a?(Array)
         raise ArgumentError, "`implements(interfaces)` must be an array, not #{interfaces.class} (#{interfaces})"
       end
 
-      @clean_interfaces = nil
-      @clean_inherited_fields = nil
-      dirty_ifaces = inherit ? @dirty_inherited_interfaces : @dirty_interfaces
-      dirty_ifaces.concat(interfaces)
+      type_memberships = inherit ? @inherited_interface_type_memberships : @interface_type_memberships
+      type_memberships_from_interfaces = interfaces.map do |iface|
+        iface.type_membership_class.new(iface, self, options)
+      end
+      type_memberships.concat(type_memberships_from_interfaces)
     end
 
     def resolve_type_proc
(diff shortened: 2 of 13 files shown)


.gitignore
```diff
@@ -22,3 +22,5 @@ yarn.lock
 OperationStoreClient.js
 spec/integration/tmp
 .vscode/
+*.swp
```
grcooper: This is to prevent me from uploading Vim .swp files...

lib/graphql/object_type.rb
```diff
@@ -117,25 +127,33 @@ def normalize_interfaces(ifaces)
     end
 
     def interface_fields
-      load_interfaces
+      load_interfaces unless @clean_inherited_fields
```
grcooper: Have to do caching a layer up now.

lib/graphql/object_type.rb
```diff
@@ -117,25 +127,33 @@ def normalize_interfaces(ifaces)
...
-    def load_interfaces
-      @clean_interfaces ||= begin
-        ensure_defined
-        clean_ifaces = normalize_interfaces(@dirty_interfaces)
-        clean_inherited_ifaces = normalize_interfaces(@dirty_inherited_interfaces)
-        inherited_fields = {}
-        clean_inherited_ifaces.each do |iface|
-          # This will be found later in schema validation:
+    def load_interfaces(ctx = GraphQL::Query::NullContext)
+      ensure_defined
```
grcooper: This is the meat of the change, a big negative is that we lose a lot of caching for interfaces.

lib/graphql/object_type.rb
```diff
@@ -38,35 +38,31 @@ def initialize
...
+      @clean_inherited_fields = nil
+      @cached_interfaces = nil
       implements(new_interfaces, inherit: true)
     end
 
-    def interfaces
-      load_interfaces
-      @clean_interfaces
+    def interfaces(ctx = GraphQL::Query::NullContext)
+      load_interfaces(ctx)
```
grcooper: I used to have a `@cached_interfaces` variable here which would cache interfaces on a null context, but it was leading to a ton of flakey tests. I can dig into it more as to why caching wasn't working.

The error I was getting was that `Cheese` (a GraphQL Object type) was being referenced as a interface on itself... but it isn't an interface so I am not sure how it was even getting assigned as an interface on it's own object.

lib/graphql/schema/object.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "graphql/query/null_context"
```
grcooper: Is there a better place for me to load this? Otherwise I can't have null context default params.

lib/graphql/schema/warden.rb
```diff
@@ -146,10 +147,29 @@ def union_memberships(obj_type)
...
-      def visible_field?(field_defn)
+      def visible_field?(field_defn, type_defn = nil)
+        return false if type_defn && !field_on_visible_interface?(field_defn, type_defn)
         visible?(field_defn) && visible_type?(field_defn.type.unwrap)
       end
 
+      # Returns true when the field is on a visible interface
+      # returns true if the field is not on an interface
+      # returns false if the field is on a hidden interface
+      def field_on_visible_interface?(field_defn, type_defn)
```
grcooper: This function is a bit funky so I will explain what I am doing:

Check if the field is on an interface that is attached to an object it is visible on.

Ie: If interface x has field y
object z implements interface x

If interface x is hidden, show field y, if it is visible, hide field x.

I need all three of these objects in order to determine a fields visibility so I grab the owner from the field to get the interface.

spec/graphql/object_type_spec.rb
```diff
@@ -189,7 +189,7 @@
       type_2.fields["nonsense"] = GraphQL::Field.define(name: "nonsense", type: type)
 
       assert_equal 4, type.interfaces.size
-      assert_equal 5, type_2.interfaces.size
+      assert_equal 4, type_2.interfaces.size
```
grcooper: I need to dig into why this changed more.

rmosolgo:
So far, so good in my opinion! Is there anything particular that's hanging you up now that I can help with?

grcooper:
@rmosolgo

 The main thing was this method https://github.com/rmosolgo/graphql-ruby/pull/2678/files#diff-edaedf574068dd9aee5988de3b6198fcR158-R169

Which encounters some really strange types as shown through the odd `rescue` statements and such.

If you have any ideas around that, it would be appreciated. If you want to be able to test it out -> Remove the `rescue` then run the tests and it encounters some weird types. Or put a `byebug` in the `rescue`

Otherwise I just wanted to make sure this was the right approach and you didn't have any major concerns before I spent a while cleaning up the last 10%.

rmosolgo:
I spent some time with this, with master merged in, over at #2791 . But then I realized you said you want it on 1.9.x. 

I don't think I can really support that -- there's a big code change between 1.9 and master (which is probably why you're waiting to upgrade 😅 ). But, I think I'm going to give it another push on master. It's ... tricky!

eapache:
We're getting closer to 1.10, and I don't think this is urgent enough to try and backport at this point, so that's fine.

grcooper:
That's unfortunate, but makes sense that we can't have it in 1.9.x.

Is there anything I can help with to merge it into master? I probably need a few more tests as well to make sure I am confident in this change.

I also still need to dive into the rescue statements I had to put for some specific tests. If you had any insight on why those might be causing issues that would be awesome otherwise I can continue to dig into them.

rmosolgo:
> anything I can help with

I'm really not _sure_. I just hacked for a while on that other branch. If you want to pull it and hack until things work, go for it! I got to the point where I thought "I'm not getting anywhere, I need to come back later" -- but I haven't really made a plan.

rmosolgo:
OK, I've got everything passing in https://github.com/rmosolgo/graphql-ruby/pull/2791, where I pulled your commits and merged in master. I have a few more TODOs there but I'm going to close this one out in favor of that.

PR #2689, opened 2020-01-22
Fix field extensions after change default value to [] from nil

durcak:
In commit https://github.com/rmosolgo/graphql-ruby/commit/2eb84f0c8bdf41912517f96fb71e89a63f0c04b2#diff-19915f765769f8f600fa85498d1ea515 was changed @extensions default value to [], one condition was not changed.


rmosolgo:
Thanks for the fix here!

PR #2699, opened 2020-01-27
Allow return raw values from resolvers

DmitryTsepelev:
This PR allows to return _raw values_ from resolvers. When Interpreter gets such a value, it should write it directly to the response, instead of visiting all the child nodes. This is how it could look like:

```ruby
  field :expansion_raw, Expansion, null: false
  
  def expansion_raw
    raw_value(sym: "RAW", name: "Raw expansion")
  end
end
```

This change is a first step of implementing partial response caching (current solutions are very limited and skip various built-in checks). Here is an example [gist](https://gist.github.com/palkan/faad9f6ff1db16fcdb1c071ec50e4190) by @palkan, which contains the same change (using metaprogramming) as well as caching implementation.

What do you think? Do you have suggestions on the API design?

### DIFF
--- a/lib/graphql/execution/interpreter/handles_raw_value.rb
+++ b/lib/graphql/execution/interpreter/handles_raw_value.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module GraphQL
+  module Execution
+    class Interpreter
+      # Wrapper for raw values
+      class RawValue
+        attr_reader :object
+
+        def initialize(obj = nil)
+          @object = obj
+        end
+
+        alias_method :resolve, :object
+      end
+
+      # Allows to return "raw" value from the resolver
+      module HandlesRawValue
+        def raw_value(obj = nil)
+          RawValue.new(obj)
+        end
+      end
+    end
+  end
+end
--- a/lib/graphql/execution/interpreter/handles_raw_value.rb
+++ b/lib/graphql/execution/interpreter/handles_raw_value.rb
@@ -5,13 +5,13 @@ module GraphQL
     class Interpreter
       # Wrapper for raw values
       class RawValue
-        attr_reader :object
-
         def initialize(obj = nil)
           @object = obj
         end
 
-        alias_method :resolve, :object
+        def resolve
+          @object
+        end
       end
 
       # Allows to return "raw" value from the resolver
--- a/lib/graphql/execution/interpreter/handles_raw_value.rb
+++ b/lib/graphql/execution/interpreter/handles_raw_value.rb
@@ -16,7 +16,7 @@ module GraphQL
 
       # Allows to return "raw" value from the resolver
       module HandlesRawValue
-        def raw_value(obj = nil)
+        def raw_value(obj)
           RawValue.new(obj)
         end
       end
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -223,7 +223,7 @@ module GraphQL
                 continue_value = continue_value(next_path, inner_result, field_defn, return_type.non_null?, ast_node)
                 if RawValue === continue_value
                   # Write raw value directly to the response without resolving nested objects
-                  write_in_response(next_path, continue_value.object)
+                  write_in_response(next_path, continue_value.resolve)
                 elsif HALT != continue_value
                   continue_field(next_path, continue_value, field_defn, return_type, ast_node, next_selections, false, object, kwarg_arguments)
                 end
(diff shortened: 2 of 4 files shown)


lib/graphql/execution/interpreter/handles_raw_value.rb
```diff
@@ -0,0 +1,25 @@
...
+        def initialize(obj = nil)
+          @object = obj
+        end
+
+        alias_method :resolve, :object
+      end
+
+      # Allows to return "raw" value from the resolver
+      module HandlesRawValue
+        def raw_value(obj = nil)
```
rmosolgo: What's the use case for `raw_value()` (without any argument)? How about requiring a value instead?
```suggestion
        def raw_value(obj)
```

lib/graphql/execution/interpreter/handles_raw_value.rb
```diff
@@ -0,0 +1,25 @@
...
+    class Interpreter
+      # Wrapper for raw values
+      class RawValue
+        attr_reader :object
+
+        def initialize(obj = nil)
+          @object = obj
+        end
+
+        alias_method :resolve, :object
```
rmosolgo: Why have an alias for this method?

```suggestion
```

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -222,7 +221,11 @@ def evaluate_selections(path, scoped_context, owner_object, owner_type, selectio
               end
               after_lazy(app_result, owner: owner_type, field: field_defn, path: next_path, scoped_context: context.scoped_context, owner_object: object, arguments: kwarg_arguments) do |inner_result|
                 continue_value = continue_value(next_path, inner_result, field_defn, return_type.non_null?, ast_node)
-                if HALT != continue_value
+                if RawValue === continue_value
+                  # Write raw value directly to the response without resolving nested objects
+                  write_in_response(next_path, continue_value.resolve)
```
rmosolgo: Could this use the other method instead? 

```suggestion
                  write_in_response(next_path, continue_value.object)
```

spec/graphql/execution/interpreter_spec.rb
```diff
@@ -480,4 +491,21 @@ def self.resolve_type(type, obj, ctx)
...
+        expansionRaw {
+          name
+          sym
+          alwaysCachedValue
+        }
+      }
+      GRAPHQL
+
+      res = InterpreterTest::Schema.execute(query_str)
+      assert_equal({ sym: "RAW", name: "Raw expansion", always_cached_value: 42 }, res["data"]["expansionRaw"])
```
rmosolgo: ✨

lib/graphql/execution/interpreter/handles_raw_value.rb
palkan: The idea is to use `resolve` everywhere for consistency.
We assume that users/libraries would be able to sub-class RawValue and implement their own resolving logic.

For example, CachedValue could be smth like:

```ruby
class CachedValue < GraphQL::Execution::Interpreter::RawValue
  def resolve
     cache_store.read(object.cache_key) || object
  end
end
```

DmitryTsepelev:
I've applied two suggestions, let's figure out if we want to go with that `alias_method`

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -222,7 +221,11 @@ def evaluate_selections(path, scoped_context, owner_object, owner_type, selectio
               end
               after_lazy(app_result, owner: owner_type, field: field_defn, path: next_path, scoped_context: context.scoped_context, owner_object: object, arguments: kwarg_arguments) do |inner_result|
                 continue_value = continue_value(next_path, inner_result, field_defn, return_type.non_null?, ast_node)
-                if HALT != continue_value
+                if RawValue === continue_value
+                  # Write raw value directly to the response without resolving nested objects
+                  write_in_response(next_path, continue_value.object)
```
rmosolgo: Sorry, I understand your choice to use `.resolve` now, and I'm in favor of it. To @palkan's point, let's use it here. 

```suggestion
                  write_in_response(next_path, continue_value.resolve)
```

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -222,7 +221,11 @@ def evaluate_selections(path, scoped_context, owner_object, owner_type, selectio
...
+                  continue_value.object
```
rmosolgo: I think this method's return value is not used, so do we need to include that last line?
```suggestion
```

lib/graphql/execution/interpreter/handles_raw_value.rb
rmosolgo: Thanks for clarifying, that makes sense to me 👌  

In that case, maybe we can get rid of `attr_reader :object` ?

DmitryTsepelev: Agreed, removed it

DmitryTsepelev:
@rmosolgo Could you please take a look at it again when you have a second?

rmosolgo:
Thanks for the bump. Looks good! It will be interesting to see where this goes 😈

PR #2700, opened 2020-01-27
Consider fields not selected in lookahead if skipped by a directive

dylanahsmith:
## Problem

GraphQL::Execution::Lookahead is primarily for determining what fields will need to be loaded for the query, so that they can be used to optimize database access.  However, it doesn't take into consideration whether or not fields are excluded by `@skip` or `@include` directives.

## Solution

I wasn't able to use `GraphQL::Schema::Directive.include?` directly, because it is given the parent object, which we won't know when looking ahead more than one level deep.  As a result, I added a `GraphQL::Schema::Directive.static_include?` method that directive classes can override so that `GraphQL::Execution::Lookahead` will use it to determine if the field will be included.

While looking at the code in lib/graphql/execution/lookahead.rb, I noticed that there was a reference to directives in a comment that precedes a conditional that is never taken.  `nil` is always passed in `ArgumentHelpers.arguments` for the `graphql_object` parameters.  I removed this unused parameters and cleaned up the dead code to avoid confusion.

### DIFF
--- a/lib/graphql/execution/lookahead.rb
+++ b/lib/graphql/execution/lookahead.rb
@@ -203,8 +203,22 @@ module GraphQL
         end
       end
 
+      def skipped_by_directive?(ast_selection)
+        ast_selection.directives.each do |directive|
+          dir_defn = @query.schema.directives.fetch(directive.name)
+          directive_class = dir_defn.type_class
+          if directive_class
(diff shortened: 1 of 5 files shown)


rmosolgo:
Thanks for the fix and the cleanup here!

PR #2705, opened 2020-01-28
Use Interface.orphan_types in schema construction

### LINKED ISSUE #2694: 1.10.0 Missing field for query returning interface, when orphan_types is called within the interface
I have a query that returns a type with an interface-typed field. All types of that interface `implements` that interface, and the interface declares the types through `orphan_types` on that interface itself. 

After upgrading to v1.10.0 from v1.9.18, when I tried to access the interface-typed field in my query, I got back an error saying the field doesn't exist. The schema dump doesn't include any types implementing the interface either.

I played around with it, and It seems this is only a problem when all of these three conditions are met:
1) using `GraphQL::Execution::Interpreter`
2) on v1.10.0
3) with `orphan_types` called from inside the interfaces instead of the schema.

Moving all the orphan_types to the schema works for now, but we prefer to have the `orphan_types` declared at each interface for better organization. I checked both the documentation and the changelogs, but didn't find any indication of the behavior of `orphan_types` changing.

I've read https://github.com/rmosolgo/graphql-ruby/issues/2674 as well as https://graphql-ruby.org/type_definitions/interfaces.html#orphan-types to make sure I'm not missing anything, but if I am, some kind of warning or verbose log would be really helpful.

rmosolgo:
Fixes #2694 

(cc @liang3404814 -- I saw this commit on your repo and I'm getting ready to cut 1.10.1, so I grabbed it!)

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1817,6 +1817,11 @@ module GraphQL
               add_type(t, owner: type, late_types: late_types)
             end
           end
+          if type.kind.interface?
+            type.orphan_types.each do |t|
+              add_type(t, owner: type, late_types: late_types)
+            end
+          end
           if type.kind.object?
             own_possible_types[type.graphql_name] = [type]
             type.interfaces.each do |i|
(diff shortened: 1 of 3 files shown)


liang3404814:
Thanks @rmosolgo, I got sidetracked by something before I was able to submit a PR. Glad it worked out! 🔥

PR #2711, opened 2020-01-31
Update InputObject undefined field error message

swalkinshaw:
InputObject's have input fields according to the spec and not arguments.

This updates the error message to use the term `Field` instead of `Argument`. This also makes the error message consistent with the legacy ones.

### DIFF
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -169,7 +169,7 @@ module GraphQL
           # Items in the input that are unexpected
           input.each do |name, value|
             if visible_arguments_map[name].nil?
-              result.add_problem("Argument is not defined on #{self.graphql_name}", [name])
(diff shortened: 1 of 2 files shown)


rmosolgo:
😖  I always found it annoying that fields have _arguments_ while input objects have _input fields_. You can see how the new DSL has only _arguments_. Did this message trip someone up?

swalkinshaw:
Long story... but we have a GraphQL client which I was updating to work on either 1.9 or 1.10 and we had a test which exposed this difference. I updated the test to work around it so it's not a big deal at all.

Though I do think these messages should reflect the actual spec terminology. In this context though it's slightly cloudy. I'll leave it up to you 🤷‍♂

rmosolgo:
Too bad graphql-cats didn't become a thing (at least, not for graphql-ruby :S)

PR #2713, opened 2020-02-01
Use Module#module_parent_name instead of Module#parent_name

tatsuyafw:
Hi,

Module#parent_name is deprecated in Rails 6.0 and will be removed in Rails 6.1.
Ref: https://github.com/rails/rails/pull/34051

To suppress a deprecation warning, use Module#module_parent_name instead.


rmosolgo:
Thanks for the update here!

Issue #2544, opened 2019-10-14
`loads:` doesn't work for field arguments


sayduck-daniel:
I would like to refactor this code:

```ruby
# types/customer_type.rb
module Types
  class CustomerType < BaseCustomObject
    field :groups, Types::GroupType.connection_type, null: false, method: :filter_groups do
      argument :parent_id, String, required: false
    end
  end
end

# models/customer.rb
class Customer
  def filter_groups(parent_id: nil)
    parent = if parent_id.present?
               SayduckSchema.object_from_id(parent_id, nil)
             else
               nil
             end

    groups.where(parent: parent)
  end
end
```

into 

```ruby
# types/customer_type.rb

module Types
  class CustomerType < BaseCustomObject
    field :groups, Types::GroupType.connection_type, null: false, method: :filter_groups do
      argument :parent_id, String, required: false, loads: Types::GroupType
    end
  end
end

# models/customer.rb
class Customer
  def filter_groups(parent: nil)
    groups.where(parent: parent)
  end
end
```

The documentation mentions this `loads:` attribute but it is not behaving correctly. 

With this sample code: 
```ruby
class Customer
  def filter_groups(parent: nil, parent_id: nil)
    pp "parent: #{parent}"
    pp "parent: #{parent.class}"
    pp "parent_id: #{parent_id}"
    groups.where(parent: parent)
  end
end
```

I get this in my log:
```log
Started POST "/graphql" for 127.0.0.1 at 2019-10-14 13:18:10 +0300
Processing by GraphqlController#execute as HTML
  Parameters: ...
  User Load (0.3ms)  ...
  ↳ app/controllers/concerns/authenticate.rb:27
  Customer Load (0.3ms)  ...
  ↳ app/graphql/sayduck_schema.rb:27
"parent: R3JvdXAtNGE1YjNlZTAtZDA5OS0wMTM3LWM4YWItMGM4NTkwOWQ4YmYw"
"parent: String"
"parent_id: "
  Group Load (0.6ms) ...
  ↳ app/graphql/sayduck_schema.rb:27
Completed 200 OK in 2249ms (Views: 0.2ms | ActiveRecord: 14.8ms)
```

Not how the `parent` is simply a string and how no queries for group have been made before the debug code. It should at least have made a find_by for the parent group.

Any idea what could be wrong here?

daemonsy:
Hi @sayduck-daniel do you have `.object_from_id` defined on `SayduckSchema`? 

[...]

PR #2720, opened 2020-02-04
loads: with plain field arguments

QWYNG:
Hi! Thank you for the great gem.
`loads:` with plain field arguments does not work now.
ref #2544

I think it is more convenient to add this.

I may have made the wrong fix or have some missing tests. If so please point.
best regards.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -648,6 +648,17 @@ module GraphQL
           # Apply any `prepare` methods. Not great code organization, can this go somewhere better?
           arguments.each do |name, arg_defn|
             ruby_kwargs_key = arg_defn.keyword
+
+            loads = arg_defn.loads
+            if ruby_kwargs.key?(ruby_kwargs_key) && loads && !arg_defn.from_resolver?
+              value = ruby_kwargs[ruby_kwargs_key]
+              ruby_kwargs[ruby_kwargs_key] = if arg_defn.type.list?
+                value.map { |val| load_application_object(arg_defn, loads, val, field_ctx.query.context) }
+              else
+                load_application_object(arg_defn, loads, value, field_ctx.query.context)
+              end
+            end
+
             if ruby_kwargs.key?(ruby_kwargs_key) && arg_defn.prepare
               ruby_kwargs[ruby_kwargs_key] = arg_defn.prepare_value(obj, ruby_kwargs[ruby_kwargs_key])
             end
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -24,21 +24,25 @@ module GraphQL
         self.class.arguments.each do |name, arg_defn|
           @arguments_by_keyword[arg_defn.keyword] = arg_defn
           ruby_kwargs_key = arg_defn.keyword
-          loads = arg_defn.loads
 
-          if @ruby_style_hash.key?(ruby_kwargs_key) && loads && !arg_defn.from_resolver? && !context.interpreter?
-            value = @ruby_style_hash[ruby_kwargs_key]
-            @ruby_style_hash[ruby_kwargs_key] = if arg_defn.type.list?
-              value.map { |val| load_application_object(arg_defn, loads, val, context) }
-            else
-              load_application_object(arg_defn, loads, value, context)
+          if @ruby_style_hash.key?(ruby_kwargs_key)
+            loads = arg_defn.loads
+            # Resolvers do this loading themselves;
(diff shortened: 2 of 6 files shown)


lib/graphql/schema/field.rb
```diff
@@ -642,12 +642,24 @@ def fetch_extra(extra_name, ctx)
       # @param field_ctx [GraphQL::Query::Context::FieldResolutionContext]
       # @return [Hash<Symbol => Any>]
       def to_ruby_args(obj, graphql_args, field_ctx)
+        @context = field_ctx.query.context
```
rmosolgo: Hmm, this field is a long-lived part of the schema -- it's created during initialization and stays around as long as the app is running. Also, this instance is shared between threads if the application is served by a multithreaded server. So I don't think caching query-specific values will work here.

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -8,6 +9,8 @@ class Interpreter
       #
       # @api private
       class Runtime
+        include GraphQL::Schema::Member::HasArguments
```
rmosolgo: `Member::HasArguments` is meant to contain behaviors that are shared between several parts of a GraphQL schema, so I don't think it's a good fit for the Runtime class. (For example, `HasArguments` implements the `argument(...)` configuration DSL!)

lib/graphql/schema/field.rb
rmosolgo: Honestly, I think you could just ignore this method. It's used by the legacy runtime, and in my opinion, we don't have to add features to that runtime. Alternatively, if supporting the legacy runtime is important for _you_, I won't stop you!

lib/graphql/execution/interpreter/runtime.rb
QWYNG: Thanks. I think I can redefine the `load_application_object` method with new arguments in another new module.
How do you think?

lib/graphql/schema/field.rb
QWYNG: I want to support as many runtimes as possible.
 I think if I create another module and move `load_application_object`, I can support this runtime.

lib/graphql/execution/interpreter/runtime.rb
QWYNG: srry We have GraphQL::Schema::Member::HasArguments::ArgumentObjectLoader
I should extend this!

QWYNG:
@rmosolgo 
Thank you for review.
The implementation was modified because the latest master was changed.
I added an argument to `load_application_object` to receive context

lib/graphql/schema/input_object.rb
```diff
@@ -5,6 +5,7 @@ class InputObject < GraphQL::Schema::Member
       extend GraphQL::Schema::Member::AcceptsDefinition
       extend Forwardable
       extend GraphQL::Schema::Member::HasArguments
+      extend GraphQL::Schema::Member::HasArguments::ArgumentObjectLoader
```
QWYNG: for InoutObject.coerce_input

spec/graphql/execution/errors_spec.rb
```diff
@@ -3,6 +3,14 @@
 
 describe "GraphQL::Execution::Errors" do
   class ErrorsTestSchema < GraphQL::Schema
+    def self.object_from_id(id, ctx)
+      if id == 1
+        :thing
+      else
+        raise ErrorD
+      end
+    end
```
rmosolgo: Is this addition still used, or should it be removed? (I don't quite see how it's used in the tests 🙈 )

QWYNG: @rmosolgo Thank you again!

spec/graphql/execution/errors_spec.rb:177

`describe "errors raised in input_object loads` need this.

Without this, `GraphQL::RequiredImplementationMissingError: ErrorsTestSchema.object_from_id(node_id, ctx) must be implemented to load by ID (tried to load from id `2`)`  will raise.

loads option always call `object_from_id`  now.

rmosolgo: Ahhh, thanks. I was confused because previously, `load_application_object` was called on the _instance_ of an InputObject type. But now, it's called on the _class_. So, to intercept the lookup, you have to define `def self.object_from_id` instead of `def object_from_id`. 

I pushed a change to update the test to reflect this change in behavior.

rmosolgo:
Thanks for your improvements here, this is awesome!

QWYNG:
@rmosolgo 
Thank you for the review! and Thank you for the useful gem!

PR #2722, opened 2020-02-04
Fix build_from_definition for built-in types

swalkinshaw:
If an SDL document contained a built-in type definition `build_from_definition` would raise an error due to type conflicts.

This defers merging in the `BUILT_IN_TYPES` and reverse merges them to keep any previously defined ones in the provided document.

### DIFF
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -42,7 +42,6 @@ module GraphQL
           end
           schema_definition = schema_defns.first
           types = {}
-          types.merge!(GraphQL::Schema::BUILT_IN_TYPES)
           directives = {}
           type_resolver = ->(type) { resolve_type(types, type)  }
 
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Looks like defining these later causes some issues with late bound types. I'll dig into it more

lib/graphql/schema/build_from_definition.rb
```diff
@@ -69,6 +68,14 @@ def build(document, default_resolve: DefaultResolve, using: {}, interpreter: tru
             end
           end
 
+          types = GraphQL::Schema::BUILT_IN_TYPES.merge(types) do |key, built_in_type, existing_type|
```
swalkinshaw: @rmosolgo had to add some merging logic in here to deal with late bound types. Otherwise when `orphan_types` was called a default scalar (which is late bound) would never be resolved. Not sure if you can think of a better solution.

rmosolgo: I didn't understand at first ... but I pulled the branch and gave it a try without the code. Now I see what you mean: if it's only _referenced_ in the given document, but not defined, then we can't ever resolve the late-bound type. I added an exploratory comment and rejiggered the code a bit so that it made more sense to simpleminded maintainer of this library 😰 .

PR #2729, opened 2020-02-06
Unify input coercion

rmosolgo:
I've added all the coercion behavior to type classes, but `runtime.rb` was still doing a lot of it itself. I've refactored it to use the type classes, instead.

### DIFF
--- a/lib/graphql/query/variables.rb
+++ b/lib/graphql/query/variables.rb
@@ -32,9 +32,11 @@ module GraphQL
             begin
               validation_result = variable_type.validate_input(provided_value, ctx)
               if validation_result.valid?
-                if value_was_provided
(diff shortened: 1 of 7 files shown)


lib/graphql/query/variables.rb
```diff
@@ -34,16 +34,22 @@ def initialize(ctx, ast_variables, provided_variables)
               if validation_result.valid?
                 if value_was_provided
                   # Add the variable if a value was provided
-                  memo[variable_name] = if provided_value.nil?
+                  memo[variable_name] = if ctx.query.interpreter?
+                    provided_value
```
rmosolgo: In order to unify these codepaths and behaviors, we don't coerce these at boot anymore. We leave them until the runtime encounters them.

PR #2739, opened 2020-02-12
update interpreter docs to point to new error handling

modosc:


rmosolgo:
Thanks for the update!

PR #2748, opened 2020-02-17
Subscriptions, prepare: and interpreter 

### LINKED ISSUE #2721: change in subscription behavior from old-style to class-based when using prepare:
not sure if this is a bug or not, but maybe it should be documented somewhere:

we're using opaque ids (ie, `object_from_id` / `id_from_object`). we had a subscription that looked like:
```
field :thingUpdated, Types::ThingType do 
  argument :id, ID, required: true, prepare: Resolvers::GlobalId.new
```
(`Resolvers::GlobalId` maps the opaque id to an integer)

when we upgraded to the class based subscriptions, things stopped working. it took me a while to track this down but it turns out there's a subtle change. previously, our graphql topic would look like:
```
someOpaqueId:thingUpdated:id:anotherOpaqueId
```

but now it looks like:
```
someOpaqueId:thingUpdated:id:integerId
```

removing the `prepare:` argument fixes this but also removes the validity check of the incoming id (which is what we had it for in the first place). 

i think the old behavior was probably incorrect (i'd expect the prepared version to be used instead) but here we are.

rmosolgo:
I'm not sure if I can actually _fix_ this or not, but I want to keep an eye on this behavior 

Closes #2721 

TODO: 

- [x] Is calling `.prepare` in subscriptions the right thing, or should I have used a different `default_value`? 
  - Yes -- when I changed the default value, it :boom:ed because the prepared value wasn't a member of the Enum.

### DIFF
--- a/lib/graphql/subscriptions.rb
+++ b/lib/graphql/subscriptions.rb
@@ -228,7 +228,10 @@ module GraphQL
         # match query arguments.
         arg_owner.arguments.each do |name, arg_defn|
           if arg_defn.default_value? && !normalized_args.key?(arg_defn.name)
-            normalized_args[arg_defn.name] = arg_defn.default_value
+            default_value = arg_defn.default_value
+            # We don't have an underlying "object" here, so it can't call methods.
+            # This is broken.
+            normalized_args[arg_defn.name] = arg_defn.prepare_value(nil, default_value, context: GraphQL::Query::NullContext)
           end
         end
 
--- a/lib/graphql/subscriptions/subscription_root.rb
+++ b/lib/graphql/subscriptions/subscription_root.rb
@@ -38,12 +38,13 @@ module GraphQL
           elsif (events = context.namespace(:subscriptions)[:events])
             # This is the first execution, so gather an Event
             # for the backend to register:
-            events << Subscriptions::Event.new(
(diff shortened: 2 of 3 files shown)


lib/graphql/subscriptions.rb
```diff
@@ -228,7 +228,10 @@ def normalize_arguments(event_name, arg_owner, args)
         # match query arguments.
         arg_owner.arguments.each do |name, arg_defn|
           if arg_defn.default_value? && !normalized_args.key?(arg_defn.name)
-            normalized_args[arg_defn.name] = arg_defn.default_value
+            default_value = arg_defn.default_value
+            # We don't have an underlying "object" here, so it can't call methods.
+            # This is broken.
+            normalized_args[arg_defn.name] = arg_defn.prepare_value(nil, default_value, context: GraphQL::Query::NullContext)
```
rmosolgo: I thought, "maybe instead of adding this, I should change the `default_value:` to the already-prepared value (that is, `"ONE"` to `"one"` in this code). But, that failed:

```
GraphQL::Schema::InvalidTypeError: Subscription is invalid: field "eventSubscription" argument "payloadType" default value "one" is not valid for type PayloadType (Can't resolve enum PayloadType for "one")
```

PR #2756, opened 2020-02-18
Sync client: wait for all chunks before resolving request.

sgwilym:
Recently we had a strange error when syncing our persisted queries:

`Failed to print sync result: SyntaxError: Unexpected end of JSON input`

After checking server logs, we found the server was returning well-formed JSON. So we thought possibly there was something up with the JS sync client, and there was: the request the sync client sends was resolved after a single chunk had been returned, rather than waiting for the stream to `end`.

Thanks to @markiz for finding the source of this issue.


sgwilym:
Just saw this was fixed by #2673, d'oh.

PR #2770, opened 2020-02-21
Remove `extend SubscriptionRoot` from subscriptions setup

rmosolgo:
There's a required `extend ...` call for using subscriptions with the interpreter that's surprising and easily overlooked. I'd like to make it work _without_ that call. 

TODO 

- [x] Make implementation not terrible 
- [x] update docs
- [x] understand the difference in behavior before-and-after

I think the big change here is how subscription fields are called in schemas from IDL. Now, you have to implement by hand the behavior that was previously default:

```ruby
->(obj, args, ctx) {
    if ctx.query.subscription_update?
      obj
    else
      ctx.skip
    end
  }
```

(Or implement it however you want.)


spec/graphql/subscriptions_spec.rb
```diff
@@ -119,12 +119,16 @@ class EventSubscription < GraphQL::Schema::Subscription
...
+    if !TESTING_INTERPRETER
       # Stub methods are required
       [:payload, :event, :my_event].each do |m|
-        define_method(m) { |*a| nil }
+        define_method(m) { |*a|
+          if context.query.subscription_update?
+            object
+          else
+            context.skip
+          end
```
rmosolgo: I think this was broken all along, but my assertion wasn't good enough to realize it 🤦‍♂

PR #2785, opened 2020-02-27
Add Query#sanitized_query_string

### LINKED ISSUE #1106: Implement Query#sanitized_query_string

### LINKED ISSUE #2333: Add a Printer to redact inlined argument values in GraphQL queries for logging
## Background 

Typically GraphQL uses `HTTP POST` for execution over the wire and POST bodies are not usually written to logs. Hence if we searched the logs, it must just consist of: 

```sh
method=POST path=/graphql format=json controller=GraphQLController  action=execute status=200

method=POST path=/graphql format=json controller=GraphQLController  action=execute status=200

method=POST path=/graphql format=json controller=GraphQLController  action=execute status=200
```

`<s>` This is super useful for debugging :)  `</s>`

### Logs are our friend 
Logging the `query` parameter can be really useful for a variety of reasons: 
- when GraphQL execution is not successful 
- Capturing field selections for deprecation of a field 
- Traces for NewRelic for slow queries

### Logging queries safely 
In web frameworks such as Rails, there are helpers to filter out sensitive parameters such as `password`, `token`. `Authorization`. 

In a GraphQL raw query , if the author did not used **inline variables**, we can already deem the query as safe as it does not contain any data. `variables` is usually a separate parameter is a `Hash`, so it's straightforward for Devs/DevOps to do filtering there. 

### The Problem and Solution
Inline variables are used frequently, especially when someone is prototyping a query. 

Hence I propose to add a `FilteredArgumentsPrinter` class that allows filtering all arguments and filtering by name matching. 

### A prototype

```graphql
query($string_variable: String!, $input: MyInput!) {
  queryEntry(stringVar: $string_variable, input_object: $input) {
    objectType(moreInlineVar: "my_var", string_var: $string_variable) {
      name 
      booleanDirectiveArgument @skip(if: true)
      customDirectiveArgument @customdirective(inline: "arg", stringVar: $string_variable)
    }
  }
}

```

```ruby
puts GraphQL::Language::FilteredPrinter.new.print(GraphQL.parse(query))
```

```graphql
query($string_variable: String!, $input: MyInput!) {
  queryEntry(stringVar: $string_variable, input_object: $input) {
    objectType(moreInlineVar: "<FILTERED>", string_var: $string_variable) {
      name
      booleanDirectiveArgument @skip(if: "<FILTERED>")
      customDirectiveArgument @customdirective(inline: "<FILTERED>", stringVar: $string_variable)
    }
  }
}
```

### The Draft PR 
Prototype PR at #2334 

### Other concerns / Todos
- A filtered query is no longer typed correctly, for example when we replace boolean values with `<FILTERED>`
- Should consider allowing the inlining of declared variables such as `$id`

rmosolgo:
The goal here is to add something that's safe for logging. All strings, except for IDs, are replaced with `"<REDACTED>"`. 

Variables are inlined and redacted the same way.

- Fixes #2334 
- Fixes #1106  
- Fixes #2333

### DIFF
--- a/lib/graphql/language.rb
+++ b/lib/graphql/language.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require "graphql/language/block_string"
 require "graphql/language/printer"
+require "graphql/language/sanitized_printer"
 require "graphql/language/document_from_schema_definition"
 require "graphql/language/generation"
 require "graphql/language/lexer"
--- a/lib/graphql/language/printer.rb
+++ b/lib/graphql/language/printer.rb
@@ -93,7 +93,7 @@ module GraphQL
       end
 
       def print_input_object(input_object)
-        "{#{input_object.arguments.map { |a| "#{a.name}: #{print(a.value)}" }.join(", ")}}"
+        "{#{input_object.arguments.map { |a| print_argument(a) }.join(", ")}}"
       end
 
       def print_list_type(list_type)
--- a/lib/graphql/language/sanitized_printer.rb
+++ b/lib/graphql/language/sanitized_printer.rb
@@ -1,0 +1,189 @@
+# frozen_string_literal: true
+module GraphQL
+  module Language
+    # A custom printer used to print sanitized queries. It inlines provided variables
+    # within the query for facilitate logging and analysis of queries.
+    #
+    # The printer returns `nil` if the query is invalid.
+    #
+    # Since the GraphQL Ruby AST for a GraphQL query doesnt contain any reference
+    # on the type of fields or arguments, we have to track the current object, field
+    # and input type while printing the query.
+    #
+    # @example Printing a scrubbed string
+    #   printer = QueryPrinter.new(query)
+    #   puts printer.sanitized_query_string
+    #
+    # @see {Query#sanitized_query_string}
+    class SanitizedPrinter < GraphQL::Language::Printer
+
+      REDACTED = "\"<REDACTED>\""
+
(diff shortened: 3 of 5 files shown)


rmosolgo:
Weird. The failing build keeps failing. The passing build passed _again_ when I re-ran it. It passes locally. I'm going for it.

PR #2790, opened 2020-02-29
Try consolidating shared resolution code

rmosolgo:
I'd love to DRY this code -- but it's very weird about the required changes to `subscription_spec` -- I gotta figure out what's up with that.


rmosolgo:
This change has similar effects to build-from-definition as #2770 , so I'm going to bring these changes in there

PR #2791, opened 2020-02-29
Interface type visibility (on master)

rmosolgo:
A branch from #2678 , trying to get it working still


lib/graphql/object_type.rb
```diff
@@ -71,24 +67,37 @@ def all_fields
...
       nil
     end
 
+    def interface_type_memberships=(interface_type_memberships)
+      @interface_type_memberships = interface_type_memberships
+    end
+
     protected
 
-    attr_reader :dirty_interfaces, :dirty_inherited_interfaces
```
rmosolgo: TODO: can the underlying ivars here be removed?

lib/graphql/schema/object.rb
```diff
@@ -69,47 +71,70 @@ def initialize(object, context)
...
               end
 
+              new_memberships << int.type_membership_class.new(int, self, options)
+
               # Include the methods here,
               # `.fields` will use the inheritance chain
               # to find inherited fields
               include(int)
+            elsif int.is_a?(String) || int.is_a?(GraphQL::Schema::LateBoundType)
+              new_memberships << int
```
rmosolgo: TODO options are silently ignored here.

lib/graphql/schema/object.rb
```diff
@@ -69,47 +71,70 @@ def initialize(object, context)
...
 
-        def interfaces
-          inherited_interfaces = (superclass.respond_to?(:interfaces) ? superclass.interfaces : nil)
-          if inherited_interfaces && !inherited_interfaces.empty?
-            own_interfaces + inherited_interfaces
-          else
-            own_interfaces
-          end
+        def interface_type_memberships
+          @interface_type_memberships ||= []
```
rmosolgo: TODO: rename with `own_`, since it doesn't include inherited ones.

lib/graphql/schema/possible_types.rb
```diff
@@ -27,13 +24,23 @@ def possible_types(type_defn, ctx)
...
           raise "Unexpected possible_types object: #{type_defn.inspect}"
         end
       end
+
+      def interface_implementers(ctx, type_defn)
+        @interface_implementers[ctx] ||= begin
+          Hash.new do |hash, key|
+            hash[key] = @object_types.select { |type| type.interfaces(ctx).include?(key) }.sort_by(&:name)
+          end
+        end
```
rmosolgo: TODO: move this block back up to the initializer

lib/graphql/schema/warden.rb
```diff
@@ -208,7 +249,7 @@ def visible_abstract_type?(type_defn)
       end
 
       def visible_possible_types?(type_defn)
-        possible_types(type_defn).any? { |t| visible_type?(t) }
+        possible_types(type_defn).any? { |t| [t.graphql_name, "visible_type?", visible_type?(t)]; visible_type?(t) }
```
rmosolgo: TODO: remove this debug info

lib/graphql/schema/object.rb
rmosolgo: Maybe this is OK for now: the `type_membership_class` comes from the _interface_, and we don't have one yet here. 

This _could_ be implemented. Eg, add a `PendingTypeMembership` class which holds the pending type and the options, and then when the type is updated, add the options into the new TypeMembership class.

lib/graphql/object_type.rb
```diff
@@ -97,25 +106,37 @@ def normalize_interfaces(ifaces)
...
+        @clean_inherited_fields
+      end
+    end
+
+    def load_interfaces(ctx = GraphQL::Query::NullContext)
+      ensure_defined
+      clean_ifaces = []
+      clean_inherited_ifaces = []
+      @interface_type_memberships.each do |type_membership|
+        if ctx.nil? || type_membership.visible?(ctx)
```
grcooper: `ctx` defaults to `GraphQL::Query::NullContext` so I believe `ctx.nil?` isn't the correct option here?

lib/graphql/object_type.rb
```diff
@@ -97,25 +106,37 @@ def normalize_interfaces(ifaces)
...
+          clean_ifaces << GraphQL::BaseType.resolve_related_type(type_membership.abstract_type)
+        end
       end
+
+      @inherited_interface_type_memberships.each do |type_membership|
+        if ctx.nil? || type_membership.visible?(ctx)
```
grcooper: Some question above about `ctx.nil?`

lib/graphql/schema/object.rb
grcooper: Is the issue that a `type_membership_class` may not exist on the interface yet? I would assume the interface would have to be loaded and have its membership class loaded in order to use it?

lib/graphql/object_type.rb
rmosolgo: I need _some_ behavior to return all interfaces, without filtering. The library does that in a few places. I've chosen `interfaces(nil)`, where passing `nil` as context is the signal to return _everything_. In that case, `ctx` is `nil`. 

I'm not sold on this approach; another approach (which is taken in `Schema::Object`) is to expose `interface_type_memberships`, and let callers who need unfiltered access use _that_ instead. 

But anyways, that's why `nil` is handled explicitly here.

lib/graphql/schema/object.rb
rmosolgo: Yup

lib/graphql/object_type.rb
grcooper: 👍 Makes sense that we need that case.

And also makes sense to make it a more explicit flag, however when I first was building this I was thinking that a NullContext was going to be that indicator if we needed to return everything, but perhaps NullContext is too widely used to be explicty return all?

Is there a case where we want to pass in a sample context of `{ return_all: true }` to be even more explicit?

rmosolgo: > a NullContext was going to be that indicator

That's a good idea. I just refactored it to remove the `.interfaces(nil)` use case, but I think you're right about using `NullContext` for it, and supporting `SomeObject.interfaces` returning _everything_.

The thing is, even if we fall back to `NullContext`, we have to hardcode a check for it, because passing `NullContext` _won't_ make `.visible?` checks pass. 

Anyhow, thanks for sharing your original intention here. Let me take another look.

lib/graphql/schema/object.rb
```diff
@@ -69,64 +71,77 @@ def initialize(object, context)
...
               # `.fields` will use the inheritance chain
               # to find inherited fields
               include(int)
+            elsif int.is_a?(String) || int.is_a?(GraphQL::Schema::LateBoundType)
+              if options.any?
+                raise ArgumentError, "`implements(...)` doesn't support options with late-loaded types yet. Remove #{options} and open an issue to request this feature."
+              end
+              new_memberships << int
+            else
+              raise ArgumentError, "Unexpected interface definition (expected module): #{int} (#{int.class})"
```
rmosolgo, line 94: TODO This is a breaking change -- can I still accept legacy interfaces here? Otherwise I'll slate this for 1.11

lib/graphql/schema/object.rb
```diff
@@ -69,64 +71,77 @@ def initialize(object, context)
...
-              int_f = {}
-              int.fields.each do |name, legacy_field|
-                int_f[name] = field_class.from_options(name, field: legacy_field)
+        # param context [Query::Context] If omitted, skip filtering.
+        def interfaces(context = GraphQL::Query::NullContext)
+          visible_interfaces = []
+          unfiltered = context == GraphQL::Query::NullContext
+          own_interface_type_memberships.each do |type_membership|
+            # During initialization, `type_memberships` can hold late-bound types
+            case type_membership
```
rmosolgo, line 127: TODO add an `else` case to handle unexpected values

lib/graphql/schema/object.rb
rmosolgo: Fixed -- added a case for legacy interfaces.

lib/graphql/schema/object.rb
rmosolgo: Fixed, added an `else` clause.

PR #2792, opened 2020-03-02
Use Argument names as configured (fix args with numbers)

### LINKED ISSUE #1839: Input object parsing with number after underscore
I have an `address_1` field in my Postgres DB schema for my address table. My GraphQL mutation has `address1` field and the input type definition on the BE has `argument :address_1, String, 'Address', required: true`.

No error is raised but in the resolve method when I access the param I have it as `address1` which is of course not ideal. Is this the intended behavior? If so, why?

### LINKED ISSUE #2611: Deserialization of arguments with numbers appears not to work

### LINKED ISSUE #2706: Is it possible to keep arguments values as original, don't snake-case always?
Hi, 

is it possible to keep the style of the graphql arguments when they're passed to the resolver?

Thanks.

rmosolgo:
Currently, GraphQL-Ruby _doesn't_ use the provided symbol to build keyword arguments, which is confusing. 

So instead, if `argument(...)` is configured with a symbol, use the given symbol for the method's keyword arguments. 

This can break existing configurations if they configured arguments with camelized symbols, but expected underscored arguments in their methods (see the change in the test suite, for example).

Fixes #2611 
Fixes #2665 (this approach maintains current behavior)
Fixes #2706
Fixes #1839

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -55,7 +55,7 @@ module GraphQL
         @owner = owner
         @as = as
         @loads = loads
-        @keyword = as || Schema::Member::BuildType.underscore(@name).to_sym
+        @keyword = as || (arg_name.is_a?(Symbol) ? arg_name : Schema::Member::BuildType.underscore(@name).to_sym)
         @prepare = prepare
         @ast_node = ast_node
         @from_resolver = from_resolver
--- a/spec/graphql/schema/input_object_spec.rb
+++ b/spec/graphql/schema/input_object_spec.rb
@@ -61,6 +61,27 @@ describe GraphQL::Schema::InputObject do
     end
   end
 
+  describe "camelizing with numbers" do
+    module InputObjectWithNumbers
+      class InputObject < GraphQL::Schema::InputObject
+        argument :number_arg1, Integer, required: false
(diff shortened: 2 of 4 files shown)


TomasBarry:
Thanks for tackling this @rmosolgo.

PR #2797, opened 2020-03-03
add entry about custom scalar changes

modosc:
per slack conversation


rmosolgo:
Thanks for the update -- sorry i missed it before!

PR #2799, opened 2020-03-03
copy GraphQL::Pagination::Conection#first/#last logic to Relay::Array…

modosc:
…Connection, Relay::RelationConnection, and Relay::BaseConnection


modosc:
@rmosolgo i think this is good to go?

modosc:
actually - it looks like this isn't necessary for us after all, feel free to close out if it's not useful.

rmosolgo:
Thanks for the fix, anyways!

PR #2803, opened 2020-03-04
[JavaScript Client] Fix module not found TypeScript Error

### LINKED ISSUE #2802: [JavaScript Client] Module not found - TypeScript compiler error
## Description

I installed `graphql-ruby-client` and successfully set it up in my Apollo configs using ActionCable. These are the imports in my `apollo.ts` file:

```ts
import ActionCable from "actioncable";
import { ActionCableLink } from "graphql-ruby-client";
```

Everything is working, but TypeScript is reporting module not found errors because I am using only the `ActionCableLink` (and `@types/actioncable`) and haven't installed the other optional dependencies in the package:

Screen Shot 2020-03-04 at 9 54 39 AM

I fixed this by installing `@types/ably` and `@types/pusher` in my repository.

## Root Cause

I think this happened when the [exports were moved to the top level index.ts](https://github.com/rmosolgo/graphql-ruby/commit/e7e114cb485740c2ceafe9d73eafece404eda63f) instead of selectively-importing them, and therefore TypeScript is reading all of the type declarations now.

## Suggested Fix

**I made a pull request: #2803**

I think the fix is actually really easy.  The following need to be moved to `dependencies` rather than `devDependencies`:

```
"@types/ably": "^1.0.0",
"@types/actioncable": "^5.2.3",
"@types/pusher-js": "^4.2.2",
```

This is because these dependencies appear in the compiled `.d.ts` files (as explained in this [stack overflow response](https://stackoverflow.com/a/46011417/5920970)).

For example:

`AblyLink.d.ts`

```ts
import { ApolloLink, Observable, FetchResult, NextLink, Operation } from "apollo-link";
import { Realtime } from "ably";

/* ... /*
```

## (Pipe dream) Other suggestion

I think a stretch goal might be to make use of a handful of tiny [scoped packages](https://docs.npmjs.com/creating-and-publishing-scoped-public-packages) instead of a single package with an index barrel file that includes code we might not need.

For example, there could be `@graphql-ruby/core`, and then, selectively, whichever implementation you needed:

- `@graphql-ruby/actioncable`
- `@graphql-ruby/pusher`
- `@graphql-ruby/ably`

but this obviously requires a bit of reworking of things, so this is just a "for your future consideration" thing 🙂

hueter:
Fixes #2802

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -9,13 +9,10 @@
   "license": "LGPL-3.0",
   "bin": "./dist/cli.js",
   "devDependencies": {
-    "@types/ably": "^1.0.0",
-    "@types/actioncable": "^5.2.3",
     "@types/glob": "^7.1.1",
     "@types/jest": "^25.1.2",
     "@types/minimist": "^1.2.0",
     "@types/node": "^13.7.1",
-    "@types/pusher-js": "^4.2.2",
     "jest": "^25.0.0",
     "nock": "^11.0.0",
     "ts-jest": "^25.2.0"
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the fix here!

rmosolgo:
🚢 In 1.7.5!

Issue #2724, opened 2020-02-05
Schema.to_definition generates multiple entries for Node under certain conditions


ajkerr:
We are attempting to upgrade our application to 1.10.2, and ran into an issue when we dump our schema to IDL and JSON.

Our schema is causing the `Node` interface to be generated twice when we dump it with `Schema.to_definition` or `to_json`:

```ruby
"""
An object with an ID.
"""
interface Node {
  """
  ID of the object.
  """
  id: ID!
}

"""
An object with an ID.
"""
interface Node {
  """
  ID of the object.
  """
  id: ID!
}
```

I was able to reproduce this with a vanilla Rails project, which I've published at https://github.com/ajkerr/reproduce_duplicate_node.

The key things that I've added which mirror what we have in our actual schema are:
```ruby
add_field(GraphQL::Types::Relay::NodeField)
```
in `Types::QueryType`, and:
```ruby
implements GraphQL::Relay::Node.interface
```
in `Types::PostType`.

We have this set up as documented in https://graphql-ruby.org/relay/object_identification.html. Our actual schema is obviously more complete and defines the `id_from_object`, etc. methods.

To see the issue, open a rails console, and run `Aktest::Schema.to_definition`. You should see two copies of the `Node` interface in the output. The breaks our Apollo client when it attempts to consume the generated json schema.

rmosolgo:
Yikes! It's probably confused by the old-style and new-style node interfaces. I'm surprised it doesn't error, though. 

As a workaround, you could change your `implements` calls to: 

```ruby
implements GraphQL::Types::Relay::Node
```

Does that work for you? If so, I'll update those docs!

ajkerr:
@rmosolgo Your workaround solution seems to solve the problem, thanks!

modosc:
just spent a few hours tracking this down locally - same thing happened to us, same workaround solved it. would be great to add this to the upgrade / relay docs.

rmosolgo:
It was trickier than I thought to catch this programatically during schema boot 😅 Hopefully people who run into it will find this issue.

PR #2806, opened 2020-03-05
Correct the interface required to implement Node.

marta-silva:
This fixes the documentation to avoid https://github.com/rmosolgo/graphql-ruby/issues/2724


rmosolgo:
Thanks for the update!

PR #2807, opened 2020-03-05
Add field complexity hooks in QueryComplexity

swalkinshaw:
The AST `QueryComplexity` analyzer is designed to return a single `max_possible_complexity` value in its `result` right now. Unfortunately there's currently no way to capture individual field level complexity details.

This adds a `field_complexity` method which acts as a hook. It gets called once for every field with its calculated complexity values. The default hook implementation is a no-op and can be overridden for custom behaviour.

Example:

```ruby
class CustomComplexityAnalyzer < GraphQL::Analysis::AST::QueryComplexity
  private

  def field_complexity(scoped_type_complexity, max_complexity:, child_complexity:)
    Rails.logger.info("Field complexity: #{scoped_type_complexity.response_path} => #{max_complexity}")
  end
end
```

### DIFF
--- a/lib/graphql/analysis/ast/query_complexity.rb
+++ b/lib/graphql/analysis/ast/query_complexity.rb
@@ -21,13 +21,17 @@ module GraphQL
           # since the lexical binding isn't important.
           HASH_CHILDREN = ->(h, k) { h[k] = {} }
 
+          attr_reader :field_definition, :response_path, :query
+
           # @param node [Language::Nodes::Field] The AST node; used for providing argument values when necessary
           # @param field_definition [GraphQL::Field, GraphQL::Schema::Field] Used for getting the `.complexity` configuration
-          # @param query [GrpahQL::Query] Used for `query.possible_types`
-          def initialize(node, field_definition, query)
+          # @param query [GraphQL::Query] Used for `query.possible_types`
+          # @param response_path [Array<String>] The path to the response key for the field
+          def initialize(node, field_definition, query, response_path)
+            @node = node
             @field_definition = field_definition
             @query = query
-            @node = node
+            @response_path = response_path
             @scoped_children = nil
           end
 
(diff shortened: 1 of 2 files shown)


lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -21,13 +21,16 @@ class ScopedTypeComplexity
...
           # @param node [Language::Nodes::Field] The AST node; used for providing argument values when necessary
           # @param field_definition [GraphQL::Field, GraphQL::Schema::Field] Used for getting the `.complexity` configuration
           # @param query [GrpahQL::Query] Used for `query.possible_types`
-          def initialize(node, field_definition, query)
+          def initialize(node, field_definition, query, response_path)
+            @node = node
             @field_definition = field_definition
             @query = query
-            @node = node
+            @response_path = response_path
```
swalkinshaw: This is the only thing specific to our use case but I can't find a better way to generalize it

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -21,13 +21,17 @@ class ScopedTypeComplexity
           # since the lexical binding isn't important.
           HASH_CHILDREN = ->(h, k) { h[k] = {} }
 
+          attr_reader :field_definition, :response_path, :query
+
           # @param node [Language::Nodes::Field] The AST node; used for providing argument values when necessary
           # @param field_definition [GraphQL::Field, GraphQL::Schema::Field] Used for getting the `.complexity` configuration
-          # @param query [GrpahQL::Query] Used for `query.possible_types`
```
swalkinshaw: 👀 😅

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -90,36 +94,21 @@ def on_leave_field(node, parent, visitor)
         end
 
         # @return [Integer]
-        def max_possible_complexity
+        def max_possible_complexity(complexity_merger = ComplexityMerger.new)
```
swalkinshaw: Now that I see this, it should probably be a kwarg?

swalkinshaw:
Yeah that's definitely a good question; it may not be worth preserving this existing design decision so I'll give it another look 👀

swalkinshaw:
Updated this and it's much simpler now 🎉 thanks for the suggestion

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -89,132 +93,142 @@ def on_leave_field(node, parent, visitor)
...
             end
 
-            # Calculate the child complexity by summing the complexity of all selections
-            complexity_for_keys.each_value.inject(0, &:+)
+            next unless scoped_children_for_key
+
+            child_complexity = merged_max_complexity_for_scopes(query, scoped_children_for_key)
+            # This is the _last_ one we visited; assume it's representative.
+            max_complexity = complexity_for_key.own_complexity(child_complexity)
+            defined_complexity = complexity_for_key.own_complexity(0)
```
rmosolgo: It seems like this is evaluated in order to be passed to `field_complexity(...)`, is that right?

Since we don't need it, and we already pass the required inputs to `field_complexity(...)` how about we don't evaluate it here, and instead, let user-defined `field_complexity` hooks generate that value if they need it, eg 

```ruby 
def field_complexity(complexity_for_key, max_complexity:, child_complexity:)
  defined_complexity = complexity_for_key.own_complexity(0)
  # ... 
end 
```

That way, we don't do any throw-away work in the default case (and I don't get any tickets about how the complexity proc is unexpectedly called twice 😅 ). Would that work here?

swalkinshaw: 🎉 yes that was specific to our needs anyway so I'll remove it by default

rmosolgo:
Thanks for upstreaming this improvement!

PR #2811, opened 2020-03-06
Use Query#arguments_for in Lookahead#arguments

rmosolgo:
A chance to dry up a lot of code here. 

Fixes #2809 by removing the problematic call altogether.

### DIFF
--- a/lib/graphql/execution/lookahead.rb
+++ b/lib/graphql/execution/lookahead.rb
@@ -54,7 +54,7 @@ module GraphQL
         if defined?(@arguments)
           @arguments
         else
-          @arguments =if @field
+          @arguments = if @field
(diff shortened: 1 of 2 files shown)


lib/graphql/execution/lookahead.rb
```diff
@@ -51,7 +51,15 @@ def initialize(query:, ast_nodes:, field: nil, root_type: nil, owner_type: nil)
 
       # @return [Hash<Symbol, Object>]
       def arguments
-        @arguments ||= @field && ArgumentHelpers.arguments(@query, @field, ast_nodes.first)
+        if defined?(@arguments)
+          @arguments
+        else
+          @arguments =if @field
```
rmosolgo:
```suggestion
          @arguments = if @field
```

PR #2812, opened 2020-03-09
Remove ruby2_keywords method

rickpr:
This method is specifically undefined in Ruby 2.7, causing a `NoMethodError`.

The method is also called on a symbol and does nothing but return `nil`.

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -7,14 +7,6 @@ require "forwardable"
 require_relative "./graphql/railtie" if defined? Rails::Railtie
 
 module GraphQL
-  # forwards-compat for argument handling
-  module Ruby2Keywords
-    if RUBY_VERSION < "2.7"
-      def ruby2_keywords(*)
-      end
-    end
-  end
-
   class Error < StandardError
   end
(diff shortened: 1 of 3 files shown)


rmosolgo:
Hi! I added this to avoid the deprecation warnings when GraphQL-Ruby is run on Ruby 2.7. Without this method, I see those deprecations in CI again, for example: 

https://travis-ci.org/rmosolgo/graphql-ruby/jobs/659993734#L432

Could you help me understand why you think these should be removed? Did you encounter a problem with them?

rickpr:
Yes, the reason is because it doesn't work at all with Ruby 2.7:
image

rmosolgo:
It looks like you're running `2.7.0.preview1`. What about updating to a stable version, like `2.7.0`? (Maybe `ruby2_keywords` wasn't added in `.preview1`!)

rickpr:
Ah ha! Thank you for your help, that was my problem. So `ruby2_keywords` is only defined for Ruby 2.7, but not for previous versions, which is the reason for declaring it.

PR #2814, opened 2020-03-09
Don't allow mixing legacy and class-based types in the same Schema

### LINKED ISSUE #2724: Schema.to_definition generates multiple entries for Node under certain conditions

### LINKED ISSUE #2784: Type mismatch on variable $pageSize and argument pageSize (Int / Int)
With the latest version , when I run queries like
```
    query($contentAreaSlug: String!,
          $contentTypes: [ContentTypes!],
          $page: Int,
          $pageSize: Int,
          $sortBy: String,
          $direction: SortDirection,
          $publishStatuses: [PublishStatuses!],
          $internalStatuses: [InternalStatuses],
          $externalTypes: [ExternalTypes],
          $keyword: String,
          $publishDateStart: ISO8601DateTime,
          $publishDateEnd: ISO8601DateTime,
          $contributorRoleSlug: String,
          $contributorProfileIds: [String]) {
            contentList(contentAreaSlug: $contentAreaSlug,
                        contentTypes: $contentTypes,
                        page: $page,
                        pageSize: $pageSize,
                        sortBy: $sortBy,
                        direction: $direction,
                        publishStatuses: $publishStatuses,
                        internalStatuses: $internalStatuses,
                        externalTypes: $externalTypes,
                        keyword: $keyword,
                        publishDateStart: $publishDateStart,
                        publishDateEnd: $publishDateEnd,
                        contributorRoleSlug: $contributorRoleSlug,
                        contributorProfileIds: $contributorProfileIds) {
                          totalPages
                          currentPage
                          totalItems
                          previousPage
                          nextPage
                          items {
                            title
                            canonicalSlug
                            contentAreaSlug
                            resourceType
                            publishDate
                            updatedAt
                            createdAt
                            internalStatus
                            publishStatus
                            userId
                            externalType
                            primaryVisuals {
                              thumbnail {
                                slug
                                preferredAspectRatio {
                                  instances {
                                    url
                                  }
                                }
                                aspectRatios {
                                  square {
                                    slug
                                    instances {
                                      url
                                    }
                                  }
                                }
                              }
                            }
                          }
                        }
            }
```

I get errors like

```
#<StandardError: [{"message"=>"Type mismatch on variable $page and argument page (Int / Int)", "locations"=>[{"line"=>17, "column"=>25}], "path"=>["query GraphQL__Client__OperationDefinition_70213633881320", "contentList", "page"], "extensions"=>{"code"=>"variableMismatch", "variableName"=>"page", "typeName"=>"Int", "argumentName"=>"page", "errorMessage"=>"Type mismatch"}}, {"message"=>"Type mismatch on variable $pageSize and argument pageSize (Int / Int)", "locations"=>[{"line"=>18, "column"=>25}], "path"=>["query GraphQL__Client__OperationDefinition_70213633881320", "contentList", "pageSize"], "extensions"=>{"code"=>"variableMismatch", "variableName"=>"pageSize", "typeName"=>"Int", "argumentName"=>"pageSize", "errorMessage"=>"Type mismatch"}}]>
```

Not sure where if I've goofed up my types or if this is a bug.

rmosolgo:
It's interestingly possible to have _both_ `T` and `T.graphql_definition` in your schema, and this happens with built-in types like `Node` and `Int`. 

We can detect this at schema build time and prevent it. 

This might break some schemas, because a schema with this inconsistency _would_ work -- but the recommended fix is not too hard. 

Fixes #2724 
Fixes #2784

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -84,7 +84,13 @@ module GraphQL
 
     class DuplicateTypeNamesError < GraphQL::Error
       def initialize(type_name:, first_definition:, second_definition:, path:)
-        super("Multiple definitions for `#{type_name}`. Previously found #{first_definition.inspect} (#{first_definition.class}), then found #{second_definition.inspect} (#{second_definition.class}) at #{path.join(".")}")
+        message = "Multiple definitions for `#{type_name}`. Previously found #{first_definition.inspect} (#{first_definition.class}), then found #{second_definition.inspect} (#{second_definition.class}) at #{path.join("")}"
+        if (first_definition.is_a?(Module) && !second_definition.is_a?(Module)) || (second_definition.is_a?(Module) && !first_definition.is_a?(Module))
+          message += "\n\nThese two GraphQL types have the _same name_, but one is a Class and the other isn't. Check for legacy type definitions in the schema, and try updating them."
+        else
+          message += "\n\nThese two types have the same name, but they're different Ruby objects. Try renaming one of them (by giving it a new class name, or with `graphql_name(\"...\")`), or remove one of the definitions."
+        end
+        super(message)
       end
     end
 
@@ -1794,6 +1800,7 @@ module GraphQL
       end
 
       def add_type(type, owner:, late_types:, path:)
+        orig_type = type
         if type.respond_to?(:metadata) && type.metadata.is_a?(Hash)
           type_class = type.metadata[:type_class]
           if type_class.nil?
@@ -1811,12 +1818,13 @@ module GraphQL
           um << owner
         end
 
+        prev_type = own_types[type.graphql_name]
         if (prev_type = own_types[type.graphql_name])
-          if prev_type != type
+          if prev_type != orig_type
             raise DuplicateTypeNamesError.new(
               type_name: type.graphql_name,
               first_definition: prev_type,
-              second_definition: type,
+              second_definition: orig_type,
               path: path,
             )
           else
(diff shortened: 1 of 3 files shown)


rmosolgo:
Interesting. As-written, it prevents _all_ legacy types from being added to the schema 😖

rmosolgo:
This issue hasn't come up again, and it's trickier than I thought to avoid, so I'm going to let this go for now.

PR #2824, opened 2020-03-13
Fix default_value: {} on Ruby 2.7

casperisfine:
Follow up: https://github.com/rmosolgo/graphql-ruby/pull/2704

This is another subtle change in 2.7 handling of keyword arguments.

It wouldn't be a problem if the target method was set as `ruby2_keywords`, but in this case it isn't.

I wish I could produce a test case for this, but my understanding of the library is too limited. 

I figured that one out, because lots of our tests were failing, and all Argument with `default_value = {}` ended up with `default_value = nil`.

### DIFF
--- a/lib/graphql/define/instance_definable.rb
+++ b/lib/graphql/define/instance_definable.rb
@@ -155,7 +155,7 @@ module GraphQL
             # Apply definition from `define(...)` kwargs
             defn.define_keywords.each do |keyword, value|
               # Don't splat string hashes, which blows up on Rubies before 2.7
-              if value.is_a?(Hash) && value.each_key.all? { |k| k.is_a?(Symbol) }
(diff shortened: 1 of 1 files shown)


rmosolgo:
Derp, thanks for the fix. It gets me every time: 

```ruby
 $ irb
irb(main):001:0> {}.all? { |k, v| k.is_a?(Symbol) }
=> true
```

It looks like the Travis config is out-of-date on the 1.9.x branch -- `ruby-head` is supposed to be an allowed failure.

casperisfine:
@rmosolgo sorry to bother you, but could this be released as 1.9.20?

rmosolgo:
Certainly! 💁‍♂️ https://rubygems.org/gems/graphql/versions/1.9.20

casperisfine:
<3

PR #2826, opened 2020-03-14
Fix so Rake task loads environment by default

scomma:
I am using this gem under rake 13.0.1 and rails 6.0.2.1. The graphql:schema:dump rake task doesn't work because it doesn't load the Rails environment so it can't find the schema to dump. To work around this I need to add `dependencies` parameter in graphql.rake:

```
require 'graphql/rake_task'
GraphQL::RakeTask.new(dependencies: :environment, schema_name: 'ApplicationSchema')
```

when it should instead be automatic. I suspect the amended line of code has stopped working and I propose this fix.

### DIFF
--- a/lib/graphql/rake_task.rb
+++ b/lib/graphql/rake_task.rb
@@ -76,7 +76,7 @@ module GraphQL
     # Set the parameters of this task by passing keyword arguments
     # or assigning attributes inside the block
     def initialize(options = {})
-      default_dependencies = if Rake::Task.task_defined?("environment")
+      default_dependencies = if Rake::Task.task_defined?(:environment)
         [:environment]
       else
         []


rmosolgo:
Were you able to confirm that this change fixed it for you? I don't _see_ how it would make a difference. 

`.task_defined?` calls `.lookup(...)`: https://github.com/ruby/rake/blob/4fe73ff6b19ea1d8490b0442c75fc3a53815c4cf/lib/rake/task.rb#L408-L411

`.lookup(...)` converts its input to a string (https://github.com/ruby/rake/blob/4fe73ff6b19ea1d8490b0442c75fc3a53815c4cf/lib/rake/task_manager.rb#L187-L205), so I wouldn't expect symbol vs. string to matter here. 

But were you able to confirm that this change fixed it for you?

louim:
It seems both version of that check don't work:
```
bin/rake graphql:schema:idl                                                                                                     
Running via Spring preloader in process 55441

[75, 84] in /Users/louim/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/graphql-1.10.5/lib/graphql/rake_task.rb
   75:
   76:     # Set the parameters of this task by passing keyword arguments
   77:     # or assigning attributes inside the block
   78:     def initialize(options = {})
   79:       debugger
=> 80:       default_dependencies = if Rake::Task.task_defined?("environment")
   81:         [:environment]
   82:       else
   83:         []
   84:       end
(byebug) Rake::Task.task_defined?("environment")
false
(byebug) Rake::Task.task_defined?(:environment)
false
```
It seems the rake tasks are not loaded when that check is made?

swalkinshaw:
I was trying to do something similar in another project and had looked at this rake task for inspiration. It didn't work for me either as you found @louim then I forgot to bring it up here.

Anyway, here's what I ended up with:
```ruby
Rake::Task['environment'].invoke if Rake::Task.task_defined?('environment')
```

This should be run within the task before anything else.

PR #2827, opened 2020-03-16
Fix TypeError in BigInt for nil input

davelooi:
`Integer(nil)` raises TypeError which is not handled by the BigInt class.

This will return `nil` for `nil` input.


rmosolgo:
Thanks for the fix!

PR #2834, opened 2020-03-19
Support lazy-loaded objects from input object loads:

### LINKED ISSUE #2671: `loads:` option on nested input doesn't resolve lazy values
#2323 fixed `loads:` on nested input objects, but if your `Schema.object_from_id` function returns a lazy value, it won't be resolved by the time the mutation's `#resolve` function is called.

### LINKED ISSUE #2832: Auto-loading field arguments doesn't resolve lazy values
**Describe the bug**

If `object_from_id` returns a lazy value, using `loads` in a field argument definition will result in an instance of `GraphQL::Execution::Lazy` being passed to the resolver method instead of the lazily-loaded object.

**Versions**

`graphql` version: master

**Steps to reproduce**

I modified an existing test case to demonstrate the issue:  https://github.com/rmosolgo/graphql-ruby/compare/master...lucthev:lt/lazy-autoloading

Using those changes, [this spec](https://github.com/lucthev/graphql-ruby/blob/15e49a95bb92e07fcffcd6ebde389d6550a0c7e5/spec/graphql/schema/argument_spec.rb#L208) fails with the following error:

```
Failure:
GraphQL::Schema::Argument::loads#test_0001_loads input object arguments [graphql-ruby/spec/graphql/schema/argument_spec.rb:215]
+++ actual
@@ -1 +1 @@
-"{:instrument=>#<struct Jazz::Models::Instrument name=\"Drum Kit\", family=\"PERCUSSION\">, :required_with_default_arg=>1}"
+"{:instrument=>#<GraphQL::Execution::Lazy:0xXXXXXX @get_value_func=#<Proc:0xXXXXXX@graphql-ruby/lib/graphql/schema.rb:1883>, @resolved=false, @path=nil, @field=nil>, :required_with_default_arg=>1}"
```

**Expected behavior**

The `instrument` argument to the resolver method should be an instance of `Jazz::Models::Instrument`

**Actual behavior**

The `instrument` argument to the resolver method is an instance of `GraphQL::Execution::Lazy` wrapping the `Jazz::Models::Instrument`.

---

Is this indeed a bug, or is it expected behaviour? I'd be happy to attempt a fix, although I'm not quite sure where to start — any pointers?

rmosolgo:
Previously, input object `loads:` didn't work when `object_from_id` returned a lazy value. 

Fixes #2832 
Fixes #2671 

_______

TODO: 

- [x]  The manual `.all(...).value` is not a good solution -- it forces a batch, but we should wait until later to do that. It will be hard to do it properly though 😖

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -646,22 +646,36 @@ module GraphQL
         if graphql_args.any? || @extras.any?
           # Splat the GraphQL::Arguments to Ruby keyword arguments
           ruby_kwargs = graphql_args.to_kwargs
+          maybe_lazies = []
           # Apply any `prepare` methods. Not great code organization, can this go somewhere better?
           arguments.each do |name, arg_defn|
             ruby_kwargs_key = arg_defn.keyword
 
-            loads = arg_defn.loads
-            if ruby_kwargs.key?(ruby_kwargs_key) && loads && !arg_defn.from_resolver?
+            if ruby_kwargs.key?(ruby_kwargs_key)
+              loads = arg_defn.loads
               value = ruby_kwargs[ruby_kwargs_key]
-              ruby_kwargs[ruby_kwargs_key] = if arg_defn.type.list?
-                value.map { |val| load_application_object(arg_defn, loads, val, field_ctx.query.context) }
+              loaded_value = if loads && !arg_defn.from_resolver?
+                if arg_defn.type.list?
+                  loaded_values = value.map { |val| load_application_object(arg_defn, loads, val, field_ctx.query.context) }
+                  if loaded_values.any? { |v| field_ctx.schema.lazy?(v) }
+                    GraphQL::Execution::Lazy.all(loaded_values)
+                  else
+                    loaded_values
+                  end
+                else
+                  load_application_object(arg_defn, loads, value, field_ctx.query.context)
+                end
               else
-                load_application_object(arg_defn, loads, value, field_ctx.query.context)
+                value
               end
-            end
 
-            if ruby_kwargs.key?(ruby_kwargs_key) && arg_defn.prepare
-              ruby_kwargs[ruby_kwargs_key] = arg_defn.prepare_value(obj, ruby_kwargs[ruby_kwargs_key])
+              maybe_lazies << field_ctx.schema.after_lazy(loaded_value) do |loaded_value|
+                ruby_kwargs[ruby_kwargs_key] = if arg_defn.prepare
+                  arg_defn.prepare_value(obj, loaded_value)
+                else
+                  loaded_value
+                end
+              end
             end
           end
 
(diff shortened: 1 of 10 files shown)


lib/graphql/schema/input_object.rb
```diff
@@ -45,6 +50,8 @@ def initialize(values = nil, ruby_kwargs: nil, context:, defaults_used:)
             end
           end
         end
+
+        GraphQL::Execution::Lazy.all(maybe_lazies).value
```
rmosolgo: This is not awesome -- it forces a lazy-loading batch before the last second. Maybe this can be moved to `#prepare` or something.

lib/graphql/schema/member/has_arguments.rb
```diff
@@ -90,25 +90,32 @@ def coerce_arguments(parent_object, values, context)
...
+                end
+                kwarg_arguments[arg_defn.keyword] = prepared_value
               end
-              kwarg_arguments[arg_defn.keyword] = prepared_value
             end
           end
+
+          # This _fixes_ lazy `loads:` in input objects, but it's a non-optimal implementation.
+          # It will force a batch-load for each set of arguments.
+          GraphQL::Execution::Lazy.all(maybe_lazies).value
```
rmosolgo: This is also not good. But refactoring it means that everywhere the runtime calls this method and expects arguments, it has to be ready for a lazy value.

spec/graphql/schema/resolver_spec.rb
```diff
@@ -365,7 +365,7 @@ class Mutation < GraphQL::Schema::Object
 
     class Query < GraphQL::Schema::Object
       class CustomField < GraphQL::Schema::Field
-        def resolve_field(*args)
+        def resolve_field(args, obj, ctx)
```
rmosolgo:
```suggestion
        def resolve_field(*args)
```

spec/support/jazz.rb
```diff
@@ -68,7 +68,7 @@ def initialize(*args, **options, &block)
       super(*args, **options, &block)
     end
 
-    def resolve_field(*)
+    def resolve_field(obj, args, ctx)
```
rmosolgo:
```suggestion
    def resolve_field(*)
```

lib/graphql/schema/field.rb
```diff
@@ -646,30 +646,50 @@ def to_ruby_args(obj, graphql_args, field_ctx)
...
           end
 
-          ruby_kwargs
+          if maybe_lazies.any? { |l| field_ctx.schema.lazy?(l) }
+            GraphQL::Execution::Lazy.all(maybe_lazies).then do
+              ruby_kwargs
+            end
+          else
+            ruby_kwargs
+          end
```
rmosolgo: This logic is duplicated so many places: "if any items in `a` are lazies, wrap `a` in a lazy, otherwise return this value. Maybe 

```
schema.after_any_lazies(l) {
  v 
}
```

Or, make `after_lazy` support Array natively 🤔

rmosolgo: Native array handling in `after_lazy` broke some existing behavior.

PR #2836, opened 2020-03-20
Fix `default_max_page_size` in documentation

nicolasrouanne:
For the "whole schema" it seems the method is `default_max_page_size` and not `max_page_size` which is what is specified in the documentation

Trying to use `max_page_size` will throw an unknown method error
```
undefined method `max_page_size' for MySchema:Class
```


rmosolgo:
Derp, thanks for the fix!

PR #2839, opened 2020-03-22
Add `--playground` option to install generator

joeytepp:
While `Graphiql` is great for testing GraphQL APIs, `GraphQL Playground` contains some features that `Graphiql` is missing such as http headers and is many developers preferred choice for testing their APIs. It can also be added to a rails project with the [`graphql_playground-rails`](https://rubygems.org/gems/graphql_playground-rails) gem. This PR adds an option to the install generator to set up `GraphQL Playground` upon installation in your project, by installing the gem and mounting it at the `/playground` route. Note that this does not prevent `Graphiql` from being installed as well (should it?) which can be done by including the `--skip-graphiql` option and also unlike `Graphiql` works for API only apps as well. Would love to hear thoughts and/or feedback on this addition!

### DIFF
--- a/guides/schema/generators.md
+++ b/guides/schema/generators.md
@@ -43,6 +43,7 @@ After installing you can see your new schema by:
 
 - `--relay` will add [Relay](https://facebook.github.io/relay/)-specific code to your schema
 - `--batch` will add [GraphQL::Batch](https://github.com/Shopify/graphql-batch) to your gemfile and include the setup in your schema
+- `--playground` will include `graphql_playground-rails` in the setup (mounted at `/playground`)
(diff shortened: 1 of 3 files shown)


joeytepp:
One job failed and it doesn't seem to be due to my changes [(link)](https://travis-ci.org/github/rmosolgo/graphql-ruby/jobs/665627120?utm_medium=notification&utm_source=github_status)

rmosolgo:
This is great, thanks so much for the contribution! Actually, I didn't know about this project before. Maybe I can take a look at it and consider fixing GraphiQL-Rails 😅

PR #2842, opened 2020-03-23
Fix actioncable unsubscribe

### LINKED ISSUE #2838: ActionCableLink does not send unsubscribe message to Action Cable
**Describe the bug**

Subscriptions set up with `subscribeToMore` in Apollo are successfully connecting and recieving messages, however when unsubscribing (by manually calling the `unsubscribe` function, or letting the component's query clean itself up) the unsubscribe message is not being sent to the backend. This leaves the subscription open on the backend, resulting in messages being sent on the cable with nothing to consume them.

**Versions**

Rails dependencies:
`graphql`: 1.10.3 
`rails`: 6.0.2.1
`graphql-batch`: 0.4.2

Javascript dependencies:
`graphql`: 14.5.8
`graphql-ruby-client`: 1.7.4
`actioncable`: 5.2.4-1
`apollo-client`: 2.6.4
`apollo-link`: 1.2.13

**Steps to reproduce**

- Call `subscribeToMore` in `getDerivedStateFromProps` in a React component
- Store the result of `subscribeToMore` in `state.unsubscribe`
- call `this.state.unsubscribe` in `componentWillUnmount`

**Expected behavior**

A `"command": "unsubscribe"` message should be sent on /cable

**Actual behavior**

No message is sent. Subsequent subscription triggers on the backend continue to send messages to the frontend, although nothing is consuming those messages

**Additional context**

I noticed that if the message payload contains `"more": false` the subscription properly unsubscribes, however if the client initiates the unsubscribe nothing happens

davidpimentel:
fixes #2838 

A few notes:
- I refactored the ActionCableLinkTest tests mainly to have the cable mock object more accurately reflect what actioncable returns. I'm also getting reference to the underlying subscription through the Observable subscriptions `_cleanup` object. Setting the reference within the mock (old line 13) doesn't give you the object after ActionCableLink has transformed it.
- I removed `this.unsubscribe()` on line 50 because I discovered that `observer.complete()` also calls unsubscribe, and the call is redundant.
- The actual fix involves using `Object.assign` instead of the spread operator for two reasons:
  - The spread operator doesn't carry over prototype methods
  - Creating a clone of the object breaks an equality check in actioncable when removing/unsubscribing a subscription

### DIFF
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -53,7 +53,7 @@ class ActionCableLink extends ApolloLink {
       })
 
       // Make the ActionCable subscription behave like an Apollo subscription
-      return { ...subscription, closed: false, unsubscribe: subscription.unsubscribe }
+      return Object.assign(subscription, {closed: false})
     })
   }
 }
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -47,14 +47,13 @@ class ActionCableLink extends ApolloLink {
           }
 
           if (!payload.more) {
-            this.unsubscribe()
             observer.complete()
           }
         }
       })
 
       // Make the ActionCable subscription behave like an Apollo subscription
-      return { ...subscription, closed: false }
+      return { ...subscription, closed: false, unsubscribe: subscription.unsubscribe }
     })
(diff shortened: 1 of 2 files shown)


davidpimentel:
@rmosolgo I believe this build failure is unrelated to my changes unless I'm mistaken?

rmosolgo:
Yeah, I'll have to look into that separately. Thanks for the fix here!

PR #2843, opened 2020-03-24
Fix nil override for max_page_size

### LINKED ISSUE #2840: How to override `default_max_page_size` in a connection to get infinite results?
**Describe the bug**

I set a `default_max_page_size` at the _Schema_ level. However, on some fields, I want the connection to return **all** the results for a connection. Setting a `max_page_size` on a _Connection_ type field can **change** the page size for the connection on this field, but **not remove it**.

**Versions**

`graphql`: 1.10.3
`rails` : 6.0.2.1
`graphql-batch`: 0.4.2

**GraphQL schema**

```ruby
class ApiSchema < GraphQL::Schema
  default_max_page_size 20
  query(Types::QueryType)

  use GraphQL::Analysis::AST

  use GraphQL::Pagination::Connections
end

module Types
    field :users, UserType.connection_type, null: false, max_page_size: -1

    def users
      User.all
    end
  end
end
```

**GraphQL query**

```graphql
query {
  users { 
    nodes {
      id
    }
 }
}
```

```json
{
  "data": {
    "users": {
      "nodes": {
        // always 20 users even if total number is greater
      }
    }
  }
}
```

**Expected behavior**

I would expect to be able to override the `default_max_page_size` at the field or connection level, that would also mean "remove it". 
Something like `0` or `nil` to convey "infinite number of items":
```ruby
max_page_size: 0
max_page_size: nil
```

**Actual behavior**

I can't remove it properly, using `nil` or `0` for instance, since the connection is checking the provided `max_page_size` against the `default_max_page_size` [w/ a `||` check](https://github.com/rmosolgo/graphql-ruby/blob/b5d1afb9ee7a2723a54a708839c0afd505cfddd3/lib/graphql/pagination/connection.rb#L68-L70), the `default_max_page_size` value always prevail:
```ruby
max_page_size: 0 # not working
max_page_size: nil # not working
```

I found a way around using what I consider a hack, providing `-1` value to `max_page_size`
```ruby
max_page_size: -1 # working
```

rmosolgo:
`max_page_size: nil` is supposed to allow unbounded results. 

Fixes #2840

### DIFF
--- a/lib/graphql/pagination/connection.rb
+++ b/lib/graphql/pagination/connection.rb
@@ -54,19 +54,39 @@ module GraphQL
       # @param last [Integer, nil] Limit parameter from the client, if provided
       # @param before [String, nil] A cursor for pagination, if the client provided one.
       # @param max_page_size [Integer, nil] A configured value to cap the result size. Applied as `first` if neither first or last are given.
-      def initialize(items, context: nil, first: nil, after: nil, max_page_size: nil, last: nil, before: nil)
+      def initialize(items, context: nil, first: nil, after: nil, max_page_size: :not_given, last: nil, before: nil)
         @items = items
         @context = context
         @first_value = first
         @after_value = after
         @last_value = last
         @before_value = before
-        @max_page_size = max_page_size
+
+        # This is only true if the object was _initialized_ with an override
+        # or if one is assigned later.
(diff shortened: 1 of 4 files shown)


rmosolgo:
That build is fixed on master 🍻

PR #2845, opened 2020-03-24
Fix has_previous and has_next page values when sliced_nodes has not been called

ptsavdar:
In `relation_connection` when `hasNextPage` and `hasPreviousPage` are being resolved before `nodes` or `edges` they don't return correct values because they're depending on `@before_offset` and `@after_offset` variables. These variables are initialised inside `sliced_nodes` method.
As a result when this method has not been called before resolving these fields, they return wrong values.

This PR tries to fix that issue by moving the `@before_offset` and `@after_offset` in their own private methods to work independently of `slices_nodes` method been called in advance or not.

### DIFF
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -12,7 +12,7 @@ module GraphQL
 
       def has_previous_page
         if @has_previous_page.nil?
-          @has_previous_page = if @after_offset && @after_offset > 0
+          @has_previous_page = if after_offset && after_offset > 0
             true
           elsif last
             # See whether there are any nodes _before_ the current offset.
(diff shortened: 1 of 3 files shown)


rmosolgo:
Great catch! Sorry about the bug and thanks for tracking down a fix 🍻 

I'll take care of that failing build separately.

vpetkar:
Tracked down this issue for a problem we're trying to resolve. It still seems to cause unexpected behaviour. Depending on where pageInfo is called within the query, it will return incorrect results. Works when `pageInfo` before `nodes` -- but not after `nodes {}`. Looks like the integration test here has moved it to the top.

rmosolgo:
Hey @vpetkar , would you mind opening a new issue to describe the bug you're seeing?

vpetkar:
@rmosolgo Done! https://github.com/rmosolgo/graphql-ruby/issues/2972

PR #2847, opened 2020-03-25
Fix several typos in docs

TolichP:


rmosolgo:
Thanks for "the the" fix 😆 !

PR #2848, opened 2020-03-26
feat: Date scalars accept Date, DateTime, and well-formatted Strings

grxy:
## Motivation
It can be easy to unintentionally resolve a Date object for a DateTime scalar field and vice versa. The intent of this PR is to make `ISO8601Date` and `ISO8601DateTime` more flexible, by allowing them to be resolved with either a `Date` object, a `DateTime` object, or a well-formatted string.

## Changes
- Updates `ISO8601Date.coerce_result` to parse the resolved value and coerce to a `Date` object
- Updates `ISO8601DateTime.coerce_result` to parse the resolved value and coerce to a `DateTime` object while retaining precision if the resolved value is already a `DateTime` object
- Updates the exception text for `ISO8601DateTime` to reflect the extended input options
- Adds test cases for all combinations of scalars + resolve options

## Alternative approach
- Each `coerce_result` could perform explicit logic for each potential type of resolved value, but this would lead to more verbose code (i.e. if `Date`, if `DateTime`, if `String`)

### DIFF
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -34,7 +34,7 @@ module GraphQL
       def self.coerce_result(value, _ctx)
         value = value.iso8601(time_precision) if value.class == DateTime
         DateTime.parse(value.to_s).iso8601(time_precision)
-      rescue ArgumentError => e
+      rescue ArgumentError
         raise GraphQL::Error, "An incompatible object (#{value.class}) was given to #{self}. Make sure that only Dates, DateTimes, and well-formatted Strings are used with this type."
       end
 
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -35,7 +35,7 @@ module GraphQL
         value = value.iso8601(time_precision) if value.class == DateTime
         DateTime.parse(value.to_s).iso8601(time_precision)
       rescue ArgumentError => e
-        raise GraphQL::Error, "An incompatible object (#{value.class}) was given to #{self}. Make sure that only DateTimes are used with this type."
+        raise GraphQL::Error, "An incompatible object (#{value.class}) was given to #{self}. Make sure that only Dates, DateTimes, and well-formatted Strings are used with this type."
       end
 
       # @param str_value [String]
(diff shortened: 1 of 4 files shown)


lib/graphql/types/iso_8601_date_time.rb
```diff
@@ -29,12 +29,13 @@ def self.time_precision=(value)
         @time_precision = value
       end
 
-      # @param value [DateTime]
+      # @param value [Date,DateTime,String]
       # @return [String]
       def self.coerce_result(value, _ctx)
-        value.iso8601(time_precision)
+        value = value.iso8601(time_precision) if value.class == DateTime
```
rmosolgo: Why use `.class ==` here, instead of `.is_a?(DateTime)`? Are you trying to exclude subclasses of `DateTime`? 

(This might be a breaking change, since this method _used_ to accept anything that responded to `.iso8601(time_precision)`)

grxy: > Why use `.class ==` here, instead of `.is_a?(DateTime)`? Are you trying to exclude subclasses of `DateTime`?

Call it a less-than-ideal understanding of Ruby semantics. :)

> (This might be a breaking change, since this method used to accept anything that responded to `.iso8601(time_precision)`)

The main reason I didn't do this is that `Date#iso8601` raises an exception when passed `time_precision`, and ignoring the precision leads to broken tests. Can you think of a better way to get around this constraint?

rmosolgo: How about: 

```ruby 
def coerce_result(value, _ctx)
  case value 
  when DateTime 
    value.iso8601(time_precision)
  when Date 
    DateTime.parse(value.to_s).iso8601(time_precision)
  when String 
    DateTime.parse(value).iso8601(time_precision)
  else 
    # In case some other API-compliant thing is given: 
    value.iso8601(time_precision)
  end 
rescue ArgumentError, NoMethodError 
  raise GraphQL::Error, "An incompatible object (#{value.class}) was given to #{self}. Make sure that only Dates, DateTimes, and well-formatted Strings are used with this type."
end 
```

?

grxy: Sounds good. I'll incorporate this. Thanks.

grxy: @rmosolgo Got stuck for a while getting the `when String` case to pass. Turns out `String` was referencing `GraphQL::Types::String` instead of `::String`. Should be good to go now.

rmosolgo: 😖 good catch!

rmosolgo:

PR #2854, opened 2020-03-30
Improve error message for failed list result

### LINKED ISSUE #2588: undefined method `map' for #<GraphQL::Relay::ArrayConnection>
This error occurs when using a 1.8+ version of the gem and DSL style and when you have defined graphql types in your schema that end with the word `Connection`.

Any types ending with the word `Connection` automatically get graphql-ruby's connection handling applied to them, which results in the error listed in the title if the resolvers on the fields in question don't return an **array** or an **ActiveRecord::Relation**.

**In short you get graphql-ruby's connection handling on types for which you may not expect that to happen.** 

This is particularly surprising when converting from old style DSL (pre.1.8+) where this automatic `Connection` handling behavior does not exist.

While this is documented [here](https://graphql-ruby.org/relay/connections.html) and [here](https://graphql-ruby.org/schema/class_based_api.html) (see the docs for a workaround), I would recommend keeping this issue open for the benefit of anyone else who might encounter this error.

Users who are not interested in graphql-ruby's built in connection support might skip over these sections of the documentation and might have difficulties figuring out the cause of the error.

rmosolgo:
Improve the error message when the returned value doesn't satisfy GraphQL's expectation. 

Fixes #2588 

new message:

```
Failed to build a GraphQL list result for field `Query.things` at path `things`.
Expected `#<GraphQL::Pagination::ArrayConnection:0x00007fe77f3d2020 @items=[{:name=>"thing1"}, {:name=>"thing2"}], @context=#<Query::Context ...>, @first_value=nil, @after_value=nil, @last_value=nil, @before_value=nil, @has_max_page_size_override=true, @max_page_size=nil>` to implement `.each` to satisfy the GraphQL return type `[ThingConnection!]!`.

This field was treated as a Relay-style connection; add `connection: false` to the `field(...)` to disable this behavior.
```

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -332,20 +332,26 @@ module GraphQL
             inner_type = type.of_type
             idx = 0
             scoped_context = context.scoped_context
-            value.each do |inner_value|
-              next_path = path.dup
-              next_path << idx
-              next_path.freeze
-              idx += 1
-              set_type_at_path(next_path, inner_type)
-              # This will update `response_list` with the lazy
-              after_lazy(inner_value, owner: inner_type, path: next_path, scoped_context: scoped_context, field: field, owner_object: owner_object, arguments: arguments) do |inner_inner_value|
-                continue_value = continue_value(next_path, inner_inner_value, field, inner_type.non_null?, ast_node)
-                if HALT != continue_value
-                  continue_field(next_path, continue_value, field, inner_type, ast_node, next_selections, false, owner_object, arguments)
+            begin
+              value.each do |inner_value|
+                next_path = path.dup
+                next_path << idx
+                next_path.freeze
+                idx += 1
+                set_type_at_path(next_path, inner_type)
+                # This will update `response_list` with the lazy
+                after_lazy(inner_value, owner: inner_type, path: next_path, scoped_context: scoped_context, field: field, owner_object: owner_object, arguments: arguments) do |inner_inner_value|
+                  continue_value = continue_value(next_path, inner_inner_value, field, inner_type.non_null?, ast_node)
+                  if HALT != continue_value
+                    continue_field(next_path, continue_value, field, inner_type, ast_node, next_selections, false, owner_object, arguments)
+                  end
                 end
               end
+            rescue NoMethodError
+              # This happens when the GraphQL schema doesn't match the implementation. Help the dev debug.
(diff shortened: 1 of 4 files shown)


rmosolgo:
I wonder where my Travis CI statuses are 🤔  It _is_ building, over at https://travis-ci.org/github/rmosolgo/graphql-ruby/builds/668818546

rmosolgo:
image

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -332,20 +332,26 @@ def continue_field(path, value, field, type, ast_node, next_selections, is_non_n
...
+                set_type_at_path(next_path, inner_type)
+                # This will update `response_list` with the lazy
+                after_lazy(inner_value, owner: inner_type, path: next_path, scoped_context: scoped_context, field: field, owner_object: owner_object, arguments: arguments) do |inner_inner_value|
+                  continue_value = continue_value(next_path, inner_inner_value, field, inner_type.non_null?, ast_node)
+                  if HALT != continue_value
+                    continue_field(next_path, continue_value, field, inner_type, ast_node, next_selections, false, owner_object, arguments)
+                  end
                 end
               end
+            rescue NoMethodError
```
dewski, line 350: @rmosolgo 👋 hope all is well.

I've ran into the situation where I'd get the `ListResultFailedError` in development and spend a lot of time trying to find out why it failed to build a list at the `SomeConnection.edges` field as the error says:

> Failed to build a GraphQL list result for field `ItemConnection.edges` at path `items.edges`.
Expected `[#<GraphQL::Pagination::Connection::Edge:0x00007fc1b7e07560 @connection=#<GraphQL::Pagination::ActiveRecordRelationConnection ...>>]` to implement `.each` to satisfy the GraphQL return type `[ItemEdge]`.

I couldn't figure out why the connection wasn't returning an array (due to the implement each method message) because I wasn't able to reproduce it when logging it in development. When I open the GraphQL gem with `bundle open graphql` and remove the rescue I get the original exception which helps me figure out what it meant exactly:

```
undefined method `organization' for nil:NilClass
/app/graphql/types/organization_type.rb:9:in `authorized?'
```

I'm not sure I understand #2588 completely for which this error was introduced for, but I'm wondering if this rescue is a bit too catch all? Curious if you have any opinions.

rmosolgo: > a bit too catch all

Ahh, I totally agree. Let me see about tightening up that rescue.

PR #2859, opened 2020-04-02
Add Query#fingerprint for grouping queries in logs

rmosolgo:
We use something like this at GitHub to connect query runs over time. You can use these identifiers to trace query performance over the long run.

__TODO:__

- Make the string a bit more human-friendly:
  - [x] Add selected operation name and/or number of root selections to query hash 
  - [x] Add number of variables to var hash 

cc @swalkinshaw I'd love to know, do y'all at shopify have something like this? (Any advice in this area?)

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -3,6 +3,7 @@ require "graphql/query/arguments"
 require "graphql/query/arguments_cache"
 require "graphql/query/context"
 require "graphql/query/executor"
+require "graphql/query/fingerprint"
 require "graphql/query/literal_input"
 require "graphql/query/null_context"
 require "graphql/query/result"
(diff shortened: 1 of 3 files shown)


lib/graphql/query.rb
```diff
@@ -265,6 +266,21 @@ def sanitized_query_string
...
     end
 
+    # @return [String] An opaque hash identifying this query-variables combination
+    def fingerprint
+      @fingerprint ||= "#{query_fingerprint}/#{variables_fingerprint}"
+    end
+
+    # @return [String] An opaque hash for identifying this query's given query string
+    def query_fingerprint
+      @query_hash ||= Fingerprint.generate(query_string)
```
jturkel, line 276: Is it worth fingerprinting a version of the query string with normalized whitespace formatting?

eapache: We do `.strip` the query before we hash it. Normalizing all whitespace would be expensive since you'd have to parse/reprint the query to avoid mangling string constants.

swalkinshaw:
Yeah as Evan said this is basically what we do. Only additional thing you mentioned is the selected operation name. What you is technically the fingerprint of the entire query string but I'm guessing in most cases people will only care about the fingerprint for the selected operation.

rmosolgo:
> entire query string ... selected operation

Wow, I didn't even think of that. In some cases, it might be downright _wrong_ to fingerprint without the selected operation name, otherwise different operations would be tracked together. (That would only happen if the _given variables_ matched too, but, hey, could happen!) 

> normalized whitespace 

I'm a bit torn on that. There's other normalization you could do too, like sorting fields alphabetically and removing comments. You could even take fragments and inline them.

On the one hand, it's good to have a fingerprint that reflects changes to client queries. On the other hand, maybe you want to ignore "insignificant" changes to those queries. 

I could go either way. Maybe measuring the perf impact, or offering both `fingerprint` and `normalized_fingerprint` would be worth it.

lib/graphql/query.rb
```diff
@@ -265,6 +266,21 @@ def sanitized_query_string
...
+    end
+
+    # @return [String] An opaque hash for identifying this query's given a variable values (not including defaults)
+    def variables_fingerprint
+      @variables_hash ||= Fingerprint.generate(provided_variables.to_json)
```
rmosolgo, line 281: TODO make ivars match method names 🤦‍♂

PR #2861, opened 2020-04-02
Change `eql?` to `==`

jhawthorn:
`eql?` is the hash equality operator. Objects which implement `eql?` must also implement a matching `hash`, otherwise using these objects as hash keys or calling uniq with them will behave unpredictably.

The implementation was also not a good match for `eql`'s hash semantics. It checked that scalars and children were equal, but not filename, line, or col.  Additionally, subclasses like `Field` and `FragmentDefinition` also had additional attributes which were not tested (EDIT: maybe this is wrong).

This commit changes `eql?` to `==` so that we have an easy way to do tree comparison (as needed for comparing children) but we don't mess up hashing semantics.

### DIFF
--- a/lib/graphql/language/nodes.rb
+++ b/lib/graphql/language/nodes.rb
@@ -39,11 +39,11 @@ module GraphQL
 
         # Value equality
         # @return [Boolean] True if `self` is equivalent to `other`
-        def eql?(other)
+        def ==(other)
           return true if equal?(other)
-          other.is_a?(self.class) &&
-            other.scalars.eql?(self.scalars) &&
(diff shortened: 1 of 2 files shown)


rmosolgo:
Seems good to me, are there any imaginable real-world use-cases this might break? (Does `.eql?` power other Ruby features, besides direct calls to `.eql?`?)

jhawthorn:
I think that any of the internal Ruby features this powers will be relying on the value of `hash` _most_ of the time, so fixing this will just make that usage more consistent.

rmosolgo:
Thanks for the fix!

PR #2863, opened 2020-04-03
Update scalars.md

christophweegen:
`Boolean` won't be found. Must be `GraphQL::Types::Boolean`. 

Another thing is: 

Defining a `GraphQL::Types::ISO8601Date` in a fields argument will raise `uninitialized constant GraphQL::Types::ISO8601Date::Date` same for `GraphQL::Types::ISO8601DateTime`.

Version `1.10.5`

Example:
```ruby
field :test, String, null: false do
  argument :foo, GraphQL::Types::ISO8601Date, required: true
end
```

Thanks and nice weekend :-)

### DIFF
--- a/guides/type_definitions/scalars.md
+++ b/guides/type_definitions/scalars.md
@@ -31,7 +31,7 @@ field :top_score, Integer, null: false
 # Float field
 field :avg_points_per_game, Float, null: false
 # Boolean field
-field :is_top_ranked, Boolean, null: false
+field :is_top_ranked, GraphQL::Types::Boolean, null: false
 # ID field
 field :id, ID, null: false
 # ISO8601DateTime field


rmosolgo:
`Boolean` works fine for me. Here's a little script to try it out: 

```ruby
# test.rb
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", "1.10.5"
end

  field :bool, Boolean, null: false
end

  query(Query)
end

puts Schema.to_definition
```

```
$ ruby test.rb
type Query {
  bool: Boolean!
}
```

Does that work for you?

-----

It sounds like the `Date*` types need `require "date"` at the top. But most people don't run into it because they're using Rails.

rmosolgo:
> which resources made you initially able to write GraphQL-Ruby?

It was pretty much just the spec. Also, I was 5 years younger and had no kids, so I had a lot of time to work in it :P 

I didn't spend much time reading the implementation of the JS library, although I did look at the schema definition API for inspiration. 

I also had some general language implementation resources that I've used over the years:

- "Ruby Under a Microscope" -- can't recommend it enough 
- Stanford Compilers course, previously on Coursera 
- "Virtual Machines", Smith & Nair -- didn't finish it tho
- "Essentials of Programming Languages" -- Inspired the design of the new runtime 


> some tips

Start with the _easiest_ thing that will work! For example, the parser was first built with Parslet, and only replaced when it was _actually_ a bottleneck. 

Hope that helps!

christophweegen:
Cool, thanks a lot for that detailed explanation 👍

PR #2866, opened 2020-04-08
Dispatch initial response in Ably handler

jscheid:
Hi Robert, as far as I can tell this is necessary for making the [initial response](https://github.com/rmosolgo/graphql-ruby/blob/master/guides/subscriptions/subscription_classes.md#adding-an-initial-response) work, or am I missing something?

This should probably also be done for ActionCable etc., but I'll leave that up to you.

There is a slight semantic change here in that a message with an empty (or otherwise malformed) payload would previously have triggered a call to `onNext({ data: undefined })` and now doesn't trigger one at all. This seems the most consistent and straightforward to me. Could do this a few different ways as well -- feel free to change.

Prettier made some formatting changes, and I've added a small bit of config to match your style. Feel free to undo/remove.

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -30,5 +30,8 @@
     "graphql": "^14.3.1",
     "minimist": "^1.2.0",
     "typescript": "^3.7.5"
+  },
+  "prettier": {
+    "semi": false
   }
 }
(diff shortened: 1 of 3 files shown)


jscheid:
Actually, regardless of the initial response I believe this is necessary so the client can surface any errors occurring during subscription setup, such as invalid arguments.

rmosolgo:
Awesome, thanks for sharing the fix here!

jscheid:
Thanks! Any chance you could cut a new release with this?

jscheid:
Ideally with #2877 in it as well.

rmosolgo:
Yes, 🚢 1.7.7 !

PR #2877, opened 2020-04-14
Pass request errors to observer in Ably handler

jscheid:
Hi Robert, here is another one where any errors in the original request (such as a 500) weren't being passed on to the subscription error handler.


rmosolgo:
Thanks for the fix here!

Issue #2869, opened 2020-04-09
Regression in GraphQL 1.10: no way to get access to query args in their original form


myronmarston:
**Describe the bug**

We've built an indexing and search framework on top of graphql-ruby that operates on GraphQL schemas generically, by translating a GraphQL query to an elastic search query to query our indexed data.  The framework is designed to work with either `camelCase` schemas or `snake_case` schemas (really, it doesn't care; as long as the user of the framework uses the same casing for a field in their GraphQL schema and their ElasticSearch JSON field schema, it'll "just work" without us having to think about it).  This all worked great with graphql-1.9, but I've been trying to upgrade to graphql-1.10 and I'm running into problems. Specifically, graphql-1.10 coerces all inputs to ruby keyword-style syntax (e.g. symbols keys instead of strings, `snake_case` over `camelCase`, etc).  This is causing us a problem because when we get a user of our framework is using a `camelCase` schema (as most GraphQL schemas are...) our framework is now receiving the field name in `snake_case` instead of `camelCase`, which causes the ElasticSearch query to fail to match anything.  I've spent a few hours trying to figure out an alternate way to get the original argument hash in whatever casing it started in, and I haven't found a way that works.

I can understand why most users probably prefer getting the query arguments in ruby keyword style syntax, but for those of us who want the argument hash as it was provided in the query, it's frustrating not to have access to it anymore, and I haven't yet found a work around I'm satisfied with.

Can the graphql gem provide access to the original args hash as it was provided in 1.9 using an alternate method or something?

**Versions**

`graphql` version: worked as desired in all 1.9.x versions, fails to work as I need it to in 1.10.x

**GraphQL schema**

See my example script below.

**GraphQL query**

See my example script below.

**Steps to reproduce**

Put this script in a file like `script/demo_args_coercion_issue.rb`:

``` ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'graphql', ENV.fetch('VERSION')
end

require 'graphql'
require 'pp'

schema_string = <<~SCHEMA
  input IdFilter {
    anyOf: [ID!]!
  }

  input IntFilter {
    anyOf: [Int!]
    gt: Int
    gte: Int
    lt: Int
    lte: Int
  }

  type Widget {
    id: ID!
    amountCents: Int!
    name: String!
  }

  input WidgetFilter {
    id: IdFilter
    amountCents: IntFilter
  }

  type Query {
[...]

PR #2881, opened 2020-04-16
Argument metadata for interpreter

### LINKED ISSUE #2841: GraphQL::Query::Arguments::ArgumentValue no longer accessible via visitor.arguments_for
# Problem

I'm working on getting [graphql-metrics](https://github.com/Shopify/graphql-metrics) compatible with graphql-ruby 1.10.x, and I've encountered an issue while trying to understand how `visitor.arguments_for` works in recent versions of graphql-ruby, in particular when the new interpreter is enabled.

On 1.9.x these calls work just fine: https://github.com/Shopify/graphql-metrics/blob/master/lib/graphql/metrics/analyzer.rb#L39-L40

We opted to recurse over `arguments.argument_values.values`  (see `#extract_arguments` in the above file), in order to get metrics metadata off of instances of `GraphQL::Query::Arguments::ArgumentValue`.

As of 1.10.0, when `interpreter? => true`, `visitor.arguments_for` now returns a hash containing only the arg names pointing to scalars, rather than a number of wrapper objects which may contain instances of `GraphQL::Query::Arguments::ArgumentValue`, via `GraphQL::Execution::Interpreter::ArgumentsCache` : https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/interpreter/arguments_cache.rb#L6

If we want to continue to gather metrics like `ArgumentValue#default_used?` and `#value` from arguments passed to a field visited via `GraphQL::Analysis::AST::Analyzer#on_leave_field` we need an alternate method to get `ArgumentValue`s, it seems.

I've looked at https://graphql-ruby.org/queries/interpreter.html and the 1.10.x changelog notes.

By the way, I recall that the reason we opted to stick with doing everything (including recursing through args) via `on_leave_field` is that `on_leave_arguments` skipped some things not in the AST, like arguments provided via variables.

# Possible solutions

I spoke today with @rmosolgo on the GraphQL Slack org's #ruby channel. Highlights included:

> Robert: `default_used` is basically not set by the interpreter, and it’s not provided to resolvers, since they all get plain hashes now
> [this] will require some improvements, because `default_used?` is not available at all right now 
> gotta decide whether to rework arguments_for(…) to include that info, or provide another API that can return it
> 
> Chris: makes sense! 
> Just so I can better understand: was the change in contract for `arguments_for` to return a simplified hash the result of _both_ the bump from 1.9 to 1.10 _and_ enabling the new interpreter? https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query.rb#L248-L249
> 
> Robert: in 1.10, that method started behaving differently when `interpreter?`
> so, i think “yes” in the sense that both 1.10 and `interpreter?` are necessary causes

Robert then suggested a few ideas for a solution in a future 1.10.x release:

> I can imagine some improvement like this:

```diff
-    def arguments_for(ast_node, definition, parent_object: nil)
+    def arguments_for(ast_node, definition, parent_object: nil, as_hash: true)
       if interpreter?
         @arguments_cache ||= Execution::Interpreter::ArgumentsCache.new(self)
-        @arguments_cache.fetch(ast_node, definition, parent_object)
+        arguments = @arguments_cache.fetch(ast_node, definition, parent_object)
+        # return an `Arguments` instance if requested, otherwise return a plain Hash
+        as_hash ? arguments.to_h : arguments 
```

> or adding `.to_h` calls in places that call `.arguments_for`. (also Interpreter::ArgumentsCache would need an upgrade)
> or maybe the solution is to stop using Intepreter::ArgumentsCache, and instead do

```ruby
if interpreter && as_hash 
  arguments.to_h
else 
  arguments 
end 
```

> Chris: yeah that's interesting. the to_h struck me as something which may be good to opt-in/out
> Is there a possible performance regression if apps want that extra detail, and opt not to get the to_h result? Or is it all in-memory anyways?
>
> Robert: it’s possible but i doubt it’s very much
>
> Chris: 👍  I'd be happy to try my hand at a PR to do this
> Robert: 👍

rmosolgo:
The interpreter runtime uses plain-Ruby hashes for arguments by default, but sometimes, you need more metadata about them (eg, `default_used?` (#2841) or original GraphQL names (#2869))

Let's add some way to access that info into: 

- field resolution (as `extras`?) 
- analyzers (via `query.arguments_for`) 

This implementation exposes fewer methods, just because it's not clear we _need_ to port over several of the others. Happy to add those when the need arises. 

I didn't reuse `GraphQL::Query::Arguments` for a few reasons: 

- It eagerly calculates `default_used?` (actually, it requires the caller to have calculated it), and although we calculate that in the interpreter, we don't pass it along the same way. In this new approach, that calculation is deferred until someone actually asks for it. 
- It doesn't support string access -- I'm trying to standardize around Ruby-style keyword access. Hopefully this doesn't come back to bite me. 
- It doesn't support method access. I think this was overkill, also, `Query::Arguments` has this expectation that it's subclassed for each field in the schema, and that just doesn't apply anymore. Plain hashes got rid of the perf problem which that system was added to mitigate. 

TODO: 

- [x] Extend `#arguments_for` & test
- [x] Add an `extras ...` option & test
- [x] Build a proper object for this data 
- [x] Don't build detailed args unless necessary? 
- [x] Add docs 

fixes #2841


rmosolgo:
I've tried to implement this to satisfy the API used in https://github.com/Shopify/graphql-metrics/blob/master/lib/graphql/metrics/analyzer.rb, with the exception that the class names have changed. 

But, I think it could support both by adding some cases: 

```diff
-       when ::GraphQL::Query::Arguments
+       when ::GraphQL::Query::Arguments, ::GraphQL::Execution::Interpreter::Arguments
          argument.each_value do |arg_val|
            extract_arguments(arg_val, field_defn, parent_input_object)
          end
-       when ::GraphQL::Query::Arguments::ArgumentValue
+       when ::GraphQL::Query::Arguments::ArgumentValue, ::GraphQL::Execution::Intepreter::Arguments::ArgumentValue
          extract_argument(argument, field_defn, parent_input_object)
          extract_arguments(argument.value, field_defn, parent_input_object)
```

I _think_ that should work ok. @chrisbutcher, do you want to give it a spin before I merge this branch, or would you rather just check it later?

chrisbutcher:
Awesome, thank you!

> I think that should work ok. @chrisbutcher, do you want to give it a spin before I merge this branch, or would you rather just check it later?

I can give it a test this morning!

chrisbutcher:
> I'm going to try this all again, with an eye on doing it the same way the old runtime did. Namely, building up a big, robust Arguments object at runtime. Even if most use-cases don't require the metadata, it'll be much easier to capture it there, where we already have everything we need. And it's probably not a big perf hit -- after all, that's the way it was done for years (and most fields have no arguments -- that's the case that needs to be optimized with a singleton null value).

This sounds great. Even if there is a small perf hit and there ends up being a way to toggle this on/off at a high level, that could be useful too IMO.

Another bit of feedback I have: going back to one of the comments on https://github.com/rmosolgo/graphql-ruby/issues/2841:

> By the way, I recall that the reason we opted to stick with doing everything (including recursing through args) via on_leave_field is that on_leave_argument skipped some things not in the AST, like arguments provided via variables.

Do you think `on_leave_argument` will likely keep that contract going forward? The graphql-metrics use case has found it convenient to use `on_leave_field` even for args logging, primarily because variable-specified arguments are available from it. If the gem provided more hooks, or provided more metadata through the existing hooks (either as an alternative to needing to recurse various gem structures), that may be an overall better analyzer API.

rmosolgo:
Yes, I think `on_leave_field` is the best place for that. 

When it comes to APIs for inspecting arguments, I can't think of a better API than providing a useful, normalized data structure for them.  The only alternative I can think of is to inline them into the AST (which has its uses), but then it becomes difficult to distinguish `default_value:` values -- is a certain AST value client-provided or server default? 

Can you imagine something better than that? I'm stumped 🤔

rmosolgo:
Maybe I closed this too soon -- it's just as easy to fix this branch as to cook up another one.

rmosolgo:
I basically did the refactor in https://github.com/rmosolgo/graphql-ruby/pull/2881/commits/a5f2744307710dc560ac8e59cba1520a215aa52f, and at least it doesn't _look_ like much overhead, just initializing a handful of objects in cases where there are actually arguments. 

And, bonus, it pretty much supports `InputObject#arguments` like the legacy runtime did. Although again, `Interpreter::Arguments` doesn't have quite as many methods as the original. If we need to add more to the `def_delegators` call, we can, etc. 

@chrisbutcher , could you give it one more go?

lib/graphql/schema/member/has_arguments.rb
```diff
@@ -110,14 +114,23 @@ def coerce_arguments(parent_object, values, context)
                   arg_defn.prepare_value(parent_object, coerced_value, context: context)
                 end
 
-                kwarg_arguments[arg_defn.keyword] = prepared_value
+                kwarg_arguments[arg_key] = prepared_value
+                # TODO code smell to access such a deeply-nested constant in a distant module
+                argument_values[arg_key] = GraphQL::Execution::Interpreter::Arguments::ArgumentValue.new(
```
rmosolgo, line 119: I feel like this smell is telling me to figure out a way to use `Query::Arguments` here, despite my efforts to the contrary 😆

chrisbutcher:
> @chrisbutcher , could you give it one more go?

Absolutely; will do shortly!

chrisbutcher:
https://github.com/Shopify/graphql-metrics/pull/23

🎉 

```
# Running:

..............

Finished in 13.087728s, 1.0697 runs/s, 3.5912 assertions/s.

14 runs, 47 assertions, 0 failures, 0 errors, 0 skips
```

lib/graphql/query.rb
```diff
@@ -246,10 +246,13 @@ def irep_selection
     # @param definition [GraphQL::Schema::Field]
     # @param parent_object [GraphQL::Schema::Object]
     # @return Hash{Symbol => Object}
-    def arguments_for(ast_node, definition, parent_object: nil)
+    def arguments_for(ast_node, definition, parent_object: nil, detailed: false)
       if interpreter?
         @arguments_cache ||= Execution::Interpreter::ArgumentsCache.new(self)
-        @arguments_cache.fetch(ast_node, definition, parent_object)
+        combined_args = @arguments_cache.fetch(ast_node, definition, parent_object)
+        schema.after_lazy(combined_args) do |args|
```
chrisbutcher, line 253: Is this here to let you get args after they are mixed in from variables?

rmosolgo: It's from #2834 -- input_object `loads:` hooks can return promises, which means `coerce_arguments` can return promises, which means `arguments_for` can return promises ...

chrisbutcher: Thanks for explaining!

rmosolgo:
We can probably remove the `detailed:` option if we make sure to unwrap with `.keyword_arguments` at runtime. I took a try in 9c982e9e3, let me know if you think it's an improvement. (It's a bit of a change because anyone who _expected_ `.arguments_for` to return a Hash will now get an `Arguments` instance.)

chrisbutcher:
Updated in https://github.com/Shopify/graphql-metrics/pull/23 and looks good to me!

> It's a bit of a change because anyone who expected .arguments_for to return a Hash will now get an Arguments instance.

Yeah, in our case at least, it's most useful to get `Arguments`.

lib/graphql/schema/member/has_arguments.rb
rmosolgo: I'm gonna leave it. I don't want to mess with `Query::Arguments`'s expectation that custom subclasses are built, or with it's normalize-to-string behavior.

PR #2882, opened 2020-04-16
Inherit superclass interfaces to children in to_graphql

### LINKED ISSUE #2873: Class inheritance stop working for `implements GraphQL::Types::Relay::Node`
For inheritance type classes structure, f.e.
```
class Types::UserType < Types::BaseObject
...
end

class Types::BaseObject < GraphQL::Schema::Object
  # or implements GraphQL::Types::Relay::Node
end
```
`Types::UserType` doesn't available for `node/nodes` query, but it was in version 1.10.5

Is it expected changes or something went wrong?

`graphql` version: 1.10.6

grcooper:
In the situation of:

```
module MyInterface
end

class MyClass
  include MyInterface
end

class MySubClass < MyClass
end
```

`MySubClass.to_graphql` was not correctly passing down its superclass interfaces. This because of a small bug in a recent PR in which the wrong method was used to pass down these interfaces:

https://github.com/rmosolgo/graphql-ruby/pull/2791/files#diff-bb60261811b3e1117b450eedea63f7fcR165

We used `own_interface_type_memberships` but previously we were using `interfaces` which also read from the superclass's interfaces.

cc @otereshin-shopify who found the bug

----- 

Fixes #2873

### DIFF
--- a/lib/graphql/schema/object.rb
+++ b/lib/graphql/schema/object.rb
@@ -162,7 +162,7 @@ module GraphQL
           obj_type = GraphQL::ObjectType.new
           obj_type.name = graphql_name
           obj_type.description = description
-          obj_type.structural_interface_type_memberships = own_interface_type_memberships
+          obj_type.structural_interface_type_memberships = interface_type_memberships
           obj_type.introspection = introspection
           obj_type.mutation = mutation
           obj_type.ast_node = ast_node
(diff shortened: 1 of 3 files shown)


rmosolgo:
Awesome, thanks so much for tracking this down. I updated the OP because I don't think issue linking works in comments. Let's see: 

Fixes #2873

rmosolgo:
Seems like GitHub is stuck in limbo. I tried to merge it and it failed, but this change is on master: 

image

But the PR is still open: 

image

rmosolgo:
😌

grcooper:
Very cool commit 😎

PR #2885, opened 2020-04-16
Fixed superclases -> superclasses

grcooper:
Caught a spelling mistake in my most recent PR :(


rmosolgo:
More like super _glasses_ 😎

PR #2886, opened 2020-04-16
Match relay spec for startCursor and endCursor

Derenge:
I was unsure on what the encoding should be if there isn't an offset.

### DIFF
--- a/lib/graphql/pagination/connection.rb
+++ b/lib/graphql/pagination/connection.rb
@@ -143,12 +143,12 @@ module GraphQL
 
       # @return [String] The cursor of the first item in {nodes}
       def start_cursor
-        nodes.first && cursor_for(nodes.first)
+        cursor_for(nodes.first)
       end
 
       # @return [String] The cursor of the last item in {nodes}
       def end_cursor
-        nodes.last && cursor_for(nodes.last)
+        cursor_for(nodes.last)
       end
 
       # Return a cursor for this item.
--- a/lib/graphql/pagination/connection.rb
+++ b/lib/graphql/pagination/connection.rb
@@ -143,12 +143,12 @@ module GraphQL
 
       # @return [String] The cursor of the first item in {nodes}
       def start_cursor
-        cursor_for(nodes.first)
+        nodes.first ? cursor_for(nodes.first) : ""
       end
 
       # @return [String] The cursor of the last item in {nodes}
       def end_cursor
-        cursor_for(nodes.last)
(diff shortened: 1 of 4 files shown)


lib/graphql/pagination/connection.rb
```diff
@@ -143,12 +143,12 @@ def has_previous_page
 
       # @return [String] The cursor of the first item in {nodes}
       def start_cursor
-        nodes.first && cursor_for(nodes.first)
+        cursor_for(nodes.first)
```
rmosolgo, line 146: This change will have downstream effects: anyone who has written a custom connection will have to update their code to handle `nil`. 

How about, instead, we handle the nil case _here_, for example: 

```suggestion
        nodes.first ? cursor_for(nodes.first) : ""
```

(I'm also not sure what to do when no node is present, but let's start with an obviously-empty value, `""`.) 

That would require fewer changes in user code after this update. Do you think it would work here and in `end_cursor` below?
```

Derenge: Yeah, much better. Thanks!

rmosolgo:
Ok, sorry to go back and forth on this, but as I was reviewing again, I asked myself, "How does Relay Modern _expect_ empty lists to be handled?" 

First, I found this issue, reporting a warning when `startCursor` or `endCursor` was null: https://github.com/facebook/relay/issues/1852

It looks like someone tried to address that issue by suppressing the warning, but that change was rejected: https://github.com/facebook/relay/pull/2122

Instead, a Relay maintainer explicitly updated the tests to expect `null` in the case of an empty list: 

https://github.com/facebook/relay/commit/a17b462b3ff7355df4858a42ddda75f58c161302#diff-344d406b6636acdb3db468e6ffdc97a6R543-R544

However, it looks like that change was not accompanied by a change to the Relay spec. So, someone else pushed a change to the spec: https://github.com/facebook/relay/pull/2655. But that change has not been merged yet. 

Here's another implementation which decided to _allow null_ for startCursor and endCursor: https://github.com/graphql-nexus/schema/pull/324#discussion_r369808892


All that to say, it looks to me like `null` is _actually_ the right behavior here, but the Relay spec hasn't been updated yet. In light of that, I think it's best to keep the current behavior (and _hope_ that the Relay spec is updated to match Relay's implementation!). 

Sorry to reverse course on this change, but does that seem like an appropriate interpretation of the changes in Relay?

Derenge:
I'm thankful you dug into it in more detail!

No change is easy, I'll go ahead and close this PR.

PR #2891, opened 2020-04-17
change Resolver#type function's YARD

igorlvicente:
RubyMine keeps complaining about wrong types when I pass an Array


rmosolgo:
Awesome, thanks for the fix!

PR #2894, opened 2020-04-20
Fix lazy loads: with list args

lucthev:
If `object_from_id` returns a lazy value, and a field has an auto-loaded list argument, the resolved list values are `GraphQL::Execution::Lazy` objects rather than the underlying application objects.


rmosolgo:
Awesome, thanks for the fix!

PR #2896, opened 2020-04-22
Handle base64 argument errors in the encoder

coding-chimp:
The new pagination system crashes when it's using the default cursor encoder and receives an invalid cursor. This was already fixed in #1855 for the relay connections. I just moved the error handling introduced there to the encoder. Now it can be used by both connection types.

I would say that it probably also makes more sense to be there because the encoder is configurable and different encoders can throw different errors. Handling all possible errors in the connections could become messy.

### DIFF
--- a/lib/graphql/schema/base_64_encoder.rb
+++ b/lib/graphql/schema/base_64_encoder.rb
@@ -14,7 +14,7 @@ module GraphQL
         # urlsafe_decode64 is for forward compatibility
         Base64Bp.urlsafe_decode64(encoded_text)
       rescue ArgumentError
-        raise GraphQL::ExecutionError, "Invalid cursor: #{encoded_text.inspect}"
+        raise GraphQL::ExecutionError, "Invalid input: #{encoded_text.inspect}"
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/graphql/schema/base_64_encoder.rb
```diff
@@ -13,6 +13,8 @@ def self.encode(unencoded_text, nonce: false)
       def self.decode(encoded_text, nonce: false)
         # urlsafe_decode64 is for forward compatibility
         Base64Bp.urlsafe_decode64(encoded_text)
+      rescue ArgumentError
+        raise GraphQL::ExecutionError, "Invalid cursor: #{encoded_text.inspect}"
```
rmosolgo, line 17: This code can also be used for base64-encoded IDs, how about changing the text a bit, eg

```suggestion
        raise GraphQL::ExecutionError, "Invalid input: #{encoded_text.inspect}"
```

?

rmosolgo:
I love the change! I just have one suggestion about the messaging.

lib/graphql/schema/base_64_encoder.rb
coding-chimp: Makes total sense, done! 🙂

rmosolgo:

PR #2900, opened 2020-04-22
Reduce stack depths of deeply nested blocks

jhawthorn:
graphql-ruby tends to end up with very large stacks of things happening recursively. This PR attempts to, wherever possible, avoid pushing extra stack frames from blocks.

Move of these changes follow the pattern of instead doing:
``` ruby
def each
  @array.each { |i| yield(i) }
```

We capture the block and pass it through.

``` ruby
def each(&block)
  @array.each(&block)
end
```

Which avoids re-visiting our `each` in the stack.

### DIFF
--- a/lib/graphql/compatibility/execution_specification/specification_schema.rb
+++ b/lib/graphql/compatibility/execution_specification/specification_schema.rb
@@ -43,8 +43,8 @@ module GraphQL
             @storage = storage
           end
 
-          def each
-            @storage.each { |i| yield(i) }
+          def each(&block)
+            @storage.each(&block)
           end
         end
 
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -412,7 +412,7 @@ module GraphQL
         # @param eager [Boolean] Set to `true` for mutation root fields only
         # @param trace [Boolean] If `false`, don't wrap this with field tracing
         # @return [GraphQL::Execution::Lazy, Object] If loading `object` will be deferred, it's a wrapper over it.
-        def after_lazy(lazy_obj, owner:, field:, path:, scoped_context:, owner_object:, arguments:, eager: false, trace: true)
+        def after_lazy(lazy_obj, owner:, field:, path:, scoped_context:, owner_object:, arguments:, eager: false, trace: true, &block)
           @interpreter_context[:current_object] = owner_object
           @interpreter_context[:current_arguments] = arguments
           @interpreter_context[:current_path] = path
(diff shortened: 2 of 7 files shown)


eapache:
What's the value in reducing stack depths? IIRC capturing blocks in ruby is fairly slow, so it's not obvious to me up front if these changes would actually perform any better.

jhawthorn:
I'm mostly interested in cleaner stacktraces, flamegraphs, etc. It's possible that having a smaller stack will make raising exceptions or anything dealing with caller/caller_locations faster, but that's just an added bonus.

> IIRC capturing blocks in ruby is fairly slow

I don't think this is true (unless you mean vs not yielding at all). Also obviously the places I add short circuit are going to be faster. However I don't think the performance of this will be relevant in either direction.

```
require "benchmark/ips"

def a
  yield
end

def with_yield
  a { yield }
end

def with_capture(&block)
  a(&block)
end

Benchmark.ips do |x|
  x.report "yield", "with_yield{ 1 }"
  x.report "capture", "with_capture{ 1 }"
end
```

```
               yield   370.695k i/100ms
             capture   377.114k i/100ms
               yield     12.601M (± 4.2%) i/s -     63.018M in   5.011469s
             capture     12.810M (± 6.3%) i/s -     63.732M in   5.006789s
```

rmosolgo:
Wow, I was using `yield` across the board because I also thought there was a performance consideration. (I even used to use `Proc.new` in some cases for the same reason, but I think those are gone now.) But it looks like that was fixed a while back (Oct 2017): https://bugs.ruby-lang.org/issues/14045

swalkinshaw:
I think it's more of a trade off when the method doesn't *always* yield so in those cases it can be better to conditionally avoid instantiating a proc `(via &block)`. But in these cases, they are always yielding so it's better to do this?

rmosolgo:
And even when the method _doesn't_ yield, the overhead of `&block` was apparently greatly decreased in recent Ruby versions. Take a benchmark like this one:

```ruby 
require "benchmark/ips"

# this method ignores the given block
def unhandled_block
  nil
end

# this method captures the given block but does nothing with it
def captured_block(&block)
  nil
end

Benchmark.ips do |x|
  x.report("unhandled block") { unhandled_block { x } }
  x.report("captured block") { captured_block { x } }
  x.compare!
end
```

And try it on an old Ruby: 

```
~/code/graphql-ruby $ rbenv local 2.3.2
~/code/graphql-ruby $ ruby --version
ruby 2.3.2p217 (2016-11-15 revision 56796) [x86_64-darwin18]
~/code/graphql-ruby $ ruby block_bench.rb
     unhandled block   269.519k i/100ms
      captured block   113.676k i/100ms
     unhandled block      9.680M (±13.2%) i/s -     47.166M in   5.000095s
      captured block      1.470M (±22.9%) i/s -      6.821M in   5.054059s

Comparison:
     unhandled block:  9679620.7 i/s
      captured block:  1470072.6 i/s - 6.58x  slower
```

then a new one: 

```
~/code/graphql-ruby $ rbenv local 2.7.0
~/code/graphql-ruby $ ruby --version
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin18]
~/code/graphql-ruby $ ruby block_bench.rb
     unhandled block   260.869k i/100ms
      captured block   230.258k i/100ms
     unhandled block     13.992M (±16.7%) i/s -     67.565M in   5.014021s
      captured block     10.479M (±10.8%) i/s -     51.808M in   5.027409s

Comparison:
     unhandled block: 13992035.8 i/s
      captured block: 10478736.7 i/s - 1.34x  slower
```

(Previously: 6.5x slower, now: 1.3x slower)

PR #2905, opened 2020-04-24
Fix deprecation on Schema.to_json parameters

volmer:
In Ruby 2.7, `Schema.to_json` prints the following warning when called
with keyword arguments.

```
graphql-ruby/lib/graphql/schema.rb:874: warning: Using the last argument
as keyword parameters is deprecated; maybe ** should be added to the
call
```

I added `**` to args to fix the deprecation.


rmosolgo:
Thanks for the fix!

Issue #2907, opened 2020-04-25
Mismatch in RootType documentation


nvh0412:
**Describe the bug**

The GraphQL query didn't match with Root types in example https://graphql-ruby.org/schema/root_types.html

![Screen Shot 2020-04-25 at 22 42 30](https://user-images.githubusercontent.com/1973623/80284039-0fc12c80-8746-11ea-9acc-e4b4b793a8f9.png)

My expectation is 

```ruby
class Types::QueryType < GraphQL::Schema::Object
  field :posts, [PostType], 'Returns all blog posts', null: false
end
```

will be matched

```
query posts {
  # `Query.posts`
  posts { ... }
}
```

**Versions**

`graphql` version: None
`rails` (or other framework): None
other applicable versions (`graphql-batch`, etc)

rmosolgo:
Fixed by #2908

PR #2908, opened 2020-04-26
Update example for root_types documentation

nvh0412:
Update documentation of RootTypes https://github.com/rmosolgo/graphql-ruby/issues/2907

Screen Shot 2020-04-26 at 09 50 31


rmosolgo:

PR #2913, opened 2020-04-28
update docs to suggest _not_ using the async activejob adapter

modosc:
it took me a week to figure out that the `async` adapter will not work correctly from background jobs or the rails console :/


rmosolgo:
😖 Thanks for improving the docs!

PR #2919, opened 2020-05-04
Add with_alias: option to Lookahead#selection and Lookahead#selects?

DmitryTsepelev:
With this change it becomes possible to use `Lookahead#selection` and `Lookahead#selects?` to find a nested selection by [alias](https://spec.graphql.org/June2018/#sec-Field-Alias) name. When `with_alias` is true it will try to find selection by field name as usual and then check for alias name. In order to keep things backward–compatible, `with_alias` is false by default, but it might make sense to change it in future. 

For instance, imagine that you have a following query, than can find a bird species by its name:

```graphql
query {
  gull: findBirdSpecies(byName: "Laughing Gull") {
    name
  }

  tanager: findBirdSpecies(byName: "Scarlet Tanager") {
    name
  }
}
```

In this case, you can get the lookahead for each selection in a following way (while searching by field name makes getting the `tanager` selection impossible):

```ruby
def find_bird_species(by_name:, lookahead:)
  if lookahead.selects?("gull")
    # ...
  else
    # ...
  end

  if lookahead.selects?("tanager")
    # ...
  else
    # ...
  end
end
```

If you wonder what was a real use case for it, I tried to use lookahead for building cache keys while working on the [graphql-ruby-fragment_cache](https://github.com/DmitryTsepelev/graphql-ruby-fragment_cache) gem. The idea was that we can use `context.namespace(:interpreter)[:current_path]` to [build](https://github.com/DmitryTsepelev/graphql-ruby-fragment_cache/blob/master/lib/graphql/fragment_cache/cache_key_builder.rb#L127) a list of fields with their arguments, but it turned out that path can contain alias names, while lookahead has no idea about their existence 🙂We've ended up with a monkey–patch, so I decided to backport it.


rmosolgo:
Sorry I haven't left a proper review for this yet. I've been trying to think of how this new suggestion meshes with the use of lookahead which I had in mind. 

In short, I use lookahead to optimize some external service calls. In a spot like this: 

```ruby 
{
  f1: f { a } 
  f2: f { b } 
}
```

I want to know that _both_ `a` and `b` are selected on `f { ... }`, even though they're selected under different aliases. 

So, in that approach, `selection("f")` would return a lookahead including _both_ `a` and `b`. But, you might use an `alias:` option to filter those results. 

Does that match what you had in mind here?

DmitryTsepelev:
> So, in that approach, selection("f") would return a lookahead including both a and b. But, you might use an alias: option to filter those results.

This is how it works now and it will keep working (I've pushed a spec for that). The problem with this approach is that it's impossible to get these selections by alias name without specifying field name.

When query contains alias, `context.namespace(:interpreter)[:current_path]` does have alias without a field name, so it's impossible to call `#selection` with a field name and then filter by alias (i.e., in the query above `current_path` for `b` inside `f2` will be `[f2 b]`, and `.selection("f2")` will give `NULL_LOOKAHEAD`).

`with_alias` option allows `#selection` to check aliases as well as field names in cases when it makes sense, otherwise the existing behaviour is kept.

lib/graphql/execution/lookahead.rb
```diff
@@ -88,26 +91,24 @@ def selected?
...
-          if next_nodes.any?
-            Lookahead.new(query: @query, ast_nodes: next_nodes, field: next_field_defn, owner_type: selected_type)
-          else
-            NULL_LOOKAHEAD
-          end
-        else
-          NULL_LOOKAHEAD
-        end
+        arguments = @query.arguments_for(alias_node, next_field_defn)
+        arguments = arguments.is_a?(::GraphQL::Execution::Interpreter::Arguments) ? arguments.keyword_arguments : arguments
```
rmosolgo, line 110: It looks like, when `with_alias: true` is given, it ignores the provided `arguments` and instead, it returns _all_ matches. 

Was this change intentional? I _expect_ the default case (`arguments: nil`) to return all matches, but then any given arguments to be used as a filter for selected lookaheads. Why not leave that behavior in place? 

```suggestion
```

DmitryTsepelev: In case when arguments are `nil` it will return _all_ fields that have the same type as the field with the provided alias, these two lines forced `#lookahead_for_selection` to find exactly one node with the corresponding alias. Without them `selection("f2", with_alias: true)` will return both `f1` and `f2` nodes, which looks confusing too 🙂

Sounds like it would be easier to keep `#selection` method as is and introduce a new one (`#alias_selection`), which will return a Lookahead with a single node matching the alias (because spec guarantees that there cannot be a second alias). I've refactored the code and specs. What do you think of this approach?

rmosolgo:
Sorry for the slow turn around here. I noticed that `lookup_alias_node` was recursive, but I wanted to pull the branch and try it myself. I've added a test that demonstrates that behavior in https://github.com/rmosolgo/graphql-ruby/pull/2919/commits/41a494bc8e7a06d16d0c5847a20c94fe27b5d329. Is that really the right thing? 

I added a test that _passes_, but it doesn't seem like the right behavior to me. It seems like `selects_alias?` should only be true if one of the field's _immediate_ selections has the given alias. Or, if you intend to preserve the recursive nature of the method, how about adding an indication to the method name, like `selects_alias_at_all?` or `any_subselections_aliased?`, or something like that?

DmitryTsepelev:
Thank you for the review! I totally agree that this spec should not pass and we should only check children of nodes of the current lookahead. Checking if the _current lookahead's node_ is alias does not look helpful too, so I've removed this line.

DmitryTsepelev:
I've just realized that I should also add support for `FragmentSpread`, will add in the next few days

DmitryTsepelev:
Added support for inline and regular fragments. Looks like failing spec fails in other PRs too 🙂

lib/graphql/execution/lookahead.rb
rmosolgo: Aaaaah, I think I finally understand what this is for. It seems like `arguments:` is being used to filter the selection set down to ones that _match the arguments of the aliased selection_. 

Unfortunately that doesn't cover all possibilities allowed by the spec. For example, this is perfectly valid: 

```graphql 
query {
  user(login: "dhh") { email } 
  dhh: user(login: "dhh") { bio }
}
```

What do you expect `lookahead.alias_selection("dhh").selections.map(&:name)` to return in that case? (I expect `bio`, is that right?) 

I pushed a test for this case in 84c1e135b, how do you think that should behave?

lib/graphql/language/nodes.rb
```diff
@@ -376,6 +380,10 @@ def initialize_node(attributes)
           @alias = attributes[:alias]
         end
 
+        def alias?(val)
+          self.alias == val
+        end
```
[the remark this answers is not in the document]
DmitryTsepelev: Agreed, done

DmitryTsepelev:
I've mulled over it a bit more, and now I feel like we should not fall back to `#selection` when there is no alias (because user always can do it by himself when needed). In this case the last assertion in the test (`lookahead.alias_selection("findBirdSpecies").selections.map(&:name)`) returns empty array, while fallback version returns `isWaterfowl` (because of alias), `otherEgret (because of field name)` and `findBirdSpecies` (because of field name too), and it looks confusing. All these changes with the adopted spec are in the last commit. What do you think?

DmitryTsepelev:
We could also make this fallback behaviour optional (`, fallback_to_selection: true`)

rmosolgo:
Thanks for the update on that. I'm still hung up on the treatment of `arguments:`, where the method accepts an `arguments:` input, but then overwrites it with `query.arguments_for(...)`. 

The _intended_ use case is to scope the lookahead based on arguments. For example, if your backend can use an optimized query when a subset of arguments are used, then you can detect that condition using a lookahead. 

I think alias lookaheads should work the same way with regards to that filtering. I've pushed a test for that filtering specifically in https://github.com/rmosolgo/graphql-ruby/pull/2919/commits/b06d7d52536772aafab54a9109f6387399b93003, what do you think of that behavior?

DmitryTsepelev:
Thanks for the feedback! I've added a check to make sure that arguments (when passed) match field arguments and rebased branch to keep sync with master

rmosolgo:
Hey, thanks for all your work on this improvement and I'm so sorry I let it sit idle for so long. It fell off my radar and I was too intimidated to read back up on your work and review it. Today I finally took the time to review it again, and it looks like everything is in order!

I've updated it with `master` :+1:

PR #2921, opened 2020-05-05
Fix link to error handling section

gustavodegeus:
Internal link was broken. Im pointing to the absolute url.
Let me know there is  a better approach. 👍

### DIFF
--- a/guides/queries/interpreter.md
+++ b/guides/queries/interpreter.md
@@ -84,7 +84,7 @@ The new runtime works with class-based schemas only. Several features are no lon
 
 - `rescue_from`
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
For me this link works fine: 

link

Please describe your issue in more detail if you think more work is required here.

gustavodegeus:
It doesnt work if you try to navigate one the markup code on GH but looks like the goal is to always use the page anyways. 👍 

Screen Shot 2020-05-11 at 11 46 40 AM

rmosolgo:
Yes, that's right -- those `.md` files are the source for graphql-ruby.org and aren't optimized for direct use :+1:

PR #2925, opened 2020-05-07
Fix keyword parameters warning in Ruby 2.7

hasghari:
Fixes the following warning:
```
/Users/hasghari/github/graphql-ruby/lib/graphql/object_type.rb:91: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/hasghari/github/graphql-ruby/lib/graphql/schema/type_membership.rb:22: warning: The called method `initialize' is defined here
```


rmosolgo:
Thanks for the fix!

PR #2934, opened 2020-05-10
Disable method access when loading schema to prevent warnings

### LINKED ISSUE #2933: Reserved name warning loading schema for INPUT_OBJECT fields
**Describe the bug**

When loading a schema via `GraphQL::Schema::Loader` containing an `INPUT_OBJECT` with a field called `values`, the library outputs a warning:

    Unable to define a helper for argument with name 'values' as this is a reserved name. Add `method_access: false` to stop this warning.

It doesn't seem possible to override and disable this warning through `GraphQL::Schema::Loader` (which I'm using via `graphql-client`).

This seems to be a regression after 1.10.6, which was the last silent version.

**Versions**

`graphql` version: 1.10.9
`graphql-client` version: 0.16.0

**Reproducer script**

```ruby
GraphQL::Schema::Loader.load(JSON.parse(<<EOS
{
  "data": {
    "__schema": {
      "types": [
        {
          "kind": "INPUT_OBJECT",
          "name": "ExampleFilter",
          "inputFields": [
            {
              "name": "values",
              "type": {
                "kind": "LIST",
                "ofType": {
                  "kind": "SCALAR",
                  "name": "String"
                }
              }
            }
          ]
        }
      ]
    }
  }
}
EOS
)).graphql_definition
```

**Steps to reproduce**

1. `ruby -Ilib -rgraphql reproducer.rb`

**Expected behavior**

No output - the schema's loaded with no warnings.

**Actual behavior**

Script outputs:

```
Unable to define a helper for argument with name 'values' as this is a reserved name. Add `method_access: false` to stop this warning.
```

**Additional context**

3e75532 (in 1.9.8) hid these method_access warnings initially when loading schemas, but it looks like cf70a75 (in 1.10.7) reintroduced it by removing the `method_access: false` when initialising arguments: https://github.com/rmosolgo/graphql-ruby/commit/cf70a75#diff-b6017af43e3a5db33958a8960449aabfL163

Since 1.10.7, the warning has reappeared. 1.10.6 is fine.

domcleal:
When loading schemas, disable method name access to prevent any warnings
that occur when method names already exist.

Re-applies 3e75532 after cf70a75 reverted the change.

The existing spec has been updated to trigger the warning which is
emitted only when the GraphQL classes are built.

Fixes #2933

### DIFF
--- a/lib/graphql/schema/loader.rb
+++ b/lib/graphql/schema/loader.rb
@@ -171,6 +171,7 @@ module GraphQL
               type: type_resolver.call(arg["type"]),
               description: arg["description"],
               required: false,
+              method_access: false,
             }
 
             if arg["defaultValue"]
--- a/spec/graphql/schema/loader_spec.rb
+++ b/spec/graphql/schema/loader_spec.rb
@@ -336,7 +336,7 @@ describe GraphQL::Schema::Loader do
               ]
             }
           }
-        })
+        }).graphql_definition
       end
     end
 


rmosolgo:
Thanks for the fix!

PR #2944, opened 2020-05-15
Fix unsubscribing and error handling for Ably

jscheid:
Hi Robert,

We've noticed that Ably sometimes raises [error 90010](https://help.ably.io/error/90010). I've investigated and found two related issues:

1. Subscriptions don't get cleaned up properly, causing this error which, afaict, effectively means that only the first 200 subscriptions work as expected.
2. Errors raised by Ably don't get propagated to the observer meaning they might go unnoticed.

See the two commit messages for more details.

Assuming you agree with these changes, could I ask you to roll a release with this ASAP?

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -32,6 +32,7 @@
     "typescript": "^3.7.5"
   },
   "prettier": {
-    "semi": false
+    "semi": false,
+    "trailingComma": "none"
   }
 }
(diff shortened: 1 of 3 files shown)


rmosolgo:
This is awesome, thanks so much for the fixes!

rmosolgo:
🚢  in graphql-ruby-client 1.7.9!

jscheid:
Thanks for the quick turnaround!

PR #2954, opened 2020-05-19
Improve object generator

### LINKED ISSUE #2951: Scaffold Generator for DB schema

When I write RESTful endpoints, I can use `rails g scaffold <model>` or `rails g resource <model>` to automatically create fetchers/mutators for the resource. But when I write GraphQL endpoints, I have to define field readers/accessors in GraphQL schema manually after creating a new model/schema.

I want a generator that creates a graphql schema and generates all field accessors for a given model based on its schema.

Manually creating GraphQL schema every time I make new models/schema

**Additional context**

Please feel free to let me know if this already exists!

tadeusz-niemiec:
Resolves #2951

Added support for existing Models in object generator. If a ActiveRecord's class exists the generator will fetch columns info based on Model's schema.

Example usage:

`rails generate graphql:object User`

You can also add custom fields which don't exist in your schema, example:

`rails generate graphql:object User custom_field:String`

Also added support for "datetime" and "date" field types.

### DIFF
--- a/lib/generators/graphql/object_generator.rb
+++ b/lib/generators/graphql/object_generator.rb
@@ -15,19 +15,72 @@ module Graphql
       desc "Create a GraphQL::ObjectType with the given name and fields"
       source_root File.expand_path('../templates', __FILE__)
 
-      argument :fields,
-        type: :array,
-        default: [],
-        banner: "name:type name:type ...",
-        desc: "Fields for this object (type may be expressed as Ruby or GraphQL)"
+      argument :custom_fields,
+               type: :array,
+               default: [],
+               banner: "name:type name:type ...",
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi! Thanks for the suggestion and sorry I've been slow to review it. The change looks good to me, but could you please add some tests for it, so I can be sure not to break it in the future?

tadeusz-niemiec:
> Hi! Thanks for the suggestion and sorry I've been slow to review it. The change looks good to me, but could you please add some tests for it, so I can be sure not to break it in the future?

Done :)

rmosolgo:

mmahalwy:
@tadeusz-niemiec @rmosolgo btw, wanted to highlight this feature as I came across it today by mistake. This is an absolute game changer. Should highlight in the docs!

PR #2956, opened 2020-05-21
Arguments#each_value should yield ArgumentValues

jturkel:
`GraphQL::Execution::Interpreter::Arguments#each_value` currently yields `[<argument_keyword>, ArgumentValue]` arrays but I believe it was meant to yield `ArgumentValue`s to be consistent with `GraphQL::Query::Arguments#each_value`.


rmosolgo:
Derp, yes, thank you!

PR #2957, opened 2020-05-21
Type specific runtime type errors

rmosolgo:
Sometimes these errors get lumped together in our bug tracker because they have the same names and backtraces. But they're really very different -- I hope to make that clear by making type-specific error classes. (We already do this with internal authorization errors which is 👌 .) 

It should be just fine from a compat standpoint, since they're all subclasses of the original.


rmosolgo:
It seems like the PR build is failing because of a lint error on master.

PR #2958, opened 2020-05-21
Add raw_value to GRAPHQL_RUBY_KEYWORDS

swalkinshaw:
All object types have the method `raw_value` via `GraphQL::Schema::Object.include(HandlesRawValue)` so this should be
considered reserved too.


rmosolgo:
Thanks for the fix -- looks like the build is failing because of a lint error on master, fixed in 018ab22b8

Issue #2846, opened 2020-03-25
Release 1.11


rmosolgo:
I think I want to focus this on some rework of subscriptions, which may include breaking-ish changes. 

- Simplify the system: 
  - [x] Remove need for `extend ...SubscriptionRoot` with Interpreter #2770 
  - ~~Simplify the implementation, too?~~ It'd be great but I have no vision for how it could be improved right now

- Support more flexible execution/delivery
  - [x] #2844 
- Implement improved subscription API 
  - Turns out, this can be done with what we already have. Only the implementations need updates 
  - [x] #2959 ActionCable
  - [x] Document how this works
  - [ ] Pro: Pusher, Ably and Pubnub should use this algorithm, too 
- [x] Default tests to `TESTING_INTERPRETER` #2860 
- [x] Migrate the schema-from-introspection-response code to build a class-based schema 
  - Going into 1.10.x #2876
- [x] Remove global tracer support #2936 

Maybe some other things will come up (or be suggested here)

apuntovanini:
Hi @rmosolgo, currently (if I don't get it terribly wrong), `createRelaySubscriptionHandler` doens't pass an observer to Relay Network layer, resulting in an Uncaught TypeError: Cannot read property 'onNext' of undefined

Offeding lines  
```
received: function (payload) {
  // When we get a response, send the update to `observer`
  var result = payload.result;
  if (result && result.errors) {
      // What kind of error stuff belongs here?
      observer.onError(result.errors);
  }
  else if (result) {
      observer.onNext({ data: result.data });
  }
  if (!payload.more) {
      // Subscription is finished
      observer.onCompleted();
  }
}
```

observer is undefined. Am I doing something wrong? If yes, maybe some additional comments in docs can be helpful (let me know if you accept PR)
Thanks!

rmosolgo:
@apuntovanini , please open a new issue if you'd like to report a bug. See https://graphql-ruby.org/development.html and https://github.com/rmosolgo/graphql-ruby/blob/master/.github/contributing.md for details about contributing to this project.

rmosolgo:
🚢 !

PR #2959, opened 2020-05-21
Add subscription broadcasts

### LINKED ISSUE #2851: Extract queries and subscribers
Currently, two concepts are coupled by necessity in graphql-ruby's subscription system:

- A query + context which are subscribed to, and can be re-evaluated in response to `.trigger`
- A listening process to that query, which receives the updated payload 

However, there are possible cases when you'd want _multiple_ listeners to a single query & result. Public data is the easiest example -- when an event happens, you can be sure that all subscribers can receive the same payload. 

Something like:

```ruby 
class MyEvent < GraphQL::Schema::Subscription
  # Returns a subset of the query context which will be used for:
  # - Persistence -- saved in the DB and reloaded 
  # - De-duping -- any subscribers with the same query string and `subscribed_context` 
  #   will be updated with the same payload 
  # Be warned, if you group subscribers that ought not to be grouped, you can send unauthorized data to a subscriber!
  # 
  # Default implementation is `context.to_h`.
  # Called during the initial subscription, only. 
  #  
  # @return [Hash] 
  def subscribed_context
    context.slice(:current_organization) # { current_organization: #<Org...> }
  end 

  # Default implementation is `schema.subscriptions.build_id`, a random ID 
  # Override it to de-dup subscriptions in the backend.
  # @return [String]
  def subscription_id 
    "#{self.class.name}:#{context[:current_organization].name}"
  end 
end 
```

rmosolgo:
If lots of people are subscribed to the same event with the same query string, and you can be sure that they should _all receive the same data_, then we should be able to calculate the payload _once_ and distribute it to all subscribers. 

I thought I would need a new API, but I think it's possible using: 

- `Query#fingerprint`, which returns a unique string for each query string + variables 
- `Event#topic`, which includes field name, args, and subscription scope

__TODO__: 

- [x] Update general broadcast docs with opt-in info and considerations 
- [x] Add `broadcastable?` helper for tests? 
- [x] Update implementation-specific docs about considerations 
- [x] Make introspection fields broadcastable by default

Fixes #2851

Part of #2846

### DIFF
--- a/lib/graphql/subscriptions.rb
+++ b/lib/graphql/subscriptions.rb
@@ -79,15 +79,13 @@ module GraphQL
     # `event` was triggered on `object`, and `subscription_id` was subscribed,
     # so it should be updated.
     #
-    # Load `subscription_id`'s GraphQL data, re-evaluate the query, and deliver the result.
-    #
-    # This is where a queue may be inserted to push updates in the background.
+    # Load `subscription_id`'s GraphQL data, re-evaluate the query and return the result.
     #
     # @param subscription_id [String]
     # @param event [GraphQL::Subscriptions::Event] The event which was triggered
     # @param object [Object] The value for the subscription field
-    # @return [void]
-    def execute(subscription_id, event, object)
+    # @return [GraphQL::Query::Result]
+    def execute_update(subscription_id, event, object)
       # Lookup the saved data for this subscription
       query_data = read_subscription(subscription_id)
       if query_data.nil?
@@ -118,6 +116,15 @@ module GraphQL
       nil
     end
 
+    # Run the update query for this subscription and deliver it
+    # @see {#execute_update}
+    # @see {#deliver}
+    # @return [void]
+    def execute(subscription_id, event, object)
+      res = execute_update(subscription_id, event, object)
+      deliver(subscription_id, res)
+    end
+
     # Event `event` occurred on `object`,
     # Update all subscribers.
     # @param event [Subscriptions::Event]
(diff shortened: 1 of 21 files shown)


lib/graphql/subscriptions.rb
```diff
@@ -118,14 +116,23 @@ def execute(subscription_id, event, object)
...
     end
 
+    # Run the update query for this subscription and deliver it
+    # @see {#execute_update}
+    # @see {#deliver}
+    # @return [void]
+    def execute(subscription_id, event, object)
+      res = execute_update(subscription_id, event, object)
+      deliver(subscription_id, res)
+    end
```
rmosolgo, line 126: This actually _unbreaks_ the API for 1.11: old code will still work; `execute_update` _can be_ called by implementations that want to separate execution and delivery.

rmosolgo:
Ok, I realized a problem why this wont work as default behavior. There are some `context`-derived fields that aren't covered by `subscription_scope`. For example:

```graphql
subscription {
  commitPushed(repositoryId: "abcd123") {
    commit {
      sha 
    }
    repository {
      name 
      viewerCanDelete
    }
  }
}
```

In this case, how could `commitPushed` _ever_ know whether `Repository.viewerCanDelete` was selected or not? It has _nothing_ to do with the subscription, but in the current implementation, _every_ subscriber would receive whatever the first subscriber happened to get for that field. 

So, I think I need a more robust solution. Something like: 

- Add an opt-in analyzer which looks at `subscription?` operations
- It checks to see if all fields are `.broadcast? -> true`, if so, it sets a flag in context 
- If that flag is truthy, _then_ we use the broadcast behavior. Otherwise, we fall back to the evaluate-for-every-subscriber behavior.

PR #2969, opened 2020-05-28
Add Scout transaction naming

### LINKED ISSUE #2968: ScoutTracing does not set the transaction name like NewRelicTracing

We switched from New Relic to Scout for APM, and noticed that the `NewRelicTracing` sets the transaction name when it's a single query, but Scout doesn't. The code for this is almost identical (just calling the Scout equivalent of set transaction name):

```ruby
# NewRelicTracing (current)
NewRelic::Agent.set_transaction_name("GraphQL/#{op_type}.#{op_name}")

# ScoutTracing (proposed)
ScoutApm::Transaction.rename("GraphQL/#{op_type}.#{op_name}")
```

`GraphQL::Tracing::ScoutTracing` sets the transaction name for `execute_query` just like `GraphQL::Tracing::NewRelicTracing` does.

We made a subclass of `GraphQL::Tracing::ScoutTracing` to do this, but I think it's a feature of the tracer itself.

**Additional context**

I am happy to PR this to add to `GraphQL::Tracing::ScoutTracing`. It's a very simple change.

bbugh:
This adds code to `GraphQL::Tracing::ScoutTracing` to write the transaction name for single query execution.

Fixes #2968 

---

This seems like an opportunity to refactor a bit since the 16 line method is identical to `GraphQL::Tracing::NewRelicTracing` except for one line. I didn't want to make this change without checking first, though.

It seems like a simple improvement would be to move the code that determines the transaction name into `PlatformTracing` as `transaction_name`, like:

```ruby
# platform_tracing.rb

# Get the transaction name based on the operation type and name
def transaction_name
  query = data[:query]
  selected_op = query.selected_operation
  if selected_op
    op_type = selected_op.operation_type
    op_name = selected_op.name || "anonymous"
  else
    op_type = "query"
    op_name = "anonymous"
  end
  "GraphQL/#{op_type}.#{op_name}"
end
```

And then in Scout/NewRelic it can do this:

```ruby
  if key == "execute_query"
    set_this_txn_name = data[:query].context[:set_scout_transaction_name]
    if set_this_txn_name == true || (set_this_txn_name.nil? && @set_transaction_name)
      ScoutApm::Transaction.rename(transaction_name)
    end
  end
```

Thoughts?

### DIFF
--- a/lib/graphql/tracing/platform_tracing.rb
+++ b/lib/graphql/tracing/platform_tracing.rb
@@ -102,6 +102,21 @@ module GraphQL
         schema_defn.tracer(tracer)
       end
 
+      protected
+
+      # Get the transaction name based on the operation type and name
+      def transaction_name(query)
+        selected_op = query.selected_operation
+        if selected_op
+          op_type = selected_op.operation_type
+          op_name = selected_op.name || "anonymous"
+        else
+          op_type = "query"
+          op_name = "anonymous"
+        end
+        "GraphQL/#{op_type}.#{op_name}"
+      end
+
       private
       attr_reader :options
 
--- a/lib/graphql/tracing/platform_tracing.rb
+++ b/lib/graphql/tracing/platform_tracing.rb
@@ -102,7 +102,7 @@ module GraphQL
         schema_defn.tracer(tracer)
       end
 
-      protected
+      private
 
       # Get the transaction name based on the operation type and name
       def transaction_name(query)
@@ -117,7 +117,6 @@ module GraphQL
         "GraphQL/#{op_type}.#{op_name}"
       end
 
-      private
       attr_reader :options
 
       def platform_key_cache(ctx)
--- a/spec/support/scout_apm.rb
+++ b/spec/support/scout_apm.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-# A stub for the NewRelic agent, so we can make assertions about how it is used
+# A stub for the Scout agent, so we can make assertions about how it is used
 if defined?(ScoutApm)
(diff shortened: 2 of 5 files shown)


rmosolgo:
:+1: To extracting `transaction_name`!

bbugh:
@rmosolgo I added the `transaction_name` to `PlatformTracing` and added some tests for `ScoutTracing`.

spec/support/scout_apm.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+# A stub for the NewRelic agent, so we can make assertions about how it is used
```
rmosolgo, line 2:
```suggestion
# A stub for the Scout agent, so we can make assertions about how it is used
```

lib/graphql/tracing/platform_tracing.rb
```diff
@@ -102,6 +102,21 @@ def self.use(schema_defn, options = {})
         schema_defn.tracer(tracer)
       end
 
+      protected
```
rmosolgo, line 105: Just curious, why `protected` instead of `private`? (I have yet to find a case where I really _want_ `protected`, maybe today's the day 😆 !)

rmosolgo: I went ahead and switched it to private. In general, I prefer the _most_ privacy I can get, so that refactoring is easier and less risky, but if I've misunderstood, please let me know!

bbugh:
Hey @rmosolgo sorry about that, I got sick and fell behind on work load. Changing `protected` to `private` should be fine. Thanks for merging and all of your hard work on this library!

PR #2973, opened 2020-06-07
Use Time objects instead of DateTime for ISO8601DateTime scalar

bdewater:
As [documented by Ruby](https://ruby-doc.org/stdlib-2.3.0/libdoc/date/rdoc/DateTime.html): "Almost certainly you'll want to use Time since your app is probably dealing with current dates and times."

The tests conflated timezones with UTC offsets - the difference is subtle but important. There's plenty of sources (e.g. [Stack Overflow](https://stackoverflow.com/tags/timezone/info)) on the internet explaining this.

Also added a small speedup by using `to_time` to coerce a Date object instead of writing out and parsing a string.

### DIFF
--- a/lib/graphql/types/iso_8601_date.rb
+++ b/lib/graphql/types/iso_8601_date.rb
@@ -15,7 +15,7 @@ module GraphQL
     class ISO8601Date < GraphQL::Schema::Scalar
       description "An ISO 8601-encoded date"
 
-      # @param value [Date,DateTime,String]
+      # @param value [Date,Time,DateTime,String]
       # @return [String]
       def self.coerce_result(value, _ctx)
         Date.parse(value.to_s).iso8601
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -1,7 +1,9 @@
+require 'time'
+
 # frozen_string_literal: true
 module GraphQL
   module Types
-    # This scalar takes `DateTime`s and transmits them as strings,
+    # This scalar takes `Time`s and transmits them as strings,
     # using ISO 8601 format.
     #
     # Use it for fields or arguments as follows:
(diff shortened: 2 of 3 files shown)


rmosolgo:
Hey, thanks for the contribution! I hadn't seen that bit in Ruby's docs before. 

Before I merge this, I'd like to understand the practical implications for people who are already using this code. It looks like the main difference is, people using `ISO8601DateTime` will now receive instances of `Time` in their application instead of `DateTime`, is that right? Is there any reason that the `Time` instance won't quack just like the `DateTime` that's currently returned? (Anything we should warn people about in the changelog?) 

In the same vein, if current users shouldn't expect any practical changes, then what's the advantage of this change?

bdewater:
There are some minute 😉 differences in public APIs:
```
irb(main):001:0> require 'time'
=> true
irb(main):002:0> DateTime.methods(false) - Time.methods(false)
=> [:jd, :ordinal, :civil, :commercial, :new, :rfc3339, :jisx0301, :_strptime]
irb(main):003:0> DateTime.instance_methods(false) - Time.instance_methods(false)
=> [:second_fraction, :new_offset, :rfc3339, :jisx0301, :minute, :second, :sec_fraction, :offset]
```
Some of those are available with a slightly different name though (e.g. `second` -> `sec` as shown in the diff). I believe that when it comes to just storing these in a database there should be no difference at all. 

When it comes to calculation the API is the same but there are some subtle differences around timezones/DST results, I've put these in a gist: https://gist.github.com/bdewater/4c483beb7f521841a915f62c56fdcea4. 

Rails offers the [TimeWithZone](https://api.rubyonrails.org/v6.0.2.1/classes/ActiveSupport/TimeWithZone.html) superset of Time to work around the UTC/`ENV["TZ"]` limitations of Ruby. Ruby 2.6 offers [improvements](https://ruby-doc.org/core-2.6/Time.html#class-Time-label-Timezone+argument) in this area but I haven't played with those myself. Overall IMO Time is preferable because it handles all this better than DateTime does.

rmosolgo:
Thanks for sharing those details, TIL 👨‍🎓 !

PR #2975, opened 2020-06-08
fix: Inherit Schema.error_handler configuration from base class

rylanc:
Currently, subclassing a `Schema` doesn't inherit the base class's error handler. This means that configuration like `use use GraphQL::Execution::Errors` isn't passed down to subclasses. This PR is an attempt to fix that.


rmosolgo:
Thanks for the fix!

PR #2976, opened 2020-06-08
Allow resolve_type to optionally return resolved object

swalkinshaw:
Often abstract types in GraphQL are be backed by "wrapped" domain objects such as unions, ADTs, etc. Currently `resolve_type` must return a GraphQL type only even though the wrapped domain object ideally should be unwrapped at this point as well instead of passing through to the next resolver.

This allows abstract type's `resolve_type` to return an optional "resolved"/unwrapped object in addition the resolved type. To optionally return a new object value, an array can be returned as `[type, object]`.

This is backwards compatible as well and the existing validation still applies to verify the first item in the array is a GraphQL type.

@rmosolgo this is a proposal right now to see what you think. If you think it's useful, I need to update documentation.

Also I'm not 100% sure how this works with lazy types yet. It doesn't break them, but I don't think I supported the `[lazy_type, object]` case yet. Should that case be supported?

cc @gmalette

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -306,19 +306,21 @@ module GraphQL
             write_in_response(path, r)
             r
           when "UNION", "INTERFACE"
-            resolved_type_or_lazy = resolve_type(type, value, path)
+            resolved_type_or_lazy, resolved_value = resolve_type(type, value, path)
+            resolved_value ||= value
+
             after_lazy(resolved_type_or_lazy, owner: type, path: path, scoped_context: context.scoped_context, field: field, owner_object: owner_object, arguments: arguments, trace: false) do |resolved_type|
               possible_types = query.possible_types(type)
 
               if !possible_types.include?(resolved_type)
                 parent_type = field.owner
                 err_class = type::UnresolvedTypeError
-                type_error = err_class.new(value, field, parent_type, resolved_type, possible_types)
+                type_error = err_class.new(resolved_value, field, parent_type, resolved_type, possible_types)
                 schema.type_error(type_error, context)
                 write_in_response(path, nil)
                 nil
               else
-                continue_field(path, value, field, resolved_type, ast_node, next_selections, is_non_null, owner_object, arguments)
+                continue_field(path, resolved_value, field, resolved_type, ast_node, next_selections, is_non_null, owner_object, arguments)
               end
             end
           when "OBJECT"
@@ -546,7 +548,7 @@ module GraphQL
 
         def resolve_type(type, value, path)
           trace_payload = { context: context, type: type, object: value, path: path }
-          resolved_type = query.trace("resolve_type", trace_payload) do
+          resolved_type, resolved_value = query.trace("resolve_type", trace_payload) do
             query.resolve_type(type, value)
           end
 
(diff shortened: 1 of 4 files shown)


lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -306,19 +306,21 @@ def continue_field(path, value, field, type, ast_node, next_selections, is_non_n
             write_in_response(path, r)
             r
           when "UNION", "INTERFACE"
-            resolved_type_or_lazy = resolve_type(type, value, path)
+            resolved_type_or_lazy, resolved_value = resolve_type(type, value, path)
+            resolved_value ||= value
+
```
swalkinshaw, line 311: @rmosolgo my other option was to add a `cast` (name tbd) method on interface and unions which would be called separately after `resolve_type`. It would probably take the resolved type, value, and maybe ctx. The default methods would just pass through `value`.

rmosolgo: I don't have a preference on the approach, bot this and something like `cast_object(type, obj, ctx)` would be fine with me.

swalkinshaw: @gmalette do you have a preference? I assume in some cases the unwrapping might involve logic which mirrors the type resolving logic so it might make more sense to keep them together. With a `cast` method you might have to duplicate some logic.

gmalette: Pretty much your arguments: using `cast` would likely require duplicated logic and would introduce new affordances for error. The naming of this method is unfortunate if it also maps the value in addition to the type, but I can live with that 🙂

swalkinshaw:
Added docs to the interface face which the union one references as well. Looks like the failing check is allowed to fail.

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: We _could_ get a twofer and add a new method, like: 

```ruby
# Implement this method to return a two-member array containing: 
# - The concrete type to use for `abstract_type` in this case 
# - A "casted" object, if `object` should be specialized in this case. 
# @return [(Class<GraphQL::Schema::Object>, Object)]
def resolve_type_and_cast_object(abstract_type, object, ctx)
  object_type = resolve_type(abstract_type, object, ctx)
  resolved_obj = cast_object(abstract_type, object_type, object, ctx)
  [object_type, resolved_obj]
end 

# Optionally override this is `obj` should be modified in this context
def cast_object(abs_t, obj_t, obj, ctx)
  obj 
end 
```

(Maybe they'd be methods on Interface/Union and have different arguments). In something like that, you get to pick _new_ method names, and keep the existing method signature the same. In your case, you could override `def resolve_type_and_cast_object`, but existing code would stay just the same. (And for better or worse, `cast_object` could exist as its own API. TBH, you could skip `def cast_object` and just return `[resolved_type, obj]`, and anyone who wants to cast can override `resolve_type_and_cast_object`.)

What do you think, worth it to pick a new method?

rmosolgo: > slightly more explicit method name?

Yeah, that'd be it 🤷

rmosolgo:
🎉 🤘 !

PR #2978, opened 2020-06-09
Avoid recomputing arguments for each object instance

jturkel:
Profiling interpreter execution of one our particularly painful queries that returns a large number of objects showed 36% of the memory used during execution was allocated in `GraphQL::Schema::Member::HasArguments.coerce_arguments`. This PR takes a crack at only applying argument coercion once to avoid these repeated allocations. There are two cases it does not cover:

1. `GraphQL::Schema::Object`'s can declare method prepare hooks which have to operate in the context of a `GraphQL::Schema::Object` instance so they can't be cached. Perhaps in the future this could be changed to call static methods on `GraphQL::Schema::Object` classes to make the results cacheable?
2. The processing of field extra's mutates the `GraphQL::Execution::Interpreter::Arguments` data structure to include the extras which will vary between object instances. We could change this without changing the resolver contract but this will change data passed to field tracers, data passed to a schema's `rescue_from` handler, and the `keyword_arguments` attribute of the `argument_details` field extra. I'm not sure if any of those will be problematic.


lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -460,7 +458,13 @@ def after_lazy(lazy_obj, owner:, field:, path:, scoped_context:, owner_object:,
         end
 
         def arguments(graphql_object, arg_owner, ast_node)
-          query.arguments_for(ast_node, arg_owner, parent_object: graphql_object)
+          # Field extras currently the arguments data structure so only cache them in the context of a given graphql object
```
swalkinshaw, line 461: Missing a word here.. "currently mutate" maybe?

jturkel: Oops. Definitely missing a word there.

PR #2983, opened 2020-06-10
Fixed NoMethodError when using extras in a subscription field 

### LINKED ISSUE #2982: NoMethodError on stringify_args when using subcriptions and extras
**Describe the bug**

When using new interpreter + `extras` + subscriptions, I get a `NoMethodError: undefined method '[]' for nil:NilClass` exception when triggering a subscription update

**Versions**

`graphql` version: 1.9.20
`rails` (or other framework): Rails 5.2.4.3
`graphql-batch`: 0.4.1

**Steps to reproduce**

Trigger a subscription update on a subscription field that has `extras`

**Expected behavior**

Update succeeds

**Actual behavior**

```
NoMethodError: undefined method `[]' for nil:NilClass
from /usr/local/var/rbenv/versions/2.5.5/lib/ruby/gems/2.5.0/gems/graphql-1.9.20/lib/graphql/subscriptions/event.rb:74:in `block in stringify_args'
```

**Additional context**

It seems like adding a nil check in `lib/graphql/subscriptions/event.rb` line 74 would fix the problem 
```ruby
next_args[normalized_arg_name] = stringify_args(arg_defn[1].type, v) if !arg_defn.nil?
```
If `arg_defn` is nil is because that argument wasn't one of the field arguments (for example when using `:query` or `:path` as extras.
However, I'm not completely sure that's the right solution, maybe the arguments should be filtered before they got to the event?

g-gaston:
This resolves #2982 by filtering out the arguments before creating the event

If we want to be explicit with the tests, I can create a new field with extras in the subscription and duplicate one test for subscribe and another one for update using that new field

### DIFF
--- a/lib/graphql/subscriptions/subscription_root.rb
+++ b/lib/graphql/subscriptions/subscription_root.rb
@@ -40,7 +40,7 @@ module GraphQL
             # for the backend to register:
             events << Subscriptions::Event.new(
               name: field.name,
-              arguments: arguments,
+              arguments: arguments_without_field_extras(arguments: arguments),
               context: context,
               field: field,
             )
@@ -48,7 +48,7 @@ module GraphQL
             value
           elsif context.query.subscription_topic == Subscriptions::Event.serialize(
               field.name,
-              arguments,
+              arguments_without_field_extras(arguments: arguments),
               field,
               scope: (field.subscription_scope ? context[field.subscription_scope] : nil),
             )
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for the fix here! I see you opened this PR into 1.9.x. Do you want me to release to rubygems or are you pulling the code from github? 

(In the meantime, I'm going to port it to master, too.)

g-gaston:
@rmosolgo if you could release to rubygems that would be great!

g-gaston:
@rmosolgo could this be released as `1.9.21`, please?

rmosolgo:
💁‍♂️ ✨ https://rubygems.org/gems/graphql/versions/1.9.21 ! Thanks again for the fix.

PR #2985, opened 2020-06-11
Only check used input arguments visibility

grcooper:
This line:

`visible_arguments_map = warden.arguments(self).reduce({}) { |m, f| m[f.name] = f; m}`

Was straightforward, but was also making visibility calls to arguments that we did not need to check.

I have sacrificed a small bit of readability here in order to reduce the number of `visible?` calls being made in warden.

There are three steps to this process to make sure it remained the same and didn't break tests:

1. Inject all required arguments as `nil` that were missing from input. This is to satisfy this test:
https://github.com/rmosolgo/graphql-ruby/blob/39b78e5e46b654d5fc9b170025e3b690e463b9e8/spec/graphql/query/executor_spec.rb#L293
2. Check to see if the input/required args are visible on the object. Satisfies this test:
https://github.com/rmosolgo/graphql-ruby/blob/39b78e5e46b654d5fc9b170025e3b690e463b9e8/spec/graphql/query/executor_spec.rb#L313
3. Confirm that the value being provided is valid. Satisfies this test: https://github.com/rmosolgo/graphql-ruby/pull/2985/files#diff-3f549290bd93783c3126a4eeded156ceR336

### DIFF
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -182,21 +182,25 @@ module GraphQL
             end
           end
 
-          visible_arguments_map = warden.arguments(self).reduce({}) { |m, f| m[f.name] = f; m}
-
-          # Items in the input that are unexpected
-          input.each do |name, value|
-            if visible_arguments_map[name].nil?
-              result.add_problem("Field is not defined on #{self.graphql_name}", [name])
+          # Inject missing required arguments
+          required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
+            if !input.key?(argument_name) && argument.type.non_null? && warden.get_argument(self, argument_name)
+              m[argument_name] = nil
             end
+
+            m
           end
 
-          # Items in the input that are expected, but have invalid values
-          visible_arguments_map.map do |name, argument|
-            argument_result = argument.type.validate_input(input[name], ctx)
-            if !argument_result.valid?
-              result.merge_result!(name, argument_result)
+          input.to_h.merge(required_inputs).each do |argument_name, value|
+            argument = warden.get_argument(self, argument_name)
+            # Items in the input that are unexpected
+            unless argument
+              result.add_problem("Field is not defined on #{self.graphql_name}", [argument_name])
+              next
             end
+            # Items in the input that are expected, but have invalid values
+            argument_result = argument.type.validate_input(value, ctx)
+            result.merge_result!(argument_name, argument_result) unless argument_result.valid?
           end
 
           result
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -169,7 +169,7 @@ module GraphQL
           # We're not actually _using_ the coerced result, we're just
           # using these methods to make sure that the object will
           # behave like a hash below, when we call `each` on it.
-          begin
+          input = begin
             input.to_h
           rescue
             begin
@@ -183,7 +183,7 @@ module GraphQL
           end
 
           # Inject missing required arguments
-          required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
+          missing_required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
             if !input.key?(argument_name) && argument.type.non_null? && warden.get_argument(self, argument_name)
               m[argument_name] = nil
             end
@@ -191,7 +191,7 @@ module GraphQL
             m
           end
 
-          input.to_h.merge(required_inputs).each do |argument_name, value|
+          input.merge(missing_required_inputs).each do |argument_name, value|
             argument = warden.get_argument(self, argument_name)
             # Items in the input that are unexpected
             unless argument
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -166,9 +166,6 @@ module GraphQL
             return result
           end
 
-          # We're not actually _using_ the coerced result, we're just
-          # using these methods to make sure that the object will
-          # behave like a hash below, when we call `each` on it.
           input = begin
             input.to_h
           rescue
--- a/lib/graphql/schema/warden.rb
+++ b/lib/graphql/schema/warden.rb
@@ -108,7 +108,7 @@ module GraphQL
 
       # @return [GraphQL::Argument, nil] The argument named `argument_name` on `parent_type`, if it exists and is visible
       def get_argument(parent_type, argument_name)
-        argument = parent_type.arguments[argument_name]
+        argument = parent_type.get_argument(argument_name)
         return argument if argument && visible_argument?(argument)
       end
 
--- a/lib/graphql/schema/warden.rb
+++ b/lib/graphql/schema/warden.rb
@@ -106,6 +106,12 @@ module GraphQL
         @visible_parent_fields[parent_type][field_name]
       end
 
+      # @return [GraphQL::Argument, nil] The argument named `argument_name` on `parent_type`, if it exists and is visible
+      def get_argument(parent_type, argument_name)
+        argument = parent_type.arguments[argument_name]
+        return argument if argument && visible_argument?(argument)
+      end
+
       # @return [Array<GraphQL::BaseType>] The types which may be member of `type_defn`
       def possible_types(type_defn)
         @visible_possible_types ||= read_through { |type_defn|
--- a/spec/graphql/query/executor_spec.rb
+++ b/spec/graphql/query/executor_spec.rb
@@ -329,5 +329,27 @@ describe GraphQL::Query::Executor do
         assert_equal(expected, result.to_h)
       end
     end
+
+    describe "for input objects with nil value for a required input" do
+      let(:variables) { {"input" => [{ "source" => nil }]} }
+      let(:query_string) {%| query Q($input: [DairyProductInput]) { searchDairy(product: $input) { __typename, ... on Cheese { id, source } } } |}
+      it "returns a variable validation error" do
+        expected = {
+          "errors"=>[
+            {
+              "message" => "Variable $input of type [DairyProductInput] was provided invalid value for 0.source (Expected value to not be null)",
+              "locations" => [{ "line" => 1, "column" => 10 }],
+              "extensions" => {
+                "value" => [{ "source" => nil }],
+                "problems" => [
+                  { "path" => [0, "source"], "explanation" => "Expected value to not be null" }
+                ]
+              }
+            }
+          ]
+        }
+        assert_equal(expected, result.to_h)
+      end
+    end
   end
 end
--- a/lib/graphql/schema/member/has_arguments.rb
+++ b/lib/graphql/schema/member/has_arguments.rb
@@ -58,6 +58,20 @@ module GraphQL
           end
         end
 
+        # @return [GraphQL::Schema::Argument, nil] Argument defined on this thing, fetched by name.
+        def get_argument(argument_name)
+          if (a = own_arguments[argument_name])
+            a
+          else
+            for ancestor in ancestors
+              if ancestor.respond_to?(:own_arguments) && a = ancestor.own_arguments[argument_name]
+                return a
+              end
+            end
+            nil
+          end
+        end
+
         # @param new_arg_class [Class] A class to use for building argument definitions
         def argument_class(new_arg_class = nil)
           self.class.argument_class(new_arg_class)


spec/graphql/query/executor_spec.rb
```diff
@@ -317,7 +317,7 @@
               "message" => "Variable $input of type [DairyProductInput] was provided invalid value for 0.foo (Field is not defined on DairyProductInput), 0.source (Expected value to not be null)",
               "locations" => [{ "line" => 1, "column" => 10 }],
               "extensions" => {
-                "value" => [{ "foo" => "bar" }],
+                "value" => [{ "foo" => "bar"}],
```
swalkinshaw, line 320: 🔥

eapache:
> Inject all required arguments as nil that were missing from input.

Is this going to break `.key?` on the input for those arguments?

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
 
-          visible_arguments_map = warden.arguments(self).reduce({}) { |m, f| m[f.name] = f; m}
-
-          # Items in the input that are unexpected
-          input.each do |name, value|
-            if visible_arguments_map[name].nil?
-              result.add_problem("Field is not defined on #{self.graphql_name}", [name])
+          # Inject missing required arguments
+          required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
+            if input[argument_name].nil? && argument.type.non_null? && warden.get_argument(self, argument_name)
```
eapache, line 189: `input[argument_name].nil?` or `!input.key?(argument_name)`

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
+
+            m
           end
 
-          # Items in the input that are expected, but have invalid values
-          visible_arguments_map.map do |name, argument|
-            argument_result = argument.type.validate_input(input[name], ctx)
-            if !argument_result.valid?
-              result.merge_result!(name, argument_result)
+          input.merge(required_inputs).each do |argument_name, value|
```
eapache, line 196: if we merged this the other way around we could build `required_inputs` much more simply?

grcooper:
@eapache Where are we using `.key?` ? `merge` only returns a merged hash, it doesn't mutated `input` so as long as I am not using it in that particular block then we are good?

eapache:
> Where are we using .key? ? merge only returns a merged hash, it doesn't mutated input so as long as I am not using it in that particular block then we are good?

KCCO I made some bad assumptions based on how you worded the PR description.

lib/graphql/schema/input_object.rb
grcooper: Yah this is likely a better way to do it and more explicit on what I am checking. Will update.

lib/graphql/schema/warden.rb
```diff
@@ -106,6 +106,12 @@ def get_field(parent_type, field_name)
         @visible_parent_fields[parent_type][field_name]
       end
 
+      # @return [GraphQL::Argument, nil] The argument named `argument_name` on `parent_type`, if it exists
```
chrisbutcher, line 109:
```suggestion
      # @return [GraphQL::Argument, nil] The argument named `argument_name` on `parent_type`, if it exists and is visible
```

This would explain the `&&` logic

lib/graphql/schema/input_object.rb
grcooper: I think the solution here using select looks something like:

```ruby
required_inputs = self.arguments.select do |argument_name, argument|
  !input.key?(argument_name) && argument.type.non_null? && warden.get_argument(self, argument_name)
end.map { |argument| nil }
```

I don't think this is too much better than the reduce, but can change if necessary.

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
 
-          visible_arguments_map = warden.arguments(self).reduce({}) { |m, f| m[f.name] = f; m}
-
-          # Items in the input that are unexpected
-          input.each do |name, value|
-            if visible_arguments_map[name].nil?
-              result.add_problem("Field is not defined on #{self.graphql_name}", [name])
+          # Inject missing required arguments
+          required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
+            if !input.key?(argument_name) && argument.type.non_null? && warden.get_argument(self, argument_name)
```
eapache, line 189: Will `get_argument` still trigger our logging? 'cause we're still calling it here on keys that are explicitly not in the input.

grcooper: Yes it will still trigger a call to visibility.

There isn't a great way around this with required arguments however, as we need to check if they are visible on the field in order to ensure that a value is being passed in for those arguments. If the argument is not visible then we don't care about it being required.

The only case we will trigger a visibility check and over-report our logging is when there is a required argument that is not supplied by the user and it is visible on the field.

If it is invisible we will still do the visibility check but it will not log.

eapache: > The only case we will trigger a visibility check and over-report our logging is when there is a required argument that is not supplied by the user and it is visible on the field.

These requests are broken anyway, correct, meaning they'll fail with an error for a missing argument?

I think this is probably a solid upstream improvement and good short-term fix for our issue (since IIRC none of the args we care about short-term are required), but the corner case worries me a bit. Let's ship this, but continue to think about a long-term solution.

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
           end
 
-          visible_arguments_map = warden.arguments(self).reduce({}) { |m, f| m[f.name] = f; m}
-
-          # Items in the input that are unexpected
-          input.each do |name, value|
-            if visible_arguments_map[name].nil?
-              result.add_problem("Field is not defined on #{self.graphql_name}", [name])
+          # Inject missing required arguments
+          required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
```
eapache, line 188: I might call this `missing_required_inputs`

lib/graphql/schema/warden.rb
```diff
@@ -106,6 +106,12 @@ def get_field(parent_type, field_name)
         @visible_parent_fields[parent_type][field_name]
       end
 
+      # @return [GraphQL::Argument, nil] The argument named `argument_name` on `parent_type`, if it exists and is visible
+      def get_argument(parent_type, argument_name)
```
eapache, line 110: I believe the intent of this method was to be cached which it is not right now?

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
-              result.merge_result!(name, argument_result)
+          input.to_h.merge(required_inputs).each do |argument_name, value|
+            argument = warden.get_argument(self, argument_name)
+            # Items in the input that are unexpected
+            unless argument
+              result.add_problem("Field is not defined on #{self.graphql_name}", [argument_name])
+              next
             end
+            # Items in the input that are expected, but have invalid values
+            argument_result = argument.type.validate_input(value, ctx)
```
eapache, line 204: @rmosolgo it's weird to me that we handle missing required inputs by adding them as nil, and then letting the validation fail that nil isn't an accepted value? Could we not just `add_problem` directly when we notice they're missing?

grcooper: Yah I was also not sure about this behaviour, I would assume this should actually be handled by a different part of code. 

We can add a specific problem if we notice they are missing rather than doing what I am currently doing.

lib/graphql/schema/warden.rb
grcooper: `visible_argument?(argument)` is cached through the `visible?` method on types. So I believe this is ok to leave non-cached?

I can make it more similar to the above methods though and cache it further.

eapache: I'll defer to @rmosolgo's judgement here. I did note that we are now calling `get_argument` twice on the same argument in some scenarios.

chrisbutcher:
👋  I'm taking over the PR as of today while Graham is 🌴

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
+
+            m
           end
 
-          # Items in the input that are expected, but have invalid values
-          visible_arguments_map.map do |name, argument|
-            argument_result = argument.type.validate_input(input[name], ctx)
-            if !argument_result.valid?
-              result.merge_result!(name, argument_result)
+          input.to_h.merge(required_inputs).each do |argument_name, value|
```
chrisbutcher, line 196: Trying this out in Shopify, `input` is sometimes a `ActionController::Parameters` instance, e.g. `<ActionController::Parameters {"barcode"=>"53941047", "id"=>"gid://shopify/ProductVariant/30322695"} permitted: false>`.

This causes `input.to_h` to raise `ActionController::UnfilteredParameters: unable to convert unpermitted parameters to hash`.

I notice that above this line we have this block:

```ruby
begin
  input.to_h
rescue
  begin
    # Handle ActionController::Parameters:
    input.to_unsafe_h
  rescue
    # We're not sure it'll act like a hash, so reject it:
    result.add_problem(INVALID_OBJECT_MESSAGE % { object: JSON.generate(input, quirks_mode: true) })
    return result
  end
end
```

Which doesn't actually convert `input` if it's a `ActionController::Parameters`. Should it do so?

Elsewhere in this gem, I see [handling for this same scenario](https://github.com/rmosolgo/graphql-ruby/blob/c67be42/lib/graphql/language/sanitized_printer.rb#L154-L160), checking if `value.respond_to?(:to_unsafe_h)`.

lib/graphql/schema/warden.rb
```diff
@@ -106,6 +106,12 @@ def get_field(parent_type, field_name)
...
+        argument = parent_type.arguments[argument_name]
```
rmosolgo, line 111: `#arguments` climbs the inheritance chain and merges argument definitions into a hash. When you only need one object, better to use `.get_argument` which searches the chain until it finds a hit, and doesn't merge any hashes: 

```suggestion
        argument = parent_type.get_argument(argument_name)
```

lib/graphql/schema/warden.rb
rmosolgo: Yeah, the call to the application's implementation would be cached, I don't know whether using `read_through { ... }` here is worth it or not, no preference without actually seeing a benchmark.

lib/graphql/schema/input_object.rb
rmosolgo: > adding them as nil

Do you mean, it worked this way in the old code? AFAIK the old code checked each argument and made sure that the provided value was valid, without special nil handling. I see that in the proposed change here, it has to handle `nil` differently.

lib/graphql/schema/input_object.rb
rmosolgo: Previously, it required that `.each` yielded key-value pairs. I think making a hash out of it is fine, if you have to, too 🤷

lib/graphql/schema/input_object.rb
rmosolgo, line 188: `self.arguments` _can_ be expensive, for input objects that inherit arguments from their parent classes, because it merges hashes. In practice, it might not happen much, and the most likely case (no inherited arguments) is optimized to return `own_arguments`. But, the nice part about `warden.arguments(self)` here was that it was cached in any case. Not a show-stopper but thought I'd mention it in case it comes up later.

lib/graphql/schema/warden.rb
chrisbutcher: 👀  I may be missing something, but I don't see `get_argument` defined on `GraphQL::Schema::InputObject`

https://graphql-ruby.org/api-doc/1.10.11/GraphQL/Schema/InputObject.html

lib/graphql/schema/input_object.rb
```diff
@@ -171,7 +171,7 @@ def validate_non_null_input(input, ctx)
           # We're not actually _using_ the coerced result, we're just
           # using these methods to make sure that the object will
           # behave like a hash below, when we call `each` on it.
-          begin
+          input = begin
```
chrisbutcher, line 174: This resolves https://github.com/rmosolgo/graphql-ruby/pull/2985#discussion_r439458182

lib/graphql/schema/warden.rb
rmosolgo: Ah, sorry, I was thinking it would be around here, but I was mistaken: https://github.com/rmosolgo/graphql-ruby/blob/12947f427d7e4b3e590b743b3c2c3ef04ece2712/lib/graphql/schema/member/has_arguments.rb#L50-L59

It could be added like get_field: 

https://github.com/rmosolgo/graphql-ruby/blob/12947f427d7e4b3e590b743b3c2c3ef04ece2712/lib/graphql/schema/member/has_fields.rb#L29-L40

chrisbutcher: Good idea! Added :)

lib/graphql/schema/input_object.rb
eapache: The comment on this block is now wrong

chrisbutcher: Good catch! 🔥

lib/graphql/schema/input_object.rb
```diff
@@ -184,21 +184,25 @@ def validate_non_null_input(input, ctx)
...
           end
 
-          visible_arguments_map = warden.arguments(self).reduce({}) { |m, f| m[f.name] = f; m}
-
-          # Items in the input that are unexpected
-          input.each do |name, value|
-            if visible_arguments_map[name].nil?
-              result.add_problem("Field is not defined on #{self.graphql_name}", [name])
+          # Inject missing required arguments
+          missing_required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
```
eapache, line 185: this hash has entirely nil values... I don't know if there's actually a nicer way to build it, or if we could get away with just an array of keys instead or something? but don't let it hold this up in the short term

lib/graphql/schema/input_object.rb
eapache: As I understand it the old code was kind of implicit - it checks every visible argument, and if the argument was required but had no value provided, then `argument.type.validate_input` would produce an error.

Now we're being a bit more explicit about generating a list of `missing_required_inputs`. Rather than merging those explicit nils into the input just so we can call `validate_input` on nil to generate the error... I think we could just generate the error *instead* of building `missing_required_inputs` at all?

rmosolgo: > generate the error instead of building `missing_required_inputs`

Ah, yeah, fine with me. (And I guess it must have worked this way in the past, if it _wasn't_ triggering `.visible?` 🤔 )

chrisbutcher: > generate the error instead of building `missing_required_inputs`

In that case, I'm a bit unsure of how this should interact/overlap with the [existing static validation of input object attributes](https://github.com/rmosolgo/graphql-ruby/blob/26d8a66/lib/graphql/static_validation/rules/required_input_object_attributes_are_present.rb#L48-L49).

A casual reading of that file initially suggested to me that this check would be [performed earlier](https://graphql-ruby.org/queries/phases_of_execution.html), removing the need to check for required + visible + not provided input arguments via `GraphQL::Schema::InputObject.validate_non_null_input`.

On a related note: the following diff produces errors only in tests related to variable coercion, which makes me think that the [passing tests for required input object arguments](https://github.com/rmosolgo/graphql-ruby/blob/26d8a66/spec/graphql/static_validation/rules/required_input_object_attributes_are_present_spec.rb) are hitting only the static validation checks.

<details><summary>diff</summary>
<p>

```diff
diff --git a/lib/graphql/schema/input_object.rb b/lib/graphql/schema/input_object.rb
index c4fb241fd..0a3a4c52d 100644
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -181,16 +181,13 @@ module GraphQL
             end
           end

-          # Inject missing required arguments
-          missing_required_inputs = self.arguments.reduce({}) do |m, (argument_name, argument)|
+          self.arguments.each do |argument_name, argument|
             if !input.key?(argument_name) && argument.type.non_null? && warden.get_argument(self, argument_name)
-              m[argument_name] = nil
+              result.add_problem("TODO on #{self.graphql_name}", [argument_name])
             end
-
-            m
           end

-          input.merge(missing_required_inputs).each do |argument_name, value|
+          input.each do |argument_name, value|
             argument = warden.get_argument(self, argument_name)
             # Items in the input that are unexpected
             unless argument
```

</p>
</details>


<details><summary>test output</summary>
<p>

```
Failure:
GraphQL::Query::Executor::variable coercion::for input objects with unknown keys in value#test_0001_returns a variable validation error [/Users/chris/src/github.com/grcooper/graphql-ruby/spec/graphql/query/executor_spec.rb:329]
+++ actual
@@ -1,11 +1,10 @@
 {"errors"=>
   [{"message"=>
-     "Variable $input of type [DairyProductInput] was provided invalid value for 0.foo (Field is not defined on DairyProductInput), 0.source (Expected value to not be null)",
+     "Variable $input of type [DairyProductInput] was provided invalid value for 0.source (TODO on DairyProductInput), 0.foo (Field is not defined on DairyProductInput)",
     "locations"=>[{"line"=>1, "column"=>10}],
     "extensions"=>
      {"value"=>[{"foo"=>"bar"}],
       "problems"=>
-       [{"path"=>[0, "foo"],
-         "explanation"=>"Field is not defined on DairyProductInput"},
-        {"path"=>[0, "source"],
-         "explanation"=>"Expected value to not be null"}]}}]}
+       [{"path"=>[0, "source"], "explanation"=>"TODO on DairyProductInput"},
+        {"path"=>[0, "foo"],
+         "explanation"=>"Field is not defined on DairyProductInput"}]}}]}


Failure:
GraphQL::Query::Executor::variable coercion::for required input object fields#test_0001_returns a variable validation error [/Users/chris/src/github.com/grcooper/graphql-ruby/spec/graphql/query/executor_spec.rb:306]
+++ actual
@@ -1,9 +1 @@
-{"errors"=>
-  [{"message"=>
-     "Variable $input of type ReplaceValuesInput! was provided invalid value for values (Expected value to not be null)",
-    "locations"=>[{"line"=>1, "column"=>13}],
-    "extensions"=>
-     {"value"=>{},
-      "problems"=>
-       [{"path"=>["values"],
-         "explanation"=>"Expected value to not be null"}]}}]}
+#<GraphQL::Query::Result @query=... @to_h={"errors"=>[{"message"=>"Variable $input of type ReplaceValuesInput! was provided invalid value for values (TODO on ReplaceValuesInput)", "locations"=>[{"line"=>1, "column"=>13}], "extensions"=>{"value"=>{}, "problems"=>[{"path"=>["values"], "explanation"=>"TODO on ReplaceValuesInput"}]}}]}>


1602 tests, 7612 assertions, 2 failures, 0 errors, 0 skips
```

</p>
</details>

eapache: My original comment here was an architecture/style question, not a behavioural one. It sounds like there's more we need to dig into here, but if we're confident that the current code works correctly could we get that out and then follow up on the other points?

rmosolgo:
Just shipped in 1.10.13! 

Thanks @grcooper for opening the followup issue. I'll add some thoughts there

PR #2987, opened 2020-06-11
Remove unused instance variable in GraphQL::Schema::InputObject

jturkel:
I was doing some code spelunking and noticed that the `@arguments_by_keyword` instance variable in `GraphQL::Schema::InputObject` was unused.


rmosolgo:
🎊 !

PR #2989, opened 2020-06-12
Fix DateTime type to handle dates without times

### LINKED ISSUE #2988: DateTime scalar change in 1.10.11 broke backwards compatibility
**Describe the bug**

The changes introduced by pull request https://github.com/rmosolgo/graphql-ruby/pull/2973 broke backwards compatibility for this gem. 

If you have a mutation with a required input of type GraphQL::Types::ISO8601DateTime, and the input value provided to the mutation is a date such as 2020-06-12, versions 1.10.10 and earlier of this gem will accept this input and set the time element to 00:00:00. However version 1.10.11 will cause the mutation to fail and return an error like the following:

```
 {"message"=>"Argument 'reference_date' on Field 'create_shape' has an invalid value (\"2020-06-12\"). Expected type 'ISO8601DateTime!'.", "locations"=>[{"line"=>4, "column"=>5}], "path"=>["mutation", "create_shape", "reference_date"], "extensions"=>{"code"=>"argumentLiteralsIncompatible", "typeName"=>"Field", "argumentName"=>"reference_date"}}
```

You can see why this behavior has changed, by running this change made by the pull request in the Rails console:

```ruby
➜  rails c
> date_str = "2020-06-12"
 => "2020-06-12"
> Time.iso8601(date_str)
        1: from (irb):8
ArgumentError (invalid date: "2020-06-12")
> DateTime.iso8601(date_str)
 => Fri, 12 Jun 2020 00:00:00 +0000
```

Granted, this particular input should have been initially modeled using GraphQL::Types::ISO8601Date instead of GraphQL::Types::ISO8601DateTime, as it is stored as a date in our database. However, it wasn't, and making this change now would alter the API's schema and break any GraphQL client that is explicitly expecting this input to be of type GraphQL::Types::ISO8601DateTime.

**Versions**

`graphql` version: 1.10.11
`rails` (or other framework): We are using 6.0.3.1, but this issue is not specific to this Rails version

**GraphQL schema**

```
create_shape(
  name: String!
  reference_date: ISO8601DateTime!
  geometry: GeoJSON!
): Shape!
```

**GraphQL query**

```graphql
mutation CreateShape {
  create_shape(
    name: "Example Shape"
    reference_date: "2020-06-12"
    geometry: {
      type: "FeatureCollection",
      features: [
        {
          type: "Feature",
          properties: {},
          geometry: {
            type: "GeometryCollection",
            geometries: [
              {
                type: "Polygon",
                coordinates: [
                  [
                    [ -95.5948085, 47.0946385 ],
                    [ -95.5885835, 47.0947915 ],
                    [ -95.5893452, 47.0933998 ],
                    [ -95.5948085, 47.0946385 ]
                  ],
                  [
                    [ -95.591592, 47.0939598 ],
                    [ -95.5895655, 47.0945269 ],
                    [ -95.5908101, 47.0944966 ],
                    [ -95.591592, 47.0939598 ]
                  ]
                ]
              }
            ]
          }
        }
      ]
    }
  ) {
    sentera_id
    color
    acres
    reference_date
    geometry
  }
}
```

```json
{
  "errors": [
    {
      "message": "Argument 'reference_date' on Field 'create_shape' has an invalid value (\"2020-06-12\"). Expected type 'ISO8601DateTime!'.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "mutation CreateShape",
        "create_shape",
        "reference_date"
      ],
      "extensions": {
        "code": "argumentLiteralsIncompatible",
        "typeName": "Field",
        "argumentName": "reference_date"
      }
    }
  ]
}
```

**Steps to reproduce**

Use the GraphQL schema and mutation listed above with version 1.10.11 of this gem to reproduce this bug.

**Expected behavior**

Passing a date formatted as ISO 8601 (ie. 2020-06-12) as a mutation input of type GraphQL::Types::ISO8601DateTime should work. The date should get converted with a 00:00:00 time element, which was the previous behavior of versions < 1.10.11).

**Actual behavior**

The mutation will fail as demonstrated by the mutation response JSON included above.

rmosolgo:
Fixes #2988 

cc @bdewater just fyi, since you were working around here recently

### DIFF
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -52,8 +52,12 @@ module GraphQL
       def self.coerce_input(str_value, _ctx)
         Time.iso8601(str_value)
       rescue ArgumentError, TypeError
-        # Invalid input
-        nil
+        begin
+          DateTime.iso8601(str_value).to_time
+        rescue ArgumentError, TypeError
+          # Invalid input
+          nil
+        end
       end
     end
   end
--- a/lib/graphql/types/iso_8601_date_time.rb
+++ b/lib/graphql/types/iso_8601_date_time.rb
@@ -53,7 +53,7 @@ module GraphQL
         Time.iso8601(str_value)
       rescue ArgumentError, TypeError
         begin
-          DateTime.iso8601(str_value).to_time
+          Date.iso8601(str_value).to_time
         rescue ArgumentError, TypeError
           # Invalid input
           nil
--- a/spec/graphql/types/iso_8601_date_time_spec.rb
+++ b/spec/graphql/types/iso_8601_date_time_spec.rb
@@ -123,6 +123,9 @@ describe GraphQL::Types::ISO8601DateTime do
 
     it "parses dates without times" do
       res = parse_date("2018-06-07")
+      # It uses the system default timezone when none is given
+      system_default_tz = Date.iso8601("2018-06-07").to_time.zone
+      system_default_offset = Date.iso8601("2018-06-07").to_time.utc_offset
       expected_res = {
         "year" => 2018,
         "month" => 6,
@@ -130,8 +133,8 @@ describe GraphQL::Types::ISO8601DateTime do
         "hour" => 0,
         "minute" => 0,
         "second" => 0,
-        "zone" => testing_rails? ? "UTC" : nil,
-        "utcOffset" => 0,
+        "zone" => system_default_tz,
+        "utcOffset" => system_default_offset,
       }
       assert_equal(expected_res, res)
     end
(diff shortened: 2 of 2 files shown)


lib/graphql/types/iso_8601_date_time.rb
```diff
@@ -52,8 +52,12 @@ def self.coerce_result(value, _ctx)
       def self.coerce_input(str_value, _ctx)
         Time.iso8601(str_value)
       rescue ArgumentError, TypeError
-        # Invalid input
-        nil
+        begin
+          DateTime.iso8601(str_value).to_time
```
bdewater, line 56: I think this should be `Date.iso8601(str_value).to_time` so it inherits the system timezone, similarly to how `coerce_result` handles `Date` objects. `DateTime` always defaults to UTC.

```
irb(main):001:0> require 'time'
=> true
irb(main):002:0> ENV['TZ'] = 'America/Toronto'
irb(main):003:0> str_value = '2018-06-07'
irb(main):004:0> DateTime.iso8601(str_value)
=> #<DateTime: 2018-06-07T00:00:00+00:00 ((2458277j,0s,0n),+0s,2299161j)>
irb(main):005:0> Date.iso8601(str_value)
=> #<Date: 2018-06-07 ((2458277j,0s,0n),+0s,2299161j)>
irb(main):006:0> DateTime.iso8601(str_value).to_time
=> 2018-06-07 00:00:00 +0000
irb(main):007:0> Date.iso8601(str_value).to_time
=> 2018-06-07 00:00:00 -0400
irb(main):008:0> ENV['TZ'] = 'Etc/UTC'
irb(main):009:0> DateTime.iso8601(str_value).to_time
=> 2018-06-07 00:00:00 +0000
irb(main):010:0> Date.iso8601(str_value).to_time
=> 2018-06-07 00:00:00 +0000
```

rmosolgo: Thanks for the suggestion :+1: !

PR #2992, opened 2020-06-13
Pubnub subscription docs

rmosolgo:
TODO: 

- [ ] Improve pubnub link implementation


rmosolgo:
I was working on this for someone, but they ended up taking a different route. Without a known use case, I'm not going to add it.

PR #2993, opened 2020-06-15
Fix issues around early disposal

jscheid:
1. In certain circumstances, the initial response can cause `dispose` to be called before the subscription has finished setting up. In this situation, the channel is never detached.
2. Likewise, if the subscription is disposed while the Ably channel is still attaching, detaching the channel doesn't do anything.

This PR fixes both of those issues. I've also simplified cleanup based on feedback from Ably.

### DIFF
--- a/javascript_client/src/subscriptions/createAblyHandler.ts
+++ b/javascript_client/src/subscriptions/createAblyHandler.ts
@@ -68,7 +68,6 @@ function createAblyHandler(options: AblyHandlerOptions) {
         observer.onCompleted()
       }
     }
-
     ;(async () => {
       try {
         // POST the subscription like a normal query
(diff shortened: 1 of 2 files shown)


javascript_client/src/subscriptions/createAblyHandler.ts
```diff
@@ -123,6 +121,8 @@ function createAblyHandler(options: AblyHandlerOptions) {
         }
         // When you get an update from ably, give it to Relay
         channel.subscribe("update", updateHandler)
+
+        dispatchResult(response.body)
```
rmosolgo, line 125:
```suggestion
        // If this body contains errors, Relay will immediately clean up the subscription. 
        // To simplify the channel state, only dispatch the result _after_ the channel is set up. 
        // That way, if Relay tries to clean it up, it will get cleaned up properly.
        dispatchResult(response.body)
```

rmosolgo: I'll add this before publishing

rmosolgo:
🚢 1.7.11 !

PR #3000, opened 2020-06-17
Remove more visible? calls to unused arguments

swalkinshaw:
Replaces more usages of `warden.arguments` during static validation (since they trigger `visible?` calls for all arguments on a type regardless if they are specified or not) with more selective `warden.get_argument`.

https://github.com/rmosolgo/graphql-ruby/pull/2985 which fixed some of these occurrences, but not all of them.

This also required additional backwards compatible work to port `get_argument` to the legacy types.

cc @eapache @grcooper @chrisbutcher

### DIFF
--- a/lib/graphql/schema/member/has_arguments.rb
+++ b/lib/graphql/schema/member/has_arguments.rb
@@ -60,7 +60,9 @@ module GraphQL
 
         # @return [GraphQL::Schema::Argument, nil] Argument defined on this thing, fetched by name.
         def get_argument(argument_name)
-          if (a = own_arguments[argument_name])
+          a = own_arguments[argument_name]
+
+          if a || !self.is_a?(Class)
             a
           else
             for ancestor in ancestors
--- a/lib/graphql/directive.rb
+++ b/lib/graphql/directive.rb
@@ -99,6 +99,10 @@ module GraphQL
     def type_class
       metadata[:type_class]
     end
+
+    def get_argument(argument_name)
+      arguments[argument_name]
+    end
   end
 end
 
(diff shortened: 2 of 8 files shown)


lib/graphql/schema/member/has_arguments.rb
```diff
@@ -58,6 +58,22 @@ def arguments
           end
         end
 
+        # @return [GraphQL::Schema::Argument, nil] Argument defined on this thing, fetched by name.
+        def get_argument(argument_name)
+          a = own_arguments[argument_name]
+
+          if a || !self.is_a?(Class)
```
swalkinshaw, line 65: Turns out this method wasn't entirely correct since `HasArguments` needs to support being both `extend`ed and `include`ed. Using `ancestors` unconditionally breaks this implicit contract but didn't cause any issues until this PR so I had to support the instance case for fields.

swalkinshaw:
Oh wait I need to redo this from `1.10.x` branch

swalkinshaw:
@grcooper it's pretty hard to write an isolated test case for that 🤔 tests do fail without it however

rmosolgo:
(sorry, evidently I typed up a review here but failed to submit it yesterday!) 

I'm game to include this as-is, but I'm concerned about the long-term plan here. Without tests in the gem that `.visible?` isn't called on arguments that aren't present, it's likely to regress. Do you plan to remove your dependency on that assumption, or should we start looking for a way to make sure that these improvements aren't accidentally lost?

swalkinshaw:
> Do you plan to remove your dependency on that assumption, or should we start looking for a way to make sure that these improvements aren't accidentally lost?

Yeah we already have some work in progress for an alternative approach 👍 

However, I think these changes are useful on their own? In most of the cases that used `arguments`, `get_argument` is a simpler replacement since only one argument was being looked up.

rmosolgo:
:+1: Agreed that many of these are outright improvements!

PR #3003, opened 2020-06-22
Fix arguments caching - check for field extensions

swalkinshaw:
Previously this only checked for field extras but field extensions should be checked as well since they can mutate arguments.

@jturkel does this makes sense to you as well? I guess you didn't have any instances of field extensions mutating args?

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -460,8 +460,9 @@ module GraphQL
         end
 
         def arguments(graphql_object, arg_owner, ast_node)
-          # Don't cache arguments if field extras are requested since extras mutate the argument data structure
-          if arg_owner.arguments_statically_coercible? && (!arg_owner.is_a?(GraphQL::Schema::Field) || arg_owner.extras.empty?)
+          # Don't cache arguments if field extras or extensions are requested since they can mutate the argument data structure
+          if arg_owner.arguments_statically_coercible? &&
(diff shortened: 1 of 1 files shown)


rmosolgo:
Oh, interesting, I guess the point is, `field.extras.empty?` _isn't enough_ to be sure that the arguments won't be modified.

I wonder if there's anything else on `Field` that could modify the arguments (Ok, it's Ruby, I'm sure the answer is yes). I guess if another one pops up, we can try to generalize about it.

swalkinshaw:
Yeah it kind of seems like since `arguments_statically_coercible` exists now, some equivalent should on field? 🤷‍♂️ This becomes more a problem if any other feature like extras/extensions is added in the future and we forget to update this check. I'm not sure how to best guarantee that though.

jturkel:
The fix makes sense to me but it's a bit pessimistic since it assumes that all field extensions will modify the arguments data structure. Perhaps the right long term direction (in 1.12?) is making `GraphQL::Interpreter::Arguments` immutable to avoid bugs like this and only take the performance penalty of a copy/mutate when necessary?

swalkinshaw:
Agreed that sounds like a good plan 👍

rmosolgo:
Thanks for the great suggestion, @jturkel, I've opened an issue to keep track of that https://github.com/rmosolgo/graphql-ruby/issues/3004

chrisbutcher:
🎉

PR #3011, opened 2020-07-01
Add validation to prevent arguments with both required and default_value

### LINKED ISSUE #3001: Arguments should prevent _both_ default_value and required: true
I don't think there's any reason to permit `default_value:` _and_ `required: true`, because a required argument will never apply a default value. 

This was previously validated by legacy schemas, but it seems like class-based schemas aren't validating it properly.

Diego-Thomaz:
Close #3001

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -61,7 +61,10 @@ module GraphQL
         @from_resolver = from_resolver
         @method_access = method_access
 
-        check_conflicting_params
+        if !@null && default_value?
+          raise ArgumentError, "Argument '#{@name}' has conflicting params, " \
+            "either use `required: false` or remove `default_value:`."
+        end
 
         if definition_block
           if definition_block.arity == 1
@@ -187,15 +190,6 @@ module GraphQL
         else
           raise "Invalid prepare for #{@owner.name}.name: #{@prepare.inspect}"
         end
-      end
-
-      private
-
-      def check_conflicting_params
-        return unless !@null && default_value?
-
-        raise ArgumentError, "Argument '#{@name}' has conflicting params! " \
-          "Arguments can't be required and have default values at the same time!"
       end
     end
   end
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -61,6 +61,8 @@ module GraphQL
         @from_resolver = from_resolver
         @method_access = method_access
 
+        check_conflicting_params
+
         if definition_block
           if definition_block.arity == 1
             instance_exec(self, &definition_block)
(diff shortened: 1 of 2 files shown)


spec/graphql/schema/argument_spec.rb
```diff
@@ -167,47 +169,60 @@ def self.object_from_id(id, ctx)
...
       GRAPHQL
 
       res = SchemaArgumentTest::Schema.execute(query_str)
-      assert_equal "Argument 'requiredWithDefaultArg' on Field 'field' has an invalid value (null). Expected type 'Int!'.", res['errors'][0]['message']
+      assert_equal "Argument 'requiredArg' on Field 'field' has an invalid value (null). Expected type 'Int!'.", res['errors'][0]['message']
+    end
+  end
+
+  describe 'invalid:' do
+    it 'field has conflicting params' do
```
lairtonlelis, line 201: how about changing this description to work with `it` in a sentence, to make things more readable:
```suggestion
    it 'raises an error when a field has conflicting params' do
```

lib/graphql/schema/argument.rb
```diff
@@ -186,6 +188,15 @@ def prepare_value(obj, value, context: nil)
...
         end
       end
+
+      private
+
+      def check_conflicting_params
+        return unless !@null && default_value?
+
+        raise ArgumentError, "Argument '#{@name}' has conflicting params! " \
+          "Arguments can't be required and have default values at the same time!"
```
rmosolgo, line 198: The one thing I'd add is a clue to the developer _how_ to fix the error, for example: 

```suggestion
          "Arguments can't be required and have default values at the same time, either use `required: false` or remove `default_value:`."
```

lib/graphql/schema/argument.rb
```diff
@@ -61,6 +61,8 @@ def initialize(arg_name = nil, type_expr = nil, desc = nil, required:, type: nil
         @from_resolver = from_resolver
         @method_access = method_access
 
+        check_conflicting_params
+
```
rmosolgo, line 65: From a maintainability standpoint, it'd be easier for me if the code was inline, here in `#initialize` instead of separated by a jump, down at the bottom of the file. 

For a reader who wants to debug this method, they'll have to break context and go read the bottom of the file if they want to know what's going on here. 

The name, `check_conflicting_params`, doesn't make it clear which params are being checked or what kind of conflicts its looking for. Since the method accepts no arguments, the reader is invited to assume that _any_ internal state of the instance is used as input by `check_conflicting_params`. 

Do you mind removing the method and putting the validation logic here in `#initialize`?

Diego-Thomaz: I'll do the changes!!

Diego-Thomaz:
@rmosolgo take a look and let me know if it is ok now!

rmosolgo:
Sorry @Diego-Thomaz , I made a mistake on this one -- apparently there are cases where default_value: and `required: true` make sense (https://github.com/graphql/graphql-spec/pull/418). Maybe I named `required:` wrongly 😖

Diego-Thomaz:
> Sorry @Diego-Thomaz , I made a mistake on this one -- apparently there are cases where default_value: and `required: true` make sense ([graphql/graphql-spec#418](https://github.com/graphql/graphql-spec/pull/418)). Maybe I named `required:` wrongly 😖

No problem @rmosolgo I was glad to help anyway.

PR #3014, opened 2020-07-07
Load deprecation reason from introspection results

jturkel:
This PR updates `GraphQL::Schema.from_introspection` to load field and enum value deprecation reasons from introspection results. I converted `spec/graphql/schema/loader_spec.rb` to use the not so new anymore class based API while I was adding test coverage.


rmosolgo:
Thanks for the fix and the test upgrade!

Issue #3028, opened 2020-07-13
`default_broadcastable: true` raises error


rgraff:
**Describe the bug**

When using subscriptions with `default_broadcastable: true` the fingerprint query attempts to call undefined methods size on ActionController::Parameters. 

```ruby
      use GraphQL::Pro::AblySubscriptions, redis: $graphql_subscriptions_redis, ably: $ably, broadcast: true, default_broadcastable: true
```

This error is not raised withe `default_broadcastable: false`

Trace:

```
undefined method `size' for #<ActionController::Parameters:0x00005567e61fd8f0>
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/query.rb:291:in `variables_fingerprint'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/query.rb:281:in `fingerprint'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/subscriptions/event.rb:63:in `fingerprint'
/usr/local/bundle/gems/graphql-pro-1.14.0/lib/graphql/pro/subscriptions/redis_storage.rb:34:in `map'
/usr/local/bundle/gems/graphql-pro-1.14.0/lib/graphql/pro/subscriptions/redis_storage.rb:34:in `write_subscription'
/usr/local/bundle/gems/graphql-pro-1.14.0/lib/graphql/pro/subscriptions.rb:82:in `write_subscription'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/subscriptions/instrumentation.rb:34:in `after_query'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/execution/instrumentation.rb:82:in `public_send'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/execution/instrumentation.rb:82:in `block in call_after_hooks'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/execution/instrumentation.rb:80:in `reverse_each'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/execution/instrumentation.rb:80:in `call_after_hooks'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/execution/instrumentation.rb:74:in `call_hooks'
/usr/local/bundle/gems/graphql-1.11.1/lib/graphql/execution/instrumentation.rb:44:in `each_query_call_hooks'
[...]

PR #3029, opened 2020-07-13
Update controller generator so that provided_variables has .size

rgraff:
One possible solution to #3028.

The variables_fingerprint calls size on the variables.  When variables are an instance of ActionController::Parameters, then this will raise an error as size in not defined. Converting ActionController::Parameters to a hash resolves the issue.

### DIFF
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -30,8 +30,10 @@ class GraphqlController < ApplicationController
       else
         {}
       end
-    when Hash, ActionController::Parameters
+    when Hash
       ambiguous_param
+    when ActionController::Parameters
+      ambiguous_param.to_unsafe_hash
     when nil
       {}
     else
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -26,7 +26,7 @@ class GraphqlController < ApplicationController
     case ambiguous_param
     when String
       if ambiguous_param.present?
-        ensure_hash(JSON.parse(ambiguous_param))
+        JSON.parse(ambiguous_param) || {}
       else
         {}
       end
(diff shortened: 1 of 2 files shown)


lib/generators/graphql/templates/graphql_controller.erb
```diff
@@ -30,8 +30,10 @@ class GraphqlController < ApplicationController
       else
         {}
       end
-    when Hash, ActionController::Parameters
+    when Hash
       ambiguous_param
+    when ActionController::Parameters
+      ambiguous_param.to_unsafe_hash
```
rmosolgo, line 36: How about adding a comment here to explain _why_ an `_unsafe` call is appropriate here? 

```suggestion
      ambiguous_param.to_unsafe_hash # GraphQL-Ruby will validate name and type of incoming variables.
```

That way, an unsuspecting reader won't be scared by seeing `unsafe` in the source!

rgraff:
@rmosolgo if you're open to bigger changes, I would rename the `ensure_hash` method to `variables`. Given the changes, it's no longer a generic method and I wouldn't want to see it used in other areas.

rmosolgo:
:+1: the rename is good by me!

rgraff:
@rmosolgo ok, renamed the method and made it more specific. If this is too big of a change, let me know.

rmosolgo:
Awesome! Thanks again for the fix. I pushed a commit to tweak the method signature a bit and update the tests.

cesarjr:
Hi!

I tried to write an automated test for the line below but I couldn't find a way to send a pure hash by controller. Is it possible that this line is not necessary?

https://github.com/rmosolgo/graphql-ruby/pull/3029/files#diff-f8e8f53d8b7477b924f19d4566869075babb63f67c7beb0c913a59031a7ce50dR34

Thanks.

rmosolgo:
Yes, I think it's possible that the code there is never called. From the diff, it looks like that method used to be recursive, but it's not anymore :+1:

PR #3030, opened 2020-07-14
Only check used input arguments visibility [for 1.11.x]

swalkinshaw:
Same as https://github.com/rmosolgo/graphql-ruby/pull/2985 but on master/1.11


rmosolgo:
Thanks!

PR #3031, opened 2020-07-14
Fix visible calls in static validation [for 1.11.x]

swalkinshaw:
Same as https://github.com/rmosolgo/graphql-ruby/pull/3000 but for master/1.11.

~Based on https://github.com/rmosolgo/graphql-ruby/pull/3030 so I'll update once that's merged.~

### DIFF
--- a/lib/graphql/directive.rb
+++ b/lib/graphql/directive.rb
@@ -99,6 +99,10 @@ module GraphQL
     def type_class
       metadata[:type_class]
     end
+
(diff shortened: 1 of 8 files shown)


swalkinshaw:
Updated this. Good to go 👍

rmosolgo:
:+1: LGTM with the usual hand wringing about how it might "break" again in the future 😅

PR #3032, opened 2020-07-14
[graphql-ruby-client] Add engines to package.json

MichaelDeBoey:


rmosolgo:
Thanks for the improvement!

PR #3033, opened 2020-07-15
Correct multiplex documentation regarding the _json field in the params variable

gbillig:
I was implementing multiplexing for my server, and in the documentation it was mentioned that Apollo sends the batched queries in a `_json` field. However when I inspected the request from the client, there was no `_json` field.

I dug deeper, and found that Apollo client sends the queries in a JSON array, and that Rails' ActionDispatch is the one who populates the "_json" field when the parsed data is not a Hash ([source](https://github.com/rails/rails/blob/e82c10b49ad9cb949f77297e6a814fd8f09d3c8f/actionpack/lib/action_dispatch/http/parameters.rb#L13)).

I also fixed some dead links.

### DIFF
--- a/guides/queries/multiplex.md
+++ b/guides/queries/multiplex.md
@@ -8,7 +8,7 @@ desc: Run multiple queries concurrently
 index: 10
 ---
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the updates!

PR #3034, opened 2020-07-16
[Docs] Improve code formatting in Connections

swalkinshaw:
These code blocks were displayed inline before and looked broken.

Looks like this right now:
image


rmosolgo:
(got it! a3796104d ⚡)

Issue #3017, opened 2020-07-09
Custom EdgeType - Failed to implement error


NicholasHazen:
I'm trying to create a new edge field "tag" on my MessageEdgeType class. The tag field needs a reference to the parent and the node object in order to properly calculate the return value. If I implement MessageEdgeType#tag, I do not have access to the parent object. When I try to follow the documentation [here](https://graphql-ruby.org/relay/connections.html#custom-edge-classes) I get back an error during execution of
 ```
Failed to implement MessageEdge.tag, tried:
        - `Types::MessageEdgeType#tag`, which did not exist
        - `GraphQL::Pagination::Connection::Edge#tag`, which did not exist
        - Looking up hash key `:tag` or `"tag"` on `#<GraphQL::Pagination::Connection::Edge:0x0000560513ee1100>`, but it wasn't a Hash

        To implement this field, define one of the methods above (and check for typos)
```

Maybe I misunderstood the documentation / implementation. Any help is appreciated.

**Versions**
`graphql` version: 1.10.10
`rails version: 6.0.3.2

**GraphQL schema**
```ruby
module Types
  class ConversationType < GraphQL::Schema::Object
    field :id, ID, null: false
    field :messages, Types::MessageConnectionType, null: false, connection: true
  end

  class MessageConnectionType < GraphQL::Types::Relay::BaseConnection
    edge_type(Types::MessageEdgeType, edge_class: Types::MessageEdge)
  end

  class MessageEdgeType < GraphQL::Types::Relay::BaseEdge
    node_type(Types::MessageType)

    field :tag, String, null: false, method: :tag
  end
  
  class MessageEdge < GraphQL::Relay::Edge
    def tag
      # I need access to both the node and the parent here
      "a tag value"
    end
  end

  class MessageType < GraphQL::Schema::Object
    field :id, ID, null: false
    field :text, String, null: true
  end
end
```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
query {
  conversations {
    messages {
      edges {
        tag
        cursor
        node {
          id
          text
        }
      }
    }
  }
}
```

```json
{
  "error": {
[...]

PR #3036, opened 2020-07-17
Use edge_class: with new pagination implementation

### LINKED ISSUE #3005: object.parent and/or edge_class not available to custom edges
**Describe the bug**

The current behavior for custom edge resolvers does not match the intended behavior [documented here](https://graphql-ruby.org/relay/connections.html#custom-edge-classes).

The documentation indicates that it is possible to define resolvers on custom edges which have access to `object.parent` (which is needed to locate rails "through relationships", for example). The documentation states that we are supposed to create a custom connection class which registers both a custom edge type class and custom edge class.

This doesn't work, as demonstrated below. The custom connection is successfully instantiated, and successfully instantiates the custom edge type class... but fails to instantiate the custom edge class, and object.parent is not available in the custom edge type class.

**Versions**

`graphql` version: graphql-1.10.10
`rails` (or other framework): rails-5.2.2.1
other applicable versions (`graphql-batch`, etc): graphql-batch-0.4.3

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok).
Are you using [interpreter](https://graphql-ruby.org/queries/interpreter.html)? Any custom instrumentation, etc?

```ruby
    class UserType < BaseObject
      field :id, GraphQL::Types::ID
      field :devices, UserDeviceConnection, null: false
    end

    class DeviceType < BaseObject
      field :os, GraphQL::Types::String
    end

    class UserDeviceConnection < GraphQL::Types::Relay::BaseConnection
      edge_type(UserDeviceEdgeType, edge_class: UserDeviceEdge)
    end

    class UserDeviceEdgeType < GraphQL::Types::Relay::BaseEdge
      node_type(Types::DeviceType)
      field :jwt_id, GraphQL::Types::String, null: true, method: :jwt_id
    end

    class UserDeviceEdge < GraphQL::Relay::Edge
      def initialize(node, connection)
        byebug #<< this never triggers
        super
      end

      def user_device
        @user_device ||= begin
          user = self.parent
          device = self.node
          UserDevice.where(user: user, device: device).first
        end
      end

      def jwt_id
        byebug #<< this never triggers
        user_device.jwt_id
      end
    end

# in app/models/device.rb
class Device
  #database fields -> id, os, ...
end

# in app/models/user.rb
class User
  #database fields -> id, name, ...
  has_many :devices, through: :user_devices
end

# in app/models/user_device.rb
class UserDevice
  #database fields -> user_id, device_id, jwt_id
end
```

**GraphQL query**

```graphql
query {
  users {      
    id
    devices {
      edges {
        jwtId # <<- everything else works, except this attribute. 
        node {
          os
        }
      }
    }
  }
}
```

```json
{
  "error": {
    "errors": [
      {
        "message": "        Failed to implement UserDeviceEdge.jwtId, tried:\n\n        - `GQL::Edges::UserDeviceEdgeType#jwt_id`, which did not exist\n        - `GraphQL::Pagination::Connection::Edge#jwt_id`, which did not exist\n        - Looking up hash key `:jwt_id` or `\"jwt_id\"` on `#<GraphQL::Pagination::Connection::Edge:0x0000555e9aefbed0>`, but it wasn't a Hash\n\n        To implement this field, define one of the methods above (and check for typos)\n"...
```

**Steps to reproduce**

Attempt to follow the documentation...

**Expected behavior**

object.parent should be accessible to resolvers in CustomEdgeType classes, or CustomEdge classes should be instantiated as documented.

</details>

**Additional context**

The following lines of code in the gem seem to ignore custom edge declarations. [BUG???]

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/relay/base_connection.rb#L109-L110

rmosolgo:
The new connection module didn't properly use `edge_class:` configurations.

Fixes #3005 
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3017

### DIFF
--- a/lib/graphql/pagination/connections.rb
+++ b/lib/graphql/pagination/connections.rb
@@ -58,7 +58,6 @@ module GraphQL
         @schema.ancestors.reverse_each do |schema_class|
           if schema_class.respond_to?(:connections) && (c = schema_class.connections)
             all_wrappers.merge!(c.wrappers)
-
           end
         end
         all_wrappers
--- a/lib/graphql/pagination/connections.rb
+++ b/lib/graphql/pagination/connections.rb
@@ -58,6 +58,7 @@ module GraphQL
         @schema.ancestors.reverse_each do |schema_class|
           if schema_class.respond_to?(:connections) && (c = schema_class.connections)
             all_wrappers.merge!(c.wrappers)
+
           end
         end
         all_wrappers
@@ -86,9 +87,21 @@ module GraphQL
           after: arguments[:after],
           last: arguments[:last],
           before: arguments[:before],
+          edge_class: edge_class_for_field(field),
         )
       end
 
+      # use an override if there is one
+      # @api private
+      def edge_class_for_field(field)
+        conn_type = field.type.unwrap
+        conn_type_edge_type = conn_type.respond_to?(:edge_class) && conn_type.edge_class
+        if conn_type_edge_type && conn_type_edge_type != Relay::Edge
+          conn_type_edge_type
+        else
+          nil
+        end
+      end
       protected
 
       attr_reader :wrappers
--- a/lib/graphql/pagination/connection.rb
+++ b/lib/graphql/pagination/connection.rb
@@ -15,11 +15,6 @@ module GraphQL
       class PaginationImplementationMissingError < GraphQL::Error
       end
 
-      # @return [Class] The class to use for wrapping items as `edges { ... }`. Defaults to `Connection::Edge`
-      def self.edge_class
-        self::Edge
-      end
-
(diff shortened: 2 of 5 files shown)


lib/graphql/pagination/connections.rb
```diff
@@ -58,6 +58,7 @@ def all_wrappers
         @schema.ancestors.reverse_each do |schema_class|
           if schema_class.respond_to?(:connections) && (c = schema_class.connections)
             all_wrappers.merge!(c.wrappers)
+
```
rmosolgo, line 61:
```suggestion
```

PR #3038, opened 2020-07-20
Update input_objects.md

vamsipavanmahesh:
Attributes must be inside the input

Here's a valid example that I tested locally

```
mutation {
  updateUser(input: {
    attributes: {
      id: 1,
      username: "updated name",
      email: "bla@chi.com"
    }  
  }){
    user {
      id
      username
      email
    }
  }
}
```

### DIFF
--- a/guides/type_definitions/input_objects.md
+++ b/guides/type_definitions/input_objects.md
@@ -12,7 +12,7 @@ Input object types are complex inputs for GraphQL operations. They're great for 
 
 ```ruby
 mutation {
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the suggestion, but I don't think it applies here. `input:` is a convention applied by `RelayClassicMutation`, but not all input objects are added as `argument :input, ...`. 

If you want to add some docs about `RelayClassicMutation`, you could add a little section to the guide you've modified here, describing that convention, or you could add a section to https://graphql-ruby.org/mutations/mutation_classes.html.

PR #3050, opened 2020-07-28
GraphQL::StaticValidation::Rules::VariablesAreUsedAndDefined - show "anonymous <operation_type>" in error message

### LINKED ISSUE #3047: GraphQL::StaticValidation::VariablesAreUsedAndDefinedError with anonymous operation has no usage or declaration context in message

When writing tests for my GraphQL queries, I noticed that if I received a `GraphQL::StaticValidation::VariablesAreUsedAndDefinedError` error when using an anonymous query, the error message would come back like so:

`Variable $undefinedVar is used by  but not declared` -- I expected to see some indication of what was using the variable similar to the behavior for named operations, rather than an extra space.

I then looked in the code and saw that this was the expected behavior for when `usage.used_by.name` is `nil`, as in the case of an anonymous operation, see: https://github.com/rmosolgo/graphql-ruby/blob/dda92abbbd9fdcddcbe65278a2301fbe3047004c/lib/graphql/static_validation/rules/variables_are_used_and_defined.rb#L143

I would like some indication of what is using the variable, even if it is an unnamed query -- perhaps:
- `Variable $undefinedVar is used by query but not declared`
- `Variable $undefinedVar is used by anonymous query but not declared`

I would also like these semantics to be applied to cases where the variable is declared but not used.

- I've considered surfacing `anonymous <operation_type>` in the case where the operation is anonymous, so the original message would become:

`Variable $undefinedVar is used by anonymous query but not declared`

Extending this to the declared-but-not-used case, we also get:

`Variable $undefinedVar is declared by anonymous query but not used`

**Additional context (links to draft PR in my fork)**
- Draft PR against my fork with proposed changeset
  - https://github.com/jhhb/graphql-ruby/pull/1

jhhb:
* This provides a default behavior for anonymous operation types in `GraphQL::StaticValidation::Rules::VariablesAreUsedAndDefined` when `usage.used_by.name` or `usaged.declared_by.name` is `nil`
* fixes #3047

### DIFF
--- a/spec/graphql/static_validation/validator_spec.rb
+++ b/spec/graphql/static_validation/validator_spec.rb
@@ -32,7 +32,7 @@ describe GraphQL::StaticValidation::Validator do
 
     it "includes message, locations, and fields keys" do
       expected_errors = [{
-        "message" => "Variable $undefinedVar is used by  but not declared",
+        "message" => "Variable $undefinedVar is used by anonymous query but not declared",
         "locations" => [{"line" => 1, "column" => 14, "filename" => "not_a_real.graphql"}],
         "path" => ["query", "cheese", "id"],
         "extensions"=>{"code"=>"variableNotDefined", "variableName"=>"undefinedVar"}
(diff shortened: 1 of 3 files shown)


spec/graphql/static_validation/validator_spec.rb
```diff
@@ -32,7 +32,7 @@
 
     it "includes message, locations, and fields keys" do
       expected_errors = [{
-        "message" => "Variable $undefinedVar is used by  but not declared",
```
rmosolgo, line 35: Oh look, a test for the _wrong_ behavior 😅

PR #3053, opened 2020-07-30
Use equal? to check for DELETE_NODE

jhawthorn:
This is a not code path and was previously calling `AbstractNode#==`, which is slower since it implements some imprecise matching. We should be treating `DELETE_NODE` as a singleton, so `.equal?` should be sufficient here.


rmosolgo:
Nice catch, thanks for the improvement!

PR #3062, opened 2020-08-05
Add `Object.instance_methods` to `CONFLICT_FIELD_NAMES` for warnings

jornvandebeek:
Naming your fields after Object instance methods will cause the resolver to call them on the resolver instance. This is normally not intentional, and confusing. This MR adds `Object.instance_methods` to the `CONFLICT_FIELD_NAMES` so a warning is generated.

There already was a test for the field name :method in spec/graphql/schema/object_spec.rb:341 . I have removed `:method` from the existing GRAPHQL_RUBY_KEYWORDS to ensure the test tests that it comes from Object.instance_methods instead. 

There is an example of the confusion between resolver instance resolution and actual object resolution in the jazz.rb test code, where `object_id` is exposed intentionally but this method is called on the resolver instance instead of the actual object instance. 

I am unsure if the intent of that test ("initializes root wrappers once") still stands. Since there is some 1 to 1 relation between resolver and underlying object, there might not be much of an issue, but this might change in the future if resolver objects are somehow still reused, but the underlying objects are not?

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -53,7 +53,7 @@ module GraphQL
         # resolve method name.
         #
         # @api private
-        CONFLICT_FIELD_NAMES = Set.new(GRAPHQL_RUBY_KEYWORDS + RUBY_KEYWORDS)
+        CONFLICT_FIELD_NAMES = Set.new(GRAPHQL_RUBY_KEYWORDS + RUBY_KEYWORDS + Object.instance_methods)
 
         # Register this field with the class, overriding a previous one if needed.
         # @param field_defn [GraphQL::Schema::Field]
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: Thanks for this improvement!

> if the intent of that test ("initializes root wrappers once") still stands

Yes, it still stands, and I think you've made the right update for it. 

That test is to make sure that given a query like this: 

```graphql
query {
  someField 
  ... SomeFragment 
}

fragment SomeFragment on Query {
  someOtherField 
}
```

We should only initialize `Types::Query` _once_, and that object should be used to resolve _both_ `someField` and `someFragment`. 

I don't expect to reuse object type objects because that object lifecycle would be too complicated. (An old GraphQL-Ruby feature, "GraphQL::Function", had an API like that, and it proved to be a foot-gun.) So much like Rails controllers, the object type objects are used once, then disposed of -- but that single use includes _all_ the selections on the object.

PR #3066, opened 2020-08-07
Revert "Add validation to prevent arguments with both required and default_value"

rmosolgo:
Reverts rmosolgo/graphql-ruby#3011 because that behavior is _actually allowed_ by the spec, see https://github.com/rmosolgo/graphql-ruby/pull/2416 and https://github.com/graphql/graphql-spec/pull/418 and https://github.com/rmosolgo/graphql-ruby/issues/3001#issuecomment-670274075

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -61,11 +61,6 @@ module GraphQL
         @from_resolver = from_resolver
         @method_access = method_access
 
-        if !@null && default_value?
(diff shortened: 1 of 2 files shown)


jfanchime:
Is the feature "arguments with both required and default_value" alive? Still getting such error:
```
GraphQL::Schema::InvalidTypeError:
  Query is invalid: field "transactions" argument "transacted_at" Variable transacted_at of type "RangeFilterDatetime!" is required and will not use the default value. Perhaps you meant to use type "RangeFilterDatetime"
```

PR #3076, opened 2020-08-12
customize ActionCable subscriptions

fdutey:
```ruby
class FooSchema < GraphQL::Schema
  use GraphQL::Subscriptions::ActionCableSubscriptions

  subscription(Foo::Types::SubscriptionType)
end

module Foo
  class SubscriptionType < ::Types::BaseObject
    field :foo_subscription, subscription: Foo::FooSubscription
  end
end

class BarSchema < GraphQL::Schema
  use GraphQL::Subscriptions::ActionCableSubscriptions

  subscription(Bar::Types::SubscriptionType)
end

module Foo
  class SubscriptionType < ::Types::BaseObject
    field :foo_subscription, subscription: Bar::FooSubscription
  end
end

# somewhere, following default configuration with channel in context
#
# FooChannel < ApplicationCable::Channel
# BarChannel < ApplicationCable::Channel

FooSchema.subscriptions.trigger(:foo_subscription, {}, {}, scope: nil) # => transmitting graphql-event:XXX
BarSchema.subscriptions.trigger(:foo_subscription, {}, {}, scope: nil) # => transmitting graphql-event:XXX
```

Multiple schemas will generate the same event stream name if parameters (object / scope) are identical and then will broadcast multiple time the same event / mixing pushes.

**BREAKING CHANGES:**

Anyone using `context[:action_cable_stream]`. But judging by [this line](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions/action_cable_subscriptions.rb#L110), it seems it was broken already anyway.

**How to use**

```ruby
class FooSchema < GraphQL::Schema
  # if not set, still defaults to 'graphql-event:' and 'graphql-subscription:'
  use(GraphQL::Subscriptions::ActionCableSubscriptions, event_prefix: 'graphql:foo:event:',
                                                        subscription_prefix: 'graphql:foo:subscription:')
end

# will transmit to graphql:foo:event:XXX

class BarSchema < GraphQL::Schema
  # if not set, still defaults to 'graphql-event:' and 'graphql-subscription:'
  use(GraphQL::Subscriptions::ActionCableSubscriptions, event_prefix: 'graphql:bar:event:',
                                                        subscription_prefix: 'graphql:bar:subscription:')
end

# will transmit to graphql:bar:event:XXX
```

Cheers

### DIFF
--- a/lib/graphql/subscriptions/action_cable_subscriptions.rb
+++ b/lib/graphql/subscriptions/action_cable_subscriptions.rb
@@ -4,7 +4,7 @@ module GraphQL
     # A subscriptions implementation that sends data
     # as ActionCable broadcastings.
     #
-    # Experimental, some things to keep in mind:
+    # Some things to keep in mind:
     #
     # - No queueing system; ActiveJob should be added
     # - Take care to reload context when re-delivering the subscription. (see {Query#subscription_update?})
@@ -87,11 +87,13 @@ module GraphQL
 
       # @param serializer [<#dump(obj), #load(string)] Used for serializing messages before handing them to `.broadcast(msg)`
       # @param namespace [string] Used to namespace events and subscriptions (default: '')
-      def initialize(serializer: Serialize, namespace: '', **rest)
+      def initialize(serializer: Serialize, namespace: '', action_cable: ActionCable, action_cable_coder: ActiveSupport::JSON, **rest)
         # A per-process map of subscriptions to deliver.
         # This is provided by Rails, so let's use it
         @subscriptions = Concurrent::Map.new
         @events = Concurrent::Map.new { |h, k| h[k] = Concurrent::Map.new { |h2, k2| h2[k2] = Concurrent::Array.new } }
+        @action_cable = action_cable
+        @action_cable_coder = action_cable_coder
         @serializer = serializer
         @transmit_ns = namespace
         super
@@ -102,14 +104,14 @@ module GraphQL
       def execute_all(event, object)
         stream = stream_event_name(event)
         message = @serializer.dump(object)
-        ActionCable.server.broadcast(stream, message)
+        @action_cable.server.broadcast(stream, message)
       end
 
       # This subscription was re-evaluated.
       # Send it to the specific stream where this client was waiting.
       def deliver(subscription_id, result)
         payload = { result: result.to_h, more: true }
-        ActionCable.server.broadcast(stream_subscription_name(subscription_id), payload)
+        @action_cable.server.broadcast(stream_subscription_name(subscription_id), payload)
       end
 
       # A query was run where these events were subscribed to.
@@ -143,7 +145,7 @@ module GraphQL
       #
       def setup_stream(channel, initial_event)
         topic = initial_event.topic
-        channel.stream_from(stream_event_name(initial_event), coder: ActiveSupport::JSON) do |message|
+        channel.stream_from(stream_event_name(initial_event), coder: @action_cable_coder) do |message|
           object = @serializer.load(message)
           events_by_fingerprint = @events[topic]
           events_by_fingerprint.each do |_fingerprint, events|
--- a/lib/graphql/subscriptions/action_cable_subscriptions.rb
+++ b/lib/graphql/subscriptions/action_cable_subscriptions.rb
@@ -82,15 +82,18 @@ module GraphQL
     #   end
     #
     class ActionCableSubscriptions < GraphQL::Subscriptions
+      SUBSCRIPTION_PREFIX = "graphql-subscription:"
+      EVENT_PREFIX = "graphql-event:"
+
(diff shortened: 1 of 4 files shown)


lib/graphql/subscriptions/action_cable_subscriptions.rb
```diff
@@ -82,23 +82,22 @@ class Subscriptions
     #   end
     #
     class ActionCableSubscriptions < GraphQL::Subscriptions
-      SUBSCRIPTION_PREFIX = "graphql-subscription:"
-      EVENT_PREFIX = "graphql-event:"
-
       # @param serializer [<#dump(obj), #load(string)] Used for serializing messages before handing them to `.broadcast(msg)`
-      def initialize(serializer: Serialize, **rest)
+      def initialize(serializer: Serialize, subscription_prefix: 'graphql-subscription:', event_prefix: 'graphql-event:', **rest)
```
rmosolgo, line 86: Why accept two options here? Is there ever a time that an application would need to specify both? 

What if, instead, we accepted: 

```
prefix: ""
```

And added that to `SUBSCRIPTION_PREFIX` and `EVENT_PREFIX` under the hood? 

That way, each schema could specify its prefix, and they'd be in different namespaces, eg: 

```ruby 
use GraphQL::Subscriptions::ActionCableSubscriptions, prefix: "foo"
# ... 
use GraphQL::Subscriptions::ActionCableSubscriptions, prefix: "bar"
```

Would that do the job?

fdutey: > Why accept two options here? Is there ever a time that an application would need to specify both?

There were 2 independent constants so I moved them "brainlessly" into 2 independent kwargs without thinking much to keep it natural to anyone who would know a bit about those internals.

Quite frankly, I think you're actually right, we should just specify one namespace. I feel a bit dumb lol.
Wouldn't really make sense to have them in 2 different namespaces. 

I will make the change. 

PS: I noticed (but my use case might not be exhaustive) that the event might actually be used only once to transmit the subscription confirmation (not even sure) and then after that, everything goes through the subscription. Would you mind enlightening me? Do we need the event at all? Am I missing something?

rmosolgo: > anyone who would know a bit about those internals.

At this point, I think that's pretty much just you and me 😆 . Oh, and maybe the folks who built graphql-anycable. 

> Do we need the event at all? 

Yeah, basically the two keys are used for different kinds of communication: 

- The event key is used for taking `.trigger` calls in one Ruby process and fanning them out to other Ruby processes. - The subscription key is used for pushing specific updates from the server to the client. 

So, with the event key, the first thing that happens is a new subscription sets up a listener on that event key: 

https://github.com/rmosolgo/graphql-ruby/blob/862e8f3ee308079c4bd890b0b592bab75d224bd8/lib/graphql/subscriptions/action_cable_subscriptions.rb#L144

Then, later, when the application code calls `.trigger`, a broadcast is sent using the event key: 

https://github.com/rmosolgo/graphql-ruby/blob/862e8f3ee308079c4bd890b0b592bab75d224bd8/lib/graphql/subscriptions/action_cable_subscriptions.rb#L98-L104

The reason we use ActionCable to implement `.trigger` is because, in a multi-process Ruby deployment, there might be some subscriptions in one process, but a `.trigger` call in another process (say, a backgroud job). We need some way to notify the _other_ server process that _this_ process called `.trigger`, and so it uses ActionCable to do it. 

After that broadcast, the listeners are called, which re-evaluates the graphql query to generate an update payload:

https://github.com/rmosolgo/graphql-ruby/blob/862e8f3ee308079c4bd890b0b592bab75d224bd8/lib/graphql/subscriptions/action_cable_subscriptions.rb#L153

After that, the subscription key is used to send the update payload from the server to the client: 

https://github.com/rmosolgo/graphql-ruby/blob/862e8f3ee308079c4bd890b0b592bab75d224bd8/lib/graphql/subscriptions/action_cable_subscriptions.rb#L110

(There's no Ruby listener for that event; instead, it's a JavaScript client that we expect to be listening.)

Does that help?

fdutey: Ok, the key here is `stream_from(...) do ...`. I didn't entirely understand it as a listener and it stopped me from getting the full picture.
Now I understand, thanks for the explanation :)

fdutey:
```ruby
class BarSchema < GraphQL::Schema
  # if not set, still defaults to 'graphql-event:' and 'graphql-subscription:'
  use(GraphQL::Subscriptions::ActionCableSubscriptions, namespace: 'bar')
end

# will result in transmitting to `graphql-event:bar:XXX` and `graphql-subscriptions:bar:XXXX`
```

rmosolgo:
Thanks again for this improvement!

PR #3083, opened 2020-08-14
Add node nullable option for edge_type

BillYu811:
just add a nullable option in "connections" to make my FE colleagues more happier.

### DIFF
--- a/lib/graphql/types/relay/base_connection.rb
+++ b/lib/graphql/types/relay/base_connection.rb
@@ -48,7 +48,7 @@ module GraphQL
           # It's called when you subclass this base connection, trying to use the
           # class name to set defaults. You can call it again in the class definition
           # to override the default (or provide a value, if the default lookup failed).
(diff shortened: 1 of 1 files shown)


lib/graphql/types/relay/base_connection.rb
```diff
@@ -90,10 +90,16 @@ def visible?(ctx)
...
+          def define_nodes_field(nullable = true)
+            if nullable
+              field :nodes, [@node_type, null: true],
+                null: true,
+                description: "A list of nodes."
+            else
+              field :nodes, [@node_type],
+                null: false,
+                description: "A list of nodes."
+            end
```
rmosolgo, line 102: What do you think of DRYing this a bit? Eg,

```suggestion
          def define_nodes_field(nullable = true)
            field :nodes, [@node_type, null: null],
              null: null,
              description: "A list of nodes."
            end
```

BillYu811:
resolved, review new change, and tell me anything I did wrong, pls. @rmosolgo

BillYu811:
There seems something i don't know caused the test CI to fail. @rmosolgo

PR #3098, opened 2020-08-20
Use module_namespacing for graphql templates

adrianna-chang-shopify:
Currently, running the generators from within an engine produces files that are not namespaced properly.

For example, if you run `rails g graphql:scalar Date` in the engine `blorgh`, you'll end up with a new ruby file that looks like:

```ruby
module Types
  class Date < Types::BaseScalar
  end
end
```

It should look like this:

```ruby
module Blorgh
  module Types
    class Date < Types::BaseScalar
    end
  end
end
```

Wrapping the template contents in [#module_namespacing](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/base.rb#L288-L292) provides the correct namespace in a generator is called from within an engine.

🎩  Results:
Screen Shot 2020-08-20 at 4 17 02 PM

### DIFF
--- a/lib/generators/graphql/core.rb
+++ b/lib/generators/graphql/core.rb
@@ -41,6 +41,14 @@ module Graphql
         end
       end
 
+      def module_namespacing_when_supported
+        if defined?(module_namespacing)
+          module_namespacing { yield }
+        else
+          yield
+        end
+      end
+
       private
 
       def schema_name
--- a/lib/generators/graphql/templates/base_argument.erb
+++ b/lib/generators/graphql/templates/base_argument.erb
@@ -1,4 +1,6 @@
+<% module_namespacing do -%>
 module Types
   class BaseArgument < GraphQL::Schema::Argument
   end
 end
+<% end -%>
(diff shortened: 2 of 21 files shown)


tadeusz-niemiec:
Dunno why the action cable test failed, it's really unrelated to my changes :(

adrianna-chang-shopify:
Hey @rmosolgo! Thanks for taking a look ❤️  Yes, my bad - looks like `#module_namespacing` was moved from `Rails::Generators::NamedBase` to `Rails::Generators::Base` in rails `v5.2.0`, which means any Rails versions prior to that won't have that method. I'll add in the change to check if we have access to the method! 👍

adrianna-chang-shopify:
[This build](https://travis-ci.com/github/rmosolgo/graphql-ruby/jobs/376895979) still seems to be failing (for unrelated reasons), and passes for me locally 🤔

tadeusz-niemiec:
> [This build](https://travis-ci.com/github/rmosolgo/graphql-ruby/jobs/376895979) still seems to be failing (for unrelated reasons), and passes for me locally

I just had the same issue, this test sometimes fails, I've just force-pushed an empty commit to my PR (couldn't find a way to re-run this test) and it passed :)

lib/generators/graphql/core.rb
```diff
@@ -41,6 +41,14 @@ def create_dir(dir)
         end
       end
 
+      def module_namespacing_when_supported
+        if defined?(module_namespacing)
```
rmosolgo, line 45: TIL `defined?` for instance methods!

PR #3101, opened 2020-08-22
Fix incorrect documentations

pocke:
This pull request contains two documentation fixes.

* `Warden#initialize` doesn't receive `deep_check` argument.
* `Warden#types` actually returns a Hash, but the documentation said "it returns a Array".


rmosolgo:
Thanks for the fixes!

PR #3107, opened 2020-08-24
Rescue SystemStackErrors during validation

### LINKED ISSUE #3089: Deeply nested query raises 'stack level too deep' even if max query depth is enforced
**Describe the bug**

Though a max query depth is enforced, executing this [query](https://gist.github.com/eapache/a4bef04ded8a624d2a41c27e36fbe2d4) raises the below error in the gem.

Screen Shot 2020-08-17 at 3 43 36 PM

**Versions**

`graphql` version: 1.10.14

graphql-batch : 0.4.3
Max query depth enforced is 17

**Steps to reproduce**

Execute the above mentioned query

**Expected behavior**

Expected an error message stating that the query depth exceeds max query depth

**Actual behavior**

SystemStackError (stack level too deep) is raised during validation phase

Screen Shot 2020-08-05 at 2 50 53 PM

rmosolgo:
Even with `max_depth ...`, if a query is too big, it will raise a stack error. 

This is the easiest, most catch-all fix. (It covers all of validation, regardless of `max_depth ...`.) 

The handling of this error can be customized in your schema definition, eg: 

```ruby 
  def self.query_stack_error(query, err)
    raise err # to raise the original Ruby error 
  end 
end 
```

Although I think this fix is appropriate in any case, more subtle changes might be possible: 

- Change `MaxQueryDepth` to _halt_ analysis when the max depth is exceeded. We'd have to change the error message, which currently contains the actual query depth. We could also add a hard cap to MaxQueryDepth (configurable, default 500?). TODO: would validation still raise a stack error?
- Refactor validation/analysis to be less stack-dependent. The current visitor implementation uses `super` to visit child nodes, but it's possible that another implementation could visit the tree without building a big Ruby stack. But this would only be useful if there's a _good reason_ to inspect a GraphQL query whose depth is greater than Ruby's stack limit!

If anyone's interested in exploring those (or another solution) further, let me know

Fixes #3089

### DIFF
--- a/spec/graphql/analysis/ast/max_query_depth_spec.rb
+++ b/spec/graphql/analysis/ast/max_query_depth_spec.rb
@@ -131,4 +131,20 @@ describe GraphQL::Analysis::AST::MaxQueryDepth do
       assert_equal "Query has depth of 7, which exceeds max depth of 4", result.message
     end
   end
+
+  describe "when the query would cause a stack error" do
+    let(:query_string) {
+      str = "query { cheese(id: 1) { ".dup
+      n = 10_000
+      n.times { str << "similarCheese(source: SHEEP) { " }
+      str << "id "
+      n.times { str << "} " }
+      str << "} }"
+      str
+    }
+
+    it "returns an error" do
+      assert_equal ["This query is too large to execute."], query.static_errors.map(&:message)
+    end
+  end
 end
--- a/lib/graphql/query/validation_pipeline.rb
+++ b/lib/graphql/query/validation_pipeline.rb
@@ -90,6 +90,9 @@ module GraphQL
         end
 
         @valid = @validation_errors.empty?
+      rescue SystemStackError => err
+        @valid = false
+        @schema.query_stack_error(@query, err)
       end
 
       # If there are max_* values, add them,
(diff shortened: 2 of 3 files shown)


spec/graphql/analysis/ast/max_query_depth_spec.rb
```diff
@@ -131,4 +131,20 @@
...
+      n = 10_000
+      n.times { str << "similarCheese(source: SHEEP) { " }
+      str << "id "
+      n.times { str << "} " }
+      str << "} }"
+      str
+    }
+
+    it "returns an error" do
+      assert_equal ["This query is too large to execute."], query.static_errors.map(&:message)
```
swalkinshaw, line 147: I'm assuming this is meant to end up in the result? The ends up as `{"errors"=>[]}` though

https://github.com/rmosolgo/graphql-ruby/blob/da513ac4a3ad3b2d86587180474ae9f96863bc41/lib/graphql/execution/multiplex.rb#L128

At this point, `static_errors` (and `analysis_errors`) appear to be reset back to `[]` but I couldn't track down where that happens.

rmosolgo: 🤦

PR #3111, opened 2020-08-25
Add 'replace' argument option

### LINKED ISSUE #3080: Add `replace_null` to optional arguments to fall back to default

The GraphQL spec does not distinguish between null and optional arguments, so if an argument is optional it is possible to just pass null instead. `graphql-ruby` then treats this as the actual value being passed, so you have to explictly check for null in those cases (as outlined in #2864). The default argument is ignored:

```ruby
field :echo, String do 
  argument :input, String, optional: true, default_value: ''
end

def echo(input: )
  if input.nil?
    input = false 
  end 
  # ...
end 
```

If you use a bunch of optional fields (for example when using [graphql-search-object](https://github.com/RStankov/SearchObjectGraphQL)) adding a check like above can become *quite* cumbersome.

An option on the argument would reduce overhead. `replace_null` was already suggested in [this comment](https://github.com/rmosolgo/graphql-ruby/issues/2864#issuecomment-610011047). 

The argument definition would then look like:

```
field :echo, String do 
  argument :input, String, optional: true, default_value: '', replace_null: true
end
```

The behavior would be such that if a null value is passed it is replaced by the default value. If the `default_value`  is not set but `replace_null` is, we may raise an error or fall back to just passing null anyways.

Using the `prepare` hook is another option, but that might be deprecated soon (?), see #2749. 

I've also considered if it makes sense to change the option a bit, so more configuration is possible. For example, to treat blank values as 'empty' as well (thus falling back to the default value). Something like this: 

```ruby
argument :input, String, optional: true, default_value: 'default', replace: :null # replace only null values with default
argument :input, String, optional: true, default_value: 'default', replace: :blank # replace blank values with default as well
```

I would actually prefer this variant, as it does not seem to be that much additional effort, and covers additional use cases as well (like dealing with empty arrays)

tadeusz-niemiec:
Resolves #3080 

It's now possible to decide what type of provided values should be replaced by the `replace` flag, which can be one of the following `:none, :nil, :blank`


rmosolgo:
👋  Hey, sorry I missed this when you first opened it! I agree this is a nice feature, and I'm game to include it as documented here. 

Do you mind updating this PR with master? (I don't care if you rebase or merge.) I think it will require you to _also_ implement `replace:...` for GraphQL::Execution::Interpreter. If you're using the legacy runtime, it's fine with me to also keep the implementation you have here -- or you can remove it.

tadeusz-niemiec:
> 👋 Hey, sorry I missed this when you first opened it! I agree this is a nice feature, and I'm game to include it as documented here.
> 
> Do you mind updating this PR with master? (I don't care if you rebase or merge.) I think it will require you to _also_ implement `replace:...` for GraphQL::Execution::Interpreter. If you're using the legacy runtime, it's fine with me to also keep the implementation you have here -- or you can remove it.

Hey, no worries, I'll update this. I'm już a bit busy this week, so I will take care of this in the next one :)

fdsEmn:
> Resolves #3080
> 
> It's now possible to decide what type of provided values should be replaced by the default value.
> The `replace` flag can be one of the following `:none, :nil, :blank`

PR #3112, opened 2020-08-25
Add gql devise to related_projects.md

00dav00:
Authentication is a common feature in most applications and Devise does great work providing this set of functionalities in full stack ruby world. When it comes to API's there are a number of of gems that modify Devise to work with REST endpoints, like devise Token Authentication for example, but with @mcelicalderon we couldn't find a gem that provides a graphql interface for devise and fit our needs so we created [graphql_devise](https://github.com/graphql-devise/graphql_devise) and have been working on it for almost a year now.

The gem has around 30 releases at the moment and we would love if it can be added to this gem related projects. If you have any reservations we would be glad to address them, just let us know!


rmosolgo:
Thanks for the addition!

PR #3117, opened 2020-08-29
Update instructions for Ably Webhook configuration

rgraff:
Adding some clarification since this tripped me up. The "Channel Lifecycle"  source option is disabled until you set the request mode to "Batch request"


rmosolgo:
Awesome, thanks the update here!

PR #3122, opened 2020-08-31
Do not wrap connections when RawValue is returned from resolver

DmitryTsepelev:
This is a followup to PR #2699, that allows returning `raw_value` from connection resolvers. In this case interpreter should stop any further processing and use the provided value (right now the value wrapped into `RawValue` will reasonably raise `Couldn't find a connection wrapper for RawValue ...`).

Example:

```ruby
  field :feed, Types::Event.connection_type, null: false

  def feed
    return raw_value(value) if (value = read_cache)

    write_cache do
      Event.where(language_id: language_id, category_id: category_id)
    end
  end
end
```

### DIFF
--- a/lib/graphql/pagination/connections.rb
+++ b/lib/graphql/pagination/connections.rb
@@ -66,6 +66,8 @@ module GraphQL
       # Used by the runtime to wrap values in connection wrappers.
       # @api Private
       def wrap(field, parent, items, arguments, context, wrappers: all_wrappers)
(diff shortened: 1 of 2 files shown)


DmitryTsepelev:
Btw seems like same specs fail on master

rmosolgo:
Makes sense to me, thanks for the fix! 

Yeah, I've had a hard time getting those ActionCable tests to work properly on CI. I can't get them to fail locally 😖

Issue #3123, opened 2020-08-31
Schema instance possible types behaves different than class method


grcooper:
**Describe the bug**

The two methods for `possible_types` do not behave the same on the schema instance vs the schema class.

The schema instance method uses the class `PossibleTypes` [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema.rb#L555).

The schema class method uses a built in hashmap [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema.rb#L1109). 

This has some correct logic for possible types on unions, but interfaces are not handled the same way they were previously. The main difference being that `context` is not taken into account for the interfaces which causes some issues for interface type membership visibility.

**Versions**

`graphql` version: `1.11.4`
`rails` (or other framework): 
other applicable versions (`graphql-batch`, etc)

**GraphQL schema**

```ruby
class TypeMembership < TypeMembershipVisibility
  def visible?(context)
    context[:private]
  end
end

module MyInterface
   include GraphQL::Schema::Interface
  type_membership_class TypeMembership

  field :test, :string, null: false
end

class Product < GraphQL::Schema::Object
  implements MyInterface

  field :id, ID, null: false
end

class QueryRoot < GraphQL::SchemaObject
  field :my_interfaces, [MyInterface], null: false
end

class ApplicationSchema < GraphQL::Schema
  query QueryType
end
```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
query {
  myInterfaces { test  ... on Product { id } }
}
```

Assuming we pass through `context = { private: false }`:

Expected:

```json
{
  "errors": [{ "Object type Product cannot be spread inside of myInterfaces }]
}
```

Actual:

```json
{
  "data": {
    "myInterfaces": [ { "test": "test", "id": 1234 }]
  }
}
```

**Additional context**

I believe this is a bug because of the different behaviour between what is expected in the instance method vs the class method. The code around the class method `possible_types` on Schema is a bit confusing and I didn't want to dive into it too hard and add `context` without first getting your opinion on if this should be solved and is indeed a bug.

rmosolgo:
Yes, agreed that it's a bug, and it should work the way you expected :+1:

PR #3124, opened 2020-09-01
Update possible types on schema class for interfaces

grcooper:
To fix: https://github.com/rmosolgo/graphql-ruby/issues/3123

Fixed up the `possible_types` method on the `Schema` class object. Previously this was not taking into account the `context` which meant that some interfaces could be shown when they shouldn't be.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1117,6 +1117,8 @@ module GraphQL
               next true unless type.kind.interface?
               next true unless possible_type.kind.object?
 
+              # Use `.graphql_name` comparison to match legacy vs class-based types. 
+              # When we don't need to support legacy `.define` types, use `.include?(type)` instead.
               possible_type.interfaces(context).any? { |interface| interface.graphql_name == type.graphql_name }
             end if stored_possible_types
             visible_possible_types ||
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1112,7 +1112,14 @@ module GraphQL
           if type.kind.union?
             type.possible_types(context: context)
           else
-            own_possible_types[type.graphql_name] ||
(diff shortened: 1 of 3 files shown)


lib/graphql/schema.rb
```diff
@@ -1112,7 +1112,13 @@ def possible_types(type = nil, context = GraphQL::Query::NullContext)
           if type.kind.union?
             type.possible_types(context: context)
           else
-            own_possible_types[type.graphql_name] ||
+            visible_possible_types = own_possible_types[type.graphql_name]&.select do |possible_type|
+              next true unless type.kind.interface?
+              next true unless possible_type.kind.object?
+
+              possible_type.interfaces(context).any? { |interface| interface.graphql_name == type.graphql_name }
```
grcooper, line 1119: I had to use `graphql_name` for comparison here rather than just the type because of: https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/interface_type_spec.rb#L10

This broke old tests because the own possible types would return `GraphQL::Schema::Interface` objects, while we were passing in a `GraphQL::InterfaceType` class to the above `possible_types` method (this also failed for a few other tests in this file). 

If we don't care about these tests as this is legacy behaviour then I would suggest changing this line to:

`possible_type.interfaces(context).include?(type)`

rmosolgo: Yeah, I doubt it's much overhead to keep legacy support in there. Let's just add a comment so that we _might_ remember to simplify it once it's possible: 

```suggestion
              # Use `.graphql_name` comparison to match legacy vs class-based types. 
              # When we don't need to support legacy `.define` types, use `.include?(type)` instead.
              possible_type.interfaces(context).any? { |interface| interface.graphql_name == type.graphql_name }
```

PR #3125, opened 2020-09-02
Avoid redundant loop through rules_to_use in Validator#validate

nvh0412:
Currently, rules attribute has fallback value which is ALL_RULES and this array contains 27 modules (latest version). None of them is Class and has `validate` method. So for this case, loop through this Module array was a complete waste of time and resources.

To deal with legacy rules which come from GraphQL pro. We should compare `rule_to_use` and ALL_RULES array before we decide to start the loop

### DIFF
--- a/lib/graphql/static_validation/validator.rb
+++ b/lib/graphql/static_validation/validator.rb
@@ -32,10 +32,13 @@ module GraphQL
 
             context = GraphQL::StaticValidation::ValidationContext.new(query, visitor_class)
 
-            # Attach legacy-style rules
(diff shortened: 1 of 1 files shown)


rmosolgo:
> which come from GraphQL pro

Yeah, this was also a public API, so applications might still be providing custom validators. I've added this to the list of things to deprecate in 1.12.0 :+1: 

nvh0412:
Thanks @rmosolgo , I didn't add more context test for my stuff because our test covered `custom validators` context already. Please let me know if you want to add more test for this PR.

PR #3126, opened 2020-09-02
Fix some grammar on ast_analysis page

dkassen:


rmosolgo:
Thanks for the fixes!

PR #3015, opened 2020-07-09
WIP: Input field deprecation support

jturkel:
This PR implements the [RFC](https://github.com/graphql/graphql-spec/pull/525) for input field deprecation support. The RFC was moved to stage 2 at the last [working group](https://docs.google.com/document/d/1vw5zEOHVPBtspoWRVQ279NSo_JfWg7ZEdGKix4O-nco/edit#heading=h.w19vr9farz2x). I don't think we want to merge this PR until the corresponding graphql-js [PR](https://github.com/graphql/graphql-js/pull/1560) merges (or the RFC moves to the accepted state) but I wanted to open a WIP PR in case there's an early feedback on my implementation and to prevent anyone from doing duplicate work.

PR #3133, opened 2020-09-04
Improve error messages for invalid input types

jturkel:
Error messages when recursively validating a field argument's type don't actually include the input type's name which makes them a bit confusing. Consider the following invalid schema:

```graphql
type MyType {
  myField(myArg: MyInput)
}

input MyInput {
  # This input field is invalid because it has an output type
  someInput: MyOtherType
}

type MyOtherType {
  foo: String
}
```
This generates the following error message that doesn't mention the `MyInput` type:
```
Argument myArg on MyType.myField is invalid: argument "someInput" type must be a valid input type (Scalar or InputObject), not GraphQL::ObjectType (MyOtherType)
```
With the proposed changes the error message becomes:
```
MyInput is invalid: argument "someInput" type must be a valid input type (Scalar or InputObject), not GraphQL::ObjectType (MyOtherType)
```

### DIFF
--- a/lib/graphql/schema/traversal.rb
+++ b/lib/graphql/schema/traversal.rb
@@ -139,7 +139,7 @@ Some late-bound types couldn't be resolved:
           prev_type = @type_map[type_defn.name]
           # Continue to visit this type if it's the first time we've seen it:
           if prev_type.nil?
-            validate_type(type_defn, context_description)
+            validate_type(type_defn)
             @type_map[type_defn.name] = type_defn
             case type_defn
             when GraphQL::ObjectType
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: This change is good by me, I just want to point out that it _won't_ apply to class-based types using the interpreter (which will someday be the only way to run graphql-ruby). I don't think it uses `traversal.rb` in that case. I recently added #3120 for a similar case, but maybe you're still using `.define`?

jturkel:
Doh. I first noticed the problem when adding validation that deprecated arguments aren't required in #3015. The tests in question use the new class based API but I guess many of them call `GraphQL::Schema.graphql_definition` which triggers the legacy validation. I can't wait till the legacy APIs are removed in 2.0 😄 

I'm going to close this PR since we're not using the legacy API. I'll open up a new PR that properly handles validation of required deprecated arguments in the class based API.

PR #3134, opened 2020-09-06
Improve Lexer

nvh0412:
Improve our GraphQL::Lexer

- [x] Using empty? instead of length.zero?

```
Benchmark.ips do |x|
  x.report("length.zero") { !"".length.zero? }
  x.report("empty?") { !"".empty? }
  x.compare!
end

         length.zero     1.395M i/100ms
              empty?     1.735M i/100ms
         length.zero     14.010M (± 1.7%) i/s -     71.165M in   5.080898s
              empty?     17.019M (± 1.9%) i/s -     86.746M in   5.099024s

Comparison:
              empty?: 17018973.1 i/s
         length.zero: 14010232.6 i/s - 1.21x  (± 0.00) slower

```

- [x] Try to reuse PACK_DIRECTIVE constant for "c*"

### DIFF
--- a/lib/graphql/language/lexer.rl
+++ b/lib/graphql/language/lexer.rl
@@ -138,7 +138,7 @@ module GraphQL
       %% write data;
 
       def self.run_lexer(query_string)
-        data = query_string.unpack("c*")
+        data = query_string.unpack(PACK_DIRECTIVE)
         eof = data.length
 
         # Since `Lexer` is a module, store all lexer state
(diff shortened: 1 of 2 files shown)


lib/graphql/language/lexer.rl
```diff
@@ -138,7 +138,7 @@ module GraphQL
       %% write data;
 
       def self.run_lexer(query_string)
-        data = query_string.unpack("c*")
```
rmosolgo, line 141: ~~The generated output from this file has `frozen_string_literal: true`, so I would _expect_ this string to be reused~~

Nevermind, I see that this is already being used elsewhere for consistency, not performance :+1:

PR #3135, opened 2020-09-07
add node_nullable option to RelayShortcuts#connection_type

QWYNG:
Thank you for maintaining a great library.

#3083 has eased the burden of the null check.
I hope to be able to use this option in shortcuts as well.

### DIFF
--- a/lib/graphql/schema/member/relay_shortcuts.rb
+++ b/lib/graphql/schema/member/relay_shortcuts.rb
@@ -31,13 +31,13 @@ module GraphQL
           end
         end
 
-        def connection_type
+        def connection_type(node_nullable: true)
           @connection_type ||= begin
             conn_name = self.graphql_name + "Connection"
             edge_type_class = self.edge_type
             Class.new(connection_type_class) do
               graphql_name(conn_name)
-              edge_type(edge_type_class)
+              edge_type(edge_type_class, node_nullable: node_nullable)
             end
           end
         end
--- a/spec/graphql/schema/member/relay_shortcut_spec.rb
+++ b/spec/graphql/schema/member/relay_shortcut_spec.rb
@@ -1,0 +1,28 @@
+require "spec_helper"
+
+describe GraphQL::Schema::Member::RelayShortcuts do
+  describe 'connection_type' do
+    module NonNullAbleConnectionShortcutsDummy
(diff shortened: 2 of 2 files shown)


lib/graphql/schema/member/relay_shortcuts.rb
```diff
@@ -31,13 +31,13 @@ def edge_type
           end
         end
 
-        def connection_type
+        def connection_type(node_nullable: true)
```
rmosolgo, line 34: The weird thing about this is: 

- We accept `node_nullable:` as true or false;
- We cache the generated type after the first call; 
- But if a later call includes a _different_ `node_nullable` value, we return the already-generated type (even though it was generated using a different `node_nullable` value. 

I don't think the library should behave like that because it'd be very unexpected behavior to pass one input to a method, but get a return value as though a different input was provided. 

Here's how you could do something like this now: 

```ruby 
# Make a custom base connection class that defaults to non-null nodes
class Types::BaseConnection < GraphQL::Types::Relay::BaseConnection 
  # set the default to false here 
  def self.edge_type(*args, node_nullable: false, **kwargs)
    super 
  end 
end 

# ... 
# Then hook it up to other base classes: 

class Types::BaseObject < ... 
  connection_type_class(Types::BaseConnection)
end 

class Types::BaseUnion < ...
  connection_type_class(Types::BaseConnection)
end 

module Types::BaseInterface 
  # ...
  connection_type_class(Types::BaseConnection)
end 
```

That way, the `.connection_type` helper will always generate a connection with `node_nullable: false`. 

Other options might be: 

- Caching the value of the first `node_nullable:` and raising an error if a connection type is generated with a different value. 
- Caching one generated type _per_ `node_nullable:` value (extending the cache to a Hash of `{ node_nullable_value => type_class }`, or using two instance variables)

Personally, I think those two are more trouble than they're worth. In my experience, adding a base class pays of in the long run. 

What do you think of those approaches?

QWYNG: I see. I wasn't aware of the possibility that another input could be made after the type cache was generated.

>Personally, I think those two are more trouble than they're worth.

I agree with you.
I'll close this PR. 
Thank you for your comment!

PR #3137, opened 2020-09-08
Validate that deprecated requirements aren't required

jturkel:
Following up on the discussion in #3133, this adds validation to prevent deprecating required arguments in the class based API. The validation added in #3015 only covered the legacy API.

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -61,7 +61,7 @@ module GraphQL
         @ast_node = ast_node
         @from_resolver = from_resolver
         @method_access = method_access
-        @deprecation_reason = deprecation_reason
+        self.deprecation_reason = deprecation_reason
 
         if definition_block
           if definition_block.arity == 1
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/argument.rb
```diff
@@ -164,6 +165,11 @@ def to_graphql
 
       def type=(new_type)
         validate_input_type(new_type)
+        # This isn't true for LateBoundTypes, but we can assume those will
+        # be updated via this codepath later in schema setup.
+        if new_type.respond_to?(:non_null?)
```
jturkel, line 170: Out of curiosity what's the expected relationship between `new_type.non_null?` and `@null`? Think we should validate that the two are consistent?

rmosolgo: I'm not sure... There are hacks that work like `argument(name, type_string, required: false)`, where type_string may include `!`. For example, that's how parsing a schema from SDL works; The SDL loader code just passes the type string along as-is, with `required: false`: 

https://github.com/rmosolgo/graphql-ruby/blob/91e7e3461871725d70c6abc719799e40c4a94f52/lib/graphql/schema/build_from_definition.rb#L261-L262

As for `@null`, I consider it a private implementation detail of `Argument`. To find out if an argument's type is non-null, you can call `argument.type.non_null?` (with the exception noted here: GraphQL internal types don't have that available right away -- but by the time you run a query, `argument.type.non_null?` will work).

rmosolgo:
The dang integration test is failing, but :+1:. I pushed a commit to cover types defined by strings.

rmosolgo:
Thanks for the improvement!

PR #3138, opened 2020-09-09
Freeze Interpreter::Arguments instances

### LINKED ISSUE #3004: Make Interpreter::Arguments immutable

We want to cache arguments during execution, but it's hard to tell when field extensions will modify them. See https://github.com/rmosolgo/graphql-ruby/pull/3003

Cache immutable objects; let those be copied when modified, so that the overhead isn't incurred by default.

rmosolgo:
Fixes #3004 , addresses issues described in #3003 

TL;DR: it's possible to make mistakes where objects from the arguments cache are reused, even after they're modified in field-specific ways. This will make that impossible, since arguments objects can't be modified.

It could well be a breaking change, because arguments hashes won't play nicely with `.delete` anymore (as shown in updated test here, same goes for other destructive Hash methods).

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -468,9 +468,7 @@ module GraphQL
         end
 
         def arguments(graphql_object, arg_owner, ast_node)
-          # Don't cache arguments if field extras or extensions are requested since they can mutate the argument data structure
-          if arg_owner.arguments_statically_coercible? &&
-              (!arg_owner.is_a?(GraphQL::Schema::Field) || (arg_owner.extras.empty? && arg_owner.extensions.empty?))
+          if arg_owner.arguments_statically_coercible? && !arg_owner.is_a?(GraphQL::Schema::Field)
             query.arguments_for(ast_node, arg_owner)
           else
             # The arguments must be prepared in the context of the given object
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -468,7 +468,7 @@ module GraphQL
         end
 
         def arguments(graphql_object, arg_owner, ast_node)
-          if arg_owner.arguments_statically_coercible? && !arg_owner.is_a?(GraphQL::Schema::Field)
+          if arg_owner.arguments_statically_coercible?
             query.arguments_for(ast_node, arg_owner)
           else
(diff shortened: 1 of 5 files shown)


lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -180,32 +180,40 @@ def evaluate_selections(path, scoped_context, owner_object, owner_type, selectio
...
-                    ast_nodes: field_ast_nodes,
-                    field: field_defn,
-                  )
-                when :argument_details
-                  kwarg_arguments[:argument_details] = resolved_arguments
-                else
-                  kwarg_arguments[extra] = field_defn.fetch_extra(extra, context)
                 end
+
+                resolved_arguments = resolved_arguments.merge_extras(extra_args)
```
jturkel, line 213: Should we remove the special handling of field extras in the `arguments` function since we're no longer mutating the arguments hash for that case? Also am I correct in assuming that the field extensions case in the `arguments` function no longer applies since field extensions can no longer mutate the arguments hash?

rmosolgo: Ahh, thanks for the reminder of that. I removed some special handling in f66a7f5b6, does that look right to you?

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -460,9 +468,7 @@ def after_lazy(lazy_obj, owner:, field:, path:, scoped_context:, owner_object:,
         end
 
         def arguments(graphql_object, arg_owner, ast_node)
-          # Don't cache arguments if field extras or extensions are requested since they can mutate the argument data structure
-          if arg_owner.arguments_statically_coercible? &&
-              (!arg_owner.is_a?(GraphQL::Schema::Field) || (arg_owner.extras.empty? && arg_owner.extensions.empty?))
+          if arg_owner.arguments_statically_coercible? && !arg_owner.is_a?(GraphQL::Schema::Field)
```
jturkel, line 471: Should we remove the `!arg_owner.is_a?(GraphQL::Schema::Field)` check? I believe that was only needed to ensure we could call `#extras` and `#extensions`.

rmosolgo: Oh, I forgot. Let's see .... 

```suggestion
          if arg_owner.arguments_statically_coercible?
```

jturkel (approved):
LGTM 🎉

PR #3140, opened 2020-09-09
Fix ActionCable integration test

rmosolgo:
Sadly, I don't actually have a plan to fix the test. 

But I've been hacking on it, trying to understand _why_ it fails on CI, and I'm gonna figure this out one way or another!


rmosolgo:
4 passing builds in a row (excepting the ones that had a typo), I'm gonna call it fixed and ... keep watching 😬

PR #3141, opened 2020-09-09
RelayClassicMutation: set description for auto-generated input argument

parkr:
We automatically generate a description for the auto-generated input type, but we don't assign that description to the mutation's argument for the new input type.

This updates the code to copy the description from the input type to the argument on the mutation.

This was prompted by the graphql-schema-linter tool saying the argument `mutationName.input` doesn't have a description.

### DIFF
--- a/lib/graphql/schema/relay_classic_mutation.rb
+++ b/lib/graphql/schema/relay_classic_mutation.rb
@@ -105,7 +105,7 @@ module GraphQL
           sig = super
           # Arguments were added at the root, but they should be nested
           sig[:arguments].clear
-          sig[:arguments][:input] = { type: input_type, required: true, description: input_type.description }
+          sig[:arguments][:input] = { type: input_type, required: true, description: "Parameters for #{graphql_name}" }
           sig
         end
 
--- a/lib/graphql/schema/relay_classic_mutation.rb
+++ b/lib/graphql/schema/relay_classic_mutation.rb
@@ -105,7 +105,7 @@ module GraphQL
           sig = super
           # Arguments were added at the root, but they should be nested
           sig[:arguments].clear
-          sig[:arguments][:input] = { type: input_type, required: true }
+          sig[:arguments][:input] = { type: input_type, required: true, description: input_type.description }
           sig
         end
 
--- a/spec/graphql/schema/relay_classic_mutation_spec.rb
+++ b/spec/graphql/schema/relay_classic_mutation_spec.rb
@@ -36,6 +36,16 @@ describe GraphQL::Schema::RelayClassicMutation do
     end
   end
 
+  describe "input argument" do
+    it "sets a description for the input argument" do
+      mutation = Class.new(GraphQL::Schema::RelayClassicMutation) do
+        graphql_name "SomeMutation"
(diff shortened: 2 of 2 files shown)


parkr:
CI failure seems unrelated.

parkr:
@rmosolgo Hey! I have a graphql linter which insists on descriptions for things and this patch fixed a problem I was having with the auto-generated input types. Let me know what else I can do to make this patch acceptable for merging. Thank you!

lib/graphql/schema/relay_classic_mutation.rb
```diff
@@ -105,7 +105,7 @@ def field_options
           sig = super
           # Arguments were added at the root, but they should be nested
           sig[:arguments].clear
-          sig[:arguments][:input] = { type: input_type, required: true }
+          sig[:arguments][:input] = { type: input_type, required: true, description: input_type.description }
```
rmosolgo, line 108: It seems like this will generate: 

```
someMutation(
  input: SomeInput # Autogenerated input type of SomeMutation
 )
 ```
 
 But that description is for the _type_, not the argument. How about using an argument-specific description instead? For example:
 
```suggestion
          sig[:arguments][:input] = { type: input_type, required: true, description: "Parameters for #{graphql_name}" }
```

rmosolgo (commented):
Hey @parkr, sorry I overlooked this before. Thanks for checking back in. Besides updating the description, do you mind adding a test to make sure this works as we expect (and _keeps_ working as we expect!)? 

For example, here: 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/schema/relay_classic_mutation_spec.rb#L4

You could add 

```
mutation = Class.new(GraphQL::Schema::RelayClassicMutation) do 
  graphql_name "SomeMutation"
end 

assert_equal "...", mutation.field_options[:arguments][:input][:description]
```

parkr:
@rmosolgo Thank you for the review! Unfortunately, I cannot seem to get GitHub Actions to fire on this branch... Would you like me to re-create this pull request? Actions does seem to run on the `master` branch of my fork.

```text
~/code/graphql-ruby#patch-1$ BUNDLE_GEMFILE=gemfiles/rails_6.0.gemfile bundle exec ruby -Ilib:spec spec/graphql/schema/relay_classic_mutation_spec.rb

# Running tests with run options --seed 12476:

......................

Finished tests in 0.036015s, 610.8566 tests/s, 1110.6483 assertions/s.


22 tests, 40 assertions, 0 failures, 0 errors, 0 skips
```

As a side note, it took me a little while to figure out how to run the tests locally. Would you be interested in something like `script/cibuild` or `script/test` (or something similar) for running the tests? I spent a while running `rspec <file>` since the tests are named with `_spec.rb`, and took a while to find the above incantation. I newcomer to this project or Ruby would have trouble figuring out how to get their local development environment setup properly!

rmosolgo:

Sorry about the trouble running the tests. For future reference, the docs on local development are here: https://graphql-ruby.org/development#running-the-tests. No, I'm not interested in maintaining those scripts, but thanks for offering.

PR #3145, opened 2020-09-10
Default to Interpreter, Analysis::AST, Execution::Errors and Pagination::Connections

rmosolgo:
These modules have been in development for a long time, and the recommended approach for a long time. Let's make them the default. 

Technically, it's still possible to choose the old behaviors: 

```ruby 
use GraphQL::Execution::Execute 
use GraphQL::Analysis
```

But this PR decreases test coverage for those, and 1.12 will also deprecate them with warnings. 

Also, `Schema.from_definition` no longer accepts an `interpreter:` option; it is always `interpreter: true`.


lib/graphql/tracing/appoptics_tracing.rb
```diff
@@ -112,6 +112,8 @@ def metadata(data, layer)
             graphql_query_string(data[key])
           when :multiplex
             graphql_multiplex(data[key])
+          when :path
+            [key, data[key].join(".")]
```
rmosolgo, line 116: I wonder if this ever worked on the interpreter. It was returning `[:path, []]`, and `[]` was flattened out of existence by the `.flatten(2)` call below, then `.each_slice(2)` raised an error about getting an odd-length array.

PR #3147, opened 2020-09-10
Implement GraphQL::Execution::Interpreter::Arguments#fetch

jturkel:
`GraphQL::Execution::Interpreter::Arguments` is a hash-like class so this PR adds the `fetch` method.


rmosolgo:
Nice, thanks!

PR #3151, opened 2020-09-12
Refactor the constructor of Schema::Argument

nvh0412:
To prepare for some new Rails style validation keywords, we should refactor the constructor of Schema::Argument a bit. Using splat keyword argument to make it shorter


rmosolgo:
No thanks, I prefer using the Ruby primitive here because it gives user-friendly error messages and it can be used to generate docs (eg https://graphql-ruby.org/api-doc/1.11.4/GraphQL/Schema/Argument.html#initialize-instance_method). Someday, I'd like to adopt Sorbet, then we'd even get _typechecking_ on keyword arguments!

nvh0412:
thanks @rmosolgo

PR #3154, opened 2020-09-16
Minor development.md improvements

casperisfine:
I just stepped upon these little annoyance while working on another PR.


rmosolgo:
Thanks for fixing it!

PR #3167, opened 2020-09-23
Make query sanitization more extensible

jturkel:
We (Salsify) are implementing logging of GraphQL queries and variables but our approach is slightly different than the default behavior of `GraphQL::Language::SanitizedPrinter`:

* We log queries and variables separately to get a more faithful representation of the requests we're getting from clients
* We extended our `GraphQL::Schema::Argument` subclass to support a `redact` option rather than just redacting all strings

We were able to implement this by subclassing `GraphQL::Language::SanitizedPrinter` but it involved some unfortunate copying of code from `GraphQL::Language::Printer` to restore overridden behavior and copying of code from `GraphQL::Language::SanitizedPrinter` where we need to inject logic in the middle of a method. Alternatively we could directly subclass `GraphQL::Language::Printer` but then we would have lost the type bookkeeping from `GraphQL::Language::SanitizedPrinter`.

This PR proposes a few changes to `GraphQL::Language::SanitizedPrinter` to make it easier to customize the behavior:

1. Making the inlining of variable values optional with a default that retains the current behavior
1. Adding `redact_argument_value?` and `redacted_argument_value` hooks that subclasses can use to more easily customize the redaction behavior

### DIFF
--- a/lib/graphql/language/sanitized_printer.rb
+++ b/lib/graphql/language/sanitized_printer.rb
@@ -19,11 +19,12 @@ module GraphQL
 
       REDACTED = "\"<REDACTED>\""
 
-      def initialize(query)
+      def initialize(query, inline_variables: true)
         @query = query
         @current_type = nil
         @current_field = nil
         @current_input_type = nil
+        @inline_variables = inline_variables
       end
 
       # @return [String, nil] A scrubbed query string, if the query was valid.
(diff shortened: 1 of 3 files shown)


jturkel:
Rebased to resolve the conflicts with #3171.

rmosolgo:
Thanks, these changes look great!

jturkel:
Thanks @rmosolgo. Any plans for a 1.11.5 release soon?

rmosolgo:
> Any plans 

I don't usually _plan_ for point releases, but feel free to ask anytime! 

🚢  💁‍♂️  https://rubygems.org/gems/graphql/versions/1.11.5

jturkel:
Woohoo! Thanks @rmosolgo.

PR #3168, opened 2020-09-25
FieldExtension: pass extended values instead of originals to after_resolve

swalkinshaw:
This is a follow-up after https://github.com/rmosolgo/graphql-ruby/pull/3138 which had breaking changes for our application without a real way to work around it. This proposes a solution to make extensions dealing with extended arguments more flexible.

Field Extensions let you extend (modify) the resolver object and arguments in the `resolve` hook. Previously the `after_resolve` hook would receive the *original* object and arguments and not the extended ones. This was a lossy process with no way for `after_resolve` to access the extended values.

This changes `after_resolve` to receive the extended values instead. If you need access to the original non-extended values, the `memo` argument can be used.

Example:

```ruby
  def apply(field)
    field.argument(:my_new_arg, :string, required: false)
  end

  def resolve(object:, arguments:, context:)
    next_args = arguments.dup
    next_args.delete(:my_new_arg)
    yield(object, next_args, { original_arguments: arguments })
  end

  def after_resolve(value:, context:, arguments:, memo:, **_args)
    # `arguments` => extended/modified arguments
    # `memo` => hash with `original_arguments`
  end
end
```

cc @jturkel what do you think?

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -725,20 +725,21 @@ module GraphQL
         if @extensions.empty?
           yield(obj, args)
         else
-          # Save these so that the originals can be re-given to `after_resolve` handlers.
-          original_args = args
-          original_obj = obj
+          extended_obj = obj
+          extended_args = args
 
           memos = []
-          value = run_extensions_before_resolve(memos, obj, args, ctx) do |extended_obj, extended_args|
-            yield(extended_obj, extended_args)
+          value = run_extensions_before_resolve(memos, obj, args, ctx) do |obj, args|
+            extended_obj = obj
+            extended_args = args
+            yield(obj, args)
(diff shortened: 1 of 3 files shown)


swalkinshaw:
Should be noted this is another breaking change after #3138 as well.

swalkinshaw:
Oh I'll probably add a more explicit test case for this behaviour

rmosolgo:
Should I wait for another test before merging this? I'm happy with the test coverage as-is, but also happy to wait if you want to push a proper test for this behavior.

swalkinshaw:
😅 I completely forgot. I do want to add another test

swalkinshaw:
Added a test

rmosolgo:
Thanks!

PR #3171, opened 2020-09-28
SanitizedPrinter fixes for lists and JSON scalars

jturkel:
This PR fixes a few problems with `GraphQL::Language::SanitizedPrinter`:

1. Variables with lists of enum values rendered as strings
1. Queries with argument values that needed to be coerced to lists raised `NoMethodError: undefined method `arguments' for #<GraphQL::Schema::List:0x00007fc4819dd508>` errors
1. Queries with JSON scalars raised `NoMethodError: undefined method `arguments' for GraphQL::Types::JSON:Class` errors

### DIFF
--- a/lib/graphql/language/sanitized_printer.rb
+++ b/lib/graphql/language/sanitized_printer.rb
@@ -60,22 +60,31 @@ module GraphQL
       end
 
       def print_argument(argument)
+        # We won't have type information if we're recursing into a custom scalar
+        return super if @current_input_type && @current_input_type.kind.scalar?
+
         arg_owner = @current_input_type || @current_directive || @current_field
         arg_def = arg_owner.arguments[argument.name]
 
         old_input_type = @current_input_type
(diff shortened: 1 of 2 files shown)


lib/graphql/language/sanitized_printer.rb
```diff
@@ -60,22 +60,31 @@ def print_node(node, indent: "")
...
+        else
+          argument.value
+        end
+        res = "#{argument.name}: #{print_node(argument_value)}".dup
+
         @current_input_type = old_input_type
         res
       end
 
-      def print_list_type(list_type)
```
jturkel, line 73: This method is only used when printing list type definitions. Thus I don't think we need to track `@current_input_type` here since we'll never recurse into arguments from here. This is not strictly required for these PR fixes but it did make it harder to understand how `@current_input_type` is used.

rmosolgo:
Thanks for your work here! 

Weird about that CI failure. Maybe one of the capybara gems needs an update for Puma 5, I'll look into it separately.

Issue #3177, opened 2020-10-02
GraphQL::Schema::UniqueWithinType error handling missing for Base64Encoder


grvillic:
**Describe the bug**

We are seeing ArgumentError logs in [here](https://github.com/rmosolgo/graphql-ruby/blob/bebd78c0d5fddae997aec959506c6f42f71564ea/lib/graphql/schema/unique_within_type.rb#L27-L32) due to missing proper error handling.

**Versions**

`graphql` version: 1.11.1
`rails` (or other framework): 2.7

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok).
Are you using [interpreter](https://graphql-ruby.org/queries/interpreter.html)? Any custom instrumentation, etc?

```ruby
def decode(node_id, separator: self.default_id_separator)
   #  error handling missing below
    Base64Bp.urlsafe_decode64(node_id).split(separator, 2)
end
```

**Steps to reproduce**

- Pass an invalid node_id that is not a valid Base64 Encoded string.

**Expected behavior**

Operation should log decoding issue. Use Base64Encoder [utility](https://github.com/rmosolgo/graphql-ruby/blob/bebd78c0d5fddae997aec959506c6f42f71564ea/lib/graphql/schema/base_64_encoder.rb#L13-L19) that already has argument error handling.

**Actual behavior**

ArgumentError is being thrown without providing insights about the invalid string.

What specifically went wrong?

<details>
<summary>Click to view exception backtrace</summary>
```
ArgumentError
invalid base64
<internal:pack>:281:in `unpack1'
/usr/local/lib/ruby/2.7.0/base64.rb:74:in `strict_decode64'
/src/app/vendor/gems/ruby/2.7.0/gems/graphql-1.11.1/lib/graphql/schema/base_64_bp.rb:24:in `urlsafe_decode64'
/src/app/vendor/gems/ruby/2.7.0/gems/graphql-1.11.1/lib/graphql/schema/unique_within_type.rb:31:in `decode'
/src/app/app/graphql/api/schema.rb:25:in `object_from_id'
/src/app/vendor/gems/ruby/2.7.0/gems/graphql-1.11.1/lib/graphql/schema.rb:670:in `call'
/src/app/vendor/gems/ruby/2.7.0/gems/graphql-1.11.1/lib/graphql/schema.rb:670:in `object_from_id'
[...]

PR #3179, opened 2020-10-02
Error handling for base 64 decoding (unique_within_type)

grvillic:
[GraphQL::Schema::UniqueWithinType error handling missing for Base64Encoder](https://github.com/rmosolgo/graphql-ruby/issues/3177)

Small change to better handle `ArgumentError` exception when decoding an invalid string. We already have a utility function that handles it [here](https://github.com/rmosolgo/graphql-ruby/blob/bebd78c0d5fddae997aec959506c6f42f71564ea/lib/graphql/schema/base_64_encoder.rb#L13-L19), we are reusing it.

Adding a test for this change as well.


rmosolgo:
Thanks for the fix!

PR #3187, opened 2020-10-08
Fix custom connection total_count example

northerner:
As of 1.11.* the previous example of `object.items.count` no longer works for providing a count of nodes on a connection.

This appears to be a result of renaming `object` to `nodes` in f768ea. This changes allows the example to work again.

### DIFF
--- a/guides/type_definitions/extensions.md
+++ b/guides/type_definitions/extensions.md
@@ -122,7 +122,7 @@ class Types::MyCustomConnection < GraphQL::Types::Relay::BaseConnection
   field :total_count, Integer, null: false
 
   def total_count
-    object.items.size
+    nodes.size
   end
 end
 ```


rmosolgo:
Hey, thanks for the suggestion. Are you using the new connections plugin (`use GraphQL::Pagination::Connections` in your schema)? That's probably the difference. Those docs apply to the new plugin, which is added by default in `rails g graphql:install` these days. 

For example, here are the two different approaches shown in the tests: 

https://github.com/rmosolgo/graphql-ruby/blob/a79ec95e6457863a97a91c2ec5705d2355b69a83/spec/support/star_wars/schema.rb#L48-L57

https://github.com/rmosolgo/graphql-ruby/blob/a79ec95e6457863a97a91c2ec5705d2355b69a83/spec/support/star_wars/schema.rb#L435-L440

northerner:
> Hey, thanks for the suggestion. Are you using the new connections plugin (`use GraphQL::Pagination::Connections` in your schema)? That's probably the difference. Those docs apply to the new plugin, which is added by default in `rails g graphql:install` these days.

Ah, sorry, did not know about that plugin.

Would you recommend using it for building any custom connections on the latest version? Is it designed to be used in place of defining a custom `BaseConnection` and setting `connection_type_class Types::BaseConnection` in the `BaseObject`?

rmosolgo:
The new plugin provides new connection _objects_ (eg, `GraphQL::Pagination::ActiveRecordRelationConnection` instead of `GraphQL::Relay::RelationConnection`), but the type definitions stay the same. 

Similar to how you have a `::User` model and a `Types::User` GraphQL type, there are connection _objects_ and then connection _types_. The new plugin uses better underlying _objects_ for backing their GraphQL types.

northerner:
> The new plugin provides new connection _objects_ (eg, `GraphQL::Pagination::ActiveRecordRelationConnection` instead of `GraphQL::Relay::RelationConnection`), but the type definitions stay the same.
> 
> Similar to how you have a `::User` model and a `Types::User` GraphQL type, there are connection _objects_ and then connection _types_. The new plugin uses better underlying _objects_ for backing their GraphQL types.

Thanks for the clear explanation on this @rmosolgo, really appreciate you taking the time.

PR #3192, opened 2020-10-14
Allow field lookup to be overwritten in custom field class

swelther:
We use a custom field class to do some magic field lookup, e.g. call the field method on a decorator when a `decorated: true` flag has been set.

This is no longer possible in the 1.11.x version of this awesome gem. I know it is a bad idea to overwrite private methods and to complain when something has changed upstream. Just want to check if extracting the raise would be an option, so we can still work with our custom class.

Are yo maybe interested in adding the `decorated` feature? Would be happy to help or provide a PR for that.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -705,17 +705,22 @@ module GraphQL
               obj.object.public_send(@method_sym)
             end
           else
-            raise <<-ERR
-          Failed to implement #{@owner.graphql_name}.#{@name}, tried:
-
-          - `#{obj.class}##{@resolver_method}`, which did not exist
-          - `#{obj.object.class}##{@method_sym}`, which did not exist
-          - Looking up hash key `#{@method_sym.inspect}` or `#{@method_str.inspect}` on `#{obj.object}`, but it wasn't a Hash
-
-          To implement this field, define one of the methods above (and check for typos)
-          ERR
+            field_lookup(obj, ruby_kwargs)
           end
         end
+      end
+
+      # May be overwritten in custom field class to do some custom field lookup
+      def field_lookup(obj, ruby_kwargs)
+        raise <<-ERR
+      Failed to implement #{@owner.graphql_name}.#{@name}, tried:
+
+      - `#{obj.class}##{@resolver_method}`, which did not exist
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! Could you share some details on what you're _doing_ with this? I don't understand what you mean by `decorated: true`, but if you could explain it a bit, maybe we can find a way to accomplish it within the existing feature set.

swelther:
Of course, sorry my comment was probably a bit confusing.

We define some of our fields like this:

```
  field :created_at_decorated, ::GraphQL::Types::String, null: false, decorated: true
```

And in our custom field class we do this:

```
  def resolve_field_method(obj, ruby_kwargs, ctx)
    return super unless @decorated

    undecorated_method = method_str.delete_suffix('_decorated')
    obj.object.decorate.send(undecorated_method) if obj.object.decorate.respond_to?(undecorated_method)
  end
```

It's a shortcut so we do not need methods like these in our type classes:

```
  def created_at_decorated
    object.decorate.created_at
  end
```

I know it is not a good idea to overwrite private methods. Guess we were lucky that it worked for so long :)

rmosolgo:
Ok, thanks for those details. Instead of buildling this into the gem, I recommend a [field extension](https://graphql-ruby.org/type_definitions/field_extensions.html) like this: 


```ruby 
class DecoratedField < GraphQL::Schema::FieldExtension
  def apply 
    @undecorated_method_name = field.method_str.delete_suffix('_decorated')
  end
  
  # When this extension is added, short-circuit the normal resolve behavior (don't `yield(object, arguments)`). 
  def resolve(object:, **_rest)
    decorator = object.decorated 
    if decorator.respond_to?(@undecorated_method_name)
      decorator.public_send(@undecorated_method_name)
    else 
      nil 
    end  
  end 
end 
```

then you could hook it up like this: 

```ruby 
class Types::BaseField < GraphQL::Schema::Field 
  def initialize(*args, decorated: false, **kwargs, &block)
    if decorated 
      extension(DecoratedField)  
    end 
  end 
end 

# ... 
field :created_at_decorated, ::GraphQL::Types::String, null: false, decorated: true
```

That way, the field will use `decorated: true` to attach that extension. The extension should work the same as the monkey-patch you shared. If not, let me know!

PR #3194, opened 2020-10-15
Fix fields nullability in subscriptions documentation

00dav00:
The documentation on subscriptions page show `Room` and `Message` as mandatory in the client side, but the ruby class show them as `null: true`. This change makes them not nullable in the ruby gql class to have consistency

```javascript
type MessageWasPostedPayload {
  room: Room!
  message: Message!
}
```


rmosolgo:
Thanks for the fix!

PR #3196, opened 2020-10-20
Accept additional options in global_id_field macro

barthez:
Accept options in `global_id_field` macro and pass them to as field definition options. This would be extremely useful with https://github.com/Gusto/apollo-federation-ruby/ turning:
```
global_id_field :id
field :id, ID, null: false, external: true
```
into:
```
global_id_field :id, external: true
```


rmosolgo (approved):
Looks great, thanks for the improvement!

PR #3204, opened 2020-10-25
Update method signature example in extension docs

bswinnerton:
While spending some time extending the graphql-ruby type system in a project of mine I noticed that the documentation that describes how the `field` DSL generates a `GraphQL::Schema::Field` object wasn't using the correct method signature. It looks like the method signature changed in 82d95dd, but this documentation was never updated.

The method signature is quite complex with a large number of keyword arguments, but hopefully only using the required arguments is sufficient for this particular documentation example.


rmosolgo:
Thanks for the fix! (👋 !)

PR #3205, opened 2020-10-26
Add InputObjectNamesAreUnique static validation

chrisbutcher:
### Context

From the GraphQL spec: 
> Input objects must not contain more than one field of the same name, otherwise an ambiguity would exist which includes an ignored portion of syntax.

https://spec.graphql.org/June2018/#sec-Input-Object-Field-Uniqueness

### Expected behaviour

This query should be considered invalid and the response should contain `errors`.

### Actual behaviour

It appears the last usage of a duplicate input object field is what graphql-ruby accepts, without any errors.

# Update

This PR now contains a proposed fix. Adding `GraphQL::StaticValidation::InputObjectNamesAreUnique`. It seems like a pretty straightforward change, consistent with how existing validations like `ArgumentNamesAreUnique` work.

### DIFF
--- a/lib/graphql/static_validation/rules/input_object_names_are_unique_error.rb
+++ b/lib/graphql/static_validation/rules/input_object_names_are_unique_error.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+module GraphQL
+  module StaticValidation
+    class InputObjectNamesAreUniqueError < StaticValidation::Error
+      attr_reader :name
+
+      def initialize(message, path: nil, nodes: [], name:)
+        super(message, path: path, nodes: nodes)
+        @name = name
+      end
+
+      # A hash representation of this Message
+      def to_h
+        extensions = {
+          "code" => code,
+          "name" => name
+        }
+
+        super.merge({
+          "extensions" => extensions
+        })
+      end
+
+      def code
+        "inputFieldNotUnique"
+      end
+    end
+  end
+end
+
--- a/spec/graphql/query_spec.rb
+++ b/spec/graphql/query_spec.rb
@@ -972,4 +972,23 @@ describe GraphQL::Query do
       assert_equal "order_by", order_by_argument_value.definition.graphql_name
     end
   end
+
+  describe "when provided input object field names are not unique" do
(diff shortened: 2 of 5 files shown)


rmosolgo:
:+1: thanks for spotting this! Are you interested in working up a fix?

chrisbutcher:
Sure thing! Here's the reference implementation: https://github.com/graphql/graphql-js/blob/master/src/validation/rules/UniqueInputFieldNamesRule.js#L34

Do you aim to keep error messages 1:1 in sync between this gem and the JS reference?

rmosolgo:
> error messages 1:1 in sync

sure, it'd be nice, but AFAIK none of the other error messages match, and there's no tests for ensuring that they match 😿

spec/graphql/query_spec.rb
```diff
@@ -972,4 +972,23 @@ def self.after_multiplex(m)
...
+      let(:query_string) {%|
+        query getCheeses{
+          searchDairy(product: [{ source: COW, source: COW }]) {
+            __typename
+          }
+        }
+      |}
+
+      it "returns errors" do
+        refute_nil(result["errors"])
```
chrisbutcher, line 990: This test was the one I added originally to confirm the issue. Do you want this improved, moved, or deleted if the validation-specific spec below is sufficient?

lib/graphql/static_validation/rules/input_object_names_are_unique_error.rb
```diff
@@ -0,0 +1,30 @@
...
+          "name" => name
+        }
+
+        super.merge({
+          "extensions" => extensions
+        })
+      end
+
+      def code
+        "inputFieldNotUnique"
```
chrisbutcher, line 25: I couldn't find a precedent for these `code`s in the JS implementation: https://github.com/graphql/graphql-js/search?q=argumentNotUnique

rmosolgo: LGTM :+1:

rmosolgo:
Thanks for the fix here!

Issue #3202, opened 2020-10-22
Int scalar type: #input_coercion does not raise error as per the spec


toneymathews:
More of a question than an issue, 

As per the [spec](http://spec.graphql.org/June2018/#sec-Int), integer input value should not be less than -2^31 or greater than (2^31 -1). If so, an error needs to be raised. This does not seem to be enforced here though #coerce_result does seem to do so.

https://github.com/rmosolgo/graphql-ruby/blob/61fbc549a2c9d77bb9a8ab0e25887b48b32043f9/lib/graphql/types/int.rb#L12-L14

We were wondering why this was not enforced?

rmosolgo:
I bet it was just an oversight in my first implementation of this scalar, and nobody has _needed_ it since then. I'm not sure why only coerce_result was updated in https://github.com/rmosolgo/graphql-ruby/pull/2101 (was there a particular reason, @eapache?)

eapache:
🤔 not that I recall. It was either a complete oversight, or I assumed that `value.is_a?(Numeric)` would already do a bounds check (though I'm not sure why I'd think that).

toneymathews:
Ok, will put up a PR to bind this type within the limits.

PR #3206, opened 2020-10-27
Bind scalar int as per spec

toneymathews:
Closes https://github.com/rmosolgo/graphql-ruby/issues/3202

The int scalar type should be bound within (2^31 - 1) and -2^31 for #coerce_input as per the graphql spec. This PR attempts to enforce it.

 @eapache @rmosolgo I think this now conforms to the spec?


lib/graphql/types/int.rb
```diff
@@ -10,7 +10,7 @@ class Int < GraphQL::Schema::Scalar
       MAX = (2**31) - 1
 
       def self.coerce_input(value, _ctx)
-        value.is_a?(Integer) ? value : nil
+        value.is_a?(Integer) && value >= MIN && value <= MAX ? value : nil
```
rmosolgo, line 13: This implementation looks fine spec-wise, but I want to make sure that we provide a path for people who want to preserve the old behavior. How about: 

- create a new error class, `GraphQL::IntegerDecodingError`, which can basically be identical to `GraphQL::IntegerEncodingError`. (But, we'll use a different class so that people can handle the two scenarios differently 
- When the value is out of range, make a new `GraphQL::IntegerDecodingError` and pass it to `ctx.schema.type_error(err, ctx)` 
- Extend the default type error handler (https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/default_type_error.rb) to handle instances of that error class by returning `nil`. (Or, should it raise the error to the client?)

That way: 

- The behavior you've proposed here is the default 
- Users of the library can retain the old behavior by adding to their schema: 

  ```ruby 
  def self.type_error(err, ctx)
    if err.is_a?(GraphQL::IntegerDecodingError)
      err.value # return it anyways, since this is how graphql-ruby used to work 
    else 
      super 
    end 
  end 
  ```


How does that sound to you?

toneymathews: > to handle instances of that error class by returning nil. (Or, should it raise the error to the client?)

I thought it should return nil since, since it would be consistent with other scalar types that seem to return nil for invalid inputs in #corece_input.

rmosolgo: > consistent

makes sense to me :+1:

rmosolgo:

PR #3210, opened 2020-11-03
Fix race conditions with Ably subscriptions

jscheid:
Specify the [`rewind`](https://www.ably.io/blog/channel-rewind/) parameter for each channel so that early messages don't get lost.

Without this change, any messages sent during a specific time window are lost. This time window begins when the subscription is established server-side and ends when the client-side call to `subscribe` takes effect, i.e. once the subscription request has been received and processed by Ably servers.  I've observed this window to be up to three seconds long,  although obviously it can be of arbitrary length depending on network latency etc.

Note that Ably promises to keep messages (with the default channel configuration) for two minutes, which should be enough even for pathological cases.  However, technically messages can still get lost when establishment of the subscription takes longer than two minutes and the Ably channel hasn't been configured for persistence, or when more than 100 messages are sent before the subscription is established. Perhaps these edge cases can be further improved in the future.

This change should be safe with respect to the danger of more-than-once delivery since every subscription gets a fresh channel.

Also, fix a second race condition, where an error was raised in the `enter` callback when the subscription is disposed immediately after creation.

Also, add an integration test for message rewinding, and make an existing integration test more robust by doing away with an arbitrary timeout value.

### DIFF
--- a/javascript_client/src/subscriptions/__tests__/createAblyHandlerTest.ts
+++ b/javascript_client/src/subscriptions/__tests__/createAblyHandlerTest.ts
@@ -329,6 +329,9 @@ describe("createAblyHandler", () => {
 
     // For executing this test you need to provide a valid Ably API key in
     // environment variable ABLY_KEY
+    //
+    // This test might take longer than the default jest timeout of 5s.
+    // Consider setting a higher timeout when running in CI.
     testWithAblyKey("can make more than 200 subscriptions", async () => {
       let caughtError = null
       const ably = new Realtime({ key, log: { level: 0 } })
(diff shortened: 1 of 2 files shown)


rmosolgo:
This fix looks great, thanks! I'm amazed at how complicated Ably is 🙈 . I used Pusher in production for years at an old job and never ran into issues like this. I wonder if it was because of our use case (less mission-critical), or if they just implement some of these behaviors by default instead of requiring configuration, or if we actually _had_ these issues but never noticed them 🤷 ! 

Anyhow, thanks for the improvements and for the writeup of the problem.

rmosolgo:
🚢  in 1.7.12!

PR #3212, opened 2020-11-04
Pass along extended args even when an extension shortcuts the normal resolve

rmosolgo:
Previously, any modifications to `args` were lost. cc @swalkinshaw 

This is a hack, but it's a _hack with a test_ 🎉 .


swalkinshaw:
cc @guialbuk since he was trying out your other branch on Shopify. We'll test this one as well 👍

swalkinshaw:
@rmosolgo this works on our end, thanks 🎉

PR #3215, opened 2020-11-09
GraphQLCollector support for prometheus' custom_labels

codebender:
I noticed that the GraphQLCollector wasn't correctly supporting custom_labels for metrics, so I added support.

Same implementation as the `prometheus_exporter` gem's build-in collectors. https://github.com/discourse/prometheus_exporter/blob/master/lib/prometheus_exporter/server/sidekiq_collector.rb#L19-L22

### DIFF
--- a/lib/graphql/tracing/prometheus_tracing/graphql_collector.rb
+++ b/lib/graphql/tracing/prometheus_tracing/graphql_collector.rb
@@ -16,7 +16,10 @@ module GraphQL
         end
 
         def collect(object)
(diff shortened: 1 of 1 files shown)


rmosolgo:
This is fine with me, but should we update the docs or add any tests for it? I have never used this so I'm not sure if that would be appropriate here. 

Also, cc @thejchap who originally added this tracing, in case you're interested.

thejchap:
Sweet! Glad to see this is getting some use :)

rmosolgo:
Thanks for the improvement!

PR #3219, opened 2020-11-12
Support generating outfile without a URL

### LINKED ISSUE #3218: Option to skip updating the server side and only create the output file
Deployment is React on Rails pro where we have SSR of React.

We would like to have the creation of the output JSON file done as part of the build process before running webpack.

Then during the release phase on Heroku, we run the DB sync after DB migrations.

Need to be able to omit the `--url` of the command, as shown here: https://graphql-ruby.org/operation_store/client_workflow#syncing 

image

We'd change:

https://github.com/rmosolgo/graphql-ruby/blob/master/javascript_client/src/sync/index.ts#L44-L47

```js
  var url = options.url
  if (!url) {
    throw new Error("URL must be provided for sync")
  }
```

to allow having no URL.

The full details of what we've been trying to do are are spelled out in here [#3067: OperationStore docs should include deployment sequence notes](https://github.com/rmosolgo/graphql-ruby/issues/3067#issuecomment-682411205).

If we can't get confirmation on this suggestion to change the `graphql-ruby-client` params, we'll create a separate NPM package that has the functionality that we need.
 

**Additional context**

TBD

rmosolgo:
This makes some deployment flows better. 

Fixes #3218 

What do you think, @justin808?


javascript_client/src/__tests__/syncTest.ts
```diff
@@ -25,6 +25,21 @@ describe("sync operations", () => {
     jest.clearAllMocks();
   })
 
+  describe("generating artifacts without syncing", () => {
+    it("works without a URL", () => {
```
justin808, line 29: Yes, exactly what we need!

javascript_client/src/sync/index.ts
```diff
@@ -27,7 +27,7 @@ interface SyncOptions {
  * @param {String} options.path - A glob to recursively search for `.graphql` files (Default is `./`)
  * @param {String} options.relayPersistedOutput - A path to a `.json` file from `relay-compiler`'s  `--persist-output` option
  * @param {String} options.secret - HMAC-SHA256 key which must match the server secret (default is no encryption)
- * @param {String} options.url - Target URL for sending prepared queries
+ * @param {String} options.url - Target URL for sending prepared queries. If omitted, then an outfile is generated without sending operations to the server.
```
justin808, line 30: 💯

justin808 (approved):
LGTM!

Have you considered:

1. Breaking the code up into smaller functions
2. Using let and const rather than var?

PR #3224, opened 2020-11-17
First-class schema directives

rmosolgo:
Add a system of directives so that schemas can: 

- Add custom configuration to types, fields, and arguments 
- Automatically include that info in `.to_definition` 
- Load configuration in `.from_definition`

TODO: 

- [x] Docs for configuring and printing 
- [x] Example for what it can be used for (visibility)
- [x] Unify the implementation of `@deprecated` to use a directive


rmosolgo:
This is passing locally, but I'm having trouble with CI. I'll merge it, but I'm working on CI over in #3251

PR #3226, opened 2020-11-19
Update README.md

bdarcet:
Change the API Documentation dead link 

From : https://www.rubydoc.info/gems/graphql
To : https://www.rubydoc.info/github/rmosolgo/graphql-ruby


rmosolgo:
Weird, it shows up here: https://www.rubydoc.info/find/gems?q=graphql But the links from that page are dead, too. Thanks for the update.

PR #3228, opened 2020-11-20
Validation: fields will merge: Faster argument comparison 

ravangen:
## Background

The Fields Will Merge validation rule seeks to ensure that when multiple field selections with the same response names are encountered during execution, the field and arguments to execute and the resulting value are to be unambiguous ([spec](http://spec.graphql.org/June2018/#sec-Field-Selection-Merging)). There are some examples of this behaviour within [GraphQL Java blog post](https://www.graphql-java.com/blog/deep-dive-merged-fields/).

## Problem

The runtime of the graphql-ruby algorithm implementation grows exponentially. As a more extreme example, for `n = 5000` repetitions of the same field: `1/2 * (n^2 - n) = 12497500`
MSP902411hdg308adc0d44c0000632ib29337c0efc8

As far as I can tell, this is not a perfectly solved problem in other GraphQL server implementations either. For example, [graphql-js reproduction repo](https://github.com/ravangen/graphql-js-merge-fields) and this [XING engineering blog post](https://tech.xing.com/graphql-overlapping-fields-can-be-merged-fast-ea6e92e0a01).

## Suggestion

I did some basic profiling, and it seemed that the majority of time was spent on `GraphQL::StaticValidation::FieldsWillMerge#possible_arguments` and `GraphQL::StaticValidation::FieldsWillMerge#find_conflicts_between_sub_selection_sets`. This work tries to optimize `possible_arguments`.

The rule was originally based on the graphql-js implementation, and I think they diverged in this case to provide an improved error message in graphql-ruby. Due to the increase in number of iterations as the number of fields grows, smaller tweaks can provide a benefit.

This proposes returning to the [graphql-js reference implementation](https://github.com/graphql/graphql-js/blob/00eab30fb269b6e2a4e8e61d097d3e249319420e/src/validation/rules/OverlappingFieldsCanBeMergedRule.js#L626-L643). The performance gain is from the happy path where there are fewer objects/allocations without the use `uniq`. This still maintains the original error details.

## Benchmarks

<details>
<summary>Figure 1: Original, Sample 1</summary>

```bash
$ bundle exec rake bench:validate
validate - introspection 
                        63.000  i/100ms
validate - abstract fragments
                       548.000  i/100ms
                       407.000  i/100ms
validate - big query     5.000  i/100ms
validate - fields will merge
                         1.000  i/100ms
validate - introspection 
                        546.473  (±11.5%) i/s -      2.709k in   5.022966s
validate - abstract fragments
                          4.160k (± 8.8%) i/s -     20.824k in   5.048538s
                          3.160k (± 8.1%) i/s -     15.873k in   5.057849s
validate - big query     45.263  (± 2.2%) i/s -    230.000  in   5.085945s
validate - fields will merge
                          0.030  (± 0.0%) i/s -      1.000  in  32.913349s
```

</details>

<details>
<summary>Figure 2: Original, Sample 2</summary>

```bash
$ bundle exec rake bench:validate
validate - introspection 
                        42.000  i/100ms
validate - abstract fragments
                       387.000  i/100ms
                       320.000  i/100ms
validate - fields will merge
                         1.000  i/100ms
validate - introspection 
                        471.898  (± 4.9%) i/s -      2.394k in   5.085980s
validate - abstract fragments
                          4.225k (± 5.0%) i/s -     21.285k in   5.051741s
                          3.517k (± 5.2%) i/s -     17.600k in   5.017897s
validate - big query     50.486  (± 4.0%) i/s -    252.000  in   5.002424s
validate - fields will merge
                          0.032  (± 0.0%) i/s -      1.000  in  31.525522s
```

</details>

<details>

<summary>Figure 3: Updated, Sample 1</summary>

```bash
$ bundle exec rake bench:validate
validate - introspection 
                        51.000  i/100ms
validate - abstract fragments
                       465.000  i/100ms
                       384.000  i/100ms
validate - big query     5.000  i/100ms
validate - fields will merge
                         1.000  i/100ms
validate - introspection 
                        508.377  (± 4.7%) i/s -      2.550k in   5.027187s
validate - abstract fragments
                          4.478k (± 4.3%) i/s -     22.785k in   5.098101s
                          3.709k (± 4.0%) i/s -     18.816k in   5.081693s
validate - big query     54.156  (± 5.5%) i/s -    270.000  in   5.006528s
validate - fields will merge
                          0.052  (± 0.0%) i/s -      1.000  in  19.127847s
```

</details>

<details>

<summary>Figure 4: Updated, Sample 2</summary>

```bash
$ bundle exec rake bench:validate
validate - introspection 
                        51.000  i/100ms
validate - abstract fragments
                       456.000  i/100ms
                       381.000  i/100ms
validate - big query     5.000  i/100ms
validate - fields will merge
                         1.000  i/100ms
validate - introspection 
                        581.214  (± 9.3%) i/s -      2.907k in   5.046322s
validate - abstract fragments
                          4.577k (±13.9%) i/s -     22.800k in   5.094230s
                          3.038k (±17.3%) i/s -     14.859k in   5.047211s
validate - big query     47.441  (±19.0%) i/s -    230.000  in   5.046090s
validate - fields will merge
                          0.051  (± 0.0%) i/s -      1.000  in  19.545309s
```

</details>

## Possible Future Improvements

We could have a configuration option to set an upper bound on how long the validation phase is permitted to run before aborting the request and returning an error. Ideally we are not spending all of our cycles on validation. EDIT: See https://github.com/rmosolgo/graphql-ruby/pull/3234

👶 I'm new to the Ruby ecosystem, so welcome any tips and suggestions on how to make this even better!

### DIFF
--- a/benchmark/run.rb
+++ b/benchmark/run.rb
@@ -19,6 +19,9 @@ module GraphQLBenchmark
   BIG_SCHEMA = GraphQL::Schema.from_definition(File.join(BENCHMARK_PATH, "big_schema.graphql"))
   BIG_QUERY = GraphQL.parse(File.read(File.join(BENCHMARK_PATH, "big_query.graphql")))
 
+  FIELDS_WILL_MERGE_SCHEMA = GraphQL::Schema.from_definition("type Query { hello: String }")
+  FIELDS_WILL_MERGE_QUERY = GraphQL.parse("{ #{Array.new(5000, "hello").join(" ")} }")
+
   module_function
   def self.run(task)
     Benchmark.ips do |x|
@@ -30,6 +33,7 @@ module GraphQLBenchmark
         x.report("validate - abstract fragments") { CARD_SCHEMA.validate(ABSTRACT_FRAGMENTS) }
         x.report("validate - abstract fragments 2") { CARD_SCHEMA.validate(ABSTRACT_FRAGMENTS_2) }
         x.report("validate - big query") { BIG_SCHEMA.validate(BIG_QUERY) }
+        x.report("validate - fields will merge") { FIELDS_WILL_MERGE_SCHEMA.validate(FIELDS_WILL_MERGE_QUERY) }
       else
         raise("Unexpected task #{task}")
       end
--- a/lib/graphql/static_validation/rules/fields_will_merge.rb
+++ b/lib/graphql/static_validation/rules/fields_will_merge.rb
@@ -202,15 +202,16 @@ module GraphQL
             )
           end
 
-          args = possible_arguments(node1, node2)
-          if args.size > 1
-            msg = "Field '#{response_key}' has an argument conflict: #{args.map { |arg| GraphQL::Language.serialize(arg) }.join(" or ")}?"
+          if !same_arguments?(node1, node2)
+            args = [serialize_field_args(node1), serialize_field_args(node2)]
+            conflicts = args.map { |arg| GraphQL::Language.serialize(arg) }.join(" or ")
+            msg = "Field '#{response_key}' has an argument conflict: #{conflicts}?"
             context.errors << GraphQL::StaticValidation::FieldsWillMergeError.new(
               msg,
               nodes: [node1, node2],
               path: [],
               field_name: response_key,
-              conflicts: args.map { |arg| GraphQL::Language.serialize(arg) }.join(" or ")
+              conflicts: conflicts
             )
           end
         end
@@ -326,20 +327,19 @@ module GraphQL
         [fields, fragment_spreads]
       end
 
-      def possible_arguments(field1, field2)
+      def same_arguments?(field1, field2)
         # Check for incompatible / non-identical arguments on this node:
-        [field1, field2].map do |n|
-          if n.arguments.any?
-            serialized_args = {}
-            n.arguments.each do |a|
-              arg_value = a.value
-              serialized_args[a.name] = serialize_arg(arg_value)
-            end
-            serialized_args
-          else
-            NO_ARGS
-          end
-        end.uniq
+        arguments1 = field1.arguments
+        arguments2 = field2.arguments
+
+        return false if arguments1.length != arguments2.length
+
+        arguments1.all? do |argument1|
+          argument2 = arguments2.find { |argument| argument.name == argument1.name }
+          return false if argument2.nil?
+
+          serialize_arg(argument1.value) == serialize_arg(argument2.value)
+        end
       end
 
       def serialize_arg(arg_value)
@@ -351,6 +351,14 @@ module GraphQL
         else
           GraphQL::Language.serialize(arg_value)
         end
+      end
+
+      def serialize_field_args(field)
+        serialized_args = {}
+        field.arguments.each do |argument|
+          serialized_args[argument.name] = serialize_arg(argument.value)
+        end
+        serialized_args
       end
 
       def compared_fragments_key(frag1, frag2, exclusive)


lib/graphql/static_validation/rules/fields_will_merge.rb
```diff
@@ -326,20 +327,19 @@ def find_fields_and_fragments(selections, owner_type:, parents:, fields:, fragme
...
+
+        arguments1.all? do |argument1|
+          argument2 = arguments2.find { |argument| argument.name == argument1.name }
+          return false if argument2.nil?
+
+          serialize_arg(argument1.value) == serialize_arg(argument2.value)
+        end
       end
 
       def serialize_arg(arg_value)
```
swalkinshaw, line 345: Not exactly in the scope of this PR, but this made me remember it might be better to use `GraphQL::Language::Printer` here since it already has this logic. @rmosolgo should that be used throughout instead of `GraphQL::Language.serialize` and `arg_value.to_query_string`?

rmosolgo: IMO the current code is _fine_. `.to_query_string` calls `Printer` under the hood: 

https://github.com/rmosolgo/graphql-ruby/blob/9cd0062330e225c90e6e4162e09c99384bf66a68/lib/graphql/language/nodes.rb#L76-L82

And `.serialize` is what `Printer` uses anyway: 

https://github.com/rmosolgo/graphql-ruby/blob/9cd0062330e225c90e6e4162e09c99384bf66a68/lib/graphql/language/printer.rb#L345-L351

So, practically speaking, I'm not sure what we'd gain from a refactor. (This method doesn't handle `Hash`es, but I guess it knows that it never receives a Hash? It would be nice if expected input classes were explicitly named by `when ...` and `else` raised an error, but that's not the end of the world.)

benchmark/run.rb
```diff
@@ -19,6 +19,9 @@ module GraphQLBenchmark
   BIG_SCHEMA = GraphQL::Schema.from_definition(File.join(BENCHMARK_PATH, "big_schema.graphql"))
   BIG_QUERY = GraphQL.parse(File.read(File.join(BENCHMARK_PATH, "big_query.graphql")))
 
+  FIELDS_WILL_MERGE_SCHEMA = GraphQL::Schema.from_definition(File.join(BENCHMARK_PATH, "fields_will_merge_schema.graphql"))
```
rmosolgo, line 22: Since the schema here is so simple, how about inlining the string instead? I imagine my future self will appreciate seeing the relevant text without opening another file. 

```suggestion
  FIELDS_WILL_MERGE_SCHEMA = GraphQL::Schema.from_definition("type Query { hello: String }")
```

benchmark/run.rb
```diff
@@ -19,6 +19,9 @@ module GraphQLBenchmark
...
+  FIELDS_WILL_MERGE_QUERY = GraphQL.parse(File.read(File.join(BENCHMARK_PATH, "fields_will_merge_query.graphql")))
```
rmosolgo, line 23: I think it would increase the clarity of this query's intent and simplify maintenance if this string was created dynamically instead of dumped to a file. For example: 

```suggestion
  FIELDS_WILL_MERGE_QUERY = GraphQL.parse("{ #{Array.new(5000, "hello").join(" ")} }")
```

(That makes it immediately clear that "this query string is 'hello' 5000 times in a row"!)

rmosolgo (commented):
👋  Welcome! This looks great, thanks for contributing it along with that benchmark. 

I left a couple of suggestions about simplifying the benchmark setup, but besides that, LGTM!

ravangen:
Thanks, I'll try to make the changes early next week.

rmosolgo:
Awesome, thanks again for this improvement 🎉 !

PR #3229, opened 2020-11-23
Update GraphQL spec link which is moved permanently

jimytc:
Current link: https://graphql.github.io/graphql-spec/draft/#example-fce18
New link: https://spec.graphql.org/draft/#example-fce18

The server uses 301 https code but with non-HTTPS scheme for new website.

Update the link to avoid further redirection and keep using HTTPS scheme.


rmosolgo:
Thanks for the fix!

PR #3234, opened 2020-11-27
Feature: Add ability to timeout on validation rules

ravangen:
Followup to https://github.com/rmosolgo/graphql-ruby/pull/3228

On a GraphQL schema, use `validate_timeout` to set an upper bound on how long static validation rules are allowed to run before returning an error. This way we can individually tune how long validation is allowed to take before aborting as this can likely be a smaller value than the query execution and HTTP timeouts.

## Implementation Consideration

This makes use of the dangerous Ruby `Timeout` module and should be used cautiously. It makes the assumption that all static validation rules are stateless Ruby code that requires no cleanup if process was interrupted (this is true of the default rules). This means rules should not include blocking IO calls, native gems, locks, or `rescue` clauses that must be reached.

I tried to provide an integration test, but it is nondeterministic so it was tricky and figured it may not be worth it. Suggestions?


rmosolgo (commented):
:+1: I think this is great! 

I agree that, even though `Timeout` is risky, it's probably ok here. How about adding a note about it to the `### Validation` doc you added? 

As far as tests go, I could go either way. I see you've got a test for the error handler, which is good. If you'd _like_ to add an integration, feel free to share what you've got so far. Maybe I can take another look and find something more deterministic. 

Thanks for the improvement! I don't have any suggestions except that update to the docs.

rmosolgo:
:+1: I'll update the docs after merging

PR #3238, opened 2020-11-30
Promote methods to public on relay's array conn

jonallured:
I recently extracted a gem called [graphql-page_cursors](https://github.com/artsy/graphql-page_cursors) that extracts Artsy's graphql pagination approach but it relies on all connections exposing first and last methods. Unless I'm missing something, I believe in the case of Relay's Array connection these methods have been marked private by mistake. Any reason to not promote these two up?

### DIFF
--- a/lib/graphql/relay/array_connection.rb
+++ b/lib/graphql/relay/array_connection.rb
@@ -31,8 +31,6 @@ module GraphQL
         end
       end
 
-      private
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: I don't suppose it'll hurt anything! Thanks for the improvement.

PR #3240, opened 2020-12-01
Convert to string in case it is not

alvaromat:
If we define an enum like this:

```ruby
class MyEnum < ::Types::Base::EnumType
    value "Zero", 0
end
```

When we try to get the schema in IDL using `GraphQL::Schema::Printer`, we will get an error like this:

```
NoMethodError: undefined method `split' for 0:Integer
/Users/alvaro/.rvm/gems/ruby-2.5.1@alkimii/gems/graphql-1.11.6/lib/graphql/language/block_string.rb:66:in `print'
/Users/alvaro/.rvm/gems/ruby-2.5.1@alkimii/gems/graphql-1.11.6/lib/graphql/language/printer.rb:257:in `print_description'
/Users/alvaro/.rvm/gems/ruby-2.5.1@alkimii/gems/graphql-1.11.6/lib/graphql/language/printer.rb:218:in `block in print_enum_type_definition'
/Users/alvaro/.rvm/gems/ruby-2.5.1@alkimii/gems/graphql-1.11.6/lib/graphql/language/printer.rb:217:in `each'
/Users/alvaro/.rvm/gems/ruby-2.5.1@alkimii/gems/graphql-1.11.6/lib/graphql/language/printer.rb:217:in `with_index'
/Users/alvaro/.rvm/gems/ruby-2.5.1@alkimii/gems/graphql-1.11.6/lib/graphql/language/printer.rb:217:in `print_enum_type_definition'
```

This avoids unexpected errors in case the description type can be converted to a string.

### DIFF
--- a/lib/graphql/language/block_string.rb
+++ b/lib/graphql/language/block_string.rb
@@ -63,7 +63,7 @@ module GraphQL
       end
 
       def self.print(str, indent: '')
-        lines = str.split("\n")
+        lines = str.to_s.split("\n")
 
         block_str = "#{indent}\"\"\"\n".dup
 


rmosolgo:
Hi! Thanks for reporting this issue and suggesting a fix. Is there any reason that GraphQL-Ruby should support non-string descriptions? 

If not, how about fixing this in the type definitions instead, for example: 

https://github.com/rmosolgo/graphql-ruby/blob/9cd0062330e225c90e6e4162e09c99384bf66a68/lib/graphql/schema/field.rb#L225

https://github.com/rmosolgo/graphql-ruby/blob/9cd0062330e225c90e6e4162e09c99384bf66a68/lib/graphql/schema/enum_value.rb#L45

https://github.com/rmosolgo/graphql-ruby/blob/9cd0062330e225c90e6e4162e09c99384bf66a68/lib/graphql/schema/argument.rb#L53

https://github.com/rmosolgo/graphql-ruby/blob/9cd0062330e225c90e6e4162e09c99384bf66a68/lib/graphql/schema/member/base_dsl_methods.rb#L42-L54

?

alvaromat:
In some edge use cases, like in the one I got this error, the description may be for example an integer. 
We solved the issue by just passing the description as a string. 

The problem in my opinion is that it took some time to understand that the problem was a not string description. Maybe the way to go is to make sure the enum description is a string.

rmosolgo:
> Maybe the way to go is to make sure the enum description is a string. 

Yes, I agree. Rather than updating all the schema tooling to support non-string descriptions, we should validate that descriptions are provided as strings. If you're interested in exploring that, please see the links above -- those are all the places I can think of where we accept descriptions.

Issue #3244, opened 2020-12-02
ActiveRecord fields of type decimal should be translated to floats by the generator


tbpeder:
Consider the following ActiveRecord schema.

```
create_table "product", force: :cascade do |t|
  t.decimal "amount", precision: 10, scale: 2
end
```

"rails g graphql:object product" will produce the following product_type.rb:

```
module Types
  class ProductType < Types::BaseObject
    field :amount, [Types::DecimalType], null: false
  end
end
```

DecimalType makes graphql explode and should be a Float instead. This only seems to happen with decimal fields that have precision and scale.

Same problem happens with uuid types. They should just be converted to String in GraphQL.

rmosolgo:
Fixed by #3246! To try it out, you can bundle `gem "graphql", github: "rmosolgo/graphql-ruby"` to get the changes on master, or you can wait for 1.11.7 to be released soon.

PR #3246, opened 2020-12-03
Update object generator to use Float for AR decimal columns

newx:
This PR is an attempt to fix `object_generator` as described on https://github.com/rmosolgo/graphql-ruby/issues/3244

Given the following schema:

```ruby
create_table "posts", force: :cascade do |t|
    t.string "title"
    t.text "body"
    t.decimal "rating", precision: 2, scale: 1
    t.datetime "created_at", precision: 6, null: false
    t.datetime "updated_at", precision: 6, null: false
  end
```

When we generate `Post` object type with:

```ruby
#> rails g graphql:object post comments:[Comment]
create  app/graphql/types/post_type.rb
```

We get the following `app/graphql/types/post_type.rb` output:

```ruby
module Types
  class PostType < Types::BaseObject
    field :id, ID, null: false
    field :title, String, null: true
    field :body, String, null: true
    field :rating, Types::DecimalType, null: true
    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
    field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
    field :comments, [Types::CommentType], null: true
  end
end
```

With this change, ActiveRecord decimal columns would use `Float` graphql type instead of custom `Types::DecimalType`.

`app/graphql/types/post_type.rb`

**NOTE**: See the change for `:rating` field

```ruby
module Types
  class PostType < Types::BaseObject
    field :id, ID, null: false
    field :title, String, null: true
    field :body, String, null: true
    field :rating, Float, null: true
    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
    field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
    field :comments, [Types::CommentType], null: true
  end
end
```

### DIFF
--- a/lib/generators/graphql/object_generator.rb
+++ b/lib/generators/graphql/object_generator.rb
@@ -40,6 +40,8 @@ module Graphql
         case type_expression
         when "Text"
           ["String", null]
+        when "Decimal"
+          ["Float", null]
         when "DateTime", "Datetime"
           ["GraphQL::Types::ISO8601DateTime", null]
         when "Date"
--- a/spec/integration/rails/generators/graphql/object_generator_spec.rb
+++ b/spec/integration/rails/generators/graphql/object_generator_spec.rb
@@ -10,6 +10,7 @@ class GraphQLGeneratorsObjectGeneratorTest < BaseGeneratorTest
       t.datetime :created_at
       t.date :birthday
       t.integer :points, null: false
+      t.decimal :rating, null: false
     end
   end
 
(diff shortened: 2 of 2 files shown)


rmosolgo (approved):
This looks great. Thanks so much for the fix!

tbpeder:
You may want to add this to the case statement as well:

```
when “UUID”
    [“String”, null]
```

rmosolgo:
🤔 @tbpeder , how about using `ID` in that case? (It's serialized as a string, but it would express the _kind_ of String that the field returns.)

tbpeder:
@rmosolgo My feeling is that it would be safer to treat UUIDs as strings since they have to be enclosed in quotes everywhere in the code. Maybe that's not a problem with GraphQL itself, but it would be safer for any code that relies on the types themselves.

PR #3247, opened 2020-12-03
Remove rewrite from Phases of execution doc

swalkinshaw:
This step/phase no longer applies when the interpreter is used which is the default now.

I figured it's easier just to remove this entirely instead of adding some note/explanation about it being interpreter dependent?


rmosolgo:
LGTM 👍

PR #3251, opened 2020-12-10
Move CI to GitHub actions

rmosolgo:
It seems like Travis CI isn't running builds anymore 🤷  I know their free plans have been ... changing. So why not just migrate? 

TODO: 

- [x] Merge workflows?
- [x] Build out the matrix 
- [x] Consider removing some gemfiles? 
- [x] Split mongodb and postgres services into jobs that need them (rather than adding the overhead every time)
- [x] Remove .travis.yml 
- [x] Disable builds on travis's website


jturkel:
Awesome to see this change! The Travis build times have been killing me for other open source projects. It looks like roughly ~80% of the graphql-ruby build time is installing gems. Is it worth using the [cache action](https://github.com/actions/cache/blob/main/examples.md#ruby---bundler)?

rmosolgo:
I'm getting really close, but: 

- Why does it think postgresql isn't running? (Well, it _isn't_ running, but why not?) 
- Why does adding `gem "mongoid", "~> 5.4.0"` cause ruby/setup-ruby to hang forever?

rmosolgo:
😹  Finally, it works! 

I guess I misunderstood the significance of `POSTGRES_PASSWORD`; I picked an arbitrary string, but I guess it _has_ to be `postgres` 🤷

rmosolgo:
😌

PR #3253, opened 2020-12-15
Clarify documentation on using field extension options

rtymchyk:
I was puzzled on how to retrieve the options passed to an `extension`, since the hash passed to `after_resolve` simply did not contain them. I poked around at the test specs and realized that `options` is actually an attribute on the `GraphQL::Schema::FieldExtension`, not a function parameter of `after_resolve`.


rmosolgo:
Thanks for the correction here! What a confusing typo 😖

Issue #3255, opened 2020-12-16
Changelog notes for 1.9.19, 1.9.20 and 1.9.21


jasonkarns:
We're looking to update from 1.9.17 to... something newer and are going through the release notes. Do a less conservative update to latest 1.9 jumps to 1.9.21 despite those versions not being listed as Releases on GitHub nor are there changelog notes for them.

Are there any details about the changes in these versions?

rmosolgo:
Hey, sorry about the confusion there. Those changelog entries probably aren't on `master` because, after 1.10 was released, changes to 1.9.x would have been committed to `1.9.x` instead of `master`. The best changelog for those 1.9 versions is found on the `1.9.x` branch:

https://github.com/rmosolgo/graphql-ruby/blob/1.9.x/CHANGELOG.md

If you think those entries should be on master, that's fine with me; feel free to open a PR for it 👍

PR #3256, opened 2020-12-17
Adds 1.9.19, 1.9.20, and 1.9.21 changelog entries

jasonkarns:
Backports the changelog entries from the 1.9.x branch so they can be easily discovered on github.

I initially didn't notice that there were changelog entries that only existed on the 1.9.x branch. This makes them discoverable on master.

from https://github.com/rmosolgo/graphql-ruby/issues/3255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,6 +153,12 @@ FieldExtension: pass extended values instead of originals to `after_resolve` #31
 - Fix compatibility of `YYYY-mm-dd` with `Types::ISO8601DateTime` #2989
 - Remove unused ivar in InputObject #2987
 
+## 1.9.21 (12 June 2020)
+
+### Bug fixes
+
+- Fix `extras:` on subscription fields #2983
+
 ## 1.10.11 (11 June 2020)
 
 ### New features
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for porting these over! Sorry I wasn't clear about this at first, but could you please order the changelog entries by _version_ instead of by _date_ (ie, put 1.9.19 after 1.9.18, and so on)? That way, people who are upgrading will find adjacent entries. (That's how 1.8.x changelog entries were done, I'm open to reconsidering if you have a case for ordering them by release date instead.)

jasonkarns:
I don't have a _strong_ case. For users, I don't think it matters too much because I would imagine that they typically would do a quick Find to search by version.

My only rationale for going by date is that generally changelog files are autogenerated by tooling. And all the tooling I've ever used (in both node and ruby land), prepends the latest release at the top, regardless what version is actually being released. So it's my anecdotal experience that changelog entries are usually ordered by date. (But then again, this is a very small sample size because there are drastically few libraries which still maintain multiple major/minor release lines.)

So that's my rationale. I'm more than happy to re-order if you still prefer.

rmosolgo:
Thanks for updating this! I'll reorder the entries later if I think of it.

PR #3278, opened 2021-01-12
Add support to Ruby 3.0.0

lucianosousa:


rmosolgo:
Thanks for starting on this! Weird that it doesn't show any builds 🤔

lucianosousa:
@rmosolgo it's showing in my fork [here](https://github.com/lucianosousa/graphql-ruby/runs/1691360567)
going to work on the fixes this week

rmosolgo:
Oh, nice! Interesting that the `.inspect` of that class changed. Maybe the way of assigning a name to a dynamically-created class is different in ruby 3?

lucianosousa:
Yeah, a lot has changed. feel free to help me here if you have time. I'm still unsure of how I'm going to fix some things

rmosolgo:
I'll share my thoughts about those failures in case it helps: 

- Wrong class name: 

  ```diff 
   -"Jazz::ReturnInvalidNull::ReturnInvalidNullPayload::InvalidNullError"
  +"#<Class:0xXXXXXX>::InvalidNullError"
  ```
  
  These classes are created here: 
  
  https://github.com/rmosolgo/graphql-ruby/blob/c60d7db6046bb4fe2bb1eda2d366b25a748ca8ff/lib/graphql/schema/object.rb#L77
  
   And the assertion that's failing is about its `inspect` output. That method is here: 
   
   https://github.com/rmosolgo/graphql-ruby/blob/f9d330c8cd1b5a7d1fcf33af51fc97ae6c14baae/lib/graphql/invalid_null_error.rb#L41-L47
   
   You can see that, on previous Rubies, it had a normal class name, but now, it's outputting the default dynamic class `inspect` output (for example, `Class.new.inspect` will output something like that). So somehow, the way GraphQL-Ruby is defining and naming that class needs to be improved. Or, maybe the inspect method needs to be improved ... 
   
 - Bad arguments 
 
   I'm surprised I didn't see warnings on previous Rubies for this, but here: 
   
    ```
    ArgumentError: wrong number of arguments (given 1, expected 2)
      /home/runner/work/graphql-ruby/graphql-ruby/lib/graphql/relay/base_connection.rb:61:in `initialize'
      /home/runner/work/graphql-ruby/graphql-ruby/spec/support/star_wars/schema.rb:313:in `initialize'
    ```
    
    and here: 
    
    ```
    ArgumentError: wrong number of arguments (given 1, expected 2)
      /home/runner/work/graphql-ruby/graphql-ruby/lib/graphql/relay/base_connection.rb:61:in `initialize'
      /home/runner/work/graphql-ruby/graphql-ruby/spec/support/star_wars/schema.rb:313:in `initialize'
    ```
    
    I suspect that these were always broken, but updated argument handling in Ruby 3 really made them :boom:. Here's the thing: 
    
     - The code path in the error message is for the "new" connection system (`GraphQL::Pagination::Connections`)
     - But the `def initialize` method there is for a class that extends the _legacy_ connection system (`GraphQL::Relay::...`):
       
       https://github.com/rmosolgo/graphql-ruby/blob/f9d330c8cd1b5a7d1fcf33af51fc97ae6c14baae/spec/support/star_wars/schema.rb#L311-L314
       
     - It seems like the test _explicitly_ calls the new connection system (`connections.add(...)`) with this legacy class: 
     
       https://github.com/rmosolgo/graphql-ruby/blob/f9d330c8cd1b5a7d1fcf33af51fc97ae6c14baae/spec/support/star_wars/schema.rb#L439
       
     I don't think that should have _ever_ worked, but I bet the previous argument handling in Ruby made it work _well enough_. I think the solution here is to: 
     
     - Make a new Connection wrapper class that extends `GraphQL::Pagination::ActiveRecordRelationConnection`
     - Make it work the same way as the legacy one 
     - Hook that new class up in the same way, instead of the legacy class

rmosolgo:
Here's a passing build on @lucianosousa's repo 🎉 https://github.com/lucianosousa/graphql-ruby/runs/1724211670

rmosolgo:
Thanks for kicking this off, @lucianosousa! I pushed a few more commits to sort out those other issues.

lucianosousa:
oh @rmosolgo, thank you a lot! Sorry for the delay here

PR #3279, opened 2021-01-12
Correct PR number in CHANGELOG

starsirius:
Hi 👋 . I am updating my application and think the breaking `authorized!` method signature changes in the CHANGELOG should link to #2520, instead of #2536.


rmosolgo:
Oops, thanks for the correction!

PR #3281, opened 2021-01-13
Add GraphQL::Types::BigInt to documentation

yokoto:
It should also be written in documentation for https://github.com/rmosolgo/graphql-ruby/pull/2150 because it doesn't work if field is defined as `BigInt`, but `GraphQL::Types::BigInt`.

Please check if the explanation of BigInt is suitable.


rmosolgo:
Thanks for the improvement!

PR #3292, opened 2021-01-21
Add GraphQL::Deprecation.warn

### LINKED ISSUE #3290: Allow usage of deprecated functionality to be treated as an error

There are lots of deprecation warnings for v2.0, and I'd like to catch those in CI and migrate off the functionality now.

We use `config.active_support.deprecation = :raise` in the development and test environments in our Rails application to do this for Rails itself.

Use a custom method to warn about deprecation, and allow configuration to raise an error rather than just warning.

Rather than implementing this entirely from scratch, it would also solve our problem if graphql-ruby used `ActiveSupport::Deprecation.warn` if it is defined, and otherwise fell back to `Kernel#warn`.

rmosolgo:
Fixes #3290

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -149,7 +149,7 @@ require "graphql/unauthorized_error"
 require "graphql/unauthorized_field_error"
 require "graphql/load_application_object_failed_error"
 require "graphql/dataloader"
-
+require "graphql/deprecation"
 
 module GraphQL
   # Ruby has `deprecate_constant`,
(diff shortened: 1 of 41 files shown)


mrcasals:
Hi @rmosolgo! After this PR I'm getting this error:

```
uninitialized constant GraphQL::Deprecation_ID_TYPE
```

It seems to be a problem with the find-and-replace, just like in commit 105de58. Unfortunately, this got into release 1.12.2, so a new release is required to fix this!

rmosolgo:
Ick, thanks for reporting this. I think you're right. I'll take a look, @mrcasals.

rmosolgo:
Fixed in 5133a157e and shipped in 1.12.3 -- let me know how that version works for you, @mrcasals !

mrcasals:
Awesome, thanks for the quick response! ❤️

PR #3298, opened 2021-01-24
Add example code for the `message` argument on validates.

connorshea:
I was stuck for a bit because I had expected the `message` argument to be an argument on `validates`, rather than inside the hash passed to `required`. This adds a code example to the docs for validators to make usage more obvious.

Also fix the link to the RequiredValidator API Docs.


rmosolgo:
Thanks for the fix here!

PR #3303, opened 2021-01-27
Pass symbol to respond_to? to fix NameError

mikepack:
I encountered this error while attempting to monkey patch `GraphQL::ExecutionError`:

```ruby
require 'graphql/execution_error'

module ExecutionErrorMonkeyPatch
  def extensions
    {
      code: 'EXECUTION_ERROR',
      group: 'GRAPHQL',
    }.merge(Hash(super))
  end
end

GraphQL::ExecutionError.prepend(ExecutionErrorMonkeyPatch)
```

This change should fix the error, but I'm going to use this opportunity to ask how I can accomplish this without monkey patching. Any ideas?

### DIFF
--- a/lib/graphql/query/variable_validation_error.rb
+++ b/lib/graphql/query/variable_validation_error.rb
@@ -23,7 +23,7 @@ module GraphQL
         # a one level deep merge explicitly. However beyond that only show the
         # latest value and problems.
         super.merge({ "extensions" => { "value" => value, "problems" => validation_result.problems }}) do |key, oldValue, newValue|
-          if oldValue.respond_to? merge
+          if oldValue.respond_to?(:merge)
             oldValue.merge(newValue)
           else
             newValue
--- a/spec/graphql/query/variable_validation_error_spec.rb
+++ b/spec/graphql/query/variable_validation_error_spec.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe GraphQL::Query::VariableValidationError do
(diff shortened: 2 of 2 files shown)


rmosolgo:
Wow, how did it _ever work_ 🤔 !? 

> accomplish this without monkey patching

It seems like you want _all_ ExecutionErrors, including ones raised by gem internals, to include those keys in their hashes. Besides extending the base class in some way, the only thing I can think of is adding those keys to the response _after_ execution has finished, for example: 

```ruby 
result = MySchema.execute(...) 
result["errors"].each do |err|
  if !err.key?("code")
    err["code"] = "EXECUTION_ERROR"
  end 
  if !err.key?("group")
    err["group"] = "GRAPHQL"
  end 
end 
```

Something like that might go in `graphql_controller.rb`, or it might be accomplished in a [query instrumenter](https://graphql-ruby.org/queries/instrumentation.html).

~~I'd love to merge this patch, but I'd like to include a test for the expected behavior (which was previously untested, evidently 😖 ). Are you willing to add a test for it?~~ Honestly, I can't think of how to test this without monkey-patching the whole gem. I'll take a look.

mikepack:
> Something like that might go in graphql_controller.rb, or it might be accomplished in a query instrumenter.

Thanks for the tips, that's helpful.

I updated this PR with a spec which does reproduce the error. I used a bunch of Structs as doubles, which may not be ideal. Happy to iterate if you think there's a more concrete way.

rmosolgo:
This is great, thanks for your work on it (and sorry it slipped off my radar til now)!

Issue #1733, opened 2018-08-02
mutation argument's `loads` option depend on gid setup


cainlevy:
It appears that the [`loads:`](https://github.com/rmosolgo/graphql-ruby/pull/1609) option available for finding and authorizing mutation inputs will only work in situations where the id argument specifies both class _and_ id (e.g. github.com/rails/globalid). Is this the Right Way to use GraphQL-Ruby, or a limitation of the first pass at implementation?

## Proposal

`loads` could accept a function that takes responsibility for the following behavior:

```
* Add a prepare hook to fetch an object with the given ID (using Schema.object_from_id)
* Check that the fetched object’s type matches the loads: type (using Schema.resolve_type)
```

For example:

```ruby
argument :employee_id, ID, required: true, loads: ->(id) { Employee.find(id) }
```

rmosolgo:
> Is this the Right Way

Kinda _yes_, it's based on the Relay practice of using UUIDs for each object. So, `loads:` leverages the same mechanism as the `Query.node(id: ID!)` field. 

But, I think we could improve the flexibility a bit by modifying the overrideable methods in `Resolver`. I've found a couple similar issues while trying this stuff out in GitHub, so I'll open a PR soon.

rmosolgo:
Resolvers now have `def load_#{argument}` methods which can be overriden, so that would probably help in this case, for example:

```ruby
argument :employee_id, ID, required: true, loads: Types::Employee

def load_employee(id)
  ::Employee.find(id)
end 
```

chrise86:
@rmosolgo I get
```
NotImplementedError (MySchema.object_from_id(node_id, ctx) must be implemented to load by ID (tried to load from id `10`)):
```
when trying to load objects like this.

PR #3304, opened 2021-01-29
Document resolvers have an overridable load method

jeffreykwan:
I noticed the overridable `load_#{argument}` method from this issue: (https://github.com/rmosolgo/graphql-ruby/issues/1733#issuecomment-510072753) on resolvers wasn't documented and found it useful.


rmosolgo:
Thanks for the improvement!

PR #3305, opened 2021-01-29
Replace .for with .with in Dataloader Overview

DmitryTsepelev:
Looks like this doc was written when there was a method `Dataloader#for`, at least now there is no such method 🙂


rmosolgo:
Thanks for the fix!

PR #3319, opened 2021-02-05
Dataloader source errors

### LINKED ISSUE #3317: Error handlers are not handling errors inside dataloader sources
**Describe the bug**

Errors raised inside dataloader sources are not being rescued by error handlers.

**Versions**

`graphql` version: current master

**GraphQL schema, query & steps to reproduce**

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "graphql", git: "git@github.com:rmosolgo/graphql-ruby.git"
end

class ErrorSchema < GraphQL::Schema
  class ErrorSource < GraphQL::Dataloader::Source
    def fetch(_)
      raise ArgumentError, "Nope"
    end
  end

    field :error, String, null: false
    field :dataloader_error, String, null: false

    def error
      raise ArgumentError, "Nope"
    end

    def dataloader_error
      dataloader.with(ErrorSource).load(123)
    end
  end

  use GraphQL::Dataloader
  query(Query)

  rescue_from(StandardError) { nil }
end

ErrorSchema.execute("{ error }")            # Error is being rescued
ErrorSchema.execute("{ dataloaderError }")  # Error isn't being rescued
```

**Expected behavior**

The error raised inside the source object should get handled.

**Actual behavior**

The error isn't handled and the script crashes.

coding-chimp:
Applying the changes suggested in #3317.

We're now rescuing all the errors happening in dataloader sources while fetching records and then re-raise the error later in the GraphQL runtime context where they can be handled by any defined error handlers.

Fixes #3317

### DIFF
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -43,7 +43,8 @@ module GraphQL
       # @param keys [Array<Object>] Loading keys which will be passed to `#fetch` (or read from the internal cache).
       # @return [Object] The result from {#fetch} for `keys`. If `keys` haven't been loaded yet, the Fiber will yield until they're loaded.
       def load_all(keys)
-        if pending_keys = keys.select { |k| !@results.key?(k) }
+        if keys.any? { |k| !@results.key?(k) }
+          pending_keys = keys.select { |k| !@results.key?(k) }
           @pending_keys.concat(pending_keys)
           sync
         end
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -3,9 +3,6 @@
 module GraphQL
   class Dataloader
     class Source
-      # @api private
-      attr_reader :results
-
       # Called by {Dataloader} to prepare the {Source}'s internal state
       # @api private
       def setup(dataloader)
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -35,11 +35,11 @@ module GraphQL
       # @return [Object] The result from {#fetch} for `key`. If `key` hasn't been loaded yet, the Fiber will yield until it's loaded.
       def load(key)
         if @results.key?(key)
-          @results[key]
+          result_for(key)
         else
           @pending_keys << key
           sync
-          @results[key]
+          result_for(key)
         end
       end
 
@@ -52,7 +52,7 @@ module GraphQL
           sync
         end
 
-        keys.map { |k| @results[k] }
+        keys.map { |k| result_for(k) }
       end
 
       # Subclasses must implement this method to return a value for each of `keys`
@@ -86,7 +86,24 @@ module GraphQL
         fetch_keys.each_with_index do |key, idx|
           @results[key] = results[idx]
         end
+      rescue StandardError => error
+        fetch_keys.each { |key| @results[key] = error }
+      ensure
         nil
+      end
+
+      private
+
+      # Reads and returns the result for the key from the internal cache, or raises an error if the result was an error
+      # @param key [Object] key passed to {#load} or {#load_all}
+      # @return [Object] The result from {#fetch} for `key`.
+      # @api private
+      def result_for(key)
+        result = @results[key]
+
+        raise result if result.class <= StandardError
+
+        result
       end
     end
   end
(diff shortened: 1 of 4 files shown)


rmosolgo:
This is looking great, thanks for your work here! I think there are a couple more uses of`results[key]` to address, which are used for `Dataloader#request(key)` and `Dataloader#request_all(key)`:

- https://github.com/rmosolgo/graphql-ruby/blob/f93b1f6eb0d8ee3ae10fce2be35eebf54f8a524b/lib/graphql/dataloader/request.rb#L11-L21
- https://github.com/rmosolgo/graphql-ruby/blob/f93b1f6eb0d8ee3ae10fce2be35eebf54f8a524b/lib/graphql/dataloader/request_all.rb#L14-L19

I think the thing to do there is: 

- Expand the current test to cover the uses of `.request(...)` and `.request_all(...)` (so that those tests fail) 
- Then, refactor `Dataloader::Request` and `Dataloader::RequestAll` to use your new `result_for` method 
- Add a new method, `Source#loaded?(key)`, to replace `results.key?(k)` usages, then remove `attr_reader :results` (this will guarantee that `@results` is a private concern of `Source`, and nobody gets results except via `loaded?` and `result_for(...)`)

How does that sound to you?

coding-chimp:
Oh yeah, somehow I imagined the `load` and `load_all` methods were also used for request resolution, but then work stuff happened and forgot to actually check. 🙈

Makes total sense, I’ll update the PR later!

rmosolgo:
>  `load` and `load_all` methods were also used for request resolution

There _might_ be some way to do it that way, I'm totally open to it if you find one!

coding-chimp:
Okay, using `load` and `load_all` in the request classes pretty much just works, however there is now some extra work being done if the result isn't already loaded. 

https://github.com/rmosolgo/graphql-ruby/blob/17b9837b4bda053a083cb993a814b2fdc422d178/lib/graphql/dataloader/source.rb#L40

https://github.com/rmosolgo/graphql-ruby/blob/17b9837b4bda053a083cb993a814b2fdc422d178/lib/graphql/dataloader/source.rb#L50-L51

I don't think those array will ever big enough for this to have any significance, though.

coding-chimp:
Simplified `#load_all` a little to get rid off the extra work.

lib/graphql/dataloader/source.rb
```diff
@@ -35,24 +32,23 @@ def request_all(keys)
...
+          result_for(key)
         end
       end
 
       # @param keys [Array<Object>] Loading keys which will be passed to `#fetch` (or read from the internal cache).
       # @return [Object] The result from {#fetch} for `keys`. If `keys` haven't been loaded yet, the Fiber will yield until they're loaded.
       def load_all(keys)
-        if keys.any? { |k| !@results.key?(k) }
-          pending_keys = keys.select { |k| !@results.key?(k) }
+        if pending_keys = keys.select { |k| !@results.key?(k) }
```
rmosolgo, line 46: Evidently there's no test for this, but I think this will not have the desired effect when all of `keys` are present in `@results`. For example: 

```
results = { a: 1, b: 2, c: 3 }
keys = [:a, :b, :c]
if pending_keys = keys.select { |k| !results.key?(k) }
  puts "has pending keys: #{pending_keys}"
else
  puts "has no pending keys"
end

# $ ruby pending_test.rb
# has pending keys: []
```

Even when all `keys` are filtered out, `pending_keys` is `[]`, which is _truthy_, so it would enter the `if ...` block here. 

Sorry there's no test for it, but how about either: 

- Adding a test to demonstrate that the current implementation isn't subject to the bug I described above 
- Or, reverting the implementation to the previous one 

? 

Alternatively, if you really want to skip the extra Array, there probably _is_ a zero-allocation approach, something like 

```ruby 
def load_all(keys)
  any_pending = false 
  
  keys.each do |k|
    if !@results.key?(k)
      any_pending = true 
      @pending_keys << k 
    end 
  end 
  
  if any_pending 
    sync 
  end 
  
  @keys.map { |k| result_for(k) }
end 
```

That'd be fine with me too. But I want to make sure we don't `sync` when all of the given `keys` are already present in `@results`.

coding-chimp: Yeah, you're right. Sorry about that. Feels like more effort than it's worth so I reverted the change.

rmosolgo:
🤔 I must have something wrong in my GitHub Actions config. The commit shows passing CI on your fork: https://github.com/coding-chimp/graphql-ruby/commit/2163752c68b437f4d7d18e3305f1cb6ca4752f69

Thanks for making this improvement happen!

PR #3325, opened 2021-02-08
Allow types to be prepended by a module

splattael:
Prior this commit the following failed with `SystemStackError`:

```ruby
  module WithNameField
    def self.prepended(base)
      base.field :name, String, null: false
    end
  end

  module NamedEntity
    include BaseInterface
    prepend WithNameField
  end

  # This raised
  # graphql-ruby/lib/graphql/schema/member/has_fields.rb:80:in `find': stack level too deep (SystemStackError)
```

This :point_up: is a reproduction of [some code from the GitLab](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/53629/diffs#note_504260946).

### DIFF
--- a/lib/graphql/schema/find_inherited_value.rb
+++ b/lib/graphql/schema/find_inherited_value.rb
@@ -20,7 +20,9 @@ module GraphQL
         if self.is_a?(Class)
           superclass.respond_to?(method_name, true) ? superclass.send(method_name) : default_value
         else
-          (ancestors - [self]).each do |ancestor|
+          ancestors_except_self = ancestors
+          ancestors_except_self.delete(self)
+          ancestors_except_self.each do |ancestor|
             if ancestor.respond_to?(method_name, true)
               return ancestor.send(method_name)
             end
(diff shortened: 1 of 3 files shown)


splattael:
@rmosolgo Do you have time to review this bugfix? :pray:

splattael:
Not sure if the failing CI pipelines are related to this PR because I've [seen them somewhere else](https://github.com/rmosolgo/graphql-ruby/actions?query=is%3Afailure++). On the other hand it also read `SystemStackError: stack level too deep` so... :shrug:

rmosolgo:
:+1: Sure, I'll take a look! Yeah, there's something messed up with some of the tests, but I think it's only when Rails is loaded multiple times (in the generator tests, maybe?). I addressed it on master by stubbing out the method that raised that error 😅

rmosolgo:
Thanks for the fix! I pushed a tiny refactor -- it looks like `field_class` was implemented before I made `find_inherited_value`, so I DRYed that up, and I switched the implementation to one that didn't allocate extra arrays.

splattael:
@rmosolgo Wow, now THAT was quick :astonished: :zap: ...and it's already released :tada: 

Thanks for merging and the nice refactor :bow: :heart:

PR #3333, opened 2021-02-11
Dataloader passes along throws

### LINKED ISSUE #3331: `throw` from GraphQL::Dataloader jobs results in `UncaughtThrowError`
This is a bit of an unconventional situation, I'm trying out the dataloader outside of a graphql context but I thought it's worth logging this problem just in case it can surface in graphql, perhaps in fields. This is probably best explained with a short example:

```ruby
catch(:hello) do
  dataloader = GraphQL::Dataloader.new
  dataloader.append_job do
    throw :hello
  end
  dataloader.run # => UncaughtThrowError: uncaught throw :hello
end
```

Currently, the `GraphQL::Dataloader` class detects error classes and raises them as expected. However, a `throw` inside the client code gets turned into a `UncaughtThrowError` instead of just propagating up until a `catch` is encountered. If the code that handles the other errors is updated to handle `UncaughtThrowError` and re-throw the `tag` attribute, it should be seamless.

**Versions**

`graphql` version: 1.12.4
`rails` (or other framework): rails 6

**Steps to reproduce**

See description above ☝️ 

**Expected behavior**

Something thrown from within a Fiber running a Dataloader job can be caught by code outside the job.

**Actual behavior**

The fiber finishes without the thrown value ever being caught, resulting in `UncaughtThrowError`.

coding-chimp:
Fixes #3331. Let me know if I should change something. 🙂

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -95,7 +95,7 @@ module GraphQL
         else
           # These fibers were previously waiting for sources to load data,
           # resume them. (They might wait again, in which case, re-enqueue them.)
-          f.resume
+          resume(f)
           if f.alive?
             next_fibers << f
           end
@@ -109,7 +109,7 @@ module GraphQL
               job.call
             end
           }
-          result = f.resume
+          result = resume(f)
           if result.is_a?(StandardError)
             raise result
           end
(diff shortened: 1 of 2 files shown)


mclark (approved):
Wow you're fast! FWIW, LGTM 👍

rmosolgo (approved):
Thanks for taking a crack at this! It looks great.

PR #3334, opened 2021-02-11
Allow using graphql@15 in javascript client

anark:
Right now, If I use the javascript client(specifically ActionCableLink) in a project that uses the javascript graphql dependency at version 15 my bundle will include 2 full versions of the graphql dependency(one at 14 for the graphql-ruby-client dependency and version 15 from the product)

This _SHOULD_ make it so you can use either graphql 14 OR 15 without having duplicate dependencies in your bundle, but I am having trouble testing that.

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -34,7 +34,7 @@
     "@types/actioncable": "^5.2.3",
     "@types/pusher-js": "^4.2.2",
     "glob": "^7.1.4",
-    "graphql": "^14.3.1",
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the improvement, judging by https://docs.npmjs.com/cli/v6/configuring-npm/package-json#dependencies, I think you're right.

rmosolgo:
🚢 graphql-ruby-client 1.9.1, please let me know if you run into any trouble with it!

anark:
That worked great for me and reduced the duplication and bundle size!

PR #3335, opened 2021-02-11
Sketch out a general-use dataloader API

rmosolgo:
We were talking about this at work, just posting it here in case anyone else is curious. 

Interestingly, @mclark also identified the fact that Dataloader might be useful _without_ Fibers in cases when the loading is determinate (ie, in Ruby source code, not in dynamic, client-provided GraphQL queries). I'll add that to this example sometime, too.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,5 @@
 name: CI Suite
 on:
-  - push
   - pull_request
 
 jobs:
--- a/guides/dataloader/testing.md
+++ b/guides/dataloader/testing.md
@@ -1,0 +1,73 @@
+---
+layout: guide
+search: true
+section: Dataloader
+title: Testing
+desc: Tips for testing Dataloader implementation
+index: 4
+experimental: true
+---
+
+There are a few techniques for testing your {{ "GraphQL::Dataloader" | api_doc }} setup.
+
+## Integration Tests
+
+One important feature of `Dataloader` is how it manages database access while GraphQL runs queries. You can test that by listening for database queries while running queries, for example, with ActiveRecord:
+
+
+```ruby
+def test_active_record_queries_are_batched_and_cached
+  # set up a listener function
+  database_queries = 0
+  callback = lambda {|_name, _started, _finished, _unique_id, _payload| database_queries += 1 }
+
+  query_str = <<-GRAPHQL
+  {
+    a1: author(id: 1) { name }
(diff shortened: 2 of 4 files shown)


bessey:
I doubt you'll be surprised to hear I am interested 😋. Just to clarify, by general-use do you mean outside of graphql-ruby, in any old Ruby code?

I have a different vision of a general-use API. Since the original Fiber based dataloader proposal, kicking in the back of my head has been the idea of the ORM itself integrating with the dataloader API, via fiber-locals (Thread.current, confusingly). 

I.e. a world where its possible for the ORM to detect prior to data fetching, whether it is running inside a dataloader and change its behaviour, without changing its API, negating the need for a verbose dataloader API for many cases.

At the simplest level this might look like
```ruby
class NotActiveRecord::Base
  def self.find(id)
    if dataloader = Thread.current[:dataloader]
      dataloader.with(self.dataloader_source).load(id)
    else
      really_find(id)
    end
  end
end
```
and to benefit from this API, _something_ up the stack has to be appending N+1 jobs, as a toy example, generic serialization:
```ruby
def render_collection(collection)
  GraphQL::Dataloader.with_dataloading { |dataloader|
    Thread.current[:dataloader] = dataloader
    collection.map do |record|
      dataloader.append_job { render(record) }
    end.join("\n")
  }
end

def render(record)
  {
    # Lets just pretend that `where` calls `NotActiveRecord::Base.find` so i don't have to write more boilerplate 😅 
    children: Log.where(record_id: record.id).map(&:name)
  }
end
```

I recognise this is a lot more "magic" though, and would need really great debug logging. If it got popular though, it would be pretty amazing to see people solving many of their N+1 problems just with
```ruby
MyORM.include Dataloader::SourceIntegrations::MyORM
MySerializers.include Dataloader::ContextIntegrations::MySerializer
```

I hope that makes some sense...

rmosolgo:
Yes, I was exploring something just like that here. I had a coworker who _was_ exploring the use of this module for other purposes (but isn't anymore). In his experimenting, though, it turned that for non-GraphQL contexts, it was simpler to _manually_ batch calls (eg, collecting IDs in an array, then running an `id: [...]` query with them), so this isn't really going anywhere. 

The point stands, though: there's nothing GraphQL-_specific_ about this dataloader, it's just that GraphQL is where it really shines.

rmosolgo:
I dusted off this PR add added some docs because, as pointed out in #3602, a method like this would be helpful for testing.

PR #3337, opened 2021-02-13
Fix the URL of "Tracking Schema Changes with GraphQL-Ruby" used in the schema_structure of guide.

kindaidai:
I could not access the URL before the fix, so I found the new URL from your(@rmosolgo ) blog and fixed it.


rmosolgo:
Sorry for the trouble, and thanks for the fix! I double-checked that there aren't any other links that need an update.

PR #3338, opened 2021-02-13
Support max_page_size config in resolver classes, fixes #3257

### LINKED ISSUE #3257: Specify max_page_size on field resolvers
Current code:

```ruby
module Objects
  class Post < Types::Object
    field :id, ID, null: false
  end
end

module Types
  class Queries < Types::Object
    field :all_posts, resolver: ::Resolvers::Post::Index
  end
end

module Resolvers::Post
  class Index < Types::Resolver
    type Objects::Post.connection_type, null: false

    def resolve(**args)
      Post.all
    end
  end
end
```

Currently, max_page_size is only support on the direct field declaration:

```ruby
field :all_posts, resolver: ::Resolvers::Post::Index, max_page_size: 20
```

But this separates logic. It would be nice to have support for max_page_size in the resolver class

```ruby
type Objects::Post.connection_type, null: false, max_page_size: 20
```

or as a main class method:

```ruby
module Resolvers::Post
  class Index < Types::Resolver
    type Objects::Post.connection_type, null: false
    max_page_size 20
  end
end
```

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3257

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -276,8 +276,29 @@ module GraphQL
           end
         end
 
+        # Get or set the `max_page_size:` which will be configured for fields using this resolver
+        # (`nil` means "unlimited max page size".)
+        # @param max_page_size [Integer, nil] Set a new value
+        # @return [Integer, nil] The `max_page_size` assigned to fields that use this resolver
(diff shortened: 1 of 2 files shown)


rmosolgo:
I wish I could figure out why the build is failing 😖

PR #3343, opened 2021-02-16
Try to fix generators loading and warnings

rmosolgo:
I don't have a clue why Generators suddenly started to fail loading. I'll see if I can get the build passing again.


rmosolgo:
😂

Issue #3348, opened 2021-02-19
graphql-ruby-client shouldn't depend on react


jaxn:
**Describe the bug**

**Versions**

`graphql` version:
`rails` (or other framework):
other applicable versions (`graphql-batch`, etc)

**GraphQL schema**

_any_

**GraphQL query**

_any_

**Steps to reproduce**

1. create a sample app with Vue and Apollo
    ```sh
    vue create grc-bug .
    vue add apollo
    yarn add graphql-ruby-client@1.8.2
    ```
1. add `PusherLink` to `src/vue-apollo.js`
    ```js
    import { PusherLink } from "graphql-ruby-client"  // eslint-disable-line no-unused-vars
    ```
1. verify older version works
    ```sh
    yarn serve
    
    >   App running at:
    >  - Local:   http://localhost:8081/ 
    >  - Network: http://192.168.185.2:8081/
    >
    >  Note that the development build is not optimized.
    >  To create a production build, run yarn build.
    ```
1. update `graphql-ruby-client` and see failed dependency on `react`
    ```sh
    yarn add graphql-ruby-client@1.9.0
    
    yarn serve
    
    > This dependency was not found:
    >
    > * react in ./node_modules/graphql-ruby-client/node_modules/@apollo/client/react/context/ApolloConsumer.js, 
    > ./node_modules/graphql-ruby-client/node_modules/@apollo/client/react/hooks/useApolloClient.js and 1 other
    >
    > To install it, you can run: npm install --save react
    ```

**Expected behavior**

`graphql-ruby-client` should not depend on `react`

**Actual behavior**

With the move to Apollo 3, `@apollo/client` has a dependency on `react`.

All imports from `@apollo/client` should be moved to `@apollo/client/core` as noted in the Apollo 3 Migration guide: [Using Apollo Client without React](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#using-apollo-client-without-react).

**Additional context**

With these details, we can efficiently hunt down the bug!

PR #3270, opened 2020-12-30
Upgrade apollo-link to @apollo/client

sirwolfgang:
`apollo-link` has since been deprecated, this works to upgrade these dependencies; With aims to work towards fixing https://github.com/rmosolgo/graphql-ruby/issues/3236. 

Part of the upgrade changes includes a change to the `Observables` support to use [Zen Obserables](https://github.com/zenparsing/zen-observable) which changes the interface for Pusher. I tried to update this as well, but I don't understand the current code well enough to get it sorted.

PR #3349, opened 2021-02-19
Remove unintentional dependency on React

jaxn:
Quick fix for #3348 

This is just a find/replace to change imports from `@apollo/client` to `@apollo/client/core`. The JS tests pass, but I have not yet done more thorough testing.  My plan is to test this more thoroughly in our application's QA environment tomorrow.

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -22,7 +22,6 @@
     "pako": "^2.0.3",
     "prettier": "^1.19.1",
     "pusher-js": "^7.0.3",
-    "react": "^17.0.1",
     "ts-jest": "^25.2.0"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


rmosolgo:
This looks great, thanks! Personally, if the tests pass, I'm :+1:, but let me know when you're ready to merge. Oh, there's one more thing, I think this can be removed: 

https://github.com/rmosolgo/graphql-ruby/blob/f88aa11da6ce6ff254cfa761dc8563d19bb6eb02/javascript_client/package.json#L25

(It was added in #3270 -- I didn't realize it would _also_ cause other projects to require `react` 🤦 )

jaxn:
I went ahead and removed `react` from `devDependencies` (good catch!). I also verified that this PR resolves the build error in our application. I am ready to merge.

Thank you for the quick reply and for all your work on this library @rmosolgo!

rmosolgo:
🚢 'd in graphql-ruby-client 1.9.2, thanks again for the fix!

PR #3354, opened 2021-02-21
Improve active record dataloader example

iguchi1124:
It is returns always empty because ActiveRecord primary key is being Integer instance by default and ID scalar type is regarded as String instance.

I suppose it needs `String#to_i` method call in this guides.


rmosolgo:
In the long run, the best approach would be to use the column to cast the value: 

https://github.com/Shopify/graphql-batch/blob/9fa2dd255e7d0801271471566e3c80083f0adbea/examples/record_loader.rb#L10

But this is a good improvement to the doc, thanks!

PR #3362, opened 2021-02-24
Suggest @rails/actioncable package for Rails 6

haines:
We had issues with our subscriptions after following the guide, because the back end is running Action Cable 6 but the front end was running Action Cable 5. The client package [got renamed to `@rails/actioncable`](https://github.com/rails/rails/pull/34905) in Rails 6.


rmosolgo:

Issue #3345, opened 2021-02-17
Some queries slower when enabling the Interpreter.


RynoPi:
When we enable the interpreter we realize that some of our queries run slower than before while others run faster. Our assumption was that enabling the interpreter should either speed up or stay the same. Clearly some configuration our setup in some of our types are causing this, but cannot seem to find anything on the internet that would indicate why.

We also tried upgrading to the latest version (which has it enabled by default), but still experience them running slower than disabling the interpreter. We noticed that most of them that run slower have a scope applied so it might be the reason we are having this experience.

Would using a Resolver have worse performance when using the runtime interpreter?

If there is any reason that might indicate why, it would be much appreciated.

rmosolgo:
Hi! It's certainly possible that something might be slower, although I don't know what. In order to debug this, I'll need some specifics: 

- An example GraphQL query string that runs slower on the interpreter than the legacy runtime 
- A benchmark of the query on the interpreter, for example: 

  ```ruby 
   StackProf.run(mode: :wall, out: "interpreter-profile.dump") do 
     MySchema.execute(...)
   end 
   ```
   
   Then share the `.dump` file or at least the output of `stackprof interpreter-profile.dump`
- If certain GraphQL fields or resolvers are identified as obviously the slowest, then sharing the source code of those would be really helpful. 


After you share some of that info, we can dig in!

RynoPi:
Sorry for taking so long to reply. Here is the top 30 with the legacy runtime:
![Screenshot from 2021-02-18 14-39-49](https://user-images.githubusercontent.com/77432729/108358269-31b8fd80-71f7-11eb-904a-13193a84b76a.png)
 
Here is the top 30 with the new interpreter from StackProf:
![Screenshot from 2021-02-18 14-12-17](https://user-images.githubusercontent.com/77432729/108355640-c7eb2480-71f3-11eb-8a95-f2b223a98e79.png)

Regarding the Graphql query I obfuscated some of the data. You will see a couple of fields name `aField` these all represent a different field. Here is the Graphql query:
```
query OrdersSearch($limit: Int!, $skip: Int!, $filter: Filter!, $itemstates: [String!], $weekstart: String, $weekend: String) {
    ordersSearch(filter: $filter) {
      orders(limit: $limit, skip: $skip) {
        ...multipleOrderFields
        __typename
      }
      totalOrders
      totalFiltered
      __typename
    }
  }

  fragment multipleOrderFields on Order {
    id
    state
[...]

PR #3364, opened 2021-02-24
Reduce interpreter overhead

rmosolgo:
Trying to improve #3345 

- Remove overhead by detecting interpreter context keys at read instead of `Context#[]=` at runtime
  - If people are using these keys for other purposes, this could break -- should i take care to preserve compat here?
- refactor set_type_at_path to set_non_null_at_path, only set it when nonnull is true


rmosolgo:
The best of these changes were rolled into #3494

PR #3373, opened 2021-03-04
Remove extra count query for hasNextPage when paginating

vahe:
This PR adds an optimization for an edge case (probably very common code path though) in pagination when querying for `hasNextPage`.

Currently, we get the value of `hasNextPage` by comparing the result of `count` query with limit `first + 1` with `first (query param) + 1`.

We can eliminate the `count` query in cases where nodes are already fetched and there's less than specified in `first` param.


rmosolgo:
This is great -- thanks so much for sharing (and testing) this fix!

PR #3374, opened 2021-03-04
Fix documentation around pusher message decompression

such:


rmosolgo:
Thanks for the fix here!

Issue #3367, opened 2021-02-28
Allow making Relay edges non-nullable


nirvdrum:
**Is your feature request related to a problem? Please describe.**

Currently, all Relay edges are marked as nullable, even when that doesn't make sense for a given application. That means every client of the API needs to account for a situation that will never occur. At best, it's a bunch of wasted checks. At worst, the caller stops believing the schema and ignores that some items could ever possibly be `NULL`.

**Describe the solution you'd like**

I'd like a way to configure the edge type to be non-null and also a way to annotate that the edge consists of non-null values. By way of illustration, _graphql-ruby_ 1.12.5 will produce a schema definition for an edge like:

```graphql
edges: [InvitationEdge]
```

I'd like to be able to control that process so the schema could instead generate:

```graphql
edges: [InvitationEdge!]!
```

**Describe alternatives you've considered**

Currently, we override the `edge_type` method to guide the schema better. Unfortunately, that's relying on _graphql-ruby_ internals that are subject to change. Somewhere between 1.11.4 and 1.12.5, the internals changed in such a way that broke our override (the `edge_class` property on fields was removed). Having an official way of achieving such control over edge definitions would allow us to avoid brittle overrides.

rmosolgo:
That'd be fine with me, how about adding an `edge_nullable` method similar to `node_nullable`? 

https://github.com/rmosolgo/graphql-ruby/blob/8e09a2bb5b615f76e4d18b4d2cfff38939808713/lib/graphql/types/relay/connection_behaviors.rb#L76-L84

Then, it could be applied here instead of the literal `true`: 

https://github.com/rmosolgo/graphql-ruby/blob/8e09a2bb5b615f76e4d18b4d2cfff38939808713/lib/graphql/types/relay/connection_behaviors.rb#L44

?

nirvdrum:
Sorry, I somehow missed your question. I'll give that a try over the next day or two and see how it works.

rmosolgo:
Fixed by #3376

PR #3376, opened 2021-03-05
added option to make edges non-nullable

gopeter:
This solves #3367


rmosolgo:
Fantastic, thanks for the improvement here!

nirvdrum:
This doesn't work in 1.12.6 due to #3386. I was in the process of reporting the issue when I noticed it was already fixed on master. I'll leave this here to help anyone that might run into the same issue prior to the next gem release.

rmosolgo:
Just shipped the fix in 1.12.7 in case anyone is keeping an eye on this!

PR #3386, opened 2021-03-12
Connections: fix logic when boolean attributes are false

rmosolgo:
Oops, these assignments and reads didn't actually work when the values were `false`, since `false` fails the `||` check.


soberstadt:
Is there any chance we can get a publish for this in the near future?

rmosolgo:
Yes, just shipped 1.12.7!

PR #3388, opened 2021-03-12
Add option for connection class level configuration of nodes_field

anark:
This adds the option to set `nodes_field(false)` in your connection classes(base classes) to make this configurable at the class level.


rmosolgo:
Thanks for this improvement! I updated the name to avoid a naming conflict with an existing method, but looks like the build is passing over at https://github.com/anark/graphql-ruby/commit/97f9da8997ffd4a868757dfd7c02b5327856420f

PR #3389, opened 2021-03-12
add class level config to allow for edge.node to be non-nullable

anark:
This adds the option to set `node_nullable(false)` in your edge classes(base edge classes) to make this configurable at the class level.


rmosolgo:
Thanks for this improvement!

PR #3391, opened 2021-03-15
Move typescript & @types packages to dev dependencies

mattvague:
Since the `graphql-ruby-client` package on npm is compiled via tsc, typescript should not be required to run it and therefore I believe `typescript`, and any `@types/**` packages should be a dev dependencies instead of a regular dependencies.  As it it now, it adds an extra 60ish mb your bundle if using `graphql-ruby-client`.

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -9,11 +9,13 @@
   "license": "LGPL-3.0",
   "bin": "./cli.js",
   "devDependencies": {
+    "@types/actioncable": "^5.2.3",
     "@types/glob": "^7.1.1",
     "@types/jest": "^25.1.2",
     "@types/minimist": "^1.2.0",
     "@types/node": "^13.7.1",
     "@types/pako": "^1.0.1",
+    "@types/pusher-js": "^4.2.2",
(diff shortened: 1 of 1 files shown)


rmosolgo:
Good catch -- thanks for the improvement!

mattvague:
> Good catch -- thanks for the improvement!

@rmosolgo No problemo, thank you for the excellent library! On the same topic, I was wondering if maybe `@apollo/client` and `graphql` should be peerDependencies instead so you're not risking having multiple versions / a version conflict in someone's app? If that sounds right am happy to PR that

rmosolgo:
I _think_ that'd be fine!

mattvague:
>I think that'd be fine!

The only issue is that the new version of graphql-ruby-client would break peoples apps and force them to add those dips if they weren't already including `@apollo/client` and `graphql` for some reason.

rmosolgo:
That'd be fine -- I'll mention it in the changelog and bump the version. It's easy to fix and breaks loudly, so I don't think anyone would be surprised by bugs in production. Thanks for thinking it through!

PR #3394, opened 2021-03-16
Remove suggestion to use deprecated functionality

ravangen:
This is now the default, it is no longer necessary to `use GraphQL::Analysis::AST` per https://github.com/rmosolgo/graphql-ruby/blob/5c3a8ec6f45319f40cb59f2ffb5961bb1071aa80/lib/graphql/analysis/ast.rb#L18


rmosolgo:
Oops, thanks for fixing this!

PR #3395, opened 2021-03-17
Move graphql and @apollo/client to peerDeps

mattvague:
As discussed [here](https://github.com/rmosolgo/graphql-ruby/pull/3391) 😃


rmosolgo:
Thanks again for this improvement. 🚢 'd in graphql-ruby-client 1.9.3!

PR #3400, opened 2021-03-18
Fix typo in errors/error_handling.md guide

MasterOdin:
This seems like a typo where `an` was meant.


rmosolgo:
Thanks for the fix!

PR #3405, opened 2021-03-23
Update validation_errors documentation string

ravangen:
Updates documentation to reflect `validation_errors` can include `GraphQL::Query::VariableValidationError` due to variable validation. https://github.com/rmosolgo/graphql-ruby/blob/46cae92e6e87be8adade48e5a5004843bab8c48f/lib/graphql/query/validation_pipeline.rb#L80

Additionally removes a `binding.pry` statement that I assume was not meant to be included in `master`.

### DIFF
--- a/lib/graphql/subscriptions/broadcast_analyzer.rb
+++ b/lib/graphql/subscriptions/broadcast_analyzer.rb
@@ -35,9 +35,6 @@ module GraphQL
           pt = @query.possible_types(current_type)
           pt.each do |object_type|
             ot_field = @query.get_field(object_type, current_field.graphql_name)
-            if !ot_field
-              binding.pry
-            end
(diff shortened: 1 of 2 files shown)


lib/graphql/subscriptions/broadcast_analyzer.rb
```diff
@@ -35,9 +35,6 @@ def on_enter_field(node, parent, visitor)
           pt = @query.possible_types(current_type)
           pt.each do |object_type|
             ot_field = @query.get_field(object_type, current_field.graphql_name)
-            if !ot_field
-              binding.pry
```
rmosolgo, line 39: 🤦  Thanks!

rmosolgo (approved):
Thanks for this improvement!

Issue #3365, opened 2021-02-24
AST Analysis on invalid queries?


ravangen:
Should an AST analysis be able to be invoked on a query that has failed static validation? Specifically errors like `FieldsAreDefinedOnTypeError`, `FieldsHaveAppropriateSelectionsError`, `ArgumentsAreDefinedError`, `FragmentTypesExistError` where the warden was used to get the field/argument/type.

For example, consider a type that used to be visible, is no longer so, resulting in static validation failure. It would be great if we could opt into running an analyzer in order to record the usage of something previously visible and no longer available.

```ruby
# In this example sketch, `Types::Comment` and `Types::UserErrors` are visible but not `Mutations::CreateComment`
class Types::UserErrors < GraphQL::Schema::Object
  field :message, String, null: false
  field :field, [String], null: true
end

class Mutations::CreateComment < Mutations::BaseMutation
  def self.visible?(context)
    # Hide mutation based on feature flag :remove_example_mutations
    super && !context[:viewer].feature_enabled?(:remove_example_mutations)
  end

  argument :body, String, required: true
  argument :post_id, ID, required: true

  field :comment, Types::Comment, null: true
  field :errors, [Types::UserErrors], null: false

  def resolve(body:, post_id:)
    post = Post.find(post_id)
    comment = post.comments.build(body: body, author: context[:current_user])
    if comment.save
      # Successful creation, return the created object with no errors
      {
        comment: comment,
        errors: [],
      }
    else
      # Failed save, return the errors to the client
      {
        comment: nil,
        errors: comment.errors
      }
    end
  end
end
```

```graphql
mutation CreateCommentTestMutation {
  createComment(id: "1", body: "hello") {
    ...InjectedTestFragment
  }
}

fragment InjectedTestFragment on CreateCommentPayload {
  errors {
    message
    field
  }
  comment {
    id
  }
}
```

```ruby
class LogHiddenByVisibility < GraphQL::Analysis::AST::Analyzer
  extend T::Sig

  sig { params(query_or_multiplex: T.any(GraphQL::Query, GraphQL::Execution::Multiplex)).void }
  def initialize(query_or_multiplex)
    super
    @query = query_or_multiplex
    @context = @query.context
  end

  sig { returns(T::Boolean) }
  def analyze?
    # NOTE: This does not consider `@context.query.valid?`, it makes a best effort from the information available
    super && !@context[:introspection]
  end

  sig { void }
  def result; end

  sig do
    params(
[...]

PR #3406, opened 2021-03-23
Store additional details in select static validation errors

ravangen:
## Objective

Ability to analyze static validation errors that can be caused due to scheme member visibility. Where possible, the analysis would look at the corresponding type definitions of related schema members (argument, enum type, enum value, field, input object, interface, mutation, object, union).

Related to [#3365 - AST Analysis on invalid queries?](https://github.com/rmosolgo/graphql-ruby/issues/3365)

The following errors would be considered for this purpose:
- `GraphQL::StaticValidation::ArgumentLiteralsAreCompatibleError`
- `GraphQL::StaticValidation::ArgumentsAreDefinedError`
- `GraphQL::StaticValidation::FieldsAreDefinedOnTypeError`
- `GraphQL::StaticValidation::FragmentTypesExistError`
- `GraphQL::Query::VariableValidationError`

## Problem

A couple static validation errors do not currently contain enough information to try and locate a schema member's definition when it exists but is hidden. While processing the static validation rules, we already lookup such definitions though the warden in order to determine if an error should be raised, and we can include the definition's reference as part of error (but not sent as part of the response to a client).

#### `ArgumentsAreDefinedError`

The existing arguments do not provide enough detail to locate the specific argument's type definition (if it even exists) from among all arguments belong to directives, fields, and input objects. We have a string value of the arguments name, but not a good sense of its surroundings.

#### `ArgumentLiteralsAreCompatibleError`

Similar to `ArgumentsAreDefinedError`, we are missing enough detail to locate the specific argument's type definition. Additionally, if the argument's expected type is an unwrapped enum, we would want to try and match the provided value (currently missing) to one of the possible enum values to check its visibility.

### DIFF
--- a/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
+++ b/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
@@ -41,7 +41,9 @@ module GraphQL
               error_options = {
                 nodes: parent,
                 type: kind_of_node,
-                argument: node.name
+                argument: node.name,
+                argument_defn: arg_defn,
+                value: node.value
               }
               if coerce_extensions
                 error_options[:coerce_extensions] = coerce_extensions
--- a/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
+++ b/lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
@@ -41,8 +41,8 @@ module GraphQL
               error_options = {
                 nodes: parent,
                 type: kind_of_node,
-                argument: node.name,
-                argument_defn: arg_defn,
+                argument_name: node.name,
+                argument: arg_defn,
                 value: node.value
               }
               if coerce_extensions
--- a/spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
+++ b/spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
@@ -333,6 +333,35 @@ describe GraphQL::StaticValidation::ArgumentLiteralsAreCompatible do
     end
   end
 
+  describe "error references argument" do
+    let(:validator) { GraphQL::StaticValidation::Validator.new(schema: schema) }
+    let(:query) { GraphQL::Query.new(schema, query_string) }
+    let(:errors) { validator.validate(query)[:errors] }
+    let(:query_string) {"
+      query {
+        cheese(id: true) { source }
+        milk(id: 1) { source @skip(if: TRUE) }
+      }
+    "}
+
+    it "works with field" do
+      id_argument = schema.types['Query'].fields['cheese'].get_argument('id')
+      error = errors.find { |error| error.argument_name == 'id' }
+
+      assert_equal id_argument, error.argument
+      assert_equal true, error.value
+    end
+
+    it "works with directive" do
+      if_argument = schema.directives['skip'].get_argument('if')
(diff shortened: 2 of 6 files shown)


lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
```diff
@@ -41,7 +41,9 @@ def on_argument(node, parent)
               error_options = {
                 nodes: parent,
                 type: kind_of_node,
-                argument: node.name
+                argument: node.name,
+                argument_defn: arg_defn,
```
rmosolgo, line 45:
```suggestion
                argument_name: node.name,
                argument: arg_defn,
```

lib/graphql/static_validation/rules/argument_literals_are_compatible.rb
```diff
@@ -41,7 +41,9 @@ def on_argument(node, parent)
...
+                value: node.value
```
ravangen, line 46: I considered storing `GraphQL::Language::Printer.new.print(node.value)` instead but figured it is better to keep the higher fidelity value

rmosolgo (commented):
In general, :+1: from me to make these errors more useful. 

I added some suggestions, regarding bikeshedding about accessor and argument names. I'd like the input names to match the accessor names, and I'd like to avoid the abbreviation `defn` in the public API if possible. Would those changes work ok for you? 

Also, in order to maintain compatibility in the future, these new methods will need some unit tests. (I imagine you were waiting for an initial 👀  before following up with those, just thought I'd mention it.) 

If you have in mind similar additions for those other classes, :+1: from me. 

Anything else I can do to help on this?

ravangen:
Thanks for the quick feedback, I will get on it and hopefully be ready later today.

Regarding tests, I wasn't sure if the existing tests cases would be enough to cover the additional arguments been passed into the initializers. I'll see what can be done to test reading these new attributes.

rmosolgo:
> arguments been passed into the initializers

Yes, I agree it covers that, but it doesn't cover the readers. (The changes here could be deleted and the existing tests would still pass -- but if your app used those readers, you would notice the regression!) 

You could probably hook into the existing tests and use `query.validation_errors` to find instances of the relevant error classes, then make an assertion about what the readers return.

spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
```diff
@@ -333,6 +333,35 @@
...
+    let(:errors) { validator.validate(query)[:errors] }
+    let(:query_string) {"
+      query {
+        cheese(id: true) { source }
+        milk(id: 1) { source @skip(if: TRUE) }
+      }
+    "}
+
+    it "works with field" do
+      id_argument = schema.types['Query'].fields['cheese'].get_argument('id')
```
ravangen, line 348: Any suggestions about better alternatives to declare the expected value?

ravangen:
@rmosolgo do you think it would be possible to get this reviewed and possibly merged today? Thanks again!

spec/graphql/static_validation/rules/argument_literals_are_compatible_spec.rb
rmosolgo: This all seems fine to me. If breaks later on, it will break loudly, and it will be straightfoward to update at that time.

rmosolgo (approved):
Thanks for this improvement!

PR #3408, opened 2021-03-25
Tracing: consistent sequence of events for Query#result

### LINKED ISSUE #3393: Tracing inconsistency
**Describe the bug**

The order of tracing events is inconsistent depending on how you execute a query.

If you instantiate a `GraphQL::Query` with a query string and schema, then call `#result` on it, then the `lex` and `parse` trace events fire outside of `execute_multiplex`.

If instead you call `Schema.execute` or `Schema.multiplex`, then the `lex` and `parse` trace events fire inside of `execute_multiplex`.

This is because `Query` calls `Multiplex.run_queries` inside of `with_prepared_ast`: https://github.com/rmosolgo/graphql-ruby/blob/5c3a8ec6f45319f40cb59f2ffb5961bb1071aa80/lib/graphql/query.rb#L198-L200

Where-as  `Schema.multiplex` calls `Multiplex.run_all` directly which fires the `execute_multiplex` tracer before it does anything that triggers lexing and parsing.

@rmosolgo maybe this isn't technically a bug but it's surprising enough it should at least be called out in the documentation.

**Versions**

1.12.5, efcedefd04d9e8d40ccdac99f867edb1cc3e667d, probably master

cc @chrisbutcher

rmosolgo:
Fixes #3393

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -18,7 +18,7 @@
     "@types/pusher-js": "^4.2.2",
     "@types/react": "^17.0.0",
     "@types/zen-observable": "^0.8.2",
-    "ably": "^1.2.4",
+    "ably": "1.2.6",
     "jest": "^25.0.0",
     "nock": "^11.0.0",
     "pako": "^2.0.3",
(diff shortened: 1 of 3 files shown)


chrisbutcher:
Thank you!

PR #3411, opened 2021-03-26
Add example of nullable options to Types::Relay::BaseConnection

gtkatakura:
Add to docs of GraphQL::Types::Relay::BaseConnection the new options to customize nullable options and presence of nodes field.


rmosolgo:
Thanks!

PR #3414, opened 2021-03-27
Optimize ActiveRecordRelationConnection#has_next_page

### LINKED ISSUE #3401: Optimise GraphQL::Pagination::RelationConnection has_next_page query

We recently updated the gem version from 1.11.5 to 1.12.6 and as a result of this the default type of connections for ActiveRecord::Relation was changed from the using the GrahpQL::Relay::RelationConnection to GraphQL::Pagination::Connection as defined by [v1.12.0](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#1120-20-january-2021)

Unfortunately we started getting timeouts on some of our complex queries that were requesting `hasNextPage` due to the different type of query being generated by `has_next_page` in the connection class. [GraphQL::Pagination::Connection](https://github.com/rmosolgo/graphql-ruby/blob/e1516208593ce15c059c4066a8b045e5e5cf9f1b/lib/graphql/pagination/relation_connection.rb#L38) is generating a count query as opposed to [GrahpQL::Relay::RelationConnection](https://github.com/rmosolgo/graphql-ruby/blob/9069ba310846dea7636dd24ca0c743eccf4beb3d/lib/graphql/relay/relation_connection.rb#L30) which uses `.exists?`

The same problem was actually addressed previously for GrahpQL::Relay::RelationConnection with [this](https://github.com/rmosolgo/graphql-ruby/issues/2332) issue and [this](https://github.com/rmosolgo/graphql-ruby/pull/2547/files) PR, handling ActiveRecord::Relations differently. If GraphQL::Pagination::Connection is going to be the new default connection, including for ActiveRecord::Relations, can we copy across that optimisation in to `has_next_page` on that class as well?

We've reverted to using the Relay connections in the meantime with `self.connections = nil` in the schema but with the new connection type written this way it's not going to work for some of the queries we're doing. So if it wasn't optimised we would probably have to roll our own connection with a different type of query in `has_next_page` but I doubt we would be the only people having this issue with complex queries and counts so I think it should be addressed by the gem.

coding-chimp:
Optimizes the ActiveRecordRelationConnection#has_next_page method by using `exists?` instead of a full count statement.

Fixes #3401.

### DIFF
--- a/lib/graphql/pagination/active_record_relation_connection.rb
+++ b/lib/graphql/pagination/active_record_relation_connection.rb
@@ -5,6 +5,26 @@ module GraphQL
   module Pagination
     # Customizes `RelationConnection` to work with `ActiveRecord::Relation`s.
     class ActiveRecordRelationConnection < Pagination::RelationConnection
+      def has_next_page
+        if @has_next_page.nil?
+          @has_next_page = if before_offset && before_offset > 0
+            true
+          elsif first
+            if @nodes && @nodes.count < first
+              false
+            else
+              initial_offset = sliced_nodes.offset_value || 0
+              sliced_nodes.offset(initial_offset + first).exists?
+            end
+          else
+            false
+          end
+        end
+        @has_next_page
+      end
+
+      private
+
       def relation_count(relation)
         int_or_hash = if relation.respond_to?(:unscope)
           relation.unscope(:order).count(:all)
--- a/lib/graphql/pagination/active_record_relation_connection.rb
+++ b/lib/graphql/pagination/active_record_relation_connection.rb
@@ -5,25 +5,12 @@ module GraphQL
   module Pagination
     # Customizes `RelationConnection` to work with `ActiveRecord::Relation`s.
     class ActiveRecordRelationConnection < Pagination::RelationConnection
-      def has_next_page
-        if @has_next_page.nil?
-          @has_next_page = if before_offset && before_offset > 0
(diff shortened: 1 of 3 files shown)


lib/graphql/pagination/active_record_relation_connection.rb
```diff
@@ -5,6 +5,26 @@ module GraphQL
...
+            else
+              initial_offset = sliced_nodes.offset_value || 0
+              sliced_nodes.offset(initial_offset + first).exists?
+            end
+          else
+            false
+          end
+        end
+        @has_next_page
+      end
```
coding-chimp, line 24: I pretty much just copied over the `has_next_page` method from RelationConnection and changed the relevant part. We could remove some duplication between the two classes:

```ruby
def has_next_page
  return @has_next_page unless @has_next_page.nil?
  return super if before_offset && before_offset > 0 || !first

  @has_next_page = if @nodes && @nodes.count < first
    false
  else
    initial_offset = sliced_nodes.offset_value || 0
    sliced_nodes.offset(initial_offset + first).exists?
  end
end
```

This would make it harder to understand the method, though.

coding-chimp: I moved the check into a private method and am now just overriding that in ActiveRecordRelationConnection.

rmosolgo:

PR #3416, opened 2021-03-29
Introspection support for deprecated directive arguments

jturkel:
Looks like I missed support for deprecated directive arguments in #3015. 🤦  This means graphql-js generated [introspection queries](https://github.com/graphql/graphql-js/blob/main/src/utilities/getIntrospectionQuery.js) with the `inputValueDeprecation: true` option don't work against graphql-ruby servers.


rmosolgo (approved):
Thanks for fixing it!

PR #3428, opened 2021-04-07
Fix serialize for single key hash

### LINKED ISSUE #3425: GraphQL::Subscriptions::Serialize fails for Single Key Hashes
**Describe the bug**

When loading subscriptions using `GraphQL::Subscriptions::Serialize`, single-size hashes are erroneously serialized to `nil`. This causes significant problems when using GraphQL subscriptions.

**Versions**

`graphql` version: 1.12.6
`rails` (or other framework):  6.1

**Steps to reproduce**

The behaviour can be reproduced by the following test: 

```ruby
test 'serialize' do
  subscription = JSON.dump(subscription: '12345', variables: { id: 1})

  result = GraphQL::Subscriptions::Serialize.load(subscription)

  assert_not_nil(result['variables'])
end
```

The above test will fail, as `{ id: 1 }` is loaded is `nil`. 

**Expected behavior**

The expected behavior would be, of course, to receive `{ id: 1 }`

**Additional context**

For context, we use a variant of the `ActionCableSubscription` where we implement `read_subscription` similar to this (we load subscriptions from a shared Redis instance):

```ruby
def read_subscription(subscription_id)
    query = @serializer.load(@subscriptions.get(subscription_id)).with_indifferent_access
    ...
end
```

I believe I have found the root cause of this issue in [this commit](https://github.com/rmosolgo/graphql-ruby/commit/8160028564ab167ad37bfb958536567fa2317f09). Previously, a condition like this was present: 

```ruby 
if value.size == 1 && value.key?(GLOBALID_KEY)
  GlobalID::Locator.locate(value[GLOBALID_KEY])
elsif value.size == 1 && value.key?(SYMBOL_KEY)
  value[SYMBOL_KEY].to_sym
else 
  # handle 'normal' hashes
end
```

However, the commit referenced above changed it to: 

```ruby 
 if value.size == 1
    case value.keys.first # there's only 1 key
    when GLOBALID_KEY
      ...
    when TIMESTAMP_KEY
     ...
    end # There is no else block here
else 
  # handle 'normal' hashes
end
```

Because of the way the conditions are structured here, the case where the value (aka. `{ id: 1}`) has size 1, but *hasn't* some sort of special key is missing - which then results in `nil` being returned. 

The issue can be fixed by adding an else statement: 

```ruby
else 
   key = value[value.keys.first]
  value = load_value(value[key]) 
  { key: value } # something like that
end
```

hschne:
This fixes #3425. 

Not sure about the tests. In order to create the error condition, we must not use `serialize_dump`, as this appends additional keys, so I just use `to_json` instead :shrug:

### DIFF
--- a/lib/graphql/subscriptions/serialize.rb
+++ b/lib/graphql/subscriptions/serialize.rb
@@ -71,6 +71,9 @@ module GraphQL
               when OPEN_STRUCT_KEY
                 ostruct_values = load_value(value[OPEN_STRUCT_KEY])
                 OpenStruct.new(ostruct_values)
+              else
+                key = value.keys.first
+                { key => load_value(value[key]) }
               end
             else
               loaded_h = {}
--- a/spec/graphql/subscriptions/serialize_spec.rb
+++ b/spec/graphql/subscriptions/serialize_spec.rb
@@ -64,4 +64,11 @@ describe GraphQL::Subscriptions::Serialize do
     reloaded = serialize_load(serialized)
     assert_equal os, reloaded, "It reloads #{os.inspect} from #{serialized.inspect}"
   end
+
+  it "can deserialize single key hash" do
+    os = { 'a' => 1 }
+    serialized = os.to_json
+    reloaded = serialize_load(serialized)
+    assert_equal os, reloaded
+  end
 end


rmosolgo:
Awesome, thanks for the fix!

rmosolgo:
🚢  in 1.12.8!

PR #3431, opened 2021-04-09
Fix find_handler_for method

ogidow:
parent_handler will be assigned the result of find_handler_for method.
https://github.com/rmosolgo/graphql-ruby/blob/ba3a0ceb7b61182270406fb61e98621afadb30ba/lib/graphql/execution/errors.rb#L142

Since the parent_handler expects a hash object with :handler and :class as keys, find_handler_for method must maintain that interface
https://github.com/rmosolgo/graphql-ruby/blob/ba3a0ceb7b61182270406fb61e98621afadb30ba/lib/graphql/execution/errors.rb#L150

### DIFF
--- a/lib/graphql/execution/errors.rb
+++ b/lib/graphql/execution/errors.rb
@@ -115,7 +115,7 @@ module GraphQL
           if obj.is_a?(GraphQL::Schema::Object)
             obj = obj.object
           end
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for the fix! I added a test, too.

rmosolgo:
🚢  in 1.12.8!

PR #3432, opened 2021-04-12
Fix error handlers inheritance

jonian:
Hi @rmosolgo, in version 1.12.7 the error handlers inheritance is broken. Method `find_handler_for` returns a proc, but when searching the schema superclass using `find_handler_for` it expects a hash to be returned.


rmosolgo:
Thanks for reporting this and contributing a fix! I just shipped 1.12.8 which included a similar fix contributed in #3431.

PR #3439, opened 2021-04-15
Documentation: Change `allow_nil` to the correct `allow_null`

kathawala:
According to the docs (in section "Constructor Details"), `allow_null` is the name of the option accepted by all validators not `allow_nil`

(docs: https://graphql-ruby.org/api-doc/1.12.8/GraphQL/Schema/Validator.html)


rmosolgo:
Thanks for the fix, sorry about that!

PR #3443, opened 2021-04-18
Subscriptions broadcast: only call deserialization once

### LINKED ISSUE #3426: GraphQL::Subscriptions stream_from makes extra DB queries when using __gid__  serialization
**Describe the bug**

I'm trying to do some triggers to broadcast a list of AR records to subscribers. I have a standalone AC app. The code the main app code that does triggering looks like this:
```ruby
MySchema.subscriptions.trigger('postUpdates', { post_id: 'Z2lkOi8vcFwdG9yLlzc3VlLzY3' }, {updates: {comments: [Comment.last, Comment.first],  action: 'comments_added' }} )
```
My `PostSubscription` is broadcastable with 
`use GraphQL::Subscriptions::ActionCableSubscriptions, broadcast: true, default_broadcastable: true`

And I have 2 active subscribers.

When I execute the trigger I can see 4 DB queries performed on the action cable side
```
  Comment Load (1.9ms)  SELECT "comments".* FROM "comments" WHERE "comments"."id" = 71 LIMIT 1
  Comment Load (2.3ms)  SELECT "comments".* FROM "comments" WHERE "comments"."id" = 14 LIMIT 1
  ...
  Comment Load (1.7ms)  SELECT "comments".* FROM "comments" WHERE "comments"."id" = 71 LIMIT 1
  Comment Load (2.0ms)  SELECT "comments".* FROM "comments" WHERE "comments"."id" = 14 LIMIT 1
```
After doing some investigation I found that deserialization [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions/action_cable_subscriptions.rb#L149) is calculated for every subscriber even though the value it returns is only used to calculate the message once for the first subscriber and then skip it for anybody else with [this](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions/action_cable_subscriptions.rb#L152) condition.

I'm wondering if that was done intentionally or there is a case I don't know about. If not I can probably make a PR to fix this 

**Versions**
    graphql (1.12.6)
    graphql-batch (0.4.3)
    graphql-pro (1.17.11)
    rails (6.0.3.6)

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok).
Are you using [interpreter](https://graphql-ruby.org/queries/interpreter.html)? Any custom instrumentation, etc?

```ruby
module Subscriptions
  class Post < BaseSubscription
    argument :post_id, ID, required: true # , loads: Types::PostType

    field :updates, Types::Subscriptions::PostBroadcastType, null: false
  end
end
```

**Expected behavior**

Only call deserialization once 

**Actual behavior**

DeSerialization is called for each subscriber

EugeneGrigorenko:
Fixes #3426

- Call `@serializer.load` only for the `initial_event` for the broadcastable subscriptions
- For non-broadcastable subscriptions cache serialization based on the message, so the same message is only serialized once.

### DIFF
--- a/lib/graphql/subscriptions/action_cable_subscriptions.rb
+++ b/lib/graphql/subscriptions/action_cable_subscriptions.rb
@@ -96,6 +96,7 @@ module GraphQL
         @action_cable_coder = action_cable_coder
         @serializer = serializer
         @transmit_ns = namespace
+        @decoded_messages = {}
         super
       end
 
@@ -146,7 +147,6 @@ module GraphQL
       def setup_stream(channel, initial_event)
         topic = initial_event.topic
         channel.stream_from(stream_event_name(initial_event), coder: @action_cable_coder) do |message|
-          object = @serializer.load(message)
           events_by_fingerprint = @events[topic]
           events_by_fingerprint.each do |_fingerprint, events|
             if events.any? && events.first == initial_event
@@ -154,6 +154,7 @@ module GraphQL
               # so just run it once, then deliver the result to every subscriber
               first_event = events.first
               first_subscription_id = first_event.context.fetch(:subscription_id)
+              object = decode_message(message)
               result = execute_update(first_subscription_id, first_event, object)
               # Having calculated the result _once_, send the same payload to all subscribers
               events.each do |event|
@@ -210,6 +211,11 @@ module GraphQL
 
       def stream_event_name(event)
         [EVENT_PREFIX, @transmit_ns, event.topic].join
+      end
+
+      # cache message in case the same message is sent to events with different fingerprints
+      def decode_message(message)
+        @decoded_messages[message] ||= @serializer.load(message)
       end
     end
   end
--- a/lib/graphql/subscriptions/action_cable_subscriptions.rb
+++ b/lib/graphql/subscriptions/action_cable_subscriptions.rb
@@ -96,7 +96,6 @@ module GraphQL
         @action_cable_coder = action_cable_coder
         @serializer = serializer
         @transmit_ns = namespace
-        @decoded_messages = {}
         super
       end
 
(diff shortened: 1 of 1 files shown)


lib/graphql/subscriptions/action_cable_subscriptions.rb
```diff
@@ -211,6 +212,11 @@ def stream_subscription_name(subscription_id)
       def stream_event_name(event)
         [EVENT_PREFIX, @transmit_ns, event.topic].join
       end
+
+      # cache message in case the same message is sent to events with different fingerprints
+      def decode_message(message)
+        @decoded_messages[message] ||= @serializer.load(message)
+      end
```
rmosolgo, line 219: I'm concerned about the lifetime of this cache, specifically, imagine: 

- An event is triggered with an object's gid in the payload: `trigger(:profile_was_updated, { user: User.find(1) }, ...)`
- That payload is passed to `decode_message`, which looks up `User.find(1)` and caches it for `"gid://User/1"`
- Later, the profile is updated again, with the same `trigger(...)` call above (so the same payload, too) 
- Won't  `decode_message(...)` return a stale `User` instance from the cache, instead of re-fetching it to get the latest changes? 

Or, is there some reason it wouldn't work out that way, and we could be sure that no stale data would be served from the cache?

rmosolgo (commented):
Looks good, besides the caching issue described above.

lib/graphql/subscriptions/action_cable_subscriptions.rb
EugeneGrigorenko: That seems to be a valid concern. 
I'm not even completely sure we need this cache. Did I understand correctly, that [this](https://github.com/rmosolgo/graphql-ruby/pull/3443/files#diff-a1ea5af1a78917fa653eff83d9ae72180fabd3d9cf616f14dd1871739963cd73R156-R158) code is gonna be run for every subscriber in case the subscription is not "broadcastable"?

rmosolgo: Yes, that's right -- if a subscription isn't "broadcastable", then it gets a unique fingerprint instead of a fingerprint based on the query and context. 

You might be able to use the lexical scope as a cache, for example, assigning `object = nil` above line 152, then using `object ||= ...` inside the loop. That way, you'd have a short-lived cache that lasted for the current update, but didn't spill into future updates.

EugeneGrigorenko: @rmosolgo updated accordingly. Please take a look

rmosolgo:
:+1: Thanks for your work on this improvement!

rmosolgo (approved):

PR #3446, opened 2021-04-19
Make it so you can opt out of object authorization

### LINKED ISSUE #3429: Allow disabling authorization checks if they are not used

We have a few large queries and the authorization checks take up a very large part of the response time, sometimes up to twice as long as the actual response. We do not use the built-in authorization at all.

Some way to disable authorization checks, or change authorization to an opt-in setting rather than always happening?

Uncertain 

**Additional context**

From this [Slack thread](https://app.slack.com/client/T044QJ50C/threads/thread/C07MVL1AB-1615923548.001800)

rmosolgo:
Fixes #3429 

I can't say I _recommend_ it, but if it's really adding up, this makes it possible (and tested) to override `Schema::Object.authorized_new` and bypass authorization. 

I also investigated an _option_ for disabling built-in authorization at runtime, but I found it put too much complexity in the code (https://github.com/rmosolgo/graphql-ruby/compare/skip-built-in-auth). 

Instead, this approach consolidates tracing and logic in `.authorized_new`, then tests to confirm that it can be overridden (including for `DynamicFields`, which implements `__typename`).

So, with this PR, you could opt out by adding: 

```ruby 
class Types::BaseObject 
  # Normally, GraphQL-Ruby calls `.authorized?` before initializing this, 
  # but we want to bypass that 
  def self.authorized_new(obj, ctx)
    new(obj, ctx)
  end 
end 
```

and 

```ruby 
  # Add a custom introspection module which will avoid `.authorized?` checks 
  # for __typename and __type(name: ...) fields
  module Introspection 
     class DynamicFields < GraphQL::Introspection::DynamicFields 
       def self.authorized_new(obj, ctx)
         new(obj, ctx)
       end 
     end 
   end 
   
   introspection(Introspection)
   
   # ... 
 end 
 ```

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -50,7 +50,7 @@ module GraphQL
           root_type = schema.root_type_for_operation(root_op_type)
           path = []
           set_all_interpreter_context(query.root_value, nil, nil, path)
-          object_proxy = authorized_new(root_type, query.root_value, context, path)
+          object_proxy = authorized_new(root_type, query.root_value, context)
           object_proxy = schema.sync_lazy(object_proxy)
           if object_proxy.nil?
             # Root .authorized? returned false.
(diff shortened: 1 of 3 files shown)


bbugh:
Seems good to me!

PR #3447, opened 2021-04-19
Update schema testing documentation

hschne:
For products where the GraphQL server serves one specific frontend, query dumps that list queries that are typically executed can be helpful. `GraphQL::SchemaComparator` lists changes that are breaking in theory, but using `StaticValidation` and aforementioned query dumps, we can get certainty when it comes to which fields can safely be changed or removed.

This PR updates the existing schema structure testing documentation to describe a simple Rake task that can be used to validate queries directly against a Schema. 

What do you think?

### DIFF
--- a/guides/testing/schema_structure.md
+++ b/guides/testing/schema_structure.md
@@ -43,10 +43,7 @@ namespace :graphql do
         puts '⏳  Validating queries...'
         puts "\n"
 
-        validator = GraphQL::StaticValidation::Validator.new(schema: schema)
-        results = queries
-                    .map { |query| query = GraphQL::Query.new(schema, query) }
-                    .map { |query| validator.validate(query) }
+        results = queries.map { |query| schema.validate(query) }
         errors = results
                    .select { |result| result[:errors].present? }
                    .map { |result| result[:errors] }
--- a/guides/testing/schema_structure.md
+++ b/guides/testing/schema_structure.md
@@ -32,7 +32,7 @@ namespace :graphql do
   namespace :queries do
     desc 'Validates GraphQL queries against the current schema'
     task validate: [:environment] do
-      queries_file = 'test/fixtures/files/queries.json
+      queries_file = 'test/fixtures/files/queries.json'
       queries = Oj.load(File.read(queries_file))
 
       Validate.run_validate(queries, MySchema)
--- a/guides/testing/schema_structure.md
+++ b/guides/testing/schema_structure.md
@@ -25,4 +25,50 @@ You can read about this approach in ["Tracking Schema Changes with GraphQL-Ruby"
 
 ## Automatically check for breaking changes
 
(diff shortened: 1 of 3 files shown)


rmosolgo:
I'd love to document this workflow, but, I'm embarrassed how hard it is to do 😅 

In fact, legacy Schema instances had a `#validate` method: 

https://github.com/rmosolgo/graphql-ruby/blob/bace1e4027900fc8779a5c2fd393ff6456046cea/lib/graphql/schema.rb#L358-L373

But, it was never ported to the class-based schema API. Would you be game for reimplementing that method on the class? 

I think the method could be moved to the `class << self` region as-is: 

https://github.com/rmosolgo/graphql-ruby/blob/bace1e4027900fc8779a5c2fd393ff6456046cea/lib/graphql/schema.rb#L855


And for compatibility, add a delegator like this to implement the instance method: 

https://github.com/rmosolgo/graphql-ruby/blob/bace1e4027900fc8779a5c2fd393ff6456046cea/lib/graphql/schema.rb#L714

After that, this doc could use `MySchema.validate(query_str)` ✨  

If you're not interested in moving that method, let me know, and I'll try to make time for it.

hschne:
> Would you be game for reimplementing that method on the class?

Oh absolutely, would gladly do it!

If you don't mind though: I'm a big fan of getting things off my plate, and updating the schema validation is kind of it's own thing. Would it be okay to merge this right now? 

As it stands this documentation update makes sense on its own, right? I would update it in a follow up PR when I also add the `validate` method.

hschne:
@rmosolgo I moved the validate method to the class method as suggested. 

Found out that doing `Schema.validate(query)` already worked before (because there was a delegator that delegated to the instance method of validate), but still went ahead with moving this. The docs have been updated accordingly.

Let me know if you think this needs anything else :slightly_smiling_face:

guides/testing/schema_structure.md
```diff
@@ -25,4 +25,47 @@ You can read about this approach in ["Tracking Schema Changes with GraphQL-Ruby"
...
 
-You can use [GraphQL::SchemaComparator](https://github.com/xuorig/graphql-schema_comparator) to check for breaking changes during development or CI.
+You can use [GraphQL::SchemaComparator](https://github.com/xuorig/graphql-schema_comparator) to check for breaking changes during development or CI. If you maintain a dump of queries that typically run against your server, you may also utilize `GraphQL::StaticValidation` to validate these queries directly. A Rake task such as the one below can be used to identify changes that are incompatible with existing queries.
+
+```ruby
+namespace :graphql do
+  namespace :queries do
+    desc 'Validates GraphQL queries against the current schema'
+    task validate: [:environment] do
+      queries_file = 'test/fixtures/files/queries.json
```
rmosolgo, line 35:
```suggestion
      queries_file = 'test/fixtures/files/queries.json'
```

rmosolgo:
Thanks for updating this! I just noticed one typo in the suggested rake task, I think that's it.

rmosolgo:

Issue #3248, opened 2020-12-03
Missing coercion or validation for default values


eapache:
**Describe the bug**

@benjie pointed out in https://github.com/graphql/graphql-spec/pull/793 that the GraphQL Spec and JS reference implementations are both somewhat confusing/broken when dealing with mistyped default values. I haven't gone through and tried to reproduce all the various combinations of this issue, but I did demonstrate a simple case to show that graphql-ruby suffers from at least a similar class of bug.

**Versions**

`graphql` version: 1.11.4 w/ interpreter, but I'm pretty sure it applies to master too

**GraphQL schema**

```ruby
  field :test, :integer, null: true do |field|
    field.argument(:arg, :integer, required: true, default_value: "string")
  end

  def _test(arg:)
    arg
  end
```

**GraphQL query**

```graphql
query {
  test
}
```

```json
{
  "data": {
    "test": null
  }
}
```

**Expected vs actual behaviour**

There are two interesting points in this example:
- GraphiQL actually says the sample query is invalid; apparently the `default_value` is not making it through into the introspection output, so GraphiQL claims the argument needs a value. However, if you actually *run* it then it works (sorta).
- When you run it, the `default_value` doesn't make it through into the resolver method, which gets `nil` instead. This could really mess with resolver code which assumes a value (since the argument is required-with-a-default at the GraphQL layer).

---

@rmosolgo there is still some debate on what the final specified behaviour should be, but I wanted to put this on your radar. The fact that the default value is getting stripped out in places suggests to me that we are validating it already somewhere, we're just not handling validation failures aggressively enough?

cc @swalkinshaw.

rmosolgo:
Thanks for the heads up on this! I bet the default value gets dropped in `coerce_input`, which returns `nil` for unmanageable values: 

https://github.com/rmosolgo/graphql-ruby/blob/74bbeea30db3cd90c55e8008d55fa14528479802/lib/graphql/types/int.rb#L13

Like you said, I bet it's never validated.

Is there any reason _not_ to apply validation during configuration (ie, during the `argument(...)` method evaluation)?

eapache:
Two potential complications came up during the Working Group meeting:
- Default values are mostly specified as language-native types, not as GraphQL AST tokens or on-the-wire transport bytes, so what it means to validate them is confusing/different from the normally-specified flow. (Though this is maybe more of a spec issue than an implementation issue.)
[...]

PR #3448, opened 2021-04-26
Validate default values

eapache:
For https://github.com/rmosolgo/graphql-ruby/issues/3248

Still some things TODO

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -301,16 +301,26 @@ module GraphQL
 
       private
 
-      def validate_input_type(input_type)
+      def validate_input_type(input_type, wrapped: false)
         if input_type.is_a?(String) || input_type.is_a?(GraphQL::Schema::LateBoundType)
           # Do nothing; assume this will be validated later
+          false
         elsif input_type.kind.non_null? || input_type.kind.list?
-          validate_input_type(input_type.unwrap)
+          type_ready = validate_input_type(input_type.unwrap, wrapped: true)
+          validate_default_value(default_value, input_type) if default_value? && type_ready && !wrapped
+          type_ready
         elsif !input_type.kind.input?
           raise ArgumentError, "Invalid input type for #{path}: #{input_type.graphql_name}. Must be scalar, enum, or input object, not #{input_type.kind.name}."
         else
-          # It's an input type, we're OK
+          # It's an input type, the type itself is OK but make sure the default conforms
+          validate_default_value(default_value, input_type) if default_value? && !wrapped
+          true
         end
+      end
+
+      def validate_default_value(default_value, input_type)
+        default_value = input_type.coerce_isolated_result(default_value) unless default_value.nil?
+        raise "BOOM" unless input_type.valid_isolated_input?(default_value)
       end
 
       def validate_deprecated_or_optional(null:, deprecation_reason:)
--- a/spec/graphql/schema/argument_spec.rb
+++ b/spec/graphql/schema/argument_spec.rb
@@ -209,6 +209,13 @@ describe GraphQL::Schema::Argument do
       res = SchemaArgumentTest::Schema.execute(query_str)
       assert_equal "Argument 'requiredWithDefaultArg' on Field 'field' has an invalid value (null). Expected type 'Int!'.", res['errors'][0]['message']
     end
+
+    it 'validates the type of the default value' do
+      arg = GraphQL::Schema::Argument.new("my_arg", GraphQL::Types::Int, required: true, owner: nil, default_value: "a_string")
+      assert_raises(StandardError) do
+        arg.type
+      end
+    end
   end
 
   describe 'loads' do


eapache:
@rmosolgo cc @benjie @RobertWSaunders this kinda/sorta does the right thing up front, but it's surfaced a bunch of other complications:

1) Enums change. 
```ruby
# Given
class DairyAnimal < GraphQL::Schema::Enum
  value("COW", "Animal with black and white spots", value: 1)
end

# This used to be valid and isn't anymore
argument :animal, DairyAnimal, required: false, default_value: 1
# Instead you need this, which used to be invalid
argument :animal, DairyAnimal, required: false, default_value: "COW"
```
Surprisingly (to me) it seems to get cast either way and the resolver still gets `1` which is important. But when you call `to_graphql` and it runs through the validator you end up with failures like:
```
GraphQL::Schema::InvalidTypeError: Field Query.cheese's return type is invalid: field "similarCheese" argument "nullableSource" default value ["COW"] is not valid for type [DairyAnimal!] (Can't resolve enum DairyAnimal for "COW")
```
And it would be a breaking change for everybody to update their enums.

2) Default values of input objects don't quite work. Given `default_value: { snake_case: "ASC" }` this almost works properly, until it gets to https://github.com/rmosolgo/graphql-ruby/blob/bace1e4027900fc8779a5c2fd393ff6456046cea/lib/graphql/schema/input_object.rb#L184
`argument_name` here is `:snake_case` but the warden only responds to camel-case strings (e.g. "camelCase") so it fails to find the argument and decides it's invalid. This is probably fixable by string-key-ing all your default values, but... gross. I assume the warden could be made string-symbol agnostic though?

3) It doesn't handle wrapped types. I think this is fixable with a bit of effort on my part though.
```ruby
# This doesn't work right now because it tries to validate the wrapped `[1]` with the unwrapped `Integer`.
argument :animal, [GraphQL::Type::Integer], required: false, default_value: [1]
```

4) For some reason it can no longer find `GraphQL::CoercionError` in tests unless I add a bunch of explicit `require`s. No idea what's going on here. Smells like a setup/config issue and not an actual code/logic issue.

eapache:
There are a lot of bad hacks here (look a commit at a time to peel them apart) but I've got the core use case passing, and there's only really two remaining kinds of test failure:
- Validating and printing schemas is broken because it tries to coerce enums on output, but default enum values are now already e.g. `COW` instead of `1`; I tried getting rid of output coercion for defaults here but that caused even more tests to fail and I gave up. I think simply converting enum defaults as I did in this PR is probably the wrong solution anyway.
- Building the schema from definition fails when there's a default value of an input object, since it generates some late-bound types nested inside the default value, which blows up in `InputObject#validate_non_null_input` because they don't respond to `non_null?` or anything normal.

eapache:
Getting closer... two hacks replaced by a more elegant solution which fixed the enum conversion issues, and #3451 has been extracted into its own PR. Still need to look into the build-schema-from-definition issues.

lib/graphql/schema/scalar.rb
```diff
@@ -49,6 +49,7 @@ def validate_non_null_input(value, ctx)
             err
           end
 
+          require 'graphql/coercion_error'
```
rmosolgo, line 52: I'm guessing this is because, previously, this code was _not_ executed while `graphql`'s own files were being loaded by Ruby. But after this change, some default values (perhaps from introspection-related code) are now taking this codepath at load-type. So, the load-time dependency situation changed. 

I encounter this from time to time when doing big refactors, too, and I consider it a code smell pointing at a larger bit of tech debt. When I started adding `require`s for this gem, I didn't really know what I was doing, I basically put them _all_ in `graphql.rb`. (Take a look, you'll see a bazillion `require`s in a row, and 🙀  they're order-dependent.) 

A better system would be if each file, y'know, contained the `require`s it needed in order to operate. It's possible that it would end up _not_ working, if there's some kind of cyclical dependency, but I'm not sure. I just haven't been able to _prioritize_ that work, since everything works fine, and _one_ more hack can't possibly make it any worse 😅 .

eapache:
On consideration, https://github.com/rmosolgo/graphql-ruby/issues/3459 is annoying and weakens any solution to this problem, but isn't technically blocking it.

However, there is one remaining blocking problem which is much trickier: handling default values of type `InputObject` where the `InputObject` type has arguments whose types are late-bound. We end up in `InputObject#validate_non_null_input` for the concrete parent input type before the nested late-bound types are resolved, which means either:
- `InputObject#validate_non_null_input` needs to learn to resolve late-bound types and deal with cycles when they occur
- `Argument#type=` doesn't need to just defer validation when the type is late-bound (as it already does), it needs to defer validation while the type _or any of that type's arguments recursively_ are late-bound, and then remember to trigger validation once that entire recursive graph has been resolved.

Originally I thought this was just an issue with schema-first development and `build_from_definition.rb` which I filed in https://github.com/rmosolgo/graphql-ruby/pull/3472, but I realized that I can build a code-first schema that also triggers this issue. (#3472 is a valid bug regardless though).

@benjie this is basically the "recursion without infinite cycles is hard" problem we were worried about when you wrote the RFC. I think the RFC is still basically correct, and if I were writing a new implementation green-field I could probably make it work, but right now backporting this validation into the existing ruby implementation would seem to require substantial refactoring.

OR, maybe, this means that `type=` isn't the best hook point and I should try building this as a validation rule or something that runs later in the schema lifecycle when all the latebound types have already been resolved. @rmosolgo are there any other hooks that you think might make sense to try?

 I don't have the time or interest to entirely rewrite how late-bound types work; I've got a few more things to explore, but barring a better solution I'm probably going to drop this.

PR #3450, opened 2021-04-27
Respect directive visibility

eapache:
@rmosolgo this validation rule wasn't going through the warden so directives were not respecting their visibility settings.

I did this change via github's UI while investigating so I hope it's right because I haven't actually run it :P

cc @lanadz I think this is sufficient to fix your problem.

### DIFF
--- a/lib/graphql/static_validation/rules/directives_are_defined.rb
+++ b/lib/graphql/static_validation/rules/directives_are_defined.rb
@@ -4,7 +4,7 @@ module GraphQL
     module DirectivesAreDefined
       def initialize(*)
         super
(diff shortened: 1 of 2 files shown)


eapache:
Worth noting: there are other validation rules and other parts of the interpreter that skip the warden when looking up directives, but I think they're all downstream of this one so they won't get hit if this fails anyway.

lanadz:
Thanks! That fixed the issue 👍

rmosolgo:
I pushed a test to confirm this fix. Weirdly, a JS test failed. I'll investigate that separately. Thanks for this improvement!

PR #3451, opened 2021-04-27
Fix test with invalid default value

eapache:
`InspectableInput` has a required argument, so this test was technically
invalid. I discovered this while trying to add validation for default
values. Replace it with a new fully optional input type in this test.

Extracted as the first stand-alone mergeable piece of #3448.

@rmosolgo


rmosolgo:
Thanks!

PR #3456, opened 2021-04-29
Fix introspection of default input objects

eapache:
It seems that the intent was to allow default input object values to be
specified in snake_case, but we weren't camelCasing them on output which
was leading to issues (in introspection and also schema printing).

See e.g. `string_value` which is actually the default for an argument `stringValue` (because graphql-ruby automatically camelCases argument names at _definition time_), in
https://github.com/rmosolgo/graphql-ruby/blob/173ad1babed18d18b43f9988e5d409db4a3ca3b7/spec/support/jazz.rb#L543

Tweak an existing spec to cover this case (as well as the "enums
specifed as values not names" case), then add the missing call to
camelize.

@rmosolgo I'm not actually 100% sure if I'm interpreting the intention here properly. The code is obviously trivial/straightforward but maybe the intention was to force people to write their default values in camelCase and it's the Jazz schema that made a mistake?

This is a second stand-alone extraction from https://github.com/rmosolgo/graphql-ruby/pull/3448.

cc @benjie

### DIFF
--- a/lib/graphql/schema/input_object.rb
+++ b/lib/graphql/schema/input_object.rb
@@ -226,8 +226,8 @@ module GraphQL
         # It's funny to think of a _result_ of an input object.
         # This is used for rendering the default value in introspection responses.
         def coerce_result(value, ctx)
-          # Allow the application to provide values as :symbols, and convert them to the strings
-          value = value.reduce({}) { |memo, (k, v)| memo[k.to_s] = v; memo }
+          # Allow the application to provide values as :snake_symbols, and convert them to the camelStrings
+          value = value.reduce({}) { |memo, (k, v)| memo[Member::BuildType.camelize(k.to_s)] = v; memo }
 
           result = {}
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
You got the intention _right_ -- I _meant_ for the `default_value:` to look like the object that would get passed into resolvers: symbol, snake-cased keys and application values for enums.

rmosolgo:
I'm trying to think, some people might've worked around this issue, will it "break" them if we fix it here? (I think that's ok, but I want to get it right in the changelog.)

eapache:
🤔  I don't think it will break anything unless the user has disabled auto-camelization of argument names in general and is creating snake_case arguments in their actual schema? Otherwise writtenAsCamel default values should continue to pass through, and written_as_snake arguments will start working where they didn't before.

rmosolgo:
Thanks for helping me think through that -- anyone who worked around this before _should_ still be good to go :+1:

PR #3457, opened 2021-04-30
Source keyword arguments

coding-chimp:
The dataloader is currently breaking in Ruby 3 when used with sources that have keyword arguments in their constructor. The splat operator in `Dataloader#with` is transforming the keyword arguments into a hash and if we then try it initialize the source with that hash, it crashes. 

This PR fixes that.

MRI supports the new syntax since v2.7 but apparently, that's [not the case for TruffleRuby](https://github.com/oracle/truffleruby/issues/2004#issuecomment-724697244). That's why I'm only using it in Ruby 3. In general, I'm not a fan of having to duplicate code for different ruby versions but I couldn't come up with a better way to make this backward-compatible. 🤔

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -29,7 +29,12 @@ module GraphQL
 
     def initialize
       @source_cache = Hash.new { |h, source_class| h[source_class] = Hash.new { |h2, batch_parameters|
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for this fix! Sorry I overlooked it in the previous implementation.

PR #3460, opened 2021-04-30
Rescue StandardError on graphql controller.

the-harry:
This PR intends to improve the error handling in the generated controller, since rescuing all exceptions may cause some issues.

### DIFF
--- a/lib/generators/graphql/templates/graphql_controller.erb
+++ b/lib/generators/graphql/templates/graphql_controller.erb
@@ -15,9 +15,9 @@ class GraphqlController < ApplicationController
     }
     result = <%= schema_name %>.execute(query, variables: variables, context: context, operation_name: operation_name)
     render json: result
-  rescue => e
+  rescue StandardError => e
     raise e unless Rails.env.development?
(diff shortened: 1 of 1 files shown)


rmosolgo:
Sorry, I'm not open to style-only changes. Everyone has a preference, and if there's something you think should be enforced by this code base, I recommend that you open an issue for it first. 

If you're interested in reducing the scope of this change, I agree that adding `StandardError` to the `rescue` condition is a good improvement. I'd be happy to accept that in this pull request or a new one.

the-harry:
> Sorry, I'm not open to style-only changes. Everyone has a preference, and if there's something you think should be enforced by this code base, I recommend that you open an issue for it first.
> 
> If you're interested in reducing the scope of this change, I agree that adding `StandardError` to the `rescue` condition is a good improvement. I'd be happy to accept that in this pull request or a new one.

Alright, makes sense! I'll update it! 😄

the-harry:
@rmosolgo i've updated the PR! Let me know if it's ok!

rmosolgo:

Issue #3449, opened 2021-04-27
`use GraphQL::Dataloader` breaks RequestStore usage


gopeter:
**Describe the bug**

I'm using the gem [`acts_as_tenant`](https://github.com/ErwinM/acts_as_tenant) which uses [`RequestStore`](https://github.com/steveklabnik/request_store) to store global values (like the current tenant). But adding `use GraphQL::Dataloader` to the schema breaks this, because `ActsAsTenant.current_tenant` is not defined then:

```ruby
module Types
  class Query < BaseObject
    field :catalog, CatalogType, null: true do
      argument :id, ID, required: true
    end

    def catalog(id:)
      puts "###########"
      puts ActsAsTenant.current_tenant
      puts "###########"

      Catalog.find!(id)
    end
  end
end
```

returns just two rows of `###` 😄

I don't know if this is something that has to be done in the RequestStore gem? Or a bug in `graphql-ruby` itself? Or are both libraries just incompatible and there is no way to resolve this?

I am not sure how Dataloader works, but since it's somewhat coupled to Fiber (?) it maybe have to do with how RequestStore stores its data: in [Thread.current](https://github.com/steveklabnik/request_store/blob/master/lib/request_store.rb) objects.

**Versions**

`graphql` version: 1.12.8
`graphql-pro` version: 1.17.14
`rails` (or other framework): 6.0.3

rmosolgo:
Hey, thanks for the detailed report.

> it maybe have to do with how RequestStore stores its data: in Thread.current objects.

Yes, I bet that's exactly it. When Dataloader is enabled, all GraphQL execution takes place inside a Fiber. As you noticed, `Thread.current` is _empty_ inside a new Fiber. (This caveat is also mentioned [in the docs](https://graphql-ruby.org/dataloader/overview.html#fibers)) 

My first recommendation is to use `context` for those values during GraphQL execution. For example, add 

```ruby 
context = {
  current_tenant: ActsAsTenant.current_tenant
  # ...
}

MySchema.execute(query_str, variables: variables, context: context)
```

To your GraphQL controller. Then, inside fields that require a `current_tenant`, you could use ActsAsTenant's [block support](https://github.com/ErwinM/acts_as_tenant#setting-the-current-tenant-for-a-block): 

```ruby 
    def catalog(id:)
      ActsAsTenant.with_tenant(context[:current_tenant]) do 
        Catalog.find!(id)
       end 
    end
```

[...]

Issue #3366, opened 2021-02-26
GraphQL::Dataloader appears to break trace chain for Datadog


srgoldman:
**Describe the bug**

When inspecting the flame graph or span list in Datadog, I expect to see the related spans aggregated under their parents. When I introduced GraphQL::Dataloader into my schema, that appeared to break the chain (in the datadog debug output, many of the graphql.execute spans ended up with a parent id of 0).

**Versions**

`graphql` version: 1.12.5
`rails` (or other framework): rails
other applicable versions (`graphql-batch`, etc)

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok).

```
class FigsCatalogSchema < GraphQL::Schema
  include ApolloFederation::Schema
  TRACING_SERVICE_NAME = ENV.fetch('DD_SERVICE', 'figs-catalog-misconfigured')

  mutation(Types::MutationType)
  query(Types::QueryType)

  use GraphQL::Dataloader
  use GraphQL::Tracing::DataDogTracing, analytics_enabled: true, service: "#{TRACING_SERVICE_NAME}-graphql"
  use ApolloFederation::Tracing

  def self.execute(query_str = nil, variables: {}, context: {}, **kwargs)
    # make sure we can see the variables passed to queries and mutations.
    # filter out sensitive data the same way we do for the logs
    f = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
    filtered_vars = variables.blank? ? variables : f.filter(variables)
    current_span = Datadog.tracer.active_span
    current_span.set_tags(query: query_str, variables: filtered_vars) if current_span

    super.tap do |result|
      h = result.to_h
      if h.dig('errors').present? || h.dig('data', 'errors').present?
        Rails.logger.warn("query: #{query_str}, variables: #{filtered_vars}, result: #{h}")
        current_span.set_tag('result', h) if current_span
      end
    end
  end
end
```

**Steps to reproduce**

Just add `use GraphQL::Dataloader` to a schema and make a query that will pull some data from the database.
Inspect the trace in Datadog.
When Dataloader is part of the schema, the database calls do not get reported properly under the graphql.execute call (some do but ones under the schema don't).

**Expected behavior**

![image (1)](https://user-images.githubusercontent.com/251189/109333416-4b200200-7814-11eb-8d59-e973139dab66.png)

**Actual behavior**

[...]

PR #3461, opened 2021-05-02
Copy parent thread local vars to dataloader fiber local vars

ianks:
By default, Fibers have their own storage for `Thread#[]`, which is entirely separate from the parent `Thread#[]` storage. Many gems (such as RequestStore, MiniRacer, etc.) rely on thread local vars to keep track of execution context. Since Dataloader relies on `Fiber.new`, gems that rely on thread local storage cannot keep track of what's happening within the GraphQL execution context, causing unexpected behavior for users.

This PR implements the solution from https://github.com/rmosolgo/graphql-ruby/issues/3449#issuecomment-830409933 to copy thread locals vars to fiber local vars when using Dataloader.

see #3366
see #3449

### DIFF
--- a/guides/dataloader/overview.md
+++ b/guides/dataloader/overview.md
@@ -33,8 +33,6 @@ At a high level, `GraphQL::Dataloader`'s usage of `Fiber` looks like this:
 - `GraphQL::Dataloader` takes the first paused Fiber and resumes it, causing the `GraphQL::Dataloader::Source` to execute its `#fetch(...)` call. That Fiber continues execution as far as it can.
 - Likewise, paused Fibers are resumed, causing GraphQL execution to continue, until all paused Fibers are evaluated completely.
 
-Since _all_ GraphQL execution is inside a Fiber, __`Thread.current[...]` won't be set__. Those assignments are _Fiber-local_, so each new Fiber has an _empty_ `Thread.current`. For `GraphQL::Dataloader`, use GraphQL-Ruby's `context` object to provide "current" values to a GraphQL query. Inside `#fetch(...)` methods, those values can be re-assigned to `Thread.current` if application code requires it.
-
 ## Getting Started
 
 To install {{ "GraphQL::Dataloader" | api_doc }}, add it to your schema with `use ...`, for example:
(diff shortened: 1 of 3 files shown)


ianks:
I've confirmed this fixes the issue for MiniRacer and RequestStore, and I imagine the other gems (datadog, etc.) would work properly as well.

guides/dataloader/overview.md
```diff
@@ -33,8 +33,6 @@ At a high level, `GraphQL::Dataloader`'s usage of `Fiber` looks like this:
 - `GraphQL::Dataloader` takes the first paused Fiber and resumes it, causing the `GraphQL::Dataloader::Source` to execute its `#fetch(...)` call. That Fiber continues execution as far as it can.
 - Likewise, paused Fibers are resumed, causing GraphQL execution to continue, until all paused Fibers are evaluated completely.
```
rmosolgo, line 35: I would like to _describe_ the feature, so folks are are aware of it in case it complicates things. How about: 

```suggestion

Whenever `GraphQL::Dataloader` creates a new `Fiber`, it copies each pair from `Thread.current[...]` and reassigns them inside the new `Fiber`.

```

rmosolgo (commented):
LGTM, I just want to keep a note of it in the docs :+1:

ianks:
Done! @rmosolgo

rmosolgo:
Thanks for taking care of this improvement! I'll ship 1.12.9 soon, in the meantime, anyone who wants to try this out can use: 

```ruby
gem "graphql", github: "rmosolgo/graphql-ruby"
```

in their gemfile.

eregon:
Just to clarify, `Thread#[]` is *not* thread-local storage, it's fiber-local storage.
`Thread#thread_variable_{get,set}` are thread-local storage.
I think the issue might be that other gems like MiniRacer/RequestStore/etc should use thread-local storage and not fiber-local storage.

coding-chimp:
> I think the issue might be that other gems like MiniRacer/RequestStore/etc should use thread-local storage and not fiber-local storage.

There are fiber-based servers like Falcon out there, so using thread- and fiber-local variables probably makes sense if you want to maximize compatibility with ruby servers. I don't think there is a one-size-fits-all solution out there, unfortunately.

zvkemp:
In addition to the issue with Falcon that @coding-chimp raises, I would also point out that third-party libraries may intentionally choose to use `Thread#[]` over `Thread.current` _because_ of its fiber-local characteristics. E.g., Skylight assumes its spans behave stackwise, and because our trace is fiber-local, it ignores events generated in other fibers. While this may result in us not collecting quite all the information that is available, the potential downsides of the alternative is worse, i.e. under the current design, spans closed out-of-order require us to discard the entire trace.

I haven't read anything in Dataloader that leads me to believe this will actually happen with this implementation (it's unlikely to `yield` across span boundaries), but it's relatively straightforward to contrive a scenario in which moving Skylight's trace to a thread-local variable does exactly this, so I would caution against assuming that this strategy will be compatible with other gems or application code.

rmosolgo:
Thanks for the heads up about the downsides of this approach. I'll keep an ear out for any bug reports about it!

Issue #3310, opened 2021-02-02
Error executing subscription, NoMethodError - undefined method `unpack


MathyouMB:
I am trying to get GraphQL subscriptions working between a Rails API only app and a create react app using the Apollo useSubscription hook. It is my first time implementing this, so it is entirely possible I did something wrong. Unfortunately, I haven't been able to find anyone with this specific issue already. This is my first time creating a github issue, so If there is any further information I can provide please let me know.

After the client has connected and has reached the ```execute``` part of the GraphqlChannel, it throws an error when running GRAPHQLSCHEMANAMEHERE.execute. 

Chrome Network tab: The client just pings over and over and never gets a response from the server because of the crash.
<img src="https://i.imgur.com/o3b5dir.png">

GraphqlChannel.rb:
```ruby
def execute(data)
    query = data["query"]
    variables = ensure_hash(data["variables"])
    operation_name = data["operationName"]
    context = {
      channel: self,
    }

    # ERROR OCCURS HERE
    result = BoardgameapiSchema.execute({
      query: query,
      context: context,
      variables: variables,
      operation_name: operation_name
    })
    ...
```
 

The error thrown is "undefined method `unpack'".

```
Could not execute command from ({"command"=>"message", "identifier"=>"{\"channel\":\"GraphqlChannel\",\"channelId\":\"17760ba95f9\"}", "data"=>"{\"query\":\"subscription {\\n  test {\\n    id\\n    __typename\\n  }\\n}\\n\",\"variables\":{},\"action\":\"execute\"}"}) [NoMethodError - undefined method `unpack' for {:query=>"subscription {\n  test {\n    id\n    __typename\n  }\n}\n", 
```

The error can be traced back to
```gems/ruby-3.0.0/gems/graphql-1.12.0/lib/graphql/language/lexer.rb```

```ruby
def self.run_lexer(query_string)
        # ERROR OCCURS ON THE LINE DIRECTLY BELOW
	data = query_string.unpack(PACK_DIRECTIVE)
	eof = data.length
	...
```

**Versions**

`ruby` version: 3.0.0
`graphql` version: 1.12.0
`rails` version: 6.1.1

**GraphQL schema**

```ruby
# SubscriptionType
module Types
  class SubscriptionType < Types::BaseObject
    field :test, Types::GameType, null: false, description: 'A game'

    def test; end
  end
end

# Type
module Types
  class GameType < Types::BaseObject
    description 'A type that represents a single Game entity.'
[...]

PR #3464, opened 2021-05-05
[Documentation] Update action cable subscription setup docs

MathyouMB:
Ruby's behavior around keyword arguments changed in Ruby 3.0. This change updates the action cable subscriptions example to account for those changes.

This was previously found in #3310


rmosolgo:
Thanks for the update!

PR #3465, opened 2021-05-06
Add use GraphQL::Execution::Errors to the error handling documentation

fabioxgn:
I was trying to use the `rescue_from` method and it didn't work. I looked at the PR that introduce this, and you need to add the `GraphQL::Execution::Errors` to the schema. I think it would be nice to have it in the documentation's example.

### DIFF
--- a/guides/errors/error_handling.md
+++ b/guides/errors/error_handling.md
@@ -19,6 +19,7 @@ Handlers are added with `rescue_from` configurations in the schema:
 ```ruby
 class MySchema < GraphQL::Schema
   # ...
+  use GraphQL::Execution::Errors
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! Sorry about the trouble. 

`use GraphQL::Execution::Errors` is not required in GraphQL-Ruby 1.12+, where it's available by default: 

https://github.com/rmosolgo/graphql-ruby/blob/e58e649241452ba673e90cf4c2062c4e970df0b4/lib/graphql/schema.rb#L1548-L1551

I'd rather not add `use ...` to the main example, but if you want to add a note about _old_ versions, I'd accept that!

fabioxgn:
@rmosolgo sorry about that, I'm running version 1.10 that's why it was necessary. If it is available by default I think that we don't need to add it in the docs. Thanks!

PR #3469, opened 2021-05-11
ActionCableLink accepts connectionParams as func where operation is first argument

### LINKED ISSUE #3468: Dynamic connection params in ActionCableLink.ts

Due to technical limitations and cross origin requests, we cannot authenticate the ActionCable connection (no option to pass additional headers, cookies are not available) and have to authenticate individual channels. 

https://github.com/rmosolgo/graphql-ruby/blob/master/javascript_client/src/subscriptions/ActionCableLink.ts#L31

If `this.connectionParams` supported function instead of object, it would be easier to pass additional auth token. Unfortunately, it cannot be provided during initialization because the auth token is also stored inside Apollo context. 

```diff
-      }, this.connectionParams), {
+      }, (typeof this.connectionParams === "function") ? this.connectionParams(operation) : this.connectionParams), {
```

I've started by "packaging" ActionCableLink inside another Apollo Link but it becomes overly complicated just for such a simple thing. I've also tried creating an additional auth subscription that would use a separate channel and set auth on the connection but same, very complex with a lot of edge cases to handle.

**Additional context**

If there's another/different way of authentication, please share. Cookies/headers are not an option in our case. 

I'm happy to submit PR with the change and test if this is acceptable proposal.

arathunku:
I've applied my suggestion from https://github.com/rmosolgo/graphql-ruby/issues/3468 and filled in specs related to the change.

Few additional questions:

  1. There's nothing in guides about connectionParams - should I fill them in on this MR or is it OK if it stays undocumented?
  2. I see there's `prettier` in package.json but it looks unconfigured(`prettierrc.json` is missing)

Resolves #3468

### DIFF
--- a/javascript_client/src/subscriptions/__tests__/ActionCableLinkTest.ts
+++ b/javascript_client/src/subscriptions/__tests__/ActionCableLinkTest.ts
@@ -7,7 +7,6 @@ describe("ActionCableLink", () => {
   var log: any[]
   var cable: any
   var options: any
-  var link: any
   var query: any
   var operation: Operation
 
@@ -15,7 +14,9 @@ describe("ActionCableLink", () => {
     log = []
     cable = {
       subscriptions: {
-        create: function(_channelName: string, options: {connected: Function, received: Function}) {
+        create: function(channelName: string | object, options: {connected: Function, received: Function}) {
+          var channel = channelName
+          var params = typeof channel === "object" ? channel : { channel }
           var alreadyConnected = false
           var subscription = Object.assign(
             Object.create({
@@ -26,6 +27,7 @@ describe("ActionCableLink", () => {
                 log.push(["unsubscribe"])
               }
             }),
+            { params },
             options
           )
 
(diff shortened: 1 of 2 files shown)


javascript_client/src/subscriptions/__tests__/ActionCableLinkTest.ts
```diff
@@ -7,15 +7,16 @@ describe("ActionCableLink", () => {
...
   var operation: Operation
 
   beforeEach(() => {
     log = []
     cable = {
       subscriptions: {
-        create: function(_channelName: string, options: {connected: Function, received: Function}) {
+        create: function(channelName: string | object, options: {connected: Function, received: Function}) {
+          var channel = channelName
+          var params = typeof channel === "object" ? channel : { channel }
```
arathunku, line 19: Copied from https://github.com/rails/rails/blob/2af458a10d336ec15a5734dba27709c2f8eb745b/actioncable/app/javascript/action_cable/subscriptions.js#L21 :shrug:

rmosolgo:
Thanks for this improvement! Prettier isn't enforced in the codebase, thanks for asking. As far as docs, either way is ok with me. I'll merge this as-is, and if you want to, you're welcome to update the docs in another PR.

rmosolgo:
🚢  graphql-ruby-client 1.9.4!

arathunku:
@rmosolgo this is amazing! Thank you very much for the swift release!

PR #3472, opened 2021-05-11
Add failing test for directives with input cycles

eapache:
It was noted in a comment in build_from_definition.rb that this might be
a problem. It is a problem (this test currently fails with
SystemStackError).

@rmosolgo this is incomplete right now given it's just a test and no actual fix, but I ran into this as part of digging around in https://github.com/rmosolgo/graphql-ruby/pull/3448.

### DIFF
--- a/spec/graphql/schema/build_from_definition_spec.rb
+++ b/spec/graphql/schema/build_from_definition_spec.rb
@@ -122,6 +122,32 @@ type Word {
       assert_schema_and_compare_output(schema.chop)
     end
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
I gave it a shot anyways, here's a proposed hack: https://github.com/rmosolgo/graphql-ruby/commit/9e4eea99c3344c4c3a3cbcbefc224b4ebe07d902 

Feel free to grab it here if it helps (or if you'd rather, I can merge that branch instead of this one -- I just couldn't push that commit here).

Issue #2670, opened 2020-01-14
Directive for query


23tux:
I hope this is no dumb question (maybe I misunderstood the use case of directives):

In my API a user can have multiple roles like admin or manager. Depending on this roles, I'm using CanCan's `accessible_by` to scope results to the users abilities.

Now the use case popped up, that I want results for a specific role, e.g. a manager. My first solution was to include an argument which I can use to create the right ability for this role

```
query {
  articles(role: "manager") { title }
}
```

The problem is, when I query multiple resources, I have to include the role argument multiple times

```
query {
  articles(role: "manager") { title }
  comments(role: "manager") { id }
}
```

My idea was to use a custom directive for this, that is defined for the whole query

```
query @ability(role: "manager") {
  articles { title }
  comments { id }
}
```

```
class Ability < Schema::Directive
  locations(GraphQL::Schema::Directive::QUERY)

  argument :role, String, required: true

  def self.resolve(object, arguments, context)
    # ...
  end
end
```

But this does not seem to work:

1. The `resolve` method is never called, only when I use `locations(GraphQL::Schema::Directive::FIELD)`
2. From the arguments of the `resolve` method I guess that it is too late to scope the result, as the objects are already loaded, am I right?
3. Is a directive the right tool for such an use case? Or are there any other ways to manipulate a context (e.g. `context.fetch(:current_ability)`) for a whole query?

rmosolgo:
Hi! Good question. I'm afraid `Directive::QUERY` is not implemented yet, because I haven't had a use for it. But I think it would be a nice improvement for it to work like you expect! 

> Is a directive the right tool?

I'm not sure... Is `role` _supposed_ to match the role of the current user? If so, then it should probably be read from the current user instead, otherwise a malicious client could inject any value they want for that directive. 


> other ways manipulate a context

You can write to context at any time; for a query-level hook, you could try query instrumentation: https://graphql-ruby.org/queries/instrumentation.html

In fact, you could even implement your directive this way _now_: during a `before_query` hook, you could inspect the AST to determine whether the directive is present on `query.selected_operation`, and if it is, you could read the argument value and write it into context. For example: 

```ruby 
def before_query(q) 
  ast_operation = q.selected_operation 
  ast_operation.directives.each do |ast_directive| 
    if ast_directive.name == "ability" 
      ast_role_argument = ast_directive.arguments.find { |a| a.name == "role" }
      if ast_role_argument
        role_value = ast_role_argument.value 
[...]

PR #3474, opened 2021-05-12
Add support of query location for directive

lanadz:
# Context
I was working on change which would benefit from having a directive working on a `query` level.
Turns out `graphql-ruby` doesn't support directives on this level.
This PRs fixes this issue.

# How do we make this change
We are able to reuse the code created for directives which use `resolve` method to alter query execution. Best example would be `Transform` directive. It uses `resolve` method and can be attached to field.
Following it as an example I was able to wrap the entry point of query interpretation (`GraphQL::Execution::Interpreter::Runtime#run_eager`) and wrap it in `resolve_with_directives` method.

`resolve_with_directives` method doesn't have any idea of what type of `node` it is working on, so it works on `field` level for `Transform` and can work on the whole query level as well.
It also supports having more than one directive attached to a query and will correctly execute all of them.

# Example usage
```Ruby
class QueryDirectiveSchema < GraphQL::Schema
   class InitInt < GraphQL::Schema::Directive
      locations(GraphQL::Schema::Directive::QUERY)
      argument(:val, Integer, "Initial integer value.", required: true)

      def self.resolve(obj, args, ctx)
         ctx[:int] = args[:val]
         yield
       end
    end

      field :int, Integer, null: false

      def int
        context[:int] ||= 0
        context[:int] += 1
      end
    end

    directive(InitInt)
    query(Query)
  end
```

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -52,22 +52,25 @@ module GraphQL
           set_all_interpreter_context(query.root_value, nil, nil, path)
           object_proxy = authorized_new(root_type, query.root_value, context)
           object_proxy = schema.sync_lazy(object_proxy)
+
           if object_proxy.nil?
             # Root .authorized? returned false.
             write_in_response(path, nil)
           else
-            gathered_selections = gather_selections(object_proxy, root_type, root_operation.selections)
-            # Make the first fiber which will begin execution
-            @dataloader.append_job {
-              evaluate_selections(
-                path,
-                context.scoped_context,
-                object_proxy,
-                root_type,
-                root_op_type == "mutation",
-                gathered_selections,
-              )
-            }
+            resolve_with_directives(object_proxy, root_operation) do # execute query level directives
+              gathered_selections = gather_selections(object_proxy, root_type, root_operation.selections)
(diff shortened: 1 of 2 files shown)


lanadz:
@eapache could you take a look please?

eapache:
I'm actually not as familiar with the guts of the interpreter, but the explanation and test case here both make a lot of sense to me. @rmosolgo and maybe @swalkinshaw will be able to give better feedback.

rmosolgo:
Wow, I'm amazed that this Just Works :tm:. I agree with your design here, and I think the place you put it should work fine. 

I went to double-check that `QUERY`/`MUTATION`/`SUBSCRIPTION` would work, and it seems like the static validation should do fine there: 

https://github.com/rmosolgo/graphql-ruby/blob/82d6dd71094db46127d343ae2a60aeced0aab82e/lib/graphql/static_validation/rules/directives_are_in_valid_locations.rb#L36-L38

Thanks for contributing this improvement, what a win!

lanadz:
Thanks @rmosolgo, that helps our project too, simplifies the approach we took first as suggested here https://github.com/rmosolgo/graphql-ruby/issues/2670! 
Yeah, I completely agree on Just Works ™️ 😄 Great gem's design!

PR #3475, opened 2021-05-12
Fix a typo in Field Resolution document

JuanitoFatas:
https://graphql-ruby.org/fields/introduction.html

CleanShot 2021-05-12 at 23 18 41@2x

`[User!]!` should return array with 1 or more `User`s.

### DIFF
--- a/guides/fields/introduction.md
+++ b/guides/fields/introduction.md
@@ -81,7 +81,7 @@ Fields with a `deprecation_reason:` will appear as "deprecated" in GraphiQL.
 
 ### Field Resolution
 
-In general, fields return Ruby values corresponding to their GraphQL return types. For example, a field with the return type `String` should return a Ruby string, and a field with the return type `[User!]!` should return a Ruby array with zero or more `User` objects in it.
+In general, fields return Ruby values corresponding to their GraphQL return types. For example, a field with the return type `String` should return a Ruby string, and a field with the return type `[User!]!` should return a Ruby array with one or more `User` objects in it.
 
 By default, fields return values by:
 


rmosolgo:
Hmm, it's not a _typo_, my reading suggests that an empty list _is_ valid for `[T!]!`. From graphql.org: 

> The Non-Null and List modifiers can be combined. For example, you can have a List of Non-Null Strings:
> 
> ```graphql
> myField: [String!]
> ```
> 
> This means that the list itself can be null, but it can't have any null members. For example, in JSON:
> 
> ```graphql
> myField: null // valid
> myField: [] // valid
> myField: ['a', 'b'] // valid
> myField: ['a', null, 'b'] // error
> ```
> 
> Now, let's say we defined a Non-Null List of Strings:
> 
> ```graphql
> myField: [String]!
> ```
> 
> This means that the list itself cannot be null, but it can contain null values:
> 
> ```graphql
> myField: null // error
> myField: [] // valid
> myField: ['a', 'b'] // valid
> myField: ['a', null, 'b'] // valid
> ```
> 
> You can arbitrarily nest any number of Non-Null and List modifiers, according to your needs.

https://graphql.org/learn/schema/#lists-and-non-null

Specifically, `[] // valid` suggests to me that an empty list is valid for non-null list types. 

The _spec_ is a little bit more ambiguous: 

> The List and Non-Null wrapping types can compose, representing more complex types. The rules for result coercion and input coercion of Lists and Non-Null types apply in a recursive fashion.
> 
> For example if the inner item type of a List is Non-Null (e.g. `[T!]`), then that List may not contain any **null** items. However if the inner type of a Non-Null is a List (e.g. `[T]!`), then **null** is not accepted however an empty list is accepted.
> 

http://spec.graphql.org/draft/#sec-Combining-List-and-Non-Null

However, I don't see anything there that suggests `[]` is invalid for non-null lists. Do you?  

If so, I bet GraphQL-Ruby will need an update:

```ruby 
require "graphql"
t = GraphQL::Schema::NonNull.new(GraphQL::Schema::List.new(GraphQL::Schema::NonNull.new(GraphQL::Types::String)
))
t.validate_input([], {})
# => #<GraphQL::Query::InputValidationResult:0x00007fb6201c2d38 @valid=true, @problems=nil>
_.valid? 
# => true
```

JuanitoFatas:
Ahh that makes sense, thanks for taking the time for detailed explanations. Awesome 🙌🏼

PR #3477, opened 2021-05-13
Simplify arg coercion

eapache:
Minor refactor, no logical change. Just something I noticed while poking around argument validation still.

Get rid of has_value by moving the "no value" case to an else in the
original if with an early return.

Diff best viewed with whitespace suppressed.

@rmosolgo this one is actually mergeable :P


rmosolgo:
🤦 derp, I thought I merged this right after you opened it. Thanks!

PR #3482, opened 2021-05-17
Evented Dataloader via Fiber.scheduler

rmosolgo:
~~Trying to leverage `Fiber.scheduler` in Ruby 3, but it's currently segfaulting.~~ It's working! I reported my problem here: https://bugs.ruby-lang.org/issues/17875

I'm using the "dummy" scheduler from the Ruby tests, ~~but looking at https://github.com/socketry/async/pull/111, there might be a non-toy implementation here: https://github.com/socketry/event/~~ It seems like that scheduler is somehow coupled to the `Async do ... end` framework, which I'm not using here. Instead, I'm testing with https://github.com/digital-fabric/libev_scheduler

TODO: 

- [x] Investigate a real scheduler (maybe from `event` above) 
  - [x]   Test with LibevScheduler and Evt::Scheduler
- [x] DRY & rename AsyncDataloader 
- [x] Docs
- [x] Make this branch into `1.13-dev`
- [x] Test per-query `context[:dataloader]`
- [x] Why does rails :boom: on this branch?


rmosolgo:
It seems like `libev_scheduler` isn't working right on GitHub actions. I've tried and tried to get to the bottom of it, but I can't figure it out. I'm going to skip it on CI and ... revisit ... later ...

PR #3488, opened 2021-05-20
Fix introspection visibility for directives

aaackerman:
https://github.com/rmosolgo/graphql-ruby/pull/3450 updated directives to respect visibility; however, it isn't respected in introspection. This PR fixes visibility for directives in introspection.

**Expected**
Directives returned during introspective respect visibility. 

**Actual**
Directives do not respect visibility and are always returned when defined. 

cc @eapache


rmosolgo (approved):
Thanks for this fix!

PR #3491, opened 2021-05-21
Update class_based_api.md with finding legacy members

danielvdao:
Hi there! I was wondering if this would be helpful with people trying to track down legacy files for large graphql-ruby repos. I recently wrote this cop and used it to find violating files. Unfortunately I didn't write an autocorrect for it, but thought it could be a good use case here: https://github.com/DmitryTsepelev/rubocop-graphql/pull/39


rmosolgo:
Thanks for the improvement!

Issue #3490, opened 2021-05-21
Benchmarking / Performance issue for Legacy vs. Class Based API


danielvdao:
**Describe the bug**

I'm having trouble replicating the proper benchmarks / performance gains seen in the 1.9.21 (and latest version) of class-based API of graphql-ruby in reference to https://github.com/rmosolgo/graphql-ruby/issues/861. I'm trying to isolate the behavior in benchmarking code and would love some eyes.

Here is an attached gist with a few things: https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd

Legacy based:
- [Code used to run benchmark for legacy](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-legacy_based_benchmark-rb)
- [Benchmarks for legacy based API on 1.9.21](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-legacy-based-api-benchmarks)
- [Memory Profiling done for legacy on 1.9.21](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-legacy_based_graphql_1-9-21)
- [Memory Profiling done for legacy on 1.6.6
](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-legacy_based_memory_profile_1-6-6-txt)

Class based:
- [Code used to run benchmark for class based](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-class_based_api_benchmark-rb)
- [Class based API benchmarks](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-class-based-api-benchmarks)
- [Memory profiling done for class based on 1.9.21](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-class_based_memory_profile-txt)
- [Ruby profiling code for execution](https://gist.github.com/danielvdao/b20a3cae4fa726728d6abd138fde3cfd#file-ruby-profile-class-based)

It looks like memory profile is showing that the new interpreter is being used. It also shows that memory allocation is a lot less too (which seems to be expected) as `FieldResolutionContext` is 🔪  in the new runtime interpreter vs. legacy based behavior on 1.6.6.

[...]

PR #3494, opened 2021-05-21
Build up the response while executing the query

rmosolgo:
There's a lot of recursion in the interpreter where it's digging back into the response to write values and check for null propagations. 

I _think_ this could be improved by passing the result object along the way, while the query is running. It makes the code more complex, but it speeds up the benchmark in #3490 by about ~~10%~~ 30%. 

I think this could be improved further, by doing a better job keeping runtime metadata separate from the accumulated response.  

Also, the changes in #3364 help a bit, so that'd improve those benchmarks, too. 

TODO: 

- [x] Make these accessors protected or something
- [x] Other todos in code


eapache:
This looks like a big step in the right direction.

> I think this could be improved further, by doing a better job keeping runtime metadata separate from the accumulated response.

I would suggest using a custom subclass of `Hash` instead of a plain `{}`. That should let you keep the non-result metadata separately, and completely avoid rewalking the result to build up a "clean" version for serialization.

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -8,6 +8,17 @@ class Interpreter
       #
       # @api private
       class Runtime
+
+        class GraphQLResultHash < Hash
+          def graphql_metadata
```
eapache, line 13: I was thinking separate accessor methods rather than putting an extra hash in your hash :P

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -17,28 +28,81 @@ class Runtime
...
+          end
+          clean_result(response)
         end
 
         def final_value
-          @response.final_value
+          clean_result(@result)
+        end
+
+        def clean_result(r)
```
eapache, line 80: I think you could get rid of this code entirely by chopping off the response immediately when an invalid `nil` propagates during execution? You should already be at the right spot in the response object since you do set `dead` there if I understand correctly.

Maybe that would involve tracking a `parent` pointer in `GraphQLResultHash`? But that doesn't seem so bad.

eapache:
> Ew, yuck: subclass hash

😂  you could also just have a plain object with an attribute that is a hash if you want. Accessing it takes a few more characters of typing but the behaviour works out about the same.

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: :+1: Agreed, that'd be less error prone and just as fast.

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: Yeah, we'd need a parent reference to do the propagation while the query is running. (Interestingly, this GraphQLResultHash is starting to look _just like_ the old `GraphQL::Query::FieldResolutionContext`! But with less overhead since it's also the result hash -- leaf fields don't create an instance of it, only objects.)

rmosolgo: Took a shot at it in https://github.com/rmosolgo/graphql-ruby/pull/3494/commits/96851cf73fa7ccb0dd4021d0d3f90fcec81b0d25

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: Done in https://github.com/rmosolgo/graphql-ruby/pull/3494/commits/96851cf73fa7ccb0dd4021d0d3f90fcec81b0d25 ... and now there's a subclass of Array, too 🙈 !

rmosolgo:
The thing is, now, what if we _leaned into_ these objects? 

The arguments to each method on `Interpreter::Runtime` are ... a lot. Because there's a lot of context at each point in execution. But now, with these "rich" response objects, we could tack runtime info onto them and just pass them through. (For example, `.graphql_object`, `.graphql_path`, maybe `.graphql_ast_node`.)

rmosolgo:
Amazingly this is looking like a 30% speedup locally, using the benchmark from #3490 : 

```diff 
Calculating -------------------------------------
  Class Based API: Querying for 10 objects - all fields  
-                         293.699  (± 4.4%) i/s -      1.485k in   5.068077s
+                         383.014  (± 8.9%) i/s -      1.900k in   5.018178s
  Class Based API: Querying for 100 objects - all fields 
-                          33.794  (± 5.9%) i/s -    171.000  in   5.070778s
+                          45.826  (± 8.7%) i/s -    232.000  in   5.109289s
  Class Based API: Querying for 1000 objects - all fields
-                           3.364  (± 0.0%) i/s -     17.000  in   5.057376s
+                           4.591  (± 0.0%) i/s -     24.000  in   5.266070s
  Class Based API: Querying for 10000 objects - all fields
-                           0.309  (± 0.0%) i/s -      2.000  in   6.473207s
+                           0.460  (± 0.0%) i/s -      3.000  in   6.527539s
```

rmosolgo:
I grabbed a few other changes from #3364 to include in this branch, and now the benchmark is a hair better: 

```
Class Based API: Querying for 10 objects - all fields  
                        41.000  i/100ms
Class Based API: Querying for 100 objects - all fields 
                         5.000  i/100ms
Class Based API: Querying for 1000 objects - all fields
                         1.000  i/100ms
Class Based API: Querying for 10000 objects - all fields
                         1.000  i/100ms
Class Based API: Querying for 10 objects - all fields  
                        421.826  (± 2.4%) i/s -      2.132k in   5.056882s
Class Based API: Querying for 100 objects - all fields 
                         47.892  (± 6.3%) i/s -    240.000  in   5.033884s
Class Based API: Querying for 1000 objects - all fields
                          4.838  (± 0.0%) i/s -     25.000  in   5.188949s
Class Based API: Querying for 10000 objects - all fields
                          0.464  (± 0.0%) i/s -      3.000  in   6.478701s

Comparison:
Class Based API: Querying for 10 objects - all fields  :      421.8 i/s
Class Based API: Querying for 100 objects - all fields :       47.9 i/s - 8.81x  (± 0.00) slower
Class Based API: Querying for 1000 objects - all fields:        4.8 i/s - 87.20x  (± 0.00) slower
Class Based API: Querying for 10000 objects - all fields:        0.5 i/s - 909.24x  (± 0.00) slower
```

rmosolgo:
Apparently benchmark/ips shipped a bad version. A fix is on the way (https://github.com/evanphx/benchmark-ips/pull/115) but I'm just going to ignore those failures for now.

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -304,43 +321,87 @@ def evaluate_selection_with_args(kwarg_arguments, field_defn, next_path, ast_nod
...
+            elsif value.is_a?(GraphQL::UnauthorizedError)
+              # this hook might raise & crash, or it might return
+              # a replacement value
+              next_value = begin
+                schema.unauthorized_object(value)
+              rescue GraphQL::ExecutionError => err
+                err
+              end
+              continue_value(path, next_value, parent_type, field, is_non_null, ast_node, result_name, selection_result)
+              HALT
```
rmosolgo, line 379: Oops, this is a bug. Previously it didn't return `HALT`, it returned `continue_value(...)`. I should remove this `HALT`.

PR #3496, opened 2021-05-24
Extract logic for schema setup

rmosolgo:
These methods were just tacked on to `Schema`, but I'm interested in adding a step or two in order to validate default argument values, as described in #3448. 

All I've done here is copy-pasted, so the code isn't any _better_, but at least these methods have a bit of isolation now. 

TODO: 

- [ ] Add a step for validating and default values on arguments that were added

(This is just like `Schema::Traversal` from the old dsl)

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require "graphql/schema/addition"
 require "graphql/schema/base_64_encoder"
 require "graphql/schema/catchall_middleware"
 require "graphql/schema/default_parse_error"
@@ -1582,10 +1583,7 @@ module GraphQL
       # @param new_directive [Class]
       # @return void
       def directive(new_directive)
-        own_directives[new_directive.graphql_name] ||= begin
-          add_type_and_traverse(new_directive, root: false)
-          new_directive
-        end
+        add_type_and_traverse(new_directive, root: false)
       end
 
       def default_directives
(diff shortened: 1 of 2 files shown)


rmosolgo:
👋 @eapache , what do you think of this approach? TL;DR: While adding types to the schema, I made it gather a list of arguments with default values. Then, after all types are added (including after resolving lazy types), I made it go through the list and call `#validate_default_value` on them. 

It seems to me like it should _work_ fine, but am I missing anything here?

rmosolgo:
Yep, it's the latter: basically, every method that adds some type info the the schema (`query(...)`,`mutation(...)`, `orphan_types(...)`, a couple others) would kick of a process of: 

- Traversal, finding every type that can be reached from the given starting point 
- Resolving late-bound types, then re-starting traversal from those resolved late-bound types 

I've added a third step: 

- Validate each argument that defined a default value  (Conceptually, that part of the change is just here: https://github.com/rmosolgo/graphql-ruby/pull/3496/commits/6a7efe6ecace9855aa0515d65de994c04bc31b71)

Yeah, that's basically a "chunk" of the schema, but because of the traversal, I'd _expect_ the chunk to have everything it needs to validate itself (ie, field's argument's types are traversed, input object argument types are traversed, and so on). In any case, I'll add a test for a deeply nested example.

I _think_ build-from-definition is OK because it uses those same entry points. If any late-bound types couldn't be resolved by the time it validates arguments, then it would've raised an error.

Anyhow, if no red flags jump out at you, I'll flesh out a few more tricky tests and see how it goes. Thanks for taking a look.

PR #3499, opened 2021-05-25
Runtime directives on fragments

rmosolgo:
Oops, runtime directives on fragment spreads and inline fragments were never actually implemented. 

It turns out to be pretty tricky because it requires these changes: 

- Gathering selections to run should be scoped by AST, _if_ the fragments at hand have runtime directives that affect execution (eg, not `@include(if: true)`) 
- Then, those selections should be run with some degree of isolation. TBD: how much isolation? Anyways, the whole point of `@defer` is to send the whole sub-tree after-the-fact, so it's got to be executed on its own.
- Then, (I guess _depending_?), the sub-tree should be merged back to the main result, so that the result can contain the whole query. (This doesn't apply to `@defer`, but one can imagine a runtime that _doesn't_ split the result tree.) 

Doing this for synchronous execution was easy enough, but once promises and dataloader get involved, 🤯 

TODO: 

- [x] DRY some code?  (I did my best, anyways.)

### DIFF
--- a/benchmark/run.rb
+++ b/benchmark/run.rb
@@ -3,7 +3,7 @@ TESTING_INTERPRETER = true
 require "graphql"
 require "jazz"
 require "benchmark/ips"
-require "ruby-prof"
+require "stackprof"
 require "memory_profiler"
 require "graphql/batch"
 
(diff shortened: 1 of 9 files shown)


rmosolgo:
Taking a quick look at `be rake bench:profile_large_result`: 

```diff 
  Querying for 1000 objects
-                           5.652  (± 0.0%) i/s -     29.000  in   5.131596s
+                           5.443  (± 0.0%) i/s -     28.000  in   5.144526s
```

~ 4% slower, that's ok

```diff 
- Total allocated: 12556544 bytes (73400 objects)
+ Total allocated: 12597264 bytes (74409 objects)
```


```diff 
  allocated memory by class
  -----------------------------------
-    6769040  Array
+    6809200  Array
-    3294864  Hash
+    2367032  Hash
      960184  GraphQL::Execution::Interpreter::Runtime::GraphQLResultHash
+     928168  GraphQL::Execution::Interpreter::Runtime::GraphQLSelectionSet
      802800  Proc
      682664  GraphQL::Execution::Interpreter::Runtime::GraphQLResultArray
       40000  GraphQLBenchmark::ProfileLargeResult::FooType
      2544  Fiber
```

Memory-wise: 

- 40k new Arrays -- probably the arrays returned from `gather_selections`? 
- 2x Hashes (counting `GraphQLSelectionSet`s 

I think these can be improved by skipping the array when there's only one selection set. I'm not sure where the _double_ Hash is coming from, but I'll track it down.

rmosolgo:
Knocked out most of those new arrays in 5e0a9d51f :+1: 

```diff 
  allocated objects by class
  -----------------------------------
-      43222  Array
+      43225  Array
```

PR #3505, opened 2021-05-27
Fix some dead marking and detection

### LINKED ISSUE #3504: Did something change about how errors are added to the response in #3494
cc @eapache , trying to replicate the change in behavior that you spotted. 

Here's a script that returns 1 error _before_ #3494 but 3 errors _after_ it. Gotta figure out why.

```ruby
require "bundler/inline"

gemfile do
  if ENV["BEFORE"]
    puts "using previous graphql-ruby ref"
    gem "graphql", github: "rmosolgo/graphql-ruby", ref: "b4a42cf8c012dee3d5866188b470b43b37c706be"
  elsif ENV["AFTER"]
    puts "using latest graphql-ruby ref"
    gem "graphql", github: "rmosolgo/graphql-ruby", ref: "master"
  else
    raise "Pass BEFORE=1 or AFTER=1"
  end
end

  def authorized?(*)
    raise GraphQL::ExecutionError, "#{name} is not authorized"
  end
end

class Thing < GraphQL::Schema::Object
  field_class BaseField
  field :title, String, null: false
end

  field :things, [Thing], null: false

  def things
    [{title: "a"}, {title: "b"}, {title: "c"}]
  end
end

  query Query
end

query = GraphQL::Query.new(Schema, "{ things { title } }")
res = query.result
puts "GraphQL: #{GraphQL::VERSION}"
pp res["errors"].size
pp res.to_h

```

rmosolgo:
Fixes #3504 

TODO: 

-  ~~Investigate marking the parent as dead. Is it in the spec?~~

    I didn't check the spec, but I'm going to keep this behavior because: 
    
      - It's how it used to work 
      - It makes it _look_ like the first error halts subsequent execution (but I think it actually _continues_ under the hood, but never updates the result, since it's dead)
- [x] Migrate the test from #3504 into a test

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -380,7 +380,6 @@ module GraphQL
                 value.ast_node ||= ast_node
                 context.errors << value
                 set_result(selection_result, result_name, nil)
-                selection_result.graphql_dead = true
               end
               HALT
             elsif value.is_a?(GraphQL::UnauthorizedError)
@@ -600,7 +599,7 @@ module GraphQL
             if eager
               lazy.value
             else
-              result[result_name] = lazy
+              set_result(result, result_name, lazy)
               lazy
             end
           else
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -12,7 +12,14 @@ module GraphQL
         module GraphQLResult
           # These methods are private concerns of GraphQL-Ruby,
           # they aren't guaranteed to continue working in the future.
-          attr_accessor :graphql_dead, :graphql_non_null, :graphql_parent, :graphql_result_name
+          attr_accessor :graphql_dead, :graphql_parent, :graphql_result_name
+          # Although these are used by only one of the Result classes,
+          # it's handy to have the methods implemented on both (even though they just return `nil`)
+          # because it makes it easy to check if anything is assigned.
+          # @return [nil, Array<String>]
+          attr_accessor :graphql_non_null_field_names
+          # @return [nil, true]
+          attr_accessor :graphql_non_null_list_items
         end
 
         class GraphQLResultHash < Hash
@@ -204,7 +211,7 @@ module GraphQL
           # the field's return type at this path in order
           # to propagate `null`
           if return_type.non_null?
-            (selections_result.graphql_non_null ||= []).push(result_name)
+            (selections_result.graphql_non_null_field_names ||= []).push(result_name)
           end
           # Set this before calling `run_with_directives`, so that the directive can have the latest path
           set_all_interpreter_context(nil, field_defn, nil, next_path)
@@ -336,8 +343,10 @@ module GraphQL
         def set_result(selection_result, result_name, value)
           if !dead_result?(selection_result)
             if value.nil? &&
-                (nn = selection_result.graphql_non_null) &&
-                (nn == true || nn.include?(result_name))
+                ( # there are two conditions under which `nil` is not allowed in the response:
+                  (selection_result.graphql_non_null_list_items) || # this value would be written into a list that doesn't allow nils
+                  ((nn = selection_result.graphql_non_null_field_names) && nn.include?(result_name)) # this value would be written into a field that doesn't allow nils
+                )
               # This is an invalid nil that should be propagated
               parent = selection_result.graphql_parent
               name_in_parent = selection_result.graphql_result_name
@@ -480,7 +489,7 @@ module GraphQL
           when "LIST"
             inner_type = current_type.of_type
             response_list = GraphQLResultArray.new
-            response_list.graphql_non_null = inner_type.non_null?
+            response_list.graphql_non_null_list_items = inner_type.non_null?
             response_list.graphql_parent = selection_result
             response_list.graphql_result_name = result_name
             set_result(selection_result, result_name, response_list)
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -348,6 +348,13 @@ module GraphQL
                   ((nn = selection_result.graphql_non_null_field_names) && nn.include?(result_name)) # this value would be written into a field that doesn't allow nils
                 )
               # This is an invalid nil that should be propagated
+              # One caller of this method passes a block,
+              # namely when application code returns a `nil` to GraphQL and it doesn't belong there.
+              # The other possibility for reaching here is when a field returns an ExecutionError, so we write
+              # `nil` to the response, not knowing whether it's an invalid `nil` or not.
+              # (And in that case, we don't have to call the schema's handler, since it's not a bug in the application.)
+              # TODO the code is trying to tell me something.
+              yield if block_given?
               parent = selection_result.graphql_parent
               name_in_parent = selection_result.graphql_result_name
               if parent.nil? # This is a top-level result hash
@@ -369,11 +376,10 @@ module GraphQL
           case value
           when nil
             if is_non_null
-              err = parent_type::InvalidNullError.new(parent_type, field, value)
-              if !dead_result?(selection_result)
+              set_result(selection_result, result_name, nil) do
+                # This block is called if `result_name` is not dead. (Maybe a previous invalid nil caused it be marked dead.)
+                err = parent_type::InvalidNullError.new(parent_type, field, value)
                 schema.type_error(err, context)
-                set_result(selection_result, result_name, nil)
-                selection_result.graphql_dead = true
               end
             else
               set_result(selection_result, result_name, nil)
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -321,8 +321,20 @@ module GraphQL
           end
         end
 
+        def dead_result?(selection_result)
+          r = selection_result
+          while r
+            if r.graphql_dead
+              return true
+            else
+              r = r.graphql_parent
+            end
+          end
+          false
+        end
+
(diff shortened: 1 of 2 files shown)


rmosolgo:
There’s still a _possible_ bug here. The difference is between using path to mark something as dead (which can identify leaves or objects/arrays) vs marking dead on the result (which is an object or list, but never a leaf). If that codepath ever marked objects as dead, using path, it won’t mark a result as dead (since I removed that line).

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -333,6 +345,9 @@ def set_result(selection_result, result_name, value)
                 @response = nil
               else
                 set_result(parent, name_in_parent, nil)
+                # This is odd, but it's how it used to work. Even if `parent` _would_ accept
+                # a `nil`, it's marked dead. TODO: check the spec, is there a reason for this?
+                parent.graphql_dead = true
```
eapache, line 350: Yeah this is weird to me too, I would expect it to just be `selection_result.graphql_dead = true` since it's the current field that actually needs marking as dead?

But I note there's something going on I don't grok with `selection_result.graphql_non_null` where it's either a boolean (whether the current field is non-null or not, makes sense, though maybe it ought to be a reference to the query AST type?) _or_ it can be a list of something for reasons that aren't clear to me. Maybe we're not always operating directly on `selection_result` as a single field or something?

eapache: Also what is the difference between this case in `set_result` and in `continue_value`, they both seem to handle part of the invalid-nil flow?

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -346,7 +361,7 @@ def continue_value(path, value, parent_type, field, is_non_null, ast_node, resul
           when nil
             if is_non_null
               err = parent_type::InvalidNullError.new(parent_type, field, value)
-              if !selection_result.graphql_dead
+              if !dead_result?(selection_result)
                 schema.type_error(err, context)
                 set_result(selection_result, result_name, nil)
```
eapache, line 366: this case (and maybe others?) know they're heading into the invalid-null flow of `set_result`, I wonder if they could call it directly, and avoid the overhead of the check in `set_result` in all the other flows?

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: Regarding `graphql_non_null`, it was basically used in two distinct modes: `true` indicated that the selection_result object was an Array which allowed no `nil`s; `Array<String>` indicated that the selection_result was a Hash which didn't allow `nil`s for the fields whose names were contained in the Array. 

It was straightforward to highlight this distinction by using two separate accessors, so I did so in a322abaac

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: I'm with you that the code is trying to tell us something... I tried to probe it a bit in 12fc019bf. What I found is: 

- There are basically two kinds of invalid `nil`s: 
    1. User code returned `nil` to GraphQL for a field that has `null: false`. We should warn the developer somehow (namely, the `schema.type_error` hook) because this is a bug in the system that should be fixed. 
    2. The field raise or returned a `GraphQL::ExecutionError`, so the runtime writes `nil` into the response. We assume that this error was raised on purpose and the developer doesn't need to be told, so we don't call `schema.type_error`. 
- Only the runtime code can distinguish between those, since it has the `field` definition. `set_result` has only the response data structures. (This could be modified, but that's how it is now.) That's why there's the `is_non_null` variable in `continue_value`. 
- `set_result` has to be prepared for receiving invalid nulls because error handlers don't bother tracking that. Maybe they _could_? With a method like `propagate_null`, `continue_value` could call that whenever it detects an invalid null. But even in that case, in order to propagate, we'd have to retain the result-level bookkeeping, in order to know whether a `parent` also has a non-null restriction. 

So I did a tiny hack that is _technically_ DRYer, and a hair faster, probably, since there's less work, but 🤷  not sure where to go next.

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: Btw, interesting consideration which I haven't explored: `graphql_non_null_fields` is currently an array, but for queries that include aliases, that array may contain duplicates. It might be that a Set is better for this. In practice I think it depends on size, so I'm not sure the best way to approach it.

rmosolgo: I take that back -- it uses `result_name`, which would be the alias if there is one. So it wouldn't ever contain duplicates. However, whenever the same selection (from the AST) is executed, it will build up duplicate arrays 🤔

rmosolgo:
I think there _are_ more improvements to be made here, but this is enough to get us back to compat, so I'm gonna roll with it.

eapache:
👍  how much of the previous perf win does this lose us?

eapache:
I was assuming _lots_, since `dead_result?` is recursive all over the place again, which is where I thought most of the previous wins had come from. Definitely curious what the benchmarks say.

rmosolgo:
I forgot about the recursion in `dead_result?`. Here's what I ran: 

```
git co 5d0239ee4e39264ab3f0cc0016d07e638eeb6eff
be rake bench:profile_large_result > before.txt
git co 12fc019bf4af28f83f411d5befc240a728362174
be rake bench:profile_large_result > after.txt
```

It is a good bit slower, ~10%: 

```diff
Querying for 1000 objects
                          1.000  i/100ms
 Calculating -------------------------------------
 Querying for 1000 objects
-                          5.613  (± 0.0%) i/s -     29.000  in   5.168206s
+                          5.014  (± 0.0%) i/s -     26.000  in   5.186973s
```

And because of the dead check:

```diff
 Measure Mode: wall_time
 Thread ID: 6200
 Fiber ID: 6220
-Total: 0.963930
+Total: 1.091016
 Sort by: self_time

  %self      total      self      wait     child     calls  name                           location
- 10.14      0.797     0.098     0.000     0.700    52002  *GraphQL::Execution::Interpreter::Runtime#after_lazy /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:562
- 10.12      0.153     0.098     0.000     0.055    73006   GraphQL::Execution::Interpreter::Runtime#set_all_interpreter_context /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:541
-  6.92      0.455     0.067     0.000     0.388    41001  *GraphQL::Execution::Interpreter::Runtime#continue_field /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:419
-  5.73      0.055     0.055     0.000     0.000   240014   GraphQL::Query::Context#[]=    /Users/rmosolgo/code/graphql-ruby/lib/graphql/query/context.rb:172
+  9.46      0.162     0.103     0.000     0.059    73006   GraphQL::Execution::Interpreter::Runtime#set_all_interpreter_context /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:570
+  9.43      0.909     0.103     0.000     0.806    52002  *GraphQL::Execution::Interpreter::Runtime#after_lazy /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:591
+  6.48      0.554     0.071     0.000     0.483    41001  *GraphQL::Execution::Interpreter::Runtime#continue_field /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:448
+  6.01      0.066     0.066     0.000     0.000    41001   GraphQL::Execution::Interpreter::Runtime#dead_result? /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:331
+  5.36      0.059     0.059     0.000     0.000   240014   GraphQL::Query::Context#[]=    /Users/rmosolgo/code/graphql-ruby/lib/graphql/query/context.rb:172
```

I wish I'd thought to check this before merging, I just forgot about that side of it 😖 .

eapache:
Those numbers aren't obviously comparable with the pre-https://github.com/rmosolgo/graphql-ruby/pull/3494 numbers for some reason (?) but I assume it's still an overall win?

Might be motivating to properly figure out the "just cut off the dead branch and let it run" approach so we could drop the recursion again?

rmosolgo:
Still a solid improvement over before the refactor: 

```
~/code/graphql-ruby % git co cd337dc89eb0ef0b82825b5d2879313ec2b9646c
Previous HEAD position was 3e5812653 Build up the response while executing the query
HEAD is now at cd337dc89 Merge pull request #3489 from alea12/patch-3
~/code/graphql-ruby % be rake bench:profile_large_result > before_refactor.txt
~/code/graphql-ruby % git diff --no-index before_refactor.txt after.txt
diff --git a/before_refactor.txt b/after.txt
index 581126632..bd59da087 100644
--- a/before_refactor.txt
+++ b/after.txt
@@ -3,281 +3,279 @@ Querying for 1000 objects
                          1.000  i/100ms
 Querying for 1000 objects
-                          3.865  (± 0.0%) i/s -     20.000  in   5.175685s
+                          5.014  (± 0.0%) i/s -     26.000  in   5.186973s
```

Which figures, since there were a couple other recursive methods (eg `write_in_response`) that were also refactored away.

Agreed, it'd be nice to figure out a solution that's just as reliable but doesn't require this recursion over-and-over. There are also contexts where `dead_result?(...)` is checked, then some custom logic is run, then it calls `set_result(...)`, which calls `dead_result?` on the same response hash. Those duplicate calls could be reduced. For example: 

https://github.com/rmosolgo/graphql-ruby/blob/a6487a0554624a235cb72f88478379270f6b7d43/lib/graphql/execution/interpreter/runtime.rb#L393-L397

Issue #3506, opened 2021-05-31
Problems resolving an enum with underscore when used as an argument to a query


thornomad:
**Describe the bug**

Banging my head on this behavior and not sure if my setup is wrong or indeed this is a bug.  When I pass an enum as an argument to a query I am getting an `UnresolvedValueError` for the enum if the enum has an underscore.

I think the [GraphQL schema supports enums with underscores](https://spec.graphql.org/June2018/#EnumValue) but this is the first time I've tried them as an argument.

#### simplified setup
```ruby
module Enums
  class PaymentStatusEnum < Types::BaseEnum
    value "PAID"
    value "UNPAID"
    value "PARTIALLY_PAID"
  end
end

module Resolvers
  class InvoicesResolver < Resolvers::BaseSearchObjectResolver
    type Types::InvoiceType.connection_type, null: false

    class InvoicesFilter < Types::BaseInputObject
      argument :payment_status, [Enums::PaymentStatusEnum], required: false
    end

    argument :filter, InvoicesFilter, required: true
  end
end
```

#### example query
```graphql
query Invoices($filter: InvoicesFilter!) {
  invoices(filter: $filter) {
    nodes {
      id
    }
  }
}
```

#### example args
```jsonb
{
	"filter": {
    "paymentStatus": "PARTIALLY_PAID"
	},
}
```

When I attempt to use `PARTIALLY_PAID` I get an exception of:

```text
Enums::InvoicePaymentStatusEnum::UnresolvedValueError: Can't resolve enum InvoicePaymentStatusEnum for \"PARTIALLY PAID\"
```

It is being raised inside of `#coerce_result` [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/enum.rb#L96-L105).

When I look at what's throwing the error I can see that the `value` being passed to `#coerce_result` is actually `"PARTIALLY PAID"` (missing the underscore).  This is throwing the error as it cannot be resolved.

I tried walking my way up the method calls but am getting lost.  Sigh.  Can't figure out where or what is stripping the underscore.  I think it's somewhere in the middle, though, because when I test it with a misspelled enum (e.g., `"PARTIALLY_PPPPPAID"`) I get a different error raised earlier saying it was expecting one of: `PAID, UNPAID, PARTIALLY_PAID`.  So it would seem at some point it's recognizing a valid enum (with underscore) then at another point that underscore is getting stripped off.

Appreciate any help you can give! Thanks.

[...]

PR #3507, opened 2021-06-01
Improve the message when a field returns an invalid enum value

rmosolgo:
To improve debugging situations like https://github.com/rmosolgo/graphql-ruby/issues/3506

### DIFF
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -36,7 +36,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "racc", "~> 1.4"
   s.add_development_dependency "rake", "~> 12"
   s.add_development_dependency "rubocop", "0.68" # for Ruby 2.2 enforcement
-  s.add_development_dependency "stackprof"
   # required for upgrader
   s.add_development_dependency "parser"
   # website stuff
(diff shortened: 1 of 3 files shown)


graphql.gemspec
```diff
@@ -36,7 +36,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "racc", "~> 1.4"
   s.add_development_dependency "rake", "~> 12"
   s.add_development_dependency "rubocop", "0.68" # for Ruby 2.2 enforcement
-  s.add_development_dependency "stackprof"
```
eregon, line 39: FWIW, one possibility might be to add it like `gem "stackprof", platforms: [:mri]` if you need it on MRI.
We'll also look into letting stackprof install, although it's rather unlikely stackprof would work at runtime on TruffleRuby without changes in stackprof: https://github.com/oracle/truffleruby/issues/2044

rmosolgo: Thanks for the suggestion -- I briefly considered a _proper_ fix, but then I remembered that when I used `ruby-prof` for profiling, it wasn't in the gemspec either 🤷 Nobody seemed to tripped up by it (it's only used by a couple rake tasks). But in any case, I'll give this a try if I turn out to need it -- thanks again!

PR #3513, opened 2021-06-03
Recursively set graphql_dead so that checking is faster

### LINKED ISSUE #3512: v1.12.11 causes "stack level too deep" with rspec
Upgrading to 1.12.11 broke one of our tests posting a graphql query with rspec. The query crashes with `stack level too deep` on ` ApiSchema.execute`.

What is strange is that it only breaks when running in rspec. The same request being called with the development server works perfectly. And the overflow is pretty since the same request on 1.12.10 and with `rails server` works with  `RUBY_THREAD_VM_STACK_SIZE=200000` (200kb) but on 1.12.11 crashes even with 10MB.

By bisecting I pinned down the problem down to that particular commit: https://github.com/rmosolgo/graphql-ruby/commit/d3e97e48f07ebb14ae00ce31c2c8a6f4d16025f8.

I imagine something in Rspec must interfere with the logic but despite trying to strip down my test from any other third-party gem, it still crashes. 

The test:
```rb
  it 'returns a 200 and the object' do
    object = create(:object)
    variables = { id: object.id }

    post graphql_path,
         params: { query: QUERY, variables: variables }
    expect(response.status).to eq(200)
  end
```

The GQL query is pretty big with 400 lines, 4 levels of nesting (complexity between O(1) and O(n) though), multiple fragments but nowhere big enough to fill the entire call stack.

Would love to hear your thoughts on what might be happening or how I can debug that further.

**Versions**

`graphql` version: 1.12.11 and 1.12.12
`rails`: 6.1.3
`rspec-rails`: 5.0.1
`graphql-batch`: 0.4.3

rmosolgo:
Fixes #3512 

cc @eapache  since we talked about improving this a bit. I feel like I'm still missing something -- why did it need the `.graphql_parent.graphql_dead` check? Gotta figure that out 👀

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -439,15 +439,10 @@ module GraphQL
         end
 
         def dead_result?(selection_result)
-          r = selection_result
-          while r
-            if r.graphql_dead
-              return true
-            else
-              r = r.graphql_parent
-            end
-          end
-          false
+          # When a result is marked dead, any of its existing children are marked dead
+          # and it accepts no more writes. But this result might have been initialized,
+          # but not yet assigned to a key in its parent. So check the parent too.
+          selection_result.graphql_dead || ((parent = selection_result.graphql_parent) && parent.graphql_dead)
         end
 
         def set_result(selection_result, result_name, value)
@@ -473,11 +468,26 @@ module GraphQL
                 set_result(parent, name_in_parent, nil)
                 # This is odd, but it's how it used to work. Even if `parent` _would_ accept
                 # a `nil`, it's marked dead. TODO: check the spec, is there a reason for this?
-                parent.graphql_dead = true
+                set_graphql_dead(parent)
               end
             else
               selection_result[result_name] = value
             end
+          end
+        end
+
+        # Mark this node and any already-registered children as dead,
+        # so that it accepts no more writes.
+        def set_graphql_dead(selection_result)
+          case selection_result
+          when Array
+            selection_result.graphql_dead = true
+            selection_result.each { |res| set_graphql_dead(res) }
+          when Hash
+            selection_result.graphql_dead = true
+            selection_result.each { |k, v| set_graphql_dead(v) }
+          else
+            # It's a scalar, no way to mark it dead.
           end
         end
 


rmosolgo:
Even this change gives another nice perf gain: 

```diff
 Calculating -------------------------------------
 Querying for 1000 objects
-                          4.926  (± 0.0%) i/s -     25.000  in   5.075798s
+                          5.301  (± 0.0%) i/s -     27.000  in   5.093446s
```

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -439,15 +439,10 @@ def evaluate_selection_with_args(kwarg_arguments, field_defn, next_path, ast_nod
...
-              return true
-            else
-              r = r.graphql_parent
-            end
-          end
-          false
+          # When a result is marked dead, any of its existing children are marked dead
+          # and it accepts no more writes. But this result might have been initialized,
+          # but not yet assigned to a key in its parent. So check the parent too.
+          selection_result.graphql_dead || ((parent = selection_result.graphql_parent) && parent.graphql_dead)
```
eapache, line 445: can the second half of the clause just use `&.` or do you support ruby versions without that still?

eapache (approved):
This change makes sense to me as-is (and the perf win is 🎉) but I also don't fully grok all the weird stuff around checking parents.

What cases start to fail if you just... don't do any stuff with the parents. Why do they fail?

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -473,14 +468,29 @@ def set_result(selection_result, result_name, value)
                 set_result(parent, name_in_parent, nil)
                 # This is odd, but it's how it used to work. Even if `parent` _would_ accept
                 # a `nil`, it's marked dead. TODO: check the spec, is there a reason for this?
```
eapache, line 470: This one just seems wrong to me. If the parent accepts a `nil` it shouldn't be marked dead, and if it doesn't then we should rely on the error bubbling up and marking it dead at that point.

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -439,15 +439,10 @@ def evaluate_selection_with_args(kwarg_arguments, field_defn, next_path, ast_nod
...
-            if r.graphql_dead
-              return true
-            else
-              r = r.graphql_parent
-            end
-          end
-          false
+          # When a result is marked dead, any of its existing children are marked dead
+          # and it accepts no more writes. But this result might have been initialized,
+          # but not yet assigned to a key in its parent. So check the parent too.
```
eapache, line 444: 🤔 how can it _have_ a parent but not yet be assigned to a key in its parent? Don't those happen at the same point?

Instead of checking the parent here, should we check "is the parent dead?" when we assign a key to its parent, and proactively mark it dead at that point? Would that solve this? Or have I totally misunderstood?

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: Yeah, technically the gem still supports Ruby 2.2, but I should probably _not_, since I think it's EOL: 

https://github.com/rmosolgo/graphql-ruby/blob/73d637d65aee3195f4e5c65a0118d94101bdee15/graphql.gemspec#L16

https://github.com/rmosolgo/graphql-ruby/blob/73d637d65aee3195f4e5c65a0118d94101bdee15/.rubocop.yml#L7

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: (Addressed by #3537)

lib/graphql/execution/interpreter/runtime.rb
rmosolgo: I'm not _exactly_ sure how it works, but without this check, the test added in https://github.com/rmosolgo/graphql-ruby/pull/3505 fails. 

In that test, the first item in a non-null list that encounters an execution error adds an error to the response and the list is marked dead. The later items in the list don't add errors to the response, since the identify that their parent (the non-null list) is marked dead. 

Maybe there's another way to implement it where the `each { ... }` call is halted when the list is marked dead. But I think there'd be some edge cases with promises, etc, where the iterations of `.each { ... }` all return, but then lazy values cause the list to be marked dead. And I doubt we have tests for those...

rmosolgo:
By taking the "heavy lift" of recursive dead marking away from the read (hot path) and on to the write (cold path), this put our synthetic benchmark back on par with `master` before #3533, speed-wise, anyway: 


```diff
 Calculating -------------------------------------
 Querying for 1000 objects
-                           5.108  (± 0.0%) i/s -     26.000  in   5.096948s
+                           5.177  (± 0.0%) i/s -     26.000  in   5.025598s
# ... 
- Total allocated: 12555624 bytes (73399 objects)
+ Total allocated: 13686480 bytes (78403 objects)
```

The memory overhead I consider unavoidable, since we need _some_ kind of container for metadata at runtime.

Issue #3525, opened 2021-06-12
Subscription with argument of type [CustomTypeHere!]! fails


HavinLeung:
**Describe the bug**

I have a subscription class which takes in an argument `input`, of type `[CustomTypeHere!]!`.
When I subscribe, the server spits out a `NoMethodError - undefined method 'arguments'` error

**Versions**

`graphql`: 1.12.12
`rails`: 6.1.3.2

**GraphQL schema**

```ruby
module Types
  class CustomTypeHere < Types::BaseInputObject
    description "Foo"

    argument :foo, GraphQL::Types::ID, required: false
    argument :bar, String, required: false
    argument :baz, Int, required: true

    validates required: { one_of: [:foo, :bar] }
  end
end

module Subscriptions
  class MySubscription < BaseSubscription
    argument :input, [Types::CustomTypeHere], required: true

    field :some_return, String, null: true
    field :errors, [Types::UserErrorType], null: false

    def authorized?(**args)
      false # it never gets this far
    end

    def subscribe(**args)
      {} # it never gets this far
    end
  end
end

```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
subscription a {
  agencyProfileAdd(input: [{
    bar: "0",
    baz: 0
  }]) {
    someReturn
    errors {
      message
      path
    }
  }
}
```

**Steps to reproduce**

Run the subscription, check the error backtrace output in `rails s`

**Expected behavior**

Expect to at least hit the `#authorized?` method

**Actual behavior**

Never gets that far, some argument resolution logic seems to go wrong.

Place full backtrace here (if a Ruby exception is involved):

<details>
<summary>Click to view exception backtrace</summary>

```
GraphqlChannel is transmitting the subscription confirmation
GraphqlChannel#execute({"query"=>"subscription a {\n  mySubscription(input: [{\n    baz: 0,\n    bar: \"0\"\n  }]) {\n    someReturn\n    errors {\n      message\n      path\n    }\n  }\n}\n", "variables"=>{}, "operationName"=>"a"})
[...]

PR #3526, opened 2021-06-15
Fix recursive and duplicate stringify_args call

TheWongGuy:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3525

Subscriptions fail when given an argument of type [CustomTypeHere!]!

Please see attached issue for more information on the problem

In the recursive stringify_args call it receives a `GraphQL::Schema::NonNull` as the `arg_owner` which does not have the arguments method that is called in `get_arg_definition`.

This patch is working for us on our project, but I am afraid I don't have full context over the guts/internals of this section of the code.

Just putting this up there to receive feedback and see what outside factors could be impacted by this change.
Todo:
- Write tests if this gets feedback from @rsmosolgo

### DIFF
--- a/lib/graphql/subscriptions/event.rb
+++ b/lib/graphql/subscriptions/event.rb
@@ -35,7 +35,7 @@ module GraphQL
           arguments
         when Hash
           if field.is_a?(GraphQL::Schema::Field)
-            stringify_args(field, arguments)
+            arguments
           else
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hey, sorry for the slow response on this. This change is certainly fine with me, assuming that there's a new test that shows it fixes the problem at hand, and that it doesn't break any of the other existing tests!

PR #3530, opened 2021-06-17
Makes the test useful.

LeFnord:
This spec was introduced here https://github.com/rmosolgo/graphql-ruby/pull/3505, so I think, the Null stuff should be tested here, right? Before an unauthorized was the error reason …

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,11 @@
 name: CI Suite
 on:
-  push
+  push:
+    branches:
+      - '*'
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi! Thanks for the suggestion. This specific test was added because a change I made to graphql-ruby broke a test case in Shopify. I isolated a reproduction and added a test here so that I could make sure I didn't break Shopify again. 

I'd like to keep the existing test case to make sure that I stay compatible for that particular behavior. If you'd like to contribute another test, please open another PR!

PR #3524, opened 2021-06-10
⚡️ Small optimization to GraphQL::Schema::InputObject#to_h

danielvdao:
I think this is a small optimization to the to_h call. I noticed it while stumbling through source code for a project I am working on. `#merge!` will avoid creating extra hash objects. I think this would be faster if the argument type had a lot of fields / nested fields on it.

PR #3539, opened 2021-06-24
Revert alias changes

rmosolgo:
Ugh! `alias_method :to_hash, :to_h` made it so that if you override `def to_h`, it isn't picked up by `.to_hash`.

Since obviously I don't understand the difference between `alias_method` and `alias`, I'm reverting all these changes from #3524

### DIFF
--- a/lib/graphql/argument.rb
+++ b/lib/graphql/argument.rb
@@ -8,7 +8,7 @@ module GraphQL
     attr_accessor :description, :name, :as, :deprecation_reason
     attr_accessor :ast_node
     attr_accessor :method_access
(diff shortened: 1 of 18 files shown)


danielvdao:
That's frustrating - I'm sorry you had to clean this up @rmosolgo 😞. Next time I'll keep my PR smaller and not be too aggressive with the changes.

rmosolgo:
No worries at all. You remember how I was hesitant about messing with `alias`? My takeaway is, I should be _more_ afraid of it, and afraid of `alias_method`, too 😆 .

PR #3543, opened 2021-06-29
[RFC] Replicate types hints to DSL methods

amomchilov:
The auto-complete experience on the DSL methods (such as in RubyMine) is quite poor, because a lot of them just accept `*args, **kwargs, &block`.

I suggest we add type hints to the DSL methods. Sure, this causes duplication, and isn't DRY, but I think the benefit to IDE quick documentation and auto-completion is worth it.

I've done it for `field` as a sample.

WDYT?

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -209,6 +209,7 @@ module GraphQL
       # @param method_conflict_warning [Boolean] If false, skip the warning if this field's method conflicts with a built-in method
       # @param validates [Array<Hash>] Configurations for validating this field
       # @param legacy_edge_class [Class, nil] (DEPRECATED) If present, pass this along to the legacy field definition
+      # @note Keep this type definition in sync with GraphQL::Schema::Member::HasFields#field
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for the suggestion. I'd love to support IDEs better (I don't use one though), but I agree, this approach is unmaintainable in the long term. 

I can think of a few approaches: 

- Use Sorbet types (would that work in RubyMine?) 
- Leverage YARD to make this work: 

    - Expand `def field(...)` to name all incoming kwargs, and add docs for them, and pass them all along by name (yuck) 
    - Add some kind of test to make sure that YARD doesn't spot any missing docs for that method or any orphaned documentation 
    
    I say to use YARD, but maybe the thing could be accomplished another way, with a regexp linter? (I don't think Rubocop works on comments.) 
    
    In theory, we could address all the current YARD warnings in the build and then add a test to make sure there aren't any _new_ YARD warnings, but I think there are a _lot_ of warnings to fix 🙈 
    
Can you think of any other more maintainable approaches, or does either of those stand out to you?

rmosolgo:
Ahh, too bad about sorbet not being supported yet. I'm game to include parameter documentations like you recommend here, but I'd like to _keep_ them in `def initialize`, too, since that's where the keywords are actually used. I find the documentation helpful from time to time when I'm working on those methods.

PR #3549, opened 2021-07-07
track deprecated arguments alongside deprecated fields

anujbiyani:
Added tracking for deprecated arguments. I didn't add tracking for all arguments as my app doesn't need it, but happy to add that if you prefer!

### DIFF
--- a/lib/graphql/analysis/ast/field_usage.rb
+++ b/lib/graphql/analysis/ast/field_usage.rb
@@ -32,8 +32,7 @@ module GraphQL
         def extract_deprecated_arguments(argument_values)
           argument_values.each_pair do |_argument_name, argument|
             if argument.definition.deprecation_reason
-              field = "#{argument.definition.owner.graphql_name}##{argument.definition.graphql_name}"
-              @used_deprecated_arguments << field
+              @used_deprecated_arguments << argument.definition.path
             end
 
             if argument.definition.type.kind.input_object?
--- a/lib/graphql/analysis/ast/field_usage.rb
+++ b/lib/graphql/analysis/ast/field_usage.rb
@@ -39,7 +39,10 @@ module GraphQL
             if argument.definition.type.kind.input_object?
               extract_deprecated_arguments(argument.value.arguments.argument_values)
             elsif argument.definition.type.list?
-              argument.value.each { |value| extract_deprecated_arguments(value.arguments.argument_values) }
+              argument
+                .value
+                .select { |value| value.respond_to?(:arguments) }
+                .each { |value| extract_deprecated_arguments(value.arguments.argument_values) }
             end
           end
         end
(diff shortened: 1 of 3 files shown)


lib/graphql/analysis/ast/field_usage.rb
```diff
@@ -7,21 +7,42 @@ def initialize(query)
...
+            if argument.definition.deprecation_reason
+              field = "#{argument.definition.owner.graphql_name}##{argument.definition.graphql_name}"
+              @used_deprecated_arguments << field
+            end
+
+            if argument.definition.type.kind.input_object?
+              extract_deprecated_arguments(argument.value.arguments.argument_values)
+            elsif argument.definition.type.list?
+              argument.value.each { |value| extract_deprecated_arguments(value.arguments.argument_values) }
+            end
```
anujbiyani, line 43: I'm not really sure if this is the best way to recurse through input objects. Do you have a better suggestion?

lib/graphql/analysis/ast/field_usage.rb
```diff
@@ -7,21 +7,42 @@ def initialize(query)
...
+            used_deprecated_arguments: @used_deprecated_arguments.to_a,
           }
         end
+
+        private
+
+        def extract_deprecated_arguments(argument_values)
+          argument_values.each_pair do |_argument_name, argument|
+            if argument.definition.deprecation_reason
+              field = "#{argument.definition.owner.graphql_name}##{argument.definition.graphql_name}"
```
anujbiyani, line 35: I chose `#` instead of `.` as an arbitrary way to differentiate between arguments and fields. Let me know if you prefer something else!

anujbiyani:
Hey @rmosolgo, would you be able to take a look at this?

lib/graphql/analysis/ast/field_usage.rb
```diff
@@ -7,21 +7,45 @@ def initialize(query)
...
+              @used_deprecated_arguments << field
```
rmosolgo, line 36: Yes, best to use `.` -- that's what we use other places, for example, `GraphQL::Schema::Argument#path` outputs something like `ObjectType.fieldName.argumentName`.  Honestly, you could probably use _that_:

```suggestion
              @used_deprecated_arguments << argument.definition.path
```

(`.path` was added after this analyzer was written, so if other code here doesn't use that method, that's probably why.)

rmosolgo (commented):
Thanks for the ping! I took a look before but I see I failed to submit my review -- it was sitting here "pending." Sorry about that!

lib/graphql/analysis/ast/field_usage.rb
anujbiyani: Fixed! [`9cc20ce` (#3549)](https://github.com/rmosolgo/graphql-ruby/pull/3549/commits/9cc20cebf9448fe50a655719d48ba566d8d40879)

PR #3559, opened 2021-07-18
configure a custom printer from the schema

ogidow:
[It is recommended that the redact condition be customized in a subclass that inherits from SanitizedPrinter](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/language/sanitized_printer.rb#L62-L63).

However, we cannot use custom SanitizedPrinter in Query#sanitized_query_string because it does not provide a way to set a custom SanitizedPrinter in the schema.
to use a custom SanitizedPrinter, you need to create a Query object, which is very inconvenient.

This PullRequest implements a way to set a custom SanitizedPrinter from the schema so that custom SanitizedPrinter can be used with Query#sanitized_query_string.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1631,6 +1631,14 @@ module GraphQL
         find_inherited_value(:multiplex_analyzers, EMPTY_ARRAY) + own_multiplex_analyzers
       end
 
+      def sanitized_printer(new_sanitized_printer = nil)
+        if new_sanitized_printer
+          @own_sanitized_printer = new_sanitized_printer
+        else
+          @own_sanitized_printer || GraphQL::Language::SanitizedPrinter
+        end
+      end
+
       # Execute a query on itself.
       # @see {Query#initialize} for arguments.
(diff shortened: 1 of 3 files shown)


lib/graphql/schema.rb
```diff
@@ -1631,6 +1631,14 @@ def multiplex_analyzers
         find_inherited_value(:multiplex_analyzers, EMPTY_ARRAY) + own_multiplex_analyzers
       end
 
+      def sanitized_printer(new_sanitized_printer = nil)
+        if new_sanitized_printer
+          @own_sanitized_printer = new_sanitized_printer
+        else
+          @own_sanitized_printer || GraphQL::Language::SanitizedPrinter
```
rmosolgo, line 1638: At some point, we may get bit because this isn't inherited. If that happens, we can use `find_inherited_value` as some of the other methods do. But it's ok with me to keep it this way for now.

rmosolgo (approved):
Thanks for this improvement! In our own usage, we have some instrumentation that manually initializes a printer, that's why I didn't need this before.

PR #3570, opened 2021-07-27
Make null context work with .authorized_new

rmosolgo:


gja:
@rmosolgo if you also define:

```ruby
def multiplex
  false
end
```

Then you'd be able t ado the following:

```ruby 
> GraphQL::Query::Context.new(query: GraphQL::Query::NullContext::NullQuery, values: nil, object: nil, schema: GraphQL::Schema).dataloader
        2: from (irb):9:in `<main>'
        1: from (irb):10:in `rescue in <main>'
NoMethodError (undefined method `multiplex' for GraphQL::Query::NullContext::NullQuery:Class)
```

thanhlecong865:
#0040

PR #3577, opened 2021-08-07
Postgresql fields support, downcased generator arguments, InputType generator, namespaced type generators

camiloforero:
There are several new things in this PR. Each of them is in its own commit with its own passing tests, so any combination of them should be selectable through cherry-picking.

1. ActiveRecord `citext` columns will be typed as `String`. `json`, `jsonb` and `hstore` columns will be typed as `GraphQL::Types::JSON`. 
2. A small bugfix; when running something like `rails g graphql:object example name:string`, if that `string` is lowercase it would be typed as `Types::StringType`, and similarly with everything, now type arguments in all lowercase should be typed correctly
3. An input type generator. Its essentially a copy paste of the object type generator with the necessary adjustments, and it can also hook with ActiveRecord to generate all arguments
4. Types are now generated within a folder, so for example an User object will go under `Types::Objects::UserType`, and so with all others that make sense. The generator templates also take this change into account. This is a change from what we have right now, but having all types mixed together in the root `Types` folder alongside the base types is quite messy, so this should help with that.
5. Mutation generator bugfix: When using the generator with a namespace, like `rails g graphql:mutation namespace/mutation_name`, the `mutation_name` that was gsubbed into the root `MutationType` kept the `/`. My way of fixing it was to refactor the mutation generator to inherit from `Rails::Generators::NamedBase` instead of `Rails::Generators::Base`, taking advantage of all the utilities that the Rails devs have created for this sort of thing. So that file ended up getting smaller, too, while still passing all tests
6. MutationCreate generator: scaffolds a `ObjectCreate`mutation for any given `Object`, that works out of the box with ActiveRecord and if previous conventions in input and object types are followed. It works very similar to the Rails generator code for controller scaffolds, and it should even work for Mongoid, although I haven't tested that.

I'll send the code for now, I can add the documentation later after I get some feedback about this


rmosolgo:
Hi! Thanks for all these suggestions. Everything except item 4 above (adding folders) sounds good to me. If you'd like to add a CLI flag to enable that behavior, that'd suit me.

camiloforero:
Alright. Over the next week/s I'll move the namespaced types into a CLI flag, add some more generators, add documentation, and check those CI flags

camiloforero:
@rmosolgo I'm finished with the CLI flag for namespaced types and the orm mutation_update and mutation_delete generators.

I saw that some of the CI tests were failing because `spec/integration/rails/generators/graphql/object_generator_spec` creates a new table, and when  `spec/integration/rails/generators/graphql/input_generator_spec` tried to create the same table, SQLite said it already existed and failed. There should be a way to solve this by cleaning the database per test case or something, but I don't have time to figure that out, so I just changed the name of the table for the input generator.

This is pretty much all I'd like to do right now to close this feature set, I want to add query type scaffolding, test cases scaffolding and seamless integration with rails model generators, but after I close this one and make sure it's well integrated.

So I can work on the documentation now, in the meantime could you re-run the CI tests and see if the code is good or if there are things I should change?

camiloforero:
@rmosolgo I've finished the documentation, it should be merge-ready now, assuming it passes all standards

lib/generators/graphql/type_generator.rb
```diff
@@ -76,23 +90,48 @@ def type_file_name
 
       # @return [Array<NormalizedField>] User-provided fields, in `(name, Ruby type name)` pairs
       def normalized_fields
+        pp "enter normalized fields"
+        pp fields
```
thomasklemm, line 94: @camiloforero Nice PR! Could these lines be leftovers from development?

lib/generators/graphql/type_generator.rb
```diff
@@ -76,23 +90,48 @@ def type_file_name
...
       end
 
+      def ruby_class_name
+        class_prefix = 
+          if options[:namespaced_types]
+            "#{graphql_type.pluralize.camelize}::"
+          else
+            ""
+          end
+          pp type_ruby_name
```
thomasklemm, line 109:
```suggestion
```

camiloforero:
@thomasklemm hi, thank you! They were indeed leftovers from development, I just removed them

camiloforero:
@thomasklemm @rmosolgo I fixed some errors with the create/update/delete generators I noticed when actually using them in one of my projects

Also, I was reviewing the latest CI results and fixing the failed test, and here are the reasons they failed:

- **`lint`:** I was using the `&.` safe navigation operator in a new class, but it was introduced in `ruby 2.3`, while `.rubocop.yml` has a `TargetRubyVersion` of `2.2`
- **`rails_3.2`:**
  - The `String#delete_prefix` method I'm using was introduced in `ruby 2.5`, while in the CI it is `ruby 2.3`
  - The `orm_instance.update` method in the ActiveModel generators was introduced in Rails 4. It was `update_attributes` on Rails 3
  - A GraphiQL name error I haven't been able to pinpoint yet
- **`rails_4.2`:**: Same `String#delete_prefix` issue. I tried running the tests with ruby 2.5.9 and they all passed.

Some of those are fixable, but the `orm_instance.update` isn't, unless I added a conditional according to the current Rails version. but I'm not sure if it's worth it. I will elaborate further on an issue/PR

brunoprietog:
Hi, any news on this? It looks very interesting. Thanks!

camiloforero:
@brunoprietog I went on a hiatus because of work responsibilities, but I have some free time now and I'll be working on getting this PR up to date

camiloforero:
Alright @rmosolgo, this one should be done now.

I made the mutation specs run faster by not requiring to run the rails generator and graphql generator once per test case, and also changed the object/interface generator to not write `null: true` for nullable fields, as per the latest defaults.

The two failing tests are also failing in other PRs. I have no idea about the truffle ruby one, but the Rails 7 one should be fixed when they release Rails 7.0.1 later this week

camiloforero:
Hey @thomasklemm @rmosolgo , I just wanted to check if there is something else I should do in order to get this ready for merge, it looks ready to me

camiloforero:
Alright, @thomasklemm, @rmosolgo , is this PR missing something before it can be merged? It looks good to me already, unless I'm missing something?

lib/generators/graphql/input_generator.rb
```diff
@@ -0,0 +1,52 @@
...
+module Graphql
+  module Generators
+    # Generate an input type by name,
+    # with the specified fields.
+    #
+    # ```
+    # rails g graphql:object PostType name:string!
+    # ```
+    #
+    # Add the Node interface with `--node`.
```
rmosolgo, line 14: I'm guessing this is left over from the object generator, right? 

```suggestion
```

lib/generators/graphql/type_generator.rb
```diff
@@ -82,15 +97,37 @@ def normalized_fields
...
 
-        def to_ruby
-          "field :#{@name}, #{@type_expr}, null: #{@null}"
+        def to_object_field
+          "field :#{@name}, #{@type_expr}#{@null ? '' : ', null: false'}"
+        end
+
+        def to_input_argument
+          "argument :#{@name}, #{@type_expr}, required: false"
         end
```
rmosolgo, line 131: How about considering`@null` when generating input arguments? 

```suggestion
        def to_input_argument
          "argument :#{@name}, #{@type_expr}#{@null ? ? ", required: false" : ""}"
        end
```

rmosolgo (commented):
Looking good so far! 

Sorry for my slow review. This PR is pretty intimidating 😅 I appreciate that you tried to put each change in its own commit, but personally, I didn't find it very helpful because you can't leave PR review comments on commits, so I had to switch between the full diff and each comment in order to leave comments. As I did so, I realized that some changes in previous commits were modified in later commits (I went looking for changes in the PR diff and couldn't find them). Of course that's totally normal, but next time, if you want a faster turn around in PR review, please open smaller PRs.

lib/generators/graphql/type_generator.rb
camiloforero: I considered it, but ultimately decided against it.

When working on my own projects I tried to make non-nullable fields required myself, but although it worked wonderfully when creating new records, it quickly broke down when editing them. It forced me to send back a value for every non-nullable field, even if I wasn't updating it, so in the end I found it more expedient to make all arguments for these sort of inputs optional, and turn database/model errors from required fields into `GraphQL::ExecutionError`s

camiloforero:
@rmosolgo don't worry, you must have been busy, and refactors plus features are always intimidating. I appreciate a lot that you put on the effort to actually review all of it even when it turned inconvenient.

This was the result of me doing a lot of work in small stretches of time, which is the way I work best, but I will take this into account and open smaller PRs next time. Probably by separating feature PRs from refactor PRs.

I merged one of your suggestions and commented on the other one.

lib/generators/graphql/type_generator.rb
rmosolgo: Fair enough, thanks for sharing your rationale! If later users report that the generator is "ignoring" their requests to make `required: true` input fields, I may update this behavior in the future, but we'll see.

rmosolgo:
Thanks again for all your work on these improvements!

Issue #3576, opened 2021-08-06
Question: is there a supported/suggestion server side Pagination implementation for Connection Types


joelwarrington-jobber:
Hey @rmosolgo  👋 ,

Is there any supported/suggested method for handling Pagination server side when querying for ActiveRecord while limiting the number of items returned?

I know of  `max_page_size` and `default_max_page_size` however this requires you to still return the entire collection to be able to paginate and also get accurate item totals. Are there any out-of-the-box implementations which are recommended?

Example:

```ruby
field :dogs, resolver: Resolvers::DogsResolver


def DogsResolver
  type Types::DogType.connection_type, null: false

  def resolve
    # assume there are a lot of dogs, and the limit number would be something manageable (like 100)
    Dog.all.limit(1)
  end
end
```

```gql
query DogsQuery {
  dogs {
    pageInfo {
      startCursor
      hasPreviousPage
      hasNextPage
      endCursor
    }
    totalCount
  }
}
```

```json
{
  "data": {
    "dogs": {
      "pageInfo": {
        "startCursor": "MQ",
        "hasPreviousPage": false,
        "hasNextPage": false,
        "endCursor": "MQ"
      },
      "totalCount": 1
    }
  }
}
```

joelwarrington-jobber:
@rmosolgo, my method to confirm may have been flawed, when adding a `binding.pry`, and getting the `.to_sql` of the query it didn't include a limit. Does the `max_page_size` limit ActiveRecord queries somehow, and will this still provide the `totalCount` that is accurate (even if limited to something below that)

rmosolgo:
Yes, I expect `max_page_size` to limit the query to fetch objects while `totalCount` returns an unbounded count.

joelwarrington-jobber:
Awesome, thanks for the clarification, will do a bit more confirmation on my side and add any findings here so that if someone in the future finds this they also can get clarification. 😊

joelwarrington-jobber:
**🔔  closing issue, as working as expected**

---

The following comment includes some investigation as to why I saw `limit` not being applied to the query when using `.connection_type` and `default_max_page_size`, but in short, limiting and database pagination works as expected when using the built in connection type and max page size for ActiveRecord.

#### Why did the query not show `limit` being correctly applied?

[...]

PR #3582, opened 2021-08-11
Update wording of the `max_page_size` option to include database limit

joelwarrington-jobber:
Not sure how you'd prefer the wording, but would like a call out that specifies that the database query is also limited, as I (and presumably others) had assumed that it would just limit the returned data but still hit the database for a large list of items.

Is related to #3576


rmosolgo:
Thanks for the improvement!

Issue #3585, opened 2021-08-17
BigInt type does not default to base 10 while parsing inputs


theanirudhvyas:
**Describe the bug**

When a number with a leading 0 is passed for a BigInt type, it assumes it to be binary and parses it incorrectly. For example "0100" is parsed as 512 instead of 1000

**Versions**

`graphql` version: v1.12
`rails` (or other framework): 6.1.3

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok).
Are you using [interpreter](https://graphql-ruby.org/queries/interpreter.html)? Any custom instrumentation, etc?

```ruby
class Mutations::ParseNumber < Mutations::BaseMutation
  argument :number, GraphQL::Types::BigInt, required: true

  field :number, GraphQL::Types::BigInt, null: false

  def resolve(number:)
    { number: number }
  end
end

class ApplicationSchema < GraphQL::Schema
   mutation(Types::MutationType)
end

class Types::MutationType < Types::BaseObject
      field :parse_number, mutation: Mutations::ParseNumber
end
```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
mutation {
    parseNumber(number: "01000") {
        number
    }
}
```

```json
{
    "data": {
        "parseNumber": {
            "number": "512"
        }
    }
}
```

**Steps to reproduce**

to the BigInt input pass a string starting with 0.

**Expected behavior**

BigInt should parse the  number with base 10 by default. strings passed with a leading 0 should not be parsed in octal format.

**Actual behavior**

"01000" is parsed as 512 for the class Types::BigInt

**Additional context**

With these details, we can efficiently hunt down the bug!

rmosolgo:
Fixed by https://github.com/rmosolgo/graphql-ruby/pull/3586

PR #3586, opened 2021-08-17
Ensures numbers are parsed with base 10 in BigInt - Issue #3585

theanirudhvyas:
Closes Issue #3585


rmosolgo:
Thanks for the fix! Agreed this is the proper behavior here.

PR #3593, opened 2021-08-24
Avoid refining String when not necessary

casperisfine:
Refinements have a very significant performance overhead. The simple fact of refining a method, even if the refinement is never used will make calls to that method 40% slower.

On rarely called methods in doesn't matter that much, but graphql-ruby refine some common String methods, so any project including the gem suffer from a significant performance penalty.

See this benchmark for more details: https://gist.github.com/casperisfine/1c46f05cccfa945cd156f445f0d3d6fa


rmosolgo:
Wow, I knew refinements slowed things down when you _used_ them, but I didn't know that just _defining_ them also came with a cost! Thanks.

PR #3600, opened 2021-08-28
Update position of keyword argument in doc example

ravangen:
Ensure specific named arguments of interest are placed before `**kwargs` catchall


rmosolgo:
Oops, thanks for fixing this!

PR #3601, opened 2021-08-31
Fix non-null connection fields to only return one non-null error

rmosolgo:
This puts the behavior _back_ to how it was in 1.12.10

Also, stop resolving lists when they become dead.

Fixes https://github.com/rmosolgo/graphql-ruby/issues/3504#issuecomment-908825696

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -10,7 +10,17 @@ module GraphQL
       class Runtime
 
         module GraphQLResult
-          attr_accessor :graphql_dead, :graphql_parent, :graphql_result_name
+          def initialize(result_name, parent_result)
+            @graphql_parent = parent_result
+            @graphql_dead = parent_result && parent_result.graphql_dead
+            @graphql_result_name = result_name
+            # Jump through some hoops to avoid creating this duplicate hash if at all possible.
+            @graphql_metadata = nil
+          end
+
+          attr_accessor :graphql_dead
+          attr_reader :graphql_parent, :graphql_result_name
+
           # Although these are used by only one of the Result classes,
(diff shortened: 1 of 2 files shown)


swalkinshaw:
:shipit: this fixes https://github.com/rmosolgo/graphql-ruby/issues/3504#issuecomment-908825696 🎉

swalkinshaw:
@rmosolgo it still might be worthwhile asking if this is worth "fixing". We had a unit test failing that expected only 1 error, but it's not behaviour we _have_  to preserve. Though I think from the other related issues the spec was ambiguous anyway?

rmosolgo:
I certainly _default_ to preserving previous behaviors, unless there's some reason to change it. I can't think of any reason why fields should keep running after one of them nullifies the parent -- is there one?

swalkinshaw:
No I think this is the best behaviour 👍  it's a bit weird that execution continues, but adding those errors basically leaks that implementation detail?

rmosolgo:
> execution continues

I fixed that, too, here: https://github.com/rmosolgo/graphql-ruby/pull/3601/files#diff-8398a090ccfa07b5ba0482e924606c920a38f2b329ffd1d6246d29e65c9ef6a8R749

So, while lists are running, if one of them causes the result to be dead, then execution of the list stops. It's possible that objects need a similar fix 👀

swalkinshaw:
👏  yeah confirmed that to be the case at least for the test gist we had

rmosolgo:
I didn't think til after merging to run a quick benchmark, fixed up some memory usage in 088ecc71e

Interestinly, `be rake bench:profile_large_result` still shows 40 more bytes of memory after that change, but I can't figure out _where_ when I read the profile 😆 🤷

Issue #3591, opened 2021-08-23
Configure complexity on connection fields


crpahl:
We're exploring rate limiting our API similar to how it's done at Shopify using a complexity based approach:

- Throttle the request if the requested cost is greater than the apps currently available complexity
- Reduce the apps available complexity for a resolved query based on the actual query cost

With every field having a default complexity of 1, the `edges` and `node` field are contributing to the cost calculated by `GraphQL::Analysis::AST::QueryComplexity`. For example, I'll get a total complexity of 15 from it for the following query when I was only expecting 5 (5 client nodes with a single `id` scalar):

```
{
  clients(first: 5) {
    edges {
      node {
        id
      }
    }
  }
}
```

Is there a way to set the complexity for the connection related fields to 0 other than redefining them in our class that extends `GraphQL::Types::Relay::BaseConnection`?

As for calculating the actual query cost, I've done this by adding a field extension to all of our fields that accumulates the total complexity of all resolved fields. This approach doesn't include the complexity cost of `edges` or `node` which is why I'm exploring setting them to 0. This also seems to be what Shopify is doing based on the values I'm getting in their throttle cost extension when experimenting with their API?

Any help would be very much appreaciated!

swalkinshaw:
Shopify doesn't use the gem's connection interface (for legacy reasons) so unfortunately I can't really help with that part. But for calculating the actual cost, we use a tracer instead of field extensions.

rmosolgo:
>  other than redefining them in our class...?

That's the only way I can think of, since that's where the fields are defined! I think you'd want to override these methods:

https://github.com/rmosolgo/graphql-ruby/blob/efd0711298bde27c7ff23cb6c70f675da564affc/lib/graphql/types/relay/connection_behaviors.rb#L121-L126

https://github.com/rmosolgo/graphql-ruby/blob/efd0711298bde27c7ff23cb6c70f675da564affc/lib/graphql/types/relay/connection_behaviors.rb#L46-L50

https://github.com/rmosolgo/graphql-ruby/blob/efd0711298bde27c7ff23cb6c70f675da564affc/lib/graphql/types/relay/edge_behaviors.rb#L23

crpahl:
Thank you both!

After following both of your suggestions, I was able to consistently get an actual query cost that matches the expected cost when the returned number nodes matches the number requested in the query so that's great!

For the tracer, I ended up doing something like this:

```
module Tracers
  class ComplexityTracing
[...]

PR #3609, opened 2021-09-02
Improve default complexity calculation for connection fields

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3591

TODO: 

- [ ] How to release this to not break current users? Push off til 1.13.0? 
- [ ] Is raising the right behavior when no page size can be determined?
- [ ] Document ☝️

### DIFF
--- a/guides/queries/complexity_and_depth.md
+++ b/guides/queries/complexity_and_depth.md
@@ -77,6 +77,48 @@ class MySchema < GraphQL::Schema
 end
 ```
 
+#### Connection fields
+
+By default, GraphQL-Ruby calculates a complexity value for connection fields by:
+
+- adding `1` for `pageInfo` and each of its subselections
+- adding `1` for `count`, `totalCount`, or `total`
+- adding `1` for the connection field itself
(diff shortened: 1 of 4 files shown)


rmosolgo:
To get the old behavior, make `def calculate_complexity(**kwargs)` use the code that used to be hardcoded in the analyzer: 


```ruby
class Types::BaseField < GraphQL::Schema::Field 
  # Override graphql-ruby 1.13's new default complexity for connection fields 
  def calculate_complexity(query:, nodes:, child_complexity:)
    defined_complexity = complexity
    case defined_complexity
    when Proc
      arguments = query.arguments_for(nodes.first, self)
      defined_complexity.call(query.context, arguments.keyword_arguments, child_complexity)
    when Numeric
      defined_complexity + child_complexity
    else
      raise("Invalid complexity: #{defined_complexity.inspect} on #{path} (#{inspect})")
    end
  end
end 
```

(You can find that same code in the `else ...` branch of the new `def calculate_complexity(...)` method in this PR.)

PR #3612, opened 2021-09-08
Make `null: true` and `required: true` the defaults, update CI Ruby versions

rmosolgo:
Currently, every field must have `null: true` or `null: false`. Likewise, every argument must have `required: true`/`required: false`. This is verbose, and I think we can recommend a better default.  So in this branch: 

- `null: true` and `required: true` are applied when those options are omitted. (I picked these because they can always be changed later, unlike `null: false` or `required: false`, which require breaking changes to change in the future)
- A couple rubocop linters to automatically update those configs. (I used the linter to update this codebase, too.)
- Update the docs
- As a side-effect, updates this project's rubocop config to 2.5+. I did this so that I could use a recent Rubocop version for writing cops. I think it will be ok to remove support for some ruby/rails versions in 1.13.0, but I may have to keep missing with the build for this to work right.

Does anyone have an opinion on this change? I've done the work to see if it's feasible (it seems to me like it is), and I see the advantage of it, but I'd love to hear any other takes on this idea. (Maybe even the defaults should be different?)

cc @swalkinshaw  did you mention once that Shopify applied defaults in these cases?


joelwarrington-jobber (commented):
> Does anyone have an opinion on this change? I've done the work to see if it's feasible (it seems to me like it is), and I see the advantage of it, but I'd love to hear any other takes on this idea. (Maybe even the defaults should be different?)

Agree with the change, and these defaults make it so that if it needs to change it won't result in a breaking change.

jgrau:
For the ones looking for the configuration to run/configure the rubocop cops mentioned in this PR [this file](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/fixtures/cop/.rubocop.yml) has the answer :)

PR #3621, opened 2021-09-25
Fix typo for Subscription#subscribe documentation

choznerol:
> #&nbsp;Default implementation returns the root object.
> ...
> def subscribe

- `Subscription#update` returns root object by default
- `Subscription#subscribe` returns :no_response by default

Therefore this line seems to be a typo.


rmosolgo:
Thanks for the correction on this!

PR #3626, opened 2021-09-29
Add missing font-weight to guides

gopeter:
This is how the guide looks in Chrome:

image

Bold text is blurry and not really bold, because the right font is missing:

image

And this is how it looks when all font weights are loaded:

image


rmosolgo:
Thanks for the fix! I wonder why it looks fine for me ... maybe my system already has the font installed 😖

Issue #3619, opened 2021-09-24
Error when parsing malformed query should raise an appropriate exception


such:
**Describe the bug**

`NoMethodError` when parsing malformed query:

```
{
  "action": "execute",
  "controller": "graphql",
  "graphql": {
    "query": {
      "default": "query {\n  clubsReady {\n    name\n    slug\n    domesticLeague {\n      displayName\n      slug\n      country {\n        slug\n      }\n    }\n    country {\n      slug\n    }\n  }\n}"
    }
  },
  "query": {
    "default": "query {\n  clubsReady {\n    name\n    slug\n    domesticLeague {\n      displayName\n      slug\n      country {\n        slug\n      }\n    }\n    country {\n      slug\n    }\n  }\n}"
  }
}
```

**Versions**

`graphql` version: 1.12.16
`rails`: 6.1.4.1

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
"query": {
    "default": "query {\n  clubsReady {\n    name\n    slug\n    domesticLeague {\n      displayName\n      slug\n      country {\n        slug\n      }\n    }\n    country {\n      slug\n    }\n  }\n}"
  }
```

**Expected behavior**

Parsing error should be returned to the caller or a proper exception should be raised

**Actual behavior**

What specifically went wrong?

Place full backtrace here (if a Ruby exception is involved):

<details>
<summary>Click to view exception backtrace</summary>

```
undefined method `unpack' for #<ActionController::Parameters {"default"=>"query {\n  clubsReady {\n    name\n    slug\n    domesticLeague {\n      displayName\n      slug\n      country {\n        slug\n      }\n    }\n    country {\n      slug\n    }\n  }\n}"} permitted: false>


lib/graphql/language/lexer.rb:170 run_lexer |  
-- | --
lib/graphql/language/lexer.rb:11 tokenize |  
lib/graphql.rb:56 scan_with_ragel |  
lib/graphql.rb:52 scan |  
parser.y:452 block in parse_document |  
lib/graphql/tracing.rb:81 call_tracers |  
lib/graphql/tracing.rb:83 block in call_tracers |  
graphql-pro (1.18.2) lib/graphql/pro/operation_store/query_instrumentation.rb:19 trace |  
lib/graphql/tracing.rb:83 call_tracers |  
lib/graphql/tracing.rb:83 block in call_tracers |  
graphql-guard (2.0.0) lib/graphql/guard.rb:42 trace |  
[...]

PR #3628, opened 2021-09-29
Assert that query is a string on GraphQL::Query#initialize

diogoosorio:
This commit aims to solve the issue #3619 - providing a query argument of an incorrect type would yield a `NoMethodError` due to the incorrect type argument to be allowed through to the lexer (which assumes it to be a string).

As this parameter is, more often than not, feed from an external source (e.g. a web request), this commit assumes a more defensive approach and ensures that when the `GraphQL::Query` is initialised, the query string is indeed of the correct type (a string).

As discussed in the issue, an `ArgumentError` is raised when a type mismatch is identified.

----

I've actually just received an error notification due to this issue. I think I'll end up addressing it at a different layer (perform the input validation/coercion at the controller layer). Regardless I tried to make the gem a bit more defensive in this regard - it's my first contribution, do let me know if there's a better place to perform the assertion and/or I'm missing some kind of test. 😄

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -117,6 +117,10 @@ module GraphQL
         raise ArgumentError, "Query should only be provided a query string or a document, not both."
       end
 
+      if @query_string && !@query_string.is_a?(String)
+        raise ArgumentError, "Query string argument should be a String, got #{@query_string.class.name} instead."
+      end
+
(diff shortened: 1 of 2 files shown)


rmosolgo:
Looks good to me! Thanks for addressing this.

such:
Thank you! When do you think you can release this?

Issue #3596, opened 2021-08-25
Stop supporting versions of ruby < 2.6, rails < 5


camiloforero:
### Ruby
source: https://www.ruby-lang.org/en/downloads/branches/

Ruby 2.5 went EOL on the last day of March, this year.
Ruby 2.3, the earliest version used on the CI tests, went EOL on 2016, two years and a half ago
Ruby 2.2, the rubocop linter target, went EOL three years and a half ago.

### Rails

Rails 3 went EOL on 2016, 5(!) years ago, with [the launch of Rails 5](https://weblog.rubyonrails.org/2016/6/30/Rails-5-0-final/)
Rails 4 went EOL on August 2019, 2 years ago, with [the launch of Rails 6](https://weblog.rubyonrails.org/2019/8/15/Rails-6-0-final-release/).

However, in the automated tests that run on the CI, backwards compatibility is maintained with all of those versions.

I'd suggest that if the official maintainers aren't supporting those versions anymore, we shouldn't either for future GraphQL versions. If someone using a very old version of Ruby/Rails wants to keep using this library, they can always lock their `graphql` version to the current one.

However, @rmosolgo, you'd know better about who is using this gem, and in which versions of ruby/rails. Is there a reason you've been supporting such old versions for this long?

If you don't want to be so drastic, and also for the purposes of merging #3577, I'd suggest
- Support rails 4.2+, to be able to use the newer ORM generators
- Support ruby 2.4+, mostly because versions 2.3 and earlier can't be easily installed on newer OSses (for example, in my Ubuntu 20.04 laptop I couldn't, as it required gcc and g++ v6, which isn't even available in the default repositories)

I can make the changes in the CI configuration as needed, according to the decision made

rmosolgo:
👋  Thanks for asking. I'm hesitant to remove support for these versions very quickly; many of them were added by people who still used those versions in production. I don't know if they still do or not, but in any case, I don't want to remove it without checking with them. 

For #3577, I recommend updating the PR to pass the current CI configuration. Then, if you're interested in removing support for these old versions, you can open new PRs for them to start the discussion around it.

PR #3640, opened 2021-10-07
CI: Rails 4+, Ruby 2.4+

camiloforero:
As per https://github.com/rmosolgo/graphql-ruby/issues/3596, I opened a PR with the CI changes needed to support at a minimum
* Ruby 2.4+
* Rails 4+

And prepare the way to support the upcoming Rails 7.

I took the opportunity to make all the changes needed in all the parts of the CI system I could find, like removing `RUBY_VERSION >= '2.4.0'`conditionals, updating rubocop version, and removing some old string refinements that shouldn't be needed anymore.

Removing support from these very old Rails and Ruby versions should speed up development moving forward.

Rails 4 and Ruby 2.4 themselves are still somewhat old and both EOL, but support for those ones is kept for a while longer

This PR can stay open for a while, for people to discuss whether they still need those versions or not.

Personally, I find it hard to believe that there are still people out there who are still using the latest `graphql-ruby` version and always keeping it updated while still using Rails 3 or Ruby 2.2. I'd argue that changing from `define` based to Class based API back in `1.8`, even with the migrator script, took a lot of effort, and anyone who can invest that effort can also invest the effort to keep their Ruby and Rails versions updated. Maybe I'm wrong on this, so let's see what other people have to say.

Also, it's not like GraphQL would stop working for them. Assuming this PR makes it to a `1.13` release, they'd still be able to use everything up to `1.12` and use later versions at their own risk, just like they are using unsupported Ruby and Rails versions that don't receive security fixes at their own risk, anyway.

Tests all worked on my local machine using the ruby versions specified in the `ci.yaml`


rmosolgo:
>  Wouldn't it make more sense to only support ruby versions that aren't EOL?

In a perfect world, everyone would be running recent Ruby versions. But in practice, some people run old Ruby versions. I'd rather not boot anyone from GraphQL-Ruby if I can help it.

I'm open to dropping support for some old Ruby versions in 1.13.0, but I haven't put a lot of thought into it yet.

rmosolgo:
Well, shoot. I wanted to remove the string refinement used in the lexer, but in order to rebuild it properly, I had to update Ragel. I added a benchmark task to make sure there wasn't a regression... but there _was_: 

```
# Before, Ragel 7.0.0.9
scan - introspection   110.000  i/100ms
parse - introspection
                        82.000  i/100ms
    scan - fragments   191.000  i/100ms
   parse - fragments   132.000  i/100ms
    scan - big query    18.000  i/100ms
   parse - big query    13.000  i/100ms
scan - introspection      1.087k (± 3.5%) i/s -      5.500k in   5.068137s
parse - introspection
                        809.097  (± 2.1%) i/s -      4.100k in   5.069787s
    scan - fragments      1.882k (± 2.2%) i/s -      9.550k in   5.076049s
   parse - fragments      1.357k (± 1.5%) i/s -      6.864k in   5.061174s
    scan - big query    186.115  (± 2.1%) i/s -    936.000  in   5.031735s
   parse - big query    136.186  (± 3.7%) i/s -    689.000  in   5.066171s

   
# After, Ragel 7.0.4
scan - introspection    66.000  i/100ms
parse - introspection
                        52.000  i/100ms
    scan - fragments   109.000  i/100ms
   parse - fragments    87.000  i/100ms
    scan - big query    10.000  i/100ms
   parse - big query     7.000  i/100ms
scan - introspection    647.735  (± 2.0%) i/s -      3.300k in   5.096960s
parse - introspection
                        532.360  (± 2.3%) i/s -      2.704k in   5.081807s
    scan - fragments      1.048k (± 2.6%) i/s -      5.341k in   5.101512s
   parse - fragments    854.413  (± 2.2%) i/s -      4.350k in   5.093800s
    scan - big query    103.480  (± 1.9%) i/s -    520.000  in   5.026922s
   parse - big query     86.099  (± 3.5%) i/s -    434.000  in   5.049648s
```

So, I guess I _don't_ want to update Ragel. I'll roll back that change and edit the generated file by hand 😖

Rakefile
```diff
@@ -61,8 +61,8 @@ ERR
     end
   end
 
-  assert_dependency_version("Ragel", "7.0.0.9", "ragel -v")
-  assert_dependency_version("Racc", "1.5.2", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
+  assert_dependency_version("Ragel", "7.0.4", "ragel -v")
+  assert_dependency_version("Racc", "1.6.0", %|ruby -e "require 'racc'; puts Racc::VERSION"|)
```
rmosolgo, line 65: TODO revert these changes

Issue #3641, opened 2021-10-07
Need access `owner_object` in evaluate_selection_with_args


digitalmoksha:
Hi.  I'm still working on upgrading to the interpreter, and I've run into another snag.

When not using the interpreter, we are able to do something like this:

```ruby
      field :web_url, GraphQL::Types::String, null: true,
            extras: [:parent]

      def web_url(parent:)
        owner = closest_parent([::Types::ProjectType, ::Types::GroupType], parent)
        ...
```

We are then able to inspect the parent structure and determine the original model we had passed in - basically `parent.object.object`

`GraphQL::Query::Context::FieldResolutionContext` exposed the `parent` value, so that `extras: [:parent]` worked.  But that class is no longer used for the interpreter.

However, if I could access the `owner_object` when processing `extra` in `evaluate_selection_with_args` (https://github.com/rmosolgo/graphql-ruby/blob/61232b03412df6685406fc46c414e11d3f447817/lib/graphql/execution/interpreter/runtime.rb#L449), then I can get what I need via `owner_object.object.parent`

But there seems to be no way to grab it.  Any ideas?

Sorry, my explanation is a little lacking. 🤦

PR #3645, opened 2021-10-11
Support extras: [:parent]

rmosolgo:
This was available in the old runtime and can be supported without much trouble. Fixes https://github.com/rmosolgo/graphql-ruby/issues/3641

cc @digitalmoksha, want to give this branch a try and see if it works for you? 

```ruby 
gem "graphql", github: "rmosolgo/graphql-ruby", ref: "extras-parent"
```


digitalmoksha:
@rmosolgo I was finally able to verify this works.  Like charm.  Much appreciated!

PR #3659, opened 2021-10-18
Add set_action_name option to AppsignalTracing

### LINKED ISSUE #3646: Appsignal tracing groups all graphql queries under GraphQLController#execute instead of separating per query
**Describe the bug**

Tracing with appsignal puts all traces of different graphql queries under Core::GraphQLController#execute . Previously we also tried tracing with skylight, and there all our queries did get neatly separated.

I configured appsignal by putting `use(GraphQL::Tracing::AppsignalTracing)` in our schema as directed in the graphql ruby docs.

**Versions**

`graphql` version: 1.12.16
`rails` 6.1
`appsingal`3.0.14

**Expected behavior**

I expect the tracing to be separated by graphql query

**Actual behavior**

The tracing is all grouped under Core::GraphQLController#execute

**Additional Info**

I did see that it may be possible to [separate these queries manually](https://github.com/appsignal/appsignal-ruby/issues/486), but it would be nice if this happened out of the box by graphql-ruby.

BoKleynen:
Closes #3646

### DIFF
--- a/lib/graphql/tracing/appsignal_tracing.rb
+++ b/lib/graphql/tracing/appsignal_tracing.rb
@@ -14,7 +14,22 @@ module GraphQL
         "execute_query_lazy" => "execute.graphql",
       }
 
+      # @param set_action_name [Boolean] If true, the GraphQL operation name will be used as the transaction name.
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for this improvement!

PR #3671, opened 2021-10-25
Make dedicated AllowBlank and AllowNull validators

### LINKED ISSUE #3492: Unexpected behavior with validation of empty string argument
**Describe the bug**

We have a string argument on a field that we want to ensure is not null and not blank/empty. We can achieve the not null by adding `required: true` to the argument definition, however achieving the not empty has been tricky. I didn't find a [built-in validator](https://graphql-ruby.org/fields/validation.html#built-in-validations) that I could use to ensure not blank, so I wrote a custom `NotBlankValidator` class to do this. Once I defined this validator on my argument, I started getting the expected "{argument name} can't be blank" validation error message, but not for the expected reason.

The "{argument name} can't be blank" validation error message is coming from the graphgql gem from [this line of code](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/validator.rb#L41). My custom validator doesn't get called when the argument's value is an empty string. It looks like the `GraphQL::Schema:: Validator#apply` method [only gets called](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/validator.rb#L125) when the argument has validators defined on it. If I remove my custom validator from the argument then the `GraphQL::Schema:: Validator#apply` method doesn't get called, and the query happily accepts an empty string for this argument.

I want the "not blank" behavior for this argument, but it seems odd to have to define some/any validator on the argument just to get the built-in "not blank" check that the `GraphQL::Schema:: Validator#apply` method provides. Perhaps there should be a built-in `not_blank` validator that can be used for an argument when you want this behavior.

Please let me know if you need me to provide a sample application that demonstrates this behavior. I'm hoping my explanation above is sufficient to define the problem. Thanks.

**Versions**

`graphql` version: 1.12.8
`rails` version: 6.1.3.1

**GraphQL schema**

**GraphQL query**

**Steps to reproduce**

**Expected behavior**

**Actual behavior**

**Additional context**

### LINKED ISSUE #3584: Allow validators to specify extensions to the error being raised
### Problem Summary
A common issue with handling execution errors client-side is understanding what the actual error was and how to proceed. Typically client's resort to regex matching on the human description. GraphQL (and this library) offers the ability add `extensions` to errors to give more context. To make it easier for the client to understand, a common approach is to add an enumerable error code to the errors extensions.

Unfortunately, the current validator implementation does not expose a way to add extensions to the error being raised.

### Proposed Options

- One possible solution would be to support returning a Hash from the `validate` method which supports an `extensions` key which is applied to the error when raised upstream.
- Alternatively, users could raise directly within a custom validator with their desired extensions?

rmosolgo:
Fixes #3492 

Fixes #3584 

- Make validators work more like rails -- `allow_blank:` and `allow_null:` can be used as "top-level" validators
- Refactor where `allow_blank` and `allow_null` are implemented so that custom errors can be easily raised 

TODO: 

- [x] Unit tests for AllowBlankValidator
- [x] Reconsider `PASSES_VALIDATION` -- how does Rails implement that? 
- [x] Add tests for custom raising in validators


rmosolgo:
Shoot, I thought this was going into 1.13-dev, but apparently not. The problem is, it'll update the website with some incorrect docs 😖 . If I get a question about it, I'll release a version at that time.

PR #3675, opened 2021-10-27
Limit Static Validation Errors in 1.11.x

mattpage:
Currently the static validator generates verbose error messages without limitation on the size of the overall error generated. This PR adds error limiting into the Static Validator and its associated rules.

By default, the error limiting is disabled. 

It is configured by defining `validate_max_errors` in the schema. When the value is set, the validator will add up to that number of errors but no more.

I have added tests to all rule specs that were modified and include a simple integration test.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -157,7 +157,7 @@ module GraphQL
 
     accepts_definitions \
       :query_execution_strategy, :mutation_execution_strategy, :subscription_execution_strategy,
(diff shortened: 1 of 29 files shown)


rmosolgo (approved):

mattpage:
@rmosolgo - do you want me to open a PR against latest for the same changes?

rmosolgo:
I did already, thanks! #3683 It's 🚢 in 1.12.18.

PR #3690, opened 2021-11-03
use validation_max_errors in validation pipeline

mattpage:
This is a follow-up to https://github.com/rmosolgo/graphql-ruby/pull/3675

That PR introduced the ability to limit static validation errors.
Unfortunately, I forgot to hook it up to the validation pipeline 😿 

Anyway, this follow-up PR does just that.

### DIFF
--- a/lib/graphql/query/validation_pipeline.rb
+++ b/lib/graphql/query/validation_pipeline.rb
@@ -72,7 +72,7 @@ module GraphQL
         elsif @operation_name_error
           @validation_errors << @operation_name_error
         else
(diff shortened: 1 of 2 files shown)


rmosolgo:
Oh, how weird! Were those validation tests taking some other codepath that didn't go by here? (Any way to add a test demonstrating what problem this fixes?)

mattpage:
@rmosolgo - I think this didn't get caught because the integration test I added to #3675  was on schema.rb. Anyway, thanks for adding the test. Would you mind publishing a 1.11.10 gem? Cheers!

rmosolgo:
D'oh, right -- shipped! Thanks again.

PR #3698, opened 2021-11-05
Improve field conflict error messages

alexkalderimis:
Following the merge of https://github.com/rmosolgo/graphql-ruby/pull/3690, we have a good general solution for excessive validation error messages.

However, we may still want to improve the field-conflict error messages, since these are confusing, and include every pair-wise combination of conflicting values. A simpler representation is a list of all conflicting values, as implemented here.

Short-circuiting behavior is preserved, honoring the `validation_max_errors` configuration.

### DIFF
--- a/lib/graphql/static_validation/error.rb
+++ b/lib/graphql/static_validation/error.rb
@@ -32,8 +32,10 @@ module GraphQL
 
       private
 
+      attr_reader :nodes
+
       def locations
-        @nodes.map do |node|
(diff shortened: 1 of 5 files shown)


alexkalderimis:
The test failures (on ruby 3.0.2) do not look related to these changes, and I notice that they appear on other PRs with different changes.

rmosolgo:
Thanks for this improvement!

PR #3705, opened 2021-11-09
Fix JSON as subscription argument

netshade:
This PR fixes a case where, when a subscription uses a `GraphQL::Types::JSON` type as an argument, the value provided is recursively inspected as if the sub hash would have corresponding argument entries for its keys. In this case a `GraphQL::Types::JSON` object is opaque at the moment is encountered and should just be passed through as a literal to allow for the topic name to be created.


lib/graphql/subscriptions/event.rb
```diff
@@ -65,12 +65,24 @@ def stringify_args(arg_owner, args)
...
               if arg_defn
                 normalized_arg_name = camelized_arg_name
-                next_args[normalized_arg_name] = stringify_args(arg_defn.type, v)
               else
                 normalized_arg_name = arg_name
                 arg_defn = get_arg_definition(arg_owner, normalized_arg_name)
               end
-              next_args[normalized_arg_name] = stringify_args(arg_defn.type, v)
+              arg_base_type = arg_defn.type
+              while arg_base_type.respond_to?(:of_type)
```
netshade, line 73: I struggled here, as this does not feel right, and that there must be some way to "unwrap" the low-level type that would be desired here; I was not able to easily find it, though, so if there is a better way to get at what I'm after happy to change it. 

I also don't know if I have a guarantee that the type chain here in fact will ever terminate and am slightly terrified this could cause an infinite loop, so happy to switch or add safeguards as well.

netshade:
Added some comments regarding things I had issues with / struggled w/ knowing the right way to resolve for the library.

netshade:
Here is the backtrace for how this fails when running the new test on `master`:

```
E

Finished tests in 0.000468s, 2136.7518 tests/s, 0.0000 assertions/s.


Error:
GraphQL::Subscriptions::Event#test_0001_should serialize a JSON argument into the topic name:
NoMethodError: undefined method `arguments' for GraphQL::Types::JSON:Class
   graphql-ruby/lib/graphql/subscriptions/event.rb:87:in `get_arg_definition'
   graphql-ruby/lib/graphql/subscriptions/event.rb:64:in `block in stringify_args'
   graphql-ruby/lib/graphql/subscriptions/event.rb:61:in `each'
   graphql-ruby/lib/graphql/subscriptions/event.rb:61:in `stringify_args'
   graphql-ruby/lib/graphql/subscriptions/event.rb:68:in `block in stringify_args'
   graphql-ruby/lib/graphql/subscriptions/event.rb:61:in `each'
   graphql-ruby/lib/graphql/subscriptions/event.rb:61:in `stringify_args'
   graphql-ruby/lib/graphql/subscriptions/event.rb:34:in `serialize'
   graphql-ruby/lib/graphql/subscriptions/event.rb:28:in `initialize'
   graphql-ruby/spec/graphql/subscriptions/event_spec.rb:26:in `new'
   graphql-ruby/spec/graphql/subscriptions/event_spec.rb:26:in `block (2 levels) in <top (required)>'
```

lib/graphql/subscriptions/event.rb
```diff
@@ -65,12 +65,24 @@ def stringify_args(arg_owner, args)
...
+                arg_base_type = arg_base_type.of_type
+              end
+              # In the case where the value being emitted is seen as a "JSON"
+              # type, treat the value as one atomic unit of serialization
+              is_json_definition = arg_base_type && arg_base_type <= GraphQL::Types::JSON
+              if is_json_definition
+                next_args[normalized_arg_name] = v.respond_to?(:to_json) ? v.to_json : v
```
rmosolgo, line 80: Usually, this method sorts hashes to make sure that clients who provide arguments _in different order_ still end up subscribing to the same topic. Should we also consider ordering here? 

In any case, how about adding a test to demonstrate the behavior with Types::JSON and ordering, so at least we can _know_ how it works?

rmosolgo (commented):
Looking good so far, thanks for working on this fix!

netshade:
@rmosolgo I made some adjustments predicated on "not ordering JSON hashes is acceptable", but happy to defer there.

lib/graphql/types/json.rb
```diff
@@ -10,6 +10,11 @@ module Types
     #
     #     argument :template_parameters, GraphQL::Types::JSON, null: false
     #
+    # Note that when the JSON scalar is used an argument for subscriptions, its value will be used to construct
+    # a topic name. You are responsible to ensure that any values provided by the client and the server are
+    # ordered identically such that both the client and the server agree on the correct topic name.
```
rmosolgo, line 15: This doesn't _seem_ right to me... 

Given an operation like 

```graphql 
subscription($filter: JSON!) {
  tweetWasTweeted(filter: $filter) { ... }
}
```

I would _expect_ that `filter: { handle: "matz", hashtags: ["#ruby"] }` would subscribe to the same events as `filter: { hashtags: ["#ruby"], handle: "matz" }`. (Arguments in GraphQL are _not_ ordered, anyways.)

Is there a reason to _prefer_ order-sensitive JSON inputs? If not, do you mind trying to normalize them on the server?

netshade:
@rmosolgo Made changes, biased to the idea that these values *should* be coming from JSON provided by the query, so ideally cycles not a thing to worry about, as well as keys not sortable with respect to each other. Does that seem right?

lib/graphql/subscriptions/event.rb
```diff
@@ -53,6 +53,34 @@ def fingerprint
 
       class << self
         private
+
+        # This method does not support cyclic references in the Hash,
+        # nor does it support Hashes whose keys are not sortable
+        # with respect to their peers ( cases where a <=> b might throw an error )
+        def deep_sort_hash_keys(hash_to_sort)
+          raise ArgumentError("Argument must be a Hash") unless hash_to_sort.is_a?(Hash)
```
rmosolgo, line 61:
```suggestion
          raise ArgumentError.new("Argument must be a Hash") unless hash_to_sort.is_a?(Hash)
```

(I don't think this will do what you expect:

```
ArgumentError("boom!")
# (irb):1:in `<main>': undefined method `ArgumentError' for main:Object (NoMethodError)
#	from /Users/rmosolgo/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
#	from /Users/rmosolgo/.rbenv/versions/3.0.2/bin/irb:23:in `load'
#	from /Users/rmosolgo/.rbenv/versions/3.0.2/bin/irb:23:in `<main>'
ArgumentError.new("boom!")
# => #<ArgumentError: boom!>
```

)

lib/graphql/subscriptions/event.rb
```diff
@@ -53,6 +53,34 @@ def fingerprint
...
+            if hash_to_sort[k].is_a?(Hash)
+              [k, deep_sort_hash_keys(hash_to_sort[k])]
+            else
+              [k, hash_to_sort[k]]
+            end
+          end.to_h
+        end
+
+        def deep_sort_array_hashes(array_to_inspect)
+          raise ArgumentError("Argument must be an Array") unless array_to_inspect.is_a?(Array)
```
rmosolgo, line 72:
```suggestion
          raise ArgumentError.new("Argument must be an Array") unless array_to_inspect.is_a?(Array)
```

lib/graphql/subscriptions/event.rb
```diff
@@ -53,6 +53,34 @@ def fingerprint
...
+        # nor does it support Hashes whose keys are not sortable
+        # with respect to their peers ( cases where a <=> b might throw an error )
+        def deep_sort_hash_keys(hash_to_sort)
+          raise ArgumentError("Argument must be a Hash") unless hash_to_sort.is_a?(Hash)
+          hash_to_sort.keys.sort.map do |k|
+            if hash_to_sort[k].is_a?(Hash)
+              [k, deep_sort_hash_keys(hash_to_sort[k])]
+            else
+              [k, hash_to_sort[k]]
+            end
```
rmosolgo, line 67: It seems like, if `hash_to_sort[k]` is an array containing hashes, then we _want_ to keep going here, right? 

```suggestion
            elsif hash_to_sort[k].is_a?(Array)
              [k, deep_sort_array_hashes(hash_to_sort[k])
            else
              [k, hash_to_sort[k]]
            end
```

rmosolgo:
> cycles not a thing to worry about, as well as keys not sortable with respect to each other. Does that seem right?

Yes, that's right!

netshade:
Made adjustments, thanks for prior reviews.

rmosolgo:
Thanks for all your work on this!

netshade:
Np, thanks for helping me with it. 🙏

PR #3707, opened 2021-11-10
Tell Sources to wait when one dataloader.yield didn't finish the load

### LINKED ISSUE #3706: Nested data source doesn't resolve dataloader in some cases
**Versions**

`graphql` version: 1.12.19

**Steps to reproduce**
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'graphql', '1.12.19'
end

class Echo < GraphQL::Dataloader::Source
  def fetch(keys)
    keys
  end
end

class Nested < GraphQL::Dataloader::Source
  def fetch(keys)
    dataloader.with(Echo).load_all(keys)
  end
end

class Nested2 < GraphQL::Dataloader::Source
  def fetch(keys)
    dataloader.with(Nested).load_all(keys)
  end
end

  field :nested, String, null: true
  field :nested2, String, null: true

  def nested
    dataloader.with(Nested).load("nested")
  end
  
  def nested2
    dataloader.with(Nested2).load("nested2")
  end
end

  query QueryType
  use GraphQL::Dataloader
end

pp TestSchema.execute(<<~GRAPHQL).to_h
{
  nested
}
GRAPHQL

#=> {"data"=>{"nested"=>"nested"}}

pp TestSchema.execute(<<~GRAPHQL).to_h
{
  nested2
}
GRAPHQL

#=> {"data"=>{"nested2"=>nil}}

pp TestSchema.execute(<<~GRAPHQL).to_h
{
  nested, nested2
}
GRAPHQL

#=> {"data"=>{"nested"=>nil, "nested2"=>nil}}

```

**Expected behavior**

- Field `nested` always returns `"nested"`
- Field `nested2` always returns `"nested2"`

**Actual behavior**

- Field `nested` returns `nil` if it's queried with field `nested2`
- Field `nested2` always returns `nil`

rmosolgo:
Fixes #3706

### DIFF
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -6,7 +6,11 @@ module GraphQL
       # Called by {Dataloader} to prepare the {Source}'s internal state
       # @api private
       def setup(dataloader)
+        # These keys have been requested but haven't been fetched yet
         @pending_keys = []
+        # These keys have been passed to `fetch` but haven't been finished yet
+        @fetching_keys = []
+        # { key => result }
         @results = {}
         @dataloader = dataloader
       end
@@ -64,7 +68,17 @@ module GraphQL
       # Then run the batch and update the cache.
       # @return [void]
       def sync
+        pending_keys = @pending_keys.dup
         @dataloader.yield
+        iterations = 0
+        while pending_keys.any? { |k| !@results.key?(k) }
+          iterations += 1
+          if iterations > 1000
+            raise "#{self.class}#sync tried 1000 times to load pending keys (#{pending_keys}), but they still weren't loaded. There is likely a circular dependency."
+          end
+          @dataloader.yield
+        end
+        nil
       end
 
       # @return [Boolean] True if this source has any pending requests for data.
@@ -76,17 +90,24 @@ module GraphQL
       # @api private
       # @return [void]
       def run_pending_keys
+        if @fetching_keys.any?
+          @pending_keys -= @fetching_keys
+        end
         return if @pending_keys.empty?
         fetch_keys = @pending_keys.uniq
+        @fetching_keys.concat(fetch_keys)
         @pending_keys = []
         results = fetch(fetch_keys)
         fetch_keys.each_with_index do |key, idx|
           @results[key] = results[idx]
         end
+        nil
       rescue StandardError => error
         fetch_keys.each { |key| @results[key] = error }
       ensure
-        nil
+        if fetch_keys
+          @fetching_keys -= fetch_keys
+        end
       end
 
       # These arguments are given to `dataloader.with(source_class, ...)`. The object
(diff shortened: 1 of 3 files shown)


lib/graphql/dataloader/source.rb
```diff
@@ -76,17 +90,24 @@ def pending?
       # @api private
       # @return [void]
       def run_pending_keys
+        if @fetching_keys.any?
```
mtsmfm, line 93: `[false, nil].any?` returns `false`.
I think It should be `unless @fetching_keys.empty?` if dataloader accepts `false` or `nil` as a key.

lib/graphql/dataloader/source.rb
```diff
@@ -76,17 +90,24 @@ def pending?
...
         results = fetch(fetch_keys)
         fetch_keys.each_with_index do |key, idx|
           @results[key] = results[idx]
         end
+        nil
       rescue StandardError => error
         fetch_keys.each { |key| @results[key] = error }
       ensure
-        nil
+        if fetch_keys
```
mtsmfm, line 108: It seems `fetch_keys` is always an instance of Array because it's the result of  `@pending_keys.uniq` and never falsy

mtsmfm:
The same test is added though, I also confirmed the code I reported works well with this branch 👍 

I can't wait 1.12.20!

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'graphql', github: 'rmosolgo/graphql-ruby', branch: 'dataloader-source-interdependency-fix'
end

class Echo < GraphQL::Dataloader::Source
  def fetch(keys)
    keys
  end
end

class Nested < GraphQL::Dataloader::Source
  def fetch(keys)
    dataloader.with(Echo).load_all(keys).tap do
    end
  end
end

class Nested2 < GraphQL::Dataloader::Source
  def fetch(keys)
    dataloader.with(Nested).load_all(keys).tap do
    end
  end
end

  field :nested, String, null: true
  field :nested2, String, null: true

  def nested
    dataloader.with(Nested).load("nested")
  end
  
  def nested2
    dataloader.with(Nested2).load("nested2")
  end
end

  query QueryType
  use GraphQL::Dataloader
end

pp TestSchema.execute(<<~GRAPHQL).to_h
{
  nested
}
GRAPHQL

#=> {"data"=>{"nested"=>"nested"}}

pp TestSchema.execute(<<~GRAPHQL).to_h
{
  nested2
}
GRAPHQL

#=> {"data"=>{"nested2"=>"nested2"}}

pp TestSchema.execute(<<~GRAPHQL).to_h
{
  nested, nested2
}
GRAPHQL

#=> {"data"=>{"nested"=>"nested", "nested2"=>"nested2"}}
```

lib/graphql/dataloader/source.rb
rmosolgo: Ahhh thank you... I _always_ miss this one 🤦

lib/graphql/dataloader/source.rb
mtsmfm: Ah, yes. You're right. Never mind, sorry.

PR #3710, opened 2021-11-17
Improve uuid and type_error implementations in the Rails generator and docs

rmosolgo:
I think this is the best way to address a couple issues:

- Recommend a better UUID implementation in the docs and generators 
- Recommend a better type_error implementation in the generator 

Fixes https://github.com/rmosolgo/graphql-ruby/pull/2106
Fixes https://github.com/rmosolgo/graphql-ruby/pull/3681

### DIFF
--- a/spec/integration/rails/generators/graphql/install_generator_spec.rb
+++ b/spec/integration/rails/generators/graphql/install_generator_spec.rb
@@ -287,8 +287,7 @@ class GraphqlController < ApplicationController
 end
 RUBY
 
-  EXPECTED_RELAY_BATCH_SCHEMA = <<-RUBY
-class DummySchema < GraphQL::Schema
+  EXPECTED_RELAY_BATCH_SCHEMA = 'class DummySchema < GraphQL::Schema
   mutation(Types::MutationType)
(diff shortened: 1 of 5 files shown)


spec/integration/rails/generators/graphql/install_generator_spec.rb
```diff
@@ -45,10 +45,22 @@ class DummySchema < GraphQL::Schema
   mutation(Types::MutationType)
   query(Types::QueryType)
 
+  # For batch-loading (see https://graphql-ruby.org/dataloader/overview.html)
+  use GraphQL::Dataloader
+
+  # GraphQL-Ruby calls this when something goes wrong while running a query:
+  def self.type_error(err)
```
hernanat, line 52: was the intent to remote the `ctx` parameter from `type_error`? the call to `super` results in an error due to the missing parameter for me (running v1.12.20)

rmosolgo: Just an oversight on my part 😖 getting a fix here: https://github.com/rmosolgo/graphql-ruby/pull/3722

PR #3712, opened 2021-11-18
Remove @client fields before syncing

rmosolgo:
Oops, if you're using Apollo's client fields (https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies/#querying), then `sync` won't work right, because those fields don't exist on the server. So, strip them out before sending the operation to `sync`. (Then, when the server runs the operation by ID, it will execute the operation _without_ `@client` fields.)


rmosolgo:
:ship: in 1.10.3!

Issue #3189, opened 2020-10-09
Clients are receiving a subscription response even with context.skip


bbugh:
**Describe the bug**

Client callbacks for subscriptions are being called for non-returning subscription methods, even when `context.skip` is used. 

**Versions**

`graphql` version: 1.11.5
`rails`: 6.0.3.2
`graphql-batch`: 0.4.3
`graphql-fragment_cache`: 1.0.3
`actioncable`: 5.2.4-4

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok).
Are you using [interpreter](https://graphql-ruby.org/queries/interpreter.html)? Any custom instrumentation, etc?

```ruby
class GraphSchema < GraphQL::Schema
  use GraphQL::Execution::Interpreter
  use GraphQL::Analysis::AST
  use GraphQL::Subscriptions::ActionCableSubscriptions
end
```

```ruby
class Types::SubscriptionType < Types::BaseObject
  field :project_added, Types::ProjectType, "XXX", null: false # or true, neither work
  field :notification, Types::NotificationType, "XXX", null: true, subscription_scope: :current_user

  def project_added
    if context.query.subscription_update?
      object
    else
      context.skip
    end
  end

  def notification
    context.skip
  end
end
```

**GraphQL query**

The ActionCable websocket request:

```json
{
  "command": "message",
  "identifier": "{\"channel\":\"GraphqlChannel\",\"channelId\":\"1750efbc531\"}",
  "data": "{\"query\":\"subscription projectAdded {\\n  projectAdded {\\n    id\\n    name\\n    clientName\\n    status\\n    __typename\\n  }\\n}\\n\",\"variables\":{},\"operationName\":\"name\",\"action\":\"execute\"}"
}
```

The query from `data` (no variables):

```gql
subscription projectAdded {
  projectAdded {
    id
    name
    clientName
    status
    __typename
  }
}
```

The ActionCable websocket response

```json
{
  "identifier": "{\"channel\":\"GraphqlChannel\",\"channelId\":\"1750efbc531\"}",
  "message": {
    "result": {
      "data": {}
    },
    "more": true
  }
}
```

**Steps to reproduce**

Not sure other than the provided code.

**Expected behavior**

The `context.skip` should prevent the client from receiving and acting on a message. 

**Actual behavior**

The client is receiving the message and running code that it should not be running. For example, the `notification` subscription tries to run the code for when a notification subscription is received, but no data arrives:

[...]

PR #3713, opened 2021-11-18
Support NO_UPDATE in ActionCable subscriptions

jqr:
Looks like this may be the fix for #3189 and similar empty subscription updates I'm seeing in my application.

Applies the same pattern to `ActionCableSubscriptions` as `Subscriptions`, which just checks if `execute_update` returns `nil` and skips the delivery. See https://github.com/rmosolgo/graphql-ruby/blob/bace1e4027900fc8779a5c2fd393ff6456046cea/lib/graphql/subscriptions.rb#L141-L146

Given this is a subclass, I'm guessing there's a better way to rely on the superclass behavior and ensure all subclasses inherit this important feature. Open to suggestions.


rmosolgo:
Sorry for my slow response on this. Looks good to me! I think putting this check in the subclass is fine. In practice, different subscription implementations have to manage their own storage and transport, so this check would go somewhere different in each one.

PR #3721, opened 2021-11-21
Remove extend GraphQL::Subscriptions::SubscriptionRoot from docs

DmitryTsepelev:
Looks like we don't need it anymore—everyone is on the Interpreter nowdays and this `extend` is added by default and we [warn](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions/subscription_root.rb#L9) when user does it.


rmosolgo:
Thanks! I guess I searched the code to clean it up, but not the docs 🤦

PR #3722, opened 2021-11-22
Fix generated type_error hook signature

rmosolgo:
Oops, I implemented the wrong thing in #3710 🤦


lib/generators/graphql/templates/schema.erb
```diff
@@ -9,7 +9,7 @@ class <%= schema_name %> < GraphQL::Schema
   use GraphQL::Dataloader
 <% end %>
   # GraphQL-Ruby calls this when something goes wrong while running a query:
-  def self.type_error(err)
+  def self.type_error(err, context)
```
hernanat, line 12: 🙏

hernanat (approved):

Issue #3723, opened 2021-11-22
:allow_blank with validate :format appears broken


joeosburn:
The `allow_blank: true` option uses with format validation appears broken. I am assuming it has something to do with the recent https://github.com/rmosolgo/graphql-ruby/pull/3671 change. I moved the `allow_blank` option to be a top level validator. I have a `postalCode` argument being validated this way:

```
argument :postal_code, String, required: false, validates: { format: { with: /[a-zA-Z\d\s\-]+/ }, allow_blank: true }
```

However, if I pass in an empty string as the argument value, I get a `postalCode is invalid` error. This was working fine with graphql 1.12.17.

**Versions**

`graphql` version: 1.12.20
ruby 3.0.2p107

rmosolgo:
Thanks for the bug report -- I agree that it should work the way you expect (and the way it used to work 😅 ). I'll take a look soon and follow up here!

butsjoh:
I am experiencing a similar issue with the length validator.

before the version bump: (i was on 1.12.16)

```
      argument :name, String, required: true, validates: {
        length: { allow_blank: false, maximum: 255 }
      }
```

We you gave an empty string "" it would throw a validation error.

In the new version it does not so i tried to adapt to:

```
      argument :name, String, required: true, validates: {
        length: { maximum: 255 }, allow_blank: false
      }
```

Still no validation error on blank string. Is this expected? I have the impression that multiple validations are consider || and not & and i don't see that in the documentation. Am i misunderstanding something?

PR #3726, opened 2021-11-23
Fix allow_blank: true with format: /.../

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3723


rmosolgo:
For the record I checked the other validators, they already had this check in them. i guess I just missed this one!

Issue #1812, opened 2018-08-27
Schema Stitching


audiolion:
Hey 👋 

Does the graphql-ruby gem provide support for schema stitching?

rmosolgo:
No, but I recommend: https://github.com/d12/graphql-remote_loader

It's a nice approach to consuming one GraphQL API from within another.

gmac:
There’s now a Ruby implementation of schema stitching at https://github.com/gmac/graphql-stitching-ruby.

PR #3728, opened 2021-11-24
get full sub-query for a field to proxy to external API

judocode:
## Use case
In order to proxy a top-level query 1:1 to a remote API, there currently is not a great solution. In https://github.com/rmosolgo/graphql-ruby/issues/1812, [graphql-remote_loader](https://github.com/d12/graphql-remote_loader) was suggested as a potential solution, but it is a bit verbose as it is on a field-by-field basis and it also doesn't seem to have the capabilities to proxy through a sub-query.

Say for example, we have a query:
```
user(id: 1) {
  name
  address
}
search(term: "ski resorts") {
  results
}
```

And we want to proxy the user query to a user API:
```ruby
def user(id:)
  # The only query string accessible seems to be the full one with user and search
  UserAPI.query(context.query.query_string)
end
```

Then the only way with current functionality is to send the entire query through or do some manual manipulation on it to properly get it proxied.

## Potential solution
If each individual resolver had access to the sub-query that is specific to it via an interface such as `context[:current_query]`, then in the example, we could have access to the user-specific sub-query that we could pass through to the UserAPI:
```
user(id: 1) {
  name
  address
}
```
I assume this can potentially cause a performance hit, if so then this could be behind a flag defined on the field, such as `show_current_query`. If performance is not a concern, then that portion is unnecessary.

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -354,6 +354,10 @@ module GraphQL
           finished_jobs = 0
           enqueued_jobs = gathered_selections.size
           gathered_selections.each do |result_name, field_ast_nodes_or_ast_node|
+            if @context[:current_object].show_current_query
+              @context[:current_query] = field_ast_nodes_or_ast_node.to_query_string
+            end
+
             @dataloader.append_job {
               evaluate_selection(
                 path, result_name, field_ast_nodes_or_ast_node, scoped_context, owner_object, owner_type, is_eager_selection, selections_result, parent_object
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -54,6 +54,21 @@ module GraphQL
           end
         end
 
+        # Whether or not :current_query should be available.
+        # Call this method to provide a new show_current_query; OR
+        # call it without an argument to get the show_current_query
+        # @param new_show_current_query [Boolean]
+        # @return [Boolean]
+        def show_current_query(new_show_current_query = nil)
+          if new_show_current_query
+            @show_current_query = new_show_current_query
+          elsif defined?(@show_current_query)
+            @show_current_query
+          else
+            nil
+          end
+        end
+
         # This pushes some configurations _down_ the inheritance tree,
         # in order to prevent repetitive lookups at runtime.
         module ConfigurationExtension


rmosolgo:
Hi! I'd love to learn the use-case of this. Could you please update the PR description to include why this would be useful and how you'd like to use it?

judocode:
@rmosolgo Thanks for your quick reply! I apologize for the pre-emptive PR as I am still working through a solution on it (doesn't currently work as-is). I have updated the PR description describing the use-case and a potential solution to it.

rmosolgo:
I think you could accomplish this with `extras: [:ast_node]` (doc: https://graphql-ruby.org/fields/introduction.html#extra-field-metadata)

Here's an example script:

```ruby
require "bundler/inline"

gemfile do
  gem "graphql", "1.13.0"
end

  class User < GraphQL::Schema::Object
    field :name, String
    field :address, String
  end

    field :user, User, extras: [:ast_node] do             # << Add `extras: [:ast_node]` to the field definition
      argument :id, ID
    end

    def user(id:, ast_node:)                              # << And add `ast_node:` to the resolver, to receive the value
      puts "AST Node:"
      puts "------"
      puts ast_node.to_query_string
      puts "------"
      {name: "Mickey Mouse", address: "Disney World"}
    end
  end

  query(Query)
end


pp Schema.execute("{ user(id: 1) { name address } }").to_h
```

Which outputs:

```
$ ruby ast_node_test.rb
AST Node:
------
user(id: 1) {
  name
  address
}
------
{"data"=>{"user"=>{"name"=>"Mickey Mouse", "address"=>"Disney World"}}}
```

What do you think of that?

judocode:
@rmosolgo Wow that is brilliant and I am blown away with how quick your reply was and how thoughtful and thorough your response was! Thanks so much! I'm not sure if this deserves its own section in the docs, but I feel like this specific use-case of this feature would be great to document and I'd be more than happy to help with that.

rmosolgo:
Glad it helped! Sure, the only mention of `extras: [:ast_node]` is on that doc I linked, and it's not very detailed. There's an "Edit" link at the bottom of that page -- please improve it if you'd like!

judocode:
@rmosolgo Excellent! Oh and just one more thing I noticed:

Let's say you are using variables and the queries are wrapped like:
```graphql
query ($id: Int!, $term: String!) {
  user(id: $id) {
    name
    address
  }
  search(term: $term) {
    results
  }
}
```

And you use the `ast_node.to_query_string`, which outputs:
```graphql
user(id: $id) {
  name
  address
}
```

This would fail if you sent it through because the `$id` var is not declared. Is there a currently a good solution for this?

Thanks again!

rmosolgo:
No, unfortunately there's nothing out-of-the-box to rewrite queries in the way you have in mind. You could use an AST visitor (https://graphql-ruby.org/language_tools/visitor.html) to inspect the `ast_node` and gather any variable names encountered, then look them up in `context.query.variables` to pass them along in the rewritten query.

Issue #3740, opened 2021-11-30
Empty description of a type causes `GraphQL::SchemaComparator.compare` to fail with a frozen string error


joelzwarrington:
**Describe the bug**

A type with an empty description causes `GraphQL::SchemaComparator.compare` to fail

**Versions**

| Gem       | Version |
|-----------|---------|
| `graphql` | 1.13.0  |
| `rails`   | 6.0     |

**GraphQL schema**

```ruby
class QueryType < GraphQL::Schema::Object
  description ""

  field :anything, String
end

class Schema < GraphQL::Schema
  query QueryType
end
```

or the equivalent definition:

```
"\"\"\"\n\"\"\"\ntype Query {\n  anything: String\n}\n"
```

**Steps to reproduce**

If you define the above schema, and perform the following:

```ruby
GraphQL::SchemaComparator.compare(Schema.to_definition, Schema.to_definition)
```

or equivalent:

```ruby
definition = "\"\"\"\n\"\"\"\ntype Query {\n  anything: String\n}\n"
GraphQL::SchemaComparator.compare(definition, definition)
```

**Expected behavior**

The schema is compared and I am provided with the returned `GraphQL::SchemaComparator::Result` class **OR** I am provided with a warning of the type containing the empty description.

**Actual behavior**

I am provided with a frozen error string with limited stack trace, so it is hard to debug the issue.

Place full backtrace here (if a Ruby exception is involved):

<details>
<summary>Click to view exception backtrace</summary>

```
FrozenError: can't modify frozen String: ""
from /Users/joelwarrington/.rvm/gems/ruby-2.7.2/gems/graphql-1.13.0/lib/graphql/language/lexer.rb:15:in `gsub!'
```

</details>

**Additional context**

Augmented the base object to include additional meta information in the description of a type. When this was removed, it unexpectedly broke a type which had an empty string defined as the description.

The cause appears to be [GraphQL::Language::BlockString.trim_whitespace](https://github.com/rmosolgo/graphql-ruby/blob/489267e0fb63651bef0579dd1ced839a9a6c2ebb/lib/graphql/language/lexer.rb#L1427) which will return a frozen string, causing the Lexer to fail at [GraphQL::Language::Lexer.replace_escaped_characters_in_place](https://github.com/rmosolgo/graphql-ruby/blob/489267e0fb63651bef0579dd1ced839a9a6c2ebb/lib/graphql/language/lexer.rb#L1440).

PR #3741, opened 2021-11-30
fix(Lexer): resolved issue when rebuilding trimmed whitespace string

joelzwarrington:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3740

When rebuilding an empty trimmed whitespace string, such as `"\n"`, which can occur when a type has an empty description, a frozen string is returned, causing in place substitution with gsub to fail when replacing escaped characters in place at [GraphQL::Language::Lexer.replace_escaped_characters_in_place](https://github.com/rmosolgo/graphql-ruby/blob/489267e0fb63651bef0579dd1ced839a9a6c2ebb/lib/graphql/language/lexer.rb#L1440).

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -1437,9 +1437,6 @@ meta[:col],
 meta[:previous_token],
 )
 else
-# As the string is modified in place below, duping the value will provide a non-frozen string to modify.
-# `GraphQL::Language::BlockString.trim_whitespace`, can return frozen strings.
-value = value.dup if value.frozen?
 replace_escaped_characters_in_place(value)
 
 if !value.valid_encoding?
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -1437,6 +1437,9 @@ meta[:col],
 meta[:previous_token],
 )
 else
+# As the string is modified in place below, duping the value will provide a non-frozen string to modify.
+# `GraphQL::Language::BlockString.trim_whitespace`, can return frozen strings.
+value = value.dup if value.frozen?
 replace_escaped_characters_in_place(value)
 
 if !value.valid_encoding?
--- a/lib/graphql/language/block_string.rb
+++ b/lib/graphql/language/block_string.rb
@@ -55,7 +55,7 @@ module GraphQL
         end
 
         # Rebuild the string
-        lines.size > 1 ? lines.join("\n") : (lines.first || "")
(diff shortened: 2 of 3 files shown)


rmosolgo:
Thanks for taking a look at this! Are you able to confirm that it fixed your problem? (I see that elsewhere in the method, it has `return ""` which would probably produce the same kind of bug.)

I'd love to get a test in the project to make sure we don't regress. Are you interested in adding it? I seems like something like: 

```ruby 
schema = GraphQL::Schema.from_definition <<-GRAPHQL
"""
"""
type Query {
  f1: Int
}

refute schema.get_type("Query").description.frozen?
```

would do the job, if I understood the bug correctly.

joelzwarrington:
@rmosolgo, I can confirm this specific fix did resolve the issue for me, I made the change very specifically to what I encountered when stepping through the code.

If you're thinking of making it so that every string is not frozen, should I instead dup things if it's frozen before the replace escaped characters happens?

Where would you suggest adding a spec test for this?

joelzwarrington:
@rmosolgo I think this is ready for review, however not familiar with the checks which are failing, maybe you could help and have a look?

lib/graphql/language/lexer.rb
```diff
@@ -1437,6 +1437,9 @@ def self.emit_string(ts, te, meta, block:)
 meta[:previous_token],
 )
 else
+# As the string is modified in place below, duping the value will provide a non-frozen string to modify.
+# `GraphQL::Language::BlockString.trim_whitespace`, can return frozen strings.
+value = value.dup if value.frozen?
```
joelzwarrington, line 1442: The only thing I'm worried about is it appears in documentation that Lexer is automatically created, would this override what I've defined?

rmosolgo: That's right -- this file is generated by `lexer.rl`, which is a pain to generate (and getting harder .... Ragel has a new dependency, and its new version generates a slower state machine than the old version 😖 ) 

IMO, better to modify `BlockString.trim_whitespace` for that reason. Was your fix there sufficient to solve your problem?

joelzwarrington: Sounds good @rmosolgo, I'll modify BlockString, and `dup` any instances of frozen strings, as it did solve my problem.

As a note, I did actually resolve the issue by updating the description so it wasn't empty, or by removing it, but thought that this might be something others encounter in the future.

rmosolgo:
Thanks for digging in on this fix! 

I haven't been able to get that truffle-ruby test passing and I don't have the heart to yank it completely :(

PR #3751, opened 2021-12-02
Add an after-define hook to field extensions

rmosolgo:
This would allow field extensions to define arguments on fields _without_ overriding something from the field's definition block. 

cc @cocoahero 

TODO: 

- [x] Name 
- [x] Docs
- [x] Remove breaking change to ConnectionExtension

### DIFF
--- a/guides/type_definitions/field_extensions.md
+++ b/guides/type_definitions/field_extensions.md
@@ -48,6 +48,8 @@ end
 
 This way, an extension can encapsulate a behavior requiring several configuration options.
 
+### TODO document after-hook
+
 ### Modifying field execution
 
 Extensions have two hooks that wrap field resolution. Since GraphQL-Ruby supports deferred execution, these hooks _might not_ be called back-to-back.
--- a/guides/type_definitions/field_extensions.md
+++ b/guides/type_definitions/field_extensions.md
@@ -48,7 +48,19 @@ end
 
 This way, an extension can encapsulate a behavior requiring several configuration options.
 
-### TODO document after-hook
+### Adding default argument configurations
+
+Extensions may provide _default_ argument configurations which are applied if the field doesn't define the argument for itself. The configuration is passed to {{ Schema::FieldExtension.default_argument | api_doc }}. For example, to define a `:query` argument if the field doesn't already have one:
+
+```ruby
+class SearchableExtension < GraphQL::Schema::FieldExtension
+  # Any field which uses this extension and _doesn't_ define
+  # its own `:query` argument will get an argument configured with this:
+  default_argument(:query, String, required: false, description: "A search query")
+end
+```
+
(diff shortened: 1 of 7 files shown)


cocoahero:
I think this is a good solution. My vote for naming would be `after_define`.

It does get a bit tedious to check for each arguments presence before applying the change, but I think for this particular edge case its fine. Perhaps a future improvement would be adding a class level `default_argument` DSL that would behind the scenes automatically apply the arguments in the `after_define` callback while checking for collisions. Example:

```ruby
  default_argument :first, Integer, required: false

  def after_define
    # this would happen in the super impl
    all_keywords = field.all_argument_definitions.map(&:keyword)
    all_keywords.uniq!
    self.default_argument_defs.each do |arg_def|
      next if all_keywords.include?(arg.keyword)
      field.add_argument(arg_def)
    end
  end
```

swalkinshaw:
Exposing the original arguments and the _added_ arguments would be useful overall. We already have a low-level extension which has to do a similar thing manually:

```ruby
Class MyExtension
  def apply
    argument_names = field.arguments.keys
    # some field.add_argument calls
    added_arguments = field.arguments.keys - argument_names

    options[:added_arguments] = added_arguments.map do |key|
      key.underscore.to_sym
    end
  end

  def resolve(object:, arguments:, context:)
    next_args = arguments.dup
    arguments = arguments.except(*options[:added_arguments])

    yield(object, arguments, next_args)
  end

  def after_resolve(value:, memo:, context:, **_args)
    arguments = memo
  end
end
```

rmosolgo:
Thanks for taking a look. I: 

- renamed it `after_define`
- added `.default_argument` and `#added_default_argument` for a workflow like the one in ConnectionExtension 
- Made ConnectionExtension only delete arguments that it added by default -- this is actually a breaking change, I should revert it. But I wanted to make sure it worked.

What do you think of that?

cocoahero:
This feels pretty good to me. I'll play with this branch in our core codebase today to see if its unblocks our upgrade.

cocoahero:
I noticed a slight issue with the current implementation where extensions added to the field after the definition block would not have their `after_define_apply` method called, and therefore default arguments and the `after_define` hook would not be used. For example:

```ruby
def custom_field_def_helper
  field_def = field(name, return_type, &block)
  field_def.extension(...) # This extension's `apply` would be called, but not `after_define_apply`.
end
```

This may be an unsupported scenario, but something that came up with our current usage. Completely possible that its a foot-gun situation 😅

rmosolgo:
It certainly _used_ to work, and I agree it _should_. But I don't know a better approach than to call both `apply` and `after_define` right away in that case.

rmosolgo:
Updated along those lines in a242d60d4 -- let me know if you see any red flags, otherwise I'm hoping to 🚢  1.13.1 this afternoon.

PR #3762, opened 2021-12-08
Validation perf from 1.12

rmosolgo:
Porting #3761 to 1.13.x


rmosolgo:
😩 Yay, new CI fails on Rails master.

PR #3763, opened 2021-12-08
📝   Add Chime to index page

danielvdao:
### What

[Chime](https://www.chime.com/) is a big user and fan of the graphql-ruby gem. We appreciate the support from you, Robert (from answering all my questions and past issues concerning class-based migrations 💚 ), and the graphQL community and would like to add our name onto this list 🙂


danielvdao:
@rmosolgo for some reason I can't tag you on this PR - but was hoping this is a pretty quick add!

rmosolgo:
cool, thanks!

PR #3774, opened 2021-12-10
Handle Rails 7 changes to ActiveSupport::TimeWithZone.name

netshade:
This PR addresses an upcoming deprecation issue in Rails 7 when changing the following framework default:

```
Rails.application.config.active_support.remove_deprecated_time_with_zone_name = true
```

Prior to this setting being changed in Rails 7 ( and prior to Rails 7 at all ), `ActiveSupport::TimeWithZone.name` would return `Time`. This allowed for `ActiveSupport::TimeWithZone` objects to be received by the deserializer and treated as `Time` objects without `graphql-ruby` being *aware* of what they actually were. 

When this new setting has been changed, prior to this PR, these objects will be emitted in a form like:

```
"{\"__timestamp__\":[\"ActiveSupport::TimeWithZone\",\"2021-12-10 15:08:18.177738000-0500\"]}"
```

which, when deserialized later on by `GraphQL::Subscriptions::Serialize#load_value`, will attempt to call `strptime` on the class `ActiveSupport::TimeWithZone` ( which doesn't exist ).

In order to prevent this from happening, the serialize code now checks the class name string at serialize time, and transforms it into a `Time` if it is the suspect class, which allows the deserialize code to continue working as it did prior to this Rails 7 change.


netshade:
This PR primarily presented as a proposal. It is definitely not ideal that `graphql-ruby` has to be "aware" of this change in `ActiveSupport`, but it may be practically necessary. Please feel free to close outright if this is too much awareness of what is happening inside Rails to match what you'd like.

netshade:
I made some changes to this to:

* Better fit how `graphql-ruby` uses `defined?` w/ external library awareness
* Better test how Rails handles the removal of the `name` facade in their upcoming deprecation

lib/graphql/subscriptions/serialize.rb
```diff
@@ -123,6 +123,17 @@ def dump_value(obj)
...
+            # This handles a case where Rails prior to 7 would
+            # make the class ActiveSupport::TimeWithZone return "Time" for
+            # its name. In Rails 7, it will now return "ActiveSupport::TimeWithZone",
+            # which happens to be incompatible with expectations we have
+            # with what a Time class supports ( notably, strptime in `load_value` ).
+            #
+            # This reverts to what was happening before, in that ActiveSupport::TimeWithZone
+            # was coerced to a Time on the deserialization side.
+            #
+            { TIMESTAMP_KEY => [Time.name, obj.strftime(TIMESTAMP_FORMAT)] }
```
rmosolgo, line 136: Will this cause `TimeWithZone`s to be serialized identical to `Time`s, then when they're reloaded, they'll be loaded as `Time`s instead of `TimeWithZone`s? 

I guess that's what _used_ to happen, but why not instead, improve it to serialize and load in such a way to reload `TimeWithZone`s?

netshade: Sure, I made that change - it requires adding awareness on the deserialize side of the incoming zone, therefore, the format of the arguments when passing a `TimeWithZone` must be changed.

rmosolgo:
Thanks again for the improvement!

PR #3776, opened 2021-12-11
Abstract out notifications engine for tracing

ianks:
It would be helpful to be able to use `dry-monitor` for a notifications engine for non-Rails apps. This PR adds support for that by abstracting out a new `NotificationsTracer` interface for that.

### Example

```ruby
$notifications_engine = Dry::Monitor::Notifications.new(:my_app)

  tracer(GraphQL::Tracing::NotificationsTracing.new($notifications_engine))
end
```

### Comments

I took care to make it so `ActiveSupportNotificationsTracing` uses this tracer under the hood in a backwards compatible way, but let me know if you would like me to change it at all.

### DIFF
--- a/lib/graphql/tracing/notifications_tracing.rb
+++ b/lib/graphql/tracing/notifications_tracing.rb
@@ -27,11 +27,6 @@ module GraphQL
 
       MAX_KEYS_SIZE = 100
 
-      # Returns all the possible event names for this tracing implementation
-      def self.event_names
-        KEYS.values
-      end
-
       # Initialize a new NotificationsTracing instance
       #
       # @param [Object] notifications_engine The notifications engine to use
--- a/lib/graphql/tracing/notifications_tracing.rb
+++ b/lib/graphql/tracing/notifications_tracing.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+module GraphQL
+  module Tracing
+    # This implementation forwards events to a notification handler (i.e.
+    # ActiveSupport::Notifications or Dry::Monitor::Notifications)
+    # with a `graphql` suffix.
+    #
+    # @see KEYS for event names
+    class NotificationsTracing
+      # A cache of frequently-used keys to avoid needless string allocations
+      KEYS = {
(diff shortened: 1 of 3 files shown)


lib/graphql/tracing/notifications_tracing.rb
```diff
@@ -0,0 +1,64 @@
...
+      # . tracer.trace("lex") { ... }
+      #
+      # @param [string] key The key for the event
+      # @param [Hash] metadata The metadata for the event
+      # @yield The block to execute for the event
+      def trace(key, metadata, &blk)
+        prefixed_key = KEYS[key] || "#{key}.graphql"
+
+        # Cache the new keys while making sure not to induce a memory leak
+        if KEYS.size < MAX_KEYS_SIZE
```
ianks, line 56: @rmosolgo This is a new enhancement to avoid more string allocations. It shouldn't cause any issues, but wanted to flag it for review.

ianks:
FYI: It looks like the test failures are unrelated due to new Rails versions.

rmosolgo:
Thanks for the suggestion here. I'm game to support this library, but I might shuffle things around a bit, implementation-wise 👀

lib/graphql/tracing/notifications_tracing.rb
```diff
@@ -0,0 +1,64 @@
...
+        "resolve_type" => "resolve_type.graphql",
+        "resolve_type_lazy" => "resolve_type.graphql",
+      }
+
+      MAX_KEYS_SIZE = 100
+
+      # Returns all the possible event names for this tracing implementation
+      def self.event_names
+        KEYS.values
+      end
```
rmosolgo, line 33: Is this used anywhere? How about removing it?

lib/graphql/tracing/notifications_tracing.rb
rmosolgo: fair enough :+1:

rmosolgo (approved):
Just one question :+1:

ianks:
@rmosolgo Removed that method 😄

PR #3778, opened 2021-12-13
Platform Tracing: fallback transaction name

### LINKED ISSUE #3769: Tracing: Resource names for queries without operation names
I am happy to contribute this change btw, but thought I would start from an issue before whipping up a PR.

Datadog APM groups GraphQL traces by their operation name, this is fairly fundamental to making those traces useful, similarly to how a Rails controller tracer would not be useful if it combined all controller + actions into a single "controllers" trace.

However, for queries with no operation name associated, that's exactly what GraphQL Ruby does today.

I could have sworn that some years ago the Datadog tracing behaviour in GraphQL Ruby was to fall back to the entire query string (or maybe truncated, but truncated very far along) as the resource name, with newlines removed. Perhaps that's my imagination.

I think that is a reasonable fall back mechanism. Another would be to has the query string, but this is more work and probably worse developer experience as they cannot be skimmed in a list

I assume such a solution just has to change this line a little:
https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/data_dog_tracing.rb#L23
That said, this seems like something all platform tracers would benefit from? Not that I am familiar with any other ones.

I tried faking an operation name in the `MySchema.execute(query, operation_name:)` API, however it reasonably raises an error that no query I am executing has that operation name.

bessey:
Enable a developer to provide the platform trace API a fallback name to
use when a query name is not provided.

Note that this is a *fallback*, the query name is preferred when
available.

Testing is limited to Datadog and New Relic.

Support is within any tracer that delegates to
`PlatformTracing#transaction_name`, which appears to include AppSignal
and Scout. Not sure if that is exhaustive.

Fixes #3769

### DIFF
--- a/lib/graphql/tracing/platform_tracing.rb
+++ b/lib/graphql/tracing/platform_tracing.rb
@@ -104,17 +104,22 @@ module GraphQL
 
       private
 
-      # Get the transaction name based on the operation type and name
+      # Get the transaction name based on the operation type and name if possible, or fall back to a user provided
+      # one. Useful for anonymous queries.
       def transaction_name(query)
         selected_op = query.selected_operation
-        if selected_op
+        txn_name = if selected_op
           op_type = selected_op.operation_type
-          op_name = selected_op.name || "anonymous"
+          op_name = selected_op.name || fallback_transaction_name(query.context) || "anonymous"
+          "#{op_type}.#{op_name}"
         else
-          op_type = "query"
-          op_name = "anonymous"
+          "query.anonymous"
         end
-        "GraphQL/#{op_type}.#{op_name}"
+        "GraphQL/#{txn_name}"
+      end
+
+      def fallback_transaction_name(context)
+        context[:tracing_fallback_transaction_name]
       end
 
       attr_reader :options
--- a/spec/graphql/tracing/data_dog_tracing_spec.rb
+++ b/spec/graphql/tracing/data_dog_tracing_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe GraphQL::Tracing::DataDogTracing do
+  module DataDogTest
+    class Query < GraphQL::Schema::Object
+      add_field GraphQL::Types::Relay::NodeField
+
+      field :int, Integer, null: false
+
+      def int
+        1
+      end
+    end
+
+    class TestSchema < GraphQL::Schema
+      query(Query)
+      use(GraphQL::Tracing::DataDogTracing)
+    end
+  end
+
+  before do
+    Datadog.clear_all
+  end
+
+  it "falls back to a :tracing_fallback_transaction_name when provided" do
+    DataDogTest::TestSchema.execute("{ int }", context: { tracing_fallback_transaction_name: "Abcd" })
+    assert_equal ["Abcd"], Datadog::SPAN_RESOURCE_NAMES
+  end
+
+  it "does not use the :tracing_fallback_transaction_name if an operation name is present" do
+    DataDogTest::TestSchema.execute(
+      "query Ab { int }",
+      context: { tracing_fallback_transaction_name: "Cd" }
+    )
+    assert_equal ["Ab"], Datadog::SPAN_RESOURCE_NAMES
+  end
+
+  it "does not require a :tracing_fallback_transaction_name even if an operation name is not present" do
+    DataDogTest::TestSchema.execute("{ int }")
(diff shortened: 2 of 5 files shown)


bessey:
@rmosolgo FYI I am finding that the test suite hangs on v1.13 when I run it locally with `rake test`, running with `TESTOPTS="--verbose"` I can see the output before it hangs is:
```
GraphQL::Dataloader::AsyncDataloader::With libev_scheduler#test_0001_runs IO in parallel by default
```
So perhaps I am missing a libev dependency and it doesn't handle it so gracefully.

So for now I am targetting v1.12 (actually, we have not yet upgraded to v1.13 so I would prefer to target v1.12 either way, but we can live on a fork until v1.13 if necessary).

lib/graphql/tracing/platform_tracing.rb
```diff
@@ -104,19 +104,27 @@ def self.use(schema_defn, options = {})
...
         if selected_op
           op_type = selected_op.operation_type
           op_name = selected_op.name || "anonymous"
+        elsif fallback_name = fallback_transaction_name(query.context)
+          return "GraphQL/#{fallback_name}"
         else
           op_type = "query"
           op_name = "anonymous"
         end
         "GraphQL/#{op_type}.#{op_name}"
```
rmosolgo, line 120: sorry, personally, I have a hard time with multi-return methods, because I end up forgetting the different "exits" that the method has. if you don't mind: 

```suggestion
        txn_name = if selected_op
          op_type = selected_op.operation_type
          op_name = selected_op.name || "anonymous"
          "#{op_type}.#{op_name}"
        elsif (fallback_name = fallback_transaction_name(query.context))
          fallback_name
        else
          "query.anonymous"
        end
        "GraphQL/#{txn_name}"
```

rmosolgo (commented):
Looking good so far! If you don't mind adding some test for this behavior, that'd be great. Do you plan to use the `.call` feature? If not, I'd rather leave it out until someone actually needs it. (Just more for me to maintain, otherwise 😅 )

>  it hangs

Oh! I found that GitHub actions failed when libev_scheduler was installed, too. Maybe that was why.

bessey:
> Looking good so far! If you don't mind adding some test for this behavior, that'd be great. Do you plan to use the `.call` feature? If not, I'd rather leave it out until someone actually needs it. (Just more for me to maintain, otherwise sweat_smile )

You made me do what I should have from the beginning: measured :D

I wanted to avoid a SHA / SHA2 digest generation when it wouldn't be used. In practice, this is so close to instant its not worth the complexity

```
irb(main):003:0> string = "ANCASDJASFSDFSDF"*1000
=> "ANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCASDJASFSDFSDFANCA...
irb(main):008:0> Benchmark.measure { Digest::SHA2.hexdigest(string) }.real
=> 0.00020213398966006935
```

If I am not mistaken, that's 0.2ms. Will remove that change. And incorporate your other comment.

> If you don't mind adding some test for this behavior, that'd be great

This is next on my to do, I have left some empty tests for now so we can agree on scope. Do these seem reasonable to you? One for my snowflake implementation in Datadog, and one for the general purpose implementation, in a tracer that relies on that implementation

rmosolgo (commented):
Yep, this scope LGTM 😎

bessey:
I have implemented the specs and got it working on master rather than 1.12, ready for review!

rmosolgo (approved):
Thanks for this improvement!

spec/graphql/tracing/data_dog_tracing_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe GraphQL::Tracing::DataDogTracing do
+  module DataDogTest
+    class Query < GraphQL::Schema::Object
+      add_field GraphQL::Types::Relay::NodeField
```
rmosolgo, line 8:
```suggestion
  include GraphQL::Types::Relay::HasNodeField
```

There was a bug here fixed in master since this branch came off: #3791

rmosolgo: I'll merge this and fix it on master

PR #3780, opened 2021-12-14
Fix: schema generation with camelized default value of input arguments

Eddayy:
Fixed an error when you're trying to generate schema with inputs with arguments names that can be camelized. 
 
I assume my fix is the proper method, as the default value expects the original keyword when writing the code.

The bottom example will fail when generating schema. 
```ruby
class ExampleInput < Types::BaseInputObject
  argument :updated_at, String, required: false
end

class ExampleQuery < BaseQuery
  argument :order_by, ExampleInput, required: false, default_value: { updated_at: 'DESC' }
end
```

### DIFF
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -196,9 +196,13 @@ module GraphQL
         when "INPUT_OBJECT"
           GraphQL::Language::Nodes::InputObject.new(
             arguments: default_value.to_h.map do |arg_name, arg_value|
-              arg = @warden.arguments(type).find { |a| a.keyword.to_s == arg_name.to_s }
+              args = @warden.arguments(type)
+              arg = args.find { |a| a.keyword.to_s == arg_name.to_s }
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hi, thanks for the fix here! It looks great. I'd just like to include a test so we can make sure this doesn't get broken in the future. Do you mind trying to add one? 

here's an existing test which _loads_ a schema from a schema dump then asserts that the printed schema dump is equivalent: 

https://github.com/rmosolgo/graphql-ruby/blob/097890bc6eaa2cd308d24be083afba9fa7a3c73c/spec/graphql/schema/build_from_definition_spec.rb#L47-L59

I think you could add a camel-cased argument to `input InputObject { ... }` and then to the default value (`= { a: 1, ... }`). IIUC, that would fail _without_ this patch, but pass after this patch is included.

rmosolgo:
Ahh, right, that wasn't a good suggestion for a test :S. I added one here and I'll merge if it passes ok 🤞 Thanks again for the fix!

PR #3783, opened 2021-12-15
Pass arguments object instead of plain Hash

rmosolgo:
Oops, this code path passes a plain hash instead of the `Interpreter::Arguments` instance. It ends up getting assigned to `context[:current_arguments]` which fails down the line.

TODO: 

- [x] Some test

### DIFF
--- a/spec/graphql/execution/interpreter_spec.rb
+++ b/spec/graphql/execution/interpreter_spec.rb
@@ -268,6 +268,20 @@ describe GraphQL::Execution::Interpreter do
       def self.resolve_type(type, obj, ctx)
         FieldCounter
       end
+
+      class EnsureArgsAreObject
+        def self.trace(event, data)
+          case event
+          when "execute_field", "execute_field_lazy"
+            args = data[:query].context[:current_arguments]
+            if !args.is_a?(GraphQL::Execution::Interpreter::Arguments)
+              raise "Expected arguments object, got #{args.class}: #{args.inspect}"
+            end
+          end
+          yield
+        end
+      end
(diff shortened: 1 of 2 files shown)


spec/graphql/execution/interpreter_spec.rb
```diff
@@ -268,6 +268,20 @@ def self.object_from_id(id, ctx)
...
+          when "execute_field", "execute_field_lazy"
+            args = data[:query].context[:current_arguments]
+            if !args.is_a?(GraphQL::Execution::Interpreter::Arguments)
+              raise "Expected arguments object, got #{args.class}: #{args.inspect}"
+            end
+          end
+          yield
+        end
+      end
+      tracer EnsureArgsAreObject
```
rmosolgo, line 284: I got this to fail without the patch: 

```
Error:
GraphQL::Execution::Interpreter::null propagation#test_0005_works with union lists that have members of different kinds, with different nullabilities:
RuntimeError: Expected arguments object, got Hash: {:ids=>["RAV", "Dark Confidant"]}
    /Users/rmosolgo/code/graphql-ruby/spec/graphql/execution/interpreter_spec.rb:278:in `trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:83:in `call_tracers'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/tracing.rb:67:in `trace'
    /Users/rmosolgo/code/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:875:in `block (2 levels) in after_lazy'
```

PR #3784, opened 2021-12-15
Add a validator for requiring an argument to be passed, even if it's null

### LINKED ISSUE #3700: `required: :nullable` argument definition
### The problem

[GraphQL Null Input Value spec](https://spec.graphql.org/October2021/#sec-Null-Value) states the following:

> GraphQL has two semantically different ways to represent the lack of a value:
> - Explicitly providing the literal value: `null`.
> - Implicitly not providing a value at all.

And also:
> ... **These two forms may be interpreted differently**. For example, a mutation representing deleting a field vs not altering a field, respectively. Neither form may be used for an input expecting a Non-Null type.

And so here's the problem with a classic `required: true` argument definition: it won't allow explicit `null` input. However, we might indeed want to have an argument (key?) required but allow explicit `null` value.

### Solution

We might introduce a new `required: :nullable` argument definition, that essentially means `required: false`, but with an extra validation that the argument was **explicitly** provided.

PS: I'm down to provide a PR if it is considered reasonable.

rmosolgo:
Fixes #3700

### DIFF
--- a/guides/fields/arguments.md
+++ b/guides/fields/arguments.md
@@ -22,6 +22,8 @@ def search_posts(category:)
 end
 ```
 
+### Nullability
+
 To make an argument optional, set `required: false`, and set default values for the corresponding keyword arguments:
 
 ```ruby
@@ -62,6 +64,10 @@ def search_posts(category:)
 end
 ```
 
+Finally, `required: :nullable` will require clients to pass the argument, although it will accept `null` as a valid input.
+
+### Deprecation
+
 **Experimental:** __Deprecated__ arguments can be marked by adding a `deprecation_reason:` keyword argument:
 
 ```ruby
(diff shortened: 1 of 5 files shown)


id-ilych:
Hey @rmosolgo :wave:
Thank you for the attention to the use-case 👍 

The resulting syntax is quite confusing:
```rb
field :update_settings, AccountSettings do
  argument :age, Integer, required: false
  validates required: { argument: :age }
end
```
Why would you say it’s `required: false` and then validate it as `required` :exploding_head:?

What do you think about the original suggestion `required: :nullable` or `required: :yes_but_nullable`?
Underlying implementation might stay the same.

andie-chernysh:
I agree with @id-ilych. His proposal looks much better for me in terms of readability.

rmosolgo:
Thanks again for taking a look, everyone. I've pushed an update to support `required: :nullable` as originally suggested! I'll ship it soon in 1.13.3. Please let me know what you think of it when you give it a try :+1:

PR #3790, opened 2021-12-17
Don't run a duplicate query when paginating already-loaded relations

### LINKED ISSUE #3775: GraphQL-Batch in connection field resolvers
**Describe the bug**

We use GraphL-Batch, but we're seeing N+1 query results in `connection_type` field resolvers.

We currently use the GraphQL-Batch [association loader](https://github.com/Shopify/graphql-batch/blob/master/examples/association_loader.rb) in any resolvers returning data through Active Record relations. For example, we basically have something like the following in our emails resolver:

```
        def emails
          AssociationsLoader.for(Client, :emails).load(object).then(&:emails)
        end
```

If emails are returned as an array and we run the following query:

```
query {
  clients {
    nodes {
      emails {
        id
      }
    }
  }
}
```

this will work as expected: all clients will be batched together, the emails association will be preloaded onto all clients and all emails will be retrieved in a single query allowing us to avoid the N+1.

But if emails are returned as a `connection_type` and we run the following query:

```
query {
  clients {
    nodes {
      emails {
        nodes {
          id
        }
      }
    }
  }
}
```

the batch loader still works and loads all emails in a single query, but we noticed the following query getting run for every client:

image

After looking at the source, I found that was happening [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/pagination/relation_connection.rb#L106). The issue I'm seeing is that [limit](https://guides.rubyonrails.org/active_record_querying.html#limit-and-offset) won't use the preloaded results and will always fetch from the database. After experimenting with this a bit in the console, I found that [take](https://guides.rubyonrails.org/active_record_querying.html#take) will [use already loaded results](https://github.com/rails/rails/blob/a35a380c2cdc7fe032f927095545aee21a192f54/activerecord/lib/active_record/relation/finder_methods.rb#L524) from preload. Would a possible solution be to update `GraphQL::Pagination::RelationConnection` to use `take` instead of `limit` to paginate the results? I know `limit` returns an `ActiveRecord_AssociationRelation` while `take` returns a `Array` and the pagination implementation would have to be changed 🙈 but I can't think of any other way to prevent the N+1 for the preloaded association.

We can get around this for now by updating the resolver to return an array so it gets paginated through `GraphQL::Pagination::ArrayConnection` instead, but I'm not sure if this is the best approach:

```
        def emails
          AssociationsLoader.for(Client, :emails).load(object).then do |client|
            client.emails.to_a
          end
        end
```

**Versions**

`graphql` version: 1.13.0
`graphql-batch` version: 0.4.3

rmosolgo:
Fixes #3775 

If the connection receives an already-loaded relation and can satisfy the query _without_ running another query, then don't run another query. Instead, choose ActiveRecord methods that will use already-loaded objects.

TODO: 

- [x] What if the relation has some other `LIMIT` or `OFFSET` applied, will this break? 
- [x] Implement without breaking Mongoid and Sequel code 

cc @crpahl

### DIFF
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -35,7 +35,7 @@ module GraphQL
             if @nodes && @nodes.count < first
               false
             else
-              relation_larger_than(sliced_nodes, first)
+              relation_larger_than(sliced_nodes, @sliced_nodes_offset, first)
             end
           else
             false
@@ -54,9 +54,10 @@ module GraphQL
       private
 
       # @param relation [Object] A database query object
+      # @param _initial_offset [Integer] The number of items already excluded from the relation
       # @param size [Integer] The value against which we check the relation size
       # @return [Boolean] True if the number of items in this relation is larger than `size`
-      def relation_larger_than(relation, size)
+      def relation_larger_than(relation, _initial_offset, size)
         relation_count(set_limit(relation, size + 1)) == size + 1
       end
 
@@ -111,36 +112,53 @@ module GraphQL
         end
       end
 
-      # Apply `before` and `after` to the underlying `items`,
-      # returning a new relation.
-      def sliced_nodes
-        @sliced_nodes ||= begin
+      def calculate_sliced_nodes_parameters
+        if defined?(@sliced_nodes_limit)
+          return
+        else
           paginated_nodes = items
 
           if after_offset
             previous_offset = relation_offset(items) || 0
-            paginated_nodes = set_offset(paginated_nodes, previous_offset + after_offset)
+            relation_offset = previous_offset + after_offset
           end
 
           if before_offset && after_offset
             if after_offset < before_offset
               # Get the number of items between the two cursors
               space_between = before_offset - after_offset - 1
-              @relation_limit = space_between
-              # paginated_nodes = set_limit(paginated_nodes, space_between)
+              relation_limit = space_between
             else
-              # TODO I think this is untested
               # The cursors overextend one another to an empty set
-              paginated_nodes = null_relation(paginated_nodes)
+              @sliced_nodes_null_relation = true
             end
           elsif before_offset
             # Use limit to cut off the tail of the relation
-            @relation_limit = before_offset - 1
-            # paginated_nodes = set_limit(paginated_nodes, before_offset - 1)
+            relation_limit = before_offset - 1
           end
 
-          if @relation_limit
-            paginated_nodes = set_limit(paginated_nodes, @relation_limit)
+          @sliced_nodes_limit = relation_limit
+          @sliced_nodes_offset = relation_offset || 0
+        end
+      end
+
+      # Apply `before` and `after` to the underlying `items`,
+      # returning a new relation.
+      def sliced_nodes
+        @sliced_nodes ||= begin
+          calculate_sliced_nodes_parameters
+          paginated_nodes = items
+
+          if @sliced_nodes_null_relation
+            paginated_nodes = null_relation(paginated_nodes)
+          else
+            if @sliced_nodes_limit
+              paginated_nodes = set_limit(paginated_nodes, @sliced_nodes_limit)
+            end
+
+            if @sliced_nodes_offset
+              paginated_nodes = set_offset(paginated_nodes, @sliced_nodes_offset)
+            end
           end
 
           paginated_nodes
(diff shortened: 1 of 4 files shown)


crpahl:
Sorry just finishing up some other work first, but I'll take a look at this soon.

lib/graphql/pagination/relation_connection.rb
```diff
@@ -156,32 +162,39 @@ def after_offset
...
           end
 
           @paged_nodes_offset = relation_offset(paginated_nodes)
-          paginated_nodes
+
+          if paginated_nodes.loaded?
+            paginated_nodes.take(@relation_limit)
+          else
+            paginated_nodes.limit(@relation_limit)
+          end
```
crpahl, line 197: This worked, but I think this if/else block can potentially just be replaced with only?

```
paginated_nodes.take(@relation_limit)
```

because `take` itself is also checking if the records are `loaded?`:

https://github.com/rails/rails/blob/a35a380c2cdc7fe032f927095545aee21a192f54/activerecord/lib/active_record/relation/finder_methods.rb#L524

I tried this out and it worked for both a preloaded association and one that wasn't preloaded. Although I'm guessing just calling `take` would break the `SequelDatasetConnection` and `MongoidRelationConnection` and another hook method would need to be added? But I guess that's already happening with `loaded?`.

crpahl:
I came across another case that triggers the N+1 even with this fix in place 😞 

```
query {
  clients {
    nodes {
      id
      emails(last: 1) {
        nodes {
          id
        }
      }
    }
  }
}
```

I noticed that calling `offset`, even on a preloaded association, will always trigger the association query, similar to `limit`. So any GQL query that causes `offset` to be called on the relation will also cause an N+1.

crpahl:
@rmosolgo Curious of what you think about this approach. What if the `ConnectionExtension` code [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/field/connection_extension.rb#L54), was updated to something roughly like this:

```
            elsif context.schema.new_connections?
              context.namespace(:connections)[:all_wrappers] ||= context.schema.connections.all_wrappers
              value = value.to_a if value.respond_to?(:loaded?) && value.respond_to?(:to_a) && value.loaded?
              context.schema.connections.wrap(field, object.object, value, original_arguments, context)
```

This is basically what we're doing in the workaround [here](https://github.com/rmosolgo/graphql-ruby/issues/3775#issuecomment-996493719): turning it into an array so it can be paginated by `ArrayConnection` when it's a loaded association.

crpahl:
Or I guess a better place to put that logic, if it's a reasonable approach, would be to put it somewhere in `GraphQL::Pagination::Connections`?

```
      def wrapper_for(items, wrappers: all_wrappers)
        impl = nil

        items.class.ancestors.each { |cls|
          impl = wrappers[cls]

          if impl == Pagination::ActiveRecordRelationConnection && items.loaded?
            items = items.to_a
            impl = Pagination::ArrayConnection
          end

          break if impl
        }

        [impl, items]
      end
```

```
        impl, items = wrapper_for(items, wrappers: wrappers)
```

rmosolgo:
> turning it into an array

I love that idea, and it'd _mostly_ work.... but as I started implementing it, I realized that it's impossible to serve a `totalCount`-type field in that case, since the array can't run a `count(*)` query. We need to retain the `Relation` to do that... Also, I think cursors would get messed up, since the array wouldn't have any information about the relation's offset or limit. 😕

lib/graphql/pagination/relation_connection.rb
rmosolgo: Yeah, definitely agreed that this can be cleaned up one way or another, as long as it works with the different backends :+1: :eyes:

rmosolgo:
It turned out to be pretty tricky. Since `.take` returns an Array, any codepath that might receive a Relation should also be prepared to take an array. 

Also, the previous implementation assumed that `relation = set_offset(relation, 1); set_offset(relation, 2)` would return a relation with an offset of `2`. But when `relation` is already loaded, the only way I could find to implement `offset` is `[offset..-1]`, which returns an array. Then, when you set the offset again, it removes more items from the array. So the implementation needed some cleanup: isolate the _calculations_ of offset and limit from the _applications_ of them, so that you could be sure you only apply the values once, depending on how/where they're needed :+1

PR #3792, opened 2021-12-17
Try to fix Rails 7 build

rmosolgo:
It's been broken for a long time ... I thought a Rails 7 release would fix it, but it didn't, so I'm trying again 😩


rmosolgo:
Well, it actually _did_ fix CRuby ... but truffle-ruby is still broken :( Anyhow, an improvement!

PR #3803, opened 2021-12-29
Fix custom field code example errors

bbugh:
There are a couple of errors in the code example that resulted in it not working. (field_options needs to be expanded apparently, and the blocks are evaluated in the instance context, not passing in a `self`). This resolves the issues.


rmosolgo:
Sorry for the trouble, and thanks for the fix!

PR #3804, opened 2021-12-29
Docs: Link to versioning policy from changelog

jaredbeck:


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+GraphQL-Ruby puts careful thought into its version numbers, but does not attempt to deliver "semantic versioning". Please see [the development guide](https://graphql-ruby.org/development.html#versioning) for details.
+
```
rmosolgo, line 4:
```suggestion
[Versioning guidelines](https://graphql-ruby.org/development.html#versioning) 
```

rmosolgo (commented):
Hi! Thanks for the suggestion. I'd be happy to include a link, but I'd rather keep it to a minimum (just a link), and if you think the versioning scheme needs further discussion, let's add that to `development.md` instead. How does that look to you?

CHANGELOG.md
jaredbeck: A link is better than nothing. Go for it.

PR #3806, opened 2021-12-30
Add code coverage

rmosolgo:
I've found this really useful in identifying dead code and untested code, and after using it on other projects for a couple years, I'd like to make the effort to get it going here, too. 

TODO: 

- [x] Make it so that nobody ever _has_ to update this locally -- GH Actions should open a PR if this file is out of date. Otherwise I'll be constantly apologizing for the terrible workflow 😅 
- [x] reorganize code and github action so that it's not a rats nest (it's still a rats nest, but at least its in its own file...)
- [x] fail if the artifacts are out-of-date
- [x] Add some docs to development.md about what this is for (and what it's not for) 
- [x] Simplify the memory benchmarks so that small changes don't cause them to be update (just memory/objects created/retained)
- [ ] Make this not flaky


rmosolgo:
I just don't know about this 🤔 It's going to add so much friction to the PR flow. I'm going to investigate a less disruptive option.

PR #3808, opened 2021-12-31
[Docs] Correction in object_from_id

dylancopeland:
Renamed the `id` parameter to `encoded_id_with_hint` in the `object_from_id` method. Previously, the method referenced an undefined variable (`encoded_id_with_hint`) that was supposed to be the global id that gets passed into `object_from_id`.


rmosolgo:
Thanks for the fix!

Issue #3809, opened 2022-01-01
Minor code error in documentation


danlo:
**Describe the bug**

A ```end``` block identifier appears to be extra code.

**Versions**

Documentation (current)

![image](https://user-images.githubusercontent.com/525181/147845323-5ca251ae-b009-4458-8666-e5ac829e4a6c.png)

**Steps to reproduce**
visit: 

https://graphql-ruby.org/authorization/can_can_integration#providing-a-custom-cancan-subject

**Expected behavior**

I think the ```end``` should be removed since there is no ```do``` to start it

PR #3811, opened 2022-01-03
Remove `end` token in code block example, fixing typo

joelzwarrington:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3809


rmosolgo:
Thanks for the fix!

PR #3818, opened 2022-01-06
Update required_validator.rb

rubendinho:
Make all arguments to be validated not required in example.


rmosolgo:
D'oh, thanks for the fix!

PR #3846, opened 2022-01-17
Fix RelationConnections when relation already has offset

### LINKED ISSUE #3843: Issue with `offset` after `v1.13.3`
**Describe the bug**

It looks like https://github.com/rmosolgo/graphql-ruby/commit/afd9b9207e19281f80ee35de1ebdc02d6c5f2342 from https://github.com/rmosolgo/graphql-ruby/pull/3790 has made using `offset` no longer work for us. The PR calls this potential out a bit which we seem to be hitting:

>  What if the relation has some other LIMIT or OFFSET applied, will this break?

We set `offset` on our relations but `graphql-ruby` is now resetting that `offset` outside of our control.

**Versions**

`graphql` version: `v1.13.3`
`rails` (or other framework): `6.1.4.4`
other applicable versions (`graphql-batch`, etc): `graphql-batch -> 0.4.3`

**Example of the problem**

This is a contrived example of what we do:

```
  field :items, Types::Items.connection_type, "A list of items.", connection: true, null: false, max_page_size: 25 do
    argument :offset, Integer, default_value: nil, required: false, description: "The number of records to skip when returning results. Useful for pagination."
  end

  def items(offset:)
    items = ::Items.all
    items = items.offset(offset) unless offset.nil?
    # Doing `pp items.to_sql` here would show that `OFFSET` is set properly but when `graphql-ruby`
    # executes this it adds the `LIMIT` properly as it's always done but resets `OFFSET` to zero.
    items
  end
```

rmosolgo:
Oops, there were some bugs in this implementation and it wasn't tested 😖 

Fixes #3843

### DIFF
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -47,7 +47,7 @@ module GraphQL
       def cursor_for(item)
         load_nodes
         # index in nodes + existing offset + 1 (because it's offset, not index)
-        offset = nodes.index(item) + 1 + (@paged_nodes_offset || 0) + (relation_offset(items) || 0)
+        offset = nodes.index(item) + 1 + (@paged_nodes_offset || 0) - (relation_offset(items) || 0)
         encode(offset.to_s)
       end
 
@@ -116,9 +116,9 @@ module GraphQL
         if defined?(@sliced_nodes_limit)
           return
         else
+          next_offset = relation_offset(items) || 0
           if after_offset
-            previous_offset = relation_offset(items) || 0
-            relation_offset = previous_offset + after_offset
+            next_offset += after_offset
           end
 
           if before_offset && after_offset
@@ -136,7 +136,7 @@ module GraphQL
           end
 
           @sliced_nodes_limit = relation_limit
-          @sliced_nodes_offset = relation_offset || 0
+          @sliced_nodes_offset = next_offset
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/graphql/pagination/relation_connection.rb
```diff
@@ -47,7 +47,7 @@ def has_next_page
       def cursor_for(item)
         load_nodes
         # index in nodes + existing offset + 1 (because it's offset, not index)
-        offset = nodes.index(item) + 1 + (@paged_nodes_offset || 0) + (relation_offset(items) || 0)
+        offset = nodes.index(item) + 1 + (@paged_nodes_offset || 0) - (relation_offset(items) || 0)
```
rmosolgo, line 50: It seems like I just ported this wrong, here's the previous implementation: 

https://github.com/rmosolgo/graphql-ruby/blob/1db571aa7bd2ec253053ab970cd167b42b13aa3b/lib/graphql/relay/relation_connection.rb#L14

lib/graphql/pagination/relation_connection.rb
```diff
@@ -116,9 +116,9 @@ def calculate_sliced_nodes_parameters
         if defined?(@sliced_nodes_limit)
           return
         else
+          next_offset = relation_offset(items) || 0
           if after_offset
-            previous_offset = relation_offset(items) || 0
```
rmosolgo, line 120: Oops, the previous offset was only applied _when_ `after:` was present 😖 . Now it's _always_ considered in the new offset value.

PR #3852, opened 2022-01-19
Remove TruffleRuby from CI

rmosolgo:
I'm tired of always seeing failed builds, and it seems like this is the flakiest one. Either Rails master is failing (as it did for a while when Rails generators wouldn't generate for `HEAD`, even when the invoked `rails ...` command was from `HEAD`) or TruffleRuby is failing (as it is currently because it's looking up `require "digest"` wrong, or something like that: https://github.com/oracle/truffleruby/issues/2511). 

If anyone wants to re-add a passing TruffleRuby build to this project (not truffleruby-head, not rails master), I'll try maintaining it again, but sheesh... 

cc @gogainda, in case you'd rather get this build _passing_ somehow ... (I can't figure out how to do it!)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,8 +51,6 @@ jobs:
             ruby: 3.0
           - gemfile: gemfiles/rails_master.gemfile
             ruby: 3.0
-          - gemfile: gemfiles/rails_master.gemfile
-            ruby: truffleruby-head
     runs-on: ubuntu-latest
     steps:
     - run: echo BUNDLE_GEMFILE=${{ matrix.gemfile }} > $GITHUB_ENV


gogainda:
Would you consider creating an issue in the truffleruby repository?

rmosolgo:
Isn't https://github.com/oracle/truffleruby/issues/2511 the issue for what's failing?

gogainda:
Created a new issue for truffleruby

bjfish:
This CI failure is tracked by the TruffleRuby team since it started happening. There is a TruffleRuby PR in progress to implement keyword argument handling to make this implementation pass:
https://github.com/rmosolgo/graphql-ruby/blob/b7fc1ae9b59fd2ee1556df85c77982f9dd09a8b0/spec/graphql/schema/dynamic_members_spec.rb#L33

I think it's unfortunate that Github Actions does not support something like continue on failure which is a suggested feature here: https://github.com/actions/toolkit/issues/399

I'll propose re-adding this to CI once the specs are all passing.

rmosolgo:
Yes, I'd definitely welcome a passing CI build!

eregon:
As a quick note the delegation code mentioned above could use `ruby2_keywords`, that would avoid code duplication, work on all Ruby version and TruffleRuby, and as a bonus is also faster on CRuby >= 3 than the current code.

eregon:
@bjfish Maybe you can make a PR with that change once the digest fix is merged?

Issue #3855, opened 2022-01-20
Support nested hash key as argument to field


joelzwarrington:
**Is your feature request related to a problem? Please describe.**

Trying to resolve nested attributes on a hash ends up creating a significant amount of boilerplate resolver methods.

```ruby
field :first_field, String
field :second_field, String

def my_user_facing_name
  object[:attributes][:first_nested_attribute]
end

def second_field
  object[:attributes][:second_nested_attribute]
end
```

**Describe the solution you'd like**

It would be nice to be able to use the `hash_key` argument on a field definition to determine how to resolve the field.

```
field :my_field, String, hash_key: [:attributes, :nested_attribute]
```

---

I would be open to working on this! My approach would be to modify the conditional block in `public_send_field` in `lib/graphql/schema/field.rb` so that it would look something like this:

```diff
- elsif obj.object.is_a?(Hash)
-   inner_object = obj.object
-   if inner_object.key?(@method_sym)
-     inner_object[@method_sym]
-   else
-     inner_object[@method_str]
-   end
+ elsif obj.object.is_a?(Hash)
+   inner_object = obj.object.with_indifferent_access
+   inner_object.dig(*@method_sym)
```

rmosolgo:
Sounds good to me! However, we can't assume that `.with_indifferent_access` will be present, since not all GraphQL-Ruby applications are also Rails applications.

Interestingly, it's _possible_ that swapping `inner_object[key]` for `inner_object.dig(*key)` would be a breaking change -- it could happen if someone was using arrays for hash keys. Maybe we could pick a _new_ keyword option for this, something like 

```ruby 
field ... dig: [:key_1, :key_2]
```

What do you think?

joelzwarrington:
- would you be opposed to including ActiveSupport?
- I'm not opposed to `dig`, but it'd likely add more complexity to the argument validation in `initialize` which isn't bad, but less code is more

re: breaking change, do you mean that someone using it like this would break?
```ruby
field :thing, String, hash_key: 0

# thing is provided an array ["Hello World"]
```

```ruby
if method && hash_key
  raise ArgumentError, "Provide `method:` _or_ `hash_key:`, not both. (called with: `method: #{method.inspect}, hash_key: #{hash_key.inspect}`)"
end

if resolver_method
  if method
[...]

PR #3856, opened 2022-01-21
Support nested keys as alternative to hash_key for fields

joelzwarrington:
closes https://github.com/rmosolgo/graphql-ruby/issues/3855

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -188,6 +188,7 @@ module GraphQL
       # @param deprecation_reason [String] If present, the field is marked "deprecated" with this message
       # @param method [Symbol] The method to call on the underlying object to resolve this field (defaults to `name`)
       # @param hash_key [String, Symbol] The hash key to lookup on the underlying object (if its a Hash) to resolve this field (defaults to `name` or `name.to_s`)
+      # @param dig [Array<String, Symbol>] The nested hash keys to lookup on the underlying hash to resolve this field using dig
       # @param resolver_method [Symbol] The method on the type to call to resolve this field (defaults to `name`)
       # @param connection [Boolean] `true` if this field should get automagic connection behavior; default is to infer by `*Connection` in the return type name
       # @param connection_extension [Class] The extension to add, to implement connections. If `nil`, no extension is added.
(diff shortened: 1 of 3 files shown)


joelzwarrington:
@rmosolgo I have something but am not entirely pleased with how I'm setting `method_sym` and overloading the shit out of it.

Also I'm explicitly not using `.to_sym` on each of the keys as it causes issues when not using an indifferent hash.

lib/graphql/schema/field.rb
```diff
@@ -236,26 +237,26 @@ def initialize(type: nil, name: nil, owner: nil, null: true, field: nil, functio
...
-          if hash_key
-            raise ArgumentError, "Provide `hash_key:` _or_ `resolver_method:`, not both. (called with: `hash_key: #{hash_key.inspect}, resolver_method: #{resolver_method.inspect}`)"
+          if hash_key || dig
+            raise ArgumentError, "Provide `hash_key:`, `dig:`, _or_ `resolver_method:`, not multiple. (called with: `hash_key: #{hash_key.inspect}, dig: #{dig.inspect}, resolver_method: #{resolver_method.inspect}`)"
           end
         end
 
         # TODO: I think non-string/symbol hash keys are wrongly normalized (eg `1` will not work)
-        method_name = method || hash_key || name_s
+        method_name = method || hash_key || dig || name_s
```
rmosolgo, line 255: I appreciate you're integrating with the existing code, but given that the previous code here is not ... _exemplary_ ... (😅 ), how about keeping `dig` _separate_ from the existing (confusing) attributes? I'll leave some suggestions to that effect...

```suggestion
        method_name = method || hash_key || name_s
        @dig_keys = dig
```

lib/graphql/schema/field.rb
```diff
@@ -236,26 +237,26 @@ def initialize(type: nil, name: nil, owner: nil, null: true, field: nil, functio
...
         resolver_method ||= name_s.to_sym
 
         @method_str = -method_name.to_s
-        @method_sym = method_name.to_sym
+        @method_sym = method_name.is_a?(Array) ? method_name : method_name.to_sym
```
rmosolgo, line 259:
```suggestion
        @method_sym =  method_name.to_sym
```

rmosolgo (commented):
The API and tests look good to me -- I just have one suggestion under the hood, let me know what you think of it.

joelzwarrington:
> The API and tests look good to me -- I just have one suggestion under the hood, let me know what you think of it.

Thanks @rmosolgo, I can make some changes based on your feedback.

Do you have anything planned to revisit some of the underlying API stuff? I'd be open to helping out and doing some refactoring if needed 😄

rmosolgo:
>  anything planned

No, nothing planned! I'd be happy to discuss in an issue or review a PR if you have something in mind, though :+1:

rmosolgo:
I'll work up a fix for rails master elsewhere. Thanks for this improvement!

PR #3874, opened 2022-01-30
Fix typo in generators.md

nerixim:
Just a small typo 🤓

### DIFF
--- a/guides/schema/generators.md
+++ b/guides/schema/generators.md
@@ -61,7 +61,7 @@ Several generators will add GraphQL types to your project. Run them with `-h` to
 
 ### ActiveRecord columns auto-extraction
 
-The `graphql:object` and `graphql:input` generators can detect the existence of an ActiveRecord class with the same name, and scaffold all database columns as fields/arguments using aproppiate GraphQL types and nullability detection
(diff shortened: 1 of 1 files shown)


guides/schema/generators.md
```diff
@@ -61,7 +61,7 @@ Several generators will add GraphQL types to your project. Run them with `-h` to
 
 ### ActiveRecord columns auto-extraction
 
-The `graphql:object` and `graphql:input` generators can detect the existence of an ActiveRecord class with the same name, and scaffold all database columns as fields/arguments using aproppiate GraphQL types and nullability detection
+The `graphql:object` and `graphql:input` generators can detect the existence of an ActiveRecord class with the same name, and scaffold all database columns as fields/arguments using aproppriate GraphQL types and nullability detection
```
rmosolgo, line 64:
```suggestion
The `graphql:object` and `graphql:input` generators can detect the existence of an ActiveRecord class with the same name, and scaffold all database columns as fields/arguments using appropriate GraphQL types and nullability detection
```

One more fix!

rmosolgo:
Thanks for the fix!

guides/schema/generators.md
nerixim: Well at least I found it 😅

PR #3879, opened 2022-01-31
[1.12.x] Implement Directive#graphql_name

rmosolgo:
Master will need this fix, too.


rmosolgo:
Added a test in https://github.com/rmosolgo/graphql-ruby/pull/3886

PR #3881, opened 2022-02-01
Provide argument to skip static validation

toneymathews:
~~This PR skips static validation for a GraphQL operation by setting `query.skip_static_validation = true`. This allows us to skip static validation for repeated operations for which this result can be cached.~~

~~Variable validation runs irrespective whether or not static validation is skipped.~~

This PR allows a GraphQL operation to skip static validation through `query.validate` property. Though the property was available earlier, it did not work as expected. Now the dependency is inverted so that `ValidationPipeline` fetches it from `Query` class.

cc @haruska

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -405,7 +405,6 @@ module GraphQL
 
       @validation_pipeline = GraphQL::Query::ValidationPipeline.new(
         query: self,
-        validate: @validate,
         parse_error: parse_error,
         operation_name_error: operation_name_error,
         max_depth: @max_depth,
--- a/lib/graphql/query/validation_pipeline.rb
+++ b/lib/graphql/query/validation_pipeline.rb
@@ -16,9 +16,8 @@ module GraphQL
     class ValidationPipeline
       attr_reader :max_depth, :max_complexity
 
-      def initialize(query:, validate:, parse_error:, operation_name_error:, max_depth:, max_complexity:)
+      def initialize(query:, parse_error:, operation_name_error:, max_depth:, max_complexity:)
         @validation_errors = []
-        @validate = validate
         @parse_error = parse_error
         @operation_name_error = operation_name_error
         @query = query
@@ -65,7 +64,7 @@ module GraphQL
         elsif @operation_name_error
           @validation_errors << @operation_name_error
         else
-          validation_result = @schema.static_validator.validate(@query, validate: @validate, timeout: @schema.validate_timeout, max_errors: @schema.validate_max_errors)
+          validation_result = @schema.static_validator.validate(@query, validate: @query.validate, timeout: @schema.validate_timeout, max_errors: @schema.validate_max_errors)
           @validation_errors.concat(validation_result[:errors])
 
           if @validation_errors.empty?
--- a/spec/graphql/query_spec.rb
+++ b/spec/graphql/query_spec.rb
@@ -699,6 +699,13 @@ describe GraphQL::Query do
       query = GraphQL::Query.new(schema, invalid_query_string, validate: true)
       assert_equal false, query.valid?
       assert_equal 1, query.static_errors.length
+
+      # Can assign attribute after calling methods that use the AST
+      query = GraphQL::Query.new(schema, invalid_query_string)
+      assert query.fingerprint
+      query.validate = false
+      assert_equal true, query.valid?
+      assert_equal 0, query.static_errors.length
     end
   end
 


swalkinshaw:
Should note some context here because `GraphQL::Query` does support a `validate` boolean already. It just doesn't really work!

So there's two options:
1. fix the existing kwarg
2. a new solution like this PR proposes

I'm kind of going off memory, but I believe the problem is here: https://github.com/rmosolgo/graphql-ruby/blob/a08c86f097395e27272fdead5037d7b3ef03401e/lib/graphql/query/validation_pipeline.rb#L68

`@validate` is passed through from `Query` except it's effectively ignored by `schema.static_validator`: https://github.com/rmosolgo/graphql-ruby/blob/a08c86f097395e27272fdead5037d7b3ef03401e/lib/graphql/schema.rb#L197-L199

This solution still leaves the existing broken/confusing situation unfortunately.

haruska:
> This solution still leaves the existing broken/confusing situation unfortunately.

I may be understanding the code wrong but it seems the `@validate` boolean is doing too much. It can be set to false or true during various logic in the gem. So, if the calling code sets validate to false, there still may be some internal code that calculates validation should occur and sets it back to true.

I like this PR as a solution because it's explicit that the caller is requesting the validation skip.

swalkinshaw:
Yeah I don't disagree! But this potentially makes it _more_ confusing by having the existing `@validate` that already does too much, plus a new setting.

Maybe `@validate` should be marked as deprecated?

@rmosolgo will know more 😄

rmosolgo:
This makes sense to me. I overlooked it before, but if I understand correctly, `validate: true|false` really covered _two_ kinds of validation: 

- Query string validation: according to the spec, a static operation taking the schema and the query string as inputs. (In fact, it's dynamic in GraphQL-Ruby, since the schema can be different for a given `context`.) 
- Variable validation: _always_ dynamic, since it's validating the current query variables against the query string and the schema. 

Makes sense to me to isolate these two and allow them to be configured separately. (Have I understood your intention correctly?) 

If that's what you have in mind, how about accepting a couple of new parameters, `validate_query_string:` and `validate_variables:`, in a backwards-compatible way. In my opinion, `true` for _either_ a new, specific parameter _or_ the previous `validate:` parameter should cause the validation process to run. That is:

```ruby
# query.rb 
# @param validate [Boolean] If true, query string validation and variable validation will run (even if `validate_query_string:` or `validate_variables:` is false. 
# @param validate_query_string [Boolean] If true, query string validation will run (even if `validate:` is false)
# @param validate_variables [Boolean] If true, variable validation will run (even if `validate:` is false)
def initialize(..., validate_query_string: nil, validate_variables: nil, validate: nil) 
   if validate_query_string.nil? && validate_variables.nil? && validate.nil?
     # No value was given for any of these options, so default to running all validations
     validate_variables = validate_query_string = validate = true
   else  
     # One or more validation options was provided, so 
     # run validation operations if _either_ of the options was `true` for that operation
     validate_query_string ||= validate 
     validate_variables ||= validate 
   end
end 
```
     
     
What do you think of an approach like that, would that work for your use case? 

> Maybe `@validate` should be marked as deprecated?

Yeah, I can understand that. It really covers too much to be helpful, and if it _really_ opens the door to broken behavior, I'm open to deprecating it.

swalkinshaw:
Wanted to clarify my [first comment](https://github.com/rmosolgo/graphql-ruby/pull/3881#issuecomment-1027172102) because it's just wrong (was going off memory).

That workflow is fine, there's even tests for it:

https://github.com/rmosolgo/graphql-ruby/blob/ae8ba32b04564e7588100e43d8bfcbe61861e7ac/spec/graphql/query_spec.rb#L684-L703

The problem for our usage is that `Query` defines the `validation_pipeline` very early within `prepare_ast` (which is triggered by almost anything). So in unit tests, it's pretty easy to just set that kwarg/attr to `false` and have it work. But in our real world usage, it's not possible to run whatever logic we need to conditionally disable `validate` _before_ `prepare_ast` is called and by then it's too late.

example:

```ruby
query = GraphQL::Query.new(schema, query_string)
query.fingerprint # maybe for logging purposes => triggers `prepare_ast` 

query.validate = false # now it's too late, doesn't do anything
query.result # static validation occurs
```

rmosolgo:
> by then it's too late

That makes sense -- I bet we could either:

- split `prepare_ast` so that it doesn't initialize `@validation_pipeline` until it really needs it (probably breaking `@validation_pipeline = ...` to a method of its own or something)
- reverse the dependency so that, instead of `Query` passing `validate` _into_ `ValidationPipeline`, instead, `ValidationPipeline` calls `query.validate` when it actually needs the value. That might put off the call so that it would work in your case. 

I can see pros and cons to each, but I think either one would probably make time for some calculations before assigning the attribute. Another thing is, we should probably make the assignment methods  raise an error if they're assigned after validation is run (otherwise, people might be surprised that their assignments make no difference). 

I realize this is growing a bit from the initial scope of this PR -- let me know if you want me to take a stab at it, I'd be happy to do that or wait here for something new to review.

swalkinshaw:
> I realize this is growing a bit from the initial scope of this PR

I'd say it's the same scope since this PR is a workaround/solution for the current limitation on `query.validate`.

I like the idea of inverting the dependencies and it's probably simpler since it looks like `ValidationPipeline.new` is fairly coupled to `prepare_ast` since it uses other values calculated within it?

It would be a bit of a breaking change though without a way to properly deprecate it I don't think.

> Another thing is, we should probably make the assignment methods raise an error if they're assigned after validation is run (otherwise, people might be surprised that their assignments make no difference).

good idea 👍

toneymathews:
Ok this helps. I'll try and work on this PR the next few days and get back with any questions/put up a PR for review.

toneymathews:
The dependency is now inverted and the pipeline now calls `query.validate`. This does have a requisite test in place as well. This works for our usecase.

https://github.com/rmosolgo/graphql-ruby/blob/79ab2eadbd56602d12b989f3dc361b1212f9cddf/spec/graphql/query_spec.rb#L684-L703


>Another thing is, we should probably make the assignment methods raise an error if they're assigned after validation is run (otherwise, people might be surprised that their assignments make no difference).

I wasn't sure how to go about this. But this would be beyond the scope of this PR?

haruska:
This is a good minimal patch that simplifies things with `query.validate` vs the pipeline `@validate` variable. 👍 

> > Another thing is, we should probably make the assignment methods raise an error if they're assigned after validation is run (otherwise, people might be surprised that their assignments make no difference).
> 
> I wasn't sure how to go about this. But this would be beyond the scope of this PR?

Maybe this can be a separate issue/enhancement.

rmosolgo:
> requisite test 

I see what you mean that it's already tested, but since y'all are building on a new behavior, I'd really like to get a test for that behavior in the gem. That is, something like: 

```ruby
     # Can assign attribute after calling methods that use the AST 
     query = GraphQL::Query.new(schema, invalid_query_string) 
     assert query.fingerprint
     query.validate = false 
     assert_equal true, query.valid? 
     assert_equal 0, query.static_errors.length 
```

Otherwise, IIUC, if I reverted the change here, all my tests would still pass, but I'd have broken it for you 😖. Did I get that right, or is there already a test in this project for the scenario I sketched out above?

rmosolgo:
Thanks for this improvement!

rmosolgo:
Cherry-picked and released in 1.13.15, too :+1:

PR #3912, opened 2022-02-11
Update getting_started.md

4geru:
fix indent

スクリーンショット 2022-02-11 10 23 24


rmosolgo:
Hmm, I don't agree with this change. 

For my part, when I write a method call whose arguments extend to the following lines, I indent those lines to give a visual cue that "these lines are subordinate to the line above", for example: 

```
validate :some_attribute, 
  presence: true,
  uniqueness: { scope: :parent_id },
  format: { with: /blah/ }
```

The example in the docs here is the same thing: `field :comments...` has a `description:` on the next line. I think it should be indented for the same reason.

PR #3913, opened 2022-02-11
Dataloader: accept a limit on the number of fibers

rmosolgo:
Now, you could pass `fiber_limit:` to tell dataloader to use no more than the given number of fibers at once. 

(I got a bug report that `Fiber#resume` can fail with `FiberError: can't set a guard page: Cannot allocate memory` when Ruby runs up against a VM's `vm.max_map_count`. Besides increasing that limit, this option should put a cap on how many fibers are spun up at a time.) 

You can either pass default fiber limit: 

```ruby 
use GraphQL::Dataloader, fiber_limit: 10_000
```

or initialize a dataloader with a limit dynamically: 

```ruby 
context = { ... } 

if feature_flags[:dataloader_fiber_limit].enabled? # for example 
  # Assign a dataloader here to override the default configs:
  context[:dataloader] = GraphQL::Dataloader.new(fiber_limit: 10_000)
end

MySchema.execute(...) 
```

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -3,6 +3,7 @@
 require "graphql/dataloader/null_dataloader"
 require "graphql/dataloader/request"
 require "graphql/dataloader/request_all"
+require "graphql/dataloader/run"
 require "graphql/dataloader/source"
 
 module GraphQL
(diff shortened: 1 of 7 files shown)


rmosolgo:
I'm going to hold off on this til we encounter a real need for it.

ganey:
@rmosolgo I recently started getting this, dataloading probably 10k items across 5 different dataloaders.

dataloaders are using ActiveRecord, and I'm also using Puma.

ganey:
I would also find this useful, we're getting Fiber out of memory errors as sometimes we're spawning about 15k Fibers on kube pods with 14gb ram.

rmosolgo:
Thanks for sharing your findings here -- I'll dust off this PR and get it merged soon!

rmosolgo:
I'm working on this over here #5132

PR #3924, opened 2022-02-13
Support dynamic schema for Subscriptions#trigger

### LINKED ISSUE #3897: Dynamic schema field for Subscription using `.visible?(context)`

[1.13.0](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#1130-24-november-2021) introduced dynamic schema based with `.visible?(context)`

> _Visibility: A schema may contain multiple members with the same name. For each name, GraphQL-Ruby will use the one that returns true for .visible?(context) for each query (and raise an error if multiple objects with the same name are visible). #3651 #3716 #3725_

The related test added in #3651 covers `query` and `mutation`, but not `subscription`.

https://github.com/rmosolgo/graphql-ruby/blob/79ab2eadbd56602d12b989f3dc361b1212f9cddf/spec/graphql/schema/dynamic_members_spec.rb#L357-L358

Looking through the doc and changed files in related PR, I'm not sure if this feature also supports subscription.

I'm using [dynamic fields using `.visible(context)`](https://graphql-ruby.org/schema/dynamic_types.html#using-visiblecontext) to separate implementations for different contexts.

```rb
module Types
  class SubscriptionType < Types::BaseObject
    field :foo, subscription: Subscriptions::Foo::ForUser
    field :foo, subscription: Subscriptions::Foo::ForAdmin
  end
end

module Subscriptions
  module Foo
    class Base < BaseSubscription
      def visible?(context)
        result = self == { User => ForUser, Admin => ForAdmin }.fetch(context[:current_account].class)
        p("visible? -> #{result}") # This prints two times, once `true` and once `false`

        result
      end
    end
  end
end

module Subscriptions
  module Foo
    class ForUser < Base
      def authorized?(**args)
        # Authorization logic for user account
      end

      def subscribe(**args)
        # Implementation for user account
      end
    end
  end
end

module Subscriptions
  module Foo
    class ForAdmin < Base
      def authorized?(**args)
        # Authorization logic for admin account
      end

      def subscribe(**args)
        # Implementation for admin account
      end
    end
  end
end
```

Doing similar things in query will work, but in subscription like above I'll get a `DuplicateNamesError` from `warden.rb` when `trigger`.

```log
GraphQL::Schema::DuplicateNamesError: Found two visible definitions for `Subscription.foo`: #<Types::BaseField Subscription.foo(...): ForUserload!>, #<Types::BaseField Subscription.foo(...): ForAdminPayload!>

/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema/warden.rb:58:in `block in visible_entry?'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema/warden.rb:53:in `each'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema/warden.rb:53:in `visible_entry?'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema/member/has_fields.rb:46:in `block in get_field'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema/member/has_fields.rb:42:in `each'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema/member/has_fields.rb:42:in `get_field'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/schema.rb:1253:in `get_field'
/Users/apple/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/graphql-1.13.2/lib/graphql/subscriptions.rb:76:in `trigger'
...
```

I haven't looked deep into #3651. If this is technically achievable then I'd love to help.

Looking through the call stack, I'm suspecting if this is related to [`GraphQL::Schema#get_field`](https://github.com/rmosolgo/graphql-ruby/blob/v1.13.2/lib/graphql/schema.rb#L1241) has its third argument `context` defaults to `GraphQL::Query::NullContext`.

Perhaps allowing passing context to [`GraphQL::Subscriptions#trigger`](https://github.com/rmosolgo/graphql-ruby/blob/v1.13.2/lib/graphql/subscriptions.rb#L67-L76) can solve this problem?
```diff
-    def trigger(event_name, args, object, scope: nil)
+    def trigger(event_name, args, object, scope: nil, context: GraphQL::Query::NullContext)
    def trigger(event_name, args, object, scope: nil)
      event_name = event_name.to_s

      # Try with the verbatim input first:
      field = @schema.get_field(@schema.subscription, event_name)

      if field.nil?
        # And if it wasn't found, normalize it:
        normalized_event_name = normalize_name(event_name)
-        field = @schema.get_field(@schema.subscription, normalized_event_name)
+        field = @schema.get_field(@schema.subscription, normalized_event_name, context)
```

If this is technically difficult to implement for subscriptions, I can also help update the [related documentation](https://graphql-ruby.org/schema/dynamic_types.html#using-visiblecontext) to mention this limitation.

**Additional context**

choznerol:
Close #3897 

a73b6d58d4280964917cf0c71dac7afd4ba82294 can reproduce the problem described in #3897 :
>

### DIFF
--- a/spec/graphql/schema/subscription_spec.rb
+++ b/spec/graphql/schema/subscription_spec.rb
@@ -20,10 +20,23 @@ describe GraphQL::Schema::Subscription do
     class Toot < GraphQL::Schema::Object
       field :handle, String, null: false
       field :body, String, null: false
+
+      def self.visible?(context)
+        !context[:legacy_schema]
+      end
+    end
+
+    class LegacyToot < Toot
+      field :likes_count, Int, null: false
+
+      def self.visible?(context)
+        !!context[:legacy_schema]
+      end
     end
 
     class Query < GraphQL::Schema::Object
       field :toots, [Toot], null: false
+      field :toots, [LegacyToot], null: false
 
       def toots
         TOOTS
@@ -38,6 +51,11 @@ describe GraphQL::Schema::Subscription do
 
       field :toot, Toot, null: false
       field :user, User, null: false
+
+      def self.visible?(context)
+        !context[:legacy_schema]
+      end
+
       # Can't subscribe to private users
       def authorized?(user:, path:, query:)
         if user[:private]
@@ -67,6 +85,14 @@ describe GraphQL::Schema::Subscription do
           # for testing that the default implementation is `return object`
           super
         end
+      end
+    end
+
+    class LegacyTootWasTooted < TootWasTooted
+      field :toot, LegacyToot
+
+      def self.visible?(context)
+        !!context[:legacy_schema]
       end
     end
 
@@ -108,6 +134,7 @@ describe GraphQL::Schema::Subscription do
 
     class Subscription < GraphQL::Schema::Object
       field :toot_was_tooted, subscription: TootWasTooted, extras: [:path, :query]
+      field :toot_was_tooted, subscription: LegacyTootWasTooted, extras: [:path, :query]
       field :direct_toot_was_tooted, subscription: DirectTootWasTooted
       field :direct_toot_was_tooted_with_optional_scope, subscription: DirectTootWasTootedWithOptionalScope
       field :users_joined, subscription: UsersJoined
--- a/spec/graphql/schema/subscription_spec.rb
+++ b/spec/graphql/schema/subscription_spec.rb
@@ -395,7 +395,7 @@ describe GraphQL::Schema::Subscription do
       GRAPHQL
       assert_equal 1, in_memory_subscription_count
       obj = OpenStruct.new(toot: { body: "I am a C programmer" }, user: SubscriptionFieldSchema::USERS["matz"])
-      SubscriptionFieldSchema.subscriptions.trigger(:toot_was_tooted, {handle: "matz"}, obj)
+      SubscriptionFieldSchema.subscriptions.trigger(:toot_was_tooted, {handle: "matz"}, obj, context: {})
 
       mailbox = res.context[:subscription_mailbox]
       update_payload = mailbox.first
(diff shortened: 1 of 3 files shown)


spec/graphql/schema/subscription_spec.rb
```diff
@@ -469,20 +496,42 @@ def in_memory_subscription_count
...
+    end
   end
 
   describe "applying `loads:`" do
     it "includes `as:` in the event topic" do
       assert_equal [], SubscriptionFieldSchema::InMemorySubscriptions::EVENT_REGISTRY.keys
       matz = SubscriptionFieldSchema::USERS["matz"]
       obj = OpenStruct.new(toot: { body: "I am a C programmer" }, user: matz)
-      SubscriptionFieldSchema.subscriptions.trigger(:toot_was_tooted, {handle: "matz"}, obj)
+      SubscriptionFieldSchema.subscriptions.trigger(:toot_was_tooted, {handle: "matz"}, obj, context: {})
```
rmosolgo, line 534: Out of curiousity, were these `context: {}`s neccessary? (I would have _expected_ this parameter's default value, `NullContext`, to be sufficient -- but maybe you added them for clarity?)

rmosolgo (commented):
LGTM! Just one question, if you don't mind taking a look.

spec/graphql/schema/subscription_spec.rb
choznerol: Good question! `DuplicatedNameError` will be raised without all the `context: {}`s because `NullContext::NullWarden#visible_field?` always return `true` (Unlike `NullContext::NullWarden`). I did worry about whether adding all the `context: {}`s indicates a breaking change, but later realized the behavior (i.e. raising `DuplicatedNameError`) did not change before or after this PR.

rmosolgo: Thanks for clarifying that ... I think it'd be nice to _not_ have to pass an empty hash when we don't really need any context values, but I'm not sure the best way to accomplish that. I'll merge this and take a look at making that further change.

rmosolgo:
Thanks for this improvement!

PR #3932, opened 2022-02-15
Same Interface Shows Up Multiple Times in "Implements"

dleavitt:
Per #3613 I put interfaces in my interfaces 🎉   

But now my types have too many interfaces 🤢 

Specifically, if you've got a schema like this:
```graphql
interface Node { ... }
interface Named implements Node { ... }
interface Timestamped implements Node { ... }
```
And you add a type `Thing` that implements `Named` and `Timestamped`, printing the schema will give you this:
```graphql
type Thing implements Named & Node & Node & Timestamped { ... }
```
I didn't see much in the spec about this but the reference implementation doesn't like it:
https://github.com/graphql/graphql-js/blob/95dac43fd4bff037e06adaa7cfb44f497bca94a7/src/type/validate.ts#L338-L348 

It's also not stable - graphql-ruby will add an additional `& Node` each time you parse and then print the SDL.

First commit: does it just in the printer.

Second commit: tries to do it for the schema/introspection as well. Wasn't 100% sure about where the deduping should happen though (`#implements` vs `#interfaces` vs `#interface_type_memberships` vs `#own_interface_type_memberships`.)

### DIFF
--- a/lib/graphql/language/printer.rb
+++ b/lib/graphql/language/printer.rb
@@ -170,7 +170,7 @@ module GraphQL
       end
 
       def print_implements(type)
-        " implements #{type.interfaces.map(&:name).join(" & ")}"
+        " implements #{type.interfaces.map(&:name).uniq.join(" & ")}"
       end
 
       def print_input_value_definition(input_value)
--- a/lib/graphql/language/printer.rb
+++ b/lib/graphql/language/printer.rb
@@ -170,7 +170,7 @@ module GraphQL
       end
 
       def print_implements(type)
-        " implements #{type.interfaces.map(&:name).uniq.join(" & ")}"
+        " implements #{type.interfaces.map(&:name).join(" & ")}"
       end
 
       def print_input_value_definition(input_value)
--- a/lib/graphql/schema/member/has_interfaces.rb
+++ b/lib/graphql/schema/member/has_interfaces.rb
@@ -33,9 +33,9 @@ module GraphQL
             end
           end
 
-          # Remove any String or late-bound interfaces which are being replaced
           own_interface_type_memberships.reject! { |old_i_m|
             if !(old_i_m.respond_to?(:abstract_type) && old_i_m.abstract_type.is_a?(Module))
+              # Remove any String or late-bound interfaces which are being replaced
               old_int_type = old_i_m.respond_to?(:abstract_type) ? old_i_m.abstract_type : old_i_m
               old_name = Schema::Member::BuildType.to_type_name(old_int_type)
 
@@ -44,6 +44,11 @@ module GraphQL
                 new_name = Schema::Member::BuildType.to_type_name(new_int_type)
 
                 new_name == old_name
+              }
+            elsif old_i_m.respond_to?(:abstract_type)
+              # Remove any existing implementations of the same interface
+              new_memberships.any? { |new_i_m|
+                new_i_m.respond_to?(:abstract_type) && new_i_m.abstract_type == old_i_m.abstract_type
               }
             end
           }
@@ -80,7 +85,7 @@ module GraphQL
             visible_interfaces.concat(superclass.interfaces(context))
           end
 
-          visible_interfaces.uniq
+          visible_interfaces
         end
       end
     end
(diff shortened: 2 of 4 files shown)


rmosolgo:
D'oh! Thanks for the fix on this. 

One thing on my mind is, what about the list of `type_memberships`, does it _also_ have duplicates in it? (It strikes me that `.uniq` might clean up the return value of _this method_, but is the internal structure of the schema still "wrong" in some way?) 

I definitely want to get this fixed, but I'd prefer a solution that _builds_ the schema properly rather than one that cleans up structure issues at the "last minute."  I _suspect_ that what we really need is some duplicate checking in this method: 

https://github.com/rmosolgo/graphql-ruby/blob/faeef68b847d2086342651530ac0a15f2d6a4fb0/lib/graphql/schema/member/has_interfaces.rb#L7

Here's where an interfaces's interfaces are added to the object type: 

https://github.com/rmosolgo/graphql-ruby/blob/faeef68b847d2086342651530ac0a15f2d6a4fb0/lib/graphql/schema/member/has_interfaces.rb#L22-L25

Maybe that line should somehow check whether this type already implemented that interface. 

Alternatively, here's some code that replaces "temporary" interface implementations with newly-created ones:

https://github.com/rmosolgo/graphql-ruby/blob/faeef68b847d2086342651530ac0a15f2d6a4fb0/lib/graphql/schema/member/has_interfaces.rb#L36-L49

Sometimes, when a schema is loading, it uses Strings or `LateBoundType` instances here, and those objects are cleaned up by that code. But maybe that code could be expanded to replace _any_ type membership that matches the newly-created ones. 

Anyhow... not sure exactly which approach would work best, but I'd like to get the schema structure "right" on the way _in_ somehow. Would you like to take a crack at it? If not, I can probably take a look soon.

dleavitt:
Thanks for the quick and thoughtful response! Agree and I'll look into a righter approach.

For any other reckless interfaciers who find this through google, here's a workaround:
```ruby
# BaseObject
def self.interfaces(...)
  super.uniq
end

# BaseInterface
definition_methods do
  def interfaces(...)
    super.uniq
  end
end
```

dleavitt:
Added another commit along the lines of your second suggestion:

https://github.com/rmosolgo/graphql-ruby/blob/0f6aead95c8b9d99c8df0c222d6f26317b356f3f/lib/graphql/schema/member/has_interfaces.rb#L48-L53

Tests seem to pass, but:
- Does the way I'm comparing the interfaces seem right? 
- Something could implement the same interface twice but with different options on the type membership. Not sure if this matters or how we'd tell. I think the last one implemented will win right now FWIW. 
- It looks like interfaces from superclasses get concatted to the return value of `.interfaces`, so we need to dedupe in there anyway.

rmosolgo:
Looking good so far, thanks again for taking another look at this!

> comparing the interfaces?

Yeah, I think that's good. The previous code uses names, but I think that's because it's trying to match Strings or LateBoundTypes (which only _have_ names).

> last one implemented will win

Given the options, seems good to me :+1: 

>  interfaces from superclasses get concatted

Good call 🍻

dleavitt:
It looks like the last implementation winning and overwriting existing memberships options is what's causing the test failures. Deduplicating when new interfaces are added means that things like this won't work:

https://github.com/rmosolgo/graphql-ruby/blob/b957fb52f5293c15bcaac36899c8752cc98c1c53/spec/graphql/schema/dynamic_members_spec.rb#L196-L197

Seems safest to keep all the type memberships around and just dedupe the returned list of interfaces.

rmosolgo:
🤦 D'oh, of course it does. I was thinking only of the case where you might have two interface definitions with the same name (but the interfaces are different in Ruby-land), but I forgot about that feature, where you use the _same_ interface but provide different options to the implementation relationship. Thanks for looking into it, but sorry I pointed you in the wrong direction. 

Thanks again for this fix! I'm going to follow up on one more thing on another branch (https://github.com/rmosolgo/graphql-ruby/issues/3954)

dleavitt:
No problem! Thanks for this amazing library!

PR #3939, opened 2022-02-16
Demonstrate errors when using method attribute

alexkalderimis:
This branch reproduces an error we saw when writing code for a connection with aggregate fields using the `:method` field attribute.

It is perfectly possible to have custom fields on connections, just not to use the `:method` attribute.


alexkalderimis:
I realised after making this whole PR that I misunderstood the docs 🤦 

This attribute is only used for delegating to the object itself, not for respecifying the method in the type.

Apologies for any notification spam.

PR #3946, opened 2022-02-17
Fix argument option name.

ycleptkellan:
`replace_null_with_default_value` -> `replace_null_with_default`.


rmosolgo:
D'oh! Thanks for the fix here.

PR #3947, opened 2022-02-18
Refactor resolving of field

### LINKED ISSUE #3861: Cleanup & refactor field class API
@rmosolgo I didn't have anything particular in mind but I'm open to helping out how I can.

Is there anything in particular you think time would be best spent? I was thinking of looking into how a field is resolved (`resolve_field`, `resolve`, `public_send_field`) unless you had other suggestions!

This issue was created as follow up to https://github.com/rmosolgo/graphql-ruby/pull/3856

### LINKED ISSUE #3944: hash_key attribute doesn't work if field is called "method"
**Describe the bug**

It's not possible to use the `hash_key` keyword if the field name is `method`, for example, the following will produce an error:

```ruby
field :method, String, null: true, hash_key: :another_method
```

Error when executing a query:

```json
{
  "errors": [
    {
      "message": "wrong number of arguments (given 0, expected 1)",
      "backtrace": [
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:692:in `method'"
      ]
    }
  ]
}
```

**Versions**

`graphql` version: `1.11.6`
`rails` (or other framework): `7.0.2`

**GraphQL schema**

Example code has been taken from [graphql-ruby Tutorial](https://www.howtographql.com/graphql-ruby/0-introduction/)

**app/models/link.rb**

```ruby
class Link < ApplicationRecord
  def another_method
    "other method"
  end
end
```

**app/graphql/types/link_type.rb**

```ruby
module Types
  class LinkType < Types::BaseObject
    field :id, ID, null: false
    field :url, String, null: false
    field :description, String, null: false
    field :testing, String, null: false

    field :method, String, null: true, hash_key: :another_method
  end
end
```

**GraphQL query**

```graphql
{
  allLinks {
    method
  }
}
```

```json
{
  "errors": [
    {
      "message": "wrong number of arguments (given 0, expected 1)",
      "backtrace": [
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:692:in `method'"
      ]
    }
  ]
}
```

**Steps to reproduce**

Define a field with the following signature:

```ruby
module Types
  class LinkType < Types::BaseObject
    field :method, String, null: true, hash_key: :another_method
  end
end
```

Implement `another_method` inside the source class:

```ruby
class Link < ApplicationRecord
  def another_method
    "other method"
  end
end
```

Execute a query:

```graphql
{
  allLinks {
    method
  }
}
```

Error is returned:

```json
{
  "errors": [
    {
      "message": "wrong number of arguments (given 0, expected 1)",
      "backtrace": [
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:692:in `method'",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:692:in `public_send'",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:692:in `block in public_send_field'",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:726:in `with_extensions'",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:672:in `public_send_field'",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/graphql-1.11.6/lib/graphql/schema/field.rb:585:in `block in resolve'",
        "<snip>",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/puma-5.6.2/lib/puma/server.rb:441:in `process_client'",
        "/Users/myname/.rvm/gems/ruby-2.7.5/gems/puma-5.6.2/lib/puma/thread_pool.rb:147:in `block in spawn_thread'"
      ]
    }
  ],
  "data": {}
}
```

If we use `resolver_method` instead:

```ruby
module Types
  class LinkType < Types::BaseObject
    field :method, String, null: true, resolver_method: :resolve_method

    def resolve_method
      object.another_method
    end
  end
end
```

The query works as expected and returns the following:

```json
{
  "data": {
    "allLinks": [
      {
        "method": "other method"
      },
      {
        "method": "other method"
      }
    ]
  }
}
```

**Expected behavior**

It should be possible to use the `hash_key` with a field named `method` _or_ the documentation should be updated to explain that this usage is not supported, and that `resolver_method` must be used instead

**Actual behavior**

Error `wrong number of arguments (given 0, expected 1)` is returned

joelzwarrington:
Fixes #3944
Fixes #3861

## Goal

Clean up the campground and refactoring the methods a field will use to resolve 🏕️

## Tasks / Todo

- [x] Remove clobbering of `@resolver_method`, `@method_sym`, `@method_str`
- [x] Confirm memory footprint isn't impacted significantly per #3959
- [x] Add test cases demonstrating #3944 was fixed


joelzwarrington:
@rmosolgo, mind letting me know if you'd be open to this approach? If not, I can look at updating the conditional branches and also likely closing https://github.com/rmosolgo/graphql-ruby/issues/3861 in the process

joelzwarrington:
Ugh, looks like specs are failing, and for the exact reason I thought this was hacky, using a field `name` would cause issues because `Object` responds to it.

I'm wondering if the check `obj.object.is_a?(Hash)` should be used before `obj.respond_to?(@resolver_method)`, gonna look a bit deeper into this issue but considering the refactor of this entire method to avoid scratching my head on edge cases 😢

rmosolgo:
Honestly, I wouldn't be surprised if the fix _began_ here: 

https://github.com/rmosolgo/graphql-ruby/blob/cee50623f88970735e93c78718a7e077d1ed104a/lib/graphql/schema/field.rb#L225-L226

so far, `Field` doesn't actually _remember_ whether a `hash_key:` was specifically configured. Instead, it clobbers all those given options into a single instance variable. (er, perhaps worse, _two_, since it makes a symbol _and_ a string.) I bet if you retained `if hash_key; @hash_key = hash_key; end`, then, in `public_send_field`, checked `if defined?(@hash_key); object[@hash_key]; end`, then you might be able to distinguish between method calls and key lookups 😅

joelzwarrington:
@rmosolgo would you mind giving early feedback on the approach I've started towards?

I would appreciate feedback in regards to:
- Do you think the approach of adding a separate class to handle resolving a field okay?
- Any suggestions as to when and how the resolver should raise errors when the underlying object doesn't respond to a method, or doesn't have a key?
- What should be accessible with `attr_reader`? I see some things already accessible which would likely break with this approach (such as `resolver_method`, `method_str`, `method_sym`)
  - Also does anything but `Field` and `FieldResolver` actually need to be aware of these? 
- Do you think methods such as `with_extensions` and `run_extensions_before_resolve` should be moved into `FieldResolver` (if the separate class is the right approach)

My goal is to get to a place where `Field#resolve` calls `FieldResolver` directly instead of `public_send_field`

(also you're more than welcome to directly commit to this branch)

rmosolgo:
> adding a separate class

Personally, I see two hangups: 

1. It adds to the memory footprint of a schema. For "normal" Rails apps, this doesn't matter, but at GitHub, we had a problem where, the bigger the schema, the more memory it needed, and the more memory our enterprise customers had to provided to run GitHub on their infrastructure. A lot of the worst offenders were cleaned up years ago, but I'm still sensitive to `GraphQL::Schema::Field` and `GraphQL::Schema::Argument` especially, since a big schema has _lots_ of both of those. Adding a new instance per `GraphQL::Schema::Field` might add a lot of overhead. It also might not -- but I'd want to confirm that before adding something to _every_ field definition. 
2. Design-wise, it's not the tradeoff I would make. I recognize it provides some isolation from the rest of `GraphQL::Schema::Field`, but in my opinion, the isolation isn't worth the indirection. I would rather slap all the implementation inside a single giant method. Even though the method itself becomes long and ungainly, I appreciate the simplicity of scanning it from top-to-bottom in order to grok the code. (I recognize that even _that_ simplicity is somewhat false comfort, since the code might call _any_ `Field` method, entangling it with all the rest of the code in the class -- but still, it's the tradeoff _I'd_ make 🤷 ) 

    "A Philosophy of Software Design" would call it a "pass-through method": 
    
    > A pass-through method is one that does little except invoke another method, whose signature is similar or identical to that of the calling method... This typically indicates that there is not a clean division of responsibility between the classes. 
    
    But maybe as you mention, moving `with_extensions` into the other class would address that smell. It's fine to move `with_extensions` -- AFAIK it's a totally private API.  

Anyhow, I wouldn't worry about that until after the tests are passing -- any refactor would come easy enough.

> when and how the resolver should raise errors

In my experience, the more the better. I expect any current behaviors to have tests, and I'd like to maintain those (or improve them). What counts is: 

- Describing _what_ went wrong in as much detail as possible (what the gem was _trying_ to do, and what kind of failure it had) 
- Describing _where_ as much as possible (where in the graphql schema, where in the ruby code)
- Describing _how_ to fix it as much as possible (can some config be added or removed? Does a method need to be implemented or improved in some way?) 

(https://www.morling.dev/blog/whats-in-a-good-error-message/)

Not all the gem's error messages do that well... but it's a goal! 

> accessible with `attr_reader`

It's essential to retain compatibility. There are enough apps using GraphQL-Ruby that _someone_ will be disappointed if any of the existing readers are broken! However, the implementation of those methods could change. (For example, instead of creating `method_str` during initialization, it could be deferred to the first time that the method is called.) 

> `Field#resolve` calls `FieldResolver` directly 

One possibility here, to increase isolation without adding memory overhead, would be to make `FieldResolver` a _module_ and move the implementation into a class method (`def self.resolve_field(field, args, ctx)`, or something). That would give you a "container" for methods and an assurance that any method calls to `field` are _explicit_, and use only public methods, but it wouldn't add any performance downside to `Field`.

joelzwarrington:
@rmosolgo, thanks for your wonderful feedback! It'll help me get this PR in a better shape.

A few questions:

> It adds to the memory footprint of a schema

I like your suggestion re: module instead of class, and will take that approach. How might one compare memory allocation if there were to be another class? If this is a somewhat convoluted process, I wouldn't worry about showing me how, but thought you might have a quick reference/link to look at.

> Not all the gem's error messages do that well... but it's a goal!

Do you think the error messages for field resolving to be one of those? It tells you what type and field failed, as well as what it tried to resolve with. I've found it to be helpful, but have seen it not be as easy to understand when not working in GraphQL-Ruby all the time. Maybe the thing that could be improved here, is mentioning how the field was attempted to be resolved if they're all split out so the error message is more specific.

---

I'll get to updating this PR with your feedback in mind and will likely ping you if I have anymore questions or want something reviewed.

rmosolgo:
> How might one compare memory allocation

There isn't currently a way... but there _should_ be! I'll get it started in https://github.com/rmosolgo/graphql-ruby/pull/3959

As for the field resolution error, yes, I'd love to see any improvements you might have in mind!

joelzwarrington:
@rmosolgo I think this is ready for a review! I decided not to get too crazy with this PR, and make some small incremental changes.

I ran the benchmark provided and got the following compared to latest master:

```diff
# master
+ Total allocated: 3079118 bytes (34855 objects)
+ Total retained:  874962 bytes (5066 objects)

# this PR
- Total allocated: 3107342 bytes (35359 objects)
- Total retained:  903186 bytes (5570 objects)
```

which equates to about 1% increase in allocated bytes which I think is not a significant regression?

lib/graphql/schema/field.rb
```diff
@@ -233,14 +233,13 @@ def initialize(type: nil, name: nil, owner: nil, null: true, description: :not_g
...
 
-        # TODO: I think non-string/symbol hash keys are wrongly normalized (eg `1` will not work)
-        method_name = method || hash_key || name_s
+        @hash_key = hash_key
         @dig_keys = dig
-        resolver_method ||= name_s.to_sym
+        @object_method = method
+        @resolver_method = (resolver_method || hash_key || method || name_s).to_sym
+        @method_sym = @resolver_method
+        @method_str = @resolver_method.to_s
```
rmosolgo, line 241: This should continue using `String#-@` ([doc](https://ruby-doc.org/core-3.0.2/String.html#method-i-2D-40)) -- that might improve the retained memory in the benchmark, too. 

```suggestion
        @method_str = -@resolver_method.to_s
```

lib/graphql/schema/field.rb
```diff
@@ -615,59 +614,73 @@ def resolve(object, args, query_ctx)
...
+
+              if inner_object.is_a?(Hash)
                 if @dig_keys
                   inner_object.dig(*@dig_keys)
-                elsif inner_object.key?(@method_sym)
-                  inner_object[@method_sym]
+                elsif inner_object.key?(method_or_key)
+                  inner_object[method_or_key]
+                elsif inner_object.key?(method_or_key.to_s)
+                  inner_object[method_or_key.to_s]
```
rmosolgo, line 652: It looks like this will create new strings every time the field is resolved. For big responses, this will add up, performance-wise. The previous implementation didn't create any new strings while running -- can we retain that advantage here?

lib/graphql/schema/field.rb
```diff
@@ -615,59 +614,73 @@ def resolve(object, args, query_ctx)
...
+              # -> 2.3 Resolver Method
+              # -> 2.4 Field Name (as a key) - through @resolver_method
+              # 3 Resolver
+              # -> 2.1 Hash Key
+              # -> 2.2 Object Method
+              # -> 2.3 Resolver Method
+              # -> 2.4 Field Name (as a key) - through @resolver_method
+
+              inner_object = obj.object
+              method_or_key = @hash_key&.to_sym || @object_method&.to_sym || @resolver_method&.to_sym
```
rmosolgo, line 644: At least `@resolver_method` is already know to be a Symbol, right? Can we remove the coercion of values that we already know to be the desired type, to reduce ambiguity in the code? (A reader might think that `@resolver_method` _isn't_ a Symbol, since it's converted to a Symbol here.)

rmosolgo (commented):
Coming right along! I left a few more suggestions regarding performance, which is especially relevant on this hot codepath. Feel free to reality-check them against the benchmarks in the project (or your own) and let me know what you find -- it's certainly possible that my concerns are unfounded.

lib/graphql/schema/field.rb
joelzwarrington: f8a7a8552

lib/graphql/schema/field.rb
joelzwarrington: 669166a2c

lib/graphql/schema/field.rb
joelzwarrington: 4ea87cdf7

joelzwarrington (commented):
@rmosolgo, updated based on your feedback.

I refactored a bit so that some logic in `public_send` is instead done at initialization time, for the purpose of removing any duplicate string/symbol assignments. Weirdly enough running the memory benchmark again didn't show any change.

I'm running `bundle exec rake bench:profile_schema_memory_footprint` is that right? any other benchmarks I should use? I tried running `bundle exec rake bench:profile` but was met with:

```
$ bundle exec rake bench:profile

/Users/joelwarrington/.rvm/gems/ruby-2.7.2/gems/pry-byebug-3.8.0/lib/pry-byebug/control_d_handler.rb:5: warning: control_d_handler's arity of 2 parameters was deprecated (eval_string, pry_instance). Now it gets passed just 1 parameter (pry_instance)
rake aborted!
: `__Type.kind` returned `#<GraphQL::TypeKinds::TypeKind:0x00007fed6b090488 @name="INPUT_OBJECT", @abstract=false, @fields=false, @wraps=false, @input=true, @composite=false, @description="Indicates this type is an input object. `inputFields` is a valid field.">` at `__schema.types.0.kind`, but this isn't a valid value for `__TypeKind`. Update the field or resolver to return one of `__TypeKind`'s values instead.
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/schema/enum.rb:143:in `coerce_result'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:684:in `continue_field'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:516:in `block (3 levels) in evaluate_selection_with_args'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:905:in `after_lazy'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:513:in `block (2 levels) in evaluate_selection_with_args'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:806:in `call_method_on_directives'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:498:in `block in evaluate_selection_with_args'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:905:in `after_lazy'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:436:in `evaluate_selection_with_args'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:425:in `evaluate_selection'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/execution/interpreter/runtime.rb:358:in `block (2 levels) in evaluate_selections'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/dataloader.rb:181:in `block in run'
/Users/joelwarrington/workspace/graphql-ruby/lib/graphql/dataloader.rb:303:in `block in spawn_fiber'
Tasks: TOP => bench:profile
```

lib/graphql/schema/field.rb
```diff
@@ -615,59 +612,72 @@ def resolve(object, args, query_ctx)
...
+              # 2 Object
+              # -> 2.1 Hash Key
+              # -> 2.2 Object Method
+              # -> 2.3 Resolver Method
+              # -> 2.4 Field Name (as a key) - through @resolver_method
+              # 3 Resolver
+              # -> 2.1 Hash Key
+              # -> 2.2 Object Method
+              # -> 2.3 Resolver Method
+              # -> 2.4 Field Name (as a key) - through @resolver_method
```
joelzwarrington, line 639: I left this comment for an explanation of my approach and what I thought the priority was, should I leave this here or remove it?

joelzwarrington:
@rmosolgo any chance I could get a review on this or when I might be able to receive one?

lib/graphql/schema/field.rb
rmosolgo: :+1: fine with me to keep it, although I can't promise it'll be faithfully updated 🙈 !

lib/graphql/schema/field.rb
```diff
@@ -615,59 +612,72 @@ def resolve(object, args, query_ctx)
...
+
+              inner_object = obj.object
+              
+              if inner_object.is_a?(Hash)
```
rmosolgo, line 643: At first I was thinking that the inversion of these conditions would break cases where fields resolved with _methods_ on returned Hashes. But I wrote up some tests and found  out that doesn't work on master, either. It seems like there's no way to configure it to call a method on a Hash, which has apparently been fine for a long time 🤷

lib/graphql/schema/field.rb
```diff
@@ -615,59 +612,72 @@ def resolve(object, args, query_ctx)
...
                 if @dig_keys
                   inner_object.dig(*@dig_keys)
                 elsif inner_object.key?(@method_sym)
-                  inner_object[@method_sym]
-                else
+                  inner_object[method_sym]
```
rmosolgo, line 647: Out of curiousity, any reason this one uses the accessor instead of the instance variable? 

```suggestion
                  inner_object[@method_sym]
```

rmosolgo (approved):
:+1: LGTM, thanks again for your work on this!

lib/graphql/schema/field.rb
rmosolgo: 😱 I have no idea why committing this caused the build to break ... maybe because master changed in the meantime? 😖

joelzwarrington: > Out of curiousity, any reason this one uses the accessor instead of the instance variable? 
> 
> 
> 
> ```suggestion
> 
>                   inner_object[@method_sym]
> 
> ```

Thanks for catching that, I did not mean to use the accessor instead of the instance variable, so thanks for fixing that 😊

joelzwarrington: @rmosolgo I merged latest master in and ran `bundle exec rake test TEST=spec/graphql/schema/field_spec.rb` without issue, is there something I'm missing here?

joelzwarrington: Ahh nevermind, there are failures in other tests, I'll have a look and investigate :eyes:

joelzwarrington:
@rmosolgo was hoping you might be able to help me out and further troubleshoot issues with my approach?

I have found that the current approach ([53c2cde](https://github.com/rmosolgo/graphql-ruby/pull/3947/commits/53c2cdefb57d88bbad317efc40f7f29a60dd7b6b)), works well in most cases but seems to break for introspection fields such as `__Field` and `__Type`

Switching the ordering of which object to look at first (`obj` vs. `inner_object`) also yields significantly different errors.

I'm not too sure where to continue troubleshooting and what to investigate, any thoughts?

rmosolgo:
My suggestion would be to take a two step approach: 

- First, add a failing test and the _minimal_ change required to make that test pass (without breaking the others). Hacks allowed here -- it's just something to get the behavior right 
- Then, re-read the resulting (_known_-to-be-working) code and consider refactoring it. 

For me, I'm not able to tell whether these bugs come from the logic change in `#initialize` or in `#resolve` -- and with so much going on, I'm not sure how to begin figuring it out!

PR #3951, opened 2022-02-18
Update enum.md missing extension on enum base class definition

aalvarado:
Added missing extension to file name on base enum definition


rmosolgo:
Thanks for the fix!

PR #3957, opened 2022-02-22
Try reducing stack trace size during queries

rmosolgo:
TODO: 

- Gather some before benchmarks on memory consumed during runtime (not just limits, but also how much is GC'ed) and stack trace size at certain points 
- See if this improves either measure
- The real work is in runtime.rb -- reducing recursive calls and nesting.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -116,6 +116,18 @@ namespace :bench do
     prepare_benchmark
     GraphQLBenchmark.profile_batch_loaders
   end
+
+  desc "Check the memory footprint of a large schema"
+  task :profile_schema_memory_footprint do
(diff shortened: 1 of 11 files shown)


rmosolgo:
New benchmark shows a very small reduction 😆 

```
Result         Depth
---------------------
Eager          229
Lazy           107
Very Lazy      79
```

rmosolgo:
New benchmark shows: 

```
Result         Depth
---------------------
Eager          196
Lazy           74
Very Lazy      45
```

rmosolgo:
For reference, the "before" profile is here: https://github.com/rmosolgo/graphql-ruby/pull/3968

PR #3961, opened 2022-02-23
Remove an unnecessary `require_relative`

CvX:
This was causing a double require which resulted in warnings:

```
(…)/query_complexity.rb:22: warning: already initialized constant GraphQL::Analysis::AST::QueryComplexity::ScopedTypeComplexity::HASH_CHILDREN
(…)/query_complexity.rb:22: warning: previous definition of HASH_CHILDREN was here
```

### DIFF
--- a/lib/graphql/analysis/ast/max_query_complexity.rb
+++ b/lib/graphql/analysis/ast/max_query_complexity.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require_relative "./query_complexity"
 module GraphQL
   module Analysis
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: To confirm, this is needless because of 

https://github.com/rmosolgo/graphql-ruby/blob/ea1c404a0c4e2a29d7e907ff8aace2ae57d3898d/lib/graphql/analysis/ast.rb#L5-L6

rmosolgo:
Thanks for the improvement!

PR #3962, opened 2022-02-25
Fix cycles in `generate graphql:install`

awall:
There are two circular dependencies in the code generated from `rails generate graphql:install`. Whenever running Rails with `eager_load` of `true` (e.g., in production), you will encounter a bug, complaining that `Types::BaseObject` can not be found (because BaseObject needs BaseEdge which need BaseObject...).

I suspect the intention was to derive from `GraphQL::Schema::Object`, rather than `BaseObject` for both of these circular dependencies. The updated version works according to my testing, and it also matches the pattern of `Relay::Types::BaseEdge` and `Relay::Types::BaseConnection`.

### DIFF
--- a/lib/generators/graphql/templates/base_connection.erb
+++ b/lib/generators/graphql/templates/base_connection.erb
@@ -4,5 +4,9 @@ module Types
     # add `nodes` and `pageInfo` fields, as well as `edge_type(...)` and `node_nullable(...)` overrides
     include GraphQL::Types::Relay::ConnectionBehaviors
   end
+
+  class BaseObject
+    connection_type_class(Types::BaseConnection)
+  end
 end
 <% end -%>
--- a/lib/generators/graphql/templates/base_edge.erb
+++ b/lib/generators/graphql/templates/base_edge.erb
@@ -4,5 +4,9 @@ module Types
     # add `node` and `cursor` fields, as well as `node_type(...)` override
     include GraphQL::Types::Relay::EdgeBehaviors
   end
+
+  class BaseObject
+    edge_type_class(Types::BaseEdge)
+  end
 end
 <% end -%>
--- a/lib/generators/graphql/templates/base_object.erb
+++ b/lib/generators/graphql/templates/base_object.erb
@@ -4,4 +4,7 @@ module Types
     field_class Types::BaseField
   end
 end
+
+require_relative "base_edge.rb"
+require_relative "base_connection.rb"
 <% end -%>


rmosolgo:
Hmm, sorry for the trouble! I would have expected Rails's autoloader to handle this fine, where it references those classes even though the file hasn't finished loading yet. 

Unfortunately, `Types::BaseObject` really _is_ the right thing for the base Edge and base Connection classes -- we want those classes to inherit any application-specific customizations from `Types::BaseObject`. 

Perhaps another approach would be: 

- Remove `edge_type_class ...` and `connection_type_class ...` configs from `Types::BaseObject` 
- Add `Types::BaseObject.edge_type_class(Types::BaseEdge)` to the end of `base_edge.rb`; likewise for BaseConnection 
- At the end of `base_object.rb`, add `require ...` statements to load BaseEdge and BaseConnection. (We have to require them manually since they aren't named by BaseObject anymore.) 

awall:
Great idea! I am on vacation and won't look at this for another week or so,
but that sounds like a solid approach.

awall:
@rmosolgo I used your approach, still appears to work from my testing, and now we inherit the user-specific changes from BaseObject. :)

lib/generators/graphql/templates/base_connection.erb
```diff
@@ -4,5 +4,9 @@ module Types
     # add `nodes` and `pageInfo` fields, as well as `edge_type(...)` and `node_nullable(...)` overrides
     include GraphQL::Types::Relay::ConnectionBehaviors
   end
+
+  class BaseObject
```
rmosolgo, line 8: I'm surprised that reopening this class would work without re-declaring its super class, for example: 

```suggestion
  class BaseObject < GraphQL::Schema::Object 
```

But, is this code working for you in your project?

awall: The code is working in my project. I couldn't find any clear documentation on whether or not you must specify the subclass when re-opening a class: my experiments show that it work. Consider the following code:

```
class A
  def hi
    puts "hi"
  end
end

class B < A
  def bye
    puts "bye"
  end
end

class B
  def aloha
    puts "aloha"
  end
end


b = B.new

b.hi
b.bye
b.aloha
puts b.is_a?(A)
```

Output:
```
hi
bye
aloha
true
```

rmosolgo:
I'm going to close this since it's gone cold, but feel free to push this work as a new branch and open a new PR if you decide to pick it up again!

rystraum:
I know this PR was closed, but just wanted to add that applying the PR changes to my app fixed production issues on circular dependencies on load so might probably be worth reopening?

PR #3964, opened 2022-02-26
Link to fiber scheduler list

bruno-:
Hi,

a `graphql-ruby` user pointed me to these docs https://graphql-ruby.org/dataloader/nonblocking.html

I've been playing with the [Fiber Scheduler](https://brunosutic.com/blog/ruby-fiber-scheduler) for the last couple weeks and I'm impressed you have a gem that's already using it. I literally thought no one in the WORLD has figured this out yet.

This PR proposes linking to [Fiber Scheduler List](https://github.com/bruno-/fiber_scheduler_list) instead of maintaining a list of available fiber schedulers in the [nonblocking dataloader doc](https://graphql-ruby.org/dataloader/nonblocking.html).

### DIFF
--- a/guides/dataloader/nonblocking.md
+++ b/guides/dataloader/nonblocking.md
@@ -33,7 +33,7 @@ Fiber.set_scheduler(MySchedulerImplementation.new)
 MySchema.execute(...)
 ```
 
-The scheduler must implement [`Fiber::SchedulerImplementation`](https://ruby-doc.org/core-3.0.2/Fiber/SchedulerInterface.html). Some implementations are available off-the-shelf:
+The scheduler must implement [`Fiber::SchedulerImplementation`](https://ruby-doc.org/core-3.0.2/Fiber/SchedulerInterface.html). Existing implementations can be found at [Fiber Scheduler List](https://github.com/bruno-/fiber_scheduler_list).
 
 - [`dsh0416/evt`](https://github.com/dsh0416/evt) supports several backends in `Evt::Scheduler`.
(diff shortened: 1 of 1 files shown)


bruno-:
If you're in favor of this change, I propose also adding a note that this gem only works with Fiber Schedulers that are using [Fiber.yield](https://github.com/rmosolgo/graphql-ruby/blob/9433eff6317b8c88445556d62433142156698cfa/lib/graphql/dataloader.rb#L95) and [Fiber#resume](https://github.com/rmosolgo/graphql-ruby/blob/9433eff6317b8c88445556d62433142156698cfa/lib/graphql/dataloader.rb#L277) for control passing.

I also saw this in the docs:

> ⚠️ [socketry/async](https://github.com/socketry/async) contains Async::Scheduler, but I haven’t figured out how to make it work with GraphQL::Dataloader yet. Please update this doc if you know how to!

`async` works with `Async::Task`s which wrap fibers. In practice, you can't just do `Fiber.new(blocking: false) { }` and expect it to work with `Async::Scheduler`like it's [done here](https://github.com/rmosolgo/graphql-ruby/blob/9433eff6317b8c88445556d62433142156698cfa/lib/graphql/dataloader.rb#L296).

Instead, the recommended approach when working with a `Fiber.scheduler` is to use [Fiber.schedule { ... }](https://docs.ruby-lang.org/en/3.1/Fiber.html#method-c-schedule) method. It will delegate to `Fiber::SchedulerInterface#fiber` method which [will do the right thing with Async::Scheduler](https://github.com/socketry/async/blob/8b9664be61debcb687c5bab23ed68faacd82376a/lib/async/scheduler.rb#L102-L110).

I'm probably just a smartass, just talking and not opening PRs, but as you probably know working with fibers is **bloody hard work**, so hopefully the comments above will help.

rmosolgo:
Thanks for the update! Looks good to me. 

Yeah, Fiber scheduler was hard to get the hang of -- segfaults here and there, when I did it wrong -- but it _worked_! 

Thanks for the pointer about `Async`. I didn't realize `nonblocking: true` wasn't the best way to do it. I bet the code can be redone to use `Fiber.schedule` instead. I'd love to support `Async` properly.

PR #3973, opened 2022-03-03
Update mutation_errors.md

jnankin:
To benefit from "Errors as Data" described above, mutation fields must not have `null: true`.

### DIFF
--- a/guides/mutations/mutation_errors.md
+++ b/guides/mutations/mutation_errors.md
@@ -129,7 +129,7 @@ Then, client apps can show the error messages to end users, so they might correc
 
 ## Nullable Mutation Payload Fields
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! Could you please help me understand this change, or share some code that's working for you? 

In my experience, mutation results and mutation errors are mutually exclusive: either you have results _or_ you have errors. If you have errors, then the result fields will be `null`. But if you configure `null: false,` then returning `nil` for any result fields will cause the whole response to be removed (since `null` would be invalid where `null: false` is configured). But, did you find something else that worked for you?

rmosolgo:
I'd be happy to consider this change further, just let me know what you think about the questions above! Closing til then.

PR #3978, opened 2022-03-07
support for ddtrace 1.0 and keep backward compatibility

danielRomero:
Datadog tracing is about to release the 1.0 version with some [changes](https://github.com/DataDog/dd-trace-rb/blob/v1.0.0.beta1/docs/UpgradeGuide.md).

The current integration is affected with one of this changes (the one about the [namespacing](https://github.com/DataDog/dd-trace-rb/blob/v1.0.0.beta1/docs/UpgradeGuide.md#namespacing--the-public-api))

This change fix the error and also keeps backward compatibility with older versions of the ddtrace gem

### DIFF
--- a/lib/graphql/tracing/data_dog_tracing.rb
+++ b/lib/graphql/tracing/data_dog_tracing.rb
@@ -48,11 +48,9 @@ module GraphQL
       end
 
       def tracer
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the fix on this!

PR #3979, opened 2022-03-09
Remove branch that uses code removed in 2.0

Tiedye:
BaseConnection was removed in 2.0 so that branch throws an error.  Context is now required

### DIFF
--- a/lib/graphql/relay/range_add.rb
+++ b/lib/graphql/relay/range_add.rb
@@ -35,10 +35,6 @@ module GraphQL
       # @param parent [Object] The owner of `collection`, will be passed to the connection if provided
       # @param edge_class [Class] The class to wrap `item` with (defaults to the connection's edge class)
       def initialize(collection:, item:, context:, parent: nil, edge_class: nil)
-        unless context.schema.new_connections?
-          raise ArgumentError, "No connections are defined for context.schema"
-        end
-
         conn_class = context.schema.connections.wrapper_for(collection)
         # The rest will be added by ConnectionExtension
         @connection = conn_class.new(collection, parent: parent, context: context, edge_class: edge_class)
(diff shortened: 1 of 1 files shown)


lib/graphql/relay/range_add.rb
```diff
@@ -31,25 +31,22 @@ class RangeAdd
...
-          # Check if this connection supports it, to support old versions of GraphQL-Pro
-          @edge = if @connection.respond_to?(:range_add_edge)
-            @connection.range_add_edge(item)
-          else
-            @connection.edge_class.new(item, @connection)
-          end
+      def initialize(collection:, item:, context:, parent: nil, edge_class: nil)
+        unless context.schema.new_connections?
+          raise ArgumentError, "No connections are defined for context.schema"
+        end
```
rmosolgo, line 40: I _think_ this could only happen if someone set `self.connections = nil` in their schema definition. IMO we don't have to handle that case nicely (I've never seen it before), do you think it'd be ok to remove this? Or does it handle another situation I haven't thought of?

rmosolgo (commented):
Thanks for the update here! I just have one question about the change. 

By the way, was this code giving you a warning in 1.13.x? (If not, I'll update 1.13 to warn in this case...)

lib/graphql/relay/range_add.rb
Tiedye: I'm not sure about this case, I didn't want to remove the check that already existed!  I'll remove it

Tiedye:
> By the way, was this code giving you a warning in 1.13.x? (If not, I'll update 1.13 to warn in this case...)

I don't believe it was.

For context we updated to 2.0.2 and this broke:
```ruby
GraphQL::Relay::RangeAdd.new(
  collection: job.job_hidden_stage_links,
  item: link,
).edge
```
adding the context fixed it

rmosolgo:
Thanks again for this fix -- I opened an issue to track that deprecation in 1.13.x :+1:

PR #3985, opened 2022-03-15
Update the max supported Ruby version to 3.1

petergoldstein:
This is actually what's already running because an unquoted '3.0' in the GitHub Actions is truncated to '3', loading the latest Ruby 3 - which at time of writing is Ruby 3.1.1.  This PR makes the change explicit.


petergoldstein:
You can see that 3.1.1 is what's already running in CI here:

Screen Shot 2022-03-14 at 5 15 33 PM

rmosolgo:
Thanks!

PR #3986, opened 2022-03-16
Allow multiples of repeatable directives

geshwho:
### What's up

When adding directives to an object, we currently overwrite any existing directives of the same class. However, if the directive is repeatable, we should be able to specify multiple directives without overwriting existing ones.

### Solution

Pretty simple: Don't remove existing directives if the directive in question is repeatable.

#### Caveats

I haven't explicitly checked the printer to make sure it does the right thing when printing multiple of the same directives. I assume it isn't really any different than if the object had multiple different directives.


rmosolgo:
:+1: Thanks for this fix! Likewise, I assume the printer won't have any trouble with it.

Issue #3173, opened 2020-09-30
Authorising Scope for non Active Record collections via Elastic Search


ashikajith:
**Unable to scope records returning from Elastic Search since they are Non Active Record collections**

I have a custom connection to fetch a collection and I'm using Elastic Search to retrieve that records. Since it's returning the collection in plain Array, it is not calling the designated Policy class and because of this, I'm unable to scope the collection.

Is there any workaround in which I can apply the scope for the collection which is Non AR? I'm using pundit for authorization [link](https://graphql-ruby.org/authorization/pundit_integration.html#scopes)

eg:
```ruby
# Query Type
module Types
  class QueryType < ::GraphQL::Schema::Object
    field :records_by_category_search, Types::Connections::RecordByCategorySearchConnection, null: false do
      argument :filter, String, "The filter string", required: false
    end

    def records_by_category_search(**args)
       Searcher::Elasticsearch.new(args).fetch_records
    end
  end
end

# Connection
class Types::Connections::RecordSearchConnection < Types::BaseConnection
  edge_type(Types::Edges::RecordSearchEdge) 
  field :filter_options, Types::Objects::FilterType, null: true

  def filter_options
    some_method
  end
end

# Node
class Types::Edges::RecordByCategoryEdge < Types::BaseEdge
  node_type(Types::Objects::RecordByCategoryType)
end

# Object Type
module Types
  module Objects
    class RecordByCategoryType < BaseObject
      field :category, Types::Objects::CategoryType, null: true
      field :records, [Types::Objects::RecordType], null: true # -> Unable to add scope inside RecordPolicy class
    end
  end
end
```

**Versions**

`graphql` version: 1.10.7
`rails`: 5.2.4
`graphql-pro`: 1.11.0

Hope the code example is helpful 
Thanks in Advance

rmosolgo:
Hey, thanks for the detailed writeup. I can think of two basic approaches: 

1. After returning an array, use Pundit to scope the array. You'll have to do two things to do this: 

  - Add a `pundit_policy_class: ...` to the field, because otherwise, Pundit will misinterpret the array of items as a "namespaced object" (#2008). 
  - Inside `RecordType.scope`, call `scope_by_pundit_policy(context, items)` (which the pundit integration _usually_ skips with arrays, because of the misinterpretation described above): 

  ```ruby 
  class Types::Objects::RecordType < ... 
    def self.scope(context, items)
      if items.is_a?(Array)
[...]

PR #3994, opened 2022-03-19
Don't re-authorize objects that were part of a scoped list

### LINKED ISSUE #3893: Pundit Integration - Scope and Object Authorization

The request relates to graphql pundit integration and the way the scopes and object authorization works together.

We are currently experiencing some performance issues with the way that the pundit integration works. Currently once an association has been scoped which should remove all unauthorized objects. We then check for each of the objects to see if they are authorized. This should have already been covered by the scope.

This would only work in the case where you authorize your objects with a `show?` like action, which we expect to have equal restrictiveness to the scope. In REST nomenclature that's easy, its show?, but in GraphQL there is no inherent equivalent, but we went with `view?`

If you look at the scope docs from pundit https://github.com/varvet/pundit#scopes once a record has been scoped the gem doesn't try to then authorize the individual objects.

I have put together a minimal reproduction steps to make it a little clearer what i mean

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'activerecord'
  gem 'graphql', '1.13.6'
  gem 'graphql-pro', '1.21.1', source: 'https://gems.graphql.pro/'
  gem 'pundit', '2.0.1'
  gem 'sqlite3'
end

require "active_record"

ActiveRecord::Base.establish_connection(
  :adapter => "sqlite3",
  :database  => ":memory:"
)

ActiveRecord::Schema.define do
  create_table :items do |table|
  end
end

class Item < ActiveRecord::Base
end

2.times { |n| Item.create }

class ApplicationPolicy
  def initialize(me, object)
  end

  class Scope
    def initialize(_account, scope)
      @scope = scope
    end

    def resolve
      puts "scoping"
      @scope.all
    end
  end
end

class ItemPolicy < ApplicationPolicy
  def view?
    puts "View ran"
    true
  end
end

  include GraphQL::Pro::PunditIntegration::ObjectIntegration
end

class ItemType < BaseObject
  pundit_role :view

  field :id, Integer, null: true
end

class QueryType < BaseObject
  pundit_role nil

  field :items, [ItemType], null: false, scope: true

  def items
    Item.all
  end
end

  query(QueryType)
end

query = <<-GRAPHQL
  query {
    items {
      id
    }
  }
GRAPHQL

TestSchema.execute(query, variables: {}, context: {})
```
which outputs

```
scoping
View ran
View ran
```

Ideally we would like to be able to define some kind of base authorization that the scope and a single action like `show?` agree on. Then if the relation has already been scoped and it would call that method it skips it as it knows that the scope has already restricted any items that would fail that check. This would remove the performance issues we are currently seeing.

I think ideally the only change that we would make is that somewhere we would define a `base_role` or something equivalent that if the pundit_role matches would skip the object level authorization. 

We currently have tried to use a cache that helps to make the object level authorization hit the db less but in an ideal world we wouldn't have to maintain that cache that even tho is more efficient than hitting the db is still wasteful. 

I hope that all makes sense, let me know what you think. Im happy to help in any way i can if you think the idea is of merit :)

rmosolgo:
After a list has been scoped, don't re-run authorization on the objects it permitted. 

Fixes #3893 

TODO: 

- [x] consider the breaking-ness of this change. Make it opt-in instead of default? 
- [x] somehow pass this thru `connection -> edges -> node` and `connection -> nodes`
- [x] use `.equal?` to test same-object, add a test for that behavior


fameoflight:
Would love to have this available in next version. We have usecase for this as well. Let me know if I can help in anyway.

rmosolgo:
Sure thing, I'll try to wrap it up soon. The trick is determining _when_ an application-defined `scope_items` method was applied. I've been trying to use the return value, but I don't think that's right: sometimes, applying `scope_items` will return the exact same list that was passed in. 

Instead, I think I'll try inspecting `.method(:scope_items)` to see if it's the default implementation or a custom implementation. The other option is, in the default implementation, checking `caller` to see if an application-defined method is calling through to it. (Maybe both of those approaches will be required.)

bessey:
@hvenables left the company (enjoy Shopify Harry :stuck_out_tongue:) so I'll have to fill in on feedback! This definitely works for us. Our GraphQL Pundit authorization architecture is essentially

- All objects are expected to have a policy by default
- All objects are authorized with `#view?` by default (we felt `show?` was inappropriate given the non RESTy nature of GraphQL, though in practice `view?` and `show?` are aliases for cross compatibility`
- `#view?` is expected to behave the same as the policies scope, i.e. it should never return false when for scoped items

So given your API we would probably implement `reauthorize_scoped_objects(false)` on our BaseObject.

We're still very keen for this API btw! Authorization induced N+1s is one of our biggest perf issues in our GraphQL API.

rmosolgo:
Thanks for sharing your thoughts about it! I'll make some time to land this PR soon.

bessey:
I've been thinking about this more lately, in the context of #4087, because we perform the same scoping that GraphQL Pro Pundit does to ActiveRecord::Relations, within all our ActiveRecord::Relation sources.

In this context, I cannot think of a way that you could indicate to the runtime "hey, I know you're returning an instance of MyRecord, but trust me, I loaded that through a dataloader which scoped it, so you do not need to authorize it again". The only option I can think that could work is

1. Set some state on the ActiveRecord::Relation object
2. Modify ActiveRecord's behaviour to persist this state through to all records instantiated from that Relation

e.g. theoretical API 

```ruby
scope = MyRecord.where(visible: true).mark_authorized!(:view) # sets @authorized = :view
results = scope.to_a
results.first.pre_authorized?(:view) => true
```

I have no idea if this is actually possible without monkeypatching ActiveRecord yet, but I thought I'd put my thoughts somewhere others with more experience might see!

ioquatix:
We appear to be running into the same issue, any chance this will make it into a release soon? Thanks!

rmosolgo:
I'll address the lint failure in another branch...

ioquatix:
Thanks for merging this. Any chance we can get an ETA on when it would potentially be released?

rmosolgo:
I just _finally_ released this in GraphQL-Ruby 2.1.0 😅

ioquatix:
Thanks, we will try it out and report back!

lib/graphql/schema/field/scope_extension.rb
```diff
@@ -10,7 +10,13 @@ def after_resolve(object:, arguments:, context:, value:, memo:)
           else
             ret_type = @field.type.unwrap
             if ret_type.respond_to?(:scope_items)
-              ret_type.scope_items(value, context)
+              scoped_items = ret_type.scope_items(value, context)
+              if !scoped_items.equal?(value) && !ret_type.reauthorize_scoped_objects
```
bmulholland, line 14: Per my comment, perhaps this should also check the type of value, so that it doesn't skip authorizing arrays twice?

rmosolgo:
Hey @bmulholland, thanks for pointing out that issue with the Changelog. I updated it in a3c093f69. 

>  move an existing method to the dataloader pattern and bam

Could you give a before-and-after example of a refactor where this would happen? I would _expect_ lists to go through `scope_items` _regardless_ of how the data was loaded, but maybe I've missed a spot!

bmulholland:
> I updated it in https://github.com/rmosolgo/graphql-ruby/commit/a3c093f692855976b0192087a14b42afefd88a07.

Thanks for clarifying, appreciate it!

> Could you give a before-and-after example of a refactor where this would happen?

Thanks for getting me to check this again. It looks like this is only in effect [when using the Pundit integration](https://graphql-ruby.org/authorization/pundit_integration.html#scopes) -- so the footgun is limited to that scenario.

It looks like this is actually a risk for us, but only from our own app's code. That's because of GraphQL-Ruby-related history: We were trying to use the Pundit integration but honestly we could never figure out how the concepts fit together so we dropped the attempt. However, because of that, we still have explicit handling of arrays in our scope_items calls, which leads to this issue. We'll discuss the best way to handle this: either somehow ensure that the "disable authorized?" option is never enabled, or have a policy that Arrays in scope_items are always scoped too.

jderose9:
The documentation for this change would lead me to believe that the default is to reauthorize scoped objects, however after observing the behavior and examining the code it would seem that the default behavior has changed (a breaking change) and that scoped objects are no longer re-authorized, meaning that consuming code that assumed an object would be authorized by the policy method would unintentionally be exposing objects after upgrading.

The check seems to be here, and I don't see anything defaulting this value to true:
https://github.com/rmosolgo/graphql-ruby/blob/109c26c5fa9118a9c69ed4c4823048c1ce1aea6d/lib/graphql/schema/field/scope_extension.rb#L14

Additionally, unless I've got something setup incorrectly, I have found that calling `reauthorize_scoped_objects(true)` in the node type's class ([as documented here](https://github.com/rmosolgo/graphql-ruby/blob/109c26c5fa9118a9c69ed4c4823048c1ce1aea6d/guides/authorization/scoping.md?plain=1#L52)) doesn't seem to do anything. I had to call it in the connection type's class.

bmulholland:
Uh, what? That's really alarming: as noted above, this is a big risk for us. It's incredibly critical that we don't expose data to the wrong users, and we need all the checks that we can get. I think that's worth it's own Issue, right?

jderose9:
> More broadly, as I've noted a few times, there's several moving parts here that lead to a high risk for disastrous integration-level bugs: Dataloaders, array handling, and scope re-authorization could easily combine together to skip all checks, exposing all data to users. Since there are several options and configuration variations of these, this is probably hard to both test thoroughly and reason about. Even if the specifics here mean there's no/low concern, I've already had a few similar conversations on this project that have ended with that same outcome. What happens when several areas of "no concern" collide?

I agree. Also, in my situation, while I would like to get as close as possible to loading only the permitted objects for performance reasons (ie. I'm not going to load objects for a completely different tenant), sometimes there is code that runs in the policy method that is difficult or impossible to reproduce in a database scope. So certainly I don't think the default behavior should have changed.

jderose9:
@rmosolgo Awesome, thank you very much for the fast resolution!

bessey:
> More broadly, as I've noted a few times, there's several moving parts here that lead to a high risk for disastrous integration-level bugs: Dataloaders, array handling, and scope re-authorization could easily combine together to skip all checks, exposing all data to users

I share this concern of yours @bmulholland. We are GraphQL Pro (and Ent) users and make heavy use of Dataloaders and the GraphQL Pro Pundit integration. I tried to adopt this new optimisation in a spike, but ran into a problem like yours: it has no effect for Dataloaders. I believe because they (or at least ours) load an Array, and Pundit can't do anything useful with scoping an already instantiated Array. 

That's not a security issue at least, but it does mean these two GraphQL Ruby provided tools are not compatible with one another.

I've been doing a lot of GraphQL Ruby perf optimisation work lately on our codebase, and I keep coming back to one thing: reasoning about this stuff is getting incredibly difficult. I have used GraphQL Ruby for years and still I have a hard time reasoning about how these factors combine:

(I know the answers to these Qs now, or so I hope, but just to illustrate how difficult this is to reason about):

- Dataloaders
  - Should a Source do its own authorization?
- Connection objects (both automatically wrapped and manually returned from resolvers)
  - Is authorization done before or after pagination?
  - What object is passed to scope_items in each case?
  - Pundit Scopes expect an ActiveRecord::Relation, what should I do with a connection object?
- The Pro Pundit integration (and more broadly the authorization framework itself)
  - Object vs field level authz
  - Scope items, and its special case for Arrays and Arrays only
  - This new scope optimisation

We've used Module#prepend to wrap the GraphQL Pro Pundit integration with verbose debug logging of the decisions it makes, and still  I regularly `bundle open` GraphQL + GraphQL Pro just to understand the control flow through these parts. 

I'm coming around to the view point that 

1. The costs of outsourcing such a rich integration as GraphQL + Pundit to a 3rd party library outweigh the benefits
2. This new scope optimisation does not belong in GraphQL Ruby at all. It is a great idea, but is better implemented by the application developer and richly integrated with their specific authorization strategy, rather than an additional layer of complexity to reason about

Is it constructive to continue this conversation, and if so is GH Discussions the best place to do so? I appreciate my problems can largely be solved by ignoring this features existence and building our own Pundit integration, but since it seems others in the community have similar concerns, perhaps its worth discussing nonetheless?

PS I don't want to come across as ungrateful, GraphQL Ruby is a wonderful piece of software ❤️

rmosolgo:
Hey @bessey and @bmulholland, thanks for your detailed writeups on the conflict between these features -- and sorry for the trouble you've run into so far on them :S I can see that they need some reconsideration. Over the next couple of days I'm going to re-read your descriptions and suggestions and open a new issue with some proposals. I'll follow up back here when I do :+1:

rmosolgo:
I went digging on the Pundit Arrays issue and found a [previous suggestion](https://github.com/rmosolgo/graphql-ruby/issues/3173#issuecomment-702736339) which seems like a good candidate for a new default behavior. I worked out the code locally and wrote up a description here: https://github.com/rmosolgo/graphql-ruby/issues/4726 

@bmulholland  and @bessey , could you let me know on that issue what you think, if you have a minute? Thanks!

iulia-b:
👋 Hey @rmosolgo. We are updating our GraphQL gem to 2.3 and want to migrate to using the `reauthorized_scoped_items` for a field A, which is a connection to a parent field P. The field A is computed through a resolver, where all the logic resides. Additionally, there is no need for authorizing this list of objects of type A once the parent (of type P) has been previously authorized.

In order to do this we want to use `reauthorize_scoped_items(false)` on type A. To match our use case, we have to return in `scope_items` a clone of the resolved items in order to bypass [this check](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/field/scope_extension.rb#L14). 

```
def self.scope_items(items, context)
  items.clone
end

reauthorize_scoped_objects(false)
```
It’s not clear to me if we are missusing this feature. Why do `scope_items` have to be different than `items` in order to skip authorization? 

Did I understand correctly the behavior? Is it possible to introduce any bugs or potential security issues by returning a clone of `items`?

rmosolgo:
Hey @iulia-b, great question. Yes, that's the best way to handle your situation. 

For lack of better implementation, GraphQL-Ruby checks `was_scoped = !prev_list.equal?(new_list)`, that is, it makes sure that `Type.scope_items(prev_list, ctx)` returns some _different_ Ruby object. That's how GraphQL-Ruby knows that scoping was applied. 

In your case, you're using the resolver to scope the list, so `scope_items` is a no-op. But, you have to return a new object so GraphQL-Ruby does the right thing. I think `.clone` is a good solution here. 

The security issue would be if any _other_ fields return lists of this same type. Do those other lists need scoping? in that case, you'd need a way of detecting whether `items` came from this particular resolver or from somewhere else (for example, you could apply a wrapper object and remove the wrapper in `.scope_items`). 

I hope this helps!

alexus37:
Thank you for the explanations, @rmosolgo! I'm curious if there might be a more efficient solution instead of cloning. Could we enhance the return type of the `scope_items` function to not only return the items but also a flag indicating if scoping was applied? Alternatively, we could consider adding a new method called `was_scoped` that allows the object to specify whether the returned items were scoped. I believe implementing either of these approaches could eliminate the need to clone items repeatedly. What do you think?

rmosolgo:
I'm open to adding some kind of flag method on the returned object, something like `graphql_list_was_scoped?`, and checking that inside GraphQL-Ruby. (I'd rather not go for multiple return ... in places where I've used that, it has added overhead and complexity 😩 )

alexus37:
I create this issue to track it:
- https://github.com/rmosolgo/graphql-ruby/issues/5051

PR #3997, opened 2022-03-21
Try to make Dataloader work with Async gem

rmosolgo:
I'm hoping to get `Dataloader` working with the `async` gem, but it seems to work a bit differently than the others. 

I tried using `Fiber.schedule { ... }`, but since it runs the fiber immediately, `Fiber.yield` doesn't work the same way from inside the fiber. 

Also, the code previously expected `spawn_fiber` to return a non-running fiber, but `Fiber.schedule` returns an already-running fiber. So, gotta handle that nicely.

cc @bruno-  Do you have any further suggestions about how this might work?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gem 'pry-byebug'
 if RUBY_VERSION >= "3.0"
   gem "libev_scheduler"
   gem "evt"
+  gem "async"
 end
 
 # Required for running `jekyll algolia ...` (via `rake site:update_search_index`)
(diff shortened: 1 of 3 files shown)


3334Aa (changes requested):
GJ

ioquatix:
I've added an example of how to make a non-async specific dataloader using `Fiber.schedule`.

https://github.com/socketry/async/compare/cfc045ea921b...b88db6f6bc8e

Issue #4003, opened 2022-03-24
Migrate dataloader to use Fiber#transfer for modern scheduler compatibility


dbackeus:
**Is your feature request related to a problem? Please describe.**

The current implementation of Dataloader uses `yield` / `resume` for Fiber management which makes it incompatible with modern Ruby 3.1 based fiber schedulers which rely on `transfer` instead (see https://github.com/bruno-/fiber_scheduler_list for a reference).

AFAICT there is no production ready Ruby 3.1 based scheduler that can be used today with the GraphQL gem.

The documentation for non-blocking dataloading at https://graphql-ruby.org/dataloader/nonblocking.html mentions "not having been able to figure out how to make Async::Scheduler work". The reason is the same - `Async::Scheduler` does not work with `yield`/ `resume`.

**Describe the solution you'd like**

I've had some Slack interaction with Bruno Sutic, author of the above scheduler list as well as the [fiber_scheduler](https://github.com/bruno-/fiber_scheduler) gem.

According to him migrating toward using `transfer` would be the right way to go for the GraphQL gem as he expects that to be the default paradigm for Fiber schedulers going forward.

I'm not sure how relevant it is to keep backwards compatibility with ruby 3.0 based schedulers. If needed then I guess we would need two different fiber management implementations and offer developers a configuration option to switch between them.

**Describe alternatives you've considered**

I spent a few hours trying to implement the change myself. I got the [async_dataloader_spec.rb](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/dataloader/async_dataloader_spec.rb) tests passing with ~~both `Async::Scheduler` and~~ `FiberScheduler` but a lot of deeper integration tests ended up failing and it appears too time consuming for me to make sense of all the parts to make progress with the time I have on hand.

rmosolgo:
Hi, thanks for sharing your research on this! 

Bummer that there are _two_ different control passing approaches. For now, I'd like to try to support both, so that anyone who's already using Fiber-based concurrency can continue using it the way they already are. 

Do you mind sharing your work on this so far? Even if it's not all passing, I'd like to see the approach you took.

dbackeus:
Sure thing: https://github.com/rmosolgo/graphql-ruby/pull/4004

Not much of an approach I'm afraid, just tried the simplest thing that could possibly work which was likely too simple :D

bruno-:
Hi,

[...]

PR #4004, opened 2022-03-24
Spike attempting to use Fiber#transfer instead of yield / resume

dbackeus:
As requested in https://github.com/rmosolgo/graphql-ruby/issues/4003 here is the diff for the spike I made for Fiber#transfer support.

I was mistaken in the issue to say that async dataloader specs were passing for both `fiber_scheduler` and `async`. In fact there are 3 failures when using `async` still. This may or may not be due to the implementation of `Async::Scheduler` itself being incomplete. According to Bruno the only gem which covers 100% of the fiber scheduler spec is his own `fiber_sheduler` gem.

As you can see it's an extremely minimal diff. A hunch as to why a lot of specs outside of `async_dataloader_spec.rb` are failing is that the `@parent_fiber` idea is too naive and needs some more sophisticated approach. I had worked neither with the GraphQL gem nor Fiber's before attempting this so felt lost pretty quickly 😅

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,9 +10,14 @@ gem 'pry-stack_explorer', platform: :ruby
 gem 'graphql-batch'
 gem 'pry-byebug'
 
-if RUBY_VERSION >= "3.0"
+if RUBY_VERSION >= "3.0" && RUBY_VERSION < "3.1"
   gem "libev_scheduler"
   gem "evt"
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for sharing your spike on this -- I'm hoping to land a similar change in #4322 !

PR #4006, opened 2022-03-25
Resolvers: properly inherit args from parent classes that aren't attached to the schema

### LINKED ISSUE #4005: Seems like #3916 broke argument inheritance in mutation resolvers?
**Describe the bug**

I have a base mutation class which implements common arguments shared by some subclasses.
With version 2.0.0, it works as it previously did.
With version 2.0.1, it no longer works, the arguments I expect to be inherited do not show up on the subclasses.

**Versions**

`graphql` version: 2.0.1
`rails` (or other framework): rails 7.0.1

**GraphQL schema**

```ruby
module Mutations
  module Carts
    class ModifyCartItemsBase < Mutations::BaseMutation
      description I18n.t('api.docs.mutation.carts.modify_items.api')

      field :errors, [Types::UserError], null: true

      argument :cart_id, ID, required: true, description: I18n.t('api.docs.mutation.carts.modify_items.args.cart_id')
      argument :details, Types::Inputs::CartDetailsInput, required: false, description: I18n.t('api.docs.types.inputs.cart_update_input.desc')
      argument :add_items, [Types::Inputs::CartAddItemInput], required: false, description: I18n.t('api.docs.mutation.carts.modify_items.args.add_items')
      argument :update_items, [Types::Inputs::CartUpdateItemInput], required: false, description: I18n.t('api.docs.mutation.carts.modify_items.args.update_items')

      def resolve(cart_id:, details: {}, add_items: [], update_items: [], email: nil)
      end
   end
  end
end

module Mutations
  module Carts
    class ModifyTeamCartItems < Mutations::Carts::ModifyCartItemsBase
      field :team_cart, Types::TeamCart, null: false

      argument :email, String, required: false, description: I18n.t('api.docs.mutation.team_carts.modify_items.args.email')

      def resolve(cart_id:, details: {}, add_items: [], update_items: [], email: nil)
      end
    end
  end
end
```

**GraphQL query**

Standard inspection of the schema.

**Steps to reproduce**

Steps to reproduce the behavior

**Expected behavior**

This is what I expect to see in the schema:
image

**Actual behavior**

This is what I actually see:
image

The breakage appears to be a result of #3916. The inherited arguments are not carrying through.

<details>

```
```

</details>

**Additional context**

With these details, we can efficiently hunt down the bug!

rmosolgo:
Fixes #4005

### DIFF
--- a/lib/graphql/schema/member/has_arguments.rb
+++ b/lib/graphql/schema/member/has_arguments.rb
@@ -148,7 +148,19 @@ module GraphQL
             end
           elsif defined?(@resolver_class) && @resolver_class
             all_defns = {}
-            all_defns.merge!(@resolver_class.own_field_arguments)
+            @resolver_class.all_field_argument_definitions.each do |arg_defn|
+              key = arg_defn.graphql_name
+              case (current_value = all_defns[key])
+              when nil
+                all_defns[key] = arg_defn
+              when Array
+                current_value << arg_defn
+              when GraphQL::Schema::Argument
+                all_defns[key] = [current_value, arg_defn]
+              else
+                raise "Invariant: Unexpected argument definition, #{current_value.class}: #{current_value.inspect}"
+              end
+            end
             all_defns.merge!(own_arguments)
           else
             all_defns = own_arguments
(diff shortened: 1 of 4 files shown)


srgoldman (approved):
Confirmed that the fix works for our use case(s).

PR #4013, opened 2022-03-30
Fix specifiedByUrl -> specifiedByURL

bubba:
The spec spells the specifiedByURL introspection field by capitalizing the acronym, and fields are case sensitive.
(Previously the inflector generated specifiedByUrl)


rmosolgo:
d'oh -- thanks for the fix on this!

PR #4014, opened 2022-03-30
Add specified_by_url for ISO8601 scalars

bubba:
Link to the RFC 3339 specification https://tools.ietf.org/html/rfc3339, which in turn specifies to use ISO 8601

### DIFF
--- a/lib/graphql/types/iso_8601_date.rb
+++ b/lib/graphql/types/iso_8601_date.rb
@@ -14,6 +14,7 @@ module GraphQL
     # own Date type.
     class ISO8601Date < GraphQL::Schema::Scalar
       description "An ISO 8601-encoded date"
+      specified_by_url "https://tools.ietf.org/html/rfc3339"
 
       # @param value [Date,Time,DateTime,String]
       # @return [String]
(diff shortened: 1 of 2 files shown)


rmosolgo (approved):
Thanks for the improvement!

sj26:
This is very nice:

```graphql
scalar ISO8601DateTime @specifiedBy(url: "https://tools.ietf.org/html/rfc3339")
```

but it seems to break when parsing IDL using earlier versions of the graphql gem:

```
No definition for @specifiedBy on ISO8601DateTime
```

Is there any way to turn this off?

sj26:
This seems to work:

```ruby
# Graphql 2.1.7 introduces `specifiedBy` directives on ISO8601 types:
#
# https://github.com/rmosolgo/graphql-ruby/pull/4014
#
# That's nice, but breaks some consumers that don't understand the directive.
#
GraphQL::Types::ISO8601Date.remove_directive(GraphQL::Schema::Directive::SpecifiedBy)
GraphQL::Types::ISO8601DateTime.remove_directive(GraphQL::Schema::Directive::SpecifiedBy)
GraphQL::Types::ISO8601Duration.remove_directive(GraphQL::Schema::Directive::SpecifiedBy)
```

rmosolgo:
Thanks for sharing what you found, @sj26 -- that looks like a good fix to me!

PR #4015, opened 2022-03-30
Fix `hash_key` attribute to work with keys commonly responded to by underlying GraphQL objects

joelzwarrington:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3944
Fixes https://github.com/rmosolgo/graphql-ruby/issues/3861

Closes #3947 as duplicate
- Removed some of the more complicated changes to focus on fixing the underlying issue first, and looking at refactoring further in later PRs to reduce the cognitive overload of having so many changes to a fundamental class

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -641,6 +641,8 @@ module GraphQL
 
               if @hash_key
                 inner_object[@hash_key]
+              elsif @dig_keys
+                inner_object.dig(*@dig_keys)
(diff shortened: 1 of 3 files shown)


joelzwarrington:
@rmosolgo I have brought out the smaller changes to fix the underlying issue with `hash_key` into this PR, and removed the more complex changes so that they could be done in isolation.

rmosolgo:
LGTM -- thanks for the fix!

PR #4016, opened 2022-04-01
Fix object id method definition

michaelvalette:
These need to be class methods.


rmosolgo (approved):
Thanks for correcting these!

PR #4025, opened 2022-04-12
Fix unexpected DateEncodingError when passing nil to optional ISO8601Date argument

### LINKED ISSUE #4019: Passing null to GraphQL::Types::ISO8601 type results in an error.
**Describe the bug**

Passing null to the argument of GraphQL::Types::ISO8601Date type of mutation occurs GraphQL::DateEncodingError.

**Versions**

`graphql` version: 1.13.6
`rails`: 6.1.5

**GraphQL schema**

```ruby
class Task < GraphQL::Schema::Object
  field :id, ID, null: false
  field :due_on, GraphQL::Types::ISO8601Date, null: true
  # …
end

class UpdateTask < Mutations::BaseMutation

  argument :due_on, GraphQL::Types::ISO8601Date, required: false
  # …

  field :result, Task, null: false
end

class MutationType < Types::BaseObject
  field :update_task, mutation: UpdateTask
end

  mutation MutationType

  def self.type_error(err, _query_ctx)
    p err
    super
  end
end
```

**GraphQL query**

```graphql
mutation {
  updateTask(input: { id: "xxx", dueOn: null }) {
    result {
      id
      dueOn
    }
  }
}
```

```json
{
  "data": {
    "updateTask": […]
  }
}
```

**Steps to reproduce**

Always occurs.

**Expected behavior**

GraphQL::DateEncodingError is not occurs.

**Actual behavior**

> #<GraphQL::DateEncodingError: Date cannot be parsed: . \nDate must be be able to be parsed as a Ruby Date object.>

**Additional context**

This change causes a GraphQL::DateEncodingError to be raised if nil is passed.
https://github.com/rmosolgo/graphql-ruby/commit/d7a723ed453a49cf1a500dec12718845342b0cb3

choznerol:
Close #4019

### DIFF
--- a/lib/graphql/types/iso_8601_date.rb
+++ b/lib/graphql/types/iso_8601_date.rb
@@ -23,7 +23,7 @@ module GraphQL
       end
 
       # @param str_value [String, Date, DateTime, Time]
-      # @return [Date]
+      # @return [Date, nil]
       def self.coerce_input(value, ctx)
         if value.is_a?(::Date)
           value
@@ -31,6 +31,8 @@ module GraphQL
           value.to_date
         elsif value.is_a?(::Time)
           value.to_date
+        elsif value.nil?
+          nil
         else
           Date.iso8601(value)
         end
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for the fix!

PR #4032, opened 2022-04-14
Updated implementation example for BaseConnection

### LINKED ISSUE #4031: Methods in example implementation for BaseConnection are in the wrong order
The implementation example in `BaseConnection` includes the methods `edge_type`, `x_nullable` but are in the wrong order.

`edge_type` should be the last method used in a connection class to ensure other class attributes have been set.

The example which should be updated:
```ruby
class Types::PostEdge < Types::BaseEdge
  node_type(Types::PostType)
end

class Types::PostConnection < Types::BaseConnection
  edge_type(Types::PostEdge)

  # The above `edge_type` method should be at the bottom of the class.
  # Alternatively, one could specify all of the options in the `edge_type` method
  # edge_type(Types::PostEdge, edges_nullable: true, edge_nullable: true, node_nullable: true, has_nodes_field: true)

  edges_nullable(true)
  edge_nullable(true)
  node_nullable(true)
  has_nodes_field(true)
end
```

In the example, the `x_nullable` options are set to true, but they don't have any effect. When making them not the default value, you'll see that the default value is used when creating the fields.

I'm planning on updating the documentation but wanted to create an issue to keep track of it!

joelzwarrington:
Resolves #4031

### DIFF
--- a/lib/graphql/types/relay/base_connection.rb
+++ b/lib/graphql/types/relay/base_connection.rb
@@ -30,9 +30,14 @@ module GraphQL
       #               edges_nullable: true,
       #               edge_nullable: true,
       #               node_nullable: true,
-      #               has_nodes_field: true
+      #               nodes_field: true
       #
-      #     # Alternatively, you can call class methods for the named arguments above - so long as `edge_type` is called afterwards. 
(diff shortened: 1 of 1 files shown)


rmosolgo (approved):
Thanks for this correction!

PR #4033, opened 2022-04-15
Simplify docs ActionCable config

cody-elhard:
Spent a bit of time figuring this out myself from the source code


rmosolgo:
Thanks for the improvement!

PR #4036, opened 2022-04-18
Fix lazy auth with dataloaded value

### LINKED ISSUE #4021: Wierd interaction between Lazy's and Dataloader.
Scenario: We have an app that currently makes extensive use of lazy loading with graphql-batch, RecordLoader, etc. I'm attempting to migrate to the Dataloader+Fibers model. That means both are currently enabled together in the schema, and I'm trying to move the first fields over to dataloader.

Symptom: In some cases, when Dataloader sources and RecordLoader are invoked in the same query, even in different objects and fields, a `SystemStackError` occurs when rendering the result to JSON. I am still working constructing a minimal repro case for this.

Investigation:
1. I added a byebug into the GraphQLController before `render json: result`
2. Inspecting the result object.
3. Seeing that a `GraphQL::Execution::Lazy` was being returned as the result for the field that was using Dataloader (teacherNote). The fields that were using RecordLoader and graphql-batch had no issues.
image
4. The lazy object is marked as `@resolved=false`, however, calling `.value` on that lazy object returns the unwrapped value immediately, without any db queries.

Theory:

The symptom is almost definitely caused by trying to `.to_json` a Lazy object.

But as to the root case, I'm at a bit of a loss on how to investigate beyond this point. My plan is to start reading the interpreter code (which I'm somewhat familiar with having read a lot of it over the past 2 years) and understanding how this could happen. Figured I would post here first in case there's something obvious I'm missing.

**Versions**

`graphql` version: 2.0.5. It was happening on 1.18.x as well, and I upgraded to see if it went away.
`rails` (or other framework): 6.1.4
`graphql-batch`: 0.5.1

**GraphQL schema**

```ruby
  query(Types::QueryType)
  use GraphQL::Batch
  use GraphQL::Dataloader

  ...
end
```

**GraphQL query**

Still working on a minimized / reliable repro query.

rmosolgo:
Fixes #4021

### DIFF
--- a/spec/graphql/dataloader_spec.rb
+++ b/spec/graphql/dataloader_spec.rb
@@ -1122,4 +1122,64 @@ describe GraphQL::Dataloader do
     assert_equal({ "data" => { "nested2" => "nested2" } }, NestedDataloaderCallsSchema.execute("{ nested2 }"))
     assert_equal({ "data" => { "nested" => "nested", "nested2" => "nested2" } }, NestedDataloaderCallsSchema.execute("{ nested nested2 }"))
   end
+
+  describe "with lazy authorization hooks" do
+    class LazyAuthHookSchema < GraphQL::Schema
+      class Source < ::GraphQL::Dataloader::Source
+        def fetch(ids)
+          return ids.map {|i| i * 2}
+        end
+      end
+
+      class BarType < GraphQL::Schema::Object
+        field :id, Integer
+
+        def id
+          object
+        end
+
+        def self.authorized?(object, context)
+          -> { true }
+        end
+      end
+
+      class FooType < GraphQL::Schema::Object
+        field :dataloader_value, BarType
+
+        def self.authorized?(object, context)
+          -> { true }
+        end
+
+        def dataloader_value
+          dataloader.with(Source).load(1)
+        end
+      end
+
+      class QueryType < GraphQL::Schema::Object
+        field :foo, FooType
+
+        def foo
+          {}
+        end
+      end
+
+      use GraphQL::Dataloader
+      query QueryType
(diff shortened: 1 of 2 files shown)


spec/graphql/dataloader_spec.rb
```diff
@@ -1122,4 +1122,64 @@ def nested2
     assert_equal({ "data" => { "nested2" => "nested2" } }, NestedDataloaderCallsSchema.execute("{ nested2 }"))
     assert_equal({ "data" => { "nested" => "nested", "nested2" => "nested2" } }, NestedDataloaderCallsSchema.execute("{ nested nested2 }"))
   end
+
+  describe "with lazy authorization hooks" do
```
palasha, line 1126: Caveat: I have fairly low context for this question so please feel free to ignore it.

Was the bug just with authorization methods returning lazy values? Or could other parts of the resolver chain like `scope_items` or field extensions (that use dataloader) also create similarly shaped issues.

Would it make sense to test various lazy/dataloader nestings of those features as well?

rmosolgo: IMO, this test is sufficient. The bug was at a lower level than any particular GraphQL feature -- instead, it was a bug about how dataloader and lazy values hand off control to one another. 

Previously, the lazy resolution code was enqueuing itself as a single job in the dataloader. This didn't work if an already-enqueued job would require multiple passes to populate some key/value pairs in the response object. (The lazy resolution code uses response objects directly to find `Lazy`s to resolve.) 

So instead of queuing up a single job and assuming that any preceding jobs will populate the response objects, now, it calls `dataloader.run` to make sure that _all_ preceding jobs have totally finished before using response objects.

PR #4038, opened 2022-04-20
Improve Datadog tracing support for ddtrace 1.0

marcotc:
This PR improves the support for GraphQL tracing with the Datadog tracing exporter.

The two changes are:
* Tag spans generated by GraphQL operations with an appropriate 'component' and 'operation' tags. This information was previously captured in the span name, as a concatenated pair. Having this information into two separate fields allows for richer grouping in the future.
* Span resource is not set to `nil` when `GraphQL::Tracing::DataDogTracing` cannot derive an appropriate rich value for it.  This allows for `ddtrace` to back-fill the resource value later in `ddtrace`s processing logic. Setting it explicitly to `nil`, as it was done before, sends a message to `ddtrace` that we explicitly don't want a resource, which was not the case here.

### DIFF
--- a/lib/graphql/tracing/data_dog_tracing.rb
+++ b/lib/graphql/tracing/data_dog_tracing.rb
@@ -17,15 +17,21 @@ module GraphQL
       def platform_trace(platform_key, key, data)
         tracer.trace(platform_key, service: service_name) do |span|
           span.span_type = 'custom'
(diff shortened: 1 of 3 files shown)


rmosolgo (approved):

marcotc:
Thank you for the quick review and release, @rmosolgo! 🙇

PR #3627, opened 2021-09-29
Suggest using rails/globalid for object identification

zavan:
I'm doing this in one application of mine and it's been working great. Comes out-of-the-box with Rails >= 4.2.
It's not encrypted, but one could use the signed version if that's desirable (much longer string though):

```
> signed = object.to_sgid.to_param
"BAh7CEkiCGdpZAY6BkVUSSIWR2lkOi8vYXBpL0FkbWluLzQGOwBUSSIMcHVycG9zZQY7AFRJIgxkZWZhdWx0BjsAVEkiD2V4cGlyZXNfYXQGOwBUSSIdMjAyMS0xMC0yOVQxNjo0OTozNy45ODdaBjsAVA==--e37d30ef25344e027994dbc198598e159b9b9646"

> GlobalID::Locator.locate_signed(signed)
```

PR #4040, opened 2022-04-23
Use GlobalID methods instead of manually handling the encoding/decoding

zavan:
Just a simple improvement.

Related: I'm curious about how this method of encoding the id is better than [what we suggested before](https://github.com/rmosolgo/graphql-ruby/pull/3627/files) (`object.to_gid.to_param`)? Why the extra work? That method produces a good looking string (like `Z2lkOi8vcmV2aWV3cy9SZXZpZXcvNDc`) too and is easier to locate.

Edit: Looks like it's even simpler: `object.to_gid_param`.

### DIFF
--- a/guides/schema/object_identification.md
+++ b/guides/schema/object_identification.md
@@ -20,24 +20,12 @@ class MySchema < GraphQL::Schema
   def self.id_from_object(object, type_definition, query_ctx)
     # Generate a unique string ID for `object` here
     # For example, use Rails' GlobalID library (https://github.com/rails/globalid):
-    object_id = object.to_global_id.to_s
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for the improvement here! I didn't know about `to_gid_param` -- I agree that'd be even better. Do you mind updating this PR? 

For reference, `to_gid_param` does the exact same thing: makes a global ID and base64-encodes it, then removes the padding: https://github.com/rails/globalid/blob/9c11e481ff11ca4f1efbc7abd99441d23cce9593/lib/global_id/identification.rb#L8

rmosolgo (approved):
Thanks for the improvement!

PR #4043, opened 2022-04-25
Fix hash_key:, tries a string as a fallback, fixes #4037

### LINKED ISSUE #4037: `hash_key` fields do no longer attempt to resolve the stringified key
**Describe the bug**

Since v2.0.6 (#4015 to be more precise) fields that use `hash_key` will no longer try to resolve the stringified key.

In previous versions of the gem, the resolver did look up both the symbolized and stringified version of the hash key.

**Versions**

`graphql` version: 2.0.6

**GraphQL schema**

```ruby
  field :name, String, hash_key: :name
end
```

**GraphQL query**

Given the hash `{ 'name' => 'foo' }` as the underlying "product" object, the following query will fail:

```graphql
query {
  products { name }
}
```

```json
{
  "data": {
    "products": null
  },
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Product.name"
    }
  ]
}
```

**Steps to reproduce**

Run the above example.

**Expected behavior**

The product's `name` field resolves to `"foo"`.

**Actual behavior**

The product's `name` field resolves to `nil` and the query fails.

rmosolgo:


rmosolgo:
cc @joelzwarrington  as an example of the _crazy_ compatibility issues I encounter with field resolution 😩 😅

PR #4050, opened 2022-04-27
Minor table formating fix

uzbekjon:
Table in "Lists, Nullable Lists, and Lists of Nulls" section was not properly formatted and GitHub did not render it correctly.

**Before:**

Screen Shot 2022-04-27 at 17 48 56

**After:**

Screen Shot 2022-04-27 at 17 49 06


rmosolgo:
Thanks for the fix!

Issue #4056, opened 2022-04-30
Getting undefined method `arguments' in extract_deprecated_arguments


digitalmoksha:
**Describe the bug**

A subclass of `GraphQL::Schema::InputObject` is used as an argument in one of our resolvers, like this:

```ruby
argument :timeframe, Types::TimeframeInputType,
         required: false,
         description: 'List items overlapping the given timeframe.'


module Types
  class TimeframeInputType < GraphQL::Schema::InputObject
    graphql_name 'Timeframe'
    description 'A time-frame defined as a closed inclusive range of two dates'
    
    argument :start, ::Types::DateType, required: true, description: 'Start of the range.'
    argument :end, ::Types::DateType, required: true, description: 'End of the range.'

    def prepare
      if self[:end] && self[:start] && self[:end] < self[:start]
        raise ::Gitlab::Graphql::Errors::ArgumentError, 'start must be before end'
      end

      to_h
    end
  end
end
```

This worked fine on `1.11.x`, but now on `1.13.12`, I'm getting

```
undefined method `arguments' for {:start=\u003eSat, 30 Apr 2022, :end=\u003eMon, 02 May 2022}:Hash
```

I've traced it down to

```ruby
            if argument.definition.type.kind.input_object?
              extract_deprecated_arguments(argument.value.arguments.argument_values) # rubocop:disable Development/ContextIsPassedCop -- runtime args instance
```

in https://github.com/rmosolgo/graphql-ruby/blob/1f8b198e00f69b08b9cd7eee00a799c4a304033b/lib/graphql/analysis/ast/field_usage.rb#L42

I think if we checked for `respond_to?(:arguments)` like is done in the `list?` branch, it would fix this. It does when I make the change locally.

```ruby
            if argument.definition.type.kind.input_object? && argument.value.respond_to?(:arguments)
              extract_deprecated_arguments(argument.value.arguments.argument_values) # rubocop:disable Development/ContextIsPassedCop -- runtime args instance
```

**Versions**

`graphql` version: 1.13.12
`rails` (or other framework): Rails 6.1.4.7

**Query**

```
{
  group(fullPath: "group1/group2"){
    milestones(timeframe: {start: "2022-05-04", end: "2022-05-06"}){
      edges {
        node {
          id
          title
          state
        }
      }
    }
  }
}
```

rmosolgo:
I'm curious how the input object was transformed into 

```ruby
[...]

PR #4057, opened 2022-05-03
Ensure argument.value responds to arguments method before calling

digitalmoksha:
This was happening in `extract_deprecated_arguments` in `lib/graphql/analysis/ast/field_usage.rb` when the `argument.value` was a Hash.

Related issue: https://github.com/rmosolgo/graphql-ruby/issues/4056

### DIFF
--- a/lib/graphql/analysis/ast/field_usage.rb
+++ b/lib/graphql/analysis/ast/field_usage.rb
@@ -39,7 +39,7 @@ module GraphQL
               @used_deprecated_arguments << argument.definition.path
             end
 
-            if argument.definition.type.kind.input_object?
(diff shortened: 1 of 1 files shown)


digitalmoksha:
This if affecting me particularly in `1.13.12`, so I hope it would be possible to to cut a new version such as `1.13.13` 🤞

rmosolgo:
Sure, I can definitely backport this to 1.13. But could you help me understand, how did you have a `Hash` here? Was it from `prepare`?

rmosolgo:
:+1: I can definitely backport it to 1.13. I'd like to get a test in place, so I'm asking for some details in the original issue.

digitalmoksha:
I guess this PR isn't actually needed, as we were using the API slightly incorrectly.  So closing it out.

Issue #4059, opened 2022-05-04
Using `hash_key` doesn't work for non-hash instances that don't respond to `fetch`


danielvdao:
**Describe the bug**

My coworker @georgexu22 and I discovered this bug..

Using `hash_key` doesn't work for non-hash instances that don't respond to `fetch`. An example would probably be an `ActiveRecord` or `OpenStruct` instance that doesn't respond to `fetch`. Ideally we should validate that the inner object is a `Hash` if we want to use `fetch` - but that could have probably have breaking changes

We saw this bug here:

```ruby
NoMethodError: undefined method `fetch'
```
**Versions**

`graphql` version: 2.0.6
`rails` (or other framework):
other applicable versions (`graphql-batch`, etc)

**GraphQL schema**

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok). Any custom extensions, etc?

```ruby
class TestType < GraphQL::Schema::Object
  field :zed, ID, hash_key: :bar
end

class Product < GraphQL::Schema::Object
  field :example, TestType
  

  def example
    OpenStruct.new(zed: 1, bar: 1)
  end
end

class ApplicationSchema < GraphQL::Schema
  query QueryType
end
```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
query {
  products { example { zed } }
}
```

**Steps to reproduce**

I think it's simple to replicate with something like this in a console actually:

```ruby
irb(main):002:0> test = OpenStruct.new(bar: 1, zed: 1)
=> #<OpenStruct bar=1, zed=1>
irb(main):003:0> test[:zed]
=> 1
irb(main):004:0> test.fetch(:zed)
Traceback (most recent call last):
        4: from /Users/daniel.dao/.rbenv/versions/2.7.4/bin/irb:23:in `<main>'
        3: from /Users/daniel.dao/.rbenv/versions/2.7.4/bin/irb:23:in `load'
        2: from /Users/daniel.dao/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        1: from (irb):4
NoMethodError (undefined method `fetch' for #<OpenStruct bar=1, zed=1>)
```
**Expected behavior**

[...]

PR #4060, opened 2022-05-04
Don't use `fetch` for fields that use `hash_key`

danielvdao:
This is a proposal to fix https://github.com/rmosolgo/graphql-ruby/issues/4059.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -643,7 +643,13 @@ module GraphQL
               inner_object = obj.object
 
               if defined?(@hash_key)
+<<<<<<< HEAD
                 inner_object[@hash_key].nil? ? inner_object[@hash_key_str] : inner_object[@hash_key]
+=======
+                inner_object.fetch(@hash_key) {
+                  inner_object[@hash_key_str]
+                }
+>>>>>>> parent of 059a5b7ea (🐛  Move the `hash_keys` call into `Hash` check)
               elsif @dig_keys
                 inner_object.dig(*@dig_keys)
               elsif obj.respond_to?(resolver_method)
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -642,6 +642,8 @@ module GraphQL
 
               inner_object = obj.object
 
+              if defined?(@hash_key)
+                inner_object[@hash_key].nil? ? inner_object[@hash_key_str] : inner_object[@hash_key]
               elsif @dig_keys
                 inner_object.dig(*@dig_keys)
               elsif obj.respond_to?(resolver_method)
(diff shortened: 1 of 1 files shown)


lib/graphql/schema/field.rb
```diff
@@ -643,9 +643,7 @@ def resolve(object, args, query_ctx)
               inner_object = obj.object
 
               if defined?(@hash_key)
-                inner_object.fetch(@hash_key) {
-                  inner_object[@hash_key_str]
-                }
+                inner_object[@hash_key].nil? ? inner_object[@hash_key_str] : inner_object[@hash_key]
```
danielvdao, line 646: Actually I am worried this might not work; if the value is actually `nil` this will revert to `@hash_key_str`

danielvdao: Curious what're your thoughts @rmosolgo? Personally it's a bit weird to see `hash_key` not be guarded by `is_a?(Hash)` -- but maybe it's meant to be used this way as noticed in previous PRs.

danielvdao:
I tried a bunch of different options in the commit tree. Maybe it's best to defer to you 😂

lib/graphql/schema/field.rb
georgexu22: We're also unsure how `hash_key` is supposed to be used. Reference [in the docs](https://graphql-ruby.org/fields/introduction.html#field-resolution): `If the underlying object is a Hash, lookup a key in that hash.`. Is this not the case anymore?

danielvdao:
Thinking out loud when visiting this method, it's checking class type and also method resolution. Ideally we should split it into:

```ruby
case inner_object
  when Hash
  ....
  else
  ....
end
```

And then do the behavior -- but.... this is risky 😅, as I imagine clients may have field names that are `Hash` / `Object` instance methods... Oh the joys of ruby 🤣

lib/graphql/schema/field.rb
rmosolgo: I chose `fetch` because it _would_ return `nil` if the key is present in `inner_object` and the value is `nil` -- I think we should retain that behavior.

How about:  

```suggestion
                inner_object.key?(@hash_key) ? inner_object[@hash_key] : inner_object[@hash_key_str]
```

(it looks like `.key?` is what the code _used_ to use: https://github.com/rmosolgo/graphql-ruby/pull/4015/files#diff-53ed7db7a95b84a82815af3a920ccd962fce9469ef21ed25111fb149bc1406eaL658-L661)

danielvdao: Sorry if I'm unclear @rmosolgo or maybe we're mis-understanding, but actually I think the problem is that we have this:

```ruby
if defined?(@hash_key)
```

check outside of checking whether or not the underlying object is a `Hash`, example code from today: [link](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/field.rb#L645). So even if we use `.key?` I think it would break... maybe the proper fix is to actually move this code block back into [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/field.rb#L660)?

rmosolgo:
Thanks for all your work on this! I've rolled your commits into #4072 and tried to address the original concern from #4059. Please let me know on that new PR (or in a new issue) if the new version doesn't properly fix your case!

danielvdao:
Thanks Robert! Funnily enough our code was broken as we used `hash_key` for something that responded to the hash accessor method but wasn't actually a hash 😂 

We used `method` to fix it -- but thought to try and fix it upstream as well. Appreciate the follow-through 💚

PR #4061, opened 2022-05-04
accept dates missing hyphen

dgodd:
https://github.com/rmosolgo/graphql-ruby/commit/d045a207e0cf8bb53e31549f82b78c72a0e661c0 changed the behaviour for parsing dates (as date times) to make the parsing much safer. However, I have several consumers of my graphql apis who currently send datetime values as a date without hyphens (eg. "20220505"). I would like to make that use case continue to work in a safe way.


rmosolgo:
Hi! Sorry for the compatibility trouble and thanks for the fix.

PR #4069, opened 2022-05-18
Add fallback_value option to Field

### LINKED ISSUE #4020: Default/Fallback Implementation of Interface Fields

Currently if you define a field implementation on an interface, it will take precedence over `public_send` and hash lookup for that field on objects that implement the interface. This is probably for the best, but on interfaces it's perhaps more useful to specify a _fallback_ implementation.

**Scenario**

I have a data model that has lots of entity types. All of them have an `id` property. Most of them have a `name` property, but some of them don't. I want all of them to implement an interface (`Node`) that has a required `id` property and an optional `name` property.

```graphql
interface Node { id: ID! name: String }
```

If I add an implementation like `def name= nil` to my interface, it'll make _all_ my nodes come back with `nil` for the name. If I don't, I'll get an error b/c some of my data objects don't have a "name" property.

This may already be achievable in some way I haven't discovered. If not, there are a bunch of possible solutions:

1. Add something to the interface DSL allowing you to define `fallback_methods` for fields. These would called instead of erroring  if `key?`, `respond_to?`, etc. fail. Seems annoying to implement/maintain.
2. ⭐  Add a new property to Field, `fallback_value` or `default` or something, that behaves similarly to option 1. Seems easier to implement and potentially useful in other situations where you might want to return nil instead of erroring.
3. Make it so `public_send` etc., _always_ take precedence over methods defined on interfaces. Backwards-incompatible, potentially surprising.
5. ⭐ Do something with `Field#resolve` so it's easier for a field implementation to provide a fallback on top of the default behavior. E.g. I want to do:
```ruby
field :name, String
def name
  context[:current_field].try_resolve_from_object(self) || "default_value"
  # or ideally
  yield || "default_value"
end
```

Happy to start a PR if any of these sound good to you.

It's not a big deal to work around the absence of this feature. If the fallback behavior you want is trivial, It's easy to add an empty implementation to each of your GQL objects or data objects. If the fallback behavior is fancier, it can be put in a mixin or something.

**Repro-ish**

Let's say my schema looks like this:

```graphql
interface Node { id: ID! name: String }
type Query { nodes: [Node!] }

type NamedObj implements Node { id: ID! name: String! }
type Obj      implements Node { id: ID! name: String }
```

here's `Node` in Ruby:

```ruby
module NodeType

  field :id, ID, null: false
  field :name, String
end
```

And here's the underlying data:
```ruby
DATA = [
  # NamedObj
  OpenStruct.new(id: "1", name: "Name"),
  # Obj
  OpenStruct.new(id: "2"),
]
```

When I run this query against the schema:
```graphql
{ nodes { id, name, __typename } }
```

I'll get "Failed to implement Node.name..." b/c my second object doesn't respond to `name`.

So I'll add an implementation of `name` to `NoteType`:

```ruby
def name
  "fallback"
end
```

No more errors, but now `name` is "fallback" for _all_ my nodes, even the ones with names.

danajackson2:
Fixes #4020. This adds a parameter called `:fallback_value` to Field, which allows you to specify a value to be returned if the usual ways (calling the method on an object, or checking for the key in a hash) fail. It's useful if there is a query that returns objects that might or might not have a certain field.

The motivation for this is a scenario where we have a data model that has lots of entity types. All of them have an `id` property. Most of them have a `name` property, but some of them don't. We want all of them to implement an interface (Node) that has a required `id` property and an optional `name` property.
``` ruby
interface Node { id: ID! name: String }
```

The default for the `:fallback_value` param in the initializer is `:not_given`, which allows us to set `nil` as the fallback if we want. The most common use case is probably with interfaces, so that's where I wrote tests. Are there any other places this should be tested? I added documentation to guides/fields/introduction.md under `Field Resolution`. Does that seem sufficient? I added a blurb to the error message about supplying a `fallback_value` to implement the field.

### DIFF
--- a/guides/fields/introduction.md
+++ b/guides/fields/introduction.md
@@ -87,6 +87,7 @@ By default, fields return values by:
 
 - Trying to call a method on the underlying object; _OR_
 - If the underlying object is a `Hash`, lookup a key in that hash.
+- An optional `:fallback_value` can be supplied that will be used if the above fail.
 
 The method name or hash key corresponds to the field name, so in this example:
 
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -645,7 +645,7 @@ module GraphQL
               inner_object = obj.object
 
               if defined?(@hash_key)
-                inner_object[@hash_key] || inner_object[@hash_key_str]
+                inner_object[@hash_key] || inner_object[@hash_key_str] || (@fallback_value != :not_given ? @fallback_value : nil)
               elsif obj.respond_to?(resolver_method)
                 method_to_call = resolver_method
                 method_receiver = obj
@@ -661,8 +661,12 @@ module GraphQL
                 elsif defined?(@hash_key)
                   if inner_object.key?(@hash_key)
                     inner_object[@hash_key]
-                  else
+                  elsif inner_object.key?(@hash_key_str)
                     inner_object[@hash_key_str]
+                  elsif @fallback_value != :not_given
+                    @fallback_value
+                  else
+                    nil
                   end
                 elsif inner_object.key?(@method_sym)
                   inner_object[@method_sym]
@@ -670,6 +674,8 @@ module GraphQL
                   inner_object[@method_str]
                 elsif @fallback_value != :not_given
                   @fallback_value
+                else
+                  nil
                 end
               elsif inner_object.respond_to?(@method_sym)
                 method_to_call = @method_sym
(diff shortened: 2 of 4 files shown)


rmosolgo:
This looks great, thanks for making this improvement! I pushed a commit to test the `hash_key: ...` behavior, although changes upstream mean there are actually _several_ hash-lookup-related branches in field resolution. I wouldn't be surprised if there are few resolution paths (`@dig_keys` comes to mind) where `fallback_value` isn't properly applied yet. But we can cross those bridges when we get to them. Thanks again!

PR #4072, opened 2022-05-24
Fix some hash lookup behaviors

### LINKED ISSUE #4059: Using `hash_key` doesn't work for non-hash instances that don't respond to `fetch`

### LINKED ISSUE #4068: Field method_str semantics is broken and ignores the hash_key
**Describe the bug**

The semantics of `field.method_str` is broken:

```Ruby
# @return [String] Method or hash key on the underlying object to look up
attr_reader :method_str
```
[see](https://github.com/rmosolgo/graphql-ruby/blob/590929323b32ca016876c7392e9ef46d19252567/lib/graphql/schema/field.rb#L29)

```Ruby
method_name = method || name_s
```
[see](https://github.com/rmosolgo/graphql-ruby/blob/590929323b32ca016876c7392e9ef46d19252567/lib/graphql/schema/field.rb#L249)

It used to be the correct order as described in the docs and the attribute comment.

```Ruby
```

This breaks the look ahead of the field names. My use case that I have an elasticsearch instance where I map grapQL queries to elasticsearch queries. This change broke the look ahead.

**Versions**

`graphql` version: 2.0.6 + 2.0.7
`rails` (or other framework): 7.0.x

[See also](https://github.com/rmosolgo/graphql-ruby/compare/v2.0.5...v2.0.6#diff-53ed7db7a95b84a82815af3a920ccd962fce9469ef21ed25111fb149bc1406eaR249)
image

rmosolgo:
Fixes #4059 
Fixes #4068

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -643,7 +643,13 @@ module GraphQL
               inner_object = obj.object
 
               if defined?(@hash_key)
+<<<<<<< HEAD
                 inner_object[@hash_key].nil? ? inner_object[@hash_key_str] : inner_object[@hash_key]
+=======
+                inner_object.fetch(@hash_key) {
+                  inner_object[@hash_key_str]
+                }
+>>>>>>> parent of 059a5b7ea (🐛  Move the `hash_keys` call into `Hash` check)
               elsif @dig_keys
                 inner_object.dig(*@dig_keys)
               elsif obj.respond_to?(resolver_method)
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -642,6 +642,8 @@ module GraphQL
 
               inner_object = obj.object
 
+              if defined?(@hash_key)
+                inner_object[@hash_key].nil? ? inner_object[@hash_key_str] : inner_object[@hash_key]
               elsif @dig_keys
                 inner_object.dig(*@dig_keys)
               elsif obj.respond_to?(resolver_method)
@@ -654,16 +656,10 @@ module GraphQL
                   obj.public_send(resolver_method)
                 end
               elsif inner_object.is_a?(Hash)
-                if defined?(@hash_key)
-                  inner_object.fetch(@hash_key) {
-                    inner_object[@hash_key_str]
-                  }
+                if inner_object.key?(@method_sym)
+                  inner_object[@method_sym]
                 else
-                  if inner_object.key?(@method_sym)
-                    inner_object[@method_sym]
-                  else
-                    inner_object[@method_str]
-                  end
+                  inner_object[@method_str]
                 end
               elsif inner_object.respond_to?(@method_sym)
                 method_to_call = @method_sym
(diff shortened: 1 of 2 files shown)


rmosolgo:
👋 Hey @alexcwatt, I'd be happy to try to backport the behaviors you're looking for. I got started over in https://github.com/rmosolgo/graphql-ruby/pull/4286, will that cover it for you? Please let me know more details on the PR if there's something else you have in mind!

PR #4078, opened 2022-05-27
Allow null input objects in arguments during field usage analysis

lyricsboy:
Problem
---

When passing in an explicit `null` value for a nullable input object-type argument on a field, field usage analysis will raise an exception like this:

```
NoMethodError: undefined method `arguments' for nil:NilClass
    graphql-ruby/lib/graphql/analysis/ast/field_usage.rb:43:in `block in extract_deprecated_arguments'
    graphql-ruby/lib/graphql/analysis/ast/field_usage.rb:37:in `each_pair'
    graphql-ruby/lib/graphql/analysis/ast/field_usage.rb:37:in `extract_deprecated_arguments'
    graphql-ruby/lib/graphql/analysis/ast/field_usage.rb:22:in `on_leave_field'
```

Solution
---

Skip arguments with null values during field usage analysis.

### DIFF
--- a/lib/graphql/analysis/ast/field_usage.rb
+++ b/lib/graphql/analysis/ast/field_usage.rb
@@ -39,9 +39,11 @@ module GraphQL
               @used_deprecated_arguments << argument.definition.path
             end
 
+            next if argument.value.nil?
+
             if argument.definition.type.kind.input_object?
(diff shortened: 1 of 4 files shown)


rmosolgo:
Thanks for this fix! It looks good to me. 

It looks like adding a new field to the dummy schema broke some assertions about the introspection result on that schema, do you mind updating those tests to include the newly-added field?

Issue #2999, opened 2020-06-17
Add `default_page_size` for relay connections


rstawarz:
**Is your feature request related to a problem? Please describe.**

In 1.9 the library would default `first` to `nil` when no first parameter was given, however 1.10 [modified the behavior](https://github.com/rmosolgo/graphql-ruby/pull/2608) to assume that the default value of `first` should be the `max_page_size` when no value is given.   Essentially there is no way to differentiate between when the `max_page_size` is being used as the `max` and when it's being used as a `default` value.

I came across this issue in upgrading to 1.10.  We have some custom connection classes that are backed by REST APIs.  With the change to the implementation of `first` it's causing us to fetch the maximum number of results; whereas when `first` was nil, the REST API implementation returned a default number of results which are totally different values in our use case (default: 3 vs. max: 100).  

**Describe the solution you'd like**

Allow for specifying a `default_page_size` on connections to allow for specifying a default value instead of assuming `max_page_size` should be the default.

**Describe alternatives you've considered**

We could lower the `max_page_size`, but that limits some usages of the fields for those who need larger result sets.

I have no other alternative implementation - using the max as the default causes performance issues for our use case, as I imagine it might for others.  For the time being I've implemented the `default_page_size` like above for our custom connection, but in doing so I had to overwrite the [implementation of first](https://github.com/rmosolgo/graphql-ruby/pull/2608/files#diff-dd6f9eae72d868abc560dbf82ebc7ac4R71) to replace the `max_page_size` with `default_page_size` which may break with future upgrades to the gem.  

I'd like to open a PR against the gem to introduce `default_page_size`, but before I go through the effort I was looking to verify it would be accepted.

rmosolgo:
Hi! Thanks for the great suggestion. I'd definitely accept a PR for this feature: 

- Add a configuration option, `default_page_size:`
- When neither `first` nor `last` is given, apply `default_page_size`
- (As before, if `first` or `last` exceed `max_page_size`, then apply `max_page_size` instead)
- For compatibility, if `max_page_size:` is given but `default_page_size:` is not, then use `max_page_size:` as `default_page_size:` 
- (If `default_page_size: nil` is explicitly configured, then return an unbounded list if `first:` and `last:` are omitted?)

If you're still interested in giving it a shot, ping me on the PR and I'll take a look!

PR #4081, opened 2022-05-30
Add default_page_size to schema, field, and resolver

connorshea:
This implements #2999.

Essentially, this allows you to set `default_page_size` on the schema, fields, and resolvers.

```ruby
class MyAppSchema < GraphQL::Schema
  default_page_size 50
end

field :items, Item.connection_type, null: false, default_page_size: 25

class ItemResolver < GraphQL::Schema::Resolver
  default_page_size 10
end
```

If no `first:` or `last:` values are specified, the default_page_size will be used to determine the size of a given page of results, and if `default_page_size` isn't set it'll fall back to the existing behavior with `max_page_size`. This should be backwards-compatible since the behavior is unchanged if `default_page_size` isn't set anywhere.

I added basic documentation and specs, and I've tested this manually with my Rails app to make sure the main use-cases all work as-expected there. I'm not sure what other specs should be written for this, if any?

cc: @rmosolgo since you asked for a ping in that issue.

### DIFF
--- a/guides/pagination/using_connections.md
+++ b/guides/pagination/using_connections.md
@@ -10,7 +10,7 @@ index: 2
 
 GraphQL-Ruby ships with a few implementations of the {% internal_link "connection pattern", "pagination/connection_concepts" %} that you can use out of the box. They support Ruby Arrays, Mongoid, Sequel, and ActiveRecord.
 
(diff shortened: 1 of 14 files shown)


rmosolgo:
This all looks great -- thanks for the contribution!

PR #4077, opened 2022-05-26
Add query execution context based tracing for API requests

toneymathews:
Closes https://github.com/rmosolgo/graphql-ruby/issues/4001

An attempt at allowing tracing per request as per suggestions in https://github.com/rmosolgo/graphql-ruby/issues/4001

- `cached_platform_key` now accepts the trace phase as a third parameter.
- `OpenTelemetryTracing` overrides `cached_platform_key` to access execution context not available in `platform_field_key`, `platform_authorized_key`, `platform_resolve_type_key`

This PR includes
- modified copy of the [OpenTelemetry implementation](https://github.com/open-telemetry/opentelemetry-ruby/blob/99bfe640cef822adb5f173937b7a98c7d5c48758/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb) to demonstrate how we would make use of `PlatformTracing`
- tests to ensure the above works as expected
- a stub for the OpenTelemetry agent

~~CI has failed due to rubocop failures in files not associated with this PR.~~

If this looks good, we are hoping to add this to 1.13.x (< 2) since we haven't yet migrated.

Issue #4001, opened 2022-03-22
`PlatformTracing` per request trace configuration


ravangen:
**Is your feature request related to a problem? Please describe.**

With `opentelemetry-instrumentation-graphql`, a `GraphQLTracer` is provided based on `GraphQL::Tracing::PlatformTracing`.

It provides [several schema level configuration options](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-instrumentation-graphql/v0.19.2/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb#L38-L40) that can enable/disable specific parts of tracing.

I am looking for a way to manage these options per execution instead of being constant. For example, a HTTP header may request verbose traces be recorded, ideally enabling [`enable_platform_field`](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-instrumentation-graphql/v0.19.2/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb#L46
) for only this request's GraphQL execution.

**Describe the solution you'd like**

When building the `platform_key` ([here](https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/lib/graphql/tracing/platform_tracing.rb#L43-L44), [here](https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/lib/graphql/tracing/platform_tracing.rb#L59-L60), [here](https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/lib/graphql/tracing/platform_tracing.rb#L66-L67)), I wonder if it makes sense to optionally provide `context`. This would enable `platform_field_key`/`platform_resolve_type_key`/`platform_authorized_key` to make request specific determinations based on the GraphQL execution context.

For example, something like this would now be possible:
```ruby
def platform_field_key(type, field, context)
  return unless config[:enable_platform_field] || context.namespace(:opentelemetry)[:enable_platform_field]

  "#{type.graphql_name}.#{field.graphql_name}"
end
```

The [platform key cache](https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/lib/graphql/tracing/platform_tracing.rb#L127-L141) should not be impacted as it keeps it data under `context`, regenerated for each request.

**Additional context**

I imagine this would be a nice to have for anything using `PlatformTracing`, not only Open Telemetry.

rmosolgo:
[...]

PR #4086, opened 2022-06-02
Remove partial `opentelemetry` implementation

ravangen:
Optional followup to https://github.com/rmosolgo/graphql-ruby/pull/4077 and a [comment](https://github.com/rmosolgo/graphql-ruby/pull/4077#issuecomment-1142365741) impacting `2.0.9` and `1.13.13`

This PR seeks to return this gem to a "stable state" and not have any misleading or inaccessible code. It is a revert of all the tracing changes with the exception of `lib/graphql/tracing/platform_tracing.rb`. I still believe there is some value in having a way to access the execution context when tracing, and this was the recommend way to achieve this per https://github.com/rmosolgo/graphql-ruby/issues/4001.

First-party OpenTelemetry instrumentation would be wonderful, but it was not my intention to force this decision either way as part of the original suggested changes.

Apologies again for the confusion and happy to try and help sort this out based on your recommendations.


rmosolgo:
Likewise, sorry I misunderstood the nature of these files in the original PR 🤦  I have no preference whether they're here or in opentelemetry-ruby. 

My only hangup, IIRC, the only test for passing `trace_phase` was the opentelemetry tracer itself. If we remove it here, then an over-enthusiastic refactor might remove that option, since it's not used anywhere in this repo.

Would you mind adding a note to the source (or a test, if you can think of a handy way to add one) to help me maintain compatibility in the future?

rmosolgo:
Added a comment in f4d99f692

ravangen:
I still will see about adding a test case soon, just to have that peace of mind. Been a bit distracted recently.

toneymathews:
Can we pls add these changes to `1.13.x` as well? I see `1.13.14` still has the opentelemetry implementation.

rmosolgo:
Oops -- just released this in 1.13.15, too!

PR #4088, opened 2022-06-02
Require Rake before referencing `Rake::DSL`

sambostock:
This file references `Rake::DSL`, and may be required in isolation, and so it should `require "rake"` before using it.

Without this, requiring this file in isolation blows up when encountering this constant.


rmosolgo:
Thanks for the fix!

Issue #4093, opened 2022-06-08
stack level too deep after 1.13.0 (not present on 1.12.24)


TSMMark:
The error exists on latest as well. v2.0.9

**Describe the bug**

It seems like when fields recursively reference each other, it has a problem.

```
Exception (SystemStackError): stack level too deep
stack level too deep
/root/.asdf/installs/ruby/3.0.0/lib/ruby/3.0.0/set.rb:119:in `each_entry'
/root/.asdf/installs/ruby/3.0.0/lib/ruby/3.0.0/set.rb:119:in `do_with_enum'
/root/.asdf/installs/ruby/3.0.0/lib/ruby/3.0.0/set.rb:442:in `merge'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:44:in `add_directives_from'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:186:in `block in add_type'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `each'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `add_type'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:185:in `block in add_type'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `each'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `add_type'
# ... ETC ...
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:185:in `block in add_type'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `each'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `add_type'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:185:in `block in add_type'
/vydia/bundle/ruby/3.0.0/gems/graphql-1.13.0/lib/graphql/schema/addition.rb:180:in `each'
[...]

PR #4097, opened 2022-06-13
Addition#add_type check if prev_type Array contains type - Solves #4093

TSMMark:
Solves #4093


rmosolgo:
Thanks for this fix! I grabbed your commit and added a test in #4107

PR #4112, opened 2022-06-22
HasFields#fields: skip ancestor's field when it's not visible locally

FunkyloverOne:
Hey @rmosolgo, I believe this should be the correct behavior, because when I want my field excluded locally - I don't want the ancestor's version to be included instead. 
Makes sense? WDYT?

That's the monkey patch I must use in my project locally right now, that's more or less how it worked in v1.12.x.
If you agree with it, I would also cover this behavior with specs.

### DIFF
--- a/lib/graphql/schema/member/has_fields.rb
+++ b/lib/graphql/schema/member/has_fields.rb
@@ -20,16 +20,18 @@ module GraphQL
           is_object = self.respond_to?(:kind) && self.kind.object?
           # Local overrides take precedence over inherited fields
           visible_fields = {}
-          for ancestor in ancestors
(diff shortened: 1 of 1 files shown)


FunkyloverOne:
Hey @rmosolgo, now I think that most likely my proposed behavior here is incorrect. So maybe it makes sense for you to have a brief look at it to confirm it is wrong, and I will close this PR :slightly_smiling_face:

rmosolgo:
Yeah, I don't think this change would be an improvement. If an object's field is not visible, but it inherits a field from an interface which _is_ visible, then that visible field is used. (To hide that interface field, it should have `visible? => false`.)

PR #4114, opened 2022-06-24
Check for nullable result value to avoid errors on unsubscription

Maaarcocr:
# Why

Currently when I run `unsubscribe` inside an `update` method for a subscription class (in ruby) this causes the front-end to fails, because the server sends a new message to the front end that looks like: 

```
{
  more: false
}
```

Now if you try to result `payload.result.data` this will error out, as result is undefined and you cannot read `.data` of undefined. 

# How I fixed it

By using conditional access to said properties, we can safely read `data` and `errors` while not trying to read some properties when `result` is undefined

### DIFF
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -46,7 +46,7 @@ class ActionCableLink extends ApolloLink {
           )
         },
         received: function(payload) {
-          if (payload.result.data || payload.result.errors) {
+          if (payload?.result?.data || payload?.result?.errors) {
             observer.next(payload.result)
           }
 
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -46,7 +46,7 @@ class ActionCableLink extends ApolloLink {
           )
         },
         received: function(payload) {
-          if (payload?.result?.data || payload?.result?.errors) {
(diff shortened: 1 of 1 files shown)


javascript_client/src/subscriptions/ActionCableLink.ts
```diff
@@ -46,7 +46,7 @@ class ActionCableLink extends ApolloLink {
           )
         },
         received: function(payload) {
-          if (payload.result.data || payload.result.errors) {
+          if (payload?.result?.data || payload?.result?.errors) {
```
rmosolgo, line 49: In retrospect, maybe this should have been `if (payload.result)`, do you think that would work in your case?

Maaarcocr: done!

Maaarcocr: @rmosolgo could you run the CI and could we merge this if it's good?

Maaarcocr:
@rmosolgo the test suite is failing because the previous behaviour of not calling the observer when data is `null` like in: 
```
{
  result: {
    data: null
  },
  more: true
}
```

has not been kept if we just check result. 

Do you want me to change the tests or should I go back to my original implementation where I was doing: `payload.result?.data`

rmosolgo:
Sorry for the slow merge on this, and thanks for this fix!

rmosolgo:
Released in 1.11.1 :+1:

PR #4132, opened 2022-07-06
Fix boolean fields with hash_key and false value provided

nevesenin:
Hi! I stumbled upon an issue after upgrading to the latest version. I added tests that reproduce the actual issue. ~~Will try to provide a fix for that later.~~ Added fix.

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -666,7 +666,12 @@ module GraphQL
               inner_object = obj.object
 
               if defined?(@hash_key)
-                inner_object[@hash_key] || inner_object[@hash_key_str] || (@fallback_value != :not_given ? @fallback_value : nil)
+                hash_value = inner_object[@hash_key] || inner_object[@hash_key_str]
+                if type.unwrap.graphql_name == "Boolean" && (hash_value == true || hash_value == false)
+                  hash_value
+                else
+                  hash_value || (@fallback_value != :not_given ? @fallback_value : nil)
+                end
               elsif obj.respond_to?(resolver_method)
                 method_to_call = resolver_method
                 method_receiver = obj
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -668,8 +668,10 @@ module GraphQL
               if defined?(@hash_key)
                 hash_value = if inner_object.respond_to?(@hash_key)
                   inner_object.public_send(@hash_key)
-                else
+                elsif inner_object.is_a?(Hash)
                   inner_object.key?(@hash_key) ? inner_object[@hash_key] : inner_object[@hash_key_str]
+                else
+                  nil
                 end
                 if hash_value == false
                   hash_value
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -666,7 +666,11 @@ module GraphQL
               inner_object = obj.object
 
               if defined?(@hash_key)
-                hash_value = inner_object.to_h.key?(@hash_key) ? inner_object[@hash_key] : inner_object[@hash_key_str]
+                hash_value = if inner_object.respond_to?(@hash_key)
+                  inner_object.public_send(@hash_key)
+                else
+                  inner_object.key?(@hash_key) ? inner_object[@hash_key] : inner_object[@hash_key_str]
+                end
                 if hash_value == false
                   hash_value
                 else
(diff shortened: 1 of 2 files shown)


nevesenin:
I don't know yet, why the system_tests are failing.

lib/graphql/schema/field.rb
```diff
@@ -253,7 +253,7 @@ def initialize(type: nil, name: nil, owner: nil, null: nil, description: :not_gi
         method_name = method || hash_key || name_s
         @dig_keys = dig
         if hash_key
-          @hash_key = hash_key
+          @hash_key = hash_key.to_sym
```
nevesenin, line 256: If `@hash_key` is a string, hash won't be checked for symbol key otherwise.

lib/graphql/schema/field.rb
```diff
@@ -666,7 +666,12 @@ def resolve(object, args, query_ctx)
               inner_object = obj.object
 
               if defined?(@hash_key)
-                inner_object[@hash_key] || inner_object[@hash_key_str] || (@fallback_value != :not_given ? @fallback_value : nil)
+                hash_value = inner_object.to_h.key?(@hash_key) ? inner_object[@hash_key] : inner_object[@hash_key_str]
```
nevesenin, line 669: Check if key exists first before trying to get the value. This is to cover the case when `inner_object[@hash_key] == false`.

lib/graphql/schema/field.rb
```diff
@@ -666,7 +666,12 @@ def resolve(object, args, query_ctx)
...
+                if type.unwrap.graphql_name == "Boolean" && (hash_value == true || hash_value == false)
```
nevesenin, line 670: Return `hash_value` in case it's a boolean field. `if type.unwrap.graphql_name == "Boolean" && hash_value == false` might be sufficient here!?

rmosolgo: Honestly, I don't think we even have to require that the field is boolean. Something like this would be ok with me too, I think: 

```suggestion
                if hash_value == false
```

Because maybe the application has some use for a `false` value _other_ than the built-in `Boolean` type. What do you think?

rmosolgo (commented):
Thanks for the fix here! Sorry for my slow review. I just have a few questions.

lib/graphql/schema/field.rb
nevesenin: Agree. Didn't think about that case.

lib/graphql/schema/field.rb
nevesenin: If a `OpenStruct` is provided instead of a `Hash` this is needed if one want to use `#key?` method on it. Will add a `#respond_to?` call before. That way no Hash copy should be created.

lib/graphql/schema/field.rb
```diff
@@ -666,7 +666,18 @@ def resolve(object, args, query_ctx)
               inner_object = obj.object
 
               if defined?(@hash_key)
-                inner_object[@hash_key] || inner_object[@hash_key_str] || (@fallback_value != :not_given ? @fallback_value : nil)
+                hash_value = if inner_object.is_a?(Hash)
+                  inner_object.key?(@hash_key) ? inner_object[@hash_key] : inner_object[@hash_key_str]
+                elsif inner_object.respond_to?(@hash_key)
+                  inner_object.public_send(@hash_key)
```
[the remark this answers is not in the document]
nevesenin: > This seems a bit odd to send @hash_key as a method name
Totally agree. On the other hand, I found it weird to check for a method to be present and then not using it. ;) Your suggestion should work as well and will leave none of us with an odd feeling.
>  ... is this new behavior? 
No, OpenStruct provides an accessor for each provided key. I just used that.

lib/graphql/schema/field.rb
rmosolgo: Honestly, I'd _prefer_ if the only lookup that GraphQL-Ruby did was with the user-provided `hash_key`. But, early versions of GraphQL-Ruby added the string-based lookup, and ever since, I have used `@hash_key_str` to maintain compatibility. But I think it would be better if it only used the configured `hash_key:`, without any conversions. 

For this `.to_sym`, is this to retain compatibility with an old GraphQL-Ruby behavior? If not, I'd rather not include it in this fix. It might even break if anyone is using numbers as hash keys.

nevesenin: I see. I wasn't sure what this is about. I should have asked. In general I agree in being as explicit as possible so will remove that.

rmosolgo:
```suggestion
          @hash_key = hash_key
```

nevesenin: Sorry for being late. Removed the according tests.

rmosolgo:
Thanks for taking care of this fix!

PR #4134, opened 2022-07-08
Add `graphql-sources` to related projects

ksylvest:
`graphql-sources` is a collection of common [sources](https://graphql-ruby.org/dataloader/sources.html) to simplify loading data from `ActiveRecord`, fetching file from `ActiveStorage`, or using `Rails.cache`. The library documentation can be found here:

https://github.com/ksylvest/graphql-sources


rmosolgo:
Cool, thanks for sharing!

PR #4138, opened 2022-07-14
Don't print empty braces when input object has no arguments

### LINKED ISSUE #4135: Correctly dealing with InputObject without any fields
**Describe the bug**

Possible invalid schema generation

**Versions**

`graphql` version: 2.0.5
`rails` (or other framework): Rails-7.0.3
other applicable versions: graphql-batch (0.5.1)

**GraphQL schema**

See above

**GraphQL query**

See below

**Steps to reproduce**

Given I define an Input Object without any fields
When I generate the schema, I get the following result:

```gql
input SomeInputType {
}
```

When I try to read this with `@graphql-tools/load` I get an error.

However If I modify the output to 

```gql
input SomeInputType
```

reading with `@graphql-tools/load` works fine.

* According to the spec (https://spec.graphql.org/June2018/#sec-Input-Objects) it seems that input field should be optional, but the validation steps say at leas on field is required. 
* According to [a thread on Stack Overflow](https://stackoverflow.com/a/68356208) an empty list notation `{ }` is indeed invalid, but the list itself is optional.

**Expected behavior**

I'm not sure what is correct here, but I'm assuming the GraphQL lib should not render empty lists `{ }`. Both for InputObject and Objects. If not, maybe there should be runtime warnings for having InputObject/Objects without any field defined.

**Actual behavior**

See above

**Additional context**

No other context

rmosolgo:
Fixes #4135

### DIFF
--- a/lib/graphql/language/printer.rb
+++ b/lib/graphql/language/printer.rb
@@ -236,12 +236,15 @@ module GraphQL
         out = print_description(input_object_type)
         out << "input #{input_object_type.name}"
         out << print_directives(input_object_type.directives)
-        out << " {\n"
-        input_object_type.fields.each.with_index do |field, i|
-          out << print_description(field, indent: '  ', first_in_block: i == 0)
-          out << "  #{print_input_value_definition(field)}\n"
+        if !input_object_type.fields.empty?
+          out << " {\n"
+          input_object_type.fields.each.with_index do |field, i|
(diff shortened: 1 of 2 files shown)


Bertg:
@rmosolgo Any ETA on that release? We would like to use this fix ;)

rmosolgo:
Thanks for the bump on this -- I just released 2.0.12 to rubygems :+1:

PR #4139, opened 2022-07-15
Update documentation

bhicks:
Thanks for the great gem, and extensive documentation! 🎉

I noticed a small repetition when perusing, this change removes an unnecessary _the_ from the `sources.md` page.


rmosolgo:
Thanks for the fix!

PR #4142, opened 2022-07-17
Use Array#+ to insert an object to frozen Array instead of #dup and #<<

Watson1978:
This patch might improve performance where it insert an object into frozen array.

### Environment
- macOS 13.0 beta
- Apple M1 Max
- Apple clang version 14.0.0 (clang-1400.0.28.1)
- Ruby 3.1.2

### Benchmark result
```
             dup, <<   735.025k i/100ms
                   +     1.141M i/100ms
             dup, <<      7.374M (± 0.4%) i/s -     74.238M in  10.067378s
                   +     11.360M (± 0.5%) i/s -    114.081M in  10.042478s

Comparison:
                   +: 11360174.1 i/s
             dup, <<:  7374162.2 i/s - 1.54x  (± 0.00) slower
```

### Benchmark code
```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.time = 10

  path = ["__schema", "types"].freeze

  x.report('dup, <<') {
    next_path = path.dup
    next_path << 1
  }

  x.report('+') {
    next_path = path + [1]
  }

  x.compare!
end
```

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -401,8 +401,7 @@ module GraphQL
           end
           return_type = field_defn.type
 
-          next_path = path.dup
-          next_path << result_name
+          next_path = path + [result_name]
           next_path.freeze
 
           # This seems janky, but we need to know
@@ -768,8 +767,7 @@ module GraphQL
                   set_result(selection_result, result_name, response_list)
                   result_was_set = true
                 end
-                next_path = path.dup
-                next_path << idx
+                next_path = path + [idx]
                 this_idx = idx
                 next_path.freeze
                 idx += 1


Watson1978:
I executed the benchmark task `bundle exec rake bench:query`.

## Before changing
```
               query     3.000  i/100ms
               query     30.132  (± 0.0%) i/s -    153.000  in   5.077925s
```

## After applying this patch
```
               query     3.000  i/100ms
               query     30.497  (± 0.0%) i/s -    153.000  in   5.017079s
```

Watson1978:
## 1. Array#dup is slightly slow

スクリーンショット 2022-07-19 23 48 58

When we call `Array#dup`, `Object#dup` will be invoked (it named `rb_obj_dup` in C land).
`Object#dup` have to copy the values of instance variables in receiver to other.
(It will be executed by `init_copy` C function)
And finally, `Object#dup` will invoke `Object#initialize_dup` and `Object#initialize_copy` methods and `Array#initialize_copy` will be called.

https://github.com/ruby/ruby/blob/631b34ef95cdcc6cb21eaae339176bd2b63a9bca/object.c#L481-L493


`Array#+` will create new instance without above method callings.

https://github.com/ruby/ruby/blob/631b34ef95cdcc6cb21eaae339176bd2b63a9bca/array.c#L5053-L5068

## 2. Array literal might not so slow
スクリーンショット 2022-07-20 0 03 39

`ary_make_partial` will be called when creating an array by Array literal.
It will invoke fast paths when there are a few objects to embed.

https://github.com/ruby/ruby/blob/631b34ef95cdcc6cb21eaae339176bd2b63a9bca/array.c#L1320-L1325

PR #4145, opened 2022-07-22
Fix typo in code example in Argument Visibility docs

duffuniverse:
https://github.com/rmosolgo/graphql-ruby/blob/master/guides/authorization/visibility.md#argument-visibility

I guess there is a typo. Probably the `visible?` method should be used there

Compare the `authorized?` and `visible?` method signatures
https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/argument.rb#L144-L146
https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/argument.rb#L136-L138

### DIFF
--- a/guides/authorization/visibility.md
+++ b/guides/authorization/visibility.md
@@ -86,7 +86,7 @@ class Types::BaseArgument < GraphQL::Schema::Argument
     super(*args, **kwargs, &block)
   end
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Derp, sorry about the typo and thanks for the fix!

PR #4146, opened 2022-07-24
Fix prometheus tracing guide

OuYangJinTing:
Look [here](https://github.com/rmosolgo/graphql-ruby/blob/v2.10.0/lib/graphql/tracing.rb#L12-L14), `GraphQL::Tracing::PrometheusTracing::GraphQLCollector` only be required when used `PrometheusExporter::Server` startup, so it throwing `NameError` exception when used other server startup.

_PS: English is not my native language; please excuse typing errors._


rmosolgo:
Thanks for the fix :+1:

PR #4159, opened 2022-08-10
Remove deprecated datadog tracer analytic options

TonyCTHsu:
**Why & What** 

Datadog GraphQL integration is deprecating `analytics_enabled`, `analytics_sample_rate`. Hence, marking it with comment, and removing it from documentation to prevent it more usage

https://github.com/DataDog/dd-trace-rb/pull/2214


rmosolgo:
Thanks for the update!

PR #4169, opened 2022-08-25
Use "an" before a vowel sound

Terris:
"The general rule is to use "a" when the indefinite article precedes a word beginning with a consonant sound and an when it precedes a word starting with a vowel sound."


rmosolgo:
Thanks for the fix!

Issue #4064, opened 2022-05-06
Support for OneOf Input Objects


erikkessler1:
**Is your feature request related to a problem? Please describe.**

The OneOf Input Objects RFC, graphql/graphql-spec#825, which introduces a form of input polymorphism to GraphQL, has entered stage 2. Given that the proposal is moving along, we should think about how we could implement the feature in this library.

**Describe the solution you'd like**

This issue is intended to foster a discussion of options, so to start the discussion, one approach would be to add a new `GraphQL::Schema::OneOfInputObject` type. For example:

```ruby
class PetInput < GraphQL::Schema::OneOfInputObject
  member :cat, CatInput
  member :dog, DogInput
end
```
_Ruby DSL_

```graphql
input PetInput @oneOf {
  cat: CatInput
  dog: DogInput
}
```
_Generated GraphQL SDL_

With this approach, we'd be able to ensure arguments are nullable and don't have defaults. However, a downside to this approach is sharing code with an application's `BaseInputObject` becomes harder unless the logic is already factored into composeable modules.

**Describe alternatives you've considered**

Another option would be to add a new DSL method to `GraphQL::Schema::InputObject` that turns it into a OneOf Input Object. For example,

```ruby
class PetInput < GraphQL::Schema::InputObject
  one_of

  argument :cat, CatInput
  argument :dog, DogInput
end
```

This approach makes it easier to share logic with a `BaseInputObject`, but makes it less clear to developers that not all the argument configuration options are available (e.g., a default can't be provided). Furthermore, it glosses over that fact that Input Objects and OneOf Input Objects are truly different concepts in the type system and serve different purposes.

rmosolgo:
:+1: I'm definitely game to start supporting this feature. 

Here's another implementation suggestion: what if the existing `required: { one_of: [...] }` validation added the `@oneOf` directive to the SDL for input objects when all their arguments are named by `one_of: [...]`? ([guide](https://graphql-ruby.org/fields/validation.html#built-in-validations), [API doc](https://graphql-ruby.org/api-doc/2.0.7/GraphQL/Schema/Validator/RequiredValidator)).


My second choice would have been the `one_of` configuration method as described above. The issue about `default_value` can be solved by checking for invalid configuration while the input object is being defined. But about this part: 

>  truly different concepts in the type system and serve different purposes

could you help me understand that better? (It seems to me like `@oneOf` is basically a _special kind_ of Input Object, but maybe I'm missing a perspective on it.)

[...]

PR #4185, opened 2022-09-06
Add support for OneOf Input Objects

erikkessler1:
This PR implements OneOf Input Objects as described in the RFC: https://github.com/graphql/graphql-spec/pull/825. Unlke https://github.com/rmosolgo/graphql-ruby/pull/4184 which uses a `one_of` DSL method on `BaseInput`, this PR allows consumers a `GraphQL::Schema::OneOfInputObject` allowing for a cleaner interface as outlined [here](https://github.com/rmosolgo/graphql-ruby/issues/4064#issuecomment-1126718392):

```ruby
class MyOneOfInput < BaseOneOfInputObject
  member :string, String
  member :int, Integer
end
```

Fixes: https://github.com/rmosolgo/graphql-ruby/issues/4064


rmosolgo:
Thanks for sharing this approach -- I cherry-picked some of your work into #4184  to make sure it would cover the same bases. I cherry-picked your static validation approach too, since it was much better than mine 😅  It should support the class-based API you had in mind if you add this: 


```ruby 
class OneOfInputObject < GraphQL::Schema::InputObject 
  one_of 
  
  def self.member(*args, **kwargs, &block)
    argument(*args, required: false, **kwargs, &block)
  end 
end 
```

I also included a test of that API in the project to make sure it keeps working in the future :+1: Please let me know if you run into any trouble with it!

Issue #4187, opened 2022-09-07
Lookahead#selects? performance


alexcwatt:
**Is your feature request related to a problem? Please describe.**

I've got a field definition with `extras: [:lookahead]` that does something like this:

```
do_extra_thing = llookahead&.selections&.any? do |sel|
  [:field_1, :field_2, :field_3].any? { |field| sel.selects?(field) }
end || false
```

It turns out that for a fairly large production GraphQL request, a significant chunk of object allocations (5% of total) happens in `normalize_name`:

https://github.com/rmosolgo/graphql-ruby/blob/9cc79a95b66b9719815fd41801e227b20944e02e/lib/graphql/execution/lookahead.rb#L200-L206

Every time `selects?` is called with a symbol, even if it's a symbol that's been seen before, it has to be camelized.

**Describe the solution you'd like**

It'd be nice if we could somehow avoid these object allocations. Perhaps the camelization could be memoized (but I'm not sure how that would work since the `Lookahead` object is new each time so I don't think it makes sense to memoize there) or we could have a one-time mapping at the query level from the GraphQL version of the field name to the Ruby version ... or do this in schema generation or something ...

I don't have a particular solution that I know would work, but I'm hoping there might be something simple to prevent fields from being re-camelized.

rmosolgo:
Thanks for the detailed report, and sorry for the trouble! I have to admit I haven't considered the performance of Lookaheads closely... and perhaps it shows. I'm sure we can improve it.  I'll take a look soon and follow up here.

rmosolgo:
(Of course, in the meantime, you could switch those symbols to strings to work around this issue!)

rmosolgo:
Were you able to give this a try on 2.0.14? Please let me know what you find if so.

alexcwatt:
@rmosolgo I have a note to check on this after we're able to upgrade to 2.0.14. Do you want to close this for now? The fix looks promising.

PR #4189, opened 2022-09-08
Lookahead: don't coerce symbol to string if a match can be made with the symbol

rmosolgo:
Part of fixing #4187

### DIFF
--- a/lib/graphql/execution/lookahead.rb
+++ b/lib/graphql/execution/lookahead.rb
@@ -87,16 +87,28 @@ module GraphQL
 
       # Like {#selects?}, but can be used for chaining.
       # It returns a null object (check with {#selected?})
+      # @param field_name [String, Symbol]
       # @return [GraphQL::Execution::Lookahead]
       def selection(field_name, selected_type: @selected_type, arguments: nil)
-        next_field_name = normalize_name(field_name)
+        next_field_defn = case field_name
+        when String
+          @query.get_field(selected_type, field_name)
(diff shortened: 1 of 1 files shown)


lib/graphql/execution/lookahead.rb
```diff
@@ -87,16 +87,28 @@ def selected?
...
+        next_field_defn = case field_name
+        when String
+          @query.get_field(selected_type, field_name)
+        when Symbol
+          # Try to avoid the `.to_s` below, if possible
+          all_fields = @query.warden.fields(selected_type)
+          if (match_by_orig_name = all_fields.find { |f| f.original_name == field_name })
+            match_by_orig_name
+          else
+            guessed_name = Schema::Member::BuildType.camelize(field_name.to_s)
```
rafaelfranca, line 102: If all you want to avoid is the new allocation when doing `field_name.to_s`, doesn't this work:

```
guessed_name = Schema::Member::BuildType.camelize(-field_name.to_s)
```

rmosolgo: Thanks for taking a look, @rafaelfranca. I investigated and found that `-field_name.to_s` _does_ allocate a new string, although it then de-duplicates it with an existing frozen string in which is already in memory. So, it would help in a place where that string was retained, but it didn't prevent the initial allocation. 

However, I did find `Symbol#name` which reduces allocations here, I'll use that it #4212

PR #4195, opened 2022-09-13
Deconstruct the interpreter

rmosolgo:
This is mostly an exploration into the question, can runtime features be _teased out_ (however ugly) of the existing execution code? If so, maybe GraphQL-Ruby could offer a mix-your-own runtime API.

Make it so that each of these can be opted in to (but all are on by default, for compatibility's sake):

- [x] Query/multiplex instrumentation 
- [x] `.trace`
- [ ] Lazy values 
- [ ] Object type instantiation / `.authorized?` 
- [ ] Custom field resolution (instead, just call methods by field name) 
- [ ] Returning errors / skip (source of many `.is_a?` checks)
- [x] `context[:current_....]`, source of many `Hash#[]=` calls
- [ ] others?


rmosolgo:
I'm exploring these performance improvements by other avenues, for example, #4344 #4342 #4330

PR #4197, opened 2022-09-15
Replace `max_validation_errors` with `validate_max_errors` in CHANGELOG

pocke:
CHANGELOG mentions `max_validation_errors` method but actually the method does not exist. There is `validate_max_errors` instead.
This PR replaces `max_validation_errors` with `validate_max_errors`.

---

In #2386, this feature was developed as `max_validation_errors` but this PR was not merged. This feature has been added as `validate_max_errors` in another PR. I guess it is the reason why there was the confusion in CHANGELOG.


rmosolgo:
Thank you for the correction!

PR #4203, opened 2022-09-20
Support apollo-style schema extensions

### LINKED ISSUE #4199: Parsing an SDL file with `extend schema` leads to `NoMethodError` (`name` on `GraphQL::Language::Nodes::SchemaExtension`)
**Describe the bug**

[Apollo requires](https://www.apollographql.com/docs/federation/federated-types/federated-directives#importing-directives) the following in a subgraph schema for Apollo Federation V2:

``` graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.0",
        import: ["@key", "@shareable"])
```

However, when I try to parse an SDL file that contains this using graphql-ruby, I get a `NoMethodError`

**Versions**

`graphql` version: 2.0.14

**GraphQL schema**

See the script below for a full runnable example of this bug.

**GraphQL query**

N/A

**Steps to reproduce**

Put this file in `tmp/graphql_ruby_bug.rb`:

``` ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", "2.0.14"
end

schema_sdl = <<~EOS
  extend schema
    @link(url: "https://specs.apollo.dev/federation/v2.0",
          import: ["@key", "@shareable"])

  type Query {
    something: Int
  }
EOS

require "graphql"

GraphQL::Schema.from_definition(schema_sdl)
```

Then run with:

```
$ ruby tmp/graphql_ruby_bug.rb
```

**Expected behavior**

This SDL string should parse without throwing any exceptions.

**Actual behavior**

```
$ ruby tmp/graphql_ruby_bug.rb                                                                                                                                                                                                                                                                        Traceback (most recent call last):
        6: from tmp/graphql_ruby_bug.rb:20:in `<main>'
        5: from /Users/myron/.rvm/gems/ruby-2.7.5/gems/graphql-2.0.14/lib/graphql/schema.rb:122:in `from_definition'
        4: from /Users/myron/.rvm/gems/ruby-2.7.5/gems/graphql-2.0.14/lib/graphql/schema/build_from_definition.rb:10:in `from_definition'
        3: from /Users/myron/.rvm/gems/ruby-2.7.5/gems/graphql-2.0.14/lib/graphql/schema/build_from_definition.rb:18:in `from_document'
        2: from /Users/myron/.rvm/gems/ruby-2.7.5/gems/graphql-2.0.14/lib/graphql/schema/build_from_definition.rb:68:in `build'
        1: from /Users/myron/.rvm/gems/ruby-2.7.5/gems/graphql-2.0.14/lib/graphql/schema/build_from_definition.rb:68:in `each'
/Users/myron/.rvm/gems/ruby-2.7.5/gems/graphql-2.0.14/lib/graphql/schema/build_from_definition.rb:74:in `block in build': 
undefined method `name' for #<GraphQL::Language::Nodes::SchemaExtension:0x0000000130a2b580> (NoMethodError)
```

**Additional context**

It looks like the `case` statement clause here is missing `GraphQL::Language::Nodes::SchemaExtension`.

https://github.com/rmosolgo/graphql-ruby/blob/v2.0.14/lib/graphql/schema/build_from_definition.rb#L70

If I add `GraphQL::Language::Nodes::SchemaExtension` to it, I no longer get the exception.

rmosolgo:
A few changes to support building definitions from Apollo Federation-style schema extensions

Fixes #4199 

- Add `Schema.schema_directives` which are the directives _on the schema itself_ (`Schema.directives` are the directives which may be used within the schema somewhere)
- Print directive definitions at the beginning of schema print-outs 
- Support `extend schema` with directives (most `extend`s are still not supported) 

__TODO__: it seems like apollo expects directives to work even when they aren't defined in the SDL, should graphql-ruby support this?

### DIFF
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -44,16 +44,18 @@ module GraphQL
       end
 
       def build_schema_node
-        GraphQL::Language::Nodes::SchemaDefinition.new(
-          query: (q = warden.root_type_for_operation("query")) && q.graphql_name,
-          mutation: (m = warden.root_type_for_operation("mutation")) && m.graphql_name,
-          subscription: (s = warden.root_type_for_operation("subscription")) && s.graphql_name,
-          # This only supports directives from parsing,
-          # use a custom printer to add to this list.
-          #
+        schema_options = {
           # `@schema.directives` is covered by `build_definition_nodes`
-          directives: ast_directives(@schema),
-        )
+          directives: definition_directives(@schema, :schema_directives),
+        }
+        if !schema_respects_root_name_conventions?(@schema)
+          options.merge!({
+            query: (q = warden.root_type_for_operation("query")) && q.graphql_name,
+            mutation: (m = warden.root_type_for_operation("mutation")) && m.graphql_name,
+            subscription: (s = warden.root_type_for_operation("subscription")) && s.graphql_name,
+          })
+        end
+        GraphQL::Language::Nodes::SchemaDefinition.new(schema_options)
       end
 
       def build_object_type_node(object_type)
@@ -283,7 +285,9 @@ module GraphQL
       private
 
       def include_schema_node?
-        always_include_schema || !schema_respects_root_name_conventions?(schema)
+        always_include_schema ||
+          !schema_respects_root_name_conventions?(schema) ||
+          !schema.schema_directives.empty?
       end
 
       def schema_respects_root_name_conventions?(schema)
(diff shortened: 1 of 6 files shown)


myronmarston:
> **TODO**: it seems like apollo expects directives to work even when they aren't defined in the SDL, should graphql-ruby support this?

I don't think so.  I chatted with the apollo folks this morning about this:

image

I think the Apollo folks hope that long term there's a standard way to import directives in an SDL file that the ecosystem supports but in the meantime, `@link` (and the other apollo directives) shouldn't be treated as being "special" by graphql-ruby (or any other language graphql library); it's just a directive that happens to have meaning to apollo's tools but that needs normal definitions like any other custom directive.  My expectation is that I have to define the `@link` and other apollo directives in my SDL file for the graphql gem to be able to parse it.

myronmarston:
Thanks for addressing the bug I reported so quickly!  I'm looking forward to the 2.0.15 release.

PR #4208, opened 2022-09-23
Strip blank lines from introspection query

sambostock:
#4184 added

```diff
 fragment FullType on __Type {
   kind
   name
   description
   #{include_specified_by_url ? "specifiedByURL" : ""}
+  #{include_is_one_of ? "isOneOf" : ""}
   fields(includeDeprecated: true) {
     name
     description
```

to the introspection query, which introduces a new blank line when the condition is false and needlessly changes the query's fingerprint.

By stripping out blank lines, we ensure that the query is stable even if new conditional lines are added that don't actually end up adding any content.


rmosolgo:
Ah -- that makes great sense. Thanks for the improvement! I'll follow up with a test after merging to make sure we don't lose this behavior in the future.

PR #4222, opened 2022-10-08
Lint lexer.rb and parser.rb after they are generated

pawandubey:
👋🏼 Hello friends. I am proposing that we lint the generated lexer and parser files at the end of the `build_parser` task.

The generated files are not properly indented, which generates a lot of warnings when loaded by client apps. (Ref. https://github.com/rmosolgo/graphql-ruby/issues/1256)

It may (should?) be possible to correct the indentations by hand-adjusting the grammar files but rubocop does the job admirably well, and as the grammar files are rarely touched (only a couple of commits per year have modified the lexer/parser in the last few years), it seems like an acceptable option to me as it gets rid of the warnings just as well.

We only make layout autocorrections (-x)  to err on the side of caution, although it does miss adding the frozen_string_literal comment.

I've committed the result of running the task on the existing files here for your perusal too.


rmosolgo:
I can't say the resulting code _looks_ any better to me: 

https://github.com/rmosolgo/graphql-ruby/blob/f4f3e04aee322e49b0b25a2ca35552adfe6b8233/lib/graphql/language/lexer.rb#L183-L190

But, could you confirm that this makes warnings go away when these files are loaded?

pawandubey:
@rmosolgo Ah, you are right. Basically it's because the `.rubocop.yml` for the project disables most cops that would be required to fix the layout issues. As it is, the warnings are not fixed but if I disable the current config then they are.

I think we can create a new rubocop config specifically for linting these files and use it to run the command, so as to not affect any of the ungenerated code. But at this point, I am happy to go with your recommendation here.

rmosolgo:
It's not important to me that these files are formatted correctly; they're not for human reading. However, if it'd work better for you to have them formatted, I'm also open to formatting them properly. A separate Rubocop config sounds like a totally appropriate solution, so please feel free to add it here, or if you're done, closing the PR is fine with me too :+1:

rmosolgo:
Feel free to take this up again if you're still interested, but I'm going to close this for now since it's inactive.

PR #4223, opened 2022-10-10
Avoid using Error.captureStackTrace

jscheid:
As described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error,) `Error.captureStackTrace` is only available in V8.


rmosolgo:
Thanks for the improvement! Weird about the failing build. I'll track it down if it resurfaces on master.

PR #4229, opened 2022-10-14
Add full stop to a description method string for consistency

devonmathershopify:
Use of a full stop in the description field appears to be missing. Adding it to be consistent with other description method calls in the code base.

Matches the use of description in other files
- [lib/graphql/types/relay/node_behaviors.rb](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/relay/node_behaviors.rb#L9)
- [lib/graphql/types/relay/edge_behaviors.rb](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/relay/edge_behaviors.rb#L8)
- [lib/graphql/types/relay/connection_behaviors.rb](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/relay/connection_behaviors.rb#L63)


rmosolgo:
Fair enough -- thanks for the fix!

Issue #4253, opened 2022-11-22
Double call of scope_items method before and after pagination


andrey-lizunov:
**Describe the bug**

Hello.
I see weird behaviour in my project when resolver with cursor pagination calls scope_items method in type two times.
First time the scope in this method is a full set of data without pagination. Second time the scope with pagination.
It affects the performance because first call works with big amounts of data.
I use scope_items method to remove records that not available for current user and policy checks for thousands of records is not what I want :)

**Versions**

`graphql` version: 2.0.15
`rails` (or other framework): 6.1.5

**GraphQL schema**
```
class OfferingType < ::Domains::Types::Base::Object
  ...
  def self.scope_items(items, context)
    items.select do |item|
      ::Core::OfferingPolicy.new(context[:current_user], item).show?
    end
  end
end
```

```
class OfferingsResolver < ::Domains::Types::Base::Query
  include SearchObject.module(:graphql)
  type ::Core::Types::OfferingType.connection_type, null: false
  ...
end
```

**GraphQL query**

```graphql
query Offerings {
  offerings(
    first: 3,
  ) {
    edges {
      cursor
    }
    nodes {
      id
    }
  }
}
```

```json
{
  "data": {
    "offerings": {
      "edges": [
        {
          "cursor": "MQ"
        },
        {
          "cursor": "Mg"
        },
        {
          "cursor": "Mw"
        }
      ],
      "nodes": [
        {
          "id": "1407"
        },
        {
          "id": "486"
        },
        {
          "id": "1190"
        }
      ]
    }
  }
}```

**Steps to reproduce**

Steps to reproduce the behavior

**Expected behavior**

scope_items method call should be made once for paginated data set

**Actual behavior**

scope_items call made for non paginated and paginated data

**Additional context**

Add any other context about the problem here.

With these details, we can efficiently hunt down the bug!

rmosolgo:
Hi, thanks for the detailed report! I agree the list should only be scoped once and that the current behavior is a bug.

I think the reason for this bug is that connection types make an _extra_ call to `.scope_items`, then the connection's `nodes` (or `edges`) list also makes a call to scope items. (`Field#scoped?` returns true for _both_ list-type fields and connection-type fields.) So, I think the fix is to somehow make `Field#scoped?` return `false` when the field is a list-type, but when it belongs to a Connection type.

andrey-lizunov:
@rmosolgo could you check the PR and provide your feedback? Does it look like valid fix?

[...]

PR #4254, opened 2022-11-22
Fix for double scope_items method call (Issue #4253)

andrey-lizunov:
You can find the description of the case in this [issue](https://github.com/rmosolgo/graphql-ruby/issues/4253).

### DIFF
--- a/lib/graphql/schema/field/scope_extension.rb
+++ b/lib/graphql/schema/field/scope_extension.rb
@@ -9,7 +9,7 @@ module GraphQL
             value
           else
             ret_type = @field.type.unwrap
-            if ret_type.respond_to?(:scope_items)
+            if ret_type.respond_to?(:scope_items) && !field.connection?
               ret_type.scope_items(value, context)
             else
               value


rmosolgo:
I think this _would_ remove one of the calls to `.scope_items`, but if I understand correctly, I think it would remove the call to the connection field instead of the call to the list of nodes and/or edges. I don't think we want this for two reasons: 

- For pagination, we want `.scope_items` to be called at the connection level, not during the list of nodes. That way, the final list of nodes and `hasNextPage`/`hasPreviousPage` would be based on the _scoped_ list instead of the unscoped list. 
- If we made a query that used both `nodes { ... }` and `edges { ... }`, we'd still get `.scope_items` called twice. (Currently in the library, it'd be called _three_ times 🤦 ). 

It strikes me now that one option might be to add `scope: false` to the default options for `nodes` and `edges` fields: 

https://github.com/rmosolgo/graphql-ruby/blob/17bf301c2a5693d63ed6caa85a043c2f0fcef3c2/lib/graphql/types/relay/connection_behaviors.rb#L47-L53

https://github.com/rmosolgo/graphql-ruby/blob/17bf301c2a5693d63ed6caa85a043c2f0fcef3c2/lib/graphql/types/relay/connection_behaviors.rb#L132-L138

I suppose you could even use that as a work-around in your app in the meantime: 

```ruby 
class Types::BaseConnection < Types::BaseObject 
  include GraphQL::Types::Relay::ConnectionBehaviors 
  # don't re-apply scoping to these fields, see https://github.com/rmosolgo/graphql-ruby/issues/4253
  nodes_field(field_options: { scope: false })
  edges_field(field_options: { scope: false }) 
end 
```

Want to give it a try?

rmosolgo:
I think this was addressed by #4263 :+1:

PR #4255, opened 2022-11-24
Update testing.md

suke:
Hi, I have added a small fix to testing.md.
Use `3.times.map` instead of `3.times` to convert to an array.


rmosolgo:
Thanks for the fix!

PR #4259, opened 2022-12-01
[DOCS] Update Pusher's message size limit link

barsumek:
When I was following docs I noticed that the previous link about Pusher's 10kb message limit does no longer work.

I updated it to the newest one, but this problem might happen again. :(


rmosolgo:
Thanks for the fix! It's certainly the best we can do for now.

PR #4275, opened 2022-12-22
Removes hubburu project

peternycander:
I have sadly decided to sunset the Hubburu project (I am the founder)


rmosolgo:
Thanks for the update!

Issue #4180, opened 2022-08-31
Build a smaller, faster runtime with fewer features


rmosolgo:
To me, the "interpreter" feels brand new, but in fact, it's almost 4 years old. Many people have run into performance problems that strike at the heart of how the interpreter works (TODO list them here, eg #3936).

Personally, I think it's time to consider a re-vamp of the GraphQL-Ruby runtime with (yet another) specific focus on performance, perhaps by compromising feature-completeness. 

In my dreams, you could build a system like this: 

- A minimal runtime module which "just" calls methods on objects based on the incoming AST, building the results into a valid response Hash (using the schema's type system). Theoretically, you could use this, along with the schema configuration API, for a _very_ minimal GraphQL runtime.
- Mixins or plugins, or whatever, for adding existing graphql-ruby features back into the system (for example, initializing `GraphQL::Schema::Object`s, `authorized?`, sophisticated field resolution, error handling, tracing, instrumentation, field extensions, lazy resolution, dataloader, `context[:current_...]` values, and so on), so that any application can grab the features it _wants_ without slowdowns from features it isn't using. 

This would be a pretty big chunk of work, but I think it's the only way to improve performance in critical cases while still supporting all the various features that have cropped up over the years. 

Fixes #3998 

If anyone has suggestions, questions, or comments while I ponder the idea, I'd welcome some discussion on this issue!

benfalk:
I've been looking into this as well for our company as we have been running into this as well.  If there is anything I can do to dig in I'd be happy to help where and when I can.  The lazy resolve and extensions phases seem to be where we loose performance.  That and we actually have the worse case scenario in the field execution where it's checking to see if the field responds to a method, is a hash, or its inner method responds to a method.  I was playing around with being able to define the primary strategy a field uses to resolve its data; and that looks like it might be a little bit of a help.

rmosolgo:
Thanks for chiming in, @benfalk -- that's definitely the kind of stuff I'd like to improve.

benfalk:
Something I have done recently to help highlight this problem potentially.  Instead of letting the GQL system drill into the fields; for connections that return the same objects every time I have setup a system to produce `RawValue` for the node.  This has ended up shaving about 200ms of render time from our requests.  I've been wondering more on how I might setup a system to dynamically compile known queries into a set of strategies that produce these `RawValue` fields where it make sense.

DmitryTsepelev:
[...]

PR #4280, opened 2022-12-29
Minimal interpreter MVP

DmitryTsepelev:
Hey, Merry Christmas 🙂

I found #4180 and decided to play around it. Here we go: a minimal interpreter that can only pass a first spec (no errors, no unions and interfaces etc), but seems to _work_. Before I keep going I wanted to get some input:

- does it look like a thing that you described or it's somehow conceptually wrong?
- what should we do to make it happen? I guess we need it to at least have errors and a plugin for the auth

### DIFF
--- a/lib/graphql/execution.rb
+++ b/lib/graphql/execution.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require "graphql/execution/directive_checks"
+require "graphql/execution/minimal_interpreter"
 require "graphql/execution/interpreter"
 require "graphql/execution/lazy"
 require "graphql/execution/lookahead"
--- a/lib/graphql/execution/minimal_interpreter.rb
+++ b/lib/graphql/execution/minimal_interpreter.rb
@@ -1,0 +1,193 @@
+module GraphQL
+  module Execution
+    class MinimalInterpreter
+      module Underscore
+        refine String do
+          def underscore
+            if match?(/\A[a-z_]+\Z/)
+              return self
+            end
+            string = self.dup
+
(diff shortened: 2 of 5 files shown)


rmosolgo:
Hey! This is definitely the _idea_ -- what could we do to provide a runtime with _minimal_ overhead? As far as making it happen, here are a few requirements in my own mind: 

- Any new runtime options must be swappable while the application is running (for example, by a feature flag), so that the new runtime can be released in production incrementally. I don't want gem users to have to cross their fingers, deploy, and hold their breath to see whether it 💥 s in production; instead, I'd like to support a "1% of traffic"-style roll-out which could be reverted without a redeploy. 
- The "maximal" and "minimal" runtimes should use the same code and same tests, somehow. GraphQL-Ruby has a lot of tests for a lot of edge cases, and I consider that to be a real asset. I'd like any new runtime to pass all the same tests as the existing one (except for features that the new runtime doesn't implement ...). _Last time_ I re-wrote query execution, I accomplished this by running CI with an environment variable (`LEGACY_RUNTIME=1`, for example), and that worked OK, although since then I've learned to appreciate using modules or inheritance to share tests between setups, and I'd pursue that approach instead if it was up to me. 
- In the same vein as above, I'd like the minimal and maximal runtimes to relate to one another like a spectrum, where features can be added or removed. Similarly to runtime swapping, this would support an incremental adoption (or removal) of runtime features with a less risky deployment.

I started trying to untangle interpreter features in #4195, and it worked _some_, but I haven't gone after the most complicated runtime features yet.

rmosolgo:
It doesn't seem like this is still in active development, so I'll close it.

PR #4286, opened 2023-01-10
1.13: lookup given hash_key: when possible

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/pull/4072#issuecomment-1375790414

When a field is configured with `hash_key:`, look up that key using `[...]`, even if the underlying object isn't a Hash. (This code checks for `.respond_to?(:[])` to avoid breaking changes, which might be possible because of how `hash_key` was previously merged into `method_*` attributes.)

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -258,6 +258,10 @@ module GraphQL
         # TODO: I think non-string/symbol hash keys are wrongly normalized (eg `1` will not work)
         method_name = method || hash_key || name_s
         @dig_keys = dig
+        if hash_key
+          @hash_key = hash_key
+        end
+
         resolver_method ||= name_s.to_sym
 
         @method_str = -method_name.to_s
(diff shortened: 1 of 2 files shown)


alexcwatt:
@rmosolgo Thank you - this is what I was looking for!

rmosolgo:
Great -- released to RubyGems in v1.13.18!

alexcwatt:
I'm not seeing it [here](https://rubygems.org/gems/graphql/versions) ... maybe due to [this](https://github.com/rmosolgo/graphql-ruby/actions/runs/3886664200/jobs/6632026985)?

rmosolgo:
🤦 🤦 🤦 Meanwhile, back in my terminal: 

image

Amazingly, entering a live code hours later worked fine 😅 💁‍♂️ https://rubygems.org/gems/graphql/versions/1.13.18

Unfortunately that GitHub Action won't pass without a bit of work, since the workflow file is so old, so the docs won't get properly published...

PR #4288, opened 2023-01-11
Avoid anonymous eval

casperisfine:
It makes it hard to locate code when profiling and such.


rmosolgo:
Thanks for the improvement!

Issue #4113, opened 2022-06-23
`UnauthorizedFieldError` should not inherit `UnauthorizedError`


FunkyloverOne:
**Describe the bug**

I believe that `GraphQL::UnauthorizedFieldError` should not be inherited from `GraphQL::UnauthorizedError`, as it opens up possibilities of "false" rescues of `UnauthorizedFieldError`, and it brings inconsistency in its handling.

**Versions**

`graphql` version: 1.13.14

**Steps to reproduce**

if there's an unauthorized field, and authorization logic is wrapped in a promise the resulting error will look as if it's `UnauthorizedError`, instead of `UnauthorizedFieldError`.

**Expected behavior**

`UnauthorizedFieldError` should be rescued with its specific logic.

**Actual behavior**

`UnauthorizedFieldError` is rescued with `UnauthorizedError`'s logic.

rmosolgo:
If anyone has specific issues with this inheritance, please share an example here or in a new issue and I'd be happy to look into further solutions :+1:

PR #4289, opened 2023-01-11
Fix wrong schema hook call on lazy denial from field authorization

id-ilych:
When field authorization is lazily denied, the interpreter correctly calls `unauthorized_field` schema hook (used to wrongly call `unauthorized_object` instead).

Additionally, what do you think about creating `GraphQL::UnauthorizedObjectError` and removing all the mentions of `GraphQL::UnauthorizedError` (except for declaration and inheritance) to encourage a conscious decision since the logic is not always the same for objects and fields (as the bug shows).

Also, what do you think about asserting both `data` AND `errors` in every case of the test not to miss other edge cases like that?

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -615,6 +615,16 @@ module GraphQL
                 end
               end
               HALT
+            elsif value.is_a?(GraphQL::UnauthorizedFieldError)
+              value.field ||= field
+              # this hook might raise & crash, or it might return
+              # a replacement value
+              next_value = begin
+                schema.unauthorized_field(value)
+              rescue GraphQL::ExecutionError => err
+                err
+              end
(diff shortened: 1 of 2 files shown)


FunkyloverOne:
I believe it fixes the issue I had when I was creating this one: https://github.com/rmosolgo/graphql-ruby/issues/4113

FunkyloverOne:
Hey @rmosolgo - all checks have passed :sweat_smile:

rmosolgo:
Cool, thanks for the fix on this! Sorry about the trouble, I agree with your fix here. 

I think it'd be fine to add `UnauthorizedObjectError` to the gem and use it as you suggest. It should still work for anyone using `UnauthorizedError`, since it inherits from it, but as you describe, it will help us keep these codepaths straight. 

I'm certainly open to asserting `data` and `errors`, too, as often as I can (and remember)!

PR #3153, opened 2020-09-15
Tweaks to improve retained memory of Schema.from_definition

rmosolgo:
We use this to parse our internal schema and the memory footprint was getting out of control. 

```ruby
require "bundler/inline"

gemfile do
  gem "memory_profiler"
  gem "graphql", path: "."
  # gem "graphql", "1.11.4"
end

require "memory_profiler"
require "graphql"


str = File.read("/path/to/schema.graphql")
schema = nil
report = MemoryProfiler.report do
  schema = GraphQL::Schema.from_definition(str)
end

report.pretty_print
```

<details>
<summary>Before.txt</summary>

<p>

```
Total allocated: 193024494 bytes (1688837 objects)
Total retained:  26853202 bytes (231866 objects)

allocated memory by gem
-----------------------------------
 166860307  graphql-1.11.4
  26164187  other

allocated memory by file
-----------------------------------
  36104072  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/schema/field.rb
  21210176  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/schema/member/has_fields.rb
  21048850  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/schema/build_from_definition.rb
  19499967  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/language/lexer.rb
  17431568  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/schema/member/accepts_definition.rb
  15252488  parser.y
  14015597  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/language/block_string.rb
   9986920  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/language/nodes.rb
   8958056  /Users/rmosolgo/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/graphql-1.11.4/lib/graphql/schema/member/build_type.rb
[...]

PR #4293, opened 2023-01-16
Improve Shape Friendliness

casperisfine:
Ruby 3.2 significantly changed how instance variables are store. It now use shapes, and in short, it's important for performance to define instance variables in a consistent order to limit the amount of shapes.

Otherwise, the number of shapes will increase past a point where MRI won't be able to cache instance variable access. The impact is even more important when YJIT is enabled.

This PR is data driven. I dump the list of Shapes from Shopify's monolith production environment, and GraphQL is the top offender:

```
Shape Edges Report
       770  @default_graphql_name
       697  @own_fields
       661  @to_non_null_type
       555  @own_interface_type_memberships
       472  @description
       389  @errors
```

NB: This PR doesn't aim to fix all of these issues, just the low hanging fruits.

cc @tenderlove @jemmaissroff

### DIFF
--- a/lib/graphql/schema/directive.rb
+++ b/lib/graphql/schema/directive.rb
@@ -93,6 +93,18 @@ module GraphQL
         def repeatable(new_value)
           @repeatable = new_value
         end
+
+        private
+
+        def inherited(subclass)
+          super
+          subclass.class_eval do
+            @default_graphql_name = nil
+            @repeatable = nil
+            @default_directive = nil
+            @locations = nil
+          end
+        end
       end
 
       # @return [GraphQL::Schema::Field, GraphQL::Schema::Argument, Class, Module]
--- a/lib/graphql/schema/late_bound_type.rb
+++ b/lib/graphql/schema/late_bound_type.rb
@@ -9,6 +9,8 @@ module GraphQL
       alias :graphql_name :name
       def initialize(local_name)
         @name = local_name
+        @to_non_null_type = nil
+        @to_list_type = nil
       end
 
       def unwrap
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -27,7 +27,7 @@ module GraphQL
         end
 
         def overridden_graphql_name
-          @graphql_name
+          defined?(@graphql_name) ? @graphql_name : nil
         end
 
         # Just a convenience method to point out that people should use graphql_name instead
@@ -47,8 +47,10 @@ module GraphQL
         def description(new_description = nil)
           if new_description
             @description = new_description
-          else
+          elsif defined?(@description)
             @description
+          else
+            nil
           end
         end
 
@@ -69,8 +71,10 @@ module GraphQL
         def introspection(new_introspection = nil)
           if !new_introspection.nil?
             @introspection = new_introspection
-          else
+          elsif defined?(@introspection)
             @introspection
+          else
+            false
           end
         end
 
@@ -83,8 +87,10 @@ module GraphQL
         def mutation(mutation_class = nil)
           if mutation_class
             @mutation = mutation_class
-          else
+          elsif defined?(@mutation)
             @mutation
+          else
+            nil
           end
         end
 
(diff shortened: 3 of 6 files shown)


casperisfine:
Hum, looks like eagerly defining `@introspection` breaks a lot of stuff. I'm not quite sure why so I reverted it for now.

casperisfine:
Probably wasn't the only problem. There is a lot of lazily initialized state in GraphQL classes, and I'm having a very hard time untangling it :/

rmosolgo:
Hey! Thanks for sharing this work. I'm really interested in playing nicer with Ruby 3.2. 

I think most of the reason behind varyingly-initialized instance variables was to reduce the static memory footprint of a schema. This was important for GitHub Enterprise which runs on customer servers and had embarrasingly high memory requirements. We found a lot of it could be reduced by removing instance variables storing `nil`. So, I tried to avoid assigning instance variables that were unused at runtime. (GraphQL::Field instances were also a big source of this, although I see that one wasn't addressed yet in this PR.) 

Then, all those `defined?` checks were added to avoid issuing warnings about accessing uninitialized instance variables. 

Other late-assigned instance variables are for caching, for example, `@default_graphql_name`: 

https://github.com/rmosolgo/graphql-ruby/blob/016d0c9f8132466223a7837b3f0cc646277284c8/lib/graphql/schema/member/base_dsl_methods.rb#L99-L108

It's generally a function of the class name, and a previous implementation re-generated it for each call. This created a lot of duplicate strings at runtime, so I started caching it. But then, we also needed that error about anonymous classes ... so it's in there too. But you couldn't run that code _during_ `Class.new(GraphQL::Types::Object)` because the application code wouldn't've had a chance to assign a `graphql_name "..."` yet. So refactoring that will probably require a bit more work, making it so that the value is generated and cached during Class setup, but deferring that check for name-less GraphQL types until later. (It could probably go here, where already-defined types are being added to the schema's `{ name => definition }` cache of types: https://github.com/rmosolgo/graphql-ruby/blob/016d0c9f8132466223a7837b3f0cc646277284c8/lib/graphql/schema/addition.rb#L143.)

So anyways, I'd really like to improve performance in this way, but I'm also sensitive to memory footprint. (This gem has `bundle exec  rake bench:profile_schema_memory_footprint` to keep an eye on this, although I haven't run it for quite a while.) I bet approaching those low-hanging fruit instance variables one-at-a-time would make it a bit easier to track down which ones need what kind of special treatment.

casperisfine:
Yeah, it's perfectly understandable, that being said:

For class level properties, eagerly defining them can lead to much better memory performance if the application uses a forking server (puma, unicorn etc), thanks to Copy-on-Write. Deferring that initialization means it will happen in every single worker, both impacting first request performance and increasing real memory usage (PSS).

So for things such as `@default_graphql_name` I think it would make sense to eagerly define them. Of course there's the anonymous class issue, but for the vast majority of cases it shouldn't be an issue.

> We found a lot of it could be reduced by removing instance variables storing nil. So, I tried to avoid assigning instance variables that were unused at runtime.

Hum, this claim surprises me. Unless you mean at the class level? Because IIRC even in 3.1 and older, the IV index table is stored on the class, so as long as one instance defined a variable once, that space will be allocated in instance, and the slot will just have `Qundef` instead of `Qnil`.

On classes properties it's a bit different though.

> embarrasingly high memory requirements

Yeah, we have the same experience at Shopify. The GraphQL static data is constantly at the top of my memory profiles.

rmosolgo:
> so as long as one instance defined a variable once, that space will be allocated in instance ... 

Hmm, thanks for double-checking that. Looking back at old PRs, I think I misremembered it. Maybe the problem was too many empty arrays and hashes assigned to instance variables but then never used. So in some cases, I assigned an empty, frozen collection and made sure that future modifications created a new collection, but in other cases, I made the assignment lazy, so the collection wasn't created (and the instance variable wasn't assigned) until something was added. For example: https://github.com/rmosolgo/graphql-ruby/pull/3153/files#diff-bce2e56fc0d34c317e91188c0d7e812305d3dd5453e07c54815dfb14460a289e

So, I guess I got my wires crossed there -- that makes more sense! Sorry for the confusion about it.

casperisfine:
Ahah no worries.

So yes, the `@foo ||= {}` pattern is fine to avoid eagerly creating objects that may never be used, as long as it's associated with an eager `@foo = nil`, which shouldn't cause any extra memory usage.

That's what I'm trying to do here, but I'm struggling a bit :/

casperisfine:
Closing in favor of smaller more targeted PRs.

PR #4295, opened 2023-01-18
Eagerly define `@default_graphql_name` variables for Shape Friendliness

casperisfine:
Ref: https://github.com/rmosolgo/graphql-ruby/pull/4293

I kinda got lost trying to fix many shapes issues at once, so I'm trying again one variable at a time.

This one the the most common shape edge in our app:

```

Shape Edges Report
       527  @default_graphql_name
```

### DIFF
--- a/lib/graphql/schema/directive.rb
+++ b/lib/graphql/schema/directive.rb
@@ -21,9 +21,9 @@ module GraphQL
         # but downcase the first letter.
         def default_graphql_name
           @default_graphql_name ||= begin
-            camelized_name = super
+            camelized_name = super.dup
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/directive.rb
```diff
@@ -21,9 +21,9 @@ def path
         # but downcase the first letter.
         def default_graphql_name
           @default_graphql_name ||= begin
-            camelized_name = super
+            camelized_name = super.dup
             camelized_name[0] = camelized_name[0].downcase
-            camelized_name
+            -camelized_name
```
casperisfine, line 26: While I was at it, I added code to deduplicate these names.

PR #4297, opened 2023-01-18
Eagerly define @to_non_null_type variables for Shape Friendliness

casperisfine:
Ref: https://github.com/rmosolgo/graphql-ruby/pull/4293

I kinda got lost trying to fix many shapes issues at once, so I'm trying again one variable at a time.

This one the the third most common shape edge in our app:

```
Shape Edges Report
       527  @default_graphql_name
       494  @own_fields
       487  @to_non_null_type
```

### DIFF
--- a/lib/graphql/schema/member/type_system_helpers.rb
+++ b/lib/graphql/schema/member/type_system_helpers.rb
@@ -4,6 +4,13 @@ module GraphQL
   class Schema
     class Member
       module TypeSystemHelpers
+        def initialize(*args, &block)
+          super
+          @to_non_null_type ||= nil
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/member/type_system_helpers.rb
```diff
@@ -4,6 +4,13 @@ module GraphQL
   class Schema
     class Member
       module TypeSystemHelpers
+        def initialize(*args, &block)
```
casperisfine, line 7: This module is a bit weird because it can be either included on extended.

rmosolgo:
Thanks for this improvement!

PR #4306, opened 2023-01-24
Allow `nil` as an enum ruby value

ojab:
Use dummy object instead of `nil` for undefined value, so we can distingush explicitly passed `nil`.

Right now we have:
```ruby
require 'graphql'

class MyEnumType < GraphQL::Schema::Enum
  value 'NO_VALUE', value: nil
  value 'ONE_VALUE', value: 1
end

p MyEnumType.coerce_input('NO_VALUE', GraphQL::Query::NullContext) # => 'NO_VALUE'
```

Not sure how to properly test it because added test is in `legacy tests` block, please advise.

### DIFF
--- a/lib/graphql/schema/enum_value.rb
+++ b/lib/graphql/schema/enum_value.rb
@@ -25,16 +25,19 @@ module GraphQL
       include GraphQL::Schema::Member::HasDirectives
       include GraphQL::Schema::Member::HasDeprecationReason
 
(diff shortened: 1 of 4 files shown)


ojab:
Single CI failure looks unrelated, sending to review.

rmosolgo:
LGTM! Thanks for this improvement.

PR #4315, opened 2023-01-28
Don't run scoping on introspection queries

rmosolgo:
This overhead seems small, but it adds up proportional to the number of lists (arguments, fields, types, enum values, interfaces, possible types) in the result. So, for big schemas, it can add up!

```diff 
-        247   (0.1%)          62   (0.0%)     GraphQL::Schema::Field::ScopeExtension#after_resolve
+
```

(Using the introspection benchmark from 8f0e9e8662edc67844074c2392e1c81a066ac3fe)

TODO: 

- [ ] Port to 2.0.x

### DIFF
--- a/lib/graphql/introspection/directive_type.rb
+++ b/lib/graphql/introspection/directive_type.rb
@@ -11,8 +11,8 @@ module GraphQL
                     "to the executor."
       field :name, String, null: false, method: :graphql_name
       field :description, String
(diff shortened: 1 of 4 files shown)


rmosolgo:
Reviewing a great big introspection benchmark: 


  | Total | Self | Method
-- | -- | -- | --
**Before** | 117.89ms (0.56%) | 31.82ms (0.15%) | GraphQL::Schema::Field::ScopeExtension#after_resolve
**After** | 0 | 0 | (not present)

PR #4316, opened 2023-01-28
Performance: reduce a duplicate call to Field#type

rmosolgo:
When building really big result sets (especially introspection, where there's almost nothing _except_ GraphQL overhead), these redundant `Field#type` calls add up. 

Amazingly, my profiles don't show any callees of this method, so I think they're all calls to the cached `@type` (https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/lib/graphql/schema/field.rb#L602). So it's just the sheer _number_ of calls that creates the observed slowness.

It didn't make a big difference in this library's benchmark, but I think it'll do the job: 

```diff
-       1813   (0.2%)        1813   (0.2%)     GraphQL::Schema::Field#type
+        819   (0.2%)         819   (0.2%)     GraphQL::Schema::Field#type
```

TODO: 

- [x] Eliminate some of these calls (at least one of the ones in `Runtime`)
- [x] Port to 2.0.x

### DIFF
--- a/benchmark/run.rb
+++ b/benchmark/run.rb
@@ -80,7 +80,7 @@ module GraphQLBenchmark
         100.times do |n|
           obj_t = Class.new(GraphQL::Schema::Object) do
             graphql_name("Object#{n}")
-            5.times do |n2|
+            20.times do |n2|
               field :"field#{n2}", String do
                 argument :arg, String
               end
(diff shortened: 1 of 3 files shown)


rmosolgo:
On a great big introspection query, this seems to have worked nicely: 

  &nbsp; | Total | Self | Method
---|---|---|---
**Before** | 762.52ms (3.6%) | 762.52ms (3.6%) | GraphQL::Schema::Field#type
**After** | 106.78ms (0.49%) | 106.78ms (0.49%) | GraphQL::Schema::Field#type

PR #4319, opened 2023-01-30
Introspection performance: don't re-encode schema member names

rmosolgo:
Looking over a profile of an introspection query, I found that duplicate strings were created for the names of the built-in scalar types. I think this was because Ruby class names are `US-ASCII`  by default, but GraphQL requires `UTF-8`. Also, `some_symbol.to_s` creates a `US-ASCII` string by default, too. 

Since ASCII is a subset of UTF-8, allow `ascii_only?` strings to be returned without further encoding. (GraphQL names [must be ASCII-only](https://spec.graphql.org/draft/#note-46408).) 

In my local benchmark: 

```diff 
-       1962   (0.3%)         926   (0.2%)     GraphQL::Types::String.coerce_result
+        517   (0.1%)         383   (0.1%)     GraphQL::Types::String.coerce_result


- Total allocated: 31279956 bytes (284706 objects)
+ Total allocated: 30961980 bytes (276750 objects)

  allocated objects by class
  -----------------------------------
      136104  Array
       93952  Hash
       10487  <<Unknown>>
       10485  GraphQL::Execution::Interpreter::Runtime::GraphQLResultHash
       10485  GraphQL::Execution::Interpreter::Runtime::GraphQLSelectionSet
-       8302  String
        5028  Proc
        ...
+        115  String
```

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -54,7 +54,15 @@ module GraphQL
       # @param replace_null_with_default [Boolean] if `true`, incoming values of `null` will be replaced with the configured `default_value`
       def initialize(arg_name = nil, type_expr = nil, desc = nil, required: true, type: nil, name: nil, loads: nil, description: nil, ast_node: nil, default_value: NO_DEFAULT, as: nil, from_resolver: false, camelize: true, prepare: nil, owner:, validates: nil, directives: nil, deprecation_reason: nil, replace_null_with_default: false, &definition_block)
         arg_name ||= name
-        @name = -(camelize ? Member::BuildType.camelize(arg_name.to_s) : arg_name.to_s)
+        name_s = camelize ? Member::BuildType.camelize(arg_name.to_s) : arg_name.to_s
+        if name_s.encoding != Encoding::UTF_8
+          if name_s.frozen?
+            name_s = name_s.encode(Encoding::UTF_8)
+          else
+            name_s.encode!(Encoding::UTF_8)
+          end
+        end
+        @name = -name_s
         @type_expr = type_expr || type
         @description = desc || description
         @null = required != true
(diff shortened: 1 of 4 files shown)


casperisfine:
@rmosolgo: this makes very little sense from an MRI perspective. Ruby encoding code knows that ASCII is a subset of UTF-8 and concatenating both together incurs no overhead.

So I suspect you might have some code somewhere that blindly calls `.encode(Encoding::UTF_8)`. I'd suggest to trace this allocation to fix the source of that.

rmosolgo:
Thanks for taking a look, @casperisfine!

> blindly calls `.encode(Encoding::UTF_8)`

Yep, if I've understood correctly, I think it's on line 13 of this bit: 

https://github.com/rmosolgo/graphql-ruby/blob/4712a3d4c71ac7a36181a1a5da0d972b634137fe/lib/graphql/types/string.rb#L10-L16

The memory profiler output points to that line: 

```
Allocated String Report
      4018  "String"
      4018  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

      2000  "arg"
      2000  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field0"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field1"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field10"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field11"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field12"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field13"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13

       100  "field14"
       100  /Users/rmosolgo/code/graphql-ruby/lib/graphql/types/string.rb:13
```

Those strings are all `US-ASCII`, for example:

```
$ irb
> require "graphql"
> GraphQL::Types::String.graphql_name.encoding
=> #<Encoding:US-ASCII>
> GraphQL::Types::Relay::BaseConnection.fields.values.first.graphql_name.encoding
=> #<Encoding:US-ASCII>
```

But do you think a better approach would be to update `Types::String` to permit US-ASCII strings too, since it's a subset of UTF-8?

casperisfine:
> But do you think a better approach would be to update `Types::String` to permit US-ASCII strings too, since it's a subset of UTF-8?

I believe so yes. A quick check may be
```ruby
str.ascii_only? || str.encoding == Encoding::UTF_8
```

rmosolgo:
Cool, thanks for the suggestion. I updated the implementation and double-checked the benchmark, looks good to me!

PR #4322, opened 2023-02-02
Support Fiber#transfer-based fiber schedulers

### LINKED ISSUE #4003: Migrate dataloader to use Fiber#transfer for modern scheduler compatibility

rmosolgo:
This adds a mode to Dataloader which works with `Fiber#transfer`-based schedulers (eg, `fiber_scheduler`, `Async::Scheduler`(?) ). It requires `use GraphQL::Dataloader, nonblocking: true, fiber_control_mode: :transfer`

Fixes #4003

TODO: 

- [x] FiberScheduler fails on Ruby 3.2
- [x] Try adding `Async::Scheduler`, too


ioquatix:
For old rubies, resume/yield and transfer cannot be mixed, but going forward, it's acceptable.

I'd suggest simply not supporting Rubies where it can't be mixed, i.e. depend on Async v2+ (Ruby 3.1+)

rmosolgo:
That's good to keep in mind. For now, I'm just trying to figure out how to get the build not to segfault 😩 

I haven't exactly figured out what's making it go wild, but evidently it doesn't like to "`sleep 0.1`" from inside fibers...

ioquatix:
If you want to chat or show me the issue, let me know. I'm happy to take a look.

ioquatix:
What kind of segfault are you seeing?

rmosolgo:
It runs fine locally, but in CI, `yield`ing from inside `Fiber.new(blocking: false) { ... }` crashes for at _least_ `Async::Scheduler` (https://github.com/rmosolgo/graphql-ruby/actions/runs/5730999250/job/15530962687) and `FiberScheduler` (https://github.com/rmosolgo/graphql-ruby/actions/runs/5732062739/job/15534264398), but not for DummyScheduler (plain-Ruby, copied from Ruby's tests). 

I just tried refactoring to pass an object instead of `yield`, but that didn't work either :S Maybe `Fiber.new(blocking: false).resume` isn't _supposed_ to work 🤔 ?

ioquatix:
What version of Ruby was failing? Can you share the logs/errors?

ioquatix:
Backticks and system should be handled identically.

rmosolgo:
How about those failures on GitHub actions? They're Ruby 3.2.2, with the segfault message logged there. I'd be happy to add more logging if there's something that'd be helpful to you!

https://github.com/rmosolgo/graphql-ruby/actions/runs/5730999250/job/15530962687

ioquatix:
Looks like it's dying on https://github.com/ruby/ruby/blob/v3_2_2/process.c#L4849

If you have spare time, is there a chance you can try to make a small repro? Otherwise, I'll take a look later this week.

spec/graphql/dataloader/async_dataloader_spec.rb
```diff
@@ -296,13 +296,39 @@ def do_sleep(duration)
...
+            query(Query)
+            use GraphQL::Dataloader, nonblocking: true, fiber_control_mode: :transfer
+          end
+
+          start = Time.now
+          puts "Starting..."
+          Fiber.set_scheduler(Async::Scheduler.new)
+          pp MySchema.execute("{ v1: value v2: value }").to_h
+          puts "Ended"
+          pp [:duration, Time.now - start]
```
ioquatix, line 329: You may want to add `Fiber.set_scheduler(nil)` to avoid leaking state.

rmosolgo:

https://github.com/rmosolgo/graphql-ruby/blob/3bade9c11a2a08fb9e932c800936e20ceac5a546/spec/graphql/dataloader/async_dataloader_spec.rb#L308-L329

It works fine if I run it on my machine ... but fails on GH Actions :S

ioquatix:
It probably means race condition.

rmosolgo:
Thanks for the hint about `IO_EVENT_SELECTOR` over on the ruby bug tracker (sorry if I opened that in the wrong spot 🤦 )! The tests now _run_ -- they just fail. But that's OK -- I'll take a closer look soon but I think it's just because of the latency in that io-event backend.

ioquatix:
It's fine, whatever gets attention :) I am glad that helped. I'll investigate further on my end.

ioquatix:
The problem seems to be this code:

```ruby
          # These fibers were previously waiting for sources to load data,
          # resume them. (They might wait again, in which case, re-enqueue them.)
          resume(f)
          if f.alive?
            next_fibers << f
          end
        end
```

Specifically, `resume(f)`.

You are resuming a fiber which is waiting in `process_wait`. If you do that, the following code is triggered:

```clang
static
VALUE process_wait_transfer(VALUE _arguments) {
	struct process_wait_arguments *arguments = (struct process_wait_arguments *)_arguments;
	
	VALUE result = IO_Event_Selector_fiber_transfer(arguments->selector->backend.loop, 0, NULL);
	
	if (arguments->waiting->result) {
		return IO_Event_Selector_process_status_wait(arguments->pid);
	} else {
		return Qfalse;
	}
}
```

Because the event loop has not actually waited successfully (on the process finishing), `arguments->waiting->result == 0`, and so we return `Qfalse`. Because of that, CRuby itself dies:

```clang
        VALUE status = rb_process_status_wait(pid, 0); # this is Qfalse.
        struct rb_process_status *data = RTYPEDDATA_DATA(status);
```

If you schedule a fiber, you must not resume it. e.g. `fiber = Fiber.schedule{}` should not call `resume` arbitrarily, unless you have specifically called `Fiber.yield`. I'll see if I can make this into an error, so that it's more obvious what the correct behaviour is and it fails with an exception rather than a segfault.

ioquatix:
By the way, if you want to wait on a `Fiber.schedule` to complete, you can basically do this:

```ruby
completions = Thread::Queue.new

work.each do |job|
  Fiber.schedule do
    completions << job.call
  end
end
```

You can count the number of jobs and completions or use some other mechanism to know when all jobs are completed.

rmosolgo:
Ok, I think that makes sense, thanks... Previously, the code could assume that if the main fiber was running, then the other fiber(s) had terminated or yielded.  But that assumption was wrong (even _then_, and now it's raising an error) because in fact, the other fiber could be waiting on IO and so the main fiber could have been given back control. 

Completion is one possibility for a stopped fiber -- thanks for the suggestion on handling that better. 

The other possibility is an explicit transfer of control (via `yield` or a call to `.transfer` on the parent fiber): 

https://github.com/rmosolgo/graphql-ruby/blob/e48f302a58f0eaf946737e007a47ec590be42c17/lib/graphql/dataloader.rb#L92-L104

Somehow I'll have to distinguish between those to determine whether the fiber should be re-enqueued to be restarted.

rmosolgo:
Working this out in #4625 and #4727 instead.

PR #4323, opened 2023-02-02
1.13.x: don't re-encode schema member names in introspection

rmosolgo:
Backport #4319


rmosolgo:
Oops, I introduced a bug here bug fixed it in 5c2547984

rmosolgo:
In practice, this didn't work out (0.9% more query time). Looking at the call stack, it seems like `.ascii_only?` takes longer than `.encoding == ...`, so I'll reverse these conditions.

  &nbsp; | Total | Self | Method 
  ---|---|---|---
  **Before** | 93.77ms (0.45%) | 47.22ms (0.22%) | GraphQL::Types::String.coerce_result
**After** | 116.31ms (0.54%) | 50.17ms (0.23%) | GraphQL::Types::String.coerce_result

PR #4330, opened 2023-02-06
Try using .then for all lazy handling

rmosolgo:
A long time ago, GraphQL-Ruby added `lazy_resolve(...)` specifically to handle GraphQL-Batch and its use of Promise.rb. `BatchLoader` also took up this API. 

Since then, Ruby added `.then` as a basic API to all objects, where an object yields itself (https://ruby-doc.org/core-3.0.2/Kernel.html#method-i-then). 

In my mind, this opens the possibility that we could greatly simplify `after_lazy`-type handling, using `.then` directly instead and reducing a lot of indirection and overhead. For most objects, it would call through directly. (However, it makes nested calls -- maybe this can be handled using the returned `Enumerator` when `.then` is called without a block?)  For promises, etc, it would return a chained Promise. Then, everything could be cleaned up at the end. 

I think this could remove a lot of overhead from code and runtime cases where lazy values (Promises, etc) are _not_ used. So, I'm exploring it here.

__TODO__ / to figure out: 
- [ ] `Lazy` blocks included some "around-" context (tracing blocks, error handling, context assignments). How can that be re-implemented here? 
    - ~~maybe error handling can be done inside methods where they might be raised instead of in runtime code~~ That's not right, those methods have already returned.
    - maybe context could be stashed with `@pending_paths`
- [ ] Consider compatibility implications of `.resolve_all` changes, will this still work with `@stream`/`@defer`? 
- [ ] Trampolining with enumerators returned from `.then` without a block? Performance is probably not good, but worth a try.

### DIFF
--- a/lib/graphql/dataloader/request.rb
+++ b/lib/graphql/dataloader/request.rb
@@ -14,6 +14,25 @@ module GraphQL
       def load
         @source.load(@key)
       end
+
+      def then(&block)
+        ThenRequest.new do
+          value = load
+          value.then do |v|
+            block.call(v)
+          end
+        end
+      end
+
+      class ThenRequest < Request
+        def initialize(&block)
+          @block = block
+        end
+
+        def load
+          @block.call
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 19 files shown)


rmosolgo:
I can't see how this would work with error handling :'(

PR #4336, opened 2023-02-07
Remove unused accessible? methods

swalkinshaw:
As far as I know, this is entirely unused now and a relic of legacy authorization. The methods weren't called anywhere, and removing them broke no tests.


rmosolgo:
I had no memory of what happened to this ... I found the deprecation here: #1841 and removal here: 20d39e606ebf270aae4657fbc09bb255d89ecc47. LGTM :+1:

Issue #4334, opened 2023-02-07
Unexpected error validating multiple arguments and `rescue_from`.


klaudia-janiec:
**Describe the bug**

I found error when validation for more than one arguments fails and there is custom `rescue_from`.

**Versions**

`graphql` version: 2.0.16

**Steps to reproduce**
```
schema = Class.new(GraphQL::Schema) do
  rescue_from(StandardError) do |exception, _obj, _args, _context, _field|
    raise exception
  end

  query_type = Class.new(GraphQL::Schema::Object) do
    graphql_name 'TestQueryType'

    field :test, Integer, null: false do
      argument :a, Integer, required: true, validates: { numericality: { greater_than_or_equal_to: 1 } }
      argument :b, Integer, required: true, validates: { numericality: { greater_than_or_equal_to: 1 } }
    end

    def test; end
  end

  query(query_type)
  lazy_resolve(Proc, :call)
end

query = '{ test(a: -4, b: -5) }'

schema.execute(query).to_h['errors']
```

**Expected behavior**

```
[22] pry(main)> schema.execute(query).to_h['errors']
=> [{"message"=>"a must be greater than or equal to 1", "locations"=>[{"line"=>1, "column"=>3}], "path"=>["test"]}, {"message"=>"b must be greater than or equal to 1", "locations"=>[{"line"=>1, "column"=>3}], "path"=>["test"]}]
```

**Actual behavior**

```
NoMethodError: undefined method `keyword_arguments' for #<GraphQL::Schema::Validator::ValidationFailedError: a must be greater than or equal to 1>
from /Users/a/.rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/graphql-2.0.16/lib/graphql/schema.rb:758:in `handle_or_reraise'
Caused by GraphQL::Schema::Validator::ValidationFailedError: b must be greater than or equal to 1
from /Users/a/.rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/gems/graphql-2.0.16/lib/graphql/schema/validator.rb:147:in `validate!'
```

**Additional notes**
I think it is because in `Schema#handle_or_reraise` we call `keyword_arguments` on `current_arguments` and after the validation of first argument fails `current_arguments` is set to the error `GraphQL::Schema::Validator::ValidationFailedError` which does not implement `keyword_arguments` method.

Can someone take a look at this issue and let me know if there is something fundamentally wrong with my code? I think this case should work in normal way.
[...]

PR #4341, opened 2023-02-09
Fix handle_or_reraise with arguments validation

klaudia-janiec:
Fixes issue described in https://github.com/rmosolgo/graphql-ruby/issues/4334


klaudia-janiec:
@rmosolgo Could you approve running workflows so that I can check if CI passes? Thanks!

rmosolgo:
Thanks for the fix!

PR #4347, opened 2023-02-16
Assert matching response shapes in fields_will_merge

rmosolgo:
Oops, it seems like one part of `FieldsWillMerge` was missing. Adding it now might _break_ some queries that used to work, so I may have to ensure there's a way to opt out of this change.

From the spec: 

> [SameResponseShape](https://spec.graphql.org/draft/#SameResponseShape())(fieldA, fieldB)
>    - Let typeA be the return type of fieldA.
>    - Let typeB be the return type of fieldB.
>    - If typeA or typeB is Non-Null:
>        - If typeA or typeB is nullable, return false.
>        - Let typeA be the nullable type of typeA.
>        - Let typeB be the nullable type of typeB.
>    - If typeA or typeB is List:
>        - If typeA or typeB is not List, return false.
>        - Let typeA be the item type of typeA.
>        - Let typeB be the item type of typeB.
>        - Repeat from step 3.
>    - If typeA or typeB is Scalar or Enum:
>        - If typeA and typeB are the same type return true, otherwise return false.

https://spec.graphql.org/draft/#SameResponseShape()

In GraphQL-JS: 

https://github.com/graphql/graphql-js/blob/f201681bf806a2c46c4ee8b2533287421327a302/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts#L691-L718

https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/validation/__tests__/OverlappingFieldsCanBeMergedRule-test.ts#L738-L767

TODO: 

- [x] Does this make other tests fail? 
  - One now-removed test was added here: https://github.com/rmosolgo/graphql-ruby/pull/2004 I think it was a test for code that doesn't exist anymore but ... will it cause a bug to make that fail?
- [x] Is this too big of a change to make at this point? 
- [x] What about people who depend on this check not being here? 
  -  Added a compat layer for deprecation and future removal
- [x] Performance implications?
  - I ran `bench:validate` and didn't see any regression. The recursion here would only occur for list/non-null types, but in the average case, the overhead is a few simple attr_reader-type method calls.


lib/graphql/static_validation/rules/fields_will_merge.rb
```diff
@@ -239,13 +232,50 @@ def find_conflict(response_key, field1, field2, mutually_exclusive: false)
...
       end
 
+      def return_types_conflict?(type1, type2)
+        if type1.list?
+          if type2.list?
+            return_types_conflict?(type1.of_type, type2.of_type)
+          else
+            true
+          end
+        elsif type2.list?
```
gmac, line 261: Comparing this with NodeJS... https://github.com/graphql/graphql-js/blob/main/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts#L694-L718. Shouldn't this return `true`?

lib/graphql/static_validation/rules/fields_will_merge.rb
```diff
@@ -239,13 +232,50 @@ def find_conflict(response_key, field1, field2, mutually_exclusive: false)
...
+        elsif type1.non_null?
+          if type2.non_null?
+            return_types_conflict?(type1.of_type, type2.of_type)
+          else
+            true
+          end
+        elsif type2.non_null?
+          true
+        elsif !type1.kind.fields? && !type2.kind.fields?
```
gmac, line 270: Again comparing this with `graphql-js`, I believe this should be an OR...

```js
if (isLeafType(type1) || isLeafType(type2)) {
    return type1 !== type2;
  }
```

So `!type1.kind.fields?` should be equivalent to `isLeafType(type1)`, at which time the compound condition shouldn't need to change.

gmac:
> Adding it now might _break_ some queries that used to work, so I may have to ensure there's a way to opt out of this change.

Unfortunately, yes... this is a breaking change that will start rejecting queries that used to be accepted. Probably the smoothest path to reaching spec would be to make this an opt-in setting on `GraphQL::Schema` that is off by default for now and activates in the next major release.

lib/graphql/static_validation/rules/fields_will_merge.rb
gmac: Revisiting this again today, I've noticed another issue with the `fields?` check in general... Right now, [Union does not belong to `fields`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/type_kinds.rb#L70) which makes the `fields?` check a rather ambiguous slice across both Leaf and Composite distinctions. Technically Union should probably belong to `fields?`, although I imagine that change could have a huge blast radius across community implementations.

My suggestion then would be to introduce a proper `leaf?` distinction (ie: `scalar? || enum?`), and then refactor the core library to revolve around formal `leaf?` and `composite?` distinctions, and allow `fields?` to fade away into the sunset (while remaining present to support existing community usage).

gmac: https://github.com/rmosolgo/graphql-ruby/issues/4351

lib/graphql/static_validation/rules/fields_will_merge.rb
rmosolgo: Yes, thanks -- fixed with a test in 6ea6508c4

lib/graphql/static_validation/rules/fields_will_merge.rb
rmosolgo: Thanks for adding `.leaf?`. I switched this code to use that check instead in 47cc81936. 

I didn't add a specific test for union types, which probably had a bug in the old implementation, because I just recently fixed a different broken validation about that: #5322. 

I think unions will now fall down to the `else false` case below and be handled by other validations.

rmosolgo:
> opt-in setting 

Done in https://github.com/rmosolgo/graphql-ruby/pull/4347/commits/75c164cf0f8f9c553c6421bf40c20ad1030d349c with a handler method for tracking occurrences of this problem.

Issue #4337, opened 2023-02-07
Pagination connection handles loaded and unloaded active record relations differently


yurik256:
**Describe the bug**
We discovered a bug with graphql-ruby, where an incorrect value for `hasNextPage` would be returned when the resolver fetches the data from the database, when the records are loaded.

Initially, we thought that the problem is related to the `ActiveRecord::Base.connected_to` that we are using to connect to the read replica. However, after some investigation, we found that under the hood, `ActiveRecord::Base.connected_to` is loading ActiveRecord relation when returning the result: https://github.com/rails/rails/blob/d18fc329993df5a583ef721330cffb248ef9a213/activerecord/lib/active_record/connection_handling.rb#L355 using `.load`

We were able to reproduce the same bug, by loading records using `.load` in the schema resolver

**Versions**

`graphql` version: `2.0.16`
`rails` (or other framework): `rails 6.1`
other applicable versions (`graphql-batch,` etc)

**GraphQL schema**
Using the integration test schema

**Steps to reproduce**

We created a test which replicates this behaviour https://github.com/rmosolgo/graphql-ruby/commit/bdc66d5967e769daba291b8a4d41272197516bce#diff-2901ee3b79d555a6ec361d354412fe1fa0a1ac5c433e147c6b6215db35896e05R51-R54

* We have 3 `base` records in the database
* Schema resolver is updated to load records https://github.com/rmosolgo/graphql-ruby/compare/master...yurik256:graphql-ruby:reproduce-a-bug?expand=1#diff-17a1a40b12e1f517921f6223660ca10de669b73f36f92afbf2905967c1475981R179
* The first query loads the first 2 records. `hasNextPage` returns `true` as expected
* The second query loads the next 2 records

**Expected behavior**
* Since there are only 3 records, `hasNextPage` should return `false`

**Actual behavior**
* `hasNextPage` returns `true`

**Additional context**
After debugging the graphql ruby library, we found the [following code](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/pagination/active_record_relation_connection.rb#L10-L16), which is likely the culprit. 
This is where we have different behaviour for loaded & unloaded records. 
We would expect both branches of the condition, to return the same result.

yurik256:
@rmosolgo unfortunately, I don't think I have enough context to fix it on my own 😔
I would really appreciate it, if you could take a look at it.

calebpowell:
[...]

Issue #2695, opened 2020-01-23
hasNextPage check causes extremely slow queries to be run


raysuelzer:
After updating to the latest verison, there were some breaking changes to pagination.  One being that the `nodes` property now was limited, and in order to get a `totalCount` the `items` property had to be used. 

https://github.com/rmosolgo/graphql-ruby/issues/2662

More of a problem is that I noticed queries like the following being executed, even when I am not requesting hasNextPage or hasPreviousPage.  

```
(4125.0ms)  SELECT COUNT(*) FROM (SELECT 1 AS one FROM "people" INNER JOIN "campaigns_people" ON "campaigns_people"."person_id" = "people"."id" WHERE "campaigns_people"."campaign_id" = $1 OFFSET $2) subquery_for_count
```
Is there a way to NOT fire this query unless we are asking for hasNextPage?  On large tables or complicated queries it is extremely slow.

raysuelzer:
I tracked down the issue to `relation_count`  I am not entirely sure what it is being used for but doing the following seemed to have no impact on cursors, pagination, or results.  I wrote my own Connections wrapper and modified it to limit the relations count to 10, as an arbitrary number to limit. 

Like I said, it seems to have had no impact on the pageInfo results or the edges / nodes being returned.  

```
class Connections::OptimizedActiveRecordConnection < GraphQL::Pagination::ActiveRecordRelationConnection
  # Overrides default usage and sets a limit when determining the relation
  # count.  This is a bit hacky, but works for now and avoids very long queries from running.
  def relation_count(relation)
    if relation.respond_to?(:unscope)
      relation.unscope(:order).limit(10).count(:all)
    else
      # Rails 3
      relation.limit(10).count
    end
  end
end
```

rmosolgo:
Thanks for the detailed report! I've opened #2697 to address it, let me know what you think.

PR #4349, opened 2023-02-17
Ensure correct hasNextPage Value When Relation is Already Loaded

calebpowell:
This PR removes the overridden `ActiveRecordRelationConnection#relation_larger_than` method. The overridden method was flawed as demonstrated by the test case on this PR and the parent class implementation appears to return the correct result. 

Note that some of the timing related tests are failing for me (but some of them fail on `master` as well?). It may be that my change introduces additional queries since the some of the prior work done in this area was to [address performance](https://github.com/rmosolgo/graphql-ruby/issues/2695). That said, correctness comes first. There is probably a better fix for this issue but I am unfamiliar with the code base 😅. Please let me know if this commit would introduce additional issues. 

See Issue #4337

### DIFF
--- a/spec/integration/rails/graphql/relay/relation_connection_spec.rb
+++ b/spec/integration/rails/graphql/relay/relation_connection_spec.rb
@@ -102,7 +102,7 @@ describe "GraphQL::Relay::RelationConnection" do
 
       log_entries = log.split("\n")
       assert_equal 1, log_entries.size, "It should run 1 sql query"
-      edges_query, has_next_page_query = log_entries
+      edges_query, = log_entries.first
       assert_includes edges_query, "ORDER BY bases.name", "The query for edges _is_ ordered"
     end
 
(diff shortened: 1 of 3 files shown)


spec/integration/rails/graphql/relay/relation_connection_spec.rb
```diff
@@ -93,10 +101,9 @@ def get_last_cursor(result)
       assert_equal(true, conn["pageInfo"]["hasNextPage"])
 
       log_entries = log.split("\n")
-      assert_equal 2, log_entries.size, "It ran 2 sql queries"
+      assert_equal 1, log_entries.size, "It should run 1 sql query"
```
calebpowell, line 104: The change seems to remove the need for the following query:

`SELECT 1 AS one FROM bases WHERE bases.id IN (SELECT bases.id FROM bases WHERE bases.faction_id = ?) LIMIT ? OFFSET ?\e[0m  [[faction_id, 2], [LIMIT, 1], [OFFSET, 2]`

spec/integration/rails/graphql/relay/relation_connection_spec.rb
```diff
@@ -93,10 +101,9 @@ def get_last_cursor(result)
...
       edges_query, has_next_page_query = log_entries
```
rmosolgo, line 105:
```suggestion
      edges_query, = log_entries.first
```

rmosolgo:
Thanks for this improvement! I think this spec was basically added to document the behavior at the time, so an improvement is very welcome :+1:

Issue #4351, opened 2023-02-17
`fields_have_appropriate_selections` does not require selections on union types


gmac:
**Describe the bug**

The `fields_have_appropriate_selections` validation presently allows the following selection:

```graphql
query { myUnion }
```

With `myUnion` being a union type, this query returns the following payload:

```json
{ "data": { "myUnion": {} } }
```

The spec would require a union type to make a selection.

**Cause**

The issue occurs here: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/rules/fields_have_appropriate_selections.rb#L34

The cause of the issue is that the [`fields?` check omits union types](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/type_kinds.rb#L70). To fix this validation, the check should use `composite?` instead of `fields?`.

**Steps to resolution**

Unfortunately this is a breaking change as sources that used to validate will now fail. Need to:

- Fix the relevant check.
- Hook breaking behavior to a schema-level setting for spec-compliance. Should also include https://github.com/rmosolgo/graphql-ruby/issues/4403

PR #4352, opened 2023-02-18
Add `leaf?` distinction to TypeKind

gmac:
### What

In this PR, I'd propose formalizing a proper `leaf?` distinction on TypeKind. 

### Why?

At present, `TypeKind` provides a `fields?` distinction (Object + Interface) that is a bit awkward because it slices across both Composite and Leaf distinctions...

- `fields?` != `Composite`, because it omits Union.
- `!fields?` != `Leaf`, because it _includes_ Union, which is not a Leaf.

**This ambiguity leads to awkward situations:**

- https://github.com/rmosolgo/graphql-ruby/pull/4347 should really be checking `.leaf?` rather than `!fields?`.
- https://github.com/rmosolgo/graphql-ruby/issues/4351 stems from ambiguity with the `fields?` distinction (spec validation should be checking `composite?`).
- As a library author, I currently need [my own helper](https://github.com/gmac/graphql-stitching-ruby/blob/main/lib/graphql/stitching/util.rb#L59-L61) to correctly assess the Leaf distinction.

### DIFF
--- a/lib/graphql/type_kinds.rb
+++ b/lib/graphql/type_kinds.rb
@@ -5,12 +5,13 @@ module GraphQL
     # These objects are singletons, eg `GraphQL::TypeKinds::UNION`, `GraphQL::TypeKinds::SCALAR`.
     class TypeKind
       attr_reader :name, :description
-      def initialize(name, abstract: false, fields: false, wraps: false, input: false, description: nil)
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for the improvement! I agree this makes sense.

PR #4355, opened 2023-02-22
Fix typo: shoulde -> should

adamnbowen:


rmosolgo:
Thanks! (Also, `GraphQL::Dataloader` _does_ work with non-blocking fibers using `yield`/`resume`, and support for `.transfer` is on the way in #4322)

PR #4356, opened 2023-02-22
Add special deprecation_reason handling for faster introspection

### LINKED ISSUE #4350: Remember `Deprecated` directive reference
This is a super minor thing, but in [`HasDeprecationReason`](https://github.com/rmosolgo/graphql-ruby/blob/0545517b38f735551dde9fbfa4501daab8f4a818/lib/graphql/schema/member/has_deprecation_reason.rb#L9), instead of searching for a specific directive type, perhaps `deprecation_reason=` can remember the reference to the directive it created?

rmosolgo:
Fixes #4350

### DIFF
--- a/lib/graphql/schema/member/has_deprecation_reason.rb
+++ b/lib/graphql/schema/member/has_deprecation_reason.rb
@@ -6,8 +6,7 @@ module GraphQL
       module HasDeprecationReason
         # @return [String, nil] Explains why this member was deprecated (if present, this will be marked deprecated in introspection)
         def deprecation_reason
-          dir = self.directives.find { |d| d.is_a?(GraphQL::Schema::Directive::Deprecated) }
-          dir && dir.arguments[:reason] # rubocop:disable Development/ContextIsPassedCop -- definition-related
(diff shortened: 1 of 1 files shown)


lib/graphql/schema/member/has_deprecation_reason.rb
```diff
@@ -16,6 +15,10 @@ def deprecation_reason=(text)
           if text.nil?
             remove_directive(GraphQL::Schema::Directive::Deprecated)
           else
+            if defined?(@deprecation_reason) && @deprecation_reason
+              remove_directive(GraphQL::Schema::Directive::Deprecated)
+            end
+            @deprecation_reason = text
```
ravangen, line 21: Should this line always be run? If `text` is `nil`, we would want to reset `@deprecation_reason` in case of future `deprecated_reason` usage?

rmosolgo: Ahh, good catch! Thanks for taking a look, I pushed a test for this case and a fix in 682e94f75

PR #4359, opened 2023-02-23
Prevent race in ActionCableLink

ball-hayden:
`channel` can still be `null` when the connected event fires. Use `this` instead, which is correctly set.


rmosolgo:
Thanks for the fix!

Issue #4299, opened 2023-01-18
Add a shapes benchmark


rmosolgo:
see https://github.com/rmosolgo/graphql-ruby/pull/4296#issuecomment-1387203377

Also TODO for shapes: 

- Improve initialization of `field.rb`, `warden.rb`, v1.23.19 `has_arguments.rb#327`, `platform_tracing.rb#42`, `has_directives.rb#64`, `has_validators.rb#21`

PR #4360, opened 2023-02-24
Improve consistency of field shapes

rmosolgo:
Initialize `GraphQL::Schema::Field` so that its instance variables are always initialized in the same order. This way Ruby's "Object Shapes" optimization will work for field instances during query execution. (However, this also requires an application's custom field state to use consistent ivar setup, too.)

Part of #4299 

This ended up improving the Schema footprint memory benchmark, though I couldn't tell you why: 

```diff
  $ bundle exec rake bench:profile_schema_memory_footprint

- Total allocated: 4060200 bytes (34279 objects)
+ Total allocated: 4058600 bytes (34279 objects)
- Total retained:  1832888 bytes (6185 objects)
+ Total retained:  1831288 bytes (6185 objects)


  allocated memory by location
  -----------------------------------
  ... 
-     199200  /Users/rmosolgo/code/graphql-ruby/lib/graphql/schema/resolver/has_payload_type.rb:92
+     197600  /Users/rmosolgo/code/graphql-ruby/lib/graphql/schema/resolver/has_payload_type.rb:92

```


lib/graphql/schema/field.rb
```diff
@@ -355,7 +357,7 @@ def initialize(type: nil, name: nil, owner: nil, null: nil, description: :not_gi
       # @return [Boolean, nil]
       # @see GraphQL::Subscriptions::BroadcastAnalyzer
       def broadcastable?
-        if defined?(@broadcastable)
+        if @broadcastable != NOT_CONFIGURED
```
casperisfine, line 360: Probably not worth it, but when using this kind of "unset" tokens, I prefer to inverse the check so that if I hit a weird user defined object with a custom `==` it won't be invoked.

You can also use `equal?` to only check identity:

```suggestion
        if !NOT_CONFIGURED.equal?(@broadcastable)
```

But I agreed that it doesn't read as well.

lib/graphql/schema/field.rb
```diff
@@ -5,6 +5,8 @@
 module GraphQL
   class Schema
     class Field
+      NOT_CONFIGURED = ::Object.new
```
casperisfine, line 8: Assuming you'd use this pattern in multiple classes, it might make sense to have a single `GraphQL::NOT_CONFIGURED`. You can also make it a `private_constant` if you don't want users to be able to pass it as argument.

lib/graphql/schema/field.rb
rmosolgo: :+1: Thanks for the suggestion, I'll update it!

lib/graphql/schema/field.rb
rmosolgo: :+1: I added it here and I've opened https://github.com/rmosolgo/graphql-ruby/issues/4363 to track rolling it out across the codebase.

lib/graphql.rb
```diff
@@ -66,6 +66,9 @@ def self.scan(graphql_string)
   def self.scan_with_ragel(graphql_string)
     GraphQL::Language::Lexer.tokenize(graphql_string)
   end
+
+  NOT_CONFIGURED = Object.new
```
casperisfine, line 70: So this is kind of ROFLscale territory, but I wonder if all these extra references might slow down major GC pause.

One way to avoid this would be to use an immediate object, like a static symbol, e.g. `:__GraphQL__NOT_DEFINED__`. This allows the GC to deretctly skip that reference.

It's not as hack proof, as it's much easier for users to pass that symbol, but I can't imagine it happening by accident.

That said, I'm no GC expert, it might not make a difference at all.

rmosolgo: I'm going to leave it as-is 😅

PR #4365, opened 2023-02-28
Add a test and improve Object type shape

rmosolgo:
Part of #4299 

This adds to the schema memory footprint so far, though: 

```diff 
- Total allocated: 4060872 bytes (34279 objects)
+ Total allocated: 4241232 bytes (34291 objects)
- Total retained:  1833560 bytes (6185 objects)
+ Total retained:  2013120 bytes (6185 objects)
```

I want to see about keeping that down.

### DIFF
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -46,7 +46,7 @@ module GraphQL
           elsif defined?(@description)
             @description
           else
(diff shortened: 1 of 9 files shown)


rmosolgo:
I cut the increase in memory footprint about in half by isolating the Relay-related instance variables (but it adds another possible shape permutation):

```
Total retained:  1926088 bytes (6185 objects)
```

PR #4366, opened 2023-02-28
Port token scanning to C

rmosolgo:
This ports `GraphQL::Language::Lexer` to C. I really just copied the old `lexer.rl` into a new one, and updated it so it worked in C. The initial benchmark makes it look plenty fast, 20x-30x faster:

```
scan c - introspection
                         45.964k (± 2.3%) i/s -    230.986k in   5.028203s
scan - introspection      1.487k (± 3.4%) i/s -      7.540k in   5.076915s
  scan c - fragments     63.817k (± 2.9%) i/s -    321.904k in   5.048363s
    scan - fragments      2.547k (± 2.7%) i/s -     12.850k in   5.048907s
  scan c - big query      7.261k (± 3.0%) i/s -     36.411k in   5.019182s
    scan - big query    253.115  (± 2.8%) i/s -      1.272k in   5.029721s
```

Looking at the parsing benchmarks, here's `master`:

```
parse - introspection
                        999.154  (± 1.8%) i/s -      5.088k in   5.093939s
   parse - fragments      1.669k (± 2.6%) i/s -      8.478k in   5.084602s
   parse - big query    173.118  (± 1.7%) i/s -    880.000  in   5.084998s
```

And then, here's the Ruby parser after migrating the Ruby lexer to use plain Arrays instead of `Token` instances: 

```
parse - introspection
                          1.088k (± 1.8%) i/s -      5.508k in   5.065722s
   parse - fragments      1.749k (± 3.0%) i/s -      8.869k in   5.076106s
   parse - big query    186.735  (± 3.2%) i/s -    936.000  in   5.019278s
```

(A little less than 10% speed-up, not bad.) 

Then, here's the Ruby parser when using the new C lexer: 

```
parse - introspection
                          3.162k (± 7.7%) i/s -     15.792k in   5.035083s
   parse - fragments      4.681k (± 6.7%) i/s -     23.641k in   5.074272s
   parse - big query    547.547  (± 1.8%) i/s -      2.754k in   5.031468s
```

About 3x faster 😎 

It also consumes less memory (this is full parsing, not scanning only):

```diff 
- Total allocated: 429808 bytes (6507 objects)
+ Total allocated: 404520 bytes (5874 objects)
```

TODO: 

- [x] Why is the new lexer failing at block strings that have backslashes? 
  - I hunted lots of reasons, but in the end, updating Ragel seems to have done the trick 😖  I was locked to an old Ragel version because new versions generated slower Ruby code, so I need to make sure that this update didn't make a regression. 
- [x] Use a nicer Ruby namespace for this new module
- [x] Reuse frozen strings for static token values
-  ~~Investigate integrating this with a yacc/C parser~~
  - I will do this in a separate PR
- ~~Add `Language::Token.wrap(...)` for backwards compat?~~
  - I will happily add this if it turns out we need it, but I doubt anyone is using Tokens directly 🤞 
- [x] Make it easy to opt in to the new lexer 
- [x] Make sure all tests pass when using this new lexer
- [x] Test that Ragel 7.0.4 doesn't make the Ruby lexer slower
- [x] Use shared strings for `_LITERAL` tokens too 

Checking the Ruby lexer, I compared `master` vs this branch:
    
```diff
  Calculating -------------------------------------
- scan - introspection      1.462k (± 1.5%) i/s -      7.314k in   5.002571s
+ scan - introspection      1.581k (± 1.6%) i/s -      8.000k in   5.060003s
-     scan - fragments      2.551k (± 1.7%) i/s -     12.800k in   5.018478s
+     scan - fragments      2.759k (± 3.1%) i/s -     13.850k in   5.026054s
-     scan - big query    246.936  (± 4.9%) i/s -      1.250k in   5.074237s
+     scan - big query    270.644  (± 3.3%) i/s -      1.377k in   5.093939s
```


rmosolgo:
I've had a really hard time getting this last test to pass. It turns out this machine: 

```
  QUOTE = '"';
  BLOCK_QUOTE =   '"""';
  ESCAPED_BLOCK_QUOTE = '\\"""';
  BLOCK_STRING_CHAR = (ESCAPED_BLOCK_QUOTE | ^QUOTE | QUOTE{1,2} ^QUOTE);
  BLOCK_STRING = (BLOCK_QUOTE BLOCK_STRING_CHAR* QUOTE{0,2} BLOCK_QUOTE);

  main := |*
    BLOCK_STRING  => {};
  *|;
```

Works on Ruby but not in C 🤷 

I've tried logging transitions and debugging the machine graph with Graphviz, but the thing is, the generated graph is the same for the Ruby lexer as the C lexer. So it must be some semantic of the host language (or bug in the code generator for one or the other) that makes one work and the other not. I'm going to try a newer Ragel version in the hopes that it will Just Work :tm:, or else, report a bug.

rmosolgo:
I updated my test script for the new code organization and it seems to work fine: 

```
require "bundler/inline"

gemfile do
  gem "graphql", github: "rmosolgo/graphql-ruby", ref: "ragel-c-gen"
  gem "graphql-c_parser", github: "rmosolgo/graphql-ruby", ref: "ragel-c-gen", glob: "**/graphql-c_parser.gemspec"
end


pp GraphQL.scan_with_c("{ a1 }")
# [[:LCURLY, 1, 1, "{", nil], ...]
```

So, I think that means the gem-within-a-gem is setup correctly :+1:

PR #4377, opened 2023-03-10
Fix docs on chaining custom tracers

jturkel:
The new tracing API introduced in 2.0.18 requires custom tracers to call `super` rather than `yield` to ensure that multiple tracers can be chained together. This PR updates the docs to reflect that. It also updates the doc for the `ActiveSupport::Notifications` tracer to use the new APIs.


rmosolgo:
Thanks for updating this!

PR #4379, opened 2023-03-10
Fix tracers on parent schema classes

### LINKED ISSUE #4378: Tracer inheritance no longer working in 2.0.18
**Describe the bug**

As of 2.0.18, tracers are no longer inherited down the `GraphQL::Schema` class hierarchy. 

**Versions**

`graphql` version: 2.0.18
`rails` (or other framework): n/a
other applicable versions (`graphql-batch`, etc): n/a

**GraphQL schema**

```ruby
module CustomTracer
  extend self

  def trace(key, data)
    puts "Custom Tracing!!!"
    yield
  end
end

class TestQueryType < GraphQL::Schema::Object
  field :ping, String, null: false

  def ping
    'ok'
  end
end

class BaseSchema < GraphQL::Schema
  tracer CustomTracer
end

class TestSchema < BaseSchema
  query TestQueryType
end
```
**GraphQL query**

```graphql
{
  ping
}
```

**Steps to reproduce**

The following script reproduces the problem:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'graphql', '2.0.18' # works in 2.0.17
end

module CustomTracer
  extend self

  def trace(key, data)
    puts "Custom Tracing!!!"
    yield
  end
end

class TestQueryType < GraphQL::Schema::Object
  field :ping, String, null: false

  def ping
    'ok'
  end
end

class BaseSchema < GraphQL::Schema
  tracer CustomTracer
end

class TestSchema < BaseSchema
  query TestQueryType
end

TestSchema.execute('{ ping }')
```

The problem can also be reproduced with the new tracing API:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'graphql', '2.0.18'
end

module CustomTrace
  def execute_multiplex(...)
    puts "Custom Tracing!!!"
    super
  end
end

class TestQueryType < GraphQL::Schema::Object
  field :ping, String, null: false

  def ping
    'ok'
  end
end

class BaseSchema < GraphQL::Schema
  trace_with CustomTrace
end

class TestSchema < BaseSchema
  query TestQueryType
end

TestSchema.execute('{ ping }')
```

**Expected behavior**

The above test case should output several lines that say `Custom Tracing!!!`.

**Actual behavior**

The above test case doesn't output anything because the customer tracer is never invoked.

rmosolgo:
Fixes #4378

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -147,7 +147,12 @@ module GraphQL
         if new_class
           @trace_class = new_class
         elsif !defined?(@trace_class)
-          @trace_class = Class.new(GraphQL::Tracing::Trace)
+          parent_trace_class = if superclass.respond_to?(:trace_class)
+            superclass.trace_class
+          else
+            GraphQL::Tracing::Trace
+          end
+          @trace_class = Class.new(parent_trace_class)
         end
         @trace_class
       end
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -960,14 +960,17 @@ module GraphQL
       # @param options [Hash] Keywords that will be passed to the tracing class during `#initialize`
       # @return [void]
       def trace_with(trace_mod, **options)
-        @trace_options ||= {}
-        @trace_options.merge!(options)
+        trace_options.merge!(options)
         trace_class.include(trace_mod)
+      end
+
+      def trace_options
+        @trace_options ||= superclass.respond_to?(:trace_options) ? superclass.trace_options.dup : {}
       end
 
       def new_trace(**options)
         if defined?(@trace_options)
-          options = @trace_options.merge(options)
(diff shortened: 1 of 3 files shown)


jturkel (commented):
Thanks for the quick fix! Should we handle inheritance for [`trace_options`](https://github.com/rmosolgo/graphql-ruby/blob/319a1fd6016120ca548a39c34f3d66a84464b007/lib/graphql/schema.rb#L958) too?

rmosolgo:
:+1: Yeah, that definitely makes sense. What do you think of [1df8f2f](https://github.com/rmosolgo/graphql-ruby/pull/4379/commits/1df8f2f878e5e92e04c70d23aec8408687761389) ?

jturkel (approved):
Looks great. Thanks again for fixing this!

PR #4394, opened 2023-03-21
Faster `GraphQL::Language::Printer` with `truncate_size` option

felix-d:
Rewrites the `GraphQL::Language::Printer` to make it more efficient and allow consumers to pass a `truncate_size` option.

Some benchmarks we ran with large queries indicated a ~2x speedup compared to the original implementation.

image

Even with the optimized implementation, printing large queries can fairly expensive. The `truncate_size:` option can be used by consumers that don't necessarily need to print the whole document to stop printing when the `truncate_size` is reached.

### DIFF
--- a/lib/graphql/language/printer.rb
+++ b/lib/graphql/language/printer.rb
@@ -2,6 +2,32 @@
 module GraphQL
   module Language
     class Printer
+      OMISSION = "... (truncated)"
+
+      class TruncatableBuffer
+        class TruncateSizeReached < StandardError; end
+
+        DEFAULT_INIT_CAPACITY = 500
+
+        def initialize(truncate_size: nil)
+          @out = String.new(capacity: truncate_size || DEFAULT_INIT_CAPACITY)
+          @truncate_size = truncate_size
+        end
+
+        def <<(other)
+          if @truncate_size && (@out.size + other.size) > @truncate_size
+            @out << other.slice(0, @truncate_size - @out.size)
+            raise(TruncateSizeReached, "Truncate size reached")
+          else
+            @out << other
+          end
+        end
+
+        def to_string
+          @out
+        end
+      end
+
       # Turn an arbitrary AST node back into a string.
       #
       # @example Turning a document into a query string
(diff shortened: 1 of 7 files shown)


felix-d:
I just need to add some tests for `truncate_size`.

lib/graphql/language/printer.rb
```diff
@@ -21,294 +47,344 @@ class Printer
...
+        print_node(variable_definition.type)
+        unless variable_definition.default_value.nil?
+          @out << " = "
+          print_node(variable_definition.default_value)
+        end
       end
 
       def print_variable_identifier(variable_identifier)
-        "$#{variable_identifier.name}".dup
+        @out << "$#{variable_identifier.name}"
```
gmac, line 190: Why not go all in...?

```suggestion
        @out << "$" << variable_identifier.name
```

lib/graphql/language/printer.rb
```diff
@@ -21,294 +47,344 @@ class Printer
...
         end
-        out
       end
 
       def print_scalar_type_definition(scalar_type)
-        out = print_description(scalar_type)
-        out << "scalar #{scalar_type.name}"
-        out << print_directives(scalar_type.directives)
+        print_description(scalar_type)
+        @out << "scalar #{scalar_type.name}"
```
gmac, line 231:
```suggestion
        @out << "scalar " << scalar_type.name
```

lib/graphql/language/printer.rb
felix-d: There was no reason other than piggy backing off of the existing code. I assume most of the performance gains came from not duplicating strings unnecessarily. Benchmarking the two it does not seem to make a big difference. Interpolations seem to be even faster if there are multiple interpolated variables in the same string. I can make the change though I don't have strong opinions there.

felix-d: I ran a benchmark again with the suggestion with all the interpolations removed and it seems like it's actually a bit slower. I suspect because of the method call indirection `TruncatableBuffer#<<`.

I renamed `#<<` to `#append` to avoid confusion and kept the interpolation.

felix-d:
I just need to fix those test 😅

rmosolgo:
Thanks for sharing this improvement!

PR #4399, opened 2023-03-23
Use a dedicated object for runtime state

rmosolgo:
Setting the current runtime state is a really hot path. This simplifies it so that, rather that creating a hash to store in `Thread.current`, we store the `Runtime` instance itself, using instance variables for the relevant bits of state. I think instance variables will be faster than hash keys. 

```diff
  $ be rake bench:profile_large_result
  # ...
  Calculating -------------------------------------
  Querying for 1000 objects
-                           3.506  (± 0.0%) i/s -     18.000  in   5.198419s
+                           4.095  (± 0.0%) i/s -     21.000  in   5.129366s
```


rmosolgo:
Ok ... I think the problem is, a single Runtime can be running in multiple fibers "at once" (concurrently, not in parallel), and then leftover state in instance variables leaks between fibers. Phew!

PR #4401, opened 2023-03-23
Use a PORO for field extension state instead of a Hash

rmosolgo:
~~When you have a field extension that _doesn't_ produce a `memo` or modify `object` or `arguments` and doesn't add any extras, this will remove the allocated Hash from that case.~~ That's not possible, but I think using a PORO will still improve performance.


rmosolgo:
Ahh -- I forgot about the case where `def resolve` returns _without_ yielding. In that case, GraphQL-Ruby has no chance to "pick up" the extended arguments at the end of the extension chain. That's why it has this Hash in the first place, it's a place to store this state _while the chain is running_, even if it ends early. We have to have some kind of out-of-bounds state object here. Maybe it could be a PORO instead, but the win there is minimal at best.

PR #4410, opened 2023-03-28
Add RubyMemcheck to CParser

rmosolgo:
This was originally part of #4368 but it got complicated so I'm going to address it separately here.


rmosolgo:
I'm not sure how to address this so I asked for help: https://stackoverflow.com/questions/75996973/how-to-free-yyptr-stack-in-bison

rmosolgo:
Ok, I have a hunch, but I'm not a C programmer, so please don't laugh if I'm wrong. 

In short, I think Bison is allocating a big block of memory called `yyptr` and referencing part of it as `yyss` (the state stack, full of integers) and `yyvs` (the value stack, full of `VALUE`s in my case). `yyss` is at the start of this block of memory; `yyvs` comes afterward. Before Bison returns, it frees `yyss` -- and this implicitly frees `yyvs` _too_. 

Does that sound possible? 

-------

My notes:

- Bison has a union that holds _either_ a stack state (integer) or a stack value (Ruby `VALUE`), called `yyalloc`: 

  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L660-L665
  
- Bison creates a stack-allocated stack first and allocates a bigger one on the heap if needed. (I learned this from: https://lists.gnu.org/archive/html/help-bison/2021-01/msg00021.html)

  Here's the stack-allocated stack (actually, two stacks in this case; one for states and another for values): 
  
  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L1679-L1687
  
  But later, when if it finds that it needs more room in the stack, it allocates a new block of memory (`yyptr`) which is an array `yyalloc` "type" (that is, each entry has room for either a state or a value). 
  
  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L1764-L1775
  
- The weird thing is, it copies _both_ the state stack(`yyss`) _and_ the value stack (`yyvs`) into that new memory: 

  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L1778-L1779

  `YYSTACK_RELOCATE` includes a direct reference to `yyptr`:

  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L678-L692
  
   That macro copies one of the stacks into `yyptr`, then moves up `yyptr` beyond the point of that stack, including `YYSTACK_GAP_MAXIMUM`: 
   
     https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L689-L690
     
     https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L667-L668
     
     So, it's a single allocated bunch of memory, but accessed via these two different ways. 
     
- During `yyreturn`, Bison frees the state stack: 

  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L3223-L3224
  
  But in cases when Bison has made a heap-allocated stack, that pointer actually _includes_ `yyvs` too, because of this: 
  
  https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L688
  
  (That expands to `yyss = &yyptr->yyss_alloc`). 
  
  I think this means that the memory used by `yyvs` is _also freed_ during this call, since that statck is at the end of `yyss`.

larry-reid-shopify:
`graphql-c_parser/ext/graphql_c_parser_ext/parser.c` is the output from your BIson parser, right? (The `.y` file output.)

larry-reid-shopify:
Your analysis seems sound. I think [this line](https://github.com/rmosolgo/graphql-ruby/blob/3fd6c53251728b73c95aef352e5337a403cb9af8/graphql-c_parser/ext/graphql_c_parser_ext/parser.c#L1767) says we're going to allocate room for two big stacks in one double-big chunk. Then there is some gymnastics to move the two stacks each to its own half of the big chunk That's why a single free can release both stacks. It just has to free the one chunk.

I would be worried that Ruby garbage collection might not play nicely with a relocation of those stacks. You said that one of those stacks holds Ruby values. I haven't kept up with how the Ruby GC works these days, so I won't venture to speculate on a specific scenario. I think about this over the weekend.

Sorry I wasn't more help so far.

rmosolgo:
👋  Thanks for taking a look. Yes, `parser.c` is the generated output from Bison. 

I _bet_ we're ok for GC because: 

- Ruby GC checks the machine stack for an Ruby objects, marking them and following their references; 
- And, my handlers all put Ruby values on the stack (`$$ = ...`), with any _other_ values being reachable from there. 

That is, the accumulated data is a tree (an AST, at that!) and the parent-most node of that tree is always on the stack (by Bison, either as `$$` or `$1, $2, ...` and so on). 

(I _did_ have some GC issues when I was first developing this, with my `static ...` C variables, but I learned to use `rb_global_variable(...)` to tell Ruby about them.)

rmosolgo:
It didn't find anything, and it's hard to get the build to pass, and it runs really slow... I'm done working on this. If anyone is interested in getting this build working, please feel free to use this PR as a starting point but I'm done for a while!

PR #4412, opened 2023-03-29
Fix returning [Type, false] from resolve_type

swalkinshaw:
https://github.com/rmosolgo/graphql-ruby/pull/4130 fixed returning `[Type, nil]` from a union's `resolve_type`, but it caused a regression for literal `false` values.

`Schema#resolve_type` had a `if resolved_value` conditional would be false for `false` values causing it _always_ return a single value (the `resolve_type`).

The runtime's `resolve_type` method would then _always_ return the tuple version (in non-lazy cases): `[type, value]`. However, since `Schema#resolve_type` was only returning a single value, the resolved value would be lost resulting in `[type, nil]` being passed on.

The fix is to remove the `if resolved_value` conditional entirely and always return the two-value tuple ensuring the original resolved value is correctly passed along.

Note: the runtime still has a branch to check if the `resolve_type` value is a tuple or not. The only case where this won't be true is for lazy resolved types.

The call stack is a bit confusing, so here's all the relevant code in call order:

1. `Schema#resolve_type`
https://github.com/rmosolgo/graphql-ruby/blob/42b341383eaeaa8ec8b927d2d5fb30d13acb746b/lib/graphql/schema.rb#L802-L806

2. `Runtime#resolve_type`
https://github.com/rmosolgo/graphql-ruby/blob/42b341383eaeaa8ec8b927d2d5fb30d13acb746b/lib/graphql/execution/interpreter/runtime.rb#L992

3. `Runtime#continue_field`
https://github.com/rmosolgo/graphql-ruby/blob/42b341383eaeaa8ec8b927d2d5fb30d13acb746b/lib/graphql/execution/interpreter/runtime.rb#L725-L730

cc @IDolgirev


rmosolgo (approved):
Wow -- thanks for the fix and test on this!

swalkinshaw:
Thanks for the backport!

PR #4416, opened 2023-03-31
Add more docs about security

QuentinN42:
I work as a DevSecOps at Escape, where we operate a tool for testing the security of GraphQL. Securing web applications is crucial, and our objective is to simplify the process of adopting security best practices.

To this end, we have enhanced the documentation of this GraphQL engine with our knowledge on GraphQL security measures.

I've created a new section named `Security` with 2 main focuses :

- Some Graphql specific threats like query batching or depth limit.
- Some more general HTTP security best practices like enforcing HTTPS or rate limiting.

I've also added some links to articles, reference or this docs when it's relevant.

Thanks and have a good day !

### DIFF
--- a/guides/guides.html
+++ b/guides/guides.html
@@ -19,6 +19,7 @@ sections:
   - name: GraphQL Enterprise - Object Cache
   - name: GraphQL Enterprise - Changesets
   - name: JavaScript Client
+  - name: Security
   - name: Language Tools
   - name: Testing
   - name: Other
--- a/guides/security/Graphql.md
+++ b/guides/security/Graphql.md
@@ -1,0 +1,84 @@
+---
+layout: guide
+doc_stub: false
+search: true
+section: Security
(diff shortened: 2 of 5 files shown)


ojab:
Spamming opensource projects with advertisement PRs is not nice.

iCarossio:
> Spamming opensource projects with advertisement PRs is not nice.

Hey @ojab, thank you for your comment. It feels like you’re bothered by this MR. 
The goal is to bring value to the community here, how can we do that together?

rmosolgo:
Thanks for the suggestion here! I'm not going to include these because they're either too broad, or, for GraphQL-specific topics, they're already covered in the existing docs (eg https://graphql-ruby.org/queries/complexity_and_depth.html#prevent-deeply-nested-queries, https://graphql-ruby.org/limiters/overview.html).

PR #4418, opened 2023-04-03
Add support for histograms to prometheus_tracing

b1zzu:
The prometheus_exporter supports histograms if enabled, but the GraphQLCollector collector always uses summary.

See: https://github.com/discourse/prometheus_exporter#histogram-mode

Similar to the [WebCollector](https://github.com/discourse/prometheus_exporter/blob/main/lib/prometheus_exporter/server/web_collector.rb#L36) this PR  use the `PrometheusExporter::Metric::Base.default_aggregation` instead of the hardcoded `Summary` metrics


rmosolgo:
Thanks for this improvement!

PR #4431, opened 2023-04-08
Method-based lazy resolution detection

rmosolgo:
###  Background 

GraphQL-Ruby handles _some_ objects specially. For example, when a field returns a `Promise`, GraphQL-Ruby sets it aside and continues running other fields until all that's left is `Promise`s, then it resolves those and continues again. Classes are registered for this special handling via `lazy_resolve` (for example, GraphQL-Batch: https://github.com/Shopify/graphql-batch/blob/57e7150695dc1b67381923965b366f64c25edb3d/lib/graphql/batch.rb#L30) 

Then, at runtime, GraphQL-Ruby checks each returned value to see if it needs lazy resolution. This is done by checking the object's class and looking up a registered method for that class (or a superclass). In my benchmark, this operation turns out to be the biggest time consumer: 

```
# before
Querying for 1000 objects
                          4.508  (± 0.0%) i/s -     23.000  in   5.103359s
==================================
  Mode: wall(1)
  Samples: 797053 (73.24% miss rate)
  GC: 7179 (0.90%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
     77852   (9.8%)       77852   (9.8%)     Kernel#class
     73140   (9.2%)       73140   (9.2%)     Kernel#hash
     34353   (4.3%)       34353   (4.3%)     GraphQL::Execution::Interpreter::Runtime#dead_result?
     32428   (4.1%)       32428   (4.1%)     GraphQL::Schema.lazy_methods
     31782   (4.0%)       31782   (4.0%)     Thread.current
```

Usually, this lookup is a waste because the returned value is a scalar (String, number, etc), not a Promise (or other lazy object). 

So, how can it be either: 

- Eliminated in the most common case; 
- Or, sped up as to not be a problem

? 

### This idea 

Instead of maintaining a custom, class-based registry of methods, what if GraphQL-Ruby let _Ruby_ do that? That is, it could use plain ol' Ruby methods instead. For example, every lazy object must provide a `object.graphql_lazy_resolve_method` which GraphQL-Ruby would use to load its value later. 

The downside is that _every_ object must implement that method, even if it returns `nil`. This ends up "polluting" the method namespace of every object -- yuck! (Refinements could theoretically solve this problem, but unfortunately they're slow.) 

Locally, this approach provides a 6.5% speedup (Although digging in, I think it could  be  another percentage more, there are other codepaths to update to this approach :eyes:) : 

```
# after 
Querying for 1000 objects
                          4.802  (± 0.0%) i/s -     25.000  in   5.210133s
==================================
  Mode: wall(1)
  Samples: 665004 (73.03% miss rate)
  GC: 6710 (1.01%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
     44396   (6.7%)       44396   (6.7%)     Object#graphql_lazy?
     39600   (6.0%)       39600   (6.0%)     Kernel#hash
     27405   (4.1%)       27405   (4.1%)     GraphQL::Execution::Interpreter::Runtime#dead_result?
     26968   (4.1%)       26968   (4.1%)     Thread.current
```

So, what do you think of this idea? Or, can you think of other options for improving this system that I might pursue?

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -1037,9 +1037,7 @@ module GraphQL
         end
 
         def lazy?(object)
-          @lazy_cache.fetch(object.class) {
-            @lazy_cache[object.class] = @schema.lazy?(object)
-          }
+          object.graphql_lazy?
         end
       end
     end
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -5,6 +5,12 @@ require "set"
 require "singleton"
 require "forwardable"
 
+class Object
+  def graphql_lazy?
+    false
+  end
+end
+
 module GraphQL
   # forwards-compat for argument handling
   module Ruby2Keywords
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -887,7 +887,31 @@ module GraphQL
         ctx.errors.push(parse_err)
       end
 
+      module GraphQLLazyTrue
+        def graphql_lazy?
+          true
+        end
+        def self.graphql_lazy_resolve(method_name)
+          Module.new do
+            module_eval <<-RUBY
+              def graphql_lazy_resolve
+                #{method_name}
+              end
+
+              def graphql_lazy?
+                true
+              end
+
+              def self.included(child_cls)
+                child_cls.const_set(:GraphQLLazyResolve, self)
+              end
+            RUBY
+          end
+        end
+      end
+
       def lazy_resolve(lazy_class, value_method)
+        lazy_class.include(GraphQLLazyTrue.graphql_lazy_resolve(value_method))
         lazy_methods.set(lazy_class, value_method)
       end
 
@@ -1070,13 +1094,13 @@ module GraphQL
       # @return [Object] A GraphQL-ready (non-lazy) object
       # @api private
       def sync_lazy(value)
-        lazy_method = lazy_method_name(value)
-        if lazy_method
-          synced_value = value.public_send(lazy_method)
-          sync_lazy(synced_value)
+        if value.graphql_lazy?
+          value = value.graphql_lazy_resolve
+          sync_lazy(value)
         else
           value
         end
+        # TODO clean up old lazy method handling
       end
 
       # @return [Symbol, nil] The method name to lazily resolve `obj`, or nil if `obj`'s class wasn't registered with {#lazy_resolve}.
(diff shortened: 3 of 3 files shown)


lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -1037,9 +1037,7 @@ def authorized_new(type, value, context)
         end
 
         def lazy?(object)
-          @lazy_cache.fetch(object.class) {
-            @lazy_cache[object.class] = @schema.lazy?(object)
-          }
+          object.graphql_lazy?
```
rafaelfranca, line 1040: Is there any overhead on calling `object.respond_to?(:graphql_lazy?) && object.graphql_lazy?` over doing the monkey patch in all objects?

rmosolgo: I _assume_ so, since I often see `Kernel#respond_to?` near the top of Stackprof output (for example: https://github.com/rmosolgo/graphql-ruby/pull/4436). On master, `Kernel#respond_to?` is about 1% of total execution time.

rmosolgo: To confirm, I tried adding `respond_to?` in bc7e7be6d and ran `bundle exec bench:profile_large_result`, and it gave:

```
Querying for 1000 objects
                          4.457  (± 0.0%) i/s -     23.000  in   5.184884s
==================================
  Mode: wall(1)
  Samples: 674069 (72.81% miss rate)
  GC: 5712 (0.85%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
     68566  (10.2%)       68566  (10.2%)     Kernel#respond_to?
     43020   (6.4%)       43020   (6.4%)     Kernel#hash
     28560   (4.2%)       28560   (4.2%)     GraphQL::Execution::Interpreter::Runtime#dead_result?
     26825   (4.0%)       26825   (4.0%)     Thread.current
     22224   (3.3%)       22224   (3.3%)     GraphQL::Schema.lazy_methods
     22211   (3.3%)       22211   (3.3%)     NilClass#===
```

casperisfine: Unfortunately `respond_to?` doesn't benefit from method caches, so yes, this is faster. Not sure it's a great idea to add a method to `Object` though (and returning a BasicObject may break it).

Another thing you could try, is to set that hash to compare by identity https://docs.ruby-lang.org/en/master/Hash.html#method-i-compare_by_identity

That will save the call to `#hash`. With that you may have similarish performance, especially since this `#graphql_lazy?` is likely to be very polymorphic, and the interpreter can only cache one class.

rmosolgo: Trying it out over here: https://github.com/rmosolgo/graphql-ruby/pull/4443

It did save a lot of time in `Kernel#hash` (and could save _more_ as mentioned in the TODOs on that PR), but didn't save any time from the `Kernel#class` time.

casperisfine: > but didn't save any time from the `Kernel#class` time.

Yes as expected. But the hope is that the difference is now small enough that it doesn't really justify a core extension into Object.

rmosolgo: I _did_ get a ~10% speedup on that PR by using `compare_by_identity` and re-working some `lazy?` checks to use that cache instead of the less-optimized schema-level cache. This would probably _still_ give another big improvement, since `Kernel#class` is now 10% of stackprof time on my microbenchmark, but it certainly worked out well over there.

rmosolgo:
This is too icky to pursue further right now.

PR #4432, opened 2023-04-11
Improve Datadog field operation and resource names

sj26:
I'm doing some work on our graphql api and am having trouble with field tracing spans in datadog. I need to query all field spans and then filter down. But there's no common operation or tag at the moment to do so.

I believe it's because they're being tagged using a resource-specific name instead of a general operation name.

Here's an example for how field spans look today:

example-1

and here's a postgres query span, for comparison:

example-2

I've thrown together a patch here based on a patch I've made in our own codebase. It demonstrates that the span can have separate operation and resource names which allow grouping the field operations then filtering into specific resources. But this feels a bit messy amongst the tracing setup. I'd love advice on how to do this nicely, or if it's something you've thought about.

### DIFF
--- a/lib/graphql/tracing/data_dog_tracing.rb
+++ b/lib/graphql/tracing/data_dog_tracing.rb
@@ -15,6 +15,13 @@ module GraphQL
       }
 
       def platform_trace(platform_key, key, data)
+        # When tracing a field, use a common span name and put the type and
+        # field name (from platform_field_key) as the resource
+        if key == 'execute_field' || key == 'execute_field_lazy'
+          field_name = platform_key
+          platform_key = 'field.graphql'
+        end
+
         tracer.trace(platform_key, service: service_name) do |span|
           span.span_type = 'custom'
           if defined?(Datadog::Tracing::Metadata::Ext) # Introduced in ddtrace 1.0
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hey, thanks for sharing this suggestion. I'm open to modifying this tracer, but what how will it affect current users when they update the gem? Will there be a break in their existing DataDog stats?

sj26:
This may be a breaking change if users are relying on the operation names reflecting the type. They would need to change queries or metrics to use the resource name instead.

I could move this change behind a flag to the datadog tracer, to ease migration?

rmosolgo:
Hey, sorry I didn't get back to your suggestion sooner. Since then, I've deprecated the `*Tracing` classes in favor of `*Trace` modules (which have less overhead at runtime). The new DataDog trace is here: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/data_dog_trace.rb

If you're interested in exploring this change there, please open a new PR on that file in a way that keeps the existing default intact. We could also ping `@tonyCTHsu` who recently updated that file to support the new dd-trace-rb library -- he might have some insight into pros and cons of the approach!

PR #4442, opened 2023-04-18
Make a way to bypass type visibility

rmosolgo:
For schemas that don't use type visibility at all, I bet we can reduce the overhead. 

TODO: 

- Can this be detected automatically? 
  - I can't figure out how
- [x] Or, make a nice API for opting in/out of this.

To opt out of graphql-ruby's visibility system, add 

```
use GraphQL::Schema::AlwaysVisible # opt out of GraphQL-Ruby's visibility feature 
```

to your schema class.


benhutton:
This or something like it is very interesting to us. We are seeing a fair bit of overhead in visibility checking when dealing with a large/complex/deep result. We'd love more flexibility in this area, to turn things on and off either globally or per-type.

PR #4445, opened 2023-04-19
Fix runtime state when fields run nested queries

rmosolgo:
Oops, if a field ran another graphql query inside its resolver body, then the inner query would clobber the thread context for the outer field. This change makes it so that the outer field will still have its runtime context after the inner query is finished. 

The performance impact looks negligible: 

```diff
  Calculating -------------------------------------
  Querying for 1000 objects
-                           4.562  (± 0.0%) i/s -     23.000  in   5.043385s
+                           4.447  (± 0.0%) i/s -     23.000  in   5.174162s
```

Plus two more objects allocated 🤷


swalkinshaw:
This fixes the regression for our use case 🎉

PR #4460, opened 2023-05-04
add `execute_multiplex.graphql` to NotificationsTrace

zvkemp:
Legacy tracing fired `execute_multiplex.graphql` to ActiveSupport::Notifications; this PR restores that behavior for the new `ActiveSupportNotificationsTrace` mixin, and adds a few missing requires (to bring it into parity with the rest of the new Trace modules).


rmosolgo:
Thanks for the fix here!

PR #4463, opened 2023-05-05
Accept optional channelName for ActionCable handler and subscriptions

### LINKED ISSUE #4459: Configurable ActionCable channel name in `graphql-ruby-client`

I use an inflection in my Rails app for GraphQL (`inflect.acronym "GraphQL"`). `graphql-ruby-client` contains [hardcoded channel names for the GraphQL channel](https://github.com/rmosolgo/graphql-ruby/blob/27762aa781de5c98039098ca5eaf6bb3fa86702f/javascript_client/src/subscriptions/createActionCableHandler.ts#L24). Rails [converts this value to a constant](https://github.com/rails/rails/blob/8049d7983f3f8fa9679e5ccfb21814554cd60297/actioncable/lib/action_cable/connection/subscriptions.rb#L36) to find the correct channel. My channel class is `GraphQLChannel` and not `GraphqlChannel`.

I think configurable channel names would be fantastic. It would be nice if this could be passed in with the `options` object.

1. Vendoring the relevant portions of `graphql-ruby-client` into my frontend so I can change the channel name.
2. Removing the GraphQL inflection.

I would be more than happy to write this feature if you think it would be accepted.

ajswis:
`ActionCableSubscriber` and `createActionCableHandler` currently do not accept a configurable `channelName` for the cable. This is an effort to support passing `channelName` down to ActionCable.

Fixes #4459

### DIFF
--- a/javascript_client/src/subscriptions/ActionCableSubscriber.ts
+++ b/javascript_client/src/subscriptions/ActionCableSubscriber.ts
@@ -11,10 +11,12 @@ interface ApolloNetworkInterface {
 class ActionCableSubscriber {
   _cable: Consumer
   _networkInterface: ApolloNetworkInterface
+  _channelName: string
 
-  constructor(cable: Consumer, networkInterface: ApolloNetworkInterface) {
+  constructor(cable: Consumer, networkInterface: ApolloNetworkInterface, channelName?: string) {
     this._cable = cable
     this._networkInterface = networkInterface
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for this improvement!

rmosolgo:
Just released in graphql-ruby-client v1.11.8, thanks again :+1:

ajswis:
Awesome! Thanks for the fast release!

PR #4369, opened 2023-03-03
Replace Ragel tokenizer with String Scanner

tenderlove:
This commit replaces the Ragel tokenizer with a "hand written" string scanner based tokenizer.  This tokenizer is about ~2x faster than the previous tokenizer.

To test, I used this benchmark:

  https://gist.github.com/tenderlove/fdce1610a3eecd8426f5845391a6c373

On this branch:

```
$ ruby -I.:lib:spec bench.rb
Warming up --------------------------------------
             graphql   506.000  i/100ms
Calculating -------------------------------------
             graphql      5.090k (± 0.4%) i/s -     25.806k in   5.070096s
```

On master:

```
$ ruby -I.:lib:spec bench.rb
Warming up --------------------------------------
             graphql   284.000  i/100ms
Calculating -------------------------------------
             graphql      2.880k (± 0.4%) i/s -     14.484k in   5.029007s
```

To ensure compatibility I ran this tokenizer in parallel with the existing tokenizer and compared the results.  There is one difference that I think is acceptable and that is when the tokenizer encounters unknown input.

For example:

```ruby
require "graphql/language/lexer"
require "graphql/language/token"

str = "😘"
p GraphQL::Language::Lexer.tokenize(str)
```

Output on master is this:

```
$ ruby -I.:lib:spec test.rb
[(UNKNOWN_CHAR "\xF0" [1:1]), (UNKNOWN_CHAR "\x9F" [1:2]), (UNKNOWN_CHAR "\x98" [1:3]), (UNKNOWN_CHAR "\x98" [1:4])]
```

Output on my branch is this:

```
$ ruby -I.:lib:spec test.rb
[(UNKNOWN_CHAR "😘" [1:1])]
```

This is because Ragel only understands byte streams where Ruby knows we have a UTF-8 string.

I didn't make many changes to the helper functions that were originally included in the Ragel parser.  Now that we're not working with byte arrays, I think could probably improve the code in those helper functions too.

PR #4467, opened 2023-05-09
Force UTF-8 encoding in lexer

swalkinshaw:
The rewritten lexer in https://github.com/rmosolgo/graphql-ruby/pull/4369 removed UTF8 forced encoding.

It's possible for applications to force encode document strings themselves before passing it to the GraphQL gem, but it should be done at this layer to ensure that invalid encoding is properly handled.

Two questions:
1. is there any reason not to restore this behaviour?
2. is this the proper place? Previously it was done in 3 separate places, but with the new simpler lexer I believe this should cover it.

cc @tenderlove

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -95,12 +95,14 @@ module GraphQL
           previous_token: nil,
         }
 
-        unless string.valid_encoding?
-          emit(:BAD_UNICODE_ESCAPE, 0, 0, meta, string)
+        value = string.dup.force_encoding(Encoding::UTF_8)
+
+        unless value.valid_encoding?
+          emit(:BAD_UNICODE_ESCAPE, 0, 0, meta, value)
           return meta[:tokens]
         end
 
-        scan = StringScanner.new string
+        scan = StringScanner.new value
 
         while !scan.eos?
           pos = scan.pos
--- a/spec/graphql/language/lexer_examples.rb
+++ b/spec/graphql/language/lexer_examples.rb
@@ -52,6 +52,12 @@ module LexerExamples
         |}
         let(:tokens) { subject.tokenize(query_string) }
 
+        it "force encodes to utf-8" do
+          # string that will be invalid utf-8 once force encoded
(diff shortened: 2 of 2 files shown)


rmosolgo:
:+1: Looks like a good solution to me, especially since it _used_ to work this way.  It'd be nice to get a test for it though, so we don't regress again. How did this surface in practice?

swalkinshaw:
Oh yeah I'll add a test.

This surfaced for requests with the `application/graphql` content type. Since that uses the POST body directly without any JSON parsing (which encodes to UTF8). The raw POST body was being tagged as ASCII-8BIT in some cases causing downstream failures (either at the DB level, or within `GraphQL::Types::String` which requires UTF8).

swalkinshaw:
Added a test. Without force encoding, the first token would be `IDENTIFIER`, now its `BAD_UNICODE_ESCAPE`

rmosolgo:

tenderlove:
Sorry, I didn't see this issue.  I don't think this is the right thing to do.  According to the spec, GraphQL queries ["are expressed as a sequence of Unicode characters"](https://spec.graphql.org/June2018/#sec-Source-Text).  When someone sends a query to the GraphQL parser, it's _up to them_ to ensure that the input string is valid UTF-8 encoding.

Imagine a situation where someone makes an HTTP Post with a `Content-Type` header that indicates the incoming POST body is encoded as `Shift-JIS`.  The encoding is correct, there is no "BAD UNICODE ESCAPE".  But now GraphQL will insist that there is.  The correct way to fix problems like this is to look at the source of the string and whatever out of band information there is that indicates the encoding.  In the case of HTTP, that would be the `Content-Type` header.  Using that information, we can correctly tag the ASCII-8BIT string as Shift-JIS, and know to transform it to UTF-8 before handing it to GraphQL.

Using `force_encoding` without external information is rarely the correct solution.  I understand doing the legwork to find where these ASCII-8BIT strings are coming from is a lot of work, but it's the correct way to handle this.  My opinion is that GraphQL should simply reject strings that are not tagged as US-ASCII or UTF-8 (or have an `invalid_encoding?`), but should never call `force_encoding`.

PR #4474, opened 2023-05-12
Fix Enum values that are procs

### LINKED ISSUE #4469: Breaking Change in 2.0.17: Enum values break if mapped to a Proc
**Describe the bug**

Prior to 2.0.17, it was possible to map Procs to enum values, and they'd be passed to the resolver without issue.
As of 2.0.17, the Proc is called in `lib/graphql/schema/enum_value.rb:37:in 'initialize'` (being passed the `UNDEFINED_VALUE` object, or `NOT_CONFIGURED` in 2.0.21) which causes unexpected results.

I think this is a result of using the case-equality `===` operator. 
(`-> (_) { puts 1 } === Object.new` calls the Proc)

**Versions**

`graphql` version: 2.0.21
`rails` (or other framework): 7.0.4.3

**GraphQL schema**

```ruby
class UserSortType < GraphQL::Schema::Enum
  value 'BY_NAME', value: -> (scope, direction: :asc) {
    scope.order(name: direction)
  }
  value 'BY_EMAIL', value: -> (scope, direction: :asc) {
    scope.order(email: direction)
  }
end
```

**Steps to reproduce**

Define an enum with Procs as mapped values.

**Expected behavior**

Prior to 2.0.16, no error is raised. The mapped value is passed to the resolver.

**Actual behavior**

As of 2.0.17:
`NoMethodError: undefined method 'order' for #<Object:0x00007ff688cb8280>`

<details>

```
NoMethodError: undefined method `order' for #<Object:0x00007ff688cb8280>
/home/runner/work/api/api/app/graphql/types/user_sort_type.rb:5:in `block in <class:UserSortType>'
/home/runner/work/api/api/vendor/bundle/ruby/3.2.0/gems/graphql-2.0.21/lib/graphql/schema/enum_value.rb:37:in `initialize'
/home/runner/work/api/api/vendor/bundle/ruby/3.2.0/gems/graphql-2.0.21/lib/graphql/schema/enum.rb:54:in `new'
/home/runner/work/api/api/vendor/bundle/ruby/3.2.0/gems/graphql-2.0.21/lib/graphql/schema/enum.rb:54:in `value'
/home/runner/work/api/api/app/graphql/types/user_sort_type.rb:11:in `<class:UserSortType>'
/home/runner/work/api/api/app/graphql/types/user_sort_type.rb:3:in `<main>'
{...rails internals...}
```

</details>

rmosolgo:
Fixes #4469

### DIFF
--- a/lib/graphql/schema/enum_value.rb
+++ b/lib/graphql/schema/enum_value.rb
@@ -34,7 +34,7 @@ module GraphQL
         @graphql_name = graphql_name.to_s
         GraphQL::NameValidator.validate!(@graphql_name)
         @description = desc || description
-        @value = value === NOT_CONFIGURED ? @graphql_name : value
+        @value = value == NOT_CONFIGURED ? @graphql_name : value
         if deprecation_reason
           self.deprecation_reason = deprecation_reason
         end
--- a/spec/graphql/schema/enum_value_spec.rb
+++ b/spec/graphql/schema/enum_value_spec.rb
@@ -13,6 +13,20 @@ describe GraphQL::Schema::EnumValue do
     end
   end
 
+  it "can accept a proc as a value" do
+    enum = Class.new(GraphQL::Schema::Enum) do
+      graphql_name "SortBy"
+      value 'BY_NAME', value: -> (scope, direction: :asc) {
+        scope.order(name: direction)
+      }
+      value 'BY_EMAIL', value: -> (scope, direction: :asc) {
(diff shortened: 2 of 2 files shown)


mintyfresh:
Tested against our codebase and this resolves the problem!

rmosolgo:
Glad to hear it! I think these test failures are unreleated; I'll investigate them separately.

PR #4483, opened 2023-05-19
GraphQL::Query::NullContext support for fetch, dig, and key?

jturkel:
This PR adds the `fetch`, `dig`, and `key?` methods to `GraphQL::Query::NullContext` for consistency with `GraphQL::Query::Context`


rmosolgo:
Awesome, thanks!

PR #4494, opened 2023-05-31
Disable `Minitest::Reporters` for RubyMine

amomchilov:
RubyMine uses its own internal reporter, and needs `Minitest::Reporters` to be disabled to function properly. This change will let RubyMine users run tests in this repo out-of-the-box without needing to patch this themselves.

I know it's editor-specific, so I'd understand if this is rejected on those grounds, but it's a small just just to the spec_helper, so it won't hurt regular consumers of the gem.


rmosolgo:
Sounds good to me! Thanks for sharing this improvement.

PR #4505, opened 2023-06-06
Document that Timeout can handle Floats

### LINKED ISSUE #4500: Should `Timeout#max_seconds` be allowed to return a `Float`?
The documented behaviour is that only `[Integer, false]` is supported:

https://github.com/rmosolgo/graphql-ruby/blob/b359fc96933f852fa6c8afa9db1769a4ed027fc0/lib/graphql/schema/timeout.rb#L93-L94

The rest of the code in https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/timeout.rb can already handle a `Float` just fine (and in fact, we've been using a Float in production for months with no issue).

Should I update the docs to widen the support return type to `[Numeric, false]`?

amomchilov:
Closes #4500

### DIFF
--- a/lib/graphql/schema/timeout.rb
+++ b/lib/graphql/schema/timeout.rb
@@ -90,7 +90,7 @@ module GraphQL
       # The default implementation returns the `max_seconds:` value from installing this plugin.
       #
       # @param query [GraphQL::Query] The query that's about to run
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for the correction!

PR #4507, opened 2023-06-07
Use `all?` to check types directly

amomchilov:
`all?` can take a "pattern" (anything that responds to `#===`).

In a synthetic benchmark, it's marginally faster to let `all?` call `Class#===` (which has the same semantics as `is_a?`) than to provide a block explicitly:

<details>
<summary>Benchmark</summary>

```ruby
#!/usr/bin/ruby

require "benchmark/ips"

a = [true, false, nil, 1, 2.3, "abc", :abc, [], {}, Object.new] * 10


def found_it!; end

Benchmark.ips do |x|
  x.time = 30
  x.warmup = 10
# x.time = 1
# x.warmup = 1
  
  x.report("all? + is_a?") do |times|
    i = 0
    while i < times
      i += 1
      
      a.all? { |v| v.is_a?(BasicObject) }
      
    end
  end
  
  x.report("just all?") do |times|
    i = 0
    while i < times
      i += 1
      
      a.all?(BasicObject)
      
    end
  end
  
  x.compare!
end
```

</details>

Result:

```
        all? + is_a?    20.952k i/100ms
           just all?    50.146k i/100ms
        all? + is_a?    218.217k (± 1.3%) i/s -      6.558M in  30.057723s
           just all?    502.744k (± 1.5%) i/s -     15.094M in  30.030166s

Comparison:
           just all?:   502744.2 i/s
        all? + is_a?:   218217.2 i/s - 2.30x  slower
```

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -705,7 +705,7 @@ module GraphQL
             end
           when Array
             # It's an array full of execution errors; add them all.
-            if value.any? && value.all? { |v| v.is_a?(GraphQL::ExecutionError) }
+            if value.any? && value.all?(GraphQL::ExecutionError)
               list_type_at_all = (field && (field.type.list?))
               if selection_result.nil? || !dead_result?(selection_result)
                 value.each_with_index do |error, index|


amomchilov:
Interesting, it looks like this overload of `all?` didn't exist in Ruby 2.4.

This'll need more benchmarking, to see if it's worth the minor savings.

rmosolgo:
Wow, TIL about `all?`! Thanks for sharing this idea. Let me know if I can help out while you continue to evaluate its effectiveness.

amomchilov:
> Wow, TIL about all?! 

IKR, it's pretty subtle. TBH, it's less readable than explicitly spelling out `is_a?`, but if it means less Ruby executing in this hot area, then it could be worth it!

> Let me know if I can help out while you continue to evaluate its effectiveness.

Looks like Ruby 2.4 support is the sticking point. What would be a reasonable way to conditionally use one or the other? I see two options which both suck:

1. Put a large `if` around the whole `#continue_value` method body, one which uses the `value.all? { |v| v.is_a?(GraphQL::ExecutionError) }` for Ruby 2.4, and the other which uses `value.all?(GraphQL::ExecutionError)`.
    * The method is huge, so this seems like a terrible idea and a non-starter
    * This has no "runtime" cost, since the branching is only run once when the file is loaded
2. Put a version check on this specific line `708`.
    * Has runtime cost, obviously.

I'm not sure how to go about measuring the cost/benefit here.

rmosolgo:
I'm open to dropping the rails 4 / ruby 2.4 build in GraphQL-Ruby 2.1. Those have been EOL for a long time. If that approach suits you, how about: 

- Removing that build from the Actions configuration
- Making this PR go into `2.1-dev` branch instead of master 
- Updating the description to include a mention of dropping 2.4 support 

What do you think?

rmosolgo:
Thanks for this improvement -- I cherry-picked it into #4760!

Issue #4468, opened 2023-05-09
Default to using resolvers for queries


radar:
:wave: Long time user, first time poster.

In the Getting Started Guide, it's suggested to use this syntax for creating a new field on the `QueryType`:

```ruby
class QueryType < GraphQL::Schema::Object
  description "The query root of this schema"

  # First describe the field signature:
  field :post, PostType, "Find a post by ID" do
    argument :id, ID
  end

  # Then provide an implementation:
  def post(id:)
    Post.find(id)
  end
end
```

An alternative way to define this field would be to separate it out into its own class so that the code that exists in `QueryType` would be:

```ruby
class QueryType < GraphQL::Schema::Object
  description "The query root of this schema"

  field :post, resolver: Resolvers::Post
end
```

This syntax aligns with what's suggested in the "Hooking up Mutations" section here: https://graphql-ruby.org/mutations/mutation_classes.html#hooking-up-mutations.

```ruby
class Types::Mutation < Types::BaseObject
  field :create_comment, mutation: Mutations::CreateComment
end
```

Whereas this syntax could be used instead:

```ruby
class Types::Mutation < Types::BaseObject
  field :create_comment, CommentType, null: false

  def create_comment(input)
  end
end
```

(Apologies if that's wrong, I'm writing it off the top my head. But I think you get the idea anyway.)

I believe having this inconsistency between how _queries_ and _mutations_ are defined is, potentially, confusing to new users of this library. 

I would like to propose that the suggested query syntax be shortened to used distinct resolver classes instead of defining the field resolution method directly in the `QueryType` class. I think this works better as any significantly complex `QueryType` class gets quite cumbersome to work with when there are too many method definitions in there. Especially when both the arguments for the `QueryType` fields are complex, as is the resolution for each field.

Can we switch the `QueryType` to use resolvers, as suggested here: https://graphql-ruby.org/fields/resolvers.html#when-do-you-really-need-a-resolver? Making resolvers the _de facto_ way of using this library, rather than defining methods inside of `QueryType`? I would really love to bring these two classes of `QueryType` and `MutationType` into alignment.

rmosolgo:
I'm definitely open to it! Mutations got special treatment in the past in order to accommodate Relay-specific schema requirements, but maybe doing root-level queries would be better this way. Are you interested in taking a crack at updating the generators and docs?

rmosolgo:
Fixed by #4513

PR #4513, opened 2023-06-13
Use resolver classes for queries by default

radar:
After raising #4468 and getting a thumbs-up from @rmosolgo about it, I've finally gotten around to issuing this PR.

I've updated the install generator to generate a base resolver class, and directed readers of the Getting Started guide to use this class when defining fields. Similar changes have been made elsewhere in the documentation as well.

Ultimately, the goal of this change is to alter the way types such as `QueryType` are defined. Currently, the suggestion is:

```ruby
module Types
    # Add `node(id: ID!) and `nodes(ids: [ID!]!)`
    include GraphQL::Types::Relay::HasNodesField

    field :repos, [RepoType], null: false

    def repos
      Repo.all
    end

    field :repo, RepoResult, null: false do
    end

    def repo(id:)
      Repo.find(id)
    end

    field :category, CategoryType, null: false do
    end

    def category(id:)
      Category.find(id)
    end
  end
end
```

After this change, it will be:

```ruby
module Types
    # Add `node(id: ID!) and `nodes(ids: [ID!]!)`
    include GraphQL::Types::Relay::HasNodesField

    field :repos, resolver: Resolvers::Repo::List
    field :repo, resolver: Resolvers::Repo::Find
    field :category, resolver: Resolvers::Category::Find
```

With each of those resolver classes defining the types, arguments and resolution techniques for the fields themselves. 

I've intentionally left the definition and structure of those resolvers "loose", as I think it's one of those areas where every developer would have an opinion about The One True Way(tm) to structure these classes.

Where I work, we follow this pattern with some decent success, with those resolver classes typically having resolution methods that pass the GraphQL args down to operations, which then resolve however they need to. The operations themselves are tested in isolation and also with GraphQL request specs. Seems to work for us!

### DIFF
--- a/guides/fields/resolvers.md
+++ b/guides/fields/resolvers.md
@@ -99,24 +99,13 @@ So, if there are other, better options, why does `Resolver` exist? Here are a fe
 
 ## Using `resolver`
 
-To add resolvers to your project, make a base class:
-
-```ruby
-# app/graphql/resolvers/base.rb
-module Resolvers
-  class Base < GraphQL::Schema::Resolver
-    # if you have a custom argument class, you can attach it:
-    argument_class Arguments::Base
-  end
-end
-```
-
-Then, extend it as needed:
+Use the base resolver class:
 
 ```ruby
 module Resolvers
-  class RecommendedItems < Resolvers::Base
+  class RecommendedItems < BaseResolver
     type [Types::Item], null: false
+    description "Items this user might like"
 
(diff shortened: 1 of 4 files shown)


rmosolgo:
Hey, thanks for this contribution and sorry it took me so long to get it merged!

radar:
Thank you so much for continuing to maintain this great gem! ❤️

PR #4529, opened 2023-06-21
Query-specific static validators

gmac:
### What

Adds a new `static_validator` parameter to the `Query` object. When assigned as a `StaticValidation::Validator` instance, the query will validate using its own validator rather than the schema's validator.

### Why

This allows trusted queries to be validated with fewer rules. In our case, we have external Rust-based validations available that are very fast and reliable aside from visibility checks, which still need to be run in Ruby. This option allows a subset of Ruby validations to be used for select requests.

### DIFF
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -45,6 +45,20 @@ module GraphQL
       end
     end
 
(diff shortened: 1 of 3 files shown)


rmosolgo:
Amazing -- thanks for the contribution!

PR #4536, opened 2023-07-02
doc: Updated mutation delete/update documentation

indyarocks:
Updated the lib/generators/graphql/mutation_delete_generator.rb AND lib/generators/graphql/mutation_update_generator.rb document/comment to reflect the actual operation


rmosolgo:
Thanks for the fix!

Issue #2398, opened 2019-07-25
NoMethodError (undefined method `line=' for #<GraphQL::Language::Nodes::Document:0x00007fe3f65ec6f8>


gkunwar:
My Gemfiel looks like:
```

gem 'graphql', '~> 1.9.7'
gem 'graphql-libgraphqlparser', '~> 1.3'
```

rmosolgo:
Hi, sorry, libgraphqlparser was broken since 1.9, when we made `Language::Nodes` immutable. See also https://github.com/rmosolgo/graphql-libgraphqlparser-ruby/issues/30

Personally, I don't think libgraphqlparser is worth the effort. The plain-Ruby parser in graphql-ruby is only a few times slower than libgraphqlparser, and it's much more stable, since it doesn't require a C extension. I recommend removing graphql-libgraphqlparser from your gemfile for that reason. 

If you want to _keep_ libgraphqlparser, please share your reasoning, along with benchmarks if applicable, and we can take a look!

PR #4539, opened 2023-07-03
Remove mention of graphql-libgraphqlparser

swrobel:
It hasn't been compatible since graphql-ruby < 1.9, and is no longer supported, as mentioned here: https://github.com/rmosolgo/graphql-ruby/issues/2398#issuecomment-515049940


rmosolgo:
Oh, yeah! There _is_ `graphql-c_parser` now, though, in case you're looking for a faster parser.

swrobel:
> There _is_ `graphql-c_parser` now, though, in case you're looking for a faster parser.

Oh, cool! Is this documented anywhere? It's the first I've heard of it, and all I could find was a [CHANGELOG](https://github.com/rmosolgo/graphql-ruby/blob/master/graphql-c_parser/CHANGELOG.md), but no usage instructions.

swrobel:
@rmosolgo

rmosolgo:
Thanks for the bump, 💁‍♂️ https://github.com/rmosolgo/graphql-ruby/pull/4548

PR #4545, opened 2023-07-06
Warning: mismatched indentations

themudassarhassan:
There is an error loading this gem locally. 
```
Gem Load Error is: /Users/mudassarhassan/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/graphql-2.0.11/lib/graphql/language/lexer.rb:25: warning: mismatched indentations at 'else' with 'if' at 20
```

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -22,7 +22,7 @@ module GraphQL
         NULL =          /null\b/
         QUERY =         /query\b/
         MUTATION =      /mutation\b/
-        SUBSCRIPTION =  /subscription\b/
+        SUBSCRIPTION=  /subscription\b/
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hi! Thanks for looking into this. I think that warning is pointing at an old (v2.0.11) version of `lexer.rb`. On that previous version, there was an `else` on line 25: 

https://github.com/rmosolgo/graphql-ruby/blob/75de90343e02cf1bf76d4a6e42a1e32a7d100f70/lib/graphql/language/lexer.rb#L25

Since then, the lexer was rewritten (https://github.com/rmosolgo/graphql-ruby/pull/4369). The new lexer was released in 2.0.18, so I think if you update to that version, the warning will be gone. Let me know if you encounter this warning after updating!

Issue #4262, opened 2022-12-05
Allow conditional introspection


sporto:
**Is your feature request related to a problem? Please describe.**

We want to disable instrospection based on the current user. So in the controller we would like to have something like

```
def graphql
  if !current_user.super_admin? && Rails.env.production?
      Schema.disable_introspection_entry_points
  end
  ...
end
```

But `disable_introspection_entry_points` cannot be easily reverted. 

**Describe the solution you'd like**

Add some way to disable this conditionally

```
allow_instrospection = ...
MySchema.execute(query_string, context: context, allow_instrospection: allow_instrospection)
```

Thanks

justisb:
I needed to perform this conditional authorization for our application as well, as we only want admin users to be able to access the full schema/field introspection in production, and conditionally build docs for GraphiQL accordingly.

I ended up using the custom introspection system to extend the existing SchemaType and FieldType classes but with authorization built into each field resolution.

e.x.:
```
class FooSchema < GraphQL::Schema
  introspection FooIntrospection
end
```

```
module FooIntrospection
  module Authorizable
    private

    def authorize_introspection!
      # Perform your custom authorization logic based on the instance `context`
      # and raise an error on failure-
      # then have your GraphQLController catch the error and respond with a 403
    end
  end

  class SchemaType < GraphQL::Introspection::SchemaType
    include Authorizable

    def types
      authorize_introspection!

      super
    end

    def query_type
      authorize_introspection!

      super
    end

    def mutation_type
      authorize_introspection!

      super
    end

    def subscription_type
      authorize_introspection!

      super
    end

    def directives
      authorize_introspection!

      super
    end
  end

  class FieldType < GraphQL::Introspection::FieldType
    include Authorizable

    def name
      authorize_introspection!

      object.name
    end

    def description
      authorize_introspection!

      object.description
    end

    def args(...)
      authorize_introspection!

      super
    end

    def type
      authorize_introspection!

      object.type
    end

    def is_deprecated
      authorize_introspection!

      super
    end

    def deprecation_reason
      authorize_introspection!

      object.deprecation_reason
    end
  end
end

```

I think ideally the introspection type fields would use the same policy extensions as the rest of the application's GQL types. However, that seemed even more complicated to implement, so manually overriding each field seemed like a straightforward solution for now.

@rmosolgo I'd be interested to know if this is similar to what you had in mind by using custom introspection system, or if you see any major issues with this kind of implementation and/or clear areas for improvement?

[...]

PR #4551, opened 2023-07-12
Proof on concept, Allow conditional introspection

cody-elhard:
This demonstrates one way to enable introspection at runtime. In our case, admin users provided by context should be able to introspect queries to make tools such as GraphiQL useful.

Fix for https://github.com/rmosolgo/graphql-ruby/issues/4262

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -712,6 +712,11 @@ module GraphQL
         @introspection_system = nil
       end
 
+      def enable_introspection_entry_points
+        @disable_introspection_entry_points = false
+        introspection_system
+      end
+
       def disable_schema_introspection_entry_point
         @disable_schema_introspection_entry_point = true
         # TODO: this clears the cache made in `def types`. But this is not a great solution.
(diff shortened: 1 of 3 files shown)


cody-elhard:
Closing, this worked from the perspective of GraphiQL.

However it does not prevent a user from running the query
`{ __schema { types { name } queryType { name } mutationType { name } } }`

lib/graphql/schema.rb
```diff
@@ -1045,6 +1050,7 @@ def execute(query_str = nil, **kwargs)
         if query_str
           kwargs[:query] = query_str
         end
+        enable_introspection_entry_points if kwargs[:enable_introspection_entry_points]
```
rmosolgo, line 1053: Unfortunately, this won't play nice with a Fiber- or Thread-based application server (like Puma) because in that case, multiple queries may be running at the same time, using the same schema. If one query disabled entry points on the schema, then another query enabled them again, the first query would use the setting from the second query. (It's a memory conflict: the schema is shared between queries, so it the schema is changed, all currently-running queries would use the modified schema.) 

Technically, it would still _work_ in forking servers (like Unicorn), but modifying the Schema class while the application is running would cause it to be copied into fork-specific memory which is slow and a poor use of space. (Although I wouldn't be surprised if some other element of Schema setup causes it to be written into the fork's own heap, so that may not be a _new_ problem...)

rmosolgo (commented):
Hey, thanks for sharing this work! I don't think it will quite work as written, but I'm happy to explore the idea a bit further... 🤔

PR #4553, opened 2023-07-13
Add Schema::Subset for caching filtered subsets of the schema

rmosolgo:
This adds `schema_subset: ...` which is a simplification of dynamic visibility. It's a way to name a subset of your schema and re-use the cached types under the hood which is easier and faster than dynamic visibility.

TODO: 

- [x] If the schema has any subsets, should `schema_subset:` be required? 
- ~~Improve internal structure of subsets for schema and object~~ probably not important since it will be cached
- [x] Implement and test shape checks for fields, arguments, other types, type memberships
- Context-based subsets (which don't use `subsets ...`)
  - [x] Test 
  - [x] Document
- ~~Do we need a `:default` subset? Can it be removed entirely?~~ Keep it as a cache of the _whole_ schema (`nil` uses no cache)

Comparing no cache vs `schema_subset: :default` (a cache of the whole schema), for an introspection query of a very large schema, it's negligibly faster but ~11% less memory: 

```
Run large introspection
                          2.186  (± 0.0%) i/s -     22.000  in  10.069807s
Run large introspection - default schema subset
                          2.235  (± 0.0%) i/s -     23.000  in  10.295016s

# ... 

No Cache:
  Total allocated: 12363238 bytes (87253 objects)
Default Schema Subset:
  Total allocated: 10893150 bytes (78592 objects)
```


rmosolgo:
I don't think I'm going to push ahead on this.

PR #4554, opened 2023-07-13
Try removing quirks_mode: true

rmosolgo:
I got a bug report of GraphQL::Subscriptions::Serialize not playing nice with Oj's JSON serialization when combined with ActiveSupport. It seems like removing `quirks_mode:` made it work again, so we _could_ remove it here.

But, looking back at why it was added: 

- https://github.com/rmosolgo/graphql-ruby/pull/306#issuecomment-253239424
- https://github.com/rmosolgo/graphql-ruby/pull/449

I think its _intention_ is to handle non-Hash and non-Array values properly, which I think is still needed in the codebase. So I think we should find another approach to solve the compatibility issue. 

(We also found that `controller: ...` was in context, and maybe that was causing an infinite loop in serialization.)

### DIFF
--- a/lib/graphql/language.rb
+++ b/lib/graphql/language.rb
@@ -29,7 +29,7 @@ module GraphQL
 
         "[#{serialized_array}]"
       else
-        JSON.generate(value, quirks_mode: true)
+        JSON.generate(value)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


rmosolgo:
Given that this was added to address compatibility issues in the past, an another approach solved this same issue in the app, I'm not going to explore this further.

PR #4565, opened 2023-07-24
Add name to static validator

gmac:
Now that a [query may define its own validator](https://github.com/rmosolgo/graphql-ruby/pull/4529), it would be useful for instrumentation if a static validator could also specify a ruleset name for itself.


gmac:
Holding on this for now...

PR #4567, opened 2023-07-25
fix: typo in pundit resolver authorization example

mjfaga:
Pretty sure this is a typo. If it's not, my apologies for misreading the docs.


rmosolgo:
You're right! It should be just like these: https://github.com/rmosolgo/graphql-ruby/blob/master/guides/authorization/pundit_integration.md#authorizing-objects Thanks for the fix :+1:

Issue #4389, opened 2023-03-17
`No query string was present` error on all requests using PersistedQuery since 2.0.18


ThomasF34:
**Describe the bug**

Following the 2.0.18 update,  I have the error `"No query string was present"` on all requests.
(Note: we are using PersistedQuery)

**Versions**

- `graphql` version: `2.0.18` and `2.0.19`
- `rails`:  `7.0.4.3`
- other applicable versions:
  - `graphql-pro`: `1.24.10`
  - `graphql-batch`: `0.5.2`

**GraphQL query**

Here is an example of what happens

Request
```json
{
        "operationId":"sweep-webapp/3bf175adebc8f3cdedef06d22a9152ee",
        "operationName":"GetCurrentUser",
        "variables": {}
}
```

Response
```json
{
	"errors": [
		{
			"message": "No query string was present"
		}
	]
}
```

**Additional context**

- We are using PersistedQuery
- The operation with alias `3bf175adebc8f3cdedef06d22a9152ee` exists in DB
- The operation `last_used` timestamp is correctly updated 
- No change was done on our schema, only the update from 2.0.17 to 2.0.18 (or 2.0.19) introduced this bug 

Also, I found this PR (#4344) that [changed the parsing](https://github.com/rmosolgo/graphql-ruby/pull/4344/files#diff-32e3c2fdba5c368945b54604800adcca75ca505ab650327ffb6cc0611aeb79b0R378) of the query string, [used to trigger or not the error I encountered ](https://github.com/rmosolgo/graphql-ruby/pull/4344/files#diff-32e3c2fdba5c368945b54604800adcca75ca505ab650327ffb6cc0611aeb79b0R400)

Let me know if you need more info

rmosolgo:
Yikes! Sorry for the trouble and thanks for the report. I'll take a look and follow up here soon.

rmosolgo:
Could you share the full backtrace of the error, please? I expect there's some problem in ordering the different tracers, but I can't quite replicate it. The backtrace would include clues of what was called when the error was raised.

ThomasF34:
Thanks for having a look

Unfortunatelly I had no stack trace as no error is raised. The error is only appearing in the request `errors` field and the only action seen in the backend is the update of the operation's `last_used_at` attribute

```ruby
Started POST "/api/q" for 172.18.0.1 at 2023-03-20 14:48:48 +0100
Processing by GraphqlController#execute as */*
[...]

PR #4568, opened 2023-07-27
Datadog: Access the query string after calling super, in case the query string was modified

rmosolgo:
This might fix order dependency issues like #4389 

TODO: 

- [x] Add a test for this behavior


rmosolgo:
This didn't fix the problem I was after...

PR #4569, opened 2023-07-27
Add Source#result_key_for

rmosolgo:
This adds a way to deduplicate `.load(...)` calls when the passed-in object is equivalent but doesn't implement `.hash` to the same value.

### DIFF
--- a/guides/dataloader/sources.md
+++ b/guides/dataloader/sources.md
@@ -168,3 +168,11 @@ end
 ```
 
 In this case, `records` will include the _first_ object for each unique `record.id` -- subsequent records with the same `.id` will be assumed to be duplicates. Under the hood, the `Source` will cache the result based on the record's `id`.
+
+Alternatively, you could use this to make the `Source` retain each incoming object, even when they would _otherwise_ be treated as duplicates. (This would come in handy when you need `def fetch` to mutate each object). For example, to treat _every_ incoming object as distinct:
+
+```ruby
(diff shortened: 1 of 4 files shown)


guides/dataloader/sources.md
```diff
@@ -148,3 +148,23 @@ dataloader.with(Sources::ActiveRecordObject, Comment).merge(comments_by_id)
 ```
 
 After that, any calls to `.load(id)` will use those already-loaded records if they're available.
+
+## De-duplicating equivalent objects
+
+Sometimes, _different_ objects in the application should load the same object from `fetch`. You can customize this behavior by implementing `def result_key_for(key)` in your application. For example, to map records from your ORM to their database ID:
```
tgwizard, line 154: The use-case I have is actually the opposite - two Active Record objects that point to the same DB row both `hash` and `eql?` the same, but for my loader (a version of an [`AssociationLoader`](https://github.com/Shopify/graphql-batch/blob/main/examples/association_loader.rb)) needs to see all the records  in `fetch`, without the deduping that happens when Active Record objects are used as keys in hashes/sets.

tgwizard (approved):
LGTM, thank you!

guides/dataloader/sources.md
rmosolgo: Got it -- I updated the doc to mention this case, as well. Thanks for clarifying!

PR #4571, opened 2023-07-28
Build out trace modes

rmosolgo:


rmosolgo:
(CI failure looks like a gem is incompatible with an old Ruby, I'll look later....)

PR #4577, opened 2023-08-04
Remove deprecated Visitor hooks API

swalkinshaw:
This API has been deprecated for a while now. Should it be removed completely?

`TypeStack` confused me but it appears to be entirely unused since the class-based visitor refactor.

Questions:
* do you want `SKIP` removed completely too? Judging by the comment it should be


rmosolgo:
Sadly, there's a downstream dependency on TypeStack in GraphQL-Client: 

https://github.com/github/graphql-client/blob/2c611765e34ded9e63085b4b40a8da59d8ce8732/lib/graphql/client/document_types.rb#L24

I don't want to break it 🤦 

I'm open to removing the rest of it in 2.1, though, thanks!

swalkinshaw:
🤔 but `TypeStack` depends on the old visitor API so wouldn't this all need to remain?


What about just constraining the `graphql` dependency? https://github.com/github/graphql-client/blob/2c611765e34ded9e63085b4b40a8da59d8ce8732/graphql-client.gemspec#L13C1-L13C1

Doesn't seem ideal to keep code around for another project which is itself unmaintained.

rmosolgo:
Yeah, that seems reasonable to pin graphql-client to an old GraphQL-Ruby version :+1: 

I think it's right to remove `SKIP`, too, since it was only used by visitor hooks.

wuarmin:
Hey @rmosolgo,
can you give some hints, how the issue can be solved at graphql-client? Pinning to an old version is just a compromise solution.

Thanks

PR #4580, opened 2023-08-08
Fix block string quote unescaping

tenderlove:
Triple quotes at the beginning and end of lines shouldn't be removed if they have been escaped.  IOW only the leading and trailing """ should be removed from the token


rmosolgo:
Nice, thanks for the fix!

Issue #3437, opened 2021-04-14
Make clientMutationId optional for RelayClassicMutation


rylanc:
**Is your feature request related to a problem? Please describe.**

I find the automatic generation of an input type a useful feature of the `RelayClassicMutation`, but the addition of the `clientMutationId` doesn't apply to my client (I use Apollo). Leaving it in the schema can be confusing for consumers who aren't familiar with Relay (it doesn't even look like Relay uses these IDs anymore). 

**Describe the solution you'd like**

It would be handy if the client mutation ID field/argument were optional. I think it would probably be the easiest to configure (while maintaining backward compatibility) to create a separate `InputArgumentMutation` (name TBD) class that `RelayClassicMutation` inherits from. The base class would include the logic for creating and handling the input argument, and `RelayClassicMutation` would just add and handle the client mutation ID field/argument.

Another option might be to have a class method on `RelayClassicMutation` called `add_client_mutation_id` that would accept a boolean and would be enabled by default. That approach seems to me to be a little more confusing than using composability/inheritance.

**Additional context**

I'm happy to do the work to add this feature, but I wanted to get some feedback before I started.

rmosolgo:
Hey, thanks for the suggestion. I've definitely heard folks confused by the presence of that argument, just like you described. 

Either approach that you proposed sounds good to me, and I'd welcome a PR for either one. Please let me know if I can help out along the way!

rmosolgo:
I don't have plans to work on this, but in case it's still an issue for anyone, here's an approach to make `clientMutationId` go away in your schema. Basically, you intercept the calls that would add that field and mutation, and make them no-ops instead. It's a bit brittle, since the gem's internal flow might change over time, so if you adopt it, I recommend adding a test to your application to make sure it's having the desired effect. That way, if the gem changes later on, you'll know when you update it.

```
require "bundler/inline"

gemfile do
  gem "graphql", "1.12.17"
end

class BaseObject < GraphQL::Schema::Object
  # ... your app's base object here
end

class BaseMutationPayload < BaseObject
  def self.add_field(field, *args, **kwargs)
    if field.graphql_name == "clientMutationId"
      # do nothing -- don't want this field
    else
      super
    end
  end
end

class BaseInputObject < GraphQL::Schema::InputObject
  # ... your app's base input object here
end

class BaseMutationInput < BaseInputObject
  def self.argument(name, *args, **kwargs, &block)
    if name == :client_mutation_id
[...]

PR #4581, opened 2023-08-14
Separate input argument into module

tueda0703:
### Summary

There are two functions mixed in RelayClassicMutation, the input relation and the clientMutationId relation.
In our project, we want to use only the input relation, so it should be separated as a single module.
I researched this request and found the following Issue.
https://github.com/rmosolgo/graphql-ruby/issues/3437
This Issue said "and I'd welcome a PR for either one." so I did it.
The reason it's Closed is because no one else was doing it at the time, not that the change itself is unacceptable, right?

### Changes

The input-related functions were split into modules from the two functions of input and clientMutationId.
The behavior has not been changed.

### Tested

- pass unit test
- work in a local environment with a graphql server


tueda0703:
> Hi, thanks for the suggestion! Does this improve GraphQL-Ruby in some way, or is it a fix for a bug? Could you please help me understand the motivation for this change?

This module will be used to improve the functionality.

RelayClassicMutation has two functions: input argument and clientMutationId.

I thought there might be cases where only the input argument functionality would be used, so I divided it into modules.

The user can use
```Ruby
def ExampleMutation < GraphQL::Schema::Mutation
    include GraphQL::Schema::HasSingleInputArgument
...
end
```
only the input argument functionality is available.

tueda0703:
The conflict was resolved.

The PR description has been rewritten with more detail.

lib/graphql/schema/relay_classic_mutation.rb
```diff
@@ -31,34 +35,14 @@ class RelayClassicMutation < GraphQL::Schema::Mutation
       def resolve_with_support(**inputs)
```
tueda0703, line 35: The behavior has not changed, although a slight modification has been made due to the effect of cutting out input-related processing.

rmosolgo:
Hi! Thanks for updating this and sorry I didn't write back sooner. 

I'm open to accepting this change except I'd rather keep the `Jazz::...` tests  and `relay_classic_mutation_spec.rb` the same. (I'd rather keep them the same so I can keep the high-level tests on RelayClassicMutation, which is widely used -- I don't want to break it!) Instead, how about making a new Schema class in your new test file, and running tests against that new schema?

Here's an example test file with a small schema in it: https://github.com/rmosolgo/graphql-ruby/blob/f6da30f70007be9600783eb1ca054dc771533c8d/spec/graphql/schema/directive/transform_spec.rb#L3C1-L19C6

Do you mind updating this test that way? Then I think we'll be good to merge.

tueda0703:
Thanks for the review.

Fixed test.
Please re-review.

PR #4585, opened 2023-08-16
Support early-return of falsey values from ready/authorized

eapache-opslevel:
If you write a resolver's authorized method with a `return false, nil` in it (to early-return a value of `nil`), it was treating this just as `return false` and raising, because the early-return value was not truthy. Instead explicitly check for an array value before splitting it in two.

I also did `ready?` while I was here since it had the same pattern.

~~The tests are "correct" but won't fail even without my "fix" because the test schema doesn't implement `def unauthorized_object(unauthorized_error)`, and there are a bunch of tests already relying on it being not implemented. Not sure what the best approach is.~~
*edit: I realized after posting that it was easier and more correct to test for a false/nil distinction which sidesteps the `unauthorized_object` issue*

@rmosolgo @dougedey

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -65,19 +65,20 @@ module GraphQL
       # @api private
       def resolve_with_support(**args)
         # First call the ready? hook which may raise
-        ready_val = if args.any?
+        raw_ready_val = if args.any?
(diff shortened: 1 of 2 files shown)


rmosolgo (approved):
Thanks for the fix!

PR #4592, opened 2023-08-22
Accept dataload: true for fields that need batching

rmosolgo:
Currently, _every_ object returned by a graphql field is checked to see if it needs lazy resolution. This wastes a lot of time (~5% of runtime in one benchmark). We could make this faster by _not_ always checking, and a first attempt would be to require opt-in for dataloaded values. (This could be more ergonomic, see below.) This only reduces overhead in fields that _don't_ require batch loading, but in my experience, that's _most_ fields (most fields are leaf fields; leaf fields don't usually use batch loading). 

### TODO 

- [ ] Make a "development mode" where it _still_ checks for lazy values and reports them, so that the config can be added 
- [ ] Could there even be a diagnostic output + Rubocop correction for automatically migrating?
- [ ] Lazy values from `resolve_type`, `authorized?`, and argument `prepare`s can "pollute" downstream, making some fields require `dataload: true` even though the field has nothing to do with that. Gotta fix this somehow.

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -550,8 +550,15 @@ module GraphQL
                 ex_err
               end
             end
-            after_lazy(app_result, field: field_defn, ast_node: ast_node, owner_object: object, arguments: resolved_arguments, result_name: result_name, result: selection_result) do |inner_result|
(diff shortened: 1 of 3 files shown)


rmosolgo:
I'm still looking for some way to reduce the overhead, but I don't think this is it.

PR #4593, opened 2023-08-22
Copy ActiveSupport:IsolatedExecutionState when spawning fibers

Goos:
This is a bit of a throwback to [this PR](https://github.com/rmosolgo/graphql-ruby/pull/3461). 

Much like @zvkemp presciently alluded to in his comment:

> [...] so I would caution against assuming that this strategy will be compatible with other gems or application code.

ActiveSupport has a construct named `IsolatedExecutionState` which, rather than storing things in `Thread.current[]`, [defines accessors on the Thread and Fiber types themselves](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/isolated_execution_state.rb#L9-L10).  This means they won't be copied when iterating over `Thread.current[]` while spawning new fibers, which in my case ended up breaking the [rails logger's tag stack](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/tagged_logging.rb#L56-L60).

While this is pretty hacky, copying this state over should keep the loggers working as expected. That being said, I would love some suggestions on how to do this in a more elegant way. I tried fixing it on my end first by adding a custom tracer that copies the state over inside the `execute_query_lazy` event, but that only handled the top-level calls, while dataloader sources were still missing the context 🤷


rmosolgo:
Hey, thanks for the write-up and sharing this suggestion. I agree that GraphQL-Ruby should cover this somehow.

Goos:
@rmosolgo thanks for having a look! Let me know if there are any more details I can share, or ideas for alternative approaches – happy to help.

rmosolgo:
My idea for an alternative approach is to expose methods on `GraphQL::Dataloader` which handle this work. Then, in application code, allow a custom subclass of `GraphQL::Dataloader` to override those methods. For example: 

```ruby 
class CustomDataloader < GraphQL::Dataloader 
  def get_fiber_state
    state_h = super 
    state_h[:active_support_isolated_execution_state] = ActiveSupport::IsolatedExecutionState.active_support_execution_state
    state_h # return the modified hash 
  end 
  
  # whatever was returned from the previous method is passed in here:
  def set_fiber_state(state_h) 
    ActiveSupport::IsolatedExecutionState.active_support_execution_state = state_h.delete(:active_support_isolated_execution_state)
    super
  end 
end 
```

And then those methods could be used here, moving all the `Thread.current` stuff into the default implementations of those methods: 

https://github.com/rmosolgo/graphql-ruby/blob/be70e18683b138a9a4b1225000319f4f73434b7f/lib/graphql/dataloader.rb#L282-L309

That's my _idea_, please give it a shot if you're interested! Otherwise I'll make some time soon to give it a try.

Goos:
@rmosolgo makes sense – that sounds a lot less brittle than the changes I made! I can try to have a go at it sometime this week.

Goos:
@rmosolgo finally got around to doing this. I pretty much took your suggestion line-by-line, with the only difference being that I called it `fiber_vars` instead of `fiber_state`, since I figured "state" might be confused with the execution state of the fiber.

Goos:
@rmosolgo is there anything else you need for us to get these changes in?

rmosolgo:
Derp -- thanks for the bump. I renamed a few methods and meant to merged right after CI passed but ... I forgot 🙈 

Thanks for this improvement!

Goos:
Haha all good, thanks for the help!

PR #4603, opened 2023-08-30
fix: onCompleted typo in createRelaySubscriptionHandler

mjfaga:
When using this handle with the `createActionCableHandler`, , an error is thrown because it expects the observer object to have an `onCompleted` method, but `onComplete` is currently passed from the `createRelaySubscriptionHandler`. 

For reference: https://github.com/rmosolgo/graphql-ruby/blob/master/javascript_client/src/subscriptions/createActionCableHandler.ts#L60

Fixes https://github.com/rmosolgo/graphql-ruby/issues/4604


mjfaga:
That failing test appears to be related to something in the rails ecosystem, which this PR doesn't touch.

rmosolgo:
Thanks for the fix! I just released in in v1.11.9.

PR #4617, opened 2023-09-05
Clear Dataloader cache around mutation root fields

### LINKED ISSUE #4600: fiber-based dataloaders prevent reloading mutated objects
**Describe the bug**

Mutations can use dataloaders to load objects from arguments, in my case, via the `loads:` argument. For example:
~~~ruby
def self.object_from_id(encoded_id, context)
  context.dataloader.with(Sources::ByGlobalID).load(encoded_id)
end
~~~

Dataloaders can also be used to authorize access to the loaded types before passing the object to the mutation's resolve method. For example:
~~~ruby
def self.authorized?(object, context)
  context.dataloader.with(MyLoader, SomeObject, context[:current_user]).load(object)
end
~~~

Whatever was loaded by dataloaders to prepare for a mutation's invocation, remains cached in that dataloader. Any fields exposed by the mutation can be traversed using regular GraphQL as expected, but any traversal from one type to another that:
A) uses a dataloader
and B) was loaded by that dataloader before the resolve function was called

Cannot be reloaded because the dataloader has that object cached internally.

**Versions**

`graphql` version: 2.0.13
`rails` (or other framework): 6.1.7.4
other applicable versions: using the fiber-based dataloaders provided with the graphql gem

**GraphQL query**

Mutation loads a B object via bId, and the mutation exposes A as a field. Traversing from a -> b in a query results in a stale lookup. In the following case, the old name is still visible.

```graphql
mutation {
  updateBName(input: { bId: "123", name: "new-name" }){
    a {
      b {
        name
      }
    }
  }
}
```

**Expected behavior**

Probably the dataloader cache should be purged directly before or maybe after resolving a mutation, but definitely before resolving the returned fields so that fresh objects can be fetched.

**Actual behavior**

What specifically went wrong?

Stale objects are returned. Mutations should be able to return fields that reflect the updates that were just made.

**Additional context**

Resetting dataloader cache in my own mutation resolve method seems to solve my issue, but its a hack since there is no exposed way to reset the dataloader cache:
`context.dataloader.instance_variable_set(:@source_cache, Hash.new { |h, k| h[k] = {} })`
-or-
`context.dataloader.instance_variable_get(:@source_cache).values.each(&:clear)`

rmosolgo:
Fixes #4600

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -104,6 +104,15 @@ module GraphQL
       nil
     end
 
+    # Clear any already-loaded objects from {Source} caches
+    # @return [void]
+    def clear_cache
+      @source_cache.each do |_source_class, batched_sources|
+        batched_sources.each_value(&:clear_cache)
+      end
+      nil
+    end
+
     # Use a self-contained queue for the work in the block.
     def run_isolated
       prev_queue = @pending_jobs
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -161,6 +161,13 @@ module GraphQL
         [*batch_args, **batch_kwargs]
       end
 
+      # Clear any already-loaded objects for this source
+      # @return [void]
+      def clear_cache
+        @results.clear
+        nil
+      end
+
       attr_reader :pending
 
       private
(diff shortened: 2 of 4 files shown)


swalkinshaw:
Is this something that could be built into lazy execution so that graphql-batch wouldn't need it's own implementation?

rmosolgo:
Yeah, I think there might be something to gain here. I'm not sure how GraphQL-Ruby would (or should?) know the internals of GraphQL-Batch, but some kind of `clear_cache` _hook_ for GraphQL-Batch to listen to would make sense 👀

Bertg:
@rmosolgo This fix has been on main almost 3 weeks. Any ETA on an official release?

PR #4619, opened 2023-09-05
Try debugging authorization bug

### LINKED ISSUE #4618: 2.0.27 broke authorized checks
**Describe the bug**

After upgrading to 2.0.27, when an `authorized?` returns false `Schema#execute` is now returning nil instead of a result hash with errors.

I have both `unauthorized_object` and `unauthorized_field` implemented to raise a GraphQL::ExecutionError.

I suspect this is related to: https://github.com/rmosolgo/graphql-ruby/pull/4585

**Versions**

`graphql` version: 2.0.27
`rails` (or other framework): 7.0.7.2

**GraphQL schema**

I'm using custom field and argument extensions to support configurable authorized checks

```ruby
# GraphQL Schema
  mutation(Types::MutationType)
  query(Types::QueryType)

  use GraphQL::Batch

  # Stop validating when it encounters this many errors:
  validate_max_errors(100)

  max_depth 15
  max_complexity 300

  disable_introspection_entry_points if Rails.env.production?

  rescue_from(ActiveRecord::RecordNotFound) do |_err, _obj, _args, _ctx, field|
    # Raise a graphql-friendly error with a custom message
    raise GraphQL::ExecutionError, "#{field.type.unwrap.graphql_name} not found"
  end

  def self.unauthorized_object(error)
    # Add a top-level error to the response instead of returning nil:
    raise GraphQL::ExecutionError, "An object of type #{error.type.graphql_name} was hidden due to permissions"
  end

  def self.unauthorized_field(error)
    # Add a top-level error to the response instead of returning nil:
    msg = "The field #{error.field.graphql_name}" \
          "on an object of type #{error.type.graphql_name} was hidden due to permissions"
    raise GraphQL::ExecutionError, msg
  end
end

module Types
  # Base class for GraphQL argument types
  class BaseArgument < GraphQL::Schema::Argument
    def initialize(*args, required_permission: [], **kwargs, &block)
      @required_permission = required_permission
      # Pass on the default args:
    end

    def authorized?(obj, arg_value, ctx)
      policy, perm = @required_permission
      allowed = policy ? policy.public_send(:"#{perm}?", ctx[:current_user], arg_value) : true

      super && allowed
    end
  end
end

module Mutations
  class AddChannelToOrganization < BaseMutation
    description 'Add a channel to an organization'

    field :errors, [Types::UserErrorType], null: false, description: 'Errors with the input provided'

    argument :channel_id, String, description: 'The ID of the channel'
    argument :organization_id, ID, loads: Types::OrganizationType,
                                   required_permission: [Profiles::Policy, :add_channel_to_organization],
                                   description: 'The ID of the organization to add the channel to'

    def load_organization(organization_id)
      #...
    end

#...
end

```

**GraphQL query**

```graphql
      mutation Mutation($input: AddChannelToOrganizationInput!) {
        addChannelToOrganization(input: $input) {
          errors {
            path
            type
          }
        }
      }
```

```json
nil
```

**Steps to reproduce**

* Override `unauthorized_field` to raise a `GraphQL::ExecutionError`
* Return false from an `authorized?` check for a field.

**Expected behavior**

When `authorized?` returns false a `GraphQL::ExecutionError` is raised and errors are added to the result hash under `errors`

**Actual behavior**

Result of `Schema#execute` is nil.

rmosolgo:
Trying to fix #4618

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -95,13 +95,20 @@ module GraphQL
                 # If the `authorized?` returned two values, `false, early_return`,
                 # then use the early return value instead of continuing
                 if authorized_val.is_a?(Array)
-                  authorized_result, early_return = authorized_val
+                  return_early = true
+                  authorized_result, early_return_value = authorized_val
+                else
+                  authorized_result = authorize_val
+                  return_early = false
+                end
+
+                if return_early
                   if authorized_result == false
-                    early_return
+                    early_return_value
                   else
-                    raise "Unexpected result from #authorized? (expected `true`, `false` or `[false, {...}]`): [#{authorized_result.inspect}, #{early_return.inspect}]"
+                    raise "Unexpected result from #authorized? (expected `true`, `false` or `[false, {...}]`): [#{authorized_result.inspect}, #{early_return_value.inspect}]"
                   end
-                elsif authorized_val
+                elsif authorized_result
                   # Finally, all the hooks have passed, so resolve it
                   if loaded_args.any?
                     public_send(self.class.resolve_method, **loaded_args)
--- a/spec/graphql/schema/resolver_spec.rb
+++ b/spec/graphql/schema/resolver_spec.rb
@@ -495,6 +495,12 @@ describe GraphQL::Schema::Resolver do
         end
       end
     end
+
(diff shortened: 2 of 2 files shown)


rmosolgo:
Nailed this down in #4839

PR #4620, opened 2023-09-05
Update mutation_errors.md

craigulliott:
The documentation is incorrect. `attribute` here is actually an ActiveModel::Error object which has a method named `attribute` and a method named `message`.  The message argument in the current example is always nil.


rmosolgo:
Thanks for the correction!

PR #4625, opened 2023-09-07
Rewrite dataloader to use Fiber#transfer

rmosolgo:
I'm trying to find a way to make GraphQL::Dataloader work with non-blocking fibers that use `.transfer`. I tried an incremental approach in #4322 , but it's not really working -- I think I need to rethink how these parts fit together. 

This is coming along nicely -- it's actually much simpler -- but it doesn't support `blocking: false` yet.


rmosolgo:
The reason I paused on this is because I _really_ wanted to get it working with the `async` gem. But I couldn't :S In any case, I think the cleaned-up algorithm is an improvement, so I'll merge it.

rmosolgo:
This change was rolled back in #4752 because `Fiber#transfer` has unpredictable control flow. See: 

- https://bugs.ruby-lang.org/issues/20081
- #4743
- #4739

PR #4627, opened 2023-09-09
makes install templates more rubocop friendly.

aaronmallen:
Adds `frozen_string_literal` comment to all templates that did not have it. reduce method size and complexity in the `graphql_controller` template.

### DIFF
--- a/lib/generators/graphql/install/templates/base_mutation.erb
+++ b/lib/generators/graphql/install/templates/base_mutation.erb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 <% module_namespacing_when_supported do -%>
 module Mutations
   class BaseMutation < GraphQL::Schema::RelayClassicMutation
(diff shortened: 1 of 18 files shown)


rmosolgo:
Hi! Thanks for the suggestion. I definitely agree that the frozen string comments are an improvement. 

However I don't think that the changes to the controller code actually improve it. It takes the core behavior of the controller action and scatters it around the file, so a new reader has to hunt around to find what it actually _does_. I'd prefer to keep it as it was, and anyone who wants to move those lines into their own methods can do it themselves. Do you mind removing those changes from this PR?

aaronmallen:
> Hi! Thanks for the suggestion. I definitely agree that the frozen string comments are an improvement.
> 
> However I don't think that the changes to the controller code actually improve it. It takes the core behavior of the controller action and scatters it around the file, so a new reader has to hunt around to find what it actually _does_. I'd prefer to keep it as it was, and anyone who wants to move those lines into their own methods can do it themselves. Do you mind removing those changes from this PR?

done

PR #4631, opened 2023-09-12
Avoid using public_send on Context#[]

mtsmfm:
Hi, thank you for maintaining awesome library ⭐
Please feel free to close this PR if it sounds like kind of an improvement that should not be made by an external contributor, such as frozen string things in the past.

---

This PR will improve performance by 1.02x in this benchmark.

<details>
<summary>

```
$ ruby bench.rb
          with patch     1.000  i/100ms
       without patch     1.000  i/100ms
          with patch      0.814  (± 0.0%) i/s -     25.000  in  30.737210s
       without patch      0.798  (± 0.0%) i/s -     24.000  in  30.094314s

Comparison:
          with patch:        0.8 i/s
       without patch:        0.8 i/s - 1.02x  slower
```

</summary>

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'benchmark-ips', require: 'benchmark/ips'
  gem 'graphql'
end

FIELD_COUNT = 100
ARTICLES_COUNT = 1000

ALL_ARTICLES = ARTICLES_COUNT.times.map do |i|
  {title: "title#{i}"}
end

module GraphQL
  module Execution
    class Interpreter
      class Runtime
        class CurrentState
          KEYS = [:current_result, :current_result_name, :current_arguments, :current_field, :current_object]

          def initialize
            @data = {}
          end

          def [](key)
            @data[key]
          end

          KEYS.each do |k|
            define_method(k) { @data[k] }
            define_method(:"#{k}=") {|v| @data[k] = v }
          end
        end
      end
    end
  end

  class Query
    class Context
      prepend(Module.new do
        def [](key)
          unless $with_patch
            return super
          end

          if @scoped_context.key?(key)
            @scoped_context[key]
          elsif @provided_values.key?(key)
            @provided_values[key]
          elsif RUNTIME_METADATA_KEYS.include?(key)
            if key == :current_path
              current_path
            else
              (current_runtime_state = Thread.current[:__graphql_runtime_info]) &&
                (query_runtime_state = current_runtime_state[@query]) &&
                (query_runtime_state[key])
            end
          else
            # not found
            nil
          end
        end
      end)
    end
  end
end

class ArticleType < GraphQL::Schema::Object
  FIELD_COUNT.times do |i|
    field "title#{i}", String, null: false do
      argument :arg, String
    end

    define_method "title#{i}" do |arg:|
      object[:title]
    end
  end
end

  field :articles, [ArticleType], null: true

  def articles
    ALL_ARTICLES
  end
end

  query QueryType
end

QUERY = <<GQL
  {
    articles { #{FIELD_COUNT.times.map {|i| "title#{i}(arg: \"a\")" }.join(",")} }
  }
GQL

Benchmark.ips do |x|
  x.time = 30

  x.report('with patch') do
    $with_patch = true
    TestSchema.execute(QUERY)
  end

  x.report('without patch') do
    $with_patch = false
    TestSchema.execute(QUERY)
  end

  x.compare!
end

__END__
$ ruby bench.rb
          with patch     1.000  i/100ms
       without patch     1.000  i/100ms
          with patch      0.814  (± 0.0%) i/s -     25.000  in  30.737210s
       without patch      0.798  (± 0.0%) i/s -     24.000  in  30.094314s

Comparison:
          with patch:        0.8 i/s
       without patch:        0.8 i/s - 1.02x  slower
```

</details>

---

I have an interest in performance improvements on graphql-ruby.

I took benchmarks for each version of the GraphQL Ruby 2 series and noticed that it has almost doubled in speed (in my example) since around 2.0.20.

image

https://github.com/mtsmfm/graphql-ruby-benchmark/blob/d4a390f392c01a9e1fc19ba8371787dde7452424/results/figs/result-2023-09-09T03%3A03%3A08%2B00%3A00.png

I was wondering which PR contributes most so I also took benchmark per PR.

image

https://github.com/mtsmfm/graphql-ruby-benchmark/blob/d4a390f392c01a9e1fc19ba8371787dde7452424/results/figs/result-performance_prs-2023-09-10T03%3A38%3A18%2B00%3A00.png

I found https://github.com/rmosolgo/graphql-ruby/pull/4399 is one of the most important PR so I read the changes on the PR.
It seems the main purpose is to make a dedicated object but I found actually that part slows down the library since `public_send` is much slower than direct method call.

<details>
<summary>

```
$ ruby foo.rb 
                send   889.676k i/100ms
              direct     2.157M i/100ms
                send      8.757M (± 7.4%) i/s -     43.594M in   5.008124s
              direct     20.391M (± 7.0%) i/s -    103.546M in   5.105176s

Comparison:
              direct: 20391490.9 i/s
                send:  8756908.0 i/s - 2.33x  slower
```

</summary>

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'benchmark-ips', require: 'benchmark/ips'
  gem 'graphql'
end

state = GraphQL::Execution::Interpreter::Runtime::CurrentState.new

Benchmark.ips do |x|
  x.report('send') do
    state.public_send(:current_arguments)
  end

  x.report('direct') do
    state.current_arguments
  end

  x.compare!
end

__END__
$ ruby foo.rb 
                send   889.676k i/100ms
              direct     2.157M i/100ms
                send      8.757M (± 7.4%) i/s -     43.594M in   5.008124s
              direct     20.391M (± 7.0%) i/s -    103.546M in   5.105176s

Comparison:
              direct: 20391490.9 i/s
                send:  8756908.0 i/s - 2.33x  slower
```
</details>

`Thread#[]` is fiber local variable and changing Hash object into `CurrentState` class doesn't affect the behavior in my understanding.

https://ruby-doc.org/3.2.2/Thread.html#method-i-5B-5D

I found actual improvement is done by this commit.

https://github.com/rmosolgo/graphql-ruby/pull/4399/commits/516265ebf90b65cda73433a3d07c76419e7e3137

If there's no argument on fields it can be huge improvement.

I wondered whether I should partially revert the original changes, but I decided to make the changes to keep backward compatibility.

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -9,17 +9,20 @@ module GraphQL
       # @api private
       class Runtime
         class CurrentState
+          KEYS = [:current_result, :current_result_name, :current_arguments, :current_field, :current_object, :was_authorized_by_scope_items]
+
           def initialize
-            @current_object = nil
-            @current_field = nil
-            @current_arguments = nil
-            @current_result_name = nil
-            @current_result = nil
-            @was_authorized_by_scope_items = nil
+            @data = {}
           end
 
-          attr_accessor :current_result, :current_result_name,
-            :current_arguments, :current_field, :current_object, :was_authorized_by_scope_items
+          def [](key)
+            @data[key]
+          end
+
+          KEYS.each do |k|
+            define_method(k) { @data[k] }
+            define_method(:"#{k}=") {|v| @data[k] = v }
+          end
         end
 
         module GraphQLResult
--- a/lib/graphql/query/context.rb
+++ b/lib/graphql/query/context.rb
@@ -132,7 +132,7 @@ module GraphQL
           else
             (current_runtime_state = Thread.current[:__graphql_runtime_info]) &&
               (query_runtime_state = current_runtime_state[@query]) &&
-              (query_runtime_state.public_send(key))
+              (query_runtime_state[key])
           end
         else
           # not found


rmosolgo:
Hi! Thanks for sharing what you found in such a detailed write-up. I've been working hard to find little things to improve in query execution and I'm glad someone _else_ noticed 😆 

I agree that using something other than `public_send` is good here. I guessed that calls to runtime state methods on `query.context...` were pretty rare. Instead, I thought that _writes_ were more common, since the runtime _always_ writes to those values, not knowing if they would be used or not. 

I suspect that adding a `Hash` will make writes slower, since the incoming key must be hashed before the value can be written. One way to reduce that overhead would be to use `compare_by_identity`, but another approach might be to _not_ used a Hash, but instead, use hardcoded method calls instead of public send. 

I tried a patch like this: 

```
class Query
    class Context
      prepend(Module.new do
        def [](key)
          if @scoped_context.key?(key)
            @scoped_context[key]
          elsif @provided_values.key?(key)
            @provided_values[key]
          elsif key == :current_path
            current_path
          elsif key == :current_argument
            query_runtime_state.current_argument
          elsif key == :current_field
            query_runtime_state.current_field
          elsif key == :current_object
            query_runtime_state.current_object
          else
            # not found
            nil
          end
        end

        def query_runtime_state
          (current_runtime_state = Thread.current[:__graphql_runtime_info]) &&
            (query_runtime_state = current_runtime_state[@query])
        end
      end)
    end
  end
```

It removes the `Set#include?` call and the `.public_send`. 

Then, I ran the benchmarks three times: once on v2.1.0 from rubygems, once with the Hash-based patch above, and once with the `elsif ...`-based patch. Here were my local results: 

```
GraphQL v2.1.0
       run benchmark     1.000  i/100ms
       run benchmark      0.911  (± 0.0%) i/s -     28.000  in  30.773497s

GraphQL v2.1.0-with-hash-storage
       run benchmark     1.000  i/100ms
       run benchmark      0.837  (± 0.0%) i/s -     26.000  in  31.073932s
       
GraphQL v2.1.0-with-hardcoded-calls
       run benchmark     1.000  i/100ms
       run benchmark      0.987  (± 0.0%) i/s -     30.000  in  30.421189s
```

So they're all very close, but the hard-coded calls approach was even a _bit_ faster. 

What do you think of an approach like that? 

If we choose it, a couple other places would need updates: 

- https://github.com/rmosolgo/graphql-ruby/blob/2781f4fbe3c6d9e0a1399534aeaf17e070047778/lib/graphql/query/context.rb#L168
- https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/query/context.rb#L186

Then `RUNTIME_METADATA_KEYS` could be removed entirely. What do you think?

swalkinshaw:
Are these benchmarks with YJIT? Ideally we always compare non YJIT with YJIT as well because there can be different characteristics

rmosolgo:
Good call, here are my local results with YJIT enabled: 

```
GraphQL v2.1.0
       run benchmark     1.000  i/100ms
       run benchmark      2.778  (± 0.0%) i/s -     84.000  in  30.259305s

GraphQL v2.1.0-with-hash-storage
       run benchmark     1.000  i/100ms
       run benchmark      2.405  (± 0.0%) i/s -     73.000  in  30.374050s

GraphQL v2.1.0-with-hardcoded-calls
       run benchmark     1.000  i/100ms
       run benchmark      2.924  (± 0.0%) i/s -     88.000  in  30.130306s
```

(Also -- the CI failures are fixed on master, sorry about that!)

rmosolgo:
I tried out that approach here: https://github.com/rmosolgo/graphql-ruby/pull/4634

mtsmfm:
> I suspect that adding a Hash will make writes slower, since the incoming key must be hashed before the value can be written. 

Good point. I agree.

> What do you think of an approach like that?

Unfortunately the code doesn't look cool though, if it's the fastest way I don't agree with it 👍 

Let me close this PR.

PR #4633, opened 2023-09-13
Add specifiedBy default directive

swalkinshaw:
SDL schemas that use a `specifiedBy` directive but _don't_ include the directive definition will fail when using `GraphQL::Schema::BuildFromDefinition`. Because `specifiedBy` is a built-in directive (in the draft spec at least), this should be supported.

This PR just defines the directive and includes it as a default directive.


swalkinshaw:
Ran my tests before adding this to default directives so I'll update these

rmosolgo:
:+1: Thanks for this addition, I didn't realize it had been added to the spec. 

I'm game to add it, but I think we should harmonize it with existing `specifiedByURL` support. I think these should be updated to read/write using directives:

https://github.com/rmosolgo/graphql-ruby/blob/9838bbee177448b91acea197eb48d157a4152bc6/lib/graphql/schema/scalar.rb#L20-L30

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/loader.rb#L145

But this one should work fine as-is: 

https://github.com/rmosolgo/graphql-ruby/blob/9838bbee177448b91acea197eb48d157a4152bc6/lib/graphql/schema/build_from_definition.rb#L306

swalkinshaw:
Should `specified_by_url` even exist on the the scalar base type? Wouldn't the equivalent now be:

```ruby
class MyCustomScalar < GrpahQL::Schema::Scalar
  directive GraphQL::Schema::Directive::SpecifiedBy, url: "http://example.com"
end
```

Guess it depends if you want to keep it for compatibility

rmosolgo:
Yes, agreed it could be migrated to `directive ...`. But I don't know about downstream usage, so could we keep `specified_by_url` as a shorthand for that, for compatibility?

swalkinshaw:
👍  gave it a shot

lib/graphql.rb
```diff
@@ -101,8 +101,8 @@ module EmptyObjects
 require "graphql/pagination"
 require "graphql/schema"
 require "graphql/query"
-require "graphql/types"
 require "graphql/dataloader"
+require "graphql/types"
```
swalkinshaw, line 105: I ran into an issue with `GraphQL::Dataloader::NullDataloader` not being defined at this point due to referencing `Directive` within `Scalar`

lib/graphql/schema/scalar.rb
```diff
@@ -19,9 +19,9 @@ def kind
 
         def specified_by_url(new_url = nil)
           if new_url
-            @specified_by_url = new_url
-          elsif defined?(@specified_by_url)
-            @specified_by_url
+            directive(GraphQL::Schema::Directive::SpecifiedBy, url: new_url)
+          elsif (directive = directives.find { |dir| dir.graphql_name == "specifiedBy" })
+            directive.arguments[:url] # rubocop:disable Development/ContextIsPassedCop
```
swalkinshaw, line 24: `Directive.arguments` is just an attr reader from what I can see but this cop doesn't know that

rmosolgo:
Thanks for this improvement!

PR #4636, opened 2023-09-18
Allow precomputed analyzers to run without visitation

gmac:
This PR adds a new `AST::Analyzer.visit?` hook that allows an analyzer to run without being included in visitation. This is different from the existing `AST::Analyzer.analyze?` hook that stops an analyzer from running entirely. This new hook is useful for analyzers that return precomputed/cached values.

### Why?

Analysis can be expensive, so we may want to cache some analysis results for reuse across repeat requests. For example, complexity scoring seems like it should be easy to reuse:

```ruby
class PrecomputedComplexity < GraphQL::Analysis::AST::QueryComplexity
  def result
    # seems good, but we still do all the visitation work...
    query.context[:precomputed_complexity] || super
  end
end
```

The above example isn't effective because the visitor pass itself is the primary overhead. If this were the only queued analyzer, it would still trigger a visitor pass (and save basically no time).

By adding a `visit?` hook, we can make the analyzer opt out of all visitation work. On its own, the following would not trigger an analysis visitor pass unless necessary:

```ruby
class PrecomputedComplexity < GraphQL::Analysis::AST::QueryComplexity
  def visit?
    query.context[:precomputed_complexity].nil?
  end

  def result
    query.context[:precomputed_complexity] || super
  end
end
```

### DIFF
--- a/lib/graphql/analysis/ast.rb
+++ b/lib/graphql/analysis/ast.rb
@@ -51,22 +51,26 @@ module GraphQL
         query.current_trace.analyze_query(query: query) do
           query_analyzers = analyzers
             .map { |analyzer| analyzer.new(query) }
-            .select { |analyzer| analyzer.analyze? }
+            .tap { _1.select!(&:analyze?) }
 
           analyzers_to_run = query_analyzers + multiplex_analyzers
           if analyzers_to_run.any?
-            visitor = GraphQL::Analysis::AST::Visitor.new(
-              query: query,
-              analyzers: analyzers_to_run
-            )
 
-            visitor.visit
+            analyzers_to_run.select!(&:visit?)
+            if analyzers_to_run.any?
+              visitor = GraphQL::Analysis::AST::Visitor.new(
+                query: query,
+                analyzers: analyzers_to_run
+              )
 
-            if visitor.rescued_errors.any?
-              visitor.rescued_errors
-            else
-              query_analyzers.map(&:result)
+              visitor.visit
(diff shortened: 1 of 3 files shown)


lib/graphql/analysis/ast.rb
```diff
@@ -49,24 +49,27 @@ def analyze_multiplex(multiplex, analyzers)
       # @return [Array<Any>] Results from those analyzers
       def analyze_query(query, analyzers, multiplex_analyzers: [])
         query.current_trace.analyze_query(query: query) do
-          query_analyzers = analyzers
-            .map { |analyzer| analyzer.new(query) }
-            .select { |analyzer| analyzer.analyze? }
+          query_analyzers = analyzers.map { |analyzer| analyzer.new(query) }
+          query_analyzers.select!(&:analyze?)
```
gmac, line 53: This should net -1 array allocs in a pretty active code path.

lib/graphql/analysis/ast.rb
```diff
@@ -51,22 +51,26 @@ def analyze_query(query, analyzers, multiplex_analyzers: [])
         query.current_trace.analyze_query(query: query) do
           query_analyzers = analyzers
             .map { |analyzer| analyzer.new(query) }
-            .select { |analyzer| analyzer.analyze? }
+            .tap { _1.select!(&:analyze?) }
```
gmac, line 54: This should net -1 array allocs.

PR #4637, opened 2023-09-18
Tidy chained array transforms

gmac:
Cleans up some pedantic details that reminded me of a Shopify Readme titled **"Ruby code that looks silly to a low-level programmer"**...

> 3. _Always_ use inline operations. Which means, prefer the use of `map! {...}` versus `map {...}`. Without the `!`, the interpreter will create a new Array. Only use the `map{...}` if the creation of a new array is desired. The same applies for other methods like `select!`, `filter!`, `rotate!`, `reverse!` and etc.

This is generally a pretty easy pattern to spot: when chaining multiple array transforms, all the later transforms should inline.

### DIFF
--- a/lib/graphql/analysis/ast.rb
+++ b/lib/graphql/analysis/ast.rb
@@ -74,7 +74,7 @@ module GraphQL
       end
 
       def analysis_errors(results)
-        results.flatten.select { |r| r.is_a?(GraphQL::AnalysisError) }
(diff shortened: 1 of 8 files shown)


swalkinshaw:
`flat_map` and `filter_map` don't have inline versions but there might be a few places where those are more efficient

rmosolgo:
Cool, thanks!

Issue #4640, opened 2023-09-19
C Parser is crashing with ruby-head debug builds


casperisfine:
**Describe the bug**

We run our application test suite against Ruby nightly builds and recently we started seeing a crash on Ruby-head debug builds (built with `RUBY_DEBUG=1`)

```
/tmp/bundle/ruby/3.3.0+0/gems/graphql-c_parser-1.0.5/lib/graphql/c_parser.rb:96: [BUG] Segmentation fault at 0x00007f1bd084ea18
ruby 3.3.0dev (2023-09-19T14:05:51Z shopify 5c5391f444) +YJIT [x86_64-linux]

-- Ruby level backtrace information ----------------------------------------
....
/bundle/ruby/3.3.0+0/gems/graphql-2.0.26/lib/graphql/query.rb:80:in `selected_operation_name'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-2.0.26/lib/graphql/query.rb:245:in `selected_operation'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-2.0.26/lib/graphql/query.rb:467:in `with_prepared_ast'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-2.0.26/lib/graphql/query.rb:407:in `prepare_ast'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-2.0.26/lib/graphql.rb:46:in `parse'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-c_parser-1.0.5/lib/graphql/c_parser.rb:11:in `parse'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-c_parser-1.0.5/lib/graphql/c_parser.rb:95:in `result'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-shopify-0.9.0/lib/graphql/shopify/trace.rb:21:in `parse'
...
/tmp/bundle/ruby/3.3.0+0/gems/graphql-c_parser-1.0.5/lib/graphql/c_parser.rb:96:in `block in result'
/tmp/bundle/ruby/3.3.0+0/gems/graphql-c_parser-1.0.5/lib/graphql/c_parser.rb:96:in `c_parse'

-- Threading information ---------------------------------------------------
Total ractor count: 1
Ruby thread count for this ractor: 6

-- Machine register context ------------------------------------------------
[...]

PR #4641, opened 2023-09-20
graphql-c_parser: Properly register global variables

casperisfine:
Fix: https://github.com/rmosolgo/graphql-ruby/issues/4640

`GraphQL_Language_Nodes_` are global variables and their value is assigned the return value of `const_get` which contrary to `rb_define_class_*` isn't pinned.

So they must be registered with `rb_global_variable`.

On a side note, `rb_global_variable` must be called before the variable is assigned, not after.

### DIFF
--- a/graphql-c_parser/ext/graphql_c_parser_ext/parser.c
+++ b/graphql-c_parser/ext/graphql_c_parser_ext/parser.c
@@ -3272,19 +3272,21 @@ void yyerror(VALUE parser, VALUE filename, const char *msg) {
   rb_exc_raise(exception);
 }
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for fixing this! I thought, since `GraphQL_Language_Nodes_...` were referencing constants that were already known in Ruby-land, that I wouldn't need to register them again with Ruby. `NONE` is an exception, but I guess I had the calls out of order in that case, anyways. Thanks for the fix!

rmosolgo:
released in graphql-c_parser 1.0.7

casperisfine:
> that I wouldn't need to register them again with Ruby

Yeah, that's the tricky part here. You don't need to register them again to prevent them from being garbage collected because they are referenced elsewhere.

But on our CI we compact the heap with `GC.compact`, so references that are not properly registered become stale and point to garbage.

PR #4644, opened 2023-09-21
Support re-printing AST extensions

### LINKED ISSUE #4635: `GraphQL::Schema.from_definition(sdl).to_definition` drops `extend` keyword when it should not
**Describe the bug**

When I have an SDL string with a schema extension, the `extend` keyword gets lost when I parse and dump the SDL using `GraphQL::Schema.from_definition(sdl).to_definition`.  In contrast, it is properly retained when using `GraphQL::Language::Printer.new.print(GraphQL.parse(sdl))`.

**Versions**

`graphql` version: `2.1.0`.

**GraphQL schema**

```graphql
  extend schema
    @customDirective

  directive @customDirective repeatable on SCHEMA

  type Query {
    foo: Int
  }
```

**GraphQL query**

N/A

**Steps to reproduce**

Put this Ruby script into a file named `extend_bug.rb`:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", "2.1.0"
end

require "graphql"

schema = <<~EOS
  extend schema
    @customDirective

  directive @customDirective repeatable on SCHEMA

  type Query {
    foo: Int
  }
EOS

puts "Parsing and print the schema:"
puts "=" * 60
puts GraphQL::Language::Printer.new.print(GraphQL.parse(schema))

puts
puts

puts "Roundtripping the schema definition:"
puts "=" * 60
puts GraphQL::Schema.from_definition(schema).to_definition
```

Run it to observe the bug.  Here's the output I get:

```
Parsing and print the schema:
============================================================
extend schema
  @customDirective

directive @customDirective repeatable on SCHEMA

type Query {
  foo: Int
}


Roundtripping the schema definition:
============================================================
schema
  @customDirective

directive @customDirective repeatable on SCHEMA

type Query {
  foo: Int
}
```

**Expected behavior**

I expect the schema printed via `GraphQL::Schema.from_definition(sdl).to_definition` to include the `extend` keyword since my schema definition has an extension (and roundtripping it through `parse` + `print` includes it!).

**Actual behavior**

The `extend` keyword is omitted.

**Additional context**

This matters because I've found that other GraphQL SDL parsers give me errors when the `extend` keyword is missing in this case.

It looks like `extend schema` support was added in #4516. I'm not sure why it's not working through `GraphQL::Schema.from_definition(sdl).to_definition`.

rmosolgo:
Fixes #4635

TODO: 

- [x] Document `from_ast: true` for `to_definition` 
- [x] DRY `extend_` methods 
- ~~Support extending default scalars?~~ This turned out to be really tricky because schemas get the built-in `@skip` and `@include` directives by default, and those include the built-in String and Boolean types. But then if the schema has `extend scalar String ...`, we don't want to modify the built-in string because that might mess up another schema. So we can make a duplicate, but it will end up clashing with the built-in one which the schema already has. (Not exactly _clashing_, but it's a duplicate type with no way to choose between them, so it doesn't _work_...) 

  If it turns out someone needs that later, it could _probably_ be added somehow, but we'd have to go back and replace the built-in directives with custom ones, etc.
  
  
Along the way I also added a benchmark for printing to SDL and improved it: 

```diff
  $ be rake bench:profile_to_definition 
  Calculating -------------------------------------

-        to_definition      2.568k (± 2.3%) i/s -     13.050k in   5.084553s
+       to_definition      2.774k (± 6.4%) i/s -     13.851k in   5.015514s

- Total allocated: 43037 bytes (336 objects) 
+ Total allocated: 29229 bytes (171 objects)
```

About half as many objects and 8% faster

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -120,6 +120,13 @@ namespace :bench do
     GraphQLBenchmark.profile_small_result
   end
 
+  desc "Dump schema to SDL"
+  task :profile_to_definition do
+    prepare_benchmark
+    GraphQLBenchmark.profile_to_definition
+  end
+
+
   desc "Compare GraphQL-Batch and GraphQL-Dataloader"
   task :profile_batch_loaders do
     prepare_benchmark
--- a/benchmark/run.rb
+++ b/benchmark/run.rb
@@ -418,6 +418,26 @@ module GraphQLBenchmark
     GRAPHQL
   end
 
+  def self.profile_to_definition
+    require_relative "./batch_loading"
+    schema = BatchLoading::GraphQLNoBatchingSchema
+
+    Benchmark.ips do |x|
+      x.report("to_definition") { schema.to_definition }
+    end
+
+    result = StackProf.run(mode: :wall, interval: 1) do
+      schema.to_definition
+    end
+    StackProf::Report.new(result).print_text
+
+    report = MemoryProfiler.report do
+      schema.to_definition
+    end
+
+    report.pretty_print
+  end
+
   def self.profile_batch_loaders
     require_relative "./batch_loading"
     include BatchLoading
--- a/lib/graphql/language/document_from_schema_definition.rb
+++ b/lib/graphql/language/document_from_schema_definition.rb
@@ -283,9 +283,13 @@ module GraphQL
       def build_type_definition_nodes(types, type_nodes, type_extensions_nodes)
         types.each do |type|
           if !include_introspection_types && type.introspection?
-            next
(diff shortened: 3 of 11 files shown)


rmosolgo:
It turns out that most of this work _wasn't_ what was requested in #4635, but I had misunderstood the issue. So I pulled the actual fix and the performance work into #4647 . I'm going to close this because I don't know that anyone actually _wants_ `from_ast: true`, but we can revive this work if it's needed later.

PR #4654, opened 2023-10-03
Fix returning hash default values with implicit keys

### LINKED ISSUE #4649: Inconsistent return value when hash_key is specified for an array field
**Describe the bug**

When a Hash with default values is passed to the object, depending on if `hash_key` is specified explicitly or implicitly the return of array field is either `nil` or `[]`.

**Versions**

The inconsistency was introduced in 2.0.8 and is there until latest version (2.1.0)

**Minimal reproduction**

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'graphql', '2.0.8'
end

class Example < GraphQL::Schema::Object
  field :nil_when_empty, [String, null: true], null: true
  field :array_when_empty, [String, null: true], null: true, hash_key: :nonexistent
end

  field :example, Example, null: false

  def example
    Hash.new { [] }
  end
end

  query(Query)
end

p AppSchema.execute('query { example { nilWhenEmpty, arrayWhenEmpty } }').to_h
```

**Steps to reproduce**

1. Try the above code with `graphql` gem set to lower version (2.0.7, 1.9.0, you name it)
2. Try the above code with `graphql` gem set to affected version (2.0.8, 2.1.0, etc)

**Expected behavior**

The behaviour is consistent regardless of if `hash_key` is set implicitly or explicitly

**Actual behavior**

One of the fields returns `nil`, another returns `[]`

**Additional context**

Discovered when upgrading the library because we use `.errors.messages.to_h` from ActiveRecord and the Hash returned from this method has the default value set to `[]`

rmosolgo:
Fixes #4649

### DIFF
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -700,16 +700,22 @@ module GraphQL
                   obj.public_send(resolver_method)
                 end
               elsif inner_object.is_a?(Hash)
-                if @dig_keys
+                hash_value = if @dig_keys
                   inner_object.dig(*@dig_keys)
                 elsif inner_object.key?(@method_sym)
                   inner_object[@method_sym]
                 elsif inner_object.key?(@method_str)
                   inner_object[@method_str]
-                elsif @fallback_value != NOT_CONFIGURED
-                  @fallback_value
                 else
-                  nil
+                  # Maybe the Hash will provide a default value, even though `.key?` was false:
+                  v = inner_object[@method_sym] || inner_object[@method_str]
+                  # Technically, this could still be wrong, if the hash-provided default value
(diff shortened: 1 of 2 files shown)


d4rky-pl:
@rmosolgo [you can detect if Hash has a default value by checking if `.default_proc` is `nil?`](https://ruby-doc.org/core-2.5.1/Hash.html#method-i-default_proc)

```
irb(main):003:0> without_default = Hash.new
=> {}
irb(main):004:0> with_default = Hash.new { 1 }
=> {}
irb(main):005:0> without_default.default_proc.nil?
=> true
irb(main):006:0> with_default.default_proc.nil?
=> false
```

PR #4656, opened 2023-10-05
Fix returning hash default values with implicit keys

d4rky-pl:
@rmosolgo I copied your spec from #4654 directly and applied my solution for the problem, the spec passed locally


rmosolgo:
Thanks for this suggestion -- I agree it's a lot better :+1: 

Fixes #4649 

Fixes #4654

Issue #3061, opened 2020-08-04
Overriding methods in GraphQL::Schema post 1.10. Breaking workaround for max_depth on introspection.


jhirn:
Hi @rmosolgo. I used to work around the Introspection query max_depth (from this issue #2437) by implementing in my Schema, but upgrading from 1.9.19 to the anything past 1.10.0 it no longer works. 

```ruby
def as_json(only: nil, except: nil, context: {})
    execute(GraphQL::Introspection::INTROSPECTION_QUERY, only: only, except:
      except, context: context, max_depth: 13).to_h
  end
```

In fact, I can't seem to override any methods in the schema object. 

Putting a pry for execute or as_json it never gets called. I'm not sure what the proper way to upgrade my Schema definition. I've used Instrumentation to work around the inability to override `execute`, but not sure from the change log exactly what is going on. I'm sure there's a way but it's unclear after perusing the documents.

rmosolgo:
Hi, great question! 

In 1.9.x, a lot of `Schema`'s methods were delegated to a singleton of the instance. However, in 1.10.x, those methods were re-implemented directly on the schema class. 

I think you could continue overriding those by moving the definition to the class, for example: 

```diff 
-  def as_json(only: nil, except: nil, context: {})
+  def self.as_json(only: nil, except: nil, context: {})
     execute(GraphQL::Introspection::INTROSPECTION_QUERY, only: only, except:
       except, context: context, max_depth: 13).to_h
   end
```

In short, this switch is part of moving graphql-ruby internals _away_ from the old `.define { ... }`-based APIs to class-based APIs. 

Does that work for you?

jhirn:
🙏 thx for the response. I will try and report back.

jhirn:
So `def self.execute` definitely works for me. I'm actually able to remove my Instrumentation class and hit a pry. Unfortunately, `def self.as_json` not hitting thje pry or setting the max_depth. I'm not too concerned with bumping up max_depth temporarily, but would love to have this working like before. Here is my entire Schema object for you 


```ruby
class MySchema < GraphQL::Schema
  max_depth 8
  mutation(Mutations::MutationType)
  query(Types::QueryType)

  use GraphQL::Execution::Interpreter
  use GraphQL::Analysis::AST

  # Add built-in connections for pagination
  use GraphQL::Pagination::Connections

  def self.execute(query, **args)
    super
  end

  def self.to_definition(only: nil, except: nil, context: {})
[...]

PR #4658, opened 2023-10-09
Add max_depth ... count_introspection_fields: false

rmosolgo:
Addressing issues as mentioned in #2437 and #3061


jhirn:
I'm not using this library anymore (big sad) but happy to see an old issue get cleaned up. Thanks Robert!

PR #4659, opened 2023-10-10
Bump required_ruby_version to 2.7

swalkinshaw:
`required_ruby_version` should be `2.7` because numbered parameters are used for blocks now.

Also I updated the Ruby versions in the CI workflow to be consistent throughout


rmosolgo:
Thanks for the update!

PR #4639, opened 2023-09-19
Reference correct variable in #merge-into

jahfer:
👋🏼 Hey @rmosolgo. I don't think we've crossed paths since the Batman.js days :)

I'm in the process of writing a new directive, and got hit by this little bug in testing. It looks like it was introduced as part of https://github.com/rmosolgo/graphql-ruby/pull/4425, which was merged relatively recently.

PR #4660, opened 2023-10-11
Include patch from #4639 in graphql_result.rb

jahfer:
We just updated one of our core Shopify apps to 2.1.1 yesterday, and after rebasing I got hit with the same error that https://github.com/rmosolgo/graphql-ruby/pull/4639 intended to fix. 

Due to some bad timing, https://github.com/rmosolgo/graphql-ruby/pull/4621/commits/cf081056e0c9a73f1492d57b5fd09a45855842da —which was authored before my patch—was merged the day _after_ https://github.com/rmosolgo/graphql-ruby/pull/4639, losing the change I contributed.

This PR includes the same change, but for the new `graphql_result.rb` file :D

Is it possible to cut a 2.1.2 with this patch included? It would be much appreciated! 🙏🏼

### DIFF
--- a/lib/graphql/execution/interpreter/runtime/graphql_result.rb
+++ b/lib/graphql/execution/interpreter/runtime/graphql_result.rb
@@ -107,7 +107,7 @@ module GraphQL
               when GraphQLResultArray
                 # There's no special handling of arrays because currently, there's no way to split the execution
                 # of a list over several concurrent flows.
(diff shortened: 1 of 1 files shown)


rmosolgo:
Sorry about the botched merge! Thanks for resubmitting this. I'll put 2.1.2 on RubyGems right now.

PR #4679, opened 2023-10-20
Support a custom Schema.query_class

### LINKED ISSUE #4354: Configurable `Query` class to be used for operation evaluation

Similar to [`context_class`](https://github.com/rmosolgo/graphql-ruby/blob/15fec5f50da50f990571be51f0e8e6e013cbaee6/lib/graphql/schema.rb#L731-L737) on `Schema`, allow for a custom `Query` class to be defined and used as part of the execution process.

I have a workaround where I override `multiplex` to use a custom subclass of `Query`.

```ruby
def multiplex(queries, **kwargs)
  queries = queries.map do |query|
    case query
    when Hash
      query_class.new(self, nil, **query)
    else
      query
    end
  end
  super(queries, **kwargs)
end
```

rmosolgo:
Fixes #4354

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -680,12 +680,20 @@ module GraphQL
         else
           string_or_document
         end
-        query = GraphQL::Query.new(self, document: doc, context: context)
+        query = query_class.new(self, document: doc, context: context)
         validator_opts = { schema: self }
         rules && (validator_opts[:rules] = rules)
         validator = GraphQL::StaticValidation::Validator.new(**validator_opts)
         res = validator.validate(query, timeout: validate_timeout, max_errors: validate_max_errors)
         res[:errors]
+      end
+
+      def query_class(new_query_class = NOT_CONFIGURED)
+        if NOT_CONFIGURED.equal?(new_query_class)
+          @query_class || (superclass.respond_to?(:query_class) ? superclass.query_class : GraphQL::Query)
+        else
+          @query_class = new_query_class
+        end
       end
 
       attr_writer :validate_max_errors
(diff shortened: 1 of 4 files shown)


lib/graphql/schema.rb
```diff
@@ -680,14 +680,22 @@ def validate(string_or_document, rules: nil, context: nil)
...
         validator_opts = { schema: self }
         rules && (validator_opts[:rules] = rules)
         validator = GraphQL::StaticValidation::Validator.new(**validator_opts)
         res = validator.validate(query, timeout: validate_timeout, max_errors: validate_max_errors)
         res[:errors]
       end
 
+      def query_class(new_query_class = NOT_CONFIGURED)
+        if NOT_CONFIGURED.equal?(new_query_class)
+          @query_class || (superclass.respond_to?(:query_class) ? superclass.query_class : GraphQL::Query)
```
ravangen, line 693: Is it worth remembering this outcome?

rmosolgo: That's a big question all around this library... what should happen in cases like: 

```ruby 
BaseSchema.query_class = CustomQuery1 
ChildSchema = Class.new(BaseSchema)
ChildSchema.query_class # CustomQuery1 
# Then: 

BaseSchema.query_class = CustomQuery2 
ChildSchema.query_class # What should this return?
```

I'd _expect_ it to return the inherited value, `CustomQuery2`, not the _previously_ inherited value, `CustomQuery1`. 

On the one hand, I wouldn't expect situations like this to happen often, but on the other hand, with Rails autoloading, and with Ruby being Ruby, I'd want to stay _open_ to it! 

So that's why I didn't memoize the result. What do you think?

ravangen: Fair point, probably not worth it at this time.

Our internal implementation has `class_attribute :query_class, default: GraphQL::Shopify::Query` currently (we were ok with the dependency on `activesupport`)

PR #4682, opened 2023-10-23
Update authorization.md to include mutation#authorized?

fwolfst:
Mention both  `#authorized?` methods of Mutations. The instance method lacks explanation, but at least its mentioned now.

### DIFF
--- a/guides/authorization/authorization.md
+++ b/guides/authorization/authorization.md
@@ -16,7 +16,7 @@ Schema members have `authorized?` methods which will be called during execution:
 - Type classes have `.authorized?(object, context)` class methods
 - Fields have `#authorized?(object, args, context)` instance methods
 - Arguments have `#authorized?(object, arg_value, context)` instance methods
-- Mutations have `#authorized?(object, context)` class method and a `#authorized?(hash)` instance method
+- Mutations and Resolvers have `.authorized?(object, context)` class methods and `#authorized?(args)` instance methods
 
 These methods are called with:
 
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hey, this is described in the mutation authorization guide: https://graphql-ruby.org/mutations/mutation_authorization.html#can-this-user-perform-this-action

How about linking to that guide instead? (Or improving that guide?)

fwolfst:
> How about linking to that guide instead? (Or improving that guide?)

That makes sense, I will update the PR.

fwolfst:
Updated the PR. So, mutation has `ready?`, but also `self.authorized?`. Does it make sense to mention the latter (second commit)? Iiuc, it is executed before any of the other mentioned methods.

guides/authorization/authorization.md
```diff
@@ -16,6 +16,7 @@ Schema members have `authorized?` methods which will be called during execution:
 - Type classes have `.authorized?(object, context)` class methods
 - Fields have `#authorized?(object, args, context)` instance methods
 - Arguments have `#authorized?(object, arg_value, context)` instance methods
+- Mutations have `#authorized?(object, context)` class method and a `#authorized?(hash)` instance method
```
rmosolgo, line 19:
```suggestion
- Mutations and Resolvers have `.authorized?(object, context)` class methods and `#authorized?(args)` instance methods
```

rmosolgo:
Thanks for this improvement -- sorry it took me so long to get back to it!

PR #3637, opened 2021-10-06
Add ISO8601Duration type

jessecurry:
Adds an `ISO8601Duration` type in the style of the existing `ISO8601Date` and `ISO8601DateTime`

Issue #3636, opened 2021-10-06
Add Support for ISO8601 Durations


jessecurry:
Rails 6.2 changes the default duration type to `ActiveSupport::Duration`, which has an `#iso8601` method.

GraphQL Ruby already supports `ISO8601Date` and `ISO8601DateTime`, I'd like to see duration added to the library.

This is a first pass class largely pulled from the existing ISO8601 types that I'm using in a project:

```
# frozen_string_literal: true

module Types
  # This scalar takes `Durations`s and transmits them as strings,
  # using ISO 8601 format.
  #
  # Use it for fields or arguments as follows:
  #
  #     field :duration, Types::ISO8601Duration, null: false
  #     argument :duration, Types::ISO8601Duration, null: false
  #
  # Add an inflection to your initializers to properly pick
  # up this class name:
  #
  # ActiveSupport::Inflector.inflections do |inflect|
  #   inflect.acronym 'ISO'
  # end
  #
  # If you're using Rubocop you may need to add this config:
  #
  # RSpec/FilePath:
  #   CustomTransform:
  #     ISO8601Duration: iso_8601_duration
  class ISO8601Duration < GraphQL::Schema::Scalar
    description 'An ISO 8601-encoded duration'

    # @param value [Duration, String]
    # @return [String]
    def self.coerce_result(value, _ctx)
      case value
      when ActiveSupport::Duration
        value.iso8601
      when ::String
        ActiveSupport::Duration.parse(value).iso8601
      else
        value.try(:iso8601)
      end
    rescue StandardError => e
      raise GraphQL::Error, <<-ERROR
            An incompatible object (#{value.class}) was
            given to #{self}. Make sure that only Durations
            and well-formatted Strings are used with this
            type. (#{e.message})"
      ERROR
    end

    # @param str_value [String]
    # @return [ActiveSupport::Duration]
    def self.coerce_input(str_value, _ctx)
      ActiveSupport::Duration.parse(str_value)
    rescue ArgumentError, TypeError
      # Invalid input
      nil
    end
  end
end
```

jessecurry:
Opened #3637 with this code for discussion.

rmosolgo:
If anyone wants to open a PR with a change like this, I'd welcome it! But since I don't have plans to discuss this further or work on it myself, I'll  close this issue.

PR #4688, opened 2023-10-26
Add ISO8601Duration scalar type

rubyroobs:
## Changelog

* adds `ISO8601Duration` Scalar
  * raises `GraphQL::Error` when `ActiveSupport::Duration` is not defined
  * added reference in the `Scalar types` guide
* adds `DurationEncodingError` for when an improperly encoded duration is passed as an input
* add tests for `ISO8601Duration`

## Context

This adds support for ISO8601 durations for projects using Rails/ActiveSupport. Implements https://github.com/rmosolgo/graphql-ruby/pull/3637

Reimplemented what was done in https://github.com/rmosolgo/graphql-ruby/issues/3636 with the suggested changes/feedback from the previous PR taken into account.


rmosolgo:
This looks great. Thanks for the contribution!

PR #4691, opened 2023-10-31
Fix ActionCable subscriptions channel example

zavan:
Query is the first positional argument for `Schema#execute`, not a kwarg.

At least 2 instances of users trying to use the example and bumping into cryptic errors, [this guy](https://stackoverflow.com/questions/66877742/ruby-graphql-and-or-actioncable-dumping-core-on-subscription) and myself ha.


rmosolgo:
Thanks for correcting it!

PR #4692, opened 2023-10-31
Don't strip Type from the end of type names

chughes87:
I've found this behavior of the library to be strange. Maybe there is a concrete reason for doing this, but I would like it if this aspect could be removed.


rmosolgo:
Hi! Thanks for the suggestion. I admit, I'm a bit rusty on how these lines of code come into play. Could you please share some info about what you were trying to, how it went wrong, and how this change would address it?

PR #4710, opened 2023-11-21
Try rewriting with Fiber.schedule and the Async gem

rmosolgo:
Continuing to simplify and try to work with the dataloader. 

This branch is an attempt at getting it to work with `Async::Scheduler` and `Fiber.schedule { ... } `, ignoring other tests:

```
be rake test TEST=spec/graphql/dataloader/async_dataloader_spec.rb
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,9 @@ if RUBY_VERSION >= "3.0"
   gem "evt"
 end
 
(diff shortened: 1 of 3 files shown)


rmosolgo:
I've got something better going in #4727

Issue #4715, opened 2023-11-30
`prepare: :symbol` syntax looks on the parent object for resolver classes


eapache-opslevel:
**Describe the bug**

If you define a mutation or resolver with an argument that uses the `prepare: :method_name` syntax, it looks for that method on the parent object instead of on the mutation/resolver.

**Versions**

`graphql` version: 2.0.27
`rails` (or other framework): 7.1.2

**GraphQL schema**

```ruby
class MyMutation < GraphQL::Schema::Mutation
  argument :foo, ID, required: true, prepare: :some_method
  class << self
    def some_method(in)
      123
    end
  end
end

class MutationRoot < GraphQL::Schema::Object
  field :my_mutation, mutation: MyMutation
end
```

**Behavior**

I would expect `foo` to be prepared by calling `some_method` on `MyMutation` but at https://github.com/rmosolgo/graphql-ruby/blob/109c26c5fa9118a9c69ed4c4823048c1ce1aea6d/lib/graphql/schema/argument.rb#L225 it looks on the object, which is `MutationRoot` (and thus hits `NoMethodError`).

eapache-opslevel:
It might be as simple as switching that call to use `owner` instead? But I'm not sure if that is correct in all cases either, we might have to type-check and switch between `owner` and `obj`?

eapache-opslevel:
> I wonder if there's a way to make this work as expected while preserving a way for existing users to maintain compatibility
Try the resolver-defined-method and if it's not defined fall back to the object-defined-method? Kind of like how we look for resolve methods?

DougEdey:
I took a stab at this since I'm the one that spoke to Evan about it originally, lemme know if I misinterepreted the proposal!

PR #4717, opened 2023-11-30
Add support for class level prepare methods on arguments

DougEdey:
Issue: https://github.com/rmosolgo/graphql-ruby/issues/4715

I _think_ this is what @eapache-opslevel was aluding to, it checks to see if the prepare method exists on the object or the owner (class method), and handles each approproiately

* For Object, don't send the context, it'll be there already
* For the Owner, send the context (like we do already for prepare blocks)
* If neither exist, give a clear error instead of a no method found error.

### DIFF
--- a/spec/support/jazz.rb
+++ b/spec/support/jazz.rb
@@ -3,7 +3,7 @@
 # Here's the "application"
 module Jazz
   module Models
-    Instrument = Struct.new(:name, :family, :alternative_name)
+    Instrument = Struct.new(:name, :family)
     Ensemble = Struct.new(:name)
     Musician = Struct.new(:name, :favorite_key)
     Key = Struct.new(:root, :sharp, :flat) do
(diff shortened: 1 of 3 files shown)


spec/support/jazz.rb
```diff
@@ -3,7 +3,7 @@
 # Here's the "application"
 module Jazz
   module Models
-    Instrument = Struct.new(:name, :family)
+    Instrument = Struct.new(:name, :family, :alternative_name)
```
eapache-opslevel, line 6: I'm not sure what this change does?

spec/support/jazz.rb
```diff
@@ -544,11 +544,17 @@ class EnsembleInput < GraphQL::Schema::InputObject
   end
 
   class AddInstrument < GraphQL::Schema::Mutation
+    class << self
+      def prepare_family(value, context)
+        value.upcase
```
eapache-opslevel, line 549: Slightly concerned that you didn't have to change an expectation anywhere to expect the upcased value?

eapache-opslevel (commented):
This makes sense based on my understanding here. Left a few questions about the tests though,

rmosolgo:
Thanks for this improvement!

PR #4720, opened 2023-12-01
Default to re-authorizing scoped items (as documented), fix skipping for connections

rmosolgo:
Oops -- the default behavior here didn't match the docs. I must have missed it when I switched the default to be opt-in 😖 

Since the implementation checked if `scope_items` returned a new object or not, this would have only affected people who implemented `scope_items` to return a new object and _didn't_ want that new object to be re-authorized. 

Also, I fixed the behavior for connections so that they get their `reauthorize_scoped_objects` config from the node type.


bmulholland:
Thanks for the fix, really appreciate it :)

PR #4727, opened 2023-12-12
Build a Dataloader on Async gem primitives

rmosolgo:
This is another attempt to get Dataloader to "Just Work" when it comes to running `Source#fetch` in the background. 

This approach uses the `async` gem to handle backgrounding, using `Async do ... end`. 

Amazingly, it worked out alright. I really don't know _how_ `Dataloader#yield` just _worked_, but it did, by calling through to `Async::Task.current.yield`. 

TODO: 

- [x] Sadly CI doesn't seem to like this. But It passes locally....
- [x] Test: during GraphQL execution, does it run resolvers in parallel? (It might, since it uses `Async { ... }.wait`, but I don't have a test for it at the moment. 
  - This was already covered in the existing tests
- [x] Try a Ruby on Rails example app and make sure this will deliver as promised with ActiveRecord, Redis, and HTTP calls inside a Source.
  - Demo: https://github.com/rmosolgo/rails-graphql-async-demo
- [x] Update the documents to recommend this implementation (remove references to other schedulers)


trevorturk:
This is a bit in the weeds for me to truly review, but I gave things a read through and it looks good! I'm happy you've been able to make some progress here, and excited for the Async-ified Dataloader!

I see your todos in the PR description, and figured I'd share a bit of code I'm using in a Rails app running Falcon/Async and doing a lot of Async::HTTP work: https://gist.github.com/trevorturk/f8baeaf8a954bc3764d70652db1692fa -- one bit is around testing that things are running in parallel and the other is a fix for integration tests.  

I'll keep an eye on this PR and report back if I have anything useful to share. Feel free to ping me if you have any questions I might be able to help with.

I'll also @ioquatix just in case he has the time/interest to put his eyes on this.

rmosolgo:
I tried it in Rails and it worked like a charm! Net::HTTP and ActiveRecord ran in parallel without a problem. Demo: https://github.com/rmosolgo/rails-graphql-async-demo

ioquatix:
You can visualise what is going on using the `async-debug` gem which is one line to drop into a test application (usually).

rmosolgo:
I think I'm all set, application-wise. It's just CI -- it runs for hours but gets nowhere: https://github.com/rmosolgo/graphql-ruby/actions/runs/7196641771/job/19601940761

Actually, the Rails build (sometimes?) gets somewhere, but not too far: https://github.com/rmosolgo/graphql-ruby/actions/runs/7196641771/job/19601942904

I'm going to dig in and straighten it out tomorrow.

ioquatix:
It’s not uncommon for CI to expose race conditions because scheduling is so much slower and thus unpredictable. I would suggest looking at scheduling operations.

ioquatix:
I should add, tests that depend on `sleep` actually sleeping the desired amount of time, and `sleep N` finishing faster than `sleep N+M` may break. It's extremely tricky to write those sorts of tests on the CI as the actual vCPU seems incredibly unpredictable (and overloaded).

spec/graphql/dataloader/async_dataloader_spec.rb
```diff
@@ -1,13 +1,16 @@
...
+if RUBY_VERSION >= "3.1.1"
+  require "async"
   describe GraphQL::Dataloader::AsyncDataloader do
     class AsyncSchema < GraphQL::Schema
       class SleepSource < GraphQL::Dataloader::Source
         def fetch(keys)
           max_sleep = keys.max
-          `sleep #{max_sleep}`
+          # t1 = Time.now
+          # puts "----- SleepSource => #{max_sleep} (from: #{keys})"
```
nunommc, line 11: clean up :)

spec/graphql/dataloader/nonblocking_dataloader_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+if Fiber.respond_to?(:scheduler) # Ruby 3+
+  describe GraphQL::Dataloader::NonblockingDataloader do
+    class NonblockingSchema < GraphQL::Schema
+      class SleepSource < GraphQL::Dataloader::Source
+        def fetch(keys)
+          max_sleep = keys.max
+          # t1 = Time.now
+          # puts "----- SleepSource => #{max_sleep} "
+          sleep(max_sleep)
+          # puts "----- SleepSource done #{max_sleep} after #{Time.now - t1}"
```
nunommc, line 13: clean up :)

spec/graphql/dataloader/nonblocking_dataloader_spec.rb
```diff
@@ -0,0 +1,261 @@
...
+          keys.map { |_k| max_sleep }
```
nunommc, line 14: is this producing an array with same size of keys and with max_sleep in all positions?

Following this approach might be more clear what's doing:
```ruby
 Array.new(keys.size, max_sleep)
 ```

(Not super clear to me what this array will be needed for)

File: spec/graphql/dataloader_spec.rb [hunk 4]
```diff
@@ -893,6 +901,16 @@ def self.included(child_class)
...
 
+  # describe "AsyncDataloader" do
+  #   let(:schema) {
+  #     Class.new(FiberSchema) {
+  #       use GraphQL::Dataloader::AsyncDataloader
+  #     }
+  #   }
+
+  #   include DataloaderAssertions
+  # end
```
nunommc, line 912: remove?

spec/spec_helper.rb
```diff
@@ -29,6 +29,8 @@
   Minitest::Reporters.use! Minitest::Reporters::DefaultReporter.new(color: true)
 end
 
+# Minitest::Reporters.use! Minitest::Reporters::SpecReporter.new(color: true)
+
```
nunommc, line 33: revert?

spec/graphql/dataloader/async_dataloader_spec.rb
rmosolgo: I have circled back enough times on this that I think I'd rather just keep it here for next time I need it 😅

rmosolgo: I have circled back enough times on this that I think I'd rather just keep it here for next time I need it 😅

rmosolgo:
🎉  

It was scheduling code. 

Previously I was relying on loops going round and round to eventually let everything resolve. But I had a test to make sure that the dataloader loop didn't exceed 1000 iterations. It _was_ -- in fact, it was neverending -- so the test suite never finished. 

I added an `Async::Condition` instead, so that Tasks which are waiting for some other data can wait on that condition, then when a dataloader loop is finished, it notifies the condition so that those tasks can start again. 

@ioquatix, this is hot stuff! Thanks for this great library. I look forward to learning more, for example, applying semaphores to the internal list of dataloader jobs if the need arises (it hasn't _yet_...).

rmosolgo:
Amazing. @ioquatix and @trevorturk, thanks again for your help along the way on this!

ioquatix:
Nice work team!

jhass:
Hey, did anyone try to run this on falcon yet? I didn't try it in a minimal example but to me it seems given `AsyncDataloader` doesn't override `spawn_fiber` to spawn an async task rather than a regular fiber, that causes the `Sync` down the line to spawn a new reactor (since there's no current task in that regular fiber), which then proceeds to replace the reactor falcon spawned on the current thread killing all async tasks running on it (=> tons of Async::Stop exceptions).

I'm probably doing something else wrong but...

rmosolgo:
👋  I never tried it out on Falcon and I don't know much about how Falcon works. But if you're willing to work up an example app that demonstrates the issue you're describing, I'd be happy to work on improving compatibility!

jhass:
@rmosolgo thanks! It seems to be as simple as https://github.com/jhass/rails-graphql-falcon to reproduce, or at least this also crashes although slightly differently than what I'm seeing in my real world app.

jhass:
I took a stab at a compatible implementation, no tests with real world traffic yet but it seems to work both for the simple example above and in local tests with a more complex application and query.

Maybe it can serve as a base for an improved version in the gem? 

https://github.com/jhass/rails-graphql-falcon/compare/main...async_dataloader

PR #4741, opened 2023-12-20
Optimize QueryComplexity algorithm

gmac:
Gives some love to the object-heavy complexity costing algorithm to make it a bit lighter and cleaner. It’s a heavy hitter on our p99 analysis times. This is a 1:1 refactor that provides some reduction to object allocations; taking the [first three specs](https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/analysis/ast/query_complexity_spec.rb#L22-L149) as examples:

- **Spec 1:** objects before 82, after 62, -24.4%
- **Spec 2:** objects before 101, after 77, -23.7%
- **Spec 3:** objects before 134, after 113, -15.6%

These reductions scale linearly with request size, so larger requests (with more object scopes) should see larger gains, thus primarily benefiting p99. Major changes:

1. `ScopedTypeComplexity` inherits from `Hash` and eliminates `scoped_children`. Before, every composite node in the costing tree allocated a Scope + ChildHash (two objects), now these are a single object. This halves the size of the composite scope tree.

2. Arithmetic steps eliminate intermediary data structures in favor of simple `reduce` ops to calculate sum and max. These savings add up as traversal recursions grow with tree size.

3. I also did a bunch of renaming and rewriting comments to make the algorithm (which is pretty dense) a bit easier to follow. A lot of the names and notes come from a Rust port I did a while back. I've only touched local variable names for code safety; method names and instance variables haven't changed.

**NOT here, but ideas to consider:**

- The only reason we need to build leaf costing nodes at all is to support callback hooks, which are probably rarely used. Otherwise leaf nodes could add their costs directly to their composite parent scopes and omit themselves from the tree entirely.

- ~It also looks like the internals of `query.possible_types` is a bit of a grind for something we do repeatedly with the same types during scope matching.~ Looks like this is [cached by warden](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/warden.rb#L185).

### DIFF
--- a/lib/graphql/analysis/ast/query_complexity.rb
+++ b/lib/graphql/analysis/ast/query_complexity.rb
@@ -16,10 +16,11 @@ module GraphQL
           max_possible_complexity
         end
 
-        class ScopedTypeComplexity
-          # A single proc for {#scoped_children} hashes. Use this to avoid repeated allocations,
-          # since the lexical binding isn't important.
-          HASH_CHILDREN = ->(h, k) { h[k] = {} }
+        # ScopedTypeComplexity models a tree of GraphQL types mapped to inner selections, ie:
+        # Hash<GraphQL::BaseType, Hash<String, ScopedTypeComplexity>>
+        class ScopedTypeComplexity < Hash
+          # A proc for defaulting empty namespace requests as a new scope hash.
+          DEFAULT_PROC = ->(h, k) { h[k] = {} }
 
           attr_reader :field_definition, :response_path, :query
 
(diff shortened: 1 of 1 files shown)


lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -65,109 +53,76 @@ def on_enter_field(node, parent, visitor)
...
-            next unless scope.kind.abstract?
-
-            missing_concrete_types = query.possible_types(scope).select { |t| !all_scopes.key?(t) }
-            # This concrete type is possible _only_ as a member of the abstract type.
-            # So, attribute to it the complexity which belongs to the abstract type.
-            missing_concrete_types.each do |concrete_scope|
-              all_scopes[concrete_scope] = all_scopes[scope]
+          # Expand abstract scope types into their concrete implementations;
+          # overlapping abstracts coalesce through their intersecting types.
+          possible_scope_types.keys.each do |possible_scope_type|
```
rmosolgo, line 96: For reducing allocations, could we use `each_key` here to avoid the intermediate array of keys? 

```suggestion
          possible_scope_types.each_key do |possible_scope_type|
```

rmosolgo (commented):
Thanks for this improvement! It looks good to me. I just have one suggestion above.

As far as calcuating leaf complexity, I'm definitely open to skipping it if there's a way to detect whether the callback is _used_ (or implemented?) or not.

lib/graphql/analysis/ast/query_complexity.rb
gmac: In this case, we need the intermediary array because the loop may modify `possible_scope_types`.

rmosolgo: Ah, right! Thanks.

Issue #4739, opened 2023-12-20
Fiber-safe ConnectionPool + Dataloader gobbles all available database connections


swrobel:
**Describe the bug**
When using the Rails 7.1 setting `config.active_support.isolation_level = :fiber`, GraphQL eventually gobbles all database connections

**Versions**

`graphql` version: 2.2.1
`rails` (or other framework): 7.1.2
other applicable versions (`graphql-batch`, etc)

* graphql-c_parser (1.0.7)
* graphql-sources (1.2.0)

**GraphQL schema**

I can't figure out exactly which part of my schema is triggering this, as I can send the same query twice in series and have it succeed the first time, and the second time it will say it can't obtain a connection from the pool. However, the fields I'm requesting don't involve any `dataloader` calls.

**GraphQL query**

This happens with many queries, but here's a simple example GraphQL query and response

```graphql
mutation InitializeUI($input: CreateSessionInput!) {
  createSession(input: $input) {
    errors
  }
}
```

```json
{
  "input": {
    "refreshToken": "8f2b217b-ae3b-4b53-aa19-611bb4512666:85d1287c-74ef-495c-9449-001b88fb416f"
  }
}
```

```json
{
  "data": {
    "createSession": {
      "errors": [],
    }
  }
}
```

**Steps to reproduce**

Send multiple queries in series

**Expected behavior**

Queries should complete successfully

**Actual behavior**

`ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use`

Full backtrace:

<details>
<summary>Click to view exception backtrace</summary>

```
7m     warn: Async::Task [oid=0x20bfc] [ec=0x20c10] [pid=17733] [2023-12-20 09:38:24 -0500]
            | Task may have ended with unhandled exception.
            |   ActiveRecord::ConnectionTimeoutError: could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use
            |   → vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.2/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:128 in `block in wait_poll'
            |     vendor/bundle/ruby/3.2.0/gems/activerecord-7.1.2/lib/active_record/connection_adapters/abstract/connection_pool/queue.rb:117 in `loop'
[...]

PR #4742, opened 2023-12-20
Add queue size option for AsyncDataloader

rmosolgo:
I think switching the ActiveSupport isolation level to `:fiber` would cause each fiber to have its own connection, so it might consume everything from the connection pool. 

See #4739

TODO: 

- [x] Add a semaphore to be used if necessary 
- [x] Add tests to ensure that it's used
- [x] Document the option
- [x] Perhaps add a default value?
- [ ] perhaps recommend upping the connection pool size in Rails config?


swrobel:
I'm still unable to use `AsyncDataloader` even after https://github.com/rmosolgo/graphql-ruby/pull/4743 was merged, as all of the pool gets gobbled up. I wanted to try this setting, but `master` needs to be merged into this branch.

rmosolgo:
:+1: I just merged master into this branch, let me know what you find!

If it's still holding on to all the connections, could you share the output of `my_stat` when you have `AsyncDataloader` enabled?

rmosolgo:
This isn't necessary now, and in #4757 I'm limiting the scope of the Async event loop significantly.

Issue #4747, opened 2023-12-23
Regression in 2.1.9: excess duplicate dataloader fetches


myronmarston:
**Describe the bug**

When attempting to upgrade to 2.1.9, I noticed that `fetch` on my dataloader gets called some extra times with arguments I already resolved.

**Versions**

`graphql` version: 2.1.9
`rails` (or other framework): N/A
`ruby` version: 3.2.2

**GraphQL schema**

```graphql
type ManufacturerConnection {
  nodes: [Manufacturer]
}

type Manufacturer {
  manufactured_parts: PartConnection
}

type PartConnection {
  nodes: [Part]
}

union Part = ElectricalPart | MechanicalPart

type ElectricalPart {
  voltage: Int
  components(filter: ComponentFilter): ComponentConnection
}

type MechanicalPart {
  weight: Int
  components(filter: ComponentFilter): ComponentConnection
}

type ComponentConnection {
  nodes: [Component]
}

type Component {
  widget: Widget
}

type Widget {
  name: String
}

input ComponentFilter {
  name: StringFilter
}

input StringFilter {
  equal_to_any_of: [String]
}

type Query {
  manufacturers: ManufacturerConnection
}
```

**GraphQL query**

```graphql
query {
  manufacturers {
    nodes {
      manufactured_parts {
        nodes {
          ... on MechanicalPart {
            components(filter: {name: {equal_to_any_of: ["c1", "c2", "c3"]}}) {
              nodes {
                widget {
                  name
                }
              }
            }
          }
          ... on ElectricalPart {
            components(filter: {name: {equal_to_any_of: ["c1", "c2", "c3"]}}) {
              nodes {
                widget {
                  name
                }
              }
            }
          }
        }
      }
    }
  }
}
```

**Steps to reproduce**

Put this into a ruby script:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", ENV.fetch("GRAPHQL_VERSION")
end

require "graphql"
require "graphql/dataloader/source"

SCHEMA_STRING = <<~EOS
  type ManufacturerConnection {
    nodes: [Manufacturer]
  }

  type Manufacturer {
    manufactured_parts: PartConnection
  }

  type PartConnection {
    nodes: [Part]
  }

  union Part = ElectricalPart | MechanicalPart

  type ElectricalPart {
    voltage: Int
    components(filter: ComponentFilter): ComponentConnection
  }

  type MechanicalPart {
[...]

Issue #4745, opened 2023-12-21
Circular dependency issue with AsyncDataloader


swrobel:
**Describe the bug**
My existing schema, which works well with the classic Dataloader, fails for certain requests with AsyncDataloader

**Versions**

`graphql` version: master commit `becd60008606`
`rails` (or other framework): 7.1.2
other applicable versions (`graphql-batch`, etc)

* graphql-c_parser (1.0.7)
* graphql-sources (1.2.0)

**GraphQL schema**

Simplified schema to repro:

```ruby
module Types
  class CategoryType < BaseObject
    field :id, ID, null: false

    field :offerings, [Types::OfferingType], null: false

    def offerings(selected: nil, hidden: nil)
      dataloader.with(Loaders::Category::OfferingsLoader, selected, hidden).load(object.id)
    end
  end
end
```

```ruby
module Loaders::Offering
  class CategoryLoader < GraphQL::Dataloader::Source
    def fetch(record_ids)
      @record_ids = record_ids

      record_ids.map do |id|
        records.find do |record|
          record.offerings.any? { |offering| offering.id == id }
        end
      end
    end

    private

    def records
      @records ||= Category
        .joins(:offerings)
        .includes(:offerings)
        .where(offerings: {id: @record_ids})
        .distinct
    end
  end
end
```

```ruby
module Types
  class OfferingType < BaseObject
    field :id, ID, null: false

    field :category, Types::CategoryType, null: true

    private

    def category
      dataloader.with(Loaders::Offering::CategoryLoader).load(object.id)
    end
  end
end
```

```ruby
class Loaders::Category::OfferingsLoader < GraphQL::Dataloader::Source
  attr_reader :selected, :hidden

  def initialize(selected, hidden)
    @selected = selected
    @hidden = hidden
  end

  def fetch(record_ids)
    @record_ids = record_ids

    record_ids.map do |id|
      records.select { |record| record.category_id == id }
    end
  end

  private

  def records
    @records ||=
      Offering.where(category_id: @record_ids)
        .tap { |query| query.where!(selected:) unless selected.nil? }
        .tap { |query| query.where!(hidden:) unless hidden.nil? }
        .order_by_position
  end
end
```

**GraphQL query**

Simplified schema to repro:

```graphql
mutation InitializeUI($input: CreateSessionInput!) {
  createSession(input: $input) {
    user {
      categories {
        id
        offerings {
          id
        }
      }
      offerings {
        id
      }
    }
  }
}
[...]

PR #4748, opened 2023-12-23
Dataloader: fix for large queries and run_isolated

rmosolgo:
Addressing the issue from https://github.com/rmosolgo/graphql-ruby/issues/4745#issuecomment-1868101409 and #4747

TODO: 

- [ ] Backport to 2.1.10

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -169,7 +169,11 @@ module GraphQL
     ensure
       @pending_jobs = prev_queue
       prev_pending_keys.each do |source_instance, pending|
-        source_instance.pending.merge!(pending)
+        pending.each do |key, value|
+          if !source_instance.results.key?(key)
+            source_instance.pending[key] = value
+          end
+        end
       end
     end
 
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -183,27 +183,37 @@ module GraphQL
       source_fibers = []
       next_source_fibers = []
       first_pass = true
+      manager = spawn_fiber do
+        while first_pass || job_fibers.any?
+          first_pass = false
 
-      while first_pass || job_fibers.any?
-        first_pass = false
-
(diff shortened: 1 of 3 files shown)


rmosolgo:
My mistake was removing the checks for `finished = ...` in #4743 . 

I needed to remove those because previously, `suspended` fibers were being left behind by Dataloader. (They would return `true`, be identified as finished, then ignored by Dataloader... but since they were still `.alive?`, Rails would continue to count them as living owners of ActiveRecord connections. But since they were suspended, they would never terminate, and then they'd hog those database connections.) 

So I removed the check for `finished = ...` and continued running the fibers until they terminated. This _worked_ for small queries, when the total number of created fibers is small, but it turns out to create mayhem when there is a large number of fibers. 

The problem is that, when a `.transfer`ed fiber returns, control is not given to the fiber which `.transfer`ed it -- instead, control is given to the top-level fiber. This means that, in nested Dataloader situations (as when argument values are resolved during a query), control may pass willy-nilly from Fiber to Fiber, resulting in mayhem. 

I think this is a bug in Ruby -- wouldn't you expect a fiber's return to transfer back to the Fiber that started it? But I may be wrong. So I opened an issue on the Ruby bug tracker: https://bugs.ruby-lang.org/issues/20081

The problem is, if _that_ bug isn't fixed, then I don't know what to do with `.transfer`-based Fibers. Maybe I can just give up altogether, and say there are two options: 

- synchronous fibers, which use `resume`/`yield`;
- Or `AsyncDataloader` which uses the Async gem. But honestly, I wouldn't be surprised if that gem was negatively affected, too -- which may be why this issue crops up: #4745

rmosolgo:
I tried `Fiber#kill` locally and found that it did _not_ solve this problem because it demonstrated the same control flow issue, where `.kill` would transfer control to the top-level fiber, not the fiber that called it. I opened another bug with Ruby here: https://bugs.ruby-lang.org/issues/20089

Issue #4750, opened 2023-12-24
uninitialized constant Types::Resolvers::BaseResolver (NameError)


henjifeng:
when I start with [getting started](https://graphql-ruby.org/getting_started). I found a bug.

**I wonder why ```BaseResolver``` is not loaded**

```ruby
module Types
    class BaseResolver < GraphQL::Schema::Resolver
    end
end

module Types
  module Resolvers
    class Post < BaseResolver
      type PostType
      argument :id, ID

      def resolve(id:)
        Post.find(id)
      end
    end
  end
end
```

```bash
$ rails c
$ Types::Resolvers::Post

# I get this error

/mnt/hgfs/share_space/rails_space/graphql_demo/app/graphql/types/resolvers/post.rb:19:in `<module:Resolvers>': uninitialized constant Types::Resolvers::BaseResolver (NameError)

    class Post < BaseResolver
                 ^^^^^^^^^^^^
Did you mean?  Types::BaseEdge
```

And then, I tried to change it but it was still wrong
```ruby
module Types
   module Resolvers
    class BaseResolver < GraphQL::Schema::Resolver
    end
  end
end

module Types
  module Resolvers
    class Post < BaseResolver
      type PostType
      argument :id, ID

      def resolve(id:)
        Post.find(id)
      end
    end
  end
end
```

When I changed the inheritance and it's working
```ruby
module Types
  module Resolvers
    class Post < GraphQL::Schema::Resolver
      type PostType, null: false
      argument :id, ID

      def resolve(id:)
        ::Post.find(id)
      end
    end
  end
end
```

rmosolgo:
Addressed by #4751 and #4755

PR #4751, opened 2023-12-24
fix getting start base resolver file name error

henjifeng:
FIx About this issue [uninitialized constant Types::Resolvers::BaseResolver (NameError) #4750](https://github.com/rmosolgo/graphql-ruby/issues/4750)


rmosolgo:
Thanks for the fix! I noticed that the generator is a bit off, too, so I'm updating that in #4755

PR #4753, opened 2023-12-27
[2.1] remove Fiber#transfer support

rmosolgo:
Backporting #4752


rmosolgo:
(Tests pass locally 😅 )

PR #4761, opened 2024-01-04
Add GraphQL Stitching to related projects guide

gmac:
Just noticed this helpful directory... Adding the stitching gem, and clarifying what the Apollo federation gem specifically does.


rmosolgo:
Thanks!

PR #4768, opened 2024-01-09
Document and test ActiveRecord::Promise usage

### LINKED ISSUE #4766: Support for ActiveRecord::Promise

Since Rails 7.1, ActiveRecord exposes methods such as `async_count`, `async_ids` which return a [ActiveRecord::Promise](https://api.rubyonrails.org/classes/ActiveRecord/Promise.html).

This allows parallel DB reads to run due to the DB I/O being async. However, graphql-ruby does not automatically unwrap the promise. Given the AsyncDataloader was implemented recently, it would be nice if promises were handled automatically in both dataloaders, with the ability to leverage the paralellism in the AsyncDataloader.

Returning a ActiveRecord::Promise as the value of a field would result in graphql-ruby automatically unwrapping it to read its return value.

I could implement this myself I guess but I thought it might be worth it to have it in the library, or maybe as an example in the docs.

**Additional context**

The value here is being able to improve performance by making select operations async, for example reading totalCount on a connection.

rmosolgo:
Fixes #4766 

Also: 

- Add a Ruby 3.3 test 
- Add a Rails 7.1 test (I think the Rails 7.0 test may have actually been running on 7.1 -- but now it runs 7.0) 
- Remove Rails 5.2 test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,8 +44,6 @@ jobs:
         include:
           - gemfile: Gemfile
             ruby: 3.2
-          - gemfile: gemfiles/rails_5.2.gemfile
-            ruby: 2.7
           # Rails 6.1 is tested with Postgresql below
           - gemfile: gemfiles/rails_7.0.gemfile
             ruby: 3.1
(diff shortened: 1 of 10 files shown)


trevorturk:
This documentation is so good, thank you! 🎉

PR #4774, opened 2024-01-10
docs: fix example given for GraphQL::Schema::Enum

0x1eef:
The example inherited from `GraphQL::Enum` - a constant that doesn't exist. 
Update the example to inherit from `GraphQL::Schema::Enum` instead.


rmosolgo:
Thanks for the fix -- sorry about that!

PR #4775, opened 2024-01-10
Add Sentry tracing

### LINKED ISSUE #3655: Add support for sentry tracing

I'd like to be able to use graphql-ruby together with sentry's tracing.

It would be nice if i could see separate traces for all graphql queries in sentry's tracing.

I've tried appsignal since they also have a plan for non-profits, but there there's also no per query tracing yet.

**Additional context**

https://docs.sentry.io/product/sentry-basics/tracing/distributed-tracing/

patch0:
Closes #3655

Adds support for sending traces to [Sentry](https://sentry.io).  The data names follow the [OpenTelemetry Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/database/graphql/), which is [how Sentry say](https://develop.sentry.dev/sdk/performance/span-data-conventions/) data should be sent.

## Points for consideration

I've aliased `execute_multiplex` / `analyze_multiplex` to `graphql.execute_multiplex` / `graphql.analyze_multiplex` operations, rather than `graphql.execute` / `graphql.analyze` which is what [other](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/prometheus_trace.rb#L22) [tracers](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/data_dog_trace.rb#L31) do.  This is because the initial entry point for queries seems to be `execute_multiplex`, followed by an `execute_query` call later.  If there is just one operation in the `execute_multiplex` span, we'd end up with two spans with identical names in the same transaction, which could be confusing.  The description could be postponed until the `execute_query` is called, but I felt it was useful to have an overall description set at the top of the span.

When an operation name is not set, I've not used `anonymous` as the identifier which is what the [base class does](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/platform_trace.rb#L108), but followed the OpenTelemetry spec which states that the name should be omitted.

If the operation name and type are both missing, this defaults to `GraphQL Operation`, which is what has been suggested by the [OpenTelemetry convention](https://opentelemetry.io/docs/specs/semconv/database/graphql/).

No attempt is made to sanitize the GraphQL document for any sensitive data.

I've not snake-cased the data names, although Sentry [suggests this is preferred](https://develop.sentry.dev/sdk/performance/span-data-conventions/).

## Questions ❓ 

In testing I've found that the tests for `GraphQL::Tracing::SentryTracing` (the legacy tracing) use `GraphQL::Tracing::SentryTrace` unless I remove `require 'graphql/tracing/sentry_trace'` form `graphql/tracing.rb`.  Is there a better way to ensure that the `SentryTracing` spec actually tests `SentryTracing`?


rmosolgo:
Hi, thanks for working this up! I'll take a look soon.

patch0:
> Hi, thanks for working this up! I'll take a look soon.

Thanks @rmosolgo :) @sl0thentr0py, one of the engineers from Sentry is also interested in it too.

rmosolgo:
Thanks again for your work on this! I took a look and: 

- Removed the legacy-style `SentryTracing` class. Since we don't have to support backwards-compatibility, let's just add the new one. 
- Updated the tracing docs to use the new-style `*Trace` modules 

Sorry about the trouble with testing. In fact, `LegacyTracing` classes are made to use the new-style modules instead, unless a specific opt-in option is given: 

https://github.com/rmosolgo/graphql-ruby/blob/0f3288796c8873686e53d646364242582be6374e/lib/graphql/tracing/platform_tracing.rb#L75-L91

(I added that option after implementing the under-the-hood migration, just in case people ran into trouble. But I never documented it, and I never heard any reports of issues with the new modules. It's time to deprecate (and eventually remove) those old ones!)

sl0thentr0py:
late to the party but the sentry data / span looks good to me from our side! thank you for adding this :)

sl0thentr0py:
alright so we like to make things as zero-configurable as possible, so what I'm going to do on our side now is

* detect this gem
* add an (opt-out) patch to automatically enable this tracing on each graphql schema class

let me know if you see a problem with that! I think this is generally useful for all our users.

sl0thentr0py:
yep something like that, thanks for the samples!

We will just have a top-level switch to turn off the whole patch and then people can of course add it manually to the ones they prefer. We just prefer to make it easier for the 99% use case to work with minimal configuration.

PR #4776, opened 2024-01-10
Remove `span_type` method call on SpanOperation

### LINKED ISSUE #4762: Question related to upgrade
👋 Hi, I am currently working on [`dd-trace-rb`](https://github.com/DataDog/dd-trace-rb) at Datadog. I have a question related to upgrading. 

Currently, I am working on `dd-trace-rb` 2.0 upgrade, which contains breaking changes of our public API. 

[This PR](https://github.com/DataDog/dd-trace-rb/pull/3330) removes the deprecated method that is used at [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/tracing/data_dog_tracing.rb#L19) in `graphql-ruby`. If I change to `span.type = 'custom'`  in `graphql-ruby`, it is still possible that it breaks application since there is no version/dependency management between `dd-trace-rb` and `graphql-ruby`.

I was wondering how would you recommend handling this situation?

TonyCTHsu:
Closes #4762 

**Why**

Datadog tracer is removing a deprecated method `span_type` for `dd-trace-rb` 2+ . [This is the PR](https://github.com/DataDog/dd-trace-rb/pull/3330).

**What**

* Change the tracing implementation in `graphql-ruby` from `span_type` to `type`.
* Remove constant reference from `dd-trace-rb` 
* Deprecate `service` option and its default
* Deprecate `tracer` option

**Additional Notes**

The test is setup with https://github.com/DataDog/dd-trace-rb/pull/3376

🚨 This change is backward compatible (All graphql-ruby versions works with `dd-trace-rb` 1+ ).  However, `graphql-ruby` users need to upgrade with this change to use `dd-trace-rb` 2+.


lib/graphql/tracing/data_dog_trace.rb
```diff
@@ -158,8 +155,7 @@ def resolve_type_lazy(object:, type:, query:)
 
       def resolve_type_span(span_key, object, type, query)
         platform_key = @platform_key_cache[DataDogTrace].platform_resolve_type_key_cache[type]
-        @tracer.trace(platform_key, service: @service_name) do |span|
-          span.span_type = 'custom'
+        @tracer.trace(platform_key, service: @service_name, type: 'custom') do |span|
```
rmosolgo, line 158: Just to make sure I understand: 

What will dd-trace-rb do with this `type: 'custom'` argument? Will it ignore it, or 💥 ? 

If this code only works with dd-trace-rb 2+, what do you think about adding a version check of some kind in `def initialize` above? If that's possible, it might be a nicer development experience than troubleshooting the argument error (if `span: ...` _does_ cause an error, that is).

TonyCTHsu: > What will dd-trace-rb do with this type: 'custom' argument?

The data model of `Span`/`SpanOperation` contains a `type` field, which can filled by multiple ways. 

1. Through `Tracer#trace(...)`, there are both `type`/`span_type` as optional keyword arguments. https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/tracer.rb#L131-L132

Both options yields the same outcome. 

```ruby Data
[2] pry(main)> span = Datadog::Tracing.trace('my_span', span_type: 'custom');
[3] pry(main)> span.type
=> "custom"
```

```ruby
[4] pry(main)> span = Datadog::Tracing.trace('my_span', type: 'custom');
[5] pry(main)> span.type
=> "custom"
```

2. Through attribute setter from the `SpanOperation`, https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/span_operation.rb#L502
Since, `span_type=` is just an alias for `type=`, the following code are the same.

```ruby
[3] pry(main)> Datadog::Tracing.trace('my_span') do |span|
[3] pry(main)*   span.span_type = 'custom'
[3] pry(main)*   puts span.type
[3] pry(main)* end
custom
```

```ruby
[3] pry(main)> Datadog::Tracing.trace('my_span') do |span|
[3] pry(main)*   span.type = 'custom'
[3] pry(main)*   puts span.type
[3] pry(main)* end
custom
```

>If this code only works with dd-trace-rb 2+

This is not entirely correct. Because the changes basically migrate the deprecated usage to a supported usage, it  works on all `dd-trace-rb` versions. 

* `dd-trace-rb` 1.x supports both `span_type=` and `type=`
* `dd-trace-rb` 2.x only supports `type=`

The existing `graphql-ruby` versions without such changes won't work with `dd-trace-rb` 2+.

PR #4780, opened 2024-01-12
Fix hasNextPage when max_page_size is set

### LINKED ISSUE #4765: Different pagination behaviour using Array connection wrapper vs ActiveRecord Associations
**Describe the bug**

When returning a paginated connection field, the behaviour is different when an array of ActiveRecord objects is returned vs an ActiveRecord association. This is most noticeable when both `before` and `after` are supplied, and the corresponding indices to the cursors are 1 apart (ie the index corresponding to the `after` cursor is 1 smaller than the index corresponding to the `before` cursor). When this happens we see a single edge in the payload where we would expect an empty array.

**Versions**

`graphql` version: 2.1.6
`rails` (or other framework): 7.1.1

**GraphQL schema**

```ruby
  field :id, ID, hash_key: :id
  # …
end

  field :products_using_array, Product.connection_type, null: false

  def products_using_array
    ::Product.all.to_a
  end

  field :products_using_association, Product.connection_type, null: false

  def products_using_association
    ::Product.all
  end
end

  query QueryType
  # …
end
```

**GraphQL query**

```graphql
query {
  productsUsingArray(
    after: "MQ", # Index 1
    before: "Mg"  # Index 2
  ) {
    products {
      edges {
        cursor
        node {
           ...
        }
     }
  }
}
```
The query above will return the following payload (with page info included)
```json
{
  "data": {
    "features": {
      "pageInfo": {
        "startCursor": "Mg",
        "endCursor": "Mg",
        "hasNextPage": false,
        "hasPreviousPage": true
      },
      "edges": [
        {
          "cursor": "Mg",
          "node": {
              ...
          }
        }
      ]
    }
  }
}
```

Notice how the edge with cursor `Mg` still appears in the payload, despite the fact that we've specified it as our `before` value (which is an exclusive boundary).

When we repeat the same query but using the ActiveRecord connection wrapper instead of the Array-based one we get a different result

```json
{
  "data": {
    "features": {
      "pageInfo": {
        "startCursor": null,
        "endCursor": null,
        "hasNextPage": true,
        "hasPreviousPage": true
      },
      "edges": []
    }
  }
}
```

This is the result I would expect from such a query, as the exclusive boundary should not have any records in its range. It seems to me like the `before` value is being ignored in this case, will take a closer look and see if I can put up a fix tomorrow!

rmosolgo:
About hasNextPage, the relay spec says: 

> [HasNextPage](https://relay.dev/graphql/connections.htm#HasNextPage())(allEdges, before, after, first, last)
> 
>     If first is set:
>         Let edges be the result of calling [ApplyCursorsToEdges](https://relay.dev/graphql/connections.htm#ApplyCursorsToEdges())(allEdges, before, after).
>         If edges contains more than first elements return true, otherwise false.
>     If before is set:
>         If the server can efficiently determine that elements exist following before, return true.
>     Return false.

But GraphQL-Ruby's `max_page_size` was causing it to take the "if first is set..." branch even when the client didn't provide `first`. 

After this change, it will fall down to the "Return false" branch of that spec unless `first` is explicitly provided by the client.

Fixes #4765

### DIFF
--- a/lib/graphql/pagination/array_connection.rb
+++ b/lib/graphql/pagination/array_connection.rb
@@ -35,10 +35,10 @@ module GraphQL
       def load_nodes
         @nodes ||= begin
           sliced_nodes = if before && after
-            end_idx = index_from_cursor(before)-1
+            end_idx = index_from_cursor(before) - 2
             end_idx < 0 ? [] : items[index_from_cursor(after)..end_idx] || []
           elsif before
-            end_idx = index_from_cursor(before)-2
+            end_idx = index_from_cursor(before) - 2
             end_idx < 0 ? [] : items[0..end_idx] || []
           elsif after
             items[index_from_cursor(after)..-1] || []
@@ -56,7 +56,7 @@ module GraphQL
             false
           end
 
-          @has_next_page = if first
+          @has_next_page = if first_value && first
             # There are more items after these items
             sliced_nodes.count > first
           elsif before
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -31,7 +31,7 @@ module GraphQL
         if @has_next_page.nil?
           @has_next_page = if before_offset && before_offset > 0
             true
-          elsif first
+          elsif first && first_value
             if @nodes && @nodes.count < first
               false
             else
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -29,14 +29,14 @@ module GraphQL
 
       def has_next_page
         if @has_next_page.nil?
-          @has_next_page = if before_offset && before_offset > 0
-            true
-          elsif first && first_value
+          @has_next_page = if first && first_value
             if @nodes && @nodes.count < first
               false
             else
               relation_larger_than(sliced_nodes, @sliced_nodes_offset, first)
             end
+          elsif before_offset && before_offset > 0
+            true
           else
             false
           end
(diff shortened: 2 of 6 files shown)


porter77:
This change broke a lot of our clients since we were dependent on the old (incorrect) behavior. Clients were not explicitly passing `first`, but rather depending on the default page size values. After upgrading we starting getting `hasNextPage: false` for all connection requests even though there were more records. If anyone else stumbles into this issue, this was our temporary fix (replace MyAppSchema with your schema...)

```ruby
module Types
  class BaseArgument < GraphQL::Schema::Argument
    def initialize(name, type, desc = nil, **kwargs)
      if name == :first && kwargs[:owner].connection?
        kwargs[:default_value] = kwargs[:owner].default_page_size || MyAppSchema.default_page_size
      end

      super
    end
  end
end
```

This will explicitly set the default value of the `first` argument to match your `default_page_size` value

eapache-opslevel:
@rmosolgo we also just had a small incident because of this surprise. Frankly, the relay spec doesn't seem to support a default max-page-size at all... if no `first` is provided, the spec "requires" (if you read it literally and with no room for flexibility) that you return all edges. But max-page-size is clearly a nice feature and otherwise integrates well.

In this part of the spec, I would consider a default max-page-size value to count as "first being set" in the absence of all other values...

eapache-opslevel:
I (think I) understand the index bug from #4765, but I don't understand the hasNextPage part of the discussion, why is `hasNextPage: true` wrong when the cursors return an empty list of edges but there are more edges available?

rmosolgo:
Yeah, sorry about the trouble ... maybe this was the wrong approach. I'm open to making it return `true` if max_page_size is set. Even though this is a violation of the spec, as you noted, the spec doesn't cover max_page_size...

rmosolgo:
Here's a PR to revert the change to `hasNextPage`, please let me know if you think we should go through with it: https://github.com/rmosolgo/graphql-ruby/pull/4866

rmosolgo:
Sorry for the disruption caused by this change -- I'm going to revert it in 2.2.12, so that `hasNextPage` is `true` (again) when `max_page_size` is applied.

rmosolgo:
I just shipped the revert (#4866) in 2.2.12. Sorry again for the trouble from this!

PR #4783, opened 2024-01-14
Unsubscribe from ActionCable on link cleanup

ball-hayden:
An `Observable` should return a cleanup function.

We should return a function to unsubscribe, rather than return an object which isn't callable.

### DIFF
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -55,8 +55,11 @@ class ActionCableLink extends ApolloLink {
           }
         }
       })
-      // Make the ActionCable subscription behave like an Apollo subscription
-      return Object.assign(channel, {closed: false})
+
+      // The Observable should return a cleanup function
+      return () => {
+        channel.unsubscribe()
+      }
     })
   }
 }


rmosolgo:
Hi! Thanks for the suggestion. I expect this object to get its `unsubscribe` method from the `channel` object. Here it's defined in the Rails source: https://github.com/rails/rails/blob/0ecfbe25688a0deb9aaee0596c99148002b9f3a4/actioncable/app/javascript/action_cable/subscription.js#L86-L88

Did you run into a bug with that implementation?

ball-hayden:
@rmosolgo just that the interface is wrong.

`Observable` doesn't expect an `unsubscribe` method, it expects a cleanup function?
https://github.com/zenparsing/zen-observable?tab=readme-ov-file#new-observablesubscribe

I might be missing it, but I can't see where Apollo would know to call an unsubscribe method instead?

Apollo's implementation returns the result of [`this.client.subscribe`](https://github.com/apollographql/apollo-client/blob/9b22974b3e0f131a50c2618cc08ea6745e79a653/src/link/subscriptions/index.ts#L56), which is a function (that unsubscribes: https://the-guild.dev/graphql/ws/docs/interfaces/client.Client#returns-4)

The documentation appears to be fairly lacking/opaque, but I'm fairly confident what is currently here isn't correct.

rmosolgo:
Ok, I've kept looking into it and I think I have something to go on. If I understand right, your question is about the function passed to `new Observable(...)`. The example you linked has the function returning another function which cleans up state.

But the docs below mention that the function may _also_ return a "subscription" object whose `.unsubscribe` method will be called: 

> The subscriber function can optionally return either a cleanup function or a **subscription object.** If it returns a cleanup function, that function will be called when the subscription has closed. If it returns a subscription object, then the subscription's **`unsubscribe` method** will be invoked when the subscription has closed.

(Emphasis mine.) 

I was double-checking the tests for this code, and they _seemed_ like they would be testing this: https://github.com/rmosolgo/graphql-ruby/blob/14796647ab906a5705fec59a94cba8057dc9a1c0/javascript_client/src/subscriptions/__tests__/ActionCableLinkTest.ts#L65-L106

But that brings me back to my first question: did you observe a bug or misbehavior that led you to think that this code should be changed? I don't want to miss that if so!

ball-hayden:
Ahh, sorry.
Yeah - you're quite right.

We were seeing some odd behaviour around subscriptions, but I think I've jumped to conclusions here.

We're working in a React Native app, so inspecting what's going on with Websockets is a bit more tricky.

I'll have a better look and re-open this if the behaviour is actually different. Sorry for the noise!

PR #4790, opened 2024-01-17
"How complexity scoring works" guide

gmac:
A frequent source of confusion internally. This summarizes the nuances of the complexity scoring process for the benefit of all.


guides/queries/complexity_and_depth.md
```diff
@@ -10,6 +10,46 @@ index: 4
 
 GraphQL-Ruby ships with some validations based on {% internal_link "query analysis", "/queries/ast_analysis" %}. You can customize them as-needed, too.
 
+## Prevent deeply-nested queries
```
gmac, line 13: Moving depth up so to keep complexity topics next to each other.

rmosolgo:
Thanks for writing this up!

PR #4793, opened 2024-01-19
ActionCable: don't use send action

### LINKED ISSUE #4792: Can't get ActionCable subscriptions to work
**Describe the bug**

There's two issues:

1. the browser fails to connect to the websocket semi-randomly. This happens almost always on firefox and sometimes on chrome. The WS connection shows up in the Network tab devtools but no messages are exchanged. This might not be a Graphql-ruby bug but I thought I'd mention it. I am still trying to figure this one out
2. ActionCable internally logs a cryptic error

Issue 2 I think is most likely a bug in graphql-ruby

The error message is: `[GraphqlChannel] Unable to process GraphqlChannel#send({"variables"=>{"userId"=>nil}, "operationName"=>"useWorkspaceLoader_Subscription",` - I cut off the rest which isn't important

I traced it here in the rails source code: https://github.com/rails/rails/blob/21eceb7db26b724d1f7a824d4cc04ddeba44244b/actioncable/lib/action_cable/channel/base.rb#L176

I think the issue is that the graphql client will try to perform a "send" operation but the channel on the ruby code only has a handler for the "execute" operation, see here: https://github.com/rmosolgo/graphql-ruby/blob/85e711771041ea7ace91003d1069fd092a8eeec9/javascript_client/src/subscriptions/createActionCableHandler.ts#L47-L48 

**Versions**

`graphql` version: 2.2.4
`rails` (or other framework): 7.1

**GraphQL schema**

I don't think it's relevant

**GraphQL query**

I don't think it's relevant

**Steps to reproduce**

Setting up a project with Relay on the frontend and ActionCable on the backend and trying to run subscriptions

**Expected behavior**

Subscriptions work fine

**Actual behavior**

I can't get them to actually work even though there is no exception with stack trace anywhere

**Additional context**

The project is a Next.js app using the "pages" folder for setting up routes

rmosolgo:
Oops -- somehow this entered back in https://github.com/rmosolgo/graphql-ruby/commit/cf0a3fe12f54e1ce3b3e3ae87fa78d8ee111f655#diff-9957e220f3cd45075c8bd11a71aa74970869a76e46eaff6007685295c806c6fdR45, but I don't think it should be there. 

Fixes #4792

### DIFF
--- a/javascript_client/src/subscriptions/__tests__/createActionCableHandlerTest.ts
+++ b/javascript_client/src/subscriptions/__tests__/createActionCableHandlerTest.ts
@@ -51,14 +51,16 @@ describe("createActionCableHandler", () => {
       {text: "", name: "", id: "abcdef"},
       {},
       {},
-      { onError: () => {}, onNext: () => {}, onCompleted: () => {} }
+      { onError: () => {}, onNext: (result: any) => { log.push(["onNext", result])}, onCompleted: () => { log.push(["onCompleted", null])} }
     )
 
     handlers.connected() // trigger the GraphQL send
+    handlers.received({ result: { data: { a: "1" } }, more: false })
 
     expect(log).toEqual([
-      ["send", { operationId: "client-1/abcdef", operationName: "", query: "", variables: {} }],
       ["execute", { operationId: "client-1/abcdef", operationName: "", query: "", variables: {} }],
+      ["onNext", { data: { a: "1" } }],
+      ["onCompleted", null],
     ])
   })
 })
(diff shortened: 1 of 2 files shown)


javascript_client/src/subscriptions/__tests__/createActionCableHandlerTest.ts
```diff
@@ -51,14 +51,16 @@ describe("createActionCableHandler", () => {
...
     )
 
     handlers.connected() // trigger the GraphQL send
+    handlers.received({ result: { data: { a: "1" } }, more: false })
 
     expect(log).toEqual([
-      ["send", { operationId: "client-1/abcdef", operationName: "", query: "", variables: {} }],
       ["execute", { operationId: "client-1/abcdef", operationName: "", query: "", variables: {} }],
+      ["onNext", { data: { a: "1" } }],
+      ["onCompleted", null],
```
rmosolgo, line 63: I assume that the previous assertion here was added after-the-fact, based on the assumption that the code worked. I'm removing that -- and adding tests for a couple other hooks in the meantime.

rmosolgo:
🚢  in graphql-ruby-client 1.13.0!

PR #4795, opened 2024-01-20
Fix typos in DataLoader testing doc + add source example

rubendinho:
## Proposed changes:
- Fix typo with numbers of records
- Remove `load` method since the dataloader returns a record, at least according to the example here: https://graphql-ruby.org/dataloader/sources.html#example-loading-activerecord-objects-by-id.  I included that implementation in the example for reference.

### DIFF
--- a/guides/dataloader/testing.md
+++ b/guides/dataloader/testing.md
@@ -43,7 +43,26 @@ You could also make specific assertions on the queries that are run (see the [`s
 
 ## Testing Dataloader Sources
 
-You can also test `Dataloader` behavior outside of GraphQL using {{ "GraphQL::Dataloader.with_dataloading" | api_doc }}. For example:
+You can also test `Dataloader` behavior outside of GraphQL using {{ "GraphQL::Dataloader.with_dataloading" | api_doc }}. For example, let's assuming have a `Sources::ActiveRecord` defined like so: 
+
+```ruby
+
+module Sources
+  class ActiveRecord < GraphQL::Dataloader::Source
+    def initialize(model_class)
+      @model_class = model_class
+    end
+
+    def fetch(ids)
+      records = @model_class.where(id: ids)
+      # return a list with `nil` for any ID that wasn't found, so the shape matches
+      ids.map { |id| records.find { |r| r.id == id.to_i } }
+    end
+  end
+end
+```
+
+You can test it like so: 
 
 ```ruby
 def test_it_fetches_objects_by_id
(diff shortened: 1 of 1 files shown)


guides/dataloader/testing.md
```diff
@@ -43,7 +43,26 @@ You could also make specific assertions on the queries that are run (see the [`s
...
 
-You can also test `Dataloader` behavior outside of GraphQL using {{ "GraphQL::Dataloader.with_dataloading" | api_doc }}. For example:
+You can also test `Dataloader` behavior outside of GraphQL using {{ "GraphQL::Dataloader.with_dataloading" | api_doc }}. For example, let's if you have a `Sources::ActiveRecord` source defined like so: 
+
+```ruby
+
+module Sources
+  class ActiveRecord < GraphQL::Dataloader::Source
+    def initialize(model_class)
+      @model_class = model_class
```
rmosolgo, line 53: I don't think this is the right `initialize` API for this example, because in the example, no `model_class` is given to `dataloader.with(...)`. Admittedly, the example below couldn't really work, unless `.id` somehow contained information about what data to fetch. But maybe instead, the example below could be clarified to be `Sources::User` instead, so that no example implementation was required.

guides/dataloader/testing.md
```diff
@@ -60,10 +79,10 @@ def test_it_fetches_objects_by_id
...
 
       # Validate source's matching up of records
-      expect(req1.load).to eq(user_1)
-      expect(req2.load).to eq(user_3)
-      expect(req3.load).to eq(user_2)
-      expect(req4.load).to be_nil
+      expect(req1).to eq(user_1)
+      expect(req2).to eq(user_2)
+      expect(req3).to eq(user_3)
+      expect(req4).to be_nil
```
rmosolgo, line 85: I don't think this change is right, for a couple reasons: 

- the examples above use `.request`, which registers the data dependency with `Dataloader` but _doesn't_ yield the fiber; it returns a pending request which must be `.load`-ed in order to return the actual value. I think we need the `.load` calls here. 
- In the example above, `req2` is a request for `user_3.id`; I would expect `req2.load` to return `user_3`, not `user_2`.

If you think that this will work without `.load`, would you mind sharing a small replication script where `.request` returns the dataloaded object?

rmosolgo (changes requested):

guides/dataloader/testing.md
rubendinho: I was going off of this example but will rename it to Sources::User to make it more obvious.

guides/dataloader/testing.md
rubendinho: the example above failed my tests. Here's roughly what the source looks like: 

```rb
class Sources::User < GraphQL::Dataloader::Source
  def initialize(model_class) # rubocop:disable Lint/MissingSuper
    @model_class = model_class
  end

   def fetch(ids)
    records = @model_class.where(id: ids)

    # Return a list of records, with `nil` for any ID that wasn't found
    ids.map do |uuid|
      # Return the record if there's a match
      records.find do |record|
        record.id == id
      end
    end
  end
end
```

rmosolgo:
Thanks for suggesting these improvements! I took a minute to make those changes and pushed them here.

guides/dataloader/testing.md
rubendinho: Ah i was doing `.load(id)` instead of `.request(id)`

PR #4798, opened 2024-01-22
Support generate-persisted-query-manifest for OperationStore

rmosolgo:
Adding support for a new (?) Apollo persisted query format: https://www.npmjs.com/package/@apollo/generate-persisted-query-manifest

TODO: 

- [x] hook up to `sync`, add a test from that level


rmosolgo:
🚢  in graphql-ruby-client 1.13.0!

PR #4800, opened 2024-01-22
Apply validate_timeout to analysis, too

### LINKED ISSUE #4629: Duplicated fields in query cause long processing times, but does not increment query complexity
**Describe the bug**

Given a query with the same field duplicated multiple times:

```ruby
query = <<QUERY
mutation($uuid: Uuid) {
  updateSomething(
    input: {
      uuid: $uuid
    }
  ) {
    #{"success " * 9999}
  }
}
QUERY
```

I expect this field to be "deduplicated" and only run once. But this query takes many seconds to execute, even though it only runs the mutation resolver and `success` method once.

Using a variation of the complexity analyzer from the docs:

```ruby
    class LogQueryComplexityAnalyzer < GraphQL::Analysis::AST::QueryComplexity
      def on_enter_field(node, parent, visitor)
        puts "on_enter_field" # this line is printed 9999 times.
        super(node, parent, visitor)
      end
      def result
        complexity = super
        message = "[GraphQL Query Complexity] #{complexity}}" 
        puts message # prints "[GraphQL Query Complexity]  5"
      end
    end
    query_analyzer(LogQueryComplexityAnalyzer)
```

I note that the AST seems to have parsed the query such that the duplicated field is represented 9999 times. I might expect the parser to deduplicate fields instead of instantiating the whole tree and then somehow skipping the duplicates while visiting them?

**Versions**

`graphql` version: `2.0.8`
`rails`:  `6.1.7.4`
`graphql-pro`: `1.22.2`

**Additional context**

This is relevant because it can be a DOS vector. A bad actor can execute relatively few queries that request any number of a single expensive field, without hitting an http request rate limit or a graphql complexity limit.

rmosolgo:
Fixes #4629 

I think this will be fair to apply to analysis too, since that's usually CPU work, like static validation. Please leave a comment here if you have a concern about it -- another implementation would be to add a separate `analysis_timeout ...` option.

### DIFF
--- a/guides/queries/timeout.md
+++ b/guides/queries/timeout.md
@@ -63,16 +63,17 @@ class MySchema < GraphQL::Schema
 end
 ```
 
-## Validation
+## Validation and Analysis
 
 Queries can originate from a user, and may be crafted in a manner to take a long time to validate against the schema.
 
-It is possible to limit how many seconds the static validation rules are allowed to run before returning a validation timeout error. The default is no timeout.
+It is possible to limit how many seconds the static validation rules and analysers are allowed to run before returning a validation timeout error. The default is no timeout.
 
 For example:
 
 ```ruby
 class MySchema < GraphQL::Schema
+  # Applies to static validation and query analysis
   validate_timeout 10
 end
 ```
--- a/lib/graphql/analysis/ast.rb
+++ b/lib/graphql/analysis/ast.rb
@@ -6,6 +6,7 @@ require "graphql/analysis/ast/query_complexity"
 require "graphql/analysis/ast/max_query_complexity"
 require "graphql/analysis/ast/query_depth"
 require "graphql/analysis/ast/max_query_depth"
+require "timeout"
 
 module GraphQL
   module Analysis
@@ -63,7 +64,10 @@ module GraphQL
                 analyzers: analyzers_to_run
               )
 
-              visitor.visit
+              # `nil` or `0` causes no timeout
+              Timeout::timeout(query.validate_timeout_remaining) do
+                visitor.visit
+              end
 
               if visitor.rescued_errors.any?
                 return visitor.rescued_errors
@@ -75,6 +79,8 @@ module GraphQL
             []
           end
         end
+      rescue Timeout::Error
+        [GraphQL::AnalysisError.new("Timeout on validation of query")]
       end
 
       def analysis_errors(results)
--- a/lib/graphql/query.rb
+++ b/lib/graphql/query.rb
@@ -317,7 +317,7 @@ module GraphQL
     end
 
     def_delegators :validation_pipeline, :validation_errors,
-                   :analyzers, :ast_analyzers, :max_depth, :max_complexity
+                   :analyzers, :ast_analyzers, :max_depth, :max_complexity, :validate_timeout_remaining
 
     attr_accessor :analysis_errors
     def valid?
--- a/lib/graphql/query/validation_pipeline.rb
+++ b/lib/graphql/query/validation_pipeline.rb
@@ -14,7 +14,7 @@ module GraphQL
     #
     # @api private
     class ValidationPipeline
-      attr_reader :max_depth, :max_complexity
+      attr_reader :max_depth, :max_complexity, :validate_timeout_remaining
 
       def initialize(query:, parse_error:, operation_name_error:, max_depth:, max_complexity:)
         @validation_errors = []
@@ -71,7 +71,7 @@ module GraphQL
           validator = @query.static_validator || @schema.static_validator
           validation_result = validator.validate(@query, validate: @query.validate, timeout: @schema.validate_timeout, max_errors: @schema.validate_max_errors)
           @validation_errors.concat(validation_result[:errors])
-
+          @validate_timeout_remaining = validation_result[:remaining_timeout]
           if @validation_errors.empty?
             @validation_errors.concat(@query.variables.errors)
           end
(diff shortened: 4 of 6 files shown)


thlacroix:
Hi @rmosolgo , (let me know if you prefer a dedicated issue to discus that)

I was looking at using validate_timeout to fail early for queries where the AST visit takes too long. I noticed that [the documentation](https://graphql-ruby.org/queries/timeout.html) has the following warning:

```
Note: This configuration uses Ruby’s built-in Timeout API, which can interrupt IO calls mid-flight, resulting in [very weird bugs](https://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/). None of GraphQL-Ruby’s validators make IO calls but if you want to use this configuration and you have custom static validators that make IO calls, open an issue to discuss implementing this in an IO-safe way.
```

After checking the code and previous issues, I found this issue https://github.com/rmosolgo/graphql-ruby/issues/5036 and this fix https://github.com/rmosolgo/graphql-ruby/pull/5071, suggesting that we might have IO side effects (and ExecutionError) during static analysis, specifically in the embedded QueryComplexity analyzer; caused by this method call https://github.com/rmosolgo/graphql-ruby/blob/v2.4.10/lib/graphql/schema/field.rb#L504 . I reproduced locally and noticed an ActiveRecord DB call when running `GraphQL::Analysis.analyze_query(query, [GraphQL::Analysis::QueryComplexity ])` , which loads objects via connection arguments (like in the issue linked above).

In this case, from what I understand, that would make the use of Timeout unsafe for this analyzer.

So a few questions:
* Is the Timeout still safe in this case ?
* Is there a way to fix the QueryComplexity analyzer to avoid side effects (as in this case I think we don't really need to load the arguments as we just check for a few static ones) ? I expected the embedded analyzers to be fully static.
* Bonus question (could be a dedicated feature request): is there a proper way to stop the analysis visitor early from an analyzer (for example when we have visited more than X nodes) ? I can raise a custom error from the analyzer which will stop the analysis process, but that's not very clean IMO. It would be great to be able to call something like `visitor.halt` for example.

Thanks,
Thomas

rmosolgo:
Hey, to answer your questions: 

1. No, timeout is not safe in this case. 
2. There might be a way to avoid side-effects. I'd be interested to see a full backtrace of the DB queries that run during this step in your app -- maybe GraphQL-Ruby is loading arguments that don't _need_ to be loaded. Maybe the `.arguments_for` call you linked could be somehow optimized to only read `first` and `last` directly from the raw hash instead, bypassing argument loading. (This would somehow have to handle cases where people have customized those arguments, though.) 
3. I would have suggested raising a `GraphQL::ExecutionError` which would terminate the analysis step and return an error to the client. If you try that approach and find something missing, please open a new issue. 

At a higher level, it might be possible to modify the timeout handling to be properly IO-safe. It could be like this: 


- Parsing continues using Ruby's `Timeout` API, based on the assumption that no I/O happens during parsing. 
- Analysis uses a manually-implemented timeout, like execution. In this case:
  - The analysis code sets a `@remaining_ms` value based on the `validate_timeout - validate_duration`
  - Somewhere during analysis, the code measures `elapsed_ms` and then does `@remaining_ms -= elapsed_ms`, and if `@remaining_ms` is less than 0, it raises a nice error. 
  
That's similar to how timeout is implemented for query execution. As a small downside, if the timeout duration passes _while_ a long-running I/O operation is running, then the actual duration of the operation may exceed the configured timeout. But IMO that's a small price to pay to I/O-safe timeout.
  
  Let me know what you think of those ideas :+1:

thlacroix:
Thanks for your response.

* I've tried raising a `GraphQL::ExecutionError` from the analyzer, but it's rescued here and the analyze_query returns an empty array https://github.com/rmosolgo/graphql-ruby/pull/5071/files#diff-0b6b87e1afefbdcb680e3bdfedada3491114642bd1e8097a86e5420d55e51ad7R86 (so no error)
* If we can ensure that the lib Analyzers don't have any side-effect, we could be fine with a standard Timeout (as ours are static).We need to check the performance impact of using Timeout first though. I'm OK with your proposed solution as well (if we check after each visit for example).
* What we experience is similar to this case in the spec https://github.com/rmosolgo/graphql-ruby/pull/5071/files#diff-4bf0809b54d9870b3be13a6fe4c3c620279aa5a4ac5ca9c23d984f93a428082bR194 where we see that the connection argument object is loaded during analysis (through ActiveRecord in our case); that was the case that the PR was addressing. I can provide a more concrete example if needed, but this original issue illustrates it well I think https://github.com/rmosolgo/graphql-ruby/issues/5036 . The fetch is called here https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/interpreter/arguments_cache.rb#L24 , and I can see the query arguments_cache filled with the resolved object after running the analyzer.

rmosolgo:
> raising a `GraphQL::ExecutionError`... (so no error)

That's a bug -- the test in that PR was trying to make sure those errors went to clients :S I'd be happy to help track it down if you could open a new issue and include the code in your app which raises that error, along with its surrounding GraphQL-Ruby-related context (type definition, method, `prepare:` proc, or whatever) 


> connection argument object is loaded during analysis

This can probably be improved so that it only gets the arguments it really _needs_. The trick will be continuing to support GraphQL-Ruby's allowed customizations there -- the complexity of backwards compatibility _might_ be too much, but it's worth a look. If you're interested in pursuing it, please open a new issue describing the desired improvement and we can continue working from there.

PR #4803, opened 2024-01-23
Fix Dataloader testing examples

rubendinho:
Update example per https://github.com/rmosolgo/graphql-ruby/pull/4795/files#r1462039017


rmosolgo:
Oops -- thanks!

PR #4804, opened 2024-01-23
Allow GraphQL::Analysis::AST::FieldUsage to detect use of deprecated enum value as argument

lukes:
Hi, @rmosolgo! What do you think about this change? Thank you!

### DIFF
--- a/spec/graphql/analysis/ast/field_usage_spec.rb
+++ b/spec/graphql/analysis/ast/field_usage_spec.rb
@@ -138,6 +138,20 @@ describe GraphQL::Analysis::AST::FieldUsage do
     end
   end
 
+  describe "query with deprecated enum argument" do
+    let(:query_string) {%|
+      query {
+        fromSource(source: YAK) {
+          id
+        }
+      }
+    |}
+
+    it "keeps track of deprecated arguments" do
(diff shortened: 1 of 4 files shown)


lukes:
I noticed there isn't support for input objects for mutations, but wasn't able to work out how to fix that so I've raised an issue: https://github.com/rmosolgo/graphql-ruby/issues/4805. 🙏

spec/graphql/analysis/ast/field_usage_spec.rb
```diff
@@ -138,6 +138,20 @@
...
+    let(:query_string) {%|
+      query {
+        fromSource(source: YAK) {
+          id
+        }
+      }
+    |}
+
+    it "keeps track of deprecated arguments" do
+      assert_equal ['Query.fromSource.source'], result[:used_deprecated_arguments]
```
lukes, line 151: The interesting thing is `Query.fromSource.source` is the argument name, but it's the argument value `YAK` in this case that's deprecated. I wonder if there's a path we should return that is more explicit about the problem?

rmosolgo:
I pushed a change that: 

- Added a new Set for deprecated enum values 
- Improved handling of non-null and list argument values 
- And after that, the example from #4805 worked 🎉

lukes:
@rmosolgo Thank you very much 🙏!

PR #4807, opened 2024-01-25
Deprecate returning `.resolve(...)` dataloader requests

rmosolgo:
This was previously supported, but it shouldn't be -- it makes Dataloader dependent on the lazy-resolve system. I'm interested in working on that system so I'd like to disentangle them first.

### DIFF
--- a/lib/graphql/dataloader/request.rb
+++ b/lib/graphql/dataloader/request.rb
@@ -14,6 +14,11 @@ module GraphQL
       def load
         @source.load(@key)
       end
+
+      def load_with_deprecation_warning
+        warn("Returning `.request(...)` from GraphQL::Dataloader is deprecated, use `.load(...)` instead. (See usage of #{@source} with #{@key.inspect}).")
+        load
+      end
     end
   end
 end
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1404,7 +1404,7 @@ module GraphQL
           else
             @lazy_methods = GraphQL::Execution::Lazy::LazyMethodMap.new
             @lazy_methods.set(GraphQL::Execution::Lazy, :value)
-            @lazy_methods.set(GraphQL::Dataloader::Request, :load)
+            @lazy_methods.set(GraphQL::Dataloader::Request, :load_with_deprecation_warning)
           end
         end
         @lazy_methods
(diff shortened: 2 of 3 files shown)


Bajena:
Hi @rmosolgo, I'm trying to figure out how to migrate from the deprecated `.request` to `.load` calls in my code.

We're currently using a dataloader source is Apollo's `resolve_reference` (https://github.com/Gusto/apollo-federation-ruby?tab=readme-ov-file#reference-resolvers) and the things are batch loaded correctly. However, as soon as I switch to `load`, each record gets loaded separately.

 Is the deprecation there, because there'll be some additional work done in the future? Or maybe we're doing something wrong? 🤔

rmosolgo:
> additional work done in the future 

Yes, my goal is to move GraphQL-Ruby's "lazy resolution" outside the core runtime, so that it doesn't add overhead to schemas that don't use it. (Lazy resolution is a pre-Dataloader approach to batch loading, used by GraphQL-Batch among others: https://graphql-ruby.org/schema/lazy_execution.html). Dataloader was hooked up to that but I think it doesn't have to be. 

I looked over the docs there but I don't quite see the usage of `.request`. Could you share some example code where `.request` is used? I'd be happy to look for a way to migrate it to `.load` (or perhaps `.load_all` in this case?) and make sure there's a way forward.

Bajena:
@rmosolgo I only found `request` here: https://graphql-ruby.org/dataloader/overview.html

Also, thank you for a very detailed explanation. I'll keep the deprecation warning for now and will be observing the development 🍿 :)

utay:
Actually I was able to make it work with `resolve_references` and `.load_many`!

`resolve_references` (plural) is not documented in README but it's working well.

rmosolgo:
:+1: Thanks for sharing what you found, @utay. I was about to make the same suggestion. It's here in the source: 

https://github.com/Gusto/apollo-federation-ruby/blob/51af77bce4e48e864e97b849053f1e5992edcd6a/lib/apollo-federation/entities_field.rb#L62-L63

You could use `.load_all` in that method, for example: 

```ruby 
def self.resolve_references(references, context)
  asset_ids = references.map { |r| r[:assetId] }
  context.dataloader.with(RecordLoader, Card).load_all(asset_ids) 
end 
```

Let me know how it goes if you give it a try!

PR #4811, opened 2024-01-26
Accept input objects when required arguments aren't provided but have defaults

### LINKED ISSUE #4728: Required input with default value cannot be omitted
**Describe the bug**

I'm only about 80% sure this is actually a bug (vs me misunderstanding the spec), cause I know this area has been the source of a lot of confusion, both in writing the spec and in implementations.

But if I define
```ruby
argument :something,
      Boolean,
      required: true,
      default_value: true,
```
resulting in
```graphql
something: Boolean! = true
```

I expect that I can omit that argument from the input of my query/mutation, and it will be treated as true. Instead, graphql-ruby produces an error:
```
Argument 'something' on InputObject 'WhateverInput' is required. Expected type Boolean!
```

**Versions**

`graphql` version: 2.0.27

rmosolgo:
Fixes #4728

### DIFF
--- a/lib/graphql/static_validation/literal_validator.rb
+++ b/lib/graphql/static_validation/literal_validator.rb
@@ -110,7 +110,7 @@ module GraphQL
         # TODO - would be nice to use these to create an error message so the caller knows
         # that required fields are missing
         required_field_names = @warden.arguments(type)
-          .select { |argument| argument.type.kind.non_null? && @warden.get_argument(type, argument.name) }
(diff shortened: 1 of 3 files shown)


rmosolgo:
Honestly, this had me wondering, "What is the point of a non-null argument with a default value?" But I re-read the spec and realized that it prevents you from passing `null` as a given value. I tested that this works according to the spec at 38f6bb9f4

PR #4816, opened 2024-01-30
Improve test helper docs

rubendinho:
Very excited about the new test helpers. I ran into a few issues getting it wired up and would love to help improve the documentation more. 

For instance, it would be great to document any additional set up / context required to use `with_resolution_context`. For example, the `with_resolution_context` method is erroring in the Object Cache Tracer:

```sh
from .../graphql-enterprise-1.3.2/lib/graphql/enterprise/object_cache.rb:521:in `trace'
```


rmosolgo:
Thanks for this fix! 

Sorry about the trouble with ObjectCache -- part of that tracer assumes that the top-level trace event has already occurred, but that's not true in this case. I improved the check so it will skip that code when it's not in a full query context and released the fix in graphql-enterprise v1.3.3. Please let me know if you run into any more trouble with it!

PR #4834, opened 2024-02-08
Don't allow line terminators in single-quoted strings

### LINKED ISSUE #4397: String literals containing `\u000A` should fail to parse
**Describe the bug**

At present, the GraphQL Ruby language parser does not handle inline string literals according to spec, because it permits `\u0004` (newlines) within single-quoted strings. This is a significant disagreement when aligning with other GraphQL tooling that parses source documents according to spec.

From [the spec](https://spec.graphql.org/October2021/#sec-String-Value):

> StringCharacter ::
>   - SourceCharacter but not `"` or `\` or `LineTerminator`
>   - `\u` EscapedUnicode
>   - `\` EscapedCharacter
>
> LineTerminator
>  - New Line (U+000A)
>  - Carriage Return (U+000D)New Line (U+000A)
>  - Carriage Return (U+000D)New Line (U+000A)

Unfortunately this is also a breaking change as sources that used to parse will now fail. Steps to resolution:
- Parameterize parser accordingly.
- Hook breaking behavior to a schema-level setting for spec-compliant behaviors. Should also include:
  - https://github.com/rmosolgo/graphql-ruby/issues/4403
  - https://github.com/rmosolgo/graphql-ruby/issues/4351

**GraphQL schema**

```ruby
require "graphql"

class ExampleSchema < GraphQL::Schema
  class Mutation < GraphQL::Schema::Object
    field :send_email, String, null: false do
      argument :message, String, required: true
    end

    def send_email(message:)
      message
    end
  end

  mutation Mutation
end
```

**GraphQL query**

This input uses a block string wrapped in `"""`, and works as expected...

```ruby
puts ExampleSchema.execute(
  query: %|
    mutation {
      sendEmail(message: """
        Hello,
          World!

        Yours,
          GraphQL.
      """)
    }
  |
).to_h.dig("data", "sendEmail")
```

However, this input uses a basic string wrapped in `"`, and should therefore fail to parse because it contains `\u0004` characters...

```ruby
puts ExampleSchema.execute(
  query: %|
    mutation {
      sendEmail(message: "
        Hello,
          World!

        Yours,
          GraphQL.
      ")
    }
  |
).to_h.dig("data", "sendEmail")
```

**Expected behavior**

We should get a consistent parse error based on the GraphQL specification. It's bad for our ecosystem when GraphQL Ruby allows non-spec inputs that are not compatible with other spec implementations of GraphQL.

**Actual behavior**

GraphQL Ruby treats `"""` and `"` string delimiters interchangeably.

rmosolgo:
Fixes #4397 

TODO: 

- [x] Add some way to retain the previous behavior, for compatibility

  I added a method which will add any missing escaped, so you could use this pre-process strings if you need to support this legacy behavior. For example: 
  
  ```ruby 
  query_str = params[:query] 
  escaped_query_str = GraphQL::Language.escape_single_quoted_newlines(query_str)
  if !query_str.equal?(escaped_query_str) # If the same string was returned, then nothing was escaped.
    Stats.count("legacy_graphql_newline") # somehow log that this is still in use.... 
    query_str = escaped_query_str 
  end 
  ```


sealabcore:
Getting failing tests from this gem version. Would be nice to have a deprecation release before the breaking change was implemented so we could use the new method `GraphQL::Language.escape_single_quoted_newlines(query_str)` to see if we have any offenses in production and fix accordingly before shipping the breaking change gem version to production.

rmosolgo:
Hey, sorry for the trouble! That _would_ have been a smoother transition, sorry I didn't think of it 😖 

If you want to backport that method into your app, you could roll back to the previous version and paste it in. Here's the source: 

https://github.com/rmosolgo/graphql-ruby/blob/d9300c46536693f9c35bf2915a72723809d9b4d2/lib/graphql/language.rb#L38-L73

PR #4837, opened 2024-02-13
Update doc, add missing parameter

fwolfst:


guides/testing/helpers.md
```diff
@@ -49,7 +49,7 @@ Additionally, it accepts some keyword arguments:
 You can use {{ "Testing::Helpers#with_resolution_context" | api_doc }} to use the same type, runtime object, and GraphQL context for multiple field resolutions. For example:
 
 ```ruby
-with_resolution_context(type: "Post", object: example_post, context: { current_user: author }) do |rc|
+with_resolution_context(YourSchema, type: "Post", object: example_post, context: { current_user: author }) do |rc|
```
rmosolgo, line 52:
```suggestion
# Assuming `include GraphQL::Testing::Helpers.for(MySchema)`
# was used above ...
with_resolution_context(type: "Post", object: example_post, context: { current_user: author }) do |rc|
```

rmosolgo (commented):
Hey, thanks for suggesting a clarification here. It _should_ work as written if the schema-based helper module was included as documented above on the page. 

How about adding a note about that instead? I suggested one possible approach.

The helper module reduces some repetition in these calls so I'd like to recommend it to readers.

fwolfst:
> Hey, thanks for suggesting a clarification here. It _should_ work as written if the schema-based helper module was included as documented above on the page.
> 
> How about adding a note about that instead? I suggested one possible approach.
> 
> The helper module reduces some repetition in these calls so I'd like to recommend it to readers.

Right. Like this, or even not as a comment, but as a statment (I tend to overlook comments ;) ).

rmosolgo:
Thanks for the improvement!

Issue #4403, opened 2023-03-24
Type comparison missing from `FieldsWillMerge` validation


gmac:
**Describe the bug**

[FieldsWillMerge.find_conflict](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/rules/fields_will_merge.rb#L213-L247) omits the [type parity check](https://github.com/graphql/graphql-js/blob/main/src/validation/rules/OverlappingFieldsCanBeMergedRule.ts#L620-L631) implemented by `graphql-js`. This deviates the GraphQL Ruby implementation from spec compliance. Unfortunately, fixing this issue is a breaking change because existing queries will start getting rejected with a validation error.

This issue has been discussed and some work has been done. Steps to resolution:
- Partial fix by @rmosolgo done in https://github.com/rmosolgo/graphql-ruby/pull/4347.
- Some adjustments needed; `fields?` checks should use `leaf?` instead.
    
**GraphQL schema**

```ruby
class TestSchema < GraphQL::Schema
  class Decimal < GraphQL::Schema::Scalar
    graphql_name "Decimal"
  end

  class MoneyV2 < GraphQL::Schema::Object
    graphql_name "MoneyV2"
    field :amount, Decimal, null: false
  end

  class PricingPercentageValue < GraphQL::Schema::Object
    graphql_name "PricingPercentageValue"
    field :percentage, Float, null: false
  end

  class PricingValue < GraphQL::Schema::Union
    graphql_name "PricingValue"
    possible_types MoneyV2, PricingPercentageValue

    def self.resolve_type(obj, ctx)
      {
        "MoneyV2" => MoneyV2,
        "PricingPercentageValue" => PricingPercentageValue,
      }.fetch(obj[:__typename])
    end
  end

  class Query < GraphQL::Schema::Object
    graphql_name "Query"
    field :pricing_values, [PricingValue], null: false

    def pricing_values
      [
        { amount: "1.0", __typename: "MoneyV2" },
        { percentage: 1.0, __typename: "PricingPercentageValue" }
      ]
    end
  end

  query Query
end
```

**GraphQL query**

```ruby
query = <<~GRAPHQL
  {
    pricingValues {
      __typename
      ...on MoneyV2 {
        discountValue: amount
      }
      ...on PricingPercentageValue {
        discountValue: percentage
      }
    }
  }
GRAPHQL

result = TestSchema.execute(query, validate: true)
pp result.to_h
```

[...]

PR #4843, opened 2024-02-15
Complexity cost bug fixes

gmac:
I noticed a `v3` milestone, so thought this would be a good time to submit a fix for several bugs in the complexity algorithm that have the potential to increase costs (_slightly and rarely_) and thus would technically be breaking.

### Here's what's been weird...

1. When multiple composite field possibilities merged, only the _final_ selection was fully costed out. This made selection order matter, and defeated idempotence. The faulty assumption here was actually [called out in a comment](https://github.com/rmosolgo/graphql-ruby/blame/5eac86139138332d17d4eb5ea3f59ae8e0aa8cfb/lib/graphql/analysis/ast/query_complexity.rb#L216C50-L216C77) for years.

```graphql
{
   entity {
     ...on Producer { cheese { kind } } # `cheese` costs 10 on Producer
     ...on Farm { cheese { kind } } # `cheese` costs 5 on Farm
     
      # last costed possibility wins => 5 + 1 + 1
      # should be MAX of possibilities => 10 + 1 + 1
   }
}
```

2. The same issue as above also happened with merged leaf fields, but for a different reason. Leaf possibilities always [directly assigned their field cost](https://github.com/rmosolgo/graphql-ruby/blob/33bc196023e471b602d97d7f82aa8f16b6159d40/lib/graphql/analysis/ast/query_complexity.rb#L159) rather than hill-climbing for a maximum. Once again, last selection wins.

3. Leaf fields would [trigger hooks N-times](https://github.com/rmosolgo/graphql-ruby/blob/33bc196023e471b602d97d7f82aa8f16b6159d40/lib/graphql/analysis/ast/query_complexity.rb#L160) for each leaf _possibility_, when really we should expect only one hook per field.

4. Compounding upon item 3, there was also a crazy side effect here from https://github.com/rmosolgo/graphql-ruby/issues/4403. Because `FieldsWillMerge` was faulty about type consistency (fix slated for `3.0`), it meant that the fields being merged could intermix leaf and composite scopes – which [each had their own workflow for triggering hooks](https://github.com/rmosolgo/graphql-ruby/blob/33bc196023e471b602d97d7f82aa8f16b6159d40/lib/graphql/analysis/ast/query_complexity.rb#L160-L173). So, in addition to leaf field hooks triggering multiple times, an additional hook could trigger once for composite possibilities.


lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -145,35 +149,38 @@ def merged_max_complexity(query, inner_selections)
...
-              if child_scope.empty?
-                field_cost = child_scope.own_complexity(0)
-                field_complexity(child_scope, max_complexity: field_cost, child_complexity: nil)
+            # Collect all child scopes for this field key;
+            # all keys come with at least one scope.
+            child_scopes = inner_selections.filter_map { _1[field_key] }
+
+            # Compute maximum possible cost of child selections;
+            # composites merge their maximums, while leaf scopes are always zero.
+            # FieldsWillMerge validation assures all scopes are uniformly one or the other.
```
gmac, line 158: With `FieldsWillMerge` getting fixed, this should now be consistent about always having scopes of the same kind. Either way, it's written to be fault-tolerant in the event that it's run without first validating.

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -145,35 +149,38 @@ def merged_max_complexity(query, inner_selections)
...
+            maximum_children_cost = if child_scopes.any?(&:composite?)
+              merged_max_complexity_for_scopes(query, child_scopes)
+            else
+              0
+            end
+
+            # Identify the maximum cost and scope among possibilities
+            maximum_cost = 0
+            maximum_scope = child_scopes.reduce(child_scopes.last) do |max_scope, possible_scope|
```
gmac, line 167: This switches to always hill-climb for the maximum possibility among scopes, regardless if they are leaf or composite scopes.

lib/graphql/analysis/ast/query_complexity.rb
```diff
@@ -145,35 +149,38 @@ def merged_max_complexity(query, inner_selections)
...
 
-            if composite_scopes
-              child_complexity = merged_max_complexity_for_scopes(query, composite_scopes)
-
-              # This is the last composite scope visited; assume it's representative (for backwards compatibility).
-              # Note: it would be more correct to score each composite scope and use the maximum possibility.
-              field_cost = composite_scopes.last.own_complexity(child_complexity)
-              field_complexity(composite_scopes.last, max_complexity: field_cost, child_complexity: child_complexity)
-            end
+            field_complexity(
```
gmac, line 177: Lastly, always trigger exactly one hook per field with the resolved maximums among scopes, costs, and children costs. Before leaf scopes would provide `nil` as their `child_complexity` – now they get `0`. This feels more inline with the `own_complexity` call that has always been sent to leaf scopes with zero child cost.

rmosolgo:
Hey, thanks so much for this detailed report and fix. As you can see, it's been a while since I dug into the complexity implementation! I'll review these closely soon.

rmosolgo:
I just updated this branch with master and started on a migration scheme to help people cross from legacy to future algorithm. I'll add some proper tests for the migration code before I merge this.

lib/graphql/analysis/query_complexity.rb
```diff
@@ -13,7 +13,32 @@ def initialize(query)
...
+        when nil
+          subject.logger.warn <<~GRAPHQL
+            GraphQL-Ruby's complexity cost system is getting some "breaking fixes" in a future version. See the migration notes at https://graphql-ruby.org/api-docs/#{GraphQL::VERSION}/Schema.html#complexity_cost_cacluation_mode-class_method
+
+            To opt into the future behavior, configure your schema (#{subject.schema.name ? subject.schema.name : subject.schema.ancestors}) with:
+
+              complexity_cost_calculation_mode(:future) # or `:legacy`, `:compare`
+
+          GRAPHQL
+          max_possible_complexity
```
thlacroix, line 38: Hi @rmosolgo , you mention that there is a need to "opt into the future behavior", but from what I can see here, `:future` is already the new mode by default if nothing is configured explicitly ? (by the way our logging system got spammed by the warning above, but we quickly noticed it)

rmosolgo: Derp, you're right -- I've opened #5343 to address it.

PR #4845, opened 2024-02-15
Add typing to createAblyHandler onError handler

### LINKED ISSUE #4832: JavaScript client should use strong typing for `onError` argument
**Describe the bug**

The typings don't currently specify what to expect from `onError`: https://github.com/rmosolgo/graphql-ruby/blob/35230d37fe0ec3c04af82eb3b1597631758d3fe3/javascript_client/src/subscriptions/createAblyHandler.ts#L12

For Ably it's de-facto an array of objects: https://github.com/rmosolgo/graphql-ruby/blob/35230d37fe0ec3c04af82eb3b1597631758d3fe3/javascript_client/src/subscriptions/createAblyHandler.ts#L56-L57

It would probably be better for the array to be wrapped in an `Error` instance, but even if you prefer to leave things as they are, it would be great to have stronger typings here so that implementing robust error handling code doesn't require being prepared for any kind of object.

**Versions**

`graphql-ruby-client` version: 1.13.0

rmosolgo:
Fixes #4832 

This typing is a bit yucky, but it describes the _current_ usage of `onError`: sometimes, it receives the `errors: ...` from GraphQL; sometimes, it receives a raised error; sometimes, it receives a wrapped network error from Ably. 

cc @jscheid your thoughts?

### DIFF
--- a/javascript_client/src/subscriptions/__tests__/createAblyHandlerTest.ts
+++ b/javascript_client/src/subscriptions/__tests__/createAblyHandlerTest.ts
@@ -1,4 +1,4 @@
-import { createAblyHandler } from "../createAblyHandler"
+import { OnErrorData, createAblyHandler } from "../createAblyHandler"
 import { Realtime, Types } from "ably"
 
 const dummyOperation = { text: "", name: "" }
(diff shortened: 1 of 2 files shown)


rmosolgo:
🚢  in graphql-ruby-client 1.13.1!

PR #4846, opened 2024-02-16
Fix compatibility with Ruby 3.4-dev

casperisfine:
- Backtrace rendering changes, so the `caller` parsing needed to be improved.
- Minitest needed an upgrade.
- Base64 is no longer bundled.


rmosolgo:

PR #4853, opened 2024-02-21
Add support for transaction names in Sentry tracing

porter77:
This essentially just copies the support built into New Relic tracing for setting the transaction name to Sentry tracing


rmosolgo:
Thanks for this improvement!

PR #4856, opened 2024-02-23
Handle unknown trace modes at runtime

### LINKED ISSUE #4855: Previously-uninitialized trace modes don't include the default tracers
**Describe the bug**

(moved from https://github.com/Shopify/graphql-batch/issues/169)

Our schema class includes up to two tracing modules depending on the rails env:
```ruby
  if Rails.env.production?
    trace_with(GraphQL::Tracing::DataDogTrace, mode: :full_tracing)
  end

  # under the hood this calls `trace_with` as of https://github.com/Shopify/graphql-batch/pull/164
  # it does not specify a trace mode
  use GraphQL::Batch
```

Our execution flow then does:
```ruby
context[:trace_mode] = :full_tracing if debug_flag_set?
result = Schema.execute(query, variables: variables, context: context, operation_name: operation_name)
```

The intent is, of course, that in prod we can turn datadog tracing on and off with a debug flag, but that the graphql-batch tracer should *always* be present, whether debug flag is set or not, whether production or development. This lines up with the docs:
> Any custom trace modes also include the default trace_with ... modules (that is, those added without any particular mode: ... configuration).

However, in a non-prod environment when a query has the debug flag set, things don't seem to work:
- the schema does not have the datadog tracer loaded, and thus does *not* a-priori know about `full_tracing` mode
- the query *does* set `context[:trace_mode] = :full_tracing` because of the debug flag

As best I can tell, this causes the graphql-batch tracer to not execute, leading to an exception as soon as any field in the query tries to do a batch load:
> Cannot create loader without an Executor. Wrap the call to for with GraphQL::Batch.batch or use GraphQL::Batch::SetupMultiplex as a query instrumenter if using with graphql-ruby

Somehow the schema seeing an unknown trace mode causes *no* traces to execute, not even the defaults, which seems incorrect?

I'll also happily admit that our tracing setup is a bit convoluted and we could probably simplify it to avoid this bug, but I'd ultimately still expect this setup to work.

**Versions**

`graphql` version: 2.1.11
`rails`: 7.1.3
`graphql-batch`: 0.5.4

rmosolgo:
Oops, there was some code here to handle unexpected mode names, but it didn't really work (and wasn't tested, etc). This makes it so that, when a Schema receives an unknown trace mode at runtime, it treats it like a `:default` mode. It creates a class and options hash for it just like it was registered ahead of time, both derived from `:default`.

Fixes #4855

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -167,7 +167,7 @@ module GraphQL
           backtrace_class.include(GraphQL::Backtrace::Trace)
           trace_mode(:default_backtrace, backtrace_class)
         end
-        trace_class_for(:default)
+        trace_class_for(:default, build: true)
       end
 
       # @return [Class] Return the trace class to use for this mode, looking one up on the superclass if this Schema doesn't have one defined.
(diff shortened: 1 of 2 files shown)


eapache-opslevel:
Code makes sense to me at the surface-level, and it does seem to fix the immediate problem we are seeing. Thanks!

PR #4860, opened 2024-02-28
Backport the compatibility fix for Ruby 3.4-dev to 1.13.x

maxprokopiev:
Backporting https://github.com/rmosolgo/graphql-ruby/pull/4846 to `1.13.x`.

I see a bunch of tests are failing, but seems like they are failing in `1.13.x` in general? (confirmed that in https://github.com/maxprokopiev/graphql-ruby/pull/2)

If this is fine and merged, I would also love to do the same for `2.0.x` and `2.1.x`.

### DIFF
--- a/graphql.gemspec
+++ b/graphql.gemspec
@@ -24,16 +24,18 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{lib}/**/*", "MIT-LICENSE", "readme.md", ".yardopts"]
 
+  s.add_runtime_dependency "base64"
+
   s.add_development_dependency "benchmark-ips"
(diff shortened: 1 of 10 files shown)


rmosolgo:
Thanks! I just shipped 1.13.22 with this patch. Let me know if you have any trouble with it. I'd be happy to release 2.0.x and 2.1.x with this patch too, if you don't mind preparing the PRs :+1:

maxprokopiev:
@rmosolgo thank you for merging this! I've created 2 follow-up PRs for remaining versions - https://github.com/rmosolgo/graphql-ruby/pull/4861 and https://github.com/rmosolgo/graphql-ruby/pull/4862 ✨

PR #4865, opened 2024-03-01
Fix fragment skip/include tracking during analysis visitation

gmac:
### What

Fixes a bug in the analysis visitor to make inline fragments and fragment spreads respond to `@skip/include` directives, [per the spec](https://spec.graphql.org/October2021/#sec--skip). This is another piece of housekeeping that may need to go into the `v3` milestone... It's a clear-cut bug to fix, though I suppose it could have implications on custom analyzer expectations out in the wild.

### Why?

At present, analysis visitation only responds to field conditions. Ignoring fragment conditions manifests numerous analysis bugs, not the least of which is over-costing our old friend complexity analysis (where a skipped fragment tree gets included in the query cost). Fixing this issue for complexity is non-breaking, because costs will only go down.

### DIFF
--- a/lib/graphql/analysis/ast/visitor.rb
+++ b/lib/graphql/analysis/ast/visitor.rb
@@ -118,8 +118,12 @@ module GraphQL
         def on_inline_fragment(node, parent)
           on_fragment_with_type(node) do
             @path.push("...#{node.type ? " on #{node.type.name}" : ""}")
+            @skipping = @skip_stack.last || skip?(node)
+            @skip_stack << @skipping
+
             call_on_enter_inline_fragment(node, parent)
             super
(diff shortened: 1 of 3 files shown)


swalkinshaw:
I'd be in favour of just releasing this as normal bug fix since, as you said, 1) it's a clear bug and 2) it only reduces complexity

gmac:
> I'd be in favour of just releasing this as normal bug fix since, as you said, 1) it's a clear bug and 2) it only reduces complexity

I'm 👍 on that. It is a bug. If your analyzer is supposed to follow the document flow, this will only make it more accurate.

rmosolgo:
Thanks for this improvement!

PR #4866, opened 2024-03-01
Revert hasNextPage change when first isn't present

rmosolgo:
This is a PR to revert the behavior change to `hasNextPage` from #4780. 

That PR made it so that `hasNextPage` was always `false` when `first:` wasn't provided. This reverts to the old behavior that `max_page_size` counts as an implicit `first:`, and `hasNextPage` may be true when `max_page_size` is applied.

I'm interested in some :+1: / :-1: on this for your situation.

### DIFF
--- a/lib/graphql/pagination/array_connection.rb
+++ b/lib/graphql/pagination/array_connection.rb
@@ -56,12 +56,12 @@ module GraphQL
             false
           end
 
-          @has_next_page = if first_value && first
-            # There are more items after these items
-            sliced_nodes.count > first
-          elsif before
+          @has_next_page = if before
             # The original array is longer than the `before` index
             index_from_cursor(before) < items.length + 1
+          elsif first
+            # There are more items after these items
+            sliced_nodes.count > first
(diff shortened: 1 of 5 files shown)


engwan:
I agree with this revert. With `max_page_size` set, it is confusing for the server to return `hasNextPage: false` when there are actually more records that aren't being returned.

gemfiles/rails_7.1.gemfile
```diff
@@ -13,5 +13,6 @@ gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
 gem "sequel"
 gem "evt"
 gem "async"
+gem "libev_scheduler"
```
eapache-opslevel, line 16: looks like this line snuck in accidentally?

rmosolgo: yes, thanks! some CI vs Local development weirdness...

rmosolgo:
🚢  in 2.2.12

PR #4869, opened 2024-03-04
Validate incoming orphan_types are object types

### LINKED ISSUE #4858: Validation for orphan types
This doesn't really cause any current problems but right now _any_ kind of type can be set as an `orphan_type` (either on the schema or an interface). However, only object types should be orphan types as the [docs](https://graphql-ruby.org/type_definitions/interfaces.html#orphan-types) describe.

We've had a few instances of developers adding non-object types which could cause a misunderstanding of the feature since it's allowed.

Raise a descriptive error when trying to pass a non-object type to `orphan_types`

We could do this in our own application but I think it belongs in the gem.

If you agree this should be added we could probably contribute it 😄

rmosolgo:
Closes #4858 

After this change, only Object types may be passed to `orphan_types(...)`. This is how it's _supposed_ to work, although it may break some edge cases. The ways I addressed this in tests are: 

- Using `extra_types(...)` to attach random types to the schema 
- Adding fields so that types are in the schema 
- Calling `add_type_and_traverse(types, root: false)`, which is a private method and I can't guarantee it will always work. But it's the right thing for implementing `Schema.load_from_definition`, etc.


swalkinshaw:
The one slight edge case to this is an interface implementing another interface. This would make it more explicit when you have to pass orphan types from one interface to another. But I still believe this change is better because as is, it obscures how GraphQL actually works. To get around this, peopel could do `orphan_types(*SomeInterface::ORPHAN_TYPES)`

rmosolgo:
I think this will be a bigger change than I realized -- I'm going to punt it back to 2.3.0.

PR #4889, opened 2024-03-22
Add `graphql-filters` to related projects

RiccardoAgatea-Moku:
Graphql Filters is a DSL to define fully typed filters for list fields and compile them to Active Record relations. The documentation is here: https://github.com/moku-io/graphql-filters


rmosolgo:
Thanks!

PR #4891, opened 2024-03-27
Fix run_graphql_field with missing query string on dummy query

ryanseys:
Bug: The `dummy_query` object created as part of `run_graphql_field` doesn't set a query string, which results in a `query_context` object that has validation errors.

Later when `object_type.wrap(inner_object, query_context)` is called with that `query_context`, it checks that the field is authorized.

If as part of checking `self.authorized?` on an object, we first ensure that the context is valid i.e. that `context.errors.empty?`, we will find that it isn't because of this missing query string.

Even if this query string isn't used, the `GraphQL::Query` object should be valid, in case there are validations that are done on these objects in a client's code. This is what we are doing in our code, which led me to this bug.

This PR sets a dummy query string value to fix this issue, and adds a simple `self.authorized?` check which ensures that the context is valid i.e. that `context.errors.empty?`

### DIFF
--- a/lib/graphql/testing/helpers.rb
+++ b/lib/graphql/testing/helpers.rb
@@ -41,7 +41,7 @@ module GraphQL
 
       def run_graphql_field(schema, field_path, object, arguments: {}, context: {})
         type_name, *field_names = field_path.split(".")
-        dummy_query = GraphQL::Query.new(schema, "query {}", context: context)
+        dummy_query = GraphQL::Query.new(schema, "{ __typename }", context: context)
         query_context = dummy_query.context
         object_type = dummy_query.get_type(type_name) # rubocop:disable Development/ContextIsPassedCop
         if object_type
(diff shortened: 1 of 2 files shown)


lib/graphql/testing/helpers.rb
```diff
@@ -41,7 +41,7 @@ def initialize(type_name:, field_name:)
 
       def run_graphql_field(schema, field_path, object, arguments: {}, context: {})
         type_name, *field_names = field_path.split(".")
-        dummy_query = GraphQL::Query.new(schema, context: context)
+        dummy_query = GraphQL::Query.new(schema, "query {}", context: context)
```
rmosolgo, line 44: I'm glad this passes _this_ test, but it might still fail somewhere because GraphQL says that selections _must_ exist (https://spec.graphql.org/draft/#sec-Validation.Fields). For example, I think GraphiQL attaches an error to this query:

image


Here's a minimal valid query: 

```suggestion
        dummy_query = GraphQL::Query.new(schema, "{ __typename }", context: context)
```

ryanseys: Yep I wondered about whether I should add that. Added! Thanks for the review :)

ryanseys:
PTAL

rmosolgo:
Thanks for the fix!

PR #4894, opened 2024-03-30
Fix alignment of a value in the "Invalid" column

Austin-X:
As highlighted in blue in the image below, that value is not aligned properly.

image

That value should be in the "Invalid" column, and the changes proposed in this PR will fix this alignment issue.


rmosolgo:
Thanks!

PR #4895, opened 2024-03-30
Fix indentation

Austin-X:
As highlighted in blue in the image below, the "Custom Runtime Directives" has an accidental indent at the beginning.

image

This is because, as highlighted in red in the image below, that `<h3>` is inside of the `<li>`. The changes in this PR make that `<h3>` **not** inside of the `<li>`, therefore fixing the indentation issue.

image


rmosolgo:
Thanks for this fix, nice find :+1:

PR #4897, opened 2024-04-02
Deduplicate TypeName#name and Field#name string

casperisfine:
Looking at deduplicated string reports from out application, GraphQL field and type are at the top:

```
Retained String Report
 203.20 kB    5080  "String"
              4902  graphql-c_parser-1.0.8/lib/graphql/c_parser.rb:73
               172  graphql-2.2.13/lib/graphql/language/lexer.rb:80
  60.77 kB    1485  "ID"
              1271  graphql-c_parser-1.0.8/lib/graphql/c_parser.rb:73
               166  graphql-2.2.13/lib/graphql/language/lexer.rb:80
                19  graphql-client-ee19269df0ee/lib/graphql/client/schema/enum_type.rb:65
  35.32 kB     883  "id"
               518  graphql-c_parser-1.0.8/lib/graphql/c_parser.rb:73
               353  graphql-2.2.13/lib/graphql/language/lexer.rb:80
  34.24 kB     856  "Int"
               800  graphql-c_parser-1.0.8/lib/graphql/c_parser.rb:73
                51  graphql-2.2.13/lib/graphql/language/lexer.rb:80
  33.08 kB     827  "Boolean"
               811  graphql-c_parser-1.0.8/lib/graphql/c_parser.rb:73
                10  graphql-2.2.13/lib/graphql/language/lexer.rb:80
```

I'm very unfamilair with the gem, but looking at which objects retain them I think I managed to locate where best to deduplicate them, but perhaps there is a better way:

```
root path to 0x7da15ef231e0:
                      ROOT (vm)
...
                      0x7da15ef02bc0 (OBJECT: GraphQL::Language::Nodes::VariableDefinition)
                      0x7da15ef6d718 (OBJECT: GraphQL::Language::Nodes::ListType)
                      0x7da15ef6d768 (OBJECT: GraphQL::Language::Nodes::NonNullType)
                      0x7da15ef6d7b8 (OBJECT: GraphQL::Language::Nodes::TypeName)
                      0x7da15ef231e0 (STRING: "String")

harb> rootpath 0x7da15da32f60
root path to 0x7da15da32f60:
                      ROOT (vm)
...
                      0x7da15d9bd7b0 (OBJECT: GraphQL::Schema::Argument)
                      0x7da15da28218 (OBJECT: GraphQL::Language::Nodes::InputValueDefinition)
                      0x7da15d9ff340 (OBJECT: GraphQL::Language::Nodes::NonNullType)
                      0x7da15d9ff390 (OBJECT: GraphQL::Language::Nodes::TypeName)
                      0x7da15da32f60 (STRING: "String")

harb> rootpath 0x7da160835c28
root path to 0x7da160835c28:
                      ROOT (vm)
...
                      0x7da16050c1c8 (OBJECT: GraphQL::Schema::Argument)
                      0x7da1606e0aa8 (OBJECT: GraphQL::Language::Nodes::InputValueDefinition)
                      0x7da160751d70 (OBJECT: GraphQL::Language::Nodes::NonNullType)
                      0x7da160751dc0 (OBJECT: GraphQL::Language::Nodes::TypeName)
                      0x7da160835c28 (STRING: "ID")

harb> rootpath 0x7da15f36ea10
root path to 0x7da15f36ea10:
                      ROOT (vm)
                      0x7da3f6ddfd20 (CLASS: Object)
...
                      0x7da15f132f80 (OBJECT: GraphQL::Schema::Field)
                      0x7da15f36ea10 (STRING: "id")

harb> rootpath 0x7da16016beb0
root path to 0x7da16016beb0:
                      ROOT (vm)
...
                      0x7da15ff16408 (OBJECT: GraphQL::Schema::Field)
                      0x7da16016beb0 (STRING: "clientMutationId")
```

### DIFF
--- a/lib/graphql/language/parser.rb
+++ b/lib/graphql/language/parser.rb
@@ -608,7 +608,7 @@ module GraphQL
       end
 
       def parse_type_name
-        TypeName.new(pos: pos, name: parse_name, filename: @filename, source_string: @graphql_str)
+        TypeName.new(pos: pos, name: -parse_name, filename: @filename, source_string: @graphql_str)
       end
 
       def parse_directives
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -228,6 +228,9 @@ module GraphQL
             raise ArgumentError, "missing second `type` argument or keyword `type:`"
           end
         end
+
+        name = -name if name.is_a?(String)
+
         @original_name = name
         name_s = -name.to_s
 


lib/graphql/language/parser.rb
```diff
@@ -608,7 +608,7 @@ def parse_enum_name
       end
 
       def parse_type_name
-        TypeName.new(pos: pos, name: parse_name, filename: @filename, source_string: @graphql_str)
+        TypeName.new(pos: pos, name: -parse_name, filename: @filename, source_string: @graphql_str)
```
rmosolgo, line 611: I see your trace as `graphql/c_parser` in it, which has its own parser and lexer. So I doubt this would affect the outcome of it. My guess would be here, where the C lexer makes tokens out of identifiers in the string: 

https://github.com/rmosolgo/graphql-ruby/blob/ec1ae00fdd36d5292096e40adb7401c4832bb7ee/graphql-c_parser/ext/graphql_c_parser_ext/lexer.rl#L264

https://github.com/rmosolgo/graphql-ruby/blob/ec1ae00fdd36d5292096e40adb7401c4832bb7ee/graphql-c_parser/ext/graphql_c_parser_ext/lexer.rl#L205-L209

Specifically the  `rb_utf8_str_new(...)` call. Maybe it would benefit from a `str_uminus` ([src](https://github.com/ruby/ruby/blob/4960a598d6bc42adcfd488c7f455b2defcdc8161/string.c#L3084)) call too?

rmosolgo: I see that this one is also a "repeat offender" in your profile: 

https://github.com/rmosolgo/graphql-ruby/blob/ec1ae00fdd36d5292096e40adb7401c4832bb7ee/lib/graphql/language/lexer.rb#L80

rmosolgo (commented):
Hey, thanks for sharing this detailed profile. I bet we can work these out! I shared a few comments above.

lib/graphql/schema/field.rb
```diff
@@ -228,6 +228,9 @@ def initialize(type: nil, name: nil, owner: nil, null: nil, description: NOT_CON
             raise ArgumentError, "missing second `type` argument or keyword `type:`"
           end
         end
+
+        name = -name if name.is_a?(String)
+
         @original_name = name
         name_s = -name.to_s
```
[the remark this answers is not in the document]
casperisfine: Yes, but it's too late because we stored `name` in `@original_name`.

`String#-@` returns a copy (unless the string is already frozen).

rmosolgo:
I added a benchmark to the project and started an improvement on it here: https://github.com/rmosolgo/graphql-ruby/pull/4899

I tried cherry-picking your commit here but I found that it didn't improve the retained strings. Did it fix your benchmark? (I think it wasn't improved because the originally-parsed strings were retained as part of the AST nodes, and the AST nodes were retained with the GraphQL schema definition objects.) 

Do you know if Ruby Strings created with `rb_enc_interned_str` are GC'ed?

casperisfine:
> Did it fix your benchmark?

Nah, I must admit I didn't retest with the patch, it wasa bit of a drive by PR, sorry :/

> Do you know if Ruby Strings created with rb_enc_interned_str are GC'ed?

Yes they aren't immortal. It's just that if it's almost certain no-one else will need the same, an interned string uses a small bit more memory than a regular one. So it's only "amortized" if it actually dedup at least another instance.

rmosolgo:
I just released graphql-c_parser 1.1.0 with the fix from #4899  -- but it will need GraphQL-Ruby 2.3.1 in order to take effect by default. If you want to use it with 2.3.0, you have to manually pass `GraphQL::CParser::SchemaParser`: 

```ruby 
MySchema.to_definition(parser: GraphQL::CParser::SchemaParser)
```

PR #4906, opened 2024-04-10
Update usage.md with updated Apollo client details

imjared:
Include updated details about `@defer` and apollo-client


rmosolgo:
Thanks for the update!

PR #4916, opened 2024-04-15
Update specs to use assert_nil

rmosolgo:
Addressing warnings from minitest


rmosolgo:
I think that's all of them!

PR #4920, opened 2024-04-17
Support lazy enumerators for @stream

### LINKED ISSUE #4907: Support Lazy Enumerators with Stream
Currently `@stream` will defer _child_ selections on list items, but all items are enumerated before any child selections are run. 

A better workflow might be that, when a `@stream`ed list is resolved, there's no call to `.each` right away. Instead, and empty list is put in the response, and then each deferral gets the item it needs and resolves its subfields. 

Here's a script to demonstrate the behavior in question: 

<details>
<summary>`@stream` with lazy enumeration</summary>

<p>

```ruby
require "bundler/inline"

gemfile do
  gem "graphql", path: "./"  # "~>2.2.0"
  gem "graphql-pro", "~>1.0"
end

  class LazyItems
    DATA = [
      OpenStruct.new(name: "Stopwatch"),
      OpenStruct.new(name: "Necktie"),
      OpenStruct.new(name: "Suitcase"),
      OpenStruct.new(name: "Chopstick"),
      OpenStruct.new(name: "Peanut"),
      OpenStruct.new(name: "Screwdriver"),
    ]


    def each
      idx = 0
      while idx < DATA.length
        item = DATA[idx]
        puts "Yielding #{idx}: #{item.name}"
        yield(item)
        idx += 1
      end
      self
    end
  end

  class Item < GraphQL::Schema::Object
    field :name, String

    def name
      puts "Resolving Item#name: #{object.name}"
      object.name
    end
  end

    field :items, [Item]

    def items
      puts "Resolving items"
      LazyItems.new
    end
  end

  query(Query)
  use GraphQL::Pro::Stream
end

query_str = "{ items @stream { name } }"

res = MySchema.execute(query_str)
res.context[:defer].each do |deferral|
  deferral.to_h
end
pp res.to_h
```

</p>
</details>

It _should_ alternate between `yield` and `resolve` messages, but instead, it does all the `yield`s then all the `resolve`s. 

If the implementation supported lazy enumeration, then you could use `@stream` to break up database requests into smaller chunks using a cursor.

rmosolgo:
Fixes #4907 

TODO: 

- [ ] Maintain compatibility with existing batching
- [ ] Add a test for this behavior in particular 
- [ ] Find an implementation other than recursion

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -658,25 +658,21 @@ module GraphQL
             inner_type_non_null = inner_type.non_null?
             response_list = GraphQLResultArray.new(result_name, selection_result, is_non_null)
             set_result(selection_result, result_name, response_list, true, is_non_null)
-            idx = nil
+            idx = 0
+            any_items_resolved = false
             list_value = begin
-              value.each do |inner_value|
-                idx ||= 0
-                this_idx = idx
-                idx += 1
-                if use_dataloader_job
-                  @dataloader.append_job do
-                    resolve_list_item(inner_value, inner_type, inner_type_non_null, ast_node, field, owner_object, arguments, this_idx, response_list, next_selections, owner_type, was_scoped, runtime_state)
-                  end
-                else
-                  resolve_list_item(inner_value, inner_type, inner_type_non_null, ast_node, field, owner_object, arguments, this_idx, response_list, next_selections, owner_type, was_scoped, runtime_state)
+              enumerator = value.to_enum
+              if use_dataloader_job
+                @dataloader.append_job do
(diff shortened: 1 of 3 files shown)


rmosolgo:
Looking at how this played out, I have a few big questions still to work out: 

- Currently, each list item calls the next one via recursion. This will make really big Ruby stacks, hogging memory. A better approach would be "trampolining", something I spiked several years ago, where different steps of execution actually _return_ to the main caller, which dispatches steps. That change might be necessary to support this feature effectively. 
- The current batching behavior expects each list item to  request its own data, then resolve requests for all list items in a single batch. That's impossible with this approach where each item is handled separately. How can I support both without expensive `.is_a?` checks on every object? Maybe detect `@stream` in particular?

rmosolgo:
This is still very much on my radar, but I don't plan to continue work on this branch in particular.

PR #4924, opened 2024-04-18
Add an option to reject numbers followed by names

### LINKED ISSUE #4871: Name start at end of numbers should fail to parse
**Describe the bug**

The GraphQL spec states that number literals (ints and floats) should not be immediately followed by a name start (with no ignored characters in between). However, `graphql-ruby` does not currently return an error for this.

From the [GraphQL spec for integers](https://spec.graphql.org/October2021/#sel-FAFdJPCAACSnBtrB):
> An [IntValue](https://spec.graphql.org/October2021/#IntValue) must not be followed by a . or [NameStart](https://spec.graphql.org/October2021/#NameStart). ... No other [NameStart](https://spec.graphql.org/October2021/#NameStart) character can follow. For example the sequences 0x123 and 123L have no valid lexical representations.

And the [GraphQL spec for floats](https://spec.graphql.org/October2021/#sel-FAFdLTCAACGlB_pR):
> A [FloatValue](https://spec.graphql.org/October2021/#FloatValue) must not be followed by a [NameStart](https://spec.graphql.org/October2021/#NameStart). For example the sequence 0x1.2p3 has no valid lexical representation.

**Versions**

`graphql` version: 2.2.11

**GraphQL query**

Example GraphQL query

```graphql
query IntegerCase {
  foo(x: 1y: 2)
}

query FloatCase {
  foo(x: 1.0y: 2)
}
```

**Steps to reproduce**

```ruby
# integer
GraphQL.parse("{ foo(x: 1y: 2) }")
# float
GraphQL.parse("{ foo(x: 1.0y: 2) }")
```

**Expected behavior**

An error stating that `1y` or `1.0y` is not a valid token in GraphQL.

**Actual behavior**

The query is parsed successfully, with `1y` treated as two separate tokens, an integer `1` and a name `y`.

rmosolgo:
Fixes #4871 

This option is off by default now but will be on by default in GraphQL-Ruby 3.0. You can opt into the new behavior with: 

```ruby
# app/graphql/my_schema.rb
GraphQL.reject_numbers_followed_by_names = true 

# ... 
```

And you can check for invalid query strings using the helper: 

```ruby 
transformed_query_str = GraphQL::Language.add_space_between_numbers_and_names(query_str) 
if transformed_query_str.equal?(query_str) 
  # The old string was returned, query_str is valid 
  # do nothing 
else 
   # A new, modified query string was returned 
   # Track this occurrence and/or use the transformed string
   BugTracer.report(:invalid_query_string, query_str)
   query_str = transformed_query_str
end 
```

TODO: 

- [x] Improve tests and implementation of string fix method to handle exponent notation
-  ~~Also make sure variable definitions are covered, eg `query($a: Int = 5$b: Int = 6) { ... }`~~
  - I think that variable definitions and values followed by directives are technically _legal_ because the following character is either `@` or `$`, not a name. I _did_ add tests for input object literals, and they were covered by the current implementation.
- [x] Are there other places where client-provided values appear, where there might be a number literal? (see above)
- [x] Add a CI build which runs with `GraphQL.reject_numbers_followed_by_names = true`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,12 +54,12 @@ jobs:
             ruby: 3.3
           - gemfile: gemfiles/rails_master.gemfile
             ruby: 3.3
-            graphql_reject_numbers_followed_by_name: 1
+            graphql_reject_numbers_followed_by_names: 1
     runs-on: ubuntu-latest
     steps:
     - run: echo BUNDLE_GEMFILE=${{ matrix.gemfile }} > $GITHUB_ENV
-    - run: echo GRAPHQL_REJECT_NUMBERS_FOLLOWED_BY_NAME=1 > $GITHUB_ENV
-      if: ${{ !!matrix.graphql_reject_numbers_followed_by_name }}
+    - run: echo GRAPHQL_REJECT_NUMBERS_FOLLOWED_BY_NAMES=1 > $GITHUB_ENV
+      if: ${{ !!matrix.graphql_reject_numbers_followed_by_names }}
     - uses: actions/checkout@v4
     - uses: ruby/setup-ruby@v1
       with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,6 +54,8 @@ jobs:
             ruby: 3.3
           - gemfile: gemfiles/rails_master.gemfile
             ruby: 3.3
+            env:
+              GRAPHQL_REJECT_NUMBERS_FOLLOWED_BY_NAMES: 1
     runs-on: ubuntu-latest
     steps:
     - run: echo BUNDLE_GEMFILE=${{ matrix.gemfile }} > $GITHUB_ENV
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,11 +54,12 @@ jobs:
             ruby: 3.3
           - gemfile: gemfiles/rails_master.gemfile
             ruby: 3.3
-            env:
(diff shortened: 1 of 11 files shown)


ravangen:
I was trying out this approach, using the `INVALID_NUMBER_FOLLOWED_BY_NAME_REGEXP` regex to detect occurrences. A couple initial observations:
- Regex may be running on untrusted input, so we likely should have a timeout (we were periodically hitting a 5s timeout)
  - Is there a way to optimize this?
- False positives in regex: `query($cursor1_pageSize: Int)` is converted to `query($cursor1 _pageSize: Int)` by `add_space_between_numbers_and_names`

rmosolgo:
👋  I took a crack at that specific case in https://github.com/rmosolgo/graphql-ruby/pull/5492. 

Good call on the Regexp timeout. It looks like Rails 8 sets a default one: https://github.com/rails/rails/pull/53490

I'm definitely open to adding one the regular expression when it's supported (Ruby 3.2+, I think), what do you think of that?

ravangen:
Sounds good. Right now I am setting it locally

```ruby
# Defined locally to add a timeout to the regex from the gem.
INVALID_NUMBER_FOLLOWED_BY_NAME_REGEXP = Regexp.new(
  GraphQL::Language::INVALID_NUMBER_FOLLOWED_BY_NAME_REGEXP,
  timeout: 1,
).freeze

def invalid_number_followed_by_name?(query)
  query.query_string.match?(INVALID_NUMBER_FOLLOWED_BY_NAME_REGEXP)
rescue Regexp::TimeoutError
  nil
end
```

PR #4902, opened 2024-04-05
Fix field usage on prepared input types with original value

rmosolgo:
Fixes #4864 

Along the way, I noticed that `InputObject#prepare` was called twice when that method returned itself. By removing the first call, this solution worked out nicely.

PR #4933, opened 2024-04-25
Call prepare on lists of input objects

engwan:
When the argument is a list of input objects we need to call prepare on each item of the list.

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -378,7 +378,7 @@ module GraphQL
           type = type.of_type
         end
 
-        if type.list?
+        if type.list? && !value.nil?
           inner_type = type.of_type
           value.map { |v| recursively_prepare_input_object(v, inner_type) }
         elsif value.is_a?(GraphQL::Schema::InputObject)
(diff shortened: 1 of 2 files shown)


engwan:
@rmosolgo I found a bug related to https://github.com/rmosolgo/graphql-ruby/pull/4902

I'm not sure if there's a better way to check for lists of input objects but I've added a spec to illustrate the bug.

rmosolgo:
Thanks for sharing this fix! Did this _used_ to work before 2.3.1?

engwan:
> Thanks for sharing this fix! Did this _used_ to work before 2.3.1?

Yes, this used to work before 2.3.1 because `coerce_input` called `prepare`.

I am trying to upgrade the GraphQL gem for GitLab and the upgrade broke our tests: https://gitlab.com/gitlab-org/gitlab/-/jobs/6711137035

I did manage to get a workaround working here: https://gitlab.com/gitlab-org/gitlab/-/commit/16e8a1035180f039cf02359b180be799ec813d68. But I think it would be better to have it fixed upstream.

rmosolgo:
Cool, thanks for clarifying. I added a commit to handle lists-of-lists, too! I'm going to investigate one other report of a bug from this change before I release the new version 👀 https://github.com/rmosolgo/graphql-ruby/pull/4902/files#r1579471558

engwan:
Thanks for the quick response!

PR #4935, opened 2024-04-26
Move runtime state into result objects

rmosolgo:
This is an experiment in managing runtime state. 

We already have a lot of per-field metadata in `current_runtime_state`, and most of it is derived from `GraphQLResult` objects. I think it should be possible to: 

- Move per-object state to `GraphQLResult` objects (not _everything_ -- can't be per-field since leaf values don't get `GraphQLResult` objects)
- Remove per-object state values from runtime method signatures, since they're accessible from the passed-in `GraphQLResult` 
- Update `current_runtime_state` to use GraphQLResult in more cases (it already does for path, at least) 

If _that_ works, then I'm hoping that `GraphQLResult` can be the basis for a run-and-return (trampoline) -style execution instead of a recursive one, where the `GraphQLResult` itself contains information about what needs to be done to it next.


rmosolgo:
I checked the benchmarks to make sure there weren't any surprises. The only thing is that Query results now retain references to their underlying GraphQL object type instances. That's the same number of allocations, they just might live a little longer.

PR #4938, opened 2024-04-29
Fix input object spec re missing optional arguments

sporto:
`values.key?([:b])` would always return false.
Should be `values.key?(:b)`


rmosolgo:
Thanks for correcting this!

PR #4967, opened 2024-05-23
Try using hardcoded queues for runtime steps

rmosolgo:
GraphQL-Ruby's runtime has a lot of general-purpose code for concurrency management (`after_lazy { ... }`, `dataloader.append_job { ... }`) and relies on Ruby call stacks for control flow. I'm investigating using dedicated data structures to manage runtime steps instead, with the hope that: 

- It will be faster because there won't be as many block-to-proc conversions 
- It will be faster because it can be more selective about when to use `after_lazy` and `append_job`
- It will be more flexible, so I could handle `@stream`-ing lazy enumerators
- It will be easier to debug since there won't be as many deep call stacks 

Some problems I can see are: 

- Adding object allocations to runtime code will slow things down; will this offset the possible speed-ups? 
- It's always hard to maintain compatibility when modifying this code 
- I might be wrong about this approach even working at all 

My first experiment will be argument resolution, which is a microcosm of runtime. It needs dataloading, lazy resolution, and a sequence of steps. I'm going to iterate on this to see if I can make a faster, more flexible implementation. If it works, I'll try expanding it.

### DIFF
--- a/lib/graphql/execution/interpreter/arguments_cache.rb
+++ b/lib/graphql/execution/interpreter/arguments_cache.rb
@@ -35,7 +35,42 @@ module GraphQL
               @storage[argument_owner][parent_object][ast_node] = resolved_args
             end
           end
+        end
 
+        class RunQueue
+          def initialize(parent: nil)
+            @parent = parent
+            @steps = []
+            @callbacks = []
+          end
+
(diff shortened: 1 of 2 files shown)


rmosolgo:
This was a useful exploration, but no intention to continue on this track in particular

PR #4968, opened 2024-05-23
Queue-based query runner

rmosolgo:
Another exploration in re-working execution for better performance and flexibility, see #4967


rmosolgo:
This was a useful exploration, but no intention to continue on this track in particular

Issue #4905, opened 2024-04-09
Possibility to generate GraphQL comments


comatory:
**Is your feature request related to a problem? Please describe.**

The problem is that the generated schema is post-processed for linting purposes. We use eslint-style comments for that but I haven't found a way to add comment to my fields, queries etc. Descriptions do not cut it and they are also semantically different things (we do not want this information to be exposed in registry / introspection).

**Describe the solution you'd like**

Provide a way to enrich definitions to add `#` comments. Having `comment` argument available on these types: type, input type, enum type, fields (inside `Query` type or `Mutation`), fields (including input types), enum members.
Having ability to add top-level `#` comment could also be useful.

**Describe alternatives you've considered**

I considered using GraphQL descriptions `"""` but that has two downsides for me:

1. Not supported by tooling such as eslint, which will only detect comments such as `eslint-disable-next-line` when they're part of GraphQL comment
2. Semantically incorrect - such information would leak to introspection and would be available as documentation in our schema registry.

**Additional context**

N/A

rmosolgo:
Hey, thanks for this suggestion. Yes, I can see how this would be useful, and how comments aren't quite the right thing. 

A third alternative might be to use directives in your schema definition. Those wouldn't necessarily leak to introspection (you could filter them out with `.visible?`), but I guess it still wouldn't work with eslint. 

Implementation-wise, it seems like one approach might be to add `attr_accessor :definition_comment` to the various schema classes (types, fields, arguments, enum values, directives, etc), then check that property when printing out a definition string. 

The catch there is that the current printing flow takes two steps: it first creates an AST, then uses the normal printer to create a string. This might be ok, we'd just have to assign the `definition_comment` to the AST node object (it would need a new property), then read from _that_ when printing a string.

comatory:
> A third alternative might be to use directives in your schema definition. 

Yes that was my go-to solution at first. It wouldn't work out of the box with eslint unfortunately but could be make to work with some effort by creating custom eslint plugin. However, in order to use custom directives, your federation standard must support them and the one I'm on does not (Apollo V1)

> Implementation-wise, it seems like one approach might be to add `attr_accessor :definition_comment` to the various schema classes (types, fields, arguments, enum values, directives, etc), then check that property when printing out a definition string.

Yes extra field on various methods would be what I'm looking for.

In this case, it would really be important to correctly differentiate various AST components. Imagine a field that takes multiple inputs (it does not really matter whether it's a mutation or query):

[...]

PR #4970, opened 2024-05-24
Support for GraphQL comments

comatory:
Resolves https://github.com/rmosolgo/graphql-ruby/issues/4905

This PR is in draft state, we want to get feedback early so we know we're on the right track.

The idea is that we expanded methods that construct the schema with new `comment` parameter (we should cover most of them like types, unions, enums). By passing string value to this new parameter, the resulting schema will contain comments such as this:

```graphql
# we added support for comments
type MyType {
  name: String!
}
```

The reasons for adding this are mentioned in the linked issue but in a nutshell, we want to leverage comments for tooling (`eslint`), using descriptions is not desireable because those would leak into documentation.

The next steps are to add support for comments to every possible place according to GraphQL specification, e.g. on definitions for enums, inputs, unions etc.
At the moment, we are fine with having single-line comments and expand to multi-line comments in subsequent PRs, which means eventually we want this to be possible:

```graphql
# ❌ not yet implemented
# first line
# second line
type MyType {
  name: String!
}
```

We haven't yet looked into edge cases such as mixing descriptions with comments, but eventually those should be supported as well.


rmosolgo:
:+1: I think this is absolutely on the right track! 

Let me know if I can do anything particular to help this work along.

comatory:
@rmosolgo could you please approve CI pipeline run? Thanks

Next steps for us is to implement support for multi-line comments, we'd need to be able to parse them. For printing (output), we were thinking of for allowing newline characters if user wished to have multiline output. This would mean the `comment` argument could accept value such as: `"first line\nsecond line"` and the output would be:

```
# first line
# second line
```

We think it's good enough API for users to format it this way instead of some complex solution, such as allowing to pass lists or having some kind of automated mechanism (e.g. break after 80 chars).

If you think that printing multiple line comments is something that's not needed or wanted, we're OK with not having it.

lib/graphql/language/parser.rb
```diff
@@ -269,6 +275,7 @@ def definition
         else
           loc = pos
           desc = at?(:STRING) ? string_value : nil
+          comment = at?(:COMMENT) ? value : comment
```
comatory, line 278: This code is not very easy to read, to be honest we added it because one of the tests was failing and this one worked.

rmosolgo:
I agree that working from line breaks is the right way to render multiline comments. Ideally, GraphQL-Ruby would be able to parse SDL with multiline comments and print them back out exactly as they were originally written. (For your use case with `es-lint`, that might even be a hard _requirement_ -- we don't want those directives getting merged with the explanatory comments above them, right?)

comatory:
> Ideally, GraphQL-Ruby would be able to parse SDL with multiline comments and print them back out exactly as they were originally written. (For your use case with `es-lint`, that might even be a hard _requirement_ -- we don't want those directives getting merged with the explanatory comments above them, right?)

Yeah exactly, we should respect that during parsing that's for sure. Yes for `eslint`, it's important the specific comment needs to be placed right above the statement it's related to.

We'll try to polish this first and have the tests pass, which will probably require us to figure out multiline comments in the process 👍

comatory:
@rmosolgo 👋 hello, is it possible to approve running pipelines on this PR? We're in the process of fixing specs and it'd be nice if we could see results in CI. 🙇

rmosolgo:
👀  I see the button to allow CI _once_, but I'm not sure how enable it for a PR. I couldn't find a doc from GitHub on how to do it -- do you know how?

comatory:
@rmosolgo Hm I thought there might be some settings in repository. I only found [this](https://stackoverflow.com/a/73317424) but not sure if it helps. Anyway, thanks!

rmosolgo:
Ah... yes, there is a setting there: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories 

I switched it so only "new GitHub accounts" need me to approve them. Hopefully that will do it 🤞

rmosolgo:
Wrong button 🤦

vaclavbohac:
@rmosolgo  Hello, we've got to a point where most of the tests are passing (of course the changes themselves are far from final) but we are stuck on the c-parser integration. It's a little mysterious how it works to us and we could use some guidance how to fix it and what the best approach here would be. Could you help us please?

cc @comatory

rmosolgo:
👋  Sure thing. I'd be happy to take care of the C-Parser changes whenever the Ruby implementation is ready to go. Or, if you're interested in investigating it yourself, I think the main thing is updating the C parser to pass `comment`s when initializing AST nodes. 

First, the lexer would need to be updated to _actually_ emit comment tokens. Currently, they're ignored: 

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/graphql-c_parser/ext/graphql_c_parser_ext/lexer.rl#L385-L388

That `if ...` clause should be removed so that comments appear in the list of tokens.

Then, the parser would need to be updated to match comments wherever they appear. Since descriptions are very similar, structurally, I think a similar approach would work. Define a rule for optionally matching a comment (eg, `comment_opt`), like this one: 

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L521-L525

Then add `comment_opt` wherever a comment may appear in the document, probably before `description_opt` in these examples:

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L539-L540

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L586-L587

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L658-L659

Then, since Yacc uses indexes to reference matched tokens, all the indexes in those rules would have to be updated. (`comment_opt` would be `$1`, so `description_opt` would now be `$2`, and all the others would increment also.)

Finally, you'd have to pass `$1` to the `MAKE_AST_NODE` macro so that the node would be initialized with the comment. This macro calls `.from_a`:

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L15

Finally, you'd have to make sure that the `.from_a` method was ready to receive the new `comment` input. That method is usually generated here: 

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/lib/graphql/language/nodes.rb#L320-L322

But a couple of nodes have custom definitions for one reason or another, so those would need to be manually updated to receive `comment` and pass it as `comment: comment`: 

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/lib/graphql/language/nodes.rb#L411

https://github.com/rmosolgo/graphql-ruby/blob/843014fb47b409614195ead4a605d6b235e3e0fb/lib/graphql/language/nodes.rb#L448

I _think_ that after that, the C Parser would be making ASTs identical to the Ruby parser's ASTs, so we'd be done. But there are always surprises 😅

vaclavbohac:
@rmosolgo Wow, thanks for such a detailed answer. The required work is more or less what I expected we'll need to do, and with the guidance of what you wrote above, I think I will be able to do it.

rmosolgo:
Happy to! Thanks for all your work on this feature so far 🍻

graphql-c_parser/ext/graphql_c_parser_ext/parser.y
```diff
@@ -524,6 +524,12 @@ type_system_definition:
       /* none */      { $$ = Qnil; }
     | description
 
+  comment: STRING
```
rmosolgo, line 527: Ahh... I realize I missed a spot when I described the changes required here. We also need a YACC token type for `COMMENT`. 

The token types are configured in two places. First, they're laid out in an enum in lexer.rl: 

https://github.com/rmosolgo/graphql-ruby/blob/18fdf6f5e9e3b2a6d656798407aaac20857b3519/graphql-c_parser/ext/graphql_c_parser_ext/lexer.rl#L150-L206

It looks like COMMENT is already on the list there :+1: 

Then, the enum values are mapped to YACC token types here: 

https://github.com/rmosolgo/graphql-ruby/blob/18fdf6f5e9e3b2a6d656798407aaac20857b3519/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L61-L98

`COMMENT` is _not_ on the list there, since it was previously excluded from the token stream. It needs to be added. Those integer values are based on their position in the enum declaration in lexer.rl, so I _think_ we need to add `%token COMMENT 240`. 

After that, you could define this rule as `comment: COMMENT`.

rmosolgo:
I did some updates to the C parser to accept comments in https://github.com/rmosolgo/graphql-ruby/commit/a70db539cc55923e8b59afa02bbf2a5add482242 (I would have pushed it to your branch but it looks like "allow commits from maintainers" isn't turned on.) 


It _parses_ again, but there's a new challenge.  When I suggested parsing comments as "real" tokens (instead of ignoring them), I didn't realize that it would require _expecting_ comments anywhere that they could occur. They could occur _anywhere_, so it would be very hard to go back and account for that possibility. Here's an example where a comment breaks the current implementation: 18dd7af84

```diff 
-        query ($sizes: [ImageSize]) # Anonymous inline comment
+        query # Comment between parts of a query
+          ($sizes: [ImageSize]) # Anonymous inline comment
```

I think we've got to go back to ignoring comments, but keeping a reference to the last one that we parsed (or something?) so that we can tack it on to parsed nodes. 🤔  I'll give something a try next week. 

Thanks so much for all your work on this change! I'm going to help get it across the finish line next week so we can get int released.

vaclavbohac:
@rmosolgo Hi, and thank you very much for helping us with this!

I was think about it over the weekend and I thought whether we need to parse the comments at all. I mean I don't see the use case for it.

What we were trying to achieve here is to be able to generate comments into the schema. We never thought we would need it other way around.

We went down the rabbit hole of parsing only because there were certain tests parsing and printing the schema and expecting the structure would be the same. If only we ignored comments there, it would work again.

The thing is the parsing as you mentioned is incredibly difficult to be done right, the comments as you said could be anywhere.

Regarding the contributes, we sent you an invite to the repo, if it doesn't work. I'll make the fork to my personal github and reopen the PR from there as I will have a full control over the settings. Let us know :-)

rmosolgo:
:+1: Sounds good to me to reduce the scope of this change to only go Ruby-to-GraphQL. I forgot that was the _real_ goal here 😅 If we want to add parsing later, we can always do it.

vaclavbohac:
Excellent, I'm on it 👍

vaclavbohac:
FYI I'm working on the changes in a separate branch. I will open the PR once I have most of the work done.

vaclavbohac:
This PR can be closed then as https://github.com/rmosolgo/graphql-ruby/pull/5067 has been merged.

cc @rmosolgo @comatory

rmosolgo:
Thanks again for all your work here ... and sorry I sent you down a road to nowhere 😖!

PR #4974, opened 2024-05-27
Defer loading types until they're needed

### LINKED ISSUE #4919: Lazily load types and fields in a schema

Right now, resolver classes are defined like this:

```rb
# app/graphql/types/query_type.rb
module Types
    field :test_field, resolver: TestFieldResolver
  end
end
```

```rb
# app/graphql/types/test_field_resolver.rb
module Types
  class TestFieldResolver < GraphQL::Schema::Resolver
    description "Test"

    type String, null: false

    def resolve
      "Test"
    end
  end
end
```

In Rails applications with lots of resolvers, this can trigger a lot of autoloads, and contribute to loading a lot of files we don't actually need to execute a query in development mode.

GraphQL types have solved this problem already by wrapping type defs in procs like this:

```rb
# app/graphql/types/test_field_resolver.rb
module Types
  class TestFieldResolver < GraphQL::Schema::Resolver
    description "Test"

    type(proc { String }, null: false)

    def resolve
      "Test"
    end
  end
end
```

I think it would be great if we could have this syntax to delay loading resolvers until actually used:

```rb
# app/graphql/types/query_type.rb
module Types
    field :test_field, resolver: proc { TestFieldResolver }
  end
end
```

But it raises an error:

```
/Users/gannon/.gem/ruby/3.2.0/gems/graphql-2.3.0/lib/graphql/schema/field.rb:158:in `scoped?': undefined method `type_expr' for #<Proc:0x0000000106aa7ec0 /Users/gannon/lazy_demo/app/graphql/types/query_type.rb:6> (NoMethodError)

          resolver_type = @resolver_class.type_expr
                                         ^^^^^^^^^^
	from /Users/gannon/.gem/ruby/3.2.0/gems/graphql-2.3.0/lib/graphql/schema/field.rb:314:in `initialize'
	from /Users/gannon/.gem/ruby/3.2.0/gems/graphql-2.3.0/lib/graphql/schema/field.rb:125:in `new'
	from /Users/gannon/.gem/ruby/3.2.0/gems/graphql-2.3.0/lib/graphql/schema/field.rb:125:in `from_options'
	from /Users/gannon/.gem/ruby/3.2.0/gems/graphql-2.3.0/lib/graphql/schema/member/has_fields.rb:12:in `field'
	from /Users/gannon/lazy_demo/app/graphql/types/query_type.rb:6:in `<class:QueryType>'
	from /Users/gannon/lazy_demo/app/graphql/types/query_type.rb:5:in `<module:Types>'
	from /Users/gannon/lazy_demo/app/graphql/types/query_type.rb:4:in `<main>'
	from <internal:/opt/rubies/3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from <internal:/opt/rubies/3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /Users/gannon/.gem/ruby/3.2.0/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
	from /Users/gannon/.gem/ruby/3.2.0/gems/zeitwerk-2.6.13/lib/zeitwerk/kernel.rb:26:in `require'
	from /Users/gannon/lazy_demo/app/graphql/lazy_demo_schema.rb:3:in `<class:LazyDemoSchema>'
	from /Users/gannon/lazy_demo/app/graphql/lazy_demo_schema.rb:1:in `<main>'
	from <internal:/opt/rubies/3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from <internal:/opt/rubies/3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /Users/gannon/.gem/ruby/3.2.0/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
...
```

I considered opening an issue for a feature that lazily loads the root `query` or `mutation`, but that seems a little more difficult from a public API perspective, and doesn't really solve the problem as granularly as I'd like. It would defer loading _all_ mutation code until the first mutation, and _all_ query code until the first query. Lazy resolvers should theoretically only load the code and types needed to execute a query/mutation while keeping the root types intact.

I also thought about creating a schema for queries and a different one for mutations to split loading the root query and mutation. This would work, but it seems like a hack that doesn't really address the root concern in the GraphQL gem.

**Additional context**

I noticed this problem when profiling my application and seeing that `AppNameSchema.execute(...)` would load both query and mutation roots (and subsequent resolvers). My application's schema is really big and takes ~3 seconds to load the mutation root and ~2 seconds to load the query root, so this feature would really help the time to first query in autoloaded environments.

cc @swalkinshaw

rmosolgo:
GraphQL-Ruby could reduce application boot time by _not_ loading the whole schema upfront. Instead, it could load the schema as-needed. 

To opt into lazy-loaded types: 

- Add `load_types_as_needed` to the _top_ of your Schema class (before any types are configured), for example: 

    ```ruby 
      load_types_as_needed if Rails.env.development? 
      # ... 
    end 
    ```

- Use strings to identify your root types (`query`, `mutation`, and/or `subscription`) and `orphan_types`, using fully-qualified Ruby constant names. For example: 

    ```diff 
    - query Types::Query 
    + query "Types::Query" 
    ```
    
    (If you use Ruby constant literals instead of strings, then Rails will load the constants right away. When you use strings, GraphQL-Ruby will `Object.const_get(...)` the type definition when it needs it.) 

-------

At a minimum, it could load the _whole_ schema the first time it's used. Ideally, it would load only the bits of schema needed for a given query. (Some queries would require the whole schema to be loaded.) 

On the other hand, it's better for a production server to have everything loaded _once_, and before any requests are received (and before any `fork`ing, to avoid duplicates in child process memory). 

TODO: 

- [x] Find tests that broke and fix them 
- [x] Add tests for internal `Schema` state not being loaded until necessary (maybe add an `unload`-type method for testing)
- [x] Migration path: make lazy-loading opt-in in development to avoid nasty surprises 
- ~~Identify calls to `Schema.` methods which require `build_types_hash`, reduce usage of them~~ I think this can be a second step
- [x] Add some APIs to `GraphQL::Schema` so that root types and orphan types can be registered without causing those files to be loaded in Rails
- [ ] Add documentation
- ~~Update `possible_types` to use type classes, not strings, as keys~~ will address in #4975 
- [ ] Check benchmarks to make sure that checking whether types have been loaded isn't too slow

Fixes #4919

### DIFF
--- a/lib/generators/graphql/core.rb
+++ b/lib/generators/graphql/core.rb
@@ -16,7 +16,7 @@ module Graphql
 
       def insert_root_type(type, name)
         log :add_root_type, type
-        sentinel = "defer_loading_types # for faster application boot\n  end\n"
+        sentinel = "load_types_as_needed # for faster application boot\n  end\n"
 
         in_root do
           if File.exist?(schema_file_path)
--- a/lib/generators/graphql/core.rb
+++ b/lib/generators/graphql/core.rb
@@ -20,7 +20,7 @@ module Graphql
 
         in_root do
           if File.exist?(schema_file_path)
-            inject_into_file schema_file_path, "  #{type}(Types::#{name})\n", after: sentinel, verbose: false, force: false
+            inject_into_file schema_file_path, "  #{type}(\"Types::#{name}\")\n", after: sentinel, verbose: false, force: false
           end
         end
       end
--- a/lib/generators/graphql/core.rb
+++ b/lib/generators/graphql/core.rb
@@ -16,7 +16,7 @@ module Graphql
 
       def insert_root_type(type, name)
         log :add_root_type, type
-        sentinel = /< GraphQL::Schema\s*\n/m
+        sentinel = "defer_loading_types # for faster application boot\n  end\n"
 
         in_root do
           if File.exist?(schema_file_path)
--- a/lib/generators/graphql/templates/schema.erb
+++ b/lib/generators/graphql/templates/schema.erb
@@ -5,7 +5,6 @@ class <%= schema_name %> < GraphQL::Schema
   if Rails.env.development?
     defer_loading_types # for faster application boot
   end
-  query("Types::QueryType")
 <% if options[:batch] %>
   # GraphQL::Batch setup:
   use GraphQL::Batch
--- a/lib/generators/graphql/templates/schema.erb
+++ b/lib/generators/graphql/templates/schema.erb
@@ -31,5 +31,9 @@ class <%= schema_name %> < GraphQL::Schema
 
   # Stop validating when it encounters this many errors:
   validate_max_errors(100)
+
+  if Rails.env.production?
+    eager_load_types!
+  end
 end
 <% end -%>
--- a/lib/generators/graphql/templates/schema.erb
+++ b/lib/generators/graphql/templates/schema.erb
@@ -2,7 +2,10 @@
 
 <% module_namespacing_when_supported do -%>
 class <%= schema_name %> < GraphQL::Schema
-  query(Types::QueryType)
+  if Rails.env.development?
+    defer_loading_types # for faster application boot
+  end
+  query("Types::QueryType")
 <% if options[:batch] %>
   # GraphQL::Batch setup:
   use GraphQL::Batch
@@ -31,9 +34,5 @@ class <%= schema_name %> < GraphQL::Schema
 
   # Stop validating when it encounters this many errors:
   validate_max_errors(100)
-
-  if Rails.env.production?
-    eager_load_types!
-  end
 end
 <% end -%>
--- a/lib/generators/graphql/templates/schema.erb
+++ b/lib/generators/graphql/templates/schema.erb
@@ -3,7 +3,7 @@
 <% module_namespacing_when_supported do -%>
 class <%= schema_name %> < GraphQL::Schema
   if Rails.env.development?
-    defer_loading_types # for faster application boot
+    load_types_as_needed # for faster application boot
   end
 <% if options[:batch] %>
   # GraphQL::Batch setup:
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -289,6 +289,7 @@ module GraphQL
       # @param context [Hash]
       # @return [String]
       def to_definition(context: {})
+        build_types_hash
         GraphQL::Schema::Printer.print_schema(self, context: context)
       end
 
@@ -362,9 +363,14 @@ module GraphQL
         visible_types
       end
 
+      def eager_load_types!
+        build_types_hash
+      end
+
       # @param type_name [String]
       # @return [Module, nil] A type, or nil if there's no type called `type_name`
       def get_type(type_name, context = GraphQL::Query::NullContext.instance)
+        build_types_hash
         local_entry = own_types[type_name]
         type_defn = case local_entry
         when nil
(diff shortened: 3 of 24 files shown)


rmosolgo:
Thanks for sharing all this thoughtful feedback ... Still quite a few things to figure out 😅 

`Proc`s are definitely :+1: from me. I'll make sure to include those in the final implementation. 

> doesn't offer much benefit over plain old autoloading of a schema

I guess I was thinking that it _would_ improve cases when you boot the app in development and use non-GraphQL features. But in that case, if your schema is autoloaded, you wouldn't've loaded it anyways. I guess I was thinking back to my old GitHub days, where the schema was always loaded at boot, regardless, but I think that was an application setup issue anyways 😖 . Yes, I can see how this would have to be pushed a little further to be useful for the _normal_ case. 

`use` will be tough. I definitely want everything to keep Just Works :tm:-ing . One possibility -- I almost hate to mention it -- is to capture the inputs to `use` and only actually install those plugins when the types are loaded. We'd probably need incrementally-loaded root types for that to be useful too: if the plugin calls `schema.mutation`, then it has to return the _real_ type -- but maybe it could somehow not have loaded its whole dependency tree... (Would that mean replacing all type/resolver references with Procs? Yuck.) 

Maybe Just Works :tm: isn't an option -- maybe future-compat via callbacks is the way to go, like you suggested.

Decoupling traversal would be huge -- I don't quite have a vision for how that'd _really_ work yet, because given a query like `{ node(id: 5) { ... on Thing { name } } }`, I have to somehow find `Thing`, which may be attached to the schema in many different ways. (The "easiest" solution is a new hook, `Schema.load_type(name)`, and make applications define it, but adding that hiccup to adoption seems bad.)

gmcgibbon:
> Maybe Just Works ™️ isn't an option -- maybe future-compat via callbacks is the way to go, like you suggested.

The way I've always approached large refactors like this is to make it work first, and then make the backwards compatibility happen later. Once we have a better idea of all the changes we need to make, we can worry about compat IMO. 

Further on the warden type traversal problem I was mentioning, I found that the way we call [this method](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/warden.rb#L286) makes it difficult to lazily load root types in general because we have to ask every type we query if it is a root type [here](https://github.com/rmosolgo/graphql-ruby/blob/b4402c989909aafc37967d6d869337c7554565b4/lib/graphql/schema/warden.rb#L359).

I'm switching gears and trying to make shallow schema type additions work like I mentioned [here](https://github.com/rmosolgo/graphql-ruby/issues/4919#issuecomment-2148044437). This way, we can keep the way type resolution works and eliminate the root problem of a root type needing to know about all of its nested types when it is loaded.

Thank you so much for your help with this, I really appreciate it!

rmosolgo:
That `root_type?` method could definitely be refactored somehow ... it was done that way because it's easy with the current implementation, but that might change!

rmosolgo:
This was a worthwhile exercise (for _me_ anyways) but not the right step forward, so I'm closing it 🍻

Issue #4993, opened 2024-06-19
Fiber-local variables are not copied when using AsyncDataloader


igorbelo:
**Describe the bug**

When using the `GraphQL::Dataloader::AsyncDataloader`, Fiber-local variables (`Thread#[]`) are missing in the context of the [source](https://graphql-ruby.org/dataloader/sources) execution.

I was on the fence about it being an actual bug, but since the patch at https://github.com/rmosolgo/graphql-ruby/pull/3461 copied the Fiber-local variables over, I expected AsyncDataloader's behavior to be the same.

I believe the issue is when spawning the source tasks [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/dataloader/async_dataloader.rb#L38). At that point, [get_fiber_variables](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/dataloader/async_dataloader.rb#L75) will be referencing the [root task](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/dataloader/async_dataloader.rb#L36)'s Fiber-local storage.

**Versions**

`graphql`: 2.3.5
`async`: 2.12.0

**Steps to reproduce**

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  ruby '3.3.3'

  gem 'async', '2.12.0'
  gem 'graphql', '2.3.5'
end

class OneSource < GraphQL::Dataloader::Source
  def fetch(_ids)
    [Thread.current[:arbitrary_context]]
  end
end

class Query < GraphQL::Schema::Object
  field :one, String, null: true

  def one
    dataloader.with(OneSource).load(1)
  end
end

class Schema < GraphQL::Schema
  query Query

  use GraphQL::Dataloader::AsyncDataloader
end

Thread.current[:arbitrary_context] = 'a'
result = Schema.execute("query { one }")
puts result.to_h
```

**Expected behavior**

Expected the output of above script to be:
```ruby
{"data"=>{"one"=>"a"}}
```
Which is the same behavior as if the script above was using `GraphQL::Dataloader` instead of `GraphQL::Dataloader::AsyncDataloader`.

**Actual behavior**
```ruby
{"data"=>{"one"=>nil}}
```

**Additional context**

[...]

PR #4994, opened 2024-06-20
Copy Fiber-local variables over to sources when using AsyncDataloader

igorbelo:
Addresses #4993.


rmosolgo:
Thanks for this fix!

PR #4998, opened 2024-06-25
Add an alternative type visiblity/filtering system

rmosolgo:
`GraphQL::Schema::Warden` has been reliably applying type visibility filtering for many years. I'm interested in exploring a new approach for a few reasons: 

- `Warden` requires schema-wide caches of type information, which means that the whole schema must be loaded to run any query. A new approach could load only what's _necessary_. (Part of #4919) 
- A new approach could cache schemas between queries, for example, by name. This could speed up queries using the same subset of the schema. This could be coupled with a DSL for making certain types visible in named subsets.
- A new approach could simplify the expectations and behaviors of filtering, reducing edge case oddities (or at least making it clear how to address them).

This PR starts down that path by adding `GraphQL::Query#types` which uses _either_ `Schema::Subset` or `Schema::Warden`. The `types` object has the same API in either case, but behaves slightly differently. (I expect to widen the gap further in behavior, carefully, so that people can migrate.)

I don't expect any changes to application behavior after merging this PR because `Query#warden` is still available and the default implementation of `Query#types` (which is now used internally by GraphQL-Ruby) behaves the same as Warden (and uses Warden under the hood).

TODO: 

- [x] Re-implement `Schema::Warden` so that one implementation can be tinkered with while the other stays in place 
  - [x] Rename to something nice
  - [x] Update CI to run both implementations
- [x] Audit changes to test suite: revert them or document them 

In later PRs:

- Build out support for lazily-loading types as-needed 
  - Lazy-loading schema root types 
  - Lazily traversing field / argument types, possible types
  - Lazy-loading types by name -- convention-based? 
- Make forward-compatible APIs for plugins, etc, to run code when types are loaded
- Optimize performance
- Build a way to re-use schema subsets
  - Cache subsets between requests 
  - Make it easy to attach types and fields to subsets

### DIFF
--- a/lib/graphql/schema/introspection_system.rb
+++ b/lib/graphql/schema/introspection_system.rb
@@ -69,7 +69,7 @@ module GraphQL
       def resolve_late_bindings
         @types.each do |name, t|
           if t.kind.fields?
-            t.fields.each do |_name, field_defn|
+            t.all_field_definitions.each do |field_defn|
               field_defn.type = resolve_late_binding(field_defn.type)
             end
           end
@@ -113,19 +113,7 @@ module GraphQL
 
       def get_fields_from_class(class_sym:)
         object_type_defn = load_constant(class_sym)
-
-        if object_type_defn.is_a?(Module)
-          object_type_defn.fields
-        else
-          extracted_field_defns = {}
-          object_class = object_type_defn.metadata[:type_class]
-          object_type_defn.all_fields.each do |field_defn|
-            inner_resolve = field_defn.resolve_proc
-            resolve_with_instantiate = PerFieldProxyResolve.new(object_class: object_class, inner_resolve: inner_resolve)
-            extracted_field_defns[field_defn.name] = field_defn.redefine(resolve: resolve_with_instantiate)
-          end
-          extracted_field_defns
-        end
+        object_type_defn.fields
       end
 
       # This is probably not 100% robust -- but it has to be good enough to avoid modifying the built-in introspection types
--- a/lib/graphql/schema/member/base_dsl_methods.rb
+++ b/lib/graphql/schema/member/base_dsl_methods.rb
@@ -108,7 +108,7 @@ module GraphQL
         end
 
         def visible?(context)
-          true
+          @mutation ? @mutation.visible?(context) : true
         end
 
         def authorized?(object, context)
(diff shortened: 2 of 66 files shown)


spec/graphql/schema/dynamic_members_spec.rb
```diff
@@ -893,6 +897,8 @@ def thing
             raise ArgumentError, "Unhandled type kind: #{type_kind.inspect}"
           end
         end
+
+        field :other, OtherObject
```
rmosolgo, line 901: TODO: What is this for?

spec/graphql/schema/dynamic_members_spec.rb
```diff
@@ -333,6 +333,7 @@ def yell(scream:)
       # just to attach these to the schema:
       field :example_locale, Locale
       field :example_region, Region
+      field :example_country, Country
```
rmosolgo, line 336: TODO why is this necessary?

spec/graphql/authorization_spec.rb
```diff
@@ -72,17 +72,19 @@ class BaseEnum < GraphQL::Schema::Enum
...
-        HiddenObject
+        def resolve_type(obj, ctx)
+          HiddenObject
+        end
       end
     end
 
     module HiddenDefaultInterface
-      include BaseInterface
+      include HiddenInterface
```
rmosolgo, line 87: TODO why this change?

lib/graphql/schema/member/base_dsl_methods.rb
```diff
@@ -108,7 +108,7 @@ def default_graphql_name
         end
 
         def visible?(context)
-          true
+          @mutation ? @mutation.visible?(context) : true
```
rmosolgo, line 111: TODO Why is this necessary?

lib/graphql/schema/introspection_system.rb
```diff
@@ -113,19 +113,7 @@ def load_constant(class_name)
...
-        else
-          extracted_field_defns = {}
-          object_class = object_type_defn.metadata[:type_class]
-          object_type_defn.all_fields.each do |field_defn|
-            inner_resolve = field_defn.resolve_proc
-            resolve_with_instantiate = PerFieldProxyResolve.new(object_class: object_class, inner_resolve: inner_resolve)
-            extracted_field_defns[field_defn.name] = field_defn.redefine(resolve: resolve_with_instantiate)
-          end
-          extracted_field_defns
-        end
```
rmosolgo, line 128: This was support for `.define`-based types ... happily long gone 😅

PR #5003, opened 2024-06-27
Remove unused `InvalidTypeError`

ravangen:
As far as I can tell, this class and file is unused and a remanent of early code.


rmosolgo:
I believe it ... thanks for the fix!

PR #5008, opened 2024-07-03
Fix Relay subscription handler sending an empty string

rmosolgo:
Oops -- this was falling back to `""` when no query string was given, which breaks when using the OperationStore.


rafaeelaudibert (approved):
Thanks for the quick fix, @rmosolgo!

rmosolgo:
🚢  to NPM in graphql-ruby-client v1.14.0

PR #5013, opened 2024-07-11
Allow validating a set of array values

sj26:
I found I needed to validate a set of values coming into a field, like this:

```ruby
field :slices do
  argument :uuids, [String], validates: { all: { format: { with: UUID::REGEXP } } } }
end
```

I thought an `all` validator which composes the other validators might be a good idea. What do you think? This is a rough spike to demonstrate the idea and see if it might work.


rmosolgo:
Yeah, I dig it! And it looks like `all: ...` basically takes any of the existing validation configs and applies it to each item, is that correct? Very cool.

sj26:
Yeah, neat!

I shipped this in our api yesterday. I’d love to land it here upstream. How far do you reckon it is? I’ll get the test suite passing. And maybe some more specs around the edges?

sj26:
Okay, I think that's a reasonable balance of coverage, and some docs. The remaining CI failures look unrelated.

What do you reckon?

rmosolgo:
LGTM! Thanks again for this improvement.

Issue #5014, opened 2024-07-12
Improve visiblity / type filtering


rmosolgo:
In #4998, I started on a new schema visibility implementation. I have a few goals here: 

- Make it possible to load GraphQL types lazily -- only loading types needed by a given query. 
- Make it possible to use cached schema subsets (don't have to recalculate the visible schema on every query -- when possible)
- Improve logging and debugging of visibility (especially when types are hidden) 
- Make a very smooth migration process: 
  - Use the `Query#types` interface to support logging when the two systems would diverge 
  - Clearly document the expectations of the new system (Warden was basically never documented, and arguably tried to do too much, but couldn't do everything) 

I'm opening this issue for my own purposes but also in case anyone wants to share any thoughts or questions on the work.

TODO: 

- [x] Build a `Subset` cache-by-name system
- [x] Make field return types lazy-loadable #5054 
- [x] Make root types lazy-loadable from the schema #5055 
- [ ] Update the install generator to add Rubocop rules to rails (when the file is present) 
- [x] Make it easier to start using `Subset` than `self.use_schema_subset = true`
- [x] Make a way to preload everything in Production
- [x] Warn when Warden is used without being explicitly included

sealabcore:
I'm having an issue with the latest version of the gem related to this work in the tests for my mutations.

In my tests I call this

```ruby
context = { current_user: user }
Mutations::AcceptDiscussionEmailInvite.new(object: nil, field: nil, context: context).resolve(**args)
```

And with the latest gem version I am getting this error:
```ruby
NoMethodError: undefined method `types' for { :current_user=> <User> }
test/graphql/mutations/accept_discussion_email_invite_test.rb:8:in `new'
```

Any thoughts on what I should be passing into the mutation as the context for my tests? Looks like a hash doesn't work anymore.

sealabcore:
Figured it out,

```ruby
context = { current_user: user }
context = GraphQL::Query::Context.new(query: GraphQL::Query.new(MapleSchema), values: context)
Mutations::AcceptDiscussionEmailInvite.new(object: nil, field: nil, context: context).resolve(**args)
```

rmosolgo:
Hey, sorry for the disruption there. Initializing Mutations outside of query execution isn't a formally supported API, but I'm glad you found something that works. In a pinch, another way to get a context is: 

```ruby
query = MySchema.query_class.new(self, "{ __typename }", context: { ... } )
context = query.context
```

[...]

PR #5018, opened 2024-07-15
Schema::Subset: Don't require schema-wide type caches at runtime

rmosolgo:
I'm hoping to make the new `Schema::Subset` system run without loading all types upfront. I'm sure this will have some incompatibilities with Warden but I'm not sure how significant they'll be yet.

Part of #5014

Part of #4919 

TODO: 

- [x] Make a compatibility layer that compares Warden vs. Subset and logs discrepancies 
- [x] Optimize `Interface#orphan_types`
- Somehow warn (or auto-add?) if an implementers reference isn't present? 
  - Unfortunately I don't see a way to do that right now... 
- [x] Review necessary changes and document them

I also profiled `Schema::Subset` using `bench:profile_large_result` and got it in the right ballpark: 

| | Warden | Subset | % change |
|---|----|----|----|
| Objects | 87453 | 87575 | +0.12% |
| Memory | 11308552 | 11321808 | +0.14% |
| Ips |  5.767 | 5.627 | n/a |


lib/graphql/schema/directive/flagged.rb
```diff
@@ -7,7 +7,7 @@ class Directive < GraphQL::Schema::Member
       # In this case, the server hides types and fields _entirely_, unless the current context has certain `:flags` present.
       class Flagged < GraphQL::Schema::Directive
         def initialize(target, **options)
-          if target.is_a?(Module) && !target.ancestors.include?(VisibleByFlag)
```
rmosolgo, line 10: This spec broke when an interface was flagged because of how the implementations of `visible?` stacked in the inheritance chain. Making object types get this included on their own fixed it.

spec/graphql/authorization_spec.rb
```diff
@@ -72,18 +72,25 @@ class BaseEnum < GraphQL::Schema::Enum
...
 
     module HiddenDefaultInterface
-      include BaseInterface
-      # visible? will call the super method
+      if GraphQL::Schema.use_schema_subset?
+        include HiddenInterface
+      else
+        # Warden will detect no possible types
+        include BaseInterface
+      end
```
rmosolgo, line 92: This is one of the biggest differences between `Warden` and `Subset`: 

- `Warden` would check if an interface (or union) had any visible possible_types; if it didn't, it would hide that type. (This caused some confusion...) 
- `Subset` won't do that: it will allow the query to use the interface (or union) but fail later if `possible_types` is called for that type (because there won't be any possible types to resolve the current object to).

spec/graphql/schema/argument_spec.rb
```diff
@@ -82,6 +82,13 @@ def context_arg_test(input:)
       end
 
       field :other_unauthorized_instruments, resolver: LoadUnauthorizedInstruments
+
+      # Warden's answer for `possible_types(UnauthorizedInstrumentType)` was an empty array
+      # because the type wasn't used in the schema except as `loads:...`
+      # This makes Subset and Warden behave the same.
+      field :example_unauthorized_instrument, UnauthorizedInstrumentType
+      # Same thing:
+      field :example_instrument, Jazz::InstrumentType
```
rmosolgo, line 91: Here's another kind of divergence I had to address. 

- When values are loaded with `loads:`, GraphQL-Ruby confirms that they match the expected type by doing `possible_types(loadable_type).include?(resolve_type(loaded_object))`. So, `possible_types(...)` may be called on plain object types.
- `Warden`  hid types from `possible_types` if they weren't attached as field return types. So in cases like this, it returned `[]`
- But `Subset` avoids that traversal if possible, so it returns `[obj_type]` without traversing. 

This divergence raised an error with TypesMigration, so I added this field to address it. 

If this becomes a problem, and people can't attach fields to the schema like this, we might be able to somehow track `@loads_only_types` in Subset to handle cases like this.

spec/graphql/schema/directive/flagged_spec.rb
```diff
@@ -5,6 +5,10 @@
   class FlaggedSchema < GraphQL::Schema
     module Animal
       include GraphQL::Schema::Interface
+      if GraphQL::Schema.use_schema_subset?
+        # It won't check possible types, so it needs this directly
+        directive GraphQL::Schema::Directive::Flagged, by: ["northPole", "southPole"]
+      end
```
rmosolgo, line 11: With `Warden`, this would be hidden when all possible types were hidden. But Subset doesn't check that, so I added a visibility rule that would do the same thing.

spec/graphql/schema/dynamic_members_spec.rb
```diff
@@ -199,6 +199,11 @@ class Country < BaseObject
 
     class Locale < BaseUnion
       possible_types Country, future_schema: true
+
+      if GraphQL::Schema.use_schema_subset?
+        # Subset won't check possible_types, this must be flagged
+        self.future_schema = true
+      end
```
rmosolgo, line 206: Same issue with possible_types

spec/graphql/schema/dynamic_members_spec.rb
```diff
@@ -841,6 +854,10 @@ class OtherObject < GraphQL::Schema::Object
         field :f, Int, null: false
       end
 
+      if GraphQL::Schema.use_schema_subset?
+        ThingInterface.orphan_types(OtherObject)
+      end
```
rmosolgo, line 859: This is necessary because, with Warden, it could be found via (unfiltered) `Schema.types`. But Subset doesn't do that, so it needed another way to reference it. 

(Its use in `ThingUnion.possible_types(...)` would make it so that `Schema.add_type_and_traverse` would find it. But Subset doesn't use the result of `add_type_and_traverse`.)

spec/graphql/schema/input_object_spec.rb
```diff
@@ -237,8 +238,6 @@ def self.object_from_id(id, ctx)
         def self.resolve_type(type, obj, ctx)
           type
         end
-
-        orphan_types [Jazz::InstrumentType]
```
rmosolgo, line 241: TODO: could this change be addressed by adding a check for `orphan_type?` inside Subset?

spec/graphql/schema/printer_spec.rb
```diff
@@ -691,9 +691,9 @@ def self.visible?(member, ctx)
             when "SCALAR"
               true
             when "OBJECT", "UNION", "INTERFACE"
-              ctx[:names].include?(member.graphql_name)
+              ctx[:names].include?(member.graphql_name) || member.introspection?
             else
-              false
+              member.introspection?
```
rmosolgo, line 696: TODO: why this change? Same question about rake_test_spec.rb

spec/graphql/schema/warden_spec.rb
```diff
@@ -524,6 +534,15 @@ class C < A; end
...
           possible_types A, B, C
+
+          if GraphQL::Schema.use_schema_subset?
+            def self.visible?(ctx)
+              (
+                possible_types.any? { |pt| ctx.schema.visible?(pt, ctx) } ||
+                ctx.schema.extra_types.include?(self)
+              ) && super
+            end
+          end
```
rmosolgo, line 545: This is the possible_types thing, where Subset doesn't know if the possible types are hidden or not.

spec/graphql/rake_task_spec.rb
```diff
@@ -31,7 +31,7 @@ def self.visible?(member, ctx)
 # Configured task
 GraphQL::RakeTask.new(idl_outfile: "tmp/configured_schema.graphql") do |t|
   t.namespace = "graphql_custom"
-  t.load_context = ->(task) { {filtered: true} }
+  t.load_context = ->(task) { {filtered: true, skip_types_migration_error: true } }
```
rmosolgo, line 34: TODO: this was necessary because `all_types` included introspection in one case, but not the other. It didn't end up mattering, since it was in the context of printing the schema. (Maybe that's the link -- the other weird issue was in `printer_spec.rb`.) 

But I need to figure out why those things were on the list in one case but not the other.

rmosolgo: OK, I got it: 

The previous implementation called `Warden#reachable_types` in the printer, see the diff here: https://github.com/rmosolgo/graphql-ruby/pull/4998/files#diff-e9dd0d295a58b66fabd1cf717040de5a78ade0feac6aeabae5247b5785c029abL268

But `reachable_types` never actually called `.visible?` on root types, it adds types immediately: 

https://github.com/rmosolgo/graphql-ruby/blob/83dabed599639c662b8d209d4218c1ad7cc1f1ab/lib/graphql/schema/warden.rb#L509-L510

So, introspection types were not checked through `.visible?` in that case. (Neither were root types or orphan_types -- I think they should have been.) 

But `Subset` passes them all through `.visible?` before adding them to this list. So I think Warden has the wrong behavior here. I'm going to update the `.visible?` implementation to permit introspection stuff.

spec/graphql/schema/printer_spec.rb
rmosolgo: `Warden` didn't send root introspection types through `.visible?` checks when using the `#reachable_types` method, hence this difference. See also https://github.com/rmosolgo/graphql-ruby/pull/5018#discussion_r1686746510

spec/graphql/schema/argument_spec.rb
rmosolgo: Improved `loadable?` tracking in ae4874452

spec/graphql/schema/input_object_spec.rb
rmosolgo: Improved `loadable?` in ae4874452

Issue #5020, opened 2024-07-15
Error paths and locations are duplicated with duplicate queries that error lazily


juanrgon:
**Describe the bug**

Error paths and locations are duplicated with duplicate queries that error lazily.

I provided a simple script to reproduce this issue. It uses a simple single field schema with a dataloader that always resolves to an error:

```rb
require 'graphql'
require 'json'

class ItemLoader < GraphQL::Dataloader::Source
  def fetch(keys)
    keys.map { |key| GraphQL::ExecutionError.new("Error for #{key}") }
  end
end

class QueryType < GraphQL::Schema::Object
  field :item, String do
    argument :key, String
  end

  def item(key:)
    dataloader.with(ItemLoader).load(key)
  end
end

class Schema < GraphQL::Schema
  query QueryType
  use GraphQL::Dataloader
end

query = <<-GRAPHQL
  query {
    query0: item(key: "a")
    query1: item(key: "a")
  }
GRAPHQL

result = Schema.execute(query)
puts JSON.pretty_generate(result.to_h)

```

When running the above, I would expect to see the following:

```json
{
  "errors": [
    {
      "message": "Error for a",
      "locations": [
        {
          "line": 2,
          "column": 5
        }
      ],
      "path": [
        "query0"
      ]
    },
    {
      "message": "Error for a",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "query1"
      ]
    }
  ],
  "data": {
    "query0": null,
    "query1": null
  }
}
```

but instead, I'm seeing this:

```json
{
  "errors": [
    {
      "message": "Error for a",
      "locations": [
        {
          "line": 2,
          "column": 5
        }
      ],
      "path": [
        "query0"
      ]
    },
    {
      "message": "Error for a",
      "locations": [
        {
          "line": 2,
          "column": 5
        }
      ],
      "path": [
        "query0"
      ]
    }
  ],
  "data": {
    "query0": null,
    "query1": null
  }
}
```

**Versions**

`graphql` version: 2.3.9

**GraphQL schema**

```ruby
class ItemLoader < GraphQL::Dataloader::Source
  def fetch(keys)
    keys.map { |key| GraphQL::ExecutionError.new("Error for #{key}") }
  end
end

class QueryType < GraphQL::Schema::Object
  field :item, String do
    argument :key, String
  end

  def item(key:)
    dataloader.with(ItemLoader).load(key)
  end
end

[...]

PR #5021, opened 2024-07-15
Use fresh path and ast_node values for errors of lazy evaluated queries

juanrgon:
Closes https://github.com/rmosolgo/graphql-ruby/issues/5020

I'm attempting to fix an issue with error paths and locations being duplicated with DataLoaders and lazy evaluation

### DIFF
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -186,8 +186,11 @@ This key should have been loaded already. This is a bug in GraphQL::Dataloader, 
 ERR
         end
         result = @results[key]
-
-        raise result if result.class <= StandardError
+        if result.is_a?(StandardError)
+          # Dup it because the rescuer may modify it.
+          # (This happens for GraphQL::ExecutionErrors, at least)
+          raise result.dup
+        end
 
         result
       end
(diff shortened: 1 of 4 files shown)


rmosolgo:
Hey, thanks for sharing this fix -- I'd love to include it! Would you mind including the script from #5020 in the test suite? You could add it to `execution_error_spec.rb`, for example.

rmosolgo:
Thanks for your work on this! In reviewing the change, I realized that the source of the problem was `Dataloader::Source`'s cache of `fetch(...)`'d values. Whenever that method returns a value for a key, it's used again for later requests for the same key. So in this case, it was returning the exact same `GraphQL::ExecutionError` for the requested key (`"a"`). 

I'm hesitant to change `runtime.rb` when I can help it -- that code affects a _lot_ of behavior! So I pushed a change that handles it inside `Dataloader::Source` instead, by `dup`'ing any error from `fetch`.

juanrgon:
@rmosolgo is it possible to get this fix back-ported to 2.1 and 2.2 in addition to 2.3? I can do the heavy-lifting for this, if possible, but will just need a little guidance

rmosolgo:
Yes, no problem. Basically my approach is to: 

- pull the `2.1.x` branch locally
- create new branch, eg, `2.1-fix-dataloader-error-path`
- cherry-pick each of these commits to that branch (or, somehow port the changes on to that branc)
- push the branch and open a PR onto the `2.1.x` branch 

Let me know if you have any trouble -- I'd be happy to help with the backports. Once the PR is ready, I can merge and release to rubygems :+1:

juanrgon:
Thanks for the quick response @rmosolgo! I opened two PRs with patches:

- [2.1 patch](https://github.com/rmosolgo/graphql-ruby/pull/5025)
- [2.2 patch](https://github.com/rmosolgo/graphql-ruby/pull/5026)

PR #5022, opened 2024-07-16
Use base class for errors in anonymous enum classes

### LINKED ISSUE #5016: Some exception class names are unstable when using `GraphQL::Schema.from_definition`
**Describe the bug**

When using `GraphQL::Schema.from_definition`, the resulting schema object sometimes produces exceptions like `#<Class:0x00000001097cbd70>::UnresolvedValueError`, where the exception class is defined underneath an anonymous class. This results in exceptions that have unstable names (if two Ruby processes running the same code hit the same error, the resulting exception class is named differently) and that in turn impacts the ability of an exception tracking system like [sentry](https://sentry.io/) to accurately group exceptions.

**Versions**

`graphql` version: 2.3.8

**GraphQL schema**

```graphql
enum Size {
  SMALL
  MEDIUM
  LARGE
}

type Query {
  size: Size
}
```

**GraphQL query**

```graphql
query {
  size
}
```

```
#<Class:0x00000001097cbd70>::UnresolvedValueError: `Query.size` returned `"MEDUM"` at `size`, but this isn't a valid value for `Size`. Update the field or resolver to return one of `Size`'s values instead.
```

**Steps to reproduce**

Put this code into an executable file:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", "2.3.8"
end

require "graphql"

SCHEMA_STRING = <<~EOS
  enum Size {
    SMALL
    MEDIUM
    LARGE
  }

  type Query {
    size: Size
  }
EOS

class Application
  def initialize
    @schema = GraphQL::Schema.from_definition(SCHEMA_STRING, default_resolve: self)
  end

  def call(parent_type, field, object, args, context)
    "MEDUM" # mispelled
  end

  def execute_query(query_string)
    @schema.execute(query_string)
  end
end

app = Application.new

begin
  app.execute_query("query { size }")
rescue => e
  puts "#{e.class.name}: #{e.message}"
end
```

Run this script--each time it'll report a different exception class name:

```
$ for i in $(seq 1 3); do ruby tmp/graphql_issue.rb; done
#<Class:0x000000010c57bfe0>::UnresolvedValueError: `Query.size` returned `"MEDUM"` at `size`, but this isn't a valid value for `Size`. Update the field or resolver to return one of `Size`'s values instead.
#<Class:0x0000000107d9c080>::UnresolvedValueError: `Query.size` returned `"MEDUM"` at `size`, but this isn't a valid value for `Size`. Update the field or resolver to return one of `Size`'s values instead.
#<Class:0x00000001055fbeb8>::UnresolvedValueError: `Query.size` returned `"MEDUM"` at `size`, but this isn't a valid value for `Size`. Update the field or resolver to return one of `Size`'s values instead.
```

**Expected behavior**

I expect exception classes to be "stable": the name should be the same each time the script runs.

**Actual behavior**

The exception class name is different each time.

**Additional context**

I think this is caused by some dynamic constant assignment here:

https://github.com/rmosolgo/graphql-ruby/blob/v2.3.8/lib/graphql/schema/enum.rb#L169

I also noticed that same sort of thing shows up here:

https://github.com/rmosolgo/graphql-ruby/blob/v2.3.8/lib/graphql/schema/member/has_unresolved_type_error.rb#L10

I haven't done a full audit, but I'd love to see this fixed for all cases like this.

rmosolgo:
Fixes #5016

### DIFF
--- a/lib/graphql/schema/enum.rb
+++ b/lib/graphql/schema/enum.rb
@@ -166,7 +166,11 @@ module GraphQL
         end
 
         def inherited(child_class)
-          child_class.const_set(:UnresolvedValueError, Class.new(Schema::Enum::UnresolvedValueError))
+          if child_class.name
+            # Don't assign a custom error class to anonymous classes
+            # because they would end up with names like `#<Class0x1234>::UnresolvedValueError` which messes up bug trackers
+            child_class.const_set(:UnresolvedValueError, Class.new(Schema::Enum::UnresolvedValueError))
+          end
           super
         end
 
--- a/spec/graphql/schema/enum_spec.rb
+++ b/spec/graphql/schema/enum_spec.rb
@@ -198,6 +198,21 @@ describe GraphQL::Schema::Enum do
       }
       expected_isolated_message = "`:nonsense` was returned for `DairyAnimal`, but this isn't a valid value for `DairyAnimal`. Update the field or resolver to return one of `DairyAnimal`'s values instead."
       assert_equal expected_isolated_message, err2.message
+      assert_equal "Dummy::DairyAnimal::UnresolvedValueError", err2.class.name
+
+      anon_enum = Class.new(GraphQL::Schema::Enum) do
+        graphql_name "AnonEnum"
+        value :one
+        value :two
+      end
+
+      err3 = assert_raises(GraphQL::Schema::Enum::UnresolvedValueError) {
+        anon_enum.coerce_isolated_result(:nonsense)
+      }
+
+      expected_anonymous_message = "`:nonsense` was returned for `AnonEnum`, but this isn't a valid value for `AnonEnum`. Update the field or resolver to return one of `AnonEnum`'s values instead."
+      assert_equal expected_anonymous_message, err3.message
+      assert_equal "GraphQL::Schema::Enum::UnresolvedValueError", err3.class.name
     end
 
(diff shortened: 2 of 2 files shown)


lib/graphql/schema/enum.rb
```diff
@@ -166,7 +166,11 @@ def coerce_input(value_name, ctx)
         end
 
         def inherited(child_class)
-          child_class.const_set(:UnresolvedValueError, Class.new(Schema::Enum::UnresolvedValueError))
+          if child_class.name
+            # Don't assign a custom error class to anonymous classes
+            # because they would end up with names like `#<Class0x1234>::UnresolvedValueError` which messes up bug trackers
+            child_class.const_set(:UnresolvedValueError, Class.new(Schema::Enum::UnresolvedValueError))
+          end
```
myronmarston, line 173: 👍 

I suspect that the same issue exists here:

https://github.com/rmosolgo/graphql-ruby/blob/v2.3.8/lib/graphql/schema/member/has_unresolved_type_error.rb#L10

Does it make sense to fix it as well?

rmosolgo: Ah, good call! https://github.com/rmosolgo/graphql-ruby/pull/5023

PR #5045, opened 2024-07-30
Fix default traces added to parent classes after default traces added to child class

### LINKED ISSUE #5044: Multiple tracers in one schema are ignored
**Describe the bug**

This bug almost appears to be a duplicate of https://github.com/rmosolgo/graphql-ruby/issues/4542 but this time around it's with the ActiveSupport and OpenTelemetry tracers.

```rb
module Graph
    ...
    
    trace_with GraphQL::Tracing::ActiveSupportNotificationsTrace  end
end
```

This schema, with the default [`OpenTelemetry` setup](https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/graphql#usage), sees the [OpenTelemetry tracer](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_trace.rb) completely ignored as spans are not included in traces yet metrics are emitted.

If I mix and match the old tracer API, I'm able to get both working:

```rb
module Graph
    ...

    tracer GraphQL::Tracing::ActiveSupportNotificationsTracing
  end
end
```

**Versions**

`graphql`: 2.3.10
`opentelemetry-instrumentation-graphql`: 0.28.2
`rails`: 7.1.3.4

**Steps to reproduce**

The test case demonstrates the opposite to the above due to how our app is built, but I think it's still sufficient to show the bug.

```rb
# frozen_string_literal: true

require "bundler/inline"

gemfile do
  gem "graphql", "2.3.10"
  gem 'opentelemetry-api'
  gem 'opentelemetry-common'
  gem 'opentelemetry-instrumentation-graphql'
  gem 'opentelemetry-sdk'
  gem 'rails'
end

module MultipleTracer
  class Thing < GraphQL::Schema::Object
    field :str, String

    def str; "blah"; end
  end

  class Named < GraphQL::Schema::Union
    possible_types Thing
    def self.resolve_type(obj, ctx)
      Thing
    end
  end

    field :int, Integer, null: false

    def int
      1
    end

    field :thing, Thing
    def thing; :thing; end

    field :named, Named, resolver_method: :thing
  end

    query(Query)

    trace_with(GraphQL::Tracing::ActiveSupportNotificationsTrace)
  end
end

# Comment from here.... 
ENV['OTEL_TRACES_EXPORTER'] = 'console'
OpenTelemetry::SDK.configure do |c|
  c.use_all('OpenTelemetry::Instrumentation::GraphQL' => { schemas: [MultipleTracer::TestSchema] })
end
# ...to here to demonstrate bug

ActiveSupport::Notifications.monotonic_subscribe("execute_query.graphql") do |name, start, finish, _message_id, data|
  puts "#{name} handled"
  puts "#{data.inspect}"
end

query = "{ int thing { str } named { ... on Thing { str } } }"
MultipleTracer::TestSchema.execute(query)
```

The console will be messy, but if you search for `execute_query.graphql handled` it will be absent. If these lines are commented out, and the command is re-run, you'll see the ActiveSupport handled messages.

```rb
ENV['OTEL_TRACES_EXPORTER'] = 'console'
OpenTelemetry::SDK.configure do |c|
  c.use_all('OpenTelemetry::Instrumentation::GraphQL' => { schemas: [MultipleTracer::TestSchema] })
end
```

```sh
$ ruby inline.rb
execute_query.graphql handled
{:query=>query ...}
```

**Expected behavior**

Both tracers should be able to emit information about the GraphQL operations. So with the above example there should be a wall of trace info, as well as the two littles lines added by the ActiveSupport subscriber.

**Actual behavior**

The span information is emitted, but nothing from the ActiveSupport subscriber.

rmosolgo:
Fixes #5044

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -205,24 +205,19 @@ module GraphQL
         @own_trace_modes ||= {}
       end
 
-      module DefaultTraceClass
-      end
-
-      private_constant :DefaultTraceClass
-
       def build_trace_mode(mode)
         case mode
         when :default
           # Use the superclass's default mode if it has one, or else start an inheritance chain at the built-in base class.
-          base_class = (superclass.respond_to?(:trace_class_for) && superclass.trace_class_for(mode)) || GraphQL::Tracing::Trace
-          Class.new(base_class) do
+          base_class = (superclass.respond_to?(:trace_class_for) && superclass.trace_class_for(mode, build: true)) || GraphQL::Tracing::Trace
+          const_set(:DefaultTrace, Class.new(base_class) do
             include DefaultTraceClass
-          end
+          end)
         when :default_backtrace
           schema_base_class = trace_class_for(:default, build: true)
-          Class.new(schema_base_class) do
+          const_set(:DefaultTraceBacktrace, Class.new(schema_base_class) do
             include(GraphQL::Backtrace::Trace)
-          end
+          end)
         else
           # First, see if the superclass has a custom-defined class for this.
           # Then, if it doesn't, use this class's default trace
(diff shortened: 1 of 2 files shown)


rmosolgo:
This case doesn't technically work, but it's not blocking #5044 so I'm not going to address it now.

PR #5048, opened 2024-07-31
fix: avoid iterating on null array if all validator permits null value

brandon-at-wrk:
When you want to validate an array only has values within a pre-defined set (`all: inclusion: []`), but also want to permit it to be `null`, right now it will crash on `undefined method `each' for nil:NilClass`

Example field and query that will crash before the patch, and won't after.
```ruby
field :pets, [Types::PetsType] do
  argument :filters, [String],
           description: 'Filters to apply -- nil for all',
           validates: {
             all: {
               allow_null: true,
               inclusion: { in: %w[cat dog turtle bird] },
             },
           }
end
```

```graphql
query MyFavoritePets {
  pets(filters: null) {
    name
  }
}
```

### DIFF
--- a/lib/graphql/schema/validator/all_validator.rb
+++ b/lib/graphql/schema/validator/all_validator.rb
@@ -28,6 +28,8 @@ module GraphQL
         end
 
         def validate(object, context, value)
+          return EMPTY_ARRAY if permitted_empty_value?(value)
+
           all_errors = EMPTY_ARRAY
 
           value.each do |subvalue|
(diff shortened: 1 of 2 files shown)


rmosolgo:
Hey, thanks for this fix -- LGTM! Do you mind adding a test case for this in https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/schema/validator/all_validator_spec.rb?

brandon-at-wrk:
Added it, but I only checked that one file  -- ie ran `bundle exec rake test TEST=spec/graphql/schema/validator/all_validator_spec.rb`

rmosolgo:
Sweet, thanks!

PR #5054, opened 2024-08-06
Defer calling field do ... end blocks until needed

rmosolgo:
This way, in theory, loading `my_object.rb` doesn't _necessary_ load the field return types of these fields. (Rails won't try to load files until the constants are actually referenced, and in this case, it can be deferred until execution.)

cc @gmcgibbon  -- would this approach work in your type-checking setup? Like this:

```ruby 
field :thing do 
  type Types::Thing 
end 

# or 
field :thing { type(Types::Thing) } 

# or, we could add sugar, 
# where the block's return value is used as the field's return type 
# if no return type was already configured: 
# ~~field :thing { Types::Thing } ~~
# Nevermind, that's not valid ruby :'( You could do it with params: 
field(:thing) { Types::Thing } # <- not implemented yet
```

I think that would give you plain Ruby constants in the source. 

The downside is the need to call `ensure_loaded` everywhere. It's too bad and reminds me of how GraphQL-Ruby used to be full of `ensure_defined`, yuck 😩 . But, I think it's the right thing, and thanks to caching by the schema / Warden / Subset, calls to that method were surprisingly manageable. 

`ensure_loaded` could even be _helpful_. For example, lots of `field.rb` checks to see if `@resolver` has a return type, arguments, etc, and uses those if they're given. But those checks could be done eagerly in `ensure_loaded`, and maybe even a `freeze` call to avoid surprises if the field is modified after `ensure_loaded` has run.

Part of #5014

TODO: 

- [x] Write up an integration test or something to make sure that this actually _delivers_ the expected Rails behavior
- [x] Write a Rubocop rule to migrate these automatically (excluding built-in scalars)
- [x] Update `Field#type(t)` to updated `connection?` as needed
- ~~Document support for this~~ I'm going to wait and document the whole thing when it's ready

### DIFF
--- a/lib/graphql/rubocop.rb
+++ b/lib/graphql/rubocop.rb
@@ -2,3 +2,4 @@
 
 require "graphql/rubocop/graphql/default_null_true"
 require "graphql/rubocop/graphql/default_required_true"
+require "graphql/rubocop/graphql/field_type_in_block"
--- a/lib/graphql/rubocop/graphql/field_type_in_block.rb
+++ b/lib/graphql/rubocop/graphql/field_type_in_block.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+require_relative "./base_cop"
+
+module GraphQL
+  module Rubocop
+    module GraphQL
+      # Identify (and auto-correct) any field whose type configuration isn't given
+      # in the configuration block.
+      #
+      # @example
+      #   # bad, immediately causes Rails to load `app/graphql/types/thing.rb`
+      #   field :thing, Types::Thing
+      #
+      #   # good, defers loading until the file is needed
+      #   field :thing do
+      #     type(Types::Thing)
+      #   end
+      #
+      class FieldTypeInBlock < BaseCop
+        MSG = "type configuration can be moved to a block to defer loading the type's file"
+
+        BUILT_IN_SCALAR_NAMES = ["Float", "Int", "Integer", "String", "ID", "Boolean"]
+        def_node_matcher :field_config_with_inline_type, <<-Pattern
+        (
+          send {nil? _} :field sym ${const array} ...
+        )
+        Pattern
+
+        def_node_matcher :field_config_with_inline_type_and_block, <<-Pattern
+        (
+          block
+            (send {nil? _} :field sym ${const array}) ...
+            (args)
+            _
+
+        )
+        Pattern
+
+        def on_block(node)
+          field_config_with_inline_type_and_block(node) do |type_const|
+            ignore_node(type_const)
+            type_const_str = get_type_argument_str(node, type_const)
+            if ignore_inline_type_str?(type_const_str)
+              # Do nothing ...
+            else
+              add_offense(type_const) do |corrector|
+                cleaned_node_source = delete_type_argument(node, type_const)
+                field_indent = determine_field_indent(node)
+                cleaned_node_source.sub!(/(\{|do)/, "\\1\n#{field_indent}  type #{type_const_str}")
+                corrector.replace(node, cleaned_node_source)
+              end
+            end
+          end
+        end
+
+        def on_send(node)
+          field_config_with_inline_type(node) do |type_const|
+            return if ignored_node?(type_const)
+            type_const_str = get_type_argument_str(node, type_const)
+            if ignore_inline_type_str?(type_const_str)
+              # Do nothing -- not loading from another file
+            else
+              add_offense(type_const) do |corrector|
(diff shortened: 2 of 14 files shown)


rmosolgo:
I was able to confirm this has the desired effect in Rails by making a new app and adding fields like this to `Query`: 

```ruby
    field :things, fallback_value: [{name: "Abacus"}] do
      type [Types::Thing]
    end

    field :other_things, fallback_value: [{name: "Abacus"}] do
      type [Types::OtherThing]
    end
```

I also added `puts ">>>> loading #{__FILE__}"` to several GraphQL-related files.

Then, from GraphiQL, I made two different queries to the backend: 


```graphql 
# first 
query {
  things { __typename }
}

# second 
query {
  otherThings { __typename }
}
```

In the Rails server log, I saw the desired load messages. The first query loaded `thing.rb`, but not `other_thing.rb`:


```
Processing by GraphqlController#execute as */*
  Parameters: {"query"=>"query {\n  things { __typename }\n}", "graphql"=>{"query"=>"query {\n  things { __typename }\n}"}}
>>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/graphql_lazy_load_test_schema.rb
>>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/types/query_type.rb
>>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/types/thing.rb
Completed 200 OK in 16ms (Views: 0.3ms | ActiveRecord: 0.0ms | Allocations: 6871)
```

The second query loaded `other_thing.rb` but not `thing.rb`: 

```
Started POST "/graphql" for 127.0.0.1 at 2024-08-07 16:08:40 -0400
Processing by GraphqlController#execute as */*
  Parameters: {"query"=>"query {\n  otherThings { __typename }\n}", "graphql"=>{"query"=>"query {\n  otherThings { __typename }\n}"}}
>>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/graphql_lazy_load_test_schema.rb
>>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/types/query_type.rb
>>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/types/other_thing.rb
Completed 200 OK in 9ms (Views: 0.1ms | ActiveRecord: 0.0ms | Allocations: 3308)
```

🎉

swalkinshaw:
We have a few places where we override `field` methods with a pattern like this:

```ruby
def field(*args, **kwargs, some_custom_option: true, &block)
  super(*T.unsafe(args), **options).tap do |field_defn|
    field_defn.extension(SomeExtension)
  end
end
```

This PR no longer yields the field instance of course. Two questions:

* should `ensure_loaded` ever be used outside of the gem?
* is there an alternative for patterns like this and would work going forward? In some simple cases (like above), we can just add the extension directly but there's cases with much more conditional logic and options.

rmosolgo:
Hey,  sorry it wasn't working the same right out of the box. I'm sure there's something that will work for this... 

What goes wrong in the example above? (I see `.tap do ...` which I would _expect_ to still work fine.) 

One thing that comes to mind is passing a _new_ block to `super`, something like: 


```ruby 
field(*args, **kwargs, &block)
  super(*args, **kwargs) do |field_inst|
     block.call(field_inst)
     field_inst.extension(SomeExtension)
  end 
end 
```

IIUC that new block would be passed to `super` and be used when needed. 

I'd love to keep `ensure_loaded` a private concern to the gem. Could you share some other examples of things that don't work right after this change?

swalkinshaw:
😓  my example was a bad one. We have a few patterns and it's actually the other one that has the issue:

```ruby
def field(*args, **kwargs, some_custom_option: true, &block)
  field_defn = super(*T.unsafe(args), **options)
  field_defn.extension(SomeExtension)
  field_defn
end
```

`super` returns `nil` now. Moving that logic to within a block passed to `super` should work because the block will be called during `ensure_loaded`.

So I think it all makes sense and _should_ work to support the same use cases.

rmosolgo:
Ok, thanks for those details. I'm surprised `super` returns `nil` now ... I would _expect_ it to still return the new instance: 

https://github.com/rmosolgo/graphql-ruby/blob/5dcdb597c4dd9dc943b8a86f11c271a32d2a97a4/lib/graphql/schema/member/has_fields.rb#L11-L15

I'd be happy to debug further if you don't mind sharing the full error message and backtrace. Maybe it would have another clue as to where that `nil` is coming from.

PR #5055, opened 2024-08-06
Accept blocks to reference schema root types

rmosolgo:
With this, the schema could defer loading a root type until necessary (thus also deferring loading all types referenced by it, resolvers, etc). 

In this implementation, calling `schema.query` returns the deferred definition which means that if, say, a plug-in references that root type, all the magic is lost. 

To be really effective, we'd also need some plugin behavior like `extend(:query) { ... }` where callbacks could be registered by the plugin, then called by GraphQL-Ruby after the type is actually loaded. I think that can be added later.

Part of #5014  

TODO: 

- [x] Add a test that proves that Rails doesn't load `types/query.rb` until a query is run when using `Schema::Subset`. 
- [x] Write a Rubocop rule for it

### DIFF
--- a/lib/graphql/rubocop.rb
+++ b/lib/graphql/rubocop.rb
@@ -3,3 +3,4 @@
 require "graphql/rubocop/graphql/default_null_true"
 require "graphql/rubocop/graphql/default_required_true"
 require "graphql/rubocop/graphql/field_type_in_block"
+require "graphql/rubocop/graphql/root_types_in_block"
--- a/lib/graphql/rubocop/graphql/root_types_in_block.rb
+++ b/lib/graphql/rubocop/graphql/root_types_in_block.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+require_relative "./base_cop"
+
+module GraphQL
+  module Rubocop
+    module GraphQL
+      # Identify (and auto-correct) any root types in your schema file.
+      #
+      # @example
+      #   # bad, immediately causes Rails to load `app/graphql/types/query.rb`
+      #   query Types::Query
+      #
+      #   # good, defers loading until the file is needed
+      #   query { Types::Query }
+      #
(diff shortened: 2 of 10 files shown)


rmosolgo:
I confirmed this has the desired effect by making a Rails app and configuring the schema with: 

```ruby
puts ">>>> loading #{__FILE__}"

class GraphqlLazyLoadTestSchema < GraphQL::Schema
  self.use_schema_subset = true

  mutation { Types::MutationType }
  query { Types::QueryType }
  # ...
end 
```

Then, I ran two queries, and watched the Rails server logs: 

```graphql
query {
  __typename
}
# Processing by GraphqlController#execute as */*
#   Parameters: {"query"=>"query {\n  __typename\n}", "graphql"=>{"query"=>"query {\n  __typename\n}"}}
# >>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/graphql_lazy_load_test_schema.rb
# >>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/types/query_type.rb
# Completed 200 OK in 15ms (Views: 0.2ms | ActiveRecord: 0.0ms | Allocations: 6307)
```

Then: 

```graphql
mutation {
  __typename
}
# Started POST "/graphql" for 127.0.0.1 at 2024-08-08 10:57:45 -0400
# Processing by GraphqlController#execute as */*
#   Parameters: {"query"=>"mutation {\n  __typename\n}", "graphql"=>{"query"=>"mutation {\n  __typename\n}"}}
# >>>> loading /Users/rmosolgo/code/graphql_lazy_load_test/app/graphql/types/mutation_type.rb
# Completed 200 OK in 3ms (Views: 0.1ms | ActiveRecord: 0.0ms | Allocations: 1025)
```

So, it works: running a `query` doesn't cause `types/mutation.rb` to be loaded 🎉

Issue #4252, opened 2022-11-21
Dataloaders are changing the response order


cesarjr:
**Describe the bug**

When a dataloader is used, the order of fields in the response is affected. 

All the fields using dataloaders are delivered after the others fields.

According to [GraphQL Specs](https://spec.graphql.org/June2018/#sec-Objects) the order of the fields in the answer must follow the order of fields in the query.

**Versions**

`ruby-graphql`: 2.0.15
`rails`: 7.0.4
`ruby`: 3.0.3

**GraphQL schema**

I've prepared a complete app reproducing the problem using rspec.

https://github.com/cesarjr/graphql-dataloader

Include relevant types and fields (in Ruby is best, in GraphQL IDL is ok). Any custom extensions, etc?

```ruby
class Types::Person < Types::BaseObject
  field :id, ID,
        null: false

  field :name, String,
        null: false,
        description: 'Name.'

  field :city_id, ID,
        null: false,
        description: 'City ID.'

  field :city, Types::City,
        null: false,
        description: 'City.'

  field :seller_id, ID,
        null: false,
        description: 'Seller ID.'

  field :seller, Types::Seller,
        null: false,
        description: 'Seller.'

  # comment the lines below if you want to see
  # the specs passing.
  def city
    dataloader
      .with(Sources::ActiveRecordObject, City)
      .load(object.city_id)
  end
end

class GraphqlDataloaderSchema < GraphQL::Schema
  mutation(Types::MutationType)
  query(Types::QueryType)

  # For batch-loading (see https://graphql-ruby.org/dataloader/overview.html)
  use GraphQL::Dataloader
  # …
end
```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
query { 
    getPeople {
        id
        name
        city {
            id
            name
            state
        }
        seller {
            id
            name
          }
    }
}
```
**Steps to reproduce**

Steps to reproduce the behavior

**Expected behavior**

```json
{
    "data": {
        "getPeople": [
            {
                "id": "1",
                "name": "Carl",
                "city": {
                    "id": "1",
                    "name": "Rio Branco",
                    "state": "Acre"
                },
                "seller": {
                    "id": "1",
                    "name": "John"
                }
            }
        ]
    }
}
```

It is expected that `city` to be returned first than `seller` as it was queried.

**Actual behavior**

```json
{
    "data": {
        "getPeople": [
            {
                "id": "1",
                "name": "Carl",
                "seller": {
[...]

PR #5060, opened 2024-08-12
Return ordered result hashes

### LINKED ISSUE #5011: Return result keys in the same order as the request
The spec says that response maps should have keys in the same order that the query requested fields (https://spec.graphql.org/draft/#sec-Serialized-Map-Ordering), but GraphQL-Ruby doesn't do that. Instead, it creates a Hash to store the result and writes keys as soon as they're available. Consequently, results from GraphQL-Ruby violate the spec: instead of coming back in the order requested by the client, they come back in a library-determined order.

The simplest implementation would be to make a new, properly ordered Hash, something like: 

```ruby
field_indexes = {}
ast_nodes.each_with_index { |node, idx| field_indexes[node.name] = idx }
ordered_result = result.sort_by { |k,v| field_indexes[k] } 
```

But I'll have to check the memory and runtime impact of adding that.

Fixes #4252

rmosolgo:
Fixes #5011 

The spec says that JSON key-value pairs should come back in the same order that they were requested in the query. I need an implementation that complies with the spec and adds as little overhead as possible (memory and runtime). Possibly, if it adds noticeable overhead, make it opt-in, since this is critical performance code

I need to somehow maintain the sequence of this result even when Dataloader or GraphQL-Batch write to it out of order (see #4252). 

I think the `selections_by_name` hash(es) have the correct order of keys here: https://github.com/rmosolgo/graphql-ruby/blob/d9bd9cf9ac6f01a9aa36c30a025bfa922eda4f46/lib/graphql/execution/interpreter/runtime.rb#L120-L192

I can't find any way to _re-order_ a Ruby Hash. So if I'm going to use a Hash, need to write the keys in the right order the _first_ time. (Technically, you can move a key to the back by doing `h[k] = h.delete(k)`, but I don't think that's a promising approach.) 

Another approach might be to use an Array to hold values and then build a Hash from there. For example, an array of `[k1, v1, k2, v2, ...]` could be used. One problem here is that it will require a second pass through the whole data structure. GraphQL-Ruby used to do this, but it was slow. I'd like to not make it do that again. 

Maybe I could mitigate the downside of the second iteration by implementing `Result#to_json` to return a properly-ordered string. This would work out-of-the-box with Rails... but I know a lot of JSON serializers are really optimized. Could GraphQL-Ruby deliver good performance in that case? (And if I take that approach, I should implement `.to_ordered_h` and test it against `.to_ordered_h.to_json` with an optimized JSON serializer.) 

`graphql-js` handles this by writing promises into the result hash. This is also how GraphQL-Ruby happens to handle `GraphQL::Execution::Lazy` instances (eg, GraphQL-Batch). Perhaps a similar approach can be made to work with Dataloader. But the challenge is cleaning up those placeholders. A `Promise` has its own `catch` block to do that, but Dataloader doesn't have that... (yet?)

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -207,7 +207,7 @@ module GraphQL
             finished_jobs = 0
             enqueued_jobs = gathered_selections.size
             gathered_selections.each do |result_name, field_ast_nodes_or_ast_node|
-
+              selections_result.set_placeholder(result_name)
               # Field resolution may pause the fiber,
               # so it wouldn't get to the `Resolve` call that happens below.
               # So instead trigger a run from this outer context.
--- a/lib/graphql/execution/interpreter/runtime/graphql_result.rb
+++ b/lib/graphql/execution/interpreter/runtime/graphql_result.rb
@@ -78,6 +78,19 @@ module GraphQL
             value
           end
 
+
+          # @api private
+          PLACEHOLDER = Object.new
+
+          def set_placeholder(key)
+            @graphql_result_data[key] = PLACEHOLDER
+            @graphql_metadata && @graphql_metadata[key] = PLACEHOLDER
+            if (t = @graphql_merged_into)
+              t.set_placeholder(key)
+            end
+            nil
+          end
+
           def delete(key)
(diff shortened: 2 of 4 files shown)


rmosolgo:
Trying again at #5315

PR #5062, opened 2024-08-12
Allow `validate_timeout` to be cleared/reset after being set

ravangen:
## Objective

Allow for `validate_timeout` to be cleared such that the timeout behaviour is removed by setting a `nil` value. Technically `0` does work already.

## Problem

`nil` is the default value so we cannot distinguish between a get and set.

## Proposal

Use existing `NOT_CONFIGURED` sentinel value as the default argument value. 

## Outcome

A schema can have a default `validate_timeout` but can be overridden to be removed.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -750,13 +750,11 @@ module GraphQL
 
       attr_writer :validate_timeout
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for this improvement!

PR #5065, opened 2024-08-13
Allow `validate_max_errors` to be cleared/reset after being set

ravangen:
## Objective

Allow for `validate_max_errors` to be cleared such that the bounding behaviour is removed by setting a `nil` value.

## Problem

`nil` is the default value so we cannot distinguish between a get and set.

## Proposal

Use existing `NOT_CONFIGURED` sentinel value as the default argument value. 

## Outcome

A schema can have a default `validate_max_errors` but can be overridden to be removed.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -787,13 +787,11 @@ module GraphQL
 
       attr_writer :validate_max_errors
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks!

PR #5066, opened 2024-08-13
Expose parser `tokens_count`

ravangen:
## Objective

Ability to apply [`max_query_string_tokens`](https://graphql-ruby.org/api-doc/2.3.14/GraphQL/Schema#max_query_string_tokens-class_method) setting without breaking any existing operations.

## Problem

We don't know the current distribution of the size of the tokens in existing operations.

## Proposal

Expose `tokens_count` on `Parser#tokens_count`. Trigger parsing if not already done.

## Outcome

To gather data, temporarily swap from the convenient `GraphQL.parse` to `GraphQL::Language::Parser.new` (or similar)

### DIFF
--- a/graphql-c_parser/lib/graphql/c_parser.rb
+++ b/graphql-c_parser/lib/graphql/c_parser.rb
@@ -123,6 +123,11 @@ module GraphQL
         @result
       end
 
+      def tokens_count
+        result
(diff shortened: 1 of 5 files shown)


rmosolgo:
Cool, thanks!

PR #5067, opened 2024-08-14
Support for comments v2

vaclavbohac:
Resolves https://github.com/rmosolgo/graphql-ruby/issues/4905

Reimplemented https://github.com/rmosolgo/graphql-ruby/pull/4970

The idea is that we expanded methods that construct the schema with new comment parameter (we should cover most of them like types, unions, enums). By passing string value to this new parameter, the resulting schema will contain comments such as this:

```graphql
# we added support for comments
type MyType {
  name: String!
}
```

The reasons for adding this are mentioned in the linked issue but in a nutshell, we want to leverage comments for tooling (eslint), using descriptions is not desirable because those would leak into documentation.

Comments are supported for the following types:
* unions
* scalars
* type definitions
* mutations
* arguments
* enums
* enum values


vaclavbohac:
Hello, @rmosolgo finally I think the implementation is ready for review. There are some tests failing, but it seems it's rather flakiness then our changes (not 100% sure, let me know).

I updated some docs as well, not sure if it's good enough.

lib/graphql/schema/argument.rb
```diff
@@ -129,6 +130,17 @@ def description(text = nil)
         end
       end
 
+      attr_writer :comment
```
rmosolgo, line 133: Why the `attr_writer` here? Can it be removed, since people can use `comment(...)` instead?

lib/graphql/language/nodes.rb
```diff
@@ -270,15 +270,17 @@ def scalars
             "col: nil",
             "pos: nil",
             "filename: nil",
-            "source: nil",
+            "source: nil"
           ]
 
+          IGNORED_MARSHALLING_KEYWORDS = [:comment]
```
rmosolgo, line 276: Why not marshal with the comment? (I'd expect a marshal-unmarshal document to be identical.)

lib/graphql/schema/interface.rb
```diff
@@ -22,7 +22,7 @@ module DefinitionMethods
         def definition_methods(&block)
           # Use an instance variable to tell whether it's been included previously or not;
           # You can't use constant detection because constants are brought into scope
-          # by `include`, which has already happened at this point.
+          # by `incllready happenude`, which has aed at this point.
```
rmosolgo, line 25:
```suggestion
          # by `include`, which has already happened at this point.
```

typo?

spec/graphql/language/printer_spec.rb
```diff
@@ -263,6 +263,174 @@
...
+        VIEWER
+      }
+    SCHEMA
+
+    assert_equal(
+      expected,
+      printer.print(schema.to_document),
+    )
+  end
+
```
rmosolgo, line 433: 😎 👍

spec/graphql/schema/object_spec.rb
```diff
@@ -375,6 +375,7 @@ def self.type_error(err, ctx)
         shape.delete(:@configs)
         shape.delete(:@future_schema)
         shape.delete(:@metadata)
+        shape.delete(:@comment)
```
rmosolgo, line 378: I think we should include `@comment` in our shape tests so that we can be sure that GraphQL-Ruby is creating efficient object shapes by default. 

We could remove this, and instead, add a line here: 

https://github.com/rmosolgo/graphql-ruby/blob/a794fc8e88660fa7d58837c25797cf529e4d45f0/lib/graphql/schema/member/base_dsl_methods.rb#L56-L59

For example, 

```ruby
child_class.comment(comment)
```

Which would initialize the instance variable and also implement comment inheritance. Alternatively, if you don't want inheritance, you could do `child_class.comment = NOT_CONFIGURED` (we'd need an `attr_writer` in that case)

rmosolgo (commented):
Thanks for taking another crack at this! It looks really good so far. I had a couple of questions and suggestions. 

(And yes, sorry about the flaky tests ... the `rails_master` test is especially flaky right now and I can't figure out why 😖 )

lib/graphql/schema/interface.rb
vaclavbohac: I have no idea what happened here :-D. Will fix.

lib/graphql/schema/argument.rb
vaclavbohac: I mirrored how the description was made for argument. If it's not necessary, (I don't have such an insight) I can remove it.

lib/graphql/language/nodes.rb
vaclavbohac: The C parser doesn't know about it and complains we require an argument it doesn't have, but we need the comment to exist on the nodes, cause the nodes are used when dumping the schema from the definition. So I made it a sort of default argument.

Maybe there's a better way.

spec/graphql/schema/object_spec.rb
vaclavbohac: Thanks, I've added the "cloning" of the comment to the child, seemed like a better option then setting it to empty default.

lib/graphql/language/nodes.rb
rmosolgo: :+1: That makes sense, thanks for the context!

lib/graphql/schema/argument.rb
vaclavbohac: @rmosolgo Is it ok, if it stays the same as description then?

rmosolgo: Yes, seems alright, thanks!

rmosolgo:
I made a few changes over in https://github.com/rmosolgo/graphql-ruby/pull/5077 - I added some debugging info to the failing tests around object shapes. I think something about the implementation of `def comment` was making things a bit weird, so I reworked it. 

I also changed the implementation so that comments are _not_ inherited. Thinking about it, if you annotate with a parent type with some kind of linter hint, I don't think the child type should have the same hint, unless it's manually configured. 

Also, I updated `def comment` to support `comment(nil)` (unsetting the comment), just in case. 

Do those changes look good to you? 

(Sorry, apparently I didn't accept the invitation to productboard/graphql-ruby in time so I can't join it. You can bring that work over to this branch, or I'll merge that PR instead of this one, either way suits me.)

vaclavbohac:
@rmosolgo I cherry picked your commits, looking great. Thank you!

rmosolgo:
Awesome. Thanks again for all your work to make this happen!

Bajena:
@rmosolgo Is there a chance you could release `2.3.15` soon(ish)? We'd love to start using the comments in our schema. 

Thank you! 🤟

rmosolgo:
Oh, yes, I just released it! Thanks again for this improvement :+1:

PR #5068, opened 2024-08-14
Use symbols for namespaced_types generator option

schpet:
this will allow the following to work:

    Rails.application.config.generators do |g|
      g.graphql namespaced_types: true

right now, the config needed is a bit quirky:

    Rails.application.config.generators do |g|
      g.graphql "namespaced_types" => true


rmosolgo:
Looks good, thanks for this improvement!

PR #5072, opened 2024-08-20
Allow variable definition directives in validation

adampetro:
Allow variable definition directives in the `DirectivesAreInValidLocations` static validation rule.

This also adds a test, but it feels a bit odd that it only tests the variable definition case and no others. I could remove the test, or add cases to cover more of the possible valid locations.


rmosolgo:
Thanks for this improvement!

PR #5080, opened 2024-09-02
Use `attr_reader` where applicable

amomchilov:
Just a small suggestion Rubocop found a while back.

Methods generated by `attr_reader` are technically ever-so-slightly faster than regular Ruby methods, though these aren't hot paths, so that doesn't matter much.


rmosolgo:

Issue #5081, opened 2024-09-02
Optimize some predicate methods with `attr_reader`


amomchilov:
When I was tinkering around and profiling GraphQL Ruby around a year ago, I found some small+easy perf wins from optimizing some commonly queried "predicate methods", by defining them via `attr_reader`. I've lost the benchmarks, and don't have time to recreate them, but I still figured it's worth sharing the idea here.

Unfortunately, Ruby makes these [unnecessarily ugly/difficult in the case of boolean attributes](https://bugs.ruby-lang.org/issues/19708), but still possible with this method swizzling dance:

```ruby
attr_reader :foo         # Define the fast method with a temporary name
alias_method :foo?, :foo # Give it a nice name
remove_method :foo       # Delete the temporary name
```

The main cuplrits were in `type_kind.rb`. I've opened a PoC PR here: 

Candidates for this kind of replacement can be found with these 2 regexes:

```regex
def ([a-z0-9_]+)\??[;\n]\s*@\1[;\n]\s*end
def ([a-z0-9_]+)\??\s*=\s*@\1
```

(They can be merged into one, but I opted not to, just to keep them somewhat legible)

PR #5082, opened 2024-09-02
PoC: Optimize predicate methods with `attr_reader`

amomchilov:
PoC for #5081

### DIFF
--- a/lib/graphql/type_kinds.rb
+++ b/lib/graphql/type_kinds.rb
@@ -16,22 +16,41 @@ module GraphQL
         @description = description
       end
 
+      # Is this TypeKind abstract?
+      attr_reader :abstract
+      alias_method :abstract?, :abstract
+      remove_method :abstract
+
       # Does this TypeKind have multiple possible implementers?
       # @deprecated Use `abstract?` instead of `resolves?`.
-      def resolves?;  @abstract;  end
-      # Is this TypeKind abstract?
-      def abstract?; @abstract; end
+      alias_method :resolves?, :abstract?
+
       # Does this TypeKind have queryable fields?
-      def fields?;    @fields;    end
+      attr_reader :fields
+      alias_method :fields?, :fields
+      remove_method :fields
+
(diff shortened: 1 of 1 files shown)


lib/graphql/type_kinds.rb
```diff
@@ -16,22 +16,41 @@ def initialize(name, abstract: false, leaf: false, fields: false, wraps: false,
         @description = description
       end
 
+      # Is this TypeKind abstract?
+      attr_reader :abstract
+      alias_method :abstract?, :abstract
+      remove_method :abstract
```
amomchilov, line 22: It's also possible to make a helper to automate these, but then IDEs won't know about the presence of the generated methods.

```ruby
module PredicateHelper # This can be mixed into Module, if you're into monkey-patching
  # Given `"foo?"`, this method generates an optimized method that reads from `@foo`.
  def attr_predicate(predicate_name)
    raise ArgumentError unless predicate_name.end_with?("?")
    
    attr_method_name = predicate_name.to_s.delete_suffix("?")
    attr_reader(attr_method_name)
    alias_method(predicate_name, attr_method_name)
    remove_method(attr_method_name)
  end
end
```

rmosolgo:
Hey, thanks for proposing this change! I'm definitely interested in it. Another implementation might be write a Rubocop rule which identifies and updates code that uses the "bad" approach. There are a couple other development-related cops here in case you're interested in try that: https://github.com/rmosolgo/graphql-ruby/tree/master/cop/development

amomchilov:
@rmosolgo Happy to hear! This is definitely rubocop-able. I'd even go so far as to add it upstream to https://github.com/rubocop/rubocop-performance

Would you mind running a quick benchmark on this, to see if it's worthwhile?

I have some env issues, and I won't have time to sort them out until at least next week

rmosolgo:
Sadly this doesn't merge into master anymore, but if you ever want to try again and share a benchmark, I'd welcome a speedup in this department!

PR #5092, opened 2024-09-09
Add frozen_string_literal to base_resolver template

tmanhkha:
Add `frozen_string_literal` to `base_resolver` template


rmosolgo:
Thanks for this improvement!

PR #5103, opened 2024-09-17
Don't raise NoMethodError on InvalidNullError.inspect

etiennebarrie:
I was listing the modules in my app with `ObjectSpace.each_object(Module)` and it failed while inspecting this class.

I also wanted to test the behavior of using the parent class's mutation but I couldn't find a way.


rmosolgo:
Thanks for this improvement!

PR #5113, opened 2024-09-29
fix logic that checks whether a request has a subscription header

bolandrm:
Hey, I'm a new Pro subscriber integrating Ably subscriptions.  It went well aside from this issue.

If you look further down in the code, `_getSubscriptionChannel` ALWAYS returns an object, regardless of whether the headers were passed:

```
  _getSubscriptionChannel(operation: Operation) {
    const response = operation.getContext().response
    // Check to see if the response has the header
    const subscriptionChannel = response.headers.get("X-Subscription-ID")
    // The server returns this header when encryption is enabled.
    const cipherKey = response.headers.get("X-Subscription-Key")
    return { channel: subscriptionChannel, key: cipherKey }
  }
```

So this logic check:

```
if (subscriptionChannelConfig) {
```

was always returning true.

### DIFF
--- a/javascript_client/src/subscriptions/AblyLink.ts
+++ b/javascript_client/src/subscriptions/AblyLink.ts
@@ -56,7 +56,7 @@ class AblyLink extends ApolloLink {
       forward(operation).subscribe({ next: (data) => {
         // If the operation has the subscription header, it's a subscription
         const subscriptionChannelConfig = this._getSubscriptionChannel(operation)
-        if (subscriptionChannelConfig) {
+        if (subscriptionChannelConfig.channel) {
           // This will keep pushing to `.next`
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hey, thanks for reporting this and sharing the fix! It looks good to me. I guess this old patch introduced the issue: https://github.com/rmosolgo/graphql-ruby/commit/191e8db766eac3305ba57e4027804024160c0c4a

rmosolgo:
I added some tests for this in 895074bec and released these changes to NPM in `graphql-ruby-client` 1.14.1 🚢 !

bolandrm:
I tried the new version and confirmed it's working for me.  Thanks!

Issue #5101, opened 2024-09-13
Incorrect/incomplete error message


fhoeben:
@rmosolgo thanks for the fix we indeed get an error message now instead of an exception.

Could you take a look whether it would be possible to change the error to take into account what is actually expected according to the schema (like we get when passing `-1`)? I now (2.3.16) get only a message and a location (no path). And the message is a bit more general than normally:
> Expected a number, but it was malformed (\"-\")

I expected a `path` to be present and a message like
> Argument 'arg' on Field 'hello' has an invalid value (-foo). Expected type 'MyEnum!'.

_Originally posted by @fhoeben in https://github.com/rmosolgo/graphql-ruby/issues/5088#issuecomment-2348298695_

rmosolgo:
Hey, sorry I didn't write back to this sooner. Basically, I don't have a vision for dramatically improving this error message. As discussed in #5115 it's much harder than most error messages because the malformation halts _lexical analysis_, not parsing, so there's no AST and no path. 

There are probably some dramatic changes to the lexer/parser that would make this possible, but I don't have plans to explore it further so I'm closing this issue. I'm happy to respond to questions in comments on this issue or to review a PR if you have any suggestions to make, code-wise!

PR #5115, opened 2024-10-02
Fix: Incorrect/incomplete error message

joaoGabriel55:
**Fix issue #5101**

Actual error message:
> _Expected a number, but it was malformed ("-")_

Expected error message:
> _Argument 'arg' on Field 'hello' has an invalid value (-foo). Expected type 'MyEnum!'._

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -72,6 +72,9 @@ module GraphQL
             # Check for a matched decimal:
             @scanner[1] ? :FLOAT : :INT
           else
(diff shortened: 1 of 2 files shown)


joaoGabriel55:
Hello @rmosolgo
I think the main goal would be to find a way to reuse this logic on our "new" error validation. The main problem is to convert the Query String passed on the `GraphQL.parse` method to an AST. I tried a lot of solutions using Regex but any of them covered complex cases like Query lists. Any ideas about it?

`lib/graphql/static_validation/rules/argument_literals_are_compatible.rb:L52`

PR #5116, opened 2024-10-04
bug fix: extensions can register more extensions via apply method

ElvinEfendi:
https://github.com/rmosolgo/graphql-ruby/pull/5054 introduced a regression where an extension registered eagerly (using the kwarg) can no longer register another extension in its `apply` method. I fix the bug here by restoring the previous `@call_after_define` logic.


rmosolgo:
Sorry for the trouble and thanks for this fix!

PR #5118, opened 2024-10-04
bug fix: respect options when using `trace_with` after `trace_class`.

ElvinEfendi:
https://github.com/rmosolgo/graphql-ruby/pull/5045 introduced a new behaviour to `add_trace_options_for` where it no longer merges options when trace class does not include `DefaultTraceClass`. While it did update `build_trace_mode` to include `DefaultTraceClass` in the trace classes, there is still `trace_class` method where it can register a new trace class in default mode without including `DefaultTraceClass`. So I fix the issue by ensuring `DefaultTraceClass` is included in the `new_class` in `trace_class`.

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -162,6 +162,7 @@ module GraphQL
           # re-apply them here
           mods = trace_modules_for(:default)
           mods.each { |mod| new_class.include(mod) }
+          new_class.include(DefaultTraceClass)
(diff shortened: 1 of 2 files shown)


ElvinEfendi:
@rmosolgo any chance we can get this released soon too?

rmosolgo:
Yes -- shipped in 2.3.18! Thanks for this fix.

PR #5130, opened 2024-10-21
Update validation message for one_of

alexcwatt:
This PR updates the error message when a `one_of` validation is not satisfied to be more explicit about what the issue is.

Before: "{field_name} has the wrong arguments"
After: "{field_name} must include exactly one of the following arguments: a, b."

Note: PR seems to be failing due to unreliable tests. [Example 1](https://github.com/rmosolgo/graphql-ruby/actions/runs/11447293332/job/31848303250?pr=5130), [Example 2](https://github.com/rmosolgo/graphql-ruby/actions/runs/11447896670/job/31850249461?pr=5130)

### DIFF
--- a/lib/graphql/schema/validator/required_validator.rb
+++ b/lib/graphql/schema/validator/required_validator.rb
@@ -37,7 +37,7 @@ module GraphQL
       class RequiredValidator < Validator
         # @param one_of [Symbol, Array<Symbol>] An argument, or a list of arguments, that represents a valid set of inputs for this field
         # @param message [String]
-        def initialize(one_of: nil, argument: nil, message: "%{validated} has the wrong arguments", **default_options)
+        def initialize(one_of: nil, argument: nil, message: nil, **default_options)
           @one_of = if one_of
             one_of
           elsif argument
@@ -49,7 +49,7 @@ module GraphQL
           super(**default_options)
         end
 
-        def validate(_object, _context, value)
+        def validate(_object, context, value)
           matched_conditions = 0
 
           if !value.nil?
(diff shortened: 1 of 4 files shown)


lib/graphql/schema/validator/required_validator.rb
```diff
@@ -76,6 +77,22 @@ def validate(_object, _context, value)
             @message
           end
         end
+
+        def build_message
+          formatted_arguments = @one_of.map do |arg|
+            if arg.is_a?(Array)
+              "(" + arg.map { |a| Schema::Member::BuildType.camelize(a.to_s) }.join(" and ") + ")"
```
rmosolgo, line 84: Hmm, `camelize` _is_ the GraphQL-Ruby default, but plenty of people disable it. This error message should reflect `camelize: false` if it applies to these arguments. 

`@validated` will be a reference to the Field or Resolver whose arguments are being validated, so I think you could do something like: 


~~`args = @validated.all_argument_definitions`~~

Wait, even that doesn't work, because the field could have _multiple_ argument definitions with the same name. I think we'd have to build this message at runtime instead of definition time to be sure that we have the currently-active definition. Instead of returning `@message`, maybe we could build it like: 

```ruby 
arg_defns = @validated.arguments(context).values 
required_names = @one_of.map do |arg_kw| 
   # TODO handle arrays too 
   arg_defn = arg_defns.find { |defn| defn.keyword == arg_kw } 
   arg_defn.graphql_name
 end
 # then build message with `arg_names`
 ```

alexcwatt: Thanks for the quick review! I pushed another commit to address this feedback. (First and last commit should be combined before merging, unless you squash everything on merge, but I didn't do this yet to make it clear what changed.)

rmosolgo (approved):
LGTM :+1: I don't have plans to modify the commit history before merging -- did you want to?

alexcwatt:
@rmosolgo great! I just squashed the first and last commit together and kept the middle one. You can see in GitHub UI that there were no changes in final state of this branch with the force-push.

PR #5131, opened 2024-10-21
No longer pin yard to specific commit

alexcwatt:
`bundle install` gave me a warning because yard was specified twice in Gemfile: once from the gemspec and once explicitly in the file.

It seems we no longer need to pin yard to a specific commit as Ruby 3.2 is supported in recent releases.


rmosolgo:
Thanks!

PR #5137, opened 2024-10-31
Require object types to have fields

### LINKED ISSUE #4509: GraphQL gem allows empty object types (in violation of GraphQL spec)
**Describe the bug**

According to the [GraphQL spec](http://spec.graphql.org/October2021/#sec-Objects.Type-Validation):

> An Object type must define one or more fields.

However, the GraphQL gem appears to treat object types that have no fields as valid.

**Versions**

`graphql` version: 2.0.22
other applicable versions (`graphql-batch`, etc): N/A

**GraphQL schema**

```graphql
type EmptyObject {
}

type Query {
  something: EmptyObject
}
```

**GraphQL query**

N/A

**Steps to reproduce**

Create a `graphql_gem_bug_no_fields.rb` file with these contents:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "graphql", "2.0.22"
end

schema_sdl = <<~EOS
  type EmptyObject {
  }

  type Query {
    something: EmptyObject
  }
EOS

require "graphql"

puts GraphQL::Schema.from_definition(schema_sdl).to_definition
```

Run the file.

**Expected behavior**

I expect the GraphQL gem's schema validation (which happens when you round-trip an SDL string through the gem's parser and dumper) to fail and indicate that `EmptyObject` isn't valid.

**Actual behavior**

The GraphQL gem treats this schema as valid valid, resulting in this output:

```
type EmptyObject

type Query {
  something: EmptyObject
}
```

**Additional context**

#1923 was previously reported for this issue and was apparently resolved by #2462.   So looks like a regression? Not sure when it crept back in.

rmosolgo:
The spec says that object types must have fields and that input object types must have arguments. This change adds that requirement to query execution and runtime. (Since Ruby classes are open, there's not a "compile-time" or "build-time" to check the definition.) 

This change keeps the current behavior when all fields are hidden by `.visible?` checks -- the type is still accessible but all fields are hidden. It might be possible to raise an error in that case but I don't think it's worth it.

For backwards compatibility (or ... _use cases_ ?), there's `has_no_fields(true)` and `has_no_arguments(true)`. Adding these configs will silence errors from GraphQL-Ruby.

Also, `RelayClassicMutation` (and `HasSingleInputArgument`) will still create empty input objects by default, with `has_no_arguments(true)` called on them. This is how it has always worked and I don't want to :boom: existing users 😅 

Fixes #4509 

**TODO**

- [x] Test errors raised when object has no fields 
- [x] Test errors raised when input object has no arguments 
- [x] Update tests for this constraint
- [x] Consider warning or backwards compat option? 
- [x] Add a deprecation warning; raise error in 3.0


gmac:
> For backwards compatibility (or ... use cases ?)

FWIW: "use cases" is legitimate. We use ActiveSupport load hooks to assemble root scopes from across the codebase, which means root scopes start as empty object types.

rmosolgo:
Thanks for sharing that, @gmac.  I would _expect_ this assertion to run during the first query that the schema executes, and by that time, any object type being used would have had its fields loaded. Is that right? Or did you find that this raised errors for you before those hooks ran? 

If you did encounter errors, I'd be interested to see the backtrace. My goal is for this to halt on execution or dumping the schema, but for it to _not_ halt on any boot-time steps.

gmac:
I think this is more on us than GraphQL Ruby; our approach is pretty novel. Our mutation root now basically looks like this:

```ruby
class MutationRoot < GraphQL::Schema::Object
  graphql_name "Mutation"

  if Rails.configuration.eager_load
    Hooks.run!
  else
    # allow no fields when running lazy load hooks
    has_no_fields(true)
  end
end
```

Then, across the codebase are load hooks that attach fields to an injected mutation object, ie:

```ruby
Hooks.configure_mutation_root do |m|
  m.field :do_stuff, true
end
```

So basically, we have to load `Mutation` to inject it into load hooks. The new validations are fine when load hooks run eagerly within the object scope, but block when hooks run lazily from the externally-curated dev loading sequence. `has_no_fields` gave us what we needed to make this work in development environments, so please don't write it off as superfluous!

rmosolgo:
Sure, I haven't heard of anyone using hooks to build the schema, but it should generally be _supported_ without errors. Ruby classes being open, etc, everything is _supposed_ to be open to extension at roughly any time. I'd be interested in seeing the backtrace on the error that this raised in your app, even just the GraphQL-Ruby parts, so I could get a sense for why it _didn't_ work in this case. 

Maybe it's because something in `Schema.mutation(MutationRoot)` looked up a field which raised this error? 

The intent of the implementation here was to allow objects without fields _until_ the object is used in a query. Because as described here, it's reasonably to have object definitions whose fields haven't been added yet. But I don't know of a use case for a object type with no fields at _query time_!

gmac:
Oh, I see what you’re saying. I’ll turn it off and see where the error happened. It was specific to tests with mutations. Sounds like dev env runs JIT hooks as part of the query itself; which wouldn’t surprise me with the aggressive startup optimizations that a dedicated team is making across the codebase.

PR #4966, opened 2024-05-21
Add DidYouMean to validation error messages

rmosolgo:
This adds `(Did you mean ...)` prompts to validation errors. You can disable it in your schema with: 

```ruby 
class MySchema < GraphQL::Schema 
  did_you_mean(nil)
end 
```

TODO: 

- [x] Add a test with multiple suggestions

Fixes #4918

PR #5144, opened 2024-11-05
handle non-existing field selection on enum type correctly

ElvinEfendi:
Since https://github.com/rmosolgo/graphql-ruby/pull/4966/files#diff-2252722dfaf6594778838471788047281b9c06a3f2419c5342a84221d75f086cR17, a query that select a (non existing) field on an enum type results in an exception: `"NoMethodError: undefined method 'fields' for class <an enum type>"`.


rmosolgo:
Thanks for this fix! Scalars might have the same problem -- I'll investigate after merging this.

PR #5145, opened 2024-11-05
Improve error message when enum has selections

rmosolgo:


rmosolgo:
I'm not sure why these tests about warnings started failing ... I'm going to track that down before I merge.

rmosolgo:
It's because a new version of `console` was released, and `console` is a dependency of `async`. I'm asking about it here: https://github.com/socketry/console/issues/63

Issue #5146, opened 2024-11-05
Intermittent bug with new Visibility code


sealabcore:
**Describe the bug**

We are seeing intermittent issues in the latest version of the graphql gem. Our app was raising on this error

```
GraphQL::Schema::DuplicateNamesError · Found two visible definitions for `Query.eventSchedules.userIds`: #<Types::BaseArgument Query.eventSchedules.userIds: [ID!]! @description="The user ids of the people to query for their event schedules">, #<Types::BaseArgument Query.eventSchedules.userIds: [ID!]! @description="The user ids of the people to query for their event schedules">
```

This was raising on every graphql request, even ones that didn't even have `Query.eventSchedules.userIds` in the query. So it seems like some weird caching issue in the new visiblity code.

We also get this error with the latest graphql code, but it's only once or twice a day right after deploy

```
FrozenError · can't modify frozen GraphQL::Schema::Field::ScopeExtension: #<GraphQL::Schema::Field::ScopeExtension:0x00007efdcc627408 @field=#<Types::BaseField Query.eventSchedules(...): EventScheduleConnection!>, @options={:call_after_define=>false}, @added_default_arguments=nil, @added_extras=nil>
```

These actually seem related, we got the FrozenError first when we deployed, and then started seeing the first error on every request after that.

**Versions**

`graphql` version: 2.4.1
`rails` (or other framework): Rails 7.2.2
other applicable versions (`graphql-batch`, etc)

**Steps to reproduce**

This would happen intermittently when deploying our app with the latest graphql gem.

**Expected behavior**

App should not raise on every graphql request.

**Actual behavior**

What specifically went wrong?

Place full backtrace here (if a Ruby exception is involved):

<details>
<summary>Click to view first frozen exception backtrace</summary>

```
vendor/bundle/ruby/3.3.0/gems/graphql-2.4.1/lib/graphql/schema/field_extension.rb:111:in `after_define_apply': can't modify frozen GraphQL::Schema::Field::ScopeExtension: #<GraphQL::Schema::Field::ScopeExtension:0x00007efdcc627408 @field=#<Types::BaseField Query.eventSchedules(...): EventScheduleConnection!>, @options={:call_after_define=>false}, @added_default_arguments=nil, @added_extras=nil> (FrozenError)
    from vendor/bundle/ruby/3.3.0/gems/graphql-2.4.1/lib/graphql/schema/field.rb:374:in `each'
[...]

PR #5147, opened 2024-11-06
Use a Mutex for Field#ensure_loaded

rmosolgo:
One possible solution to #5146 

But I'm not sure this is really the _best_ fix -- instead, it would be better if things were really preloaded in production.


rmosolgo:
I think #5148 will be a better approach

PR #5150, opened 2024-11-08
Fix double unsubscribe calls for Urql and Relay subscriptions

### LINKED ISSUE #5149: Issue with duplicate unsubscribe calls when using Urql & ActionCable
Hi,

This is follow up for PR https://github.com/rmosolgo/graphql-ruby/pull/5109. Not sure if my comments visible on closed PR, so I decided to create issue here.

---

I have spent some time debugging that issue with double unsubscriptions. Looks like the problem, at least in my case is as follows. Note that I'm using Urql with ActionCable.

Preconditions:
1. Have one Urql.Client for React application in Urql.Provider
2. Have 2 components, both are using subscription (Urql.useSubscription)

The problem exposes itself, when first component is unmounting (that triggers also unsubscribe() call), and the second is starting subscription negotiation.

In this case, new handler for second component receives {more: false} AC message, which trigger complete() callback, and I suspect that complete() callback in turn calls unsubscribe(), causing subscription to interrupt negotiation. This leaves second component unsubscribed.

I have reworked initial monkey patch from my previous message to following, which fixes doubled unsubscription calls.
Well, at least, in my case.

```javascript
import { Consumer, Subscription, logger } from "@rails/actioncable"
type ForwardCallback = (...args: any[]) => void

function createUrqlActionCableSubscription(consumer: Consumer, channelName: string = "GraphqlChannel") {
    return function (operation: Operation) {
        const subscribe = ({ next, error, complete }: { next: ForwardCallback, error: ForwardCallback, complete: ForwardCallback }) => {
            let subscribed = false;

            const subscription: Subscription = consumer.subscriptions.create(channelName, {
                connected() {
                    subscription.perform("execute", { query: operation.query, variables: operation.variables });
                    subscribed = true;
                },
                received(data: any) {
                    if (data?.result?.errors) {
                        error(data.errors);
                    }
                    if (data?.result?.data) {
                        next(data.result);
                    }
                    if (!data.more && subscribed) {
                        complete();
                    }
                }
            } as any);

            const unsubscribe = () => {
                if (subscribed) {
                    subscription?.unsubscribe();
                    subscribed = false;
                }
            };

            return { unsubscribe };
        };

        return { subscribe };
    };
}

const forwardToActionCable = createUrqlActionCableSubscription(actionCable);


const client = createClient({
    url: '/graphql',
    exchanges: [
        cacheExchange,
        fetchExchange,
        subscriptionExchange({
            forwardSubscription: forwardToActionCable as any
        }),
    ],
});
```

The change that matters at large is here:
```javascript
                    if (!data.more && subscribed) {
                        complete();
                    }
```

I'm not sure how this translates to createActionCableHandler, which is targeted in this PR https://github.com/rmosolgo/graphql-ruby/pull/5109, but here are my 2c. PR https://github.com/rmosolgo/graphql-ruby/pull/5109 may need revision.

vbatychko-modeln:
Fixes #5149

- prevent double unsubscribe calls in createUrqlActionCableSubscription
- call onCompleted only when we subscribed to AC for Relay & Urql handlers

### DIFF
--- a/javascript_client/src/subscriptions/SubscriptionExchange.ts
+++ b/javascript_client/src/subscriptions/SubscriptionExchange.ts
@@ -99,40 +99,45 @@ function createPusherSubscription(pusher: Pusher) {
 
 
 function createUrqlActionCableSubscription(consumer: Consumer, channelName: string = "GraphqlChannel") {
-  return function(operation: Urql.Operation) {
-    let subscription: Subscription | null = null
+  return function (operation: Urql.Operation) {
+    const subscribe = ({ next, error, complete }: { next: ForwardCallback, error: ForwardCallback, complete: ForwardCallback }) => {
+        let subscribed = false;
+
+        const subscription: Subscription = consumer.subscriptions.create(channelName, {
+            connected() {
+                subscription.perform("execute", { query: operation.query, variables: operation.variables });
+                subscribed = true;
+            },
+            received(data: any) {
+                if (data?.result?.errors) {
+                    error(data.errors);
+                }
+                if (data?.result?.data) {
+                    next(data.result);
+                }
+                if (!data.more && subscribed) {
+                    complete();
+                }
+            }
+        });
+
+        // urql will call `.unsubscribe()` if it's returned here:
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for sharing this fix! I just released it to NPM in v1.14.4.

PR #5167, opened 2024-11-21
Fix bug in DataLoader::Source#load_all when sending only falsy values

Baxxx:
In ruby:
```rb
[nil].any? #=> false
[false].any? #=> false
```

When sending [nil] to load_all, it was making the source think that there are no pending keys, and raised GraphQL::InvariantError.

### DIFF
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -73,7 +73,7 @@ module GraphQL
           end
         }
 
-        if pending_keys.any?
+        if pending_keys.size.positive?
           sync(pending_keys)
         end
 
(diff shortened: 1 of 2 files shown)


lib/graphql/dataloader/source.rb
```diff
@@ -73,7 +73,7 @@ def load_all(values)
           end
         }
 
-        if pending_keys.any?
+        if pending_keys.size.positive?
```
rmosolgo, line 76:
```suggestion
        if !pending_keys.empty?
```

rmosolgo:
Thanks for this fix! I have a Rubocop rule that checks for `.none?` for the same reason, I'm going to update it to also check for `.any?`.

PR #5178, opened 2024-11-29
Autoload nested constants

gmcgibbon:
Reduces require time from ~160ms to ~5ms. Autoloads all constants under `GraphQL` to optimize boot time. Also autoloads `GraphQL::Types` and `GraphQL::Query` because they can't be required in sequence.

Script:
```rb
#!/usr/bin/env ruby
# frozen_string_literal: true

require "bundler/setup"
require "benchmark"

ms = 1000 * Benchmark.realtime do
  require "graphql"
end

puts "#{ms}ms"
```

Before:

```
> bin/boot
169.55200000666082ms
```

After:

```
> bin/boot
5.60200004838407ms
```

Rails will automatically eager load GraphQL, but other frameworks / setups will need to call `GraphQL.eager_load!`. If we agree with the approach, I can document this. Because we don't require on Active Support, we need to manually implement eager loading. `ActiveSupport::Autoload` implements the `eager_load!` method for you.

Also, if you're interested, I can implement testing for eager loading (example), but we'll need to roll our own test isolation feature.

### DIFF
--- a/guides/code_loading.md
+++ b/guides/code_loading.md
@@ -1,0 +1,22 @@
+---
+layout: guide
+doc_stub: false
+search: true
+title: Code Loading
+section: Other
+desc: Read this before deploying GraphQL to production.
+---
+
+## Autoloading and Eager Loading
+
+GraphQL Ruby is autoloaded, which means most code won't be loaded until it is referenced. This is optimal for development and test environments where you want to boot your application as fast as possible. However, this is not optimal for production enviromnets.
+
+Production environments typically include multiple workers, and need to load an application upfront as much as possible. This ensures requests are as fast as possible at the cost of increased boot time, and forked processes don't need to load additional code. Unfortunately, there is no approach to eager code loading that is accepted by all web application frameworks.
+
+- For Rails applications, a Railtie is included that automatically eager-loads the GraphQL Ruby library for you. No action is required by the developer to opt into this behaviour.
+
+- For Sinatra applications, please put `configure(:production) { GraphQL.eager_load! }` in your application file.
+
+- For Hanami applications, please put `environment(:production) { GraphQL.eager_load! }` in your application file.
+
+- Other frameworks need to manually call `GraphQL.eager_load!` when their application is booting in production mode. If this is not done properly, GraphQL Ruby will log an warning.
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -5,8 +5,11 @@ require "set"
 require "singleton"
 require "forwardable"
 require "fiber/storage"
+require "graphql/autoload"
 
 module GraphQL
+  extend Autoload
+
   class Error < StandardError
   end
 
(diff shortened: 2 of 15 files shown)


gmcgibbon:
Looks like I have some tests to fix. I'm on it 👀

gmcgibbon:
I think the rest are happening on the main branch. Let me know if I'm wrong.

rmosolgo:
Hey, thanks so much for putting in the time on this. It's been in the back of my mind for a long time to improve the organization of all those `require`s, but I didn't think of approaching it this way. 


> other frameworks / setups will need to call `GraphQL.eager_load!` 

This seems alright to me. I'm trying to think if we can detect these and warn in cases where this should be called. Are there other setups that come to mind besides Sinatra and Hanami? (I mean, there's plain Rack, and a long tail of other possibilities, but any other big ones?)


>  testing for eager loading


I'm comfortable assuming that `autoload` works... but is there any way that we could make sure `eager_load!` will stay in-sync with our `autoload ...` configurations? For example, something like: 

```ruby 
all_constants = GraphQL.constants.map { |const_name| GraphQL.const_get(const_name) }
eager_loadable_constants = all_constants - [
  # Any specific exceptions to autoloading can go here, along with a comment for why they're skipped
]
assert_equal eager_loadable_constants, GraphQL.eager_load!
```

And then do something similar for `GraphQL::Query`. 

Yes, those other tests are failing on `master` 😿 . I haven't been able to get the `system_tests` to fail locally and I can't figure out how to fix the flaky test where database tables are somehow missing...

gmcgibbon:
> Are there other setups that come to mind besides Sinatra and Hanami?

I'll investigate this. Let's narrow the scope to Sinatra, Hanami, and Rails. Everything else can be supported later, and we can instruct folks to call that method themselves in more exotic cases.

> I'm comfortable assuming that autoload works... but is there any way that we could make sure eager_load! will stay in-sync with our autoload ... configurations?

Sounds good. I'll re-implement a minimal version of `ActiveSupport::Autload` so that we can automate the eager_load! method.

gmcgibbon:
I wrote a document page, added `GraphQL::Autoload`, added a warning for production not eager loading, and some rudimentary autoload tests. I can test the warning, but that would require an isolated test suite and re-implementing `ActiveSupport::Testing::Isolation`.

rmosolgo:
I'm looking over it one more time and adding more docs... it seems like `eager_load!` should continue recursively, calling `.eager_load!` on any loaded constants which respond to that method (eg, `GraphQL.eager_load!` should call `GraphQL::Query.eager_load!`). Does that sound right to you? I'll open a PR with the change to review.

rmosolgo:
🚢  in 2.4.5! Thanks again.

gmcgibbon:
Ah, you're right. Thanks for the correction!

sealabcore:
We are seeing this new error in our Rails app on web server boot, production console boot, and some rake tasks. Should we be seeing this? The docs in this PR mention that nothing should be needed when using the Graphql-Ruby gem with a Rails app.

Screenshot 2024-12-04 at 1 49 28 PM

Screenshot 2024-12-04 at 4 28 11 PM
Screenshot 2024-12-04 at 4 28 26 PM
Screenshot 2024-12-04 at 4 34 05 PM

rmosolgo:
Hey @sealabcore, no... it's not supposed to show that all the time :S I was able to replicate this immediately with `rails new ...`.  Digging in, its a problem with the _warning_ -- its detection for whether autoloading is happening or not is not sophisticated enough. 

First, Rails calls `Zeitwerk::Loader.eager_load_all`, which loads application code (including `app/graphql/my_schema.rb`), which references `GraphQL::Schema`, which emits this warning. 

Then, it does `eager_load_namespaces.each(&:eager_load!)` which eager-loads GraphQL's constants. 

Here's the source for that: https://github.com/rails/rails/blob/91d456366638ac6c3f6dec38670c8ada5e7c69b1/railties/lib/rails/application/finisher.rb#L79-L81

So I think everything _works_ -- everything is loaded when the application starts -- but the integration needs to be improved. I'll take a closer look tomorrow, sorry for the trouble!

gmcgibbon:
Hm, the application could reference any GraphQL constant in theory, so it might make more sense to use `before_eager_load { GraphQL.eager_load! }` instead of appending to `eager_load_namesapces`.

gmcgibbon:
I can confirm this fixes the issue. If we want to make the the gem is eager loaded before we eager load the app, I think this is our best bet, so I'll issue the patch. Sorry for the issues folks.

rmosolgo:
Thanks for taking care of it, @gmcgibbon! I just shipped that patch in 2.4.6. 

@sealabcore, thanks for reporting this and sorry for the trouble. Please give that new version a try and let me know if it gives you any more issues!

sealabcore:
Thanks to you both for the quick resolution! I’ll give this a try today.

zvkemp:
We've had a report that this change breaks constant loading if any of the nested files are required during app boot (I can open a new issue on this if necessary).

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'graphql', '2.4.7'
end

require 'graphql'
require 'graphql/tracing/active_support_notifications_trace'

GraphQL.eager_load!
```

What I think is happening is that the nested graphql/tracing/**.rb files don't require their parent (tracing.rb). So requiring this file:
 - calls `require "graphql/tracing/platform_trace"`, 
 - which loads platform_trace.rb, 
 - which encounters the autoload-registered constant `GraphQL::Tracing`
 - which calls `require 'graphql/tracing'
 - which calls `require 'graphql/tracing/platform_trace'`, which is skipped because this file has already been required — at this point the loading of the `PlatformTrace` module is higher up in the callstack, so it's not going to complete until after the rest of tracing.rb is loaded
 - Unfortunately, the rest of tracing.rb includes files that refer to PlatformTrace, so this error is raised:
`NameError: uninitialized constant GraphQL::Tracing::AppOpticsTrace::PlatformTrace`

I believe all of these nested files should include a top-level `require 'graphql/tracing'` to ensure the order is correct.

PR #5179, opened 2024-12-02
Add Autoload docs, make it recursive

rmosolgo:
Addresses the question in https://github.com/rmosolgo/graphql-ruby/pull/5178#issuecomment-2511846840


rmosolgo:
I initially implemented this with recursive `eager_load!` calls, but given that they're hardcoded now, I'm much less worried about the scope and impact of this change.

PR #5182, opened 2024-12-05
Use config.before_eager_load

gmcgibbon:
We want to make sure the gem is fully loaded in a production context before the application begins eagerly loading, otherwise we emit a warning. So, let's use `config.before_eager_load` to make sure gem eager loading happens before the app.

Before:
```
> RAILS_ENV=production bin/rails runner ""
GraphQL-Ruby thinks this is a production deployment but didn't eager-load its constants. Address this by:

  - Calling `GraphQL.eager_load!` in a production-only initializer or setup hook
  - Assign `GraphQL.env = "..."` to something _other_ than `"production"` (for example, `GraphQL.env = "development"`)

More details: https://graphql-ruby.org/schema/definition#production-considerations
```

After:
```
RAILS_ENV=production bin/rails runner ""
```

### DIFF
--- a/lib/graphql/railtie.rb
+++ b/lib/graphql/railtie.rb
@@ -10,7 +10,9 @@ module GraphQL
   class Railtie < Rails::Railtie
     config.graphql = ActiveSupport::OrderedOptions.new
     config.graphql.parser_cache = false
-    config.eager_load_namespaces << GraphQL
+    config.before_eager_load do
+      GraphQL.eager_load!
+    end
 
     initializer("graphql.cache") do |app|
       if config.graphql.parser_cache


rmosolgo:
Awesome, thanks!

sealabcore:
Looks like we are still getting this logger message on the latest version. Anything else I can try / info I can send over to help track it down?

Screenshot 2024-12-05 at 2 59 01 PM

rmosolgo:
Ugh, sorry 😖 Yes, to help debug, It'd be great to get a backtrace when that warning is created. That would show exactly what part of Rails caused GraphQL::Schema to be loaded. 

Here's a branch that will print the backtrace when that warning happens: https://github.com/rmosolgo/graphql-ruby/pull/5185

Does it print if you start the rails server locally? Could you bundle that version and share the backtrace? For example: 

```ruby 
gem "graphql", github: "rmosolgo/graphql-ruby", ref: "debug-eager-load-warning"
```

(You could also `bundle open graphql` and add that line of code yourself.) If you can share that backtrace, maybe it will have a clue about why our eager-loading hasn't already happened by that point!

gmcgibbon:
Ideally, if GraphQL client gem uses parts of this GraphQL gem, it would make sense for it to also be autoloaded. As for https://github.com/rmosolgo/graphql-ruby/pull/5182#issuecomment-2521660147, I would suspect something similar like a gem dependency is loading GraphQL before we expect it to. I need a backtrace to verify, though.

The problem mainly stems from the warning assuming that the schema will be loaded by the app, but it could be loaded by any dependent library at bundler require time. Most gems don't bother with this warning, and this may be part of why. We don't control what loads our code in consuming apps, and the worst case scenario is in more exotic use-cases, we don't load as much as we really should upfront (but it still works).

rmosolgo:
Yeah, I'm inclined to remove this warning since I can't exactly nail down when it should be logged: https://github.com/rmosolgo/graphql-ruby/pull/5187

What do you think, @gmcgibbon ?

gmcgibbon:
I agree removing it is probably the least error-prone and more future proof) approach. Let's remove it.

rmosolgo:
I removed it altogether and shipped 2.4.7. Thanks for your help tracking this down, everyone, and sorry the trouble along the way!

PR #5183, opened 2024-12-05
Add GraphQL::Query::Partial

### LINKED ISSUE #5136: Support running partial graphql queries
For `ObjectCache`, I could improve the granularity of caching if I could re-execute _parts_ (eg, uncacheable parts) of queries without re-running the whole thing. This comes with some possible foot-guns, since fields sometimes depend on "upstream" context for their resolution, but it could also work pretty well in most cases. 

I think the best API for me would be something like: 

```ruby
query = GraphQL::Query.new(...) # make a query somehow 
query_partial = query.partial_for("currentUser", "posts")
partial_result = query_partial.execute(@current_user.posts) # run the selected portion of the query using the given object as the underlying data
```

That would also support multiplexing, eg: 

```ruby 
query = GraphQLQuery.new(...)
# Run these partials concurrently (sharing a dataloader cache, etc) 
results = query.execute_partials({
  ["currentUser", "posts"] => @current_user.posts, 
  ["currentUser", "profile", "auditLog"] => AuditLog.events(user: @current_user), 
  # Etc 
})
```

Then it's up to the caller how to handle those results.

rmosolgo:
This is to implement partial execution, where you can take a prepared `GraphQL::Query` and run part of it in isolation. 

Fixes #5136 

### TODO 

- [x] Finish unit tests 
- [x] Add specific tests for Dataloader and lazy resolution 
- [x] DRY between Query and Query::Partial
- [x] Address TODOs in execution code
- [x] Dry `run_partials` and `run_partial_eager`?
- [x] What about when selections are duplicated in the AST? Have to merge fields appropriately
- [x] Add tests for continuing from non-null fields, unions, interfaces, scalars, enums
- [x] Add some way to spawn a partial from a query, or halt execution in a query. `ctx.skip`? 
  - `context.skip` works for now -- I can explore sugar for that down the road.
- [x] Test lazy root object in Partial -- object and scalar 
- [x] Test base_path usage, current_path setting in Partial


lib/graphql/query/context.rb
```diff
@@ -53,9 +53,6 @@ def initialize(query:, schema: query.schema, values:)
         @storage = Hash.new { |h, k| h[k] = {} }
         @storage[nil] = @provided_values
         @errors = []
-        @path = []
-        @value = nil
-        @context = self # for SharedMethods TODO delete sharedmethods
```
rmosolgo, line 58: These were leftover from a previous `GraphQL::Query::Context` implementation. I rediscovered them while working on this feature because at first, I thought I was going to need to use `query.context.dup`. I didn't end up doing that (instead, `Query::Context.new` inside `Partial#initialize`) but I'm going to keep these clean-ups.

lib/graphql/query/partial.rb
```diff
@@ -0,0 +1,90 @@
...
+      def analyzers
+        EmptyObjects::EMPTY_ARRAY
+      end
+
+      def current_trace
+        @query.current_trace
+      end
+
+      def analysis_errors=(_errs)
+      end
```
rmosolgo, line 38: TODO raise an error if this is actually assigned; or find some way to skip analysis checks altogether; it's not supported.

lib/graphql/query/partial.rb
```diff
@@ -0,0 +1,90 @@
...
+      def subscription?
+        false
+      end
+
+      def selected_operation
+        selection = @query.selected_operation
+        @path.each do |name_in_doc|
+          selection = selection.selections.find { |sel| sel.alias == name_in_doc || sel.name == name_in_doc }
+        end
+        selection
```
rmosolgo, line 49: TODO merge this with `root_type` -- they do the same dig into the AST, the duplicate work could be avoided.

PR #5185, opened 2024-12-06
Print backtrace when warning

rmosolgo:
Trying to hunt down why this is still printing in production sometimes.


rmosolgo:
(I broke a test that asserts that nothing is printed to stdout when this warning is emitted ...)

rmosolgo:
This warning was removed in #5187

PR #5186, opened 2024-12-06
Don't print production eager loading warning if already eager loaded

fusion2004:
This is my attempt to get the warning to not print if we've already eager loaded. Before, it was only checking if we were in the middle of eager loading, which always goes back to `false` when the eager load is complete.


fusion2004:
Ah, I see you're probably going to take a different approach now tho in #5187

rmosolgo:
Yes -- thanks for the suggestion but I think I have to give up on the idea of the warning 😩 !

Issue #5188, opened 2024-12-08
Upgraded from 2.4.4 to 2.4.7 getting uninitialized constant GraphQL::Schema::Object


janosrusiczki:
**Describe the bug**

As the title says, I upgraded my gem from 2.4.4 (which was working fine) to 2.4.7 (also tried with 2.4.6) and I'm getting the following error:

```
NameError: uninitialized constant GraphQL::Schema::Object (NameError)

@object_class || find_inherited_value(:object_class, GraphQL::Schema::Object)
                                                                    ^^^^^^^^
/home/runner/work/photonia/photonia/vendor/bundle/ruby/3.3.0/gems/graphql-2.4.7/lib/graphql/schema/resolver/has_payload_type.rb:56:in `object_class'
```

This is on the build machine when running `bin/rails db:schema:load` but it's the same on the local environment too when running anything that would initialize Rails.

To me it seems to be something related to autoloading but I can't figure it out.

**Versions**

`graphql` version: 2.4.7 (and 2.4.6)
`rails` (or other framework): 7.2.2
other applicable versions (`graphql-batch`, etc)

**GraphQL schema**

Not relevant.

**GraphQL query**

Not relevant.

**Steps to reproduce**

Run `rails server` or `rails db:schema:load`.

**Expected behavior**

Rails starts up.

**Actual behavior**

What specifically went wrong?

Place full backtrace here (if a Ruby exception is involved):

<details>
<summary>Click to view exception backtrace</summary>

```
bin/rails aborted!
NameError: uninitialized constant GraphQL::Schema::Object (NameError)

            @object_class || find_inherited_value(:object_class, GraphQL::Schema::Object)
                                                                                ^^^^^^^^
/home/runner/work/photonia/photonia/vendor/bundle/ruby/3.3.0/gems/graphql-2.4.7/lib/graphql/schema/resolver/has_payload_type.rb:56:in `object_class'
/home/runner/work/photonia/photonia/vendor/bundle/ruby/3.3.0/gems/graphql-2.4.7/lib/graphql/schema/resolver/has_payload_type.rb:92:in `generate_payload_type'
/home/runner/work/photonia/photonia/vendor/bundle/ruby/3.3.0/gems/graphql-2.4.7/lib/graphql/schema/mutation.rb:85:in `generate_payload_type'
[...]

PR #5190, opened 2024-12-09
require parent Tracing module before defining children

zvkemp:
(Continuing discussion from https://github.com/rmosolgo/graphql-ruby/pull/5178)

We've had a report that this change breaks constant loading if any of the nested files are required during app boot (I can open a new issue on this if necessary).

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'graphql', '2.4.7'
end

require 'graphql'
require 'graphql/tracing/active_support_notifications_trace'

GraphQL.eager_load!
```

What I think is happening is that the nested graphql/tracing/**.rb files don't require their parent (tracing.rb). So requiring this file:

- calls require "graphql/tracing/platform_trace",
- which loads platform_trace.rb,
- which encounters the autoload-registered constant GraphQL::Tracing
- which calls `require 'graphql/tracing'
- which calls require 'graphql/tracing/platform_trace', which is skipped because this file has already been required — at this point the loading of the PlatformTrace module is higher up in the callstack, so it's not going to complete until after the rest of tracing.rb is loaded
- Unfortunately, the rest of tracing.rb includes files that refer to PlatformTrace, so this error is raised:
`NameError: uninitialized constant GraphQL::Tracing::AppOpticsTrace::PlatformTrace`

I believe all of these nested files should include a top-level require 'graphql/tracing' to ensure the order is correct.

### DIFF
--- a/lib/graphql/tracing.rb
+++ b/lib/graphql/tracing.rb
@@ -1,38 +1,36 @@
 # frozen_string_literal: true
-require "graphql/tracing/trace"
-require "graphql/tracing/legacy_trace"
-require "graphql/tracing/legacy_hooks_trace"
 
-# Legacy tracing:
-require "graphql/tracing/active_support_notifications_tracing"
-require "graphql/tracing/platform_tracing"
-require "graphql/tracing/appoptics_tracing"
-require "graphql/tracing/appsignal_tracing"
-require "graphql/tracing/data_dog_tracing"
-require "graphql/tracing/new_relic_tracing"
-require "graphql/tracing/scout_tracing"
-require "graphql/tracing/statsd_tracing"
-require "graphql/tracing/prometheus_tracing"
-
-# New Tracing:
-require "graphql/tracing/active_support_notifications_trace"
-require "graphql/tracing/platform_trace"
-require "graphql/tracing/appoptics_trace"
-require "graphql/tracing/appsignal_trace"
-require "graphql/tracing/data_dog_trace"
-require "graphql/tracing/new_relic_trace"
-require "graphql/tracing/notifications_trace"
-require "graphql/tracing/sentry_trace"
-require "graphql/tracing/scout_trace"
-require "graphql/tracing/statsd_trace"
-require "graphql/tracing/prometheus_trace"
-if defined?(PrometheusExporter::Server)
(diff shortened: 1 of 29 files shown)


zvkemp:
^ Obviously tests are failing there — will address shortly. It's also throwing warnings about the circular requires, though I'm not sure there's a way to avoid that given the current structure.

rmosolgo:
Hey, thanks for the detailed report and this suggested fix. The same kind of problem popped up here: https://github.com/rmosolgo/graphql-ruby/issues/5188

I'm open to supporting these single-file requires, but do you mind sharing why it's important? For example, what if you remove that `require "graphql/tracing/active_support_notifications_trace"` from your app? 

I'm open to supported these nested requires, but I'd like to understand a little more about why they're necessary. (I'd _expect_ `require "graphql"` to make `GraphQL::Tracing::ActiveSupportNotificationsTrace` available.)

zvkemp:
Well we have a library that requires `"graphql/tracing/active_support_notifications_trace"`, so while we _can_ remove it, there are already published versions that refer to that file specifically, it's the only core tracing functionality we need (so it makes sense to have a targeted require), and we didn't anticipate this moving to an autoload-based strategy.

I'm not really sure of the utility of using autoloading here in the first place — as the readme correctly points out, this isn't recommended in production, so is this just a development/test mode optimization? FWIW I would happily trade the 160ms  saved for more predictable loading behavior.

rmosolgo:
Yeah, that's fair -- if it's code that's already out there, then let's make it work again. 

To clarify, do you need _all_ these traces, or just ActiveSupportNotificationsTrace?

rmosolgo:
To avoid the circular reference, I updated them all to `require "graphql/tracing/platform_trace"`. It looks to me like this'll work -- what do you think, @zvkemp ?

zvkemp:
I think the underlying issue is probably still the same — the file I used in my example already requires platform_trace.

rmosolgo:
Oh, I see -- my test wasn't enough because it didn't `require "graphql"` first 😖 . I'll take another look.

rmosolgo:
I slathered on some more `autoload` and I think it's working now 😅

PR #5193, opened 2024-12-17
Update C lexer to handle strings with null bytes

swalkinshaw:
`StringValueCStr` returns a null-terminated char* for a String but a Ruby string might contain null bytes and `StringValueCStr` will raise an `ArgumentError` ("string contains null byte").

This switches to using a `StringValuePtr` to get a (possibly unterminated) char* combined with `RSTRING_LEN` to get the string's length.

Note: I can't build or test this because I need ragel 7 which I have no idea how to install. But in theory this solution is good? 😅


rmosolgo (approved):

rmosolgo:
You can download Ragel directly from here: http://www.colm.net/open-source/ragel/

I think the zip file has instructions for installing 😅  I re-ran it and pushed the changes here, LGTM

rmosolgo:
🚢  in graphql_c-parser 1.1.2, thanks for the fix!

PR #5195, opened 2024-12-26
Make resolvers inherit description from superclass

rmosolgo:
Oops -- I think this was just an oversight that `description` wasn't passed down to children.

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -22,6 +22,7 @@ module GraphQL
       include Schema::Member::GraphQLTypeNames
       # Really we only need description & comment from here, but:
       extend Schema::Member::BaseDSLMethods
+      extend Member::BaseDSLMethods::ConfigurationExtension
       extend GraphQL::Schema::Member::HasArguments
       extend GraphQL::Schema::Member::HasValidators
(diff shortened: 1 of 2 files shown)


lib/graphql/schema/resolver.rb
```diff
@@ -22,6 +22,7 @@ class Resolver
       include Schema::Member::GraphQLTypeNames
       # Really we only need description & comment from here, but:
       extend Schema::Member::BaseDSLMethods
+      extend Member::BaseDSLMethods::ConfigurationExtension
```
gmac, line 25: There's a side-effect from this change where resolvers that inherit from other resolvers must now _explicitly_ declare a typename, otherwise they inherit the parent's typename and become a duplicate definition name.

Was this intentional? It's a bit counter-intuitive to the default precedent of elements following their class names.

gmac: FWIW, @swalkinshaw and I talked about the new behavior and agreed that it isn’t all bad. A resolver is effectively a representation of a Type, and type inheritance is generally discouraged in favor of sharing behavior through mixins. That’s internal precedent though. The former behavior is probably more consistent with the general library.

PR #5204, opened 2025-01-09
Fix definition.md copy from autoload to eager load

lukes:
Fix the copy from autoload to eager load.


lukes:
@rmosolgo Hello! I think I spotted a typo in the docs. Could you please review? 🙇

rmosolgo:
Thanks! Yes, that's the whole point 😆

PR #5206, opened 2025-01-12
Allow accessing enum value names via methods

DmitryTsepelev:
This is a pretty minor change, but might be helpful: now you can grab GraphQL name of the enum value using the method instead of a complex spell.

Before: `Jazz::Family.values["STRING"].graphql_name`
Now: `Jazz::Family.string`

### DIFF
--- a/guides/type_definitions/enums.md
+++ b/guides/type_definitions/enums.md
@@ -76,5 +76,5 @@ Enum classes are never instantiated and their methods are never called.
 You can get the GraphQL name of the enum value using the method matching its downcased name:
 
 ```ruby
-Types::MediaCategory.audio # => "Audio"
+Types::MediaCategory.audio # => "AUDIO"
 ```
--- a/guides/type_definitions/enums.md
+++ b/guides/type_definitions/enums.md
@@ -78,3 +78,19 @@ You can get the GraphQL name of the enum value using the method matching its dow
 ```ruby
 Types::MediaCategory.audio # => "AUDIO"
 ```
+
+You can pass a `value_method:` to override the value of the generated method:
+
+```ruby
+value "AUDIO", value: :audio, value_method: :lo_fi_audio
+
+# ...
+
+Types::MediaCategory.lo_fi_audio # => "AUDIO"
+```
+
+Also, you can completely skip the method generation by setting `value_method` to `false`
+
+```ruby
+value "AUDIO", value: :audio, value_method: false
+```
--- a/guides/type_definitions/enums.md
+++ b/guides/type_definitions/enums.md
@@ -72,3 +72,9 @@ value "AUDIO", value: :audio
 Then, GraphQL inputs of `AUDIO` will be converted to `:audio` and Ruby values of `:audio` will be converted to `"AUDIO"` in GraphQL responses.
 
 Enum classes are never instantiated and their methods are never called.
+
+You can get the GraphQL name of the enum value using the method matching its downcased name:
+
+```ruby
+Types::MediaCategory.audio # => "Audio"
+```
--- a/lib/graphql/schema/enum.rb
+++ b/lib/graphql/schema/enum.rb
@@ -61,13 +61,16 @@ module GraphQL
         # @option kwargs [String] :description, the GraphQL description for this value, present in documentation
         # @option kwargs [String] :comment, the GraphQL comment for this value, present in documentation
         # @option kwargs [::Object] :value the translated Ruby value for this object (defaults to `graphql_name`)
+        # @option kwargs [::Object] :value_method, the method name to fetch `graphql_name` (defaults to `graphql_name.downcase`)
         # @option kwargs [String] :deprecation_reason if this object is deprecated, include a message here
         # @return [void]
         # @see {Schema::EnumValue} which handles these inputs by default
-        def value(*args, **kwargs, &block)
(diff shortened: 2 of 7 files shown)


guides/type_definitions/enums.md
```diff
@@ -72,3 +72,9 @@ value "AUDIO", value: :audio
 Then, GraphQL inputs of `AUDIO` will be converted to `:audio` and Ruby values of `:audio` will be converted to `"AUDIO"` in GraphQL responses.
 
 Enum classes are never instantiated and their methods are never called.
+
+You can get the GraphQL name of the enum value using the method matching its downcased name:
+
+```ruby
+Types::MediaCategory.audio # => "Audio"
```
rmosolgo, line 79: Would this be `"AUDIO"`?

lib/graphql/schema/enum.rb
```diff
@@ -67,6 +67,7 @@ class << self
         def value(*args, **kwargs, &block)
           kwargs[:owner] = self
           value = enum_value_class.new(*args, **kwargs, &block)
+          singleton_class.define_method(value.graphql_name.downcase) { value.graphql_name }
```
rmosolgo, line 70: Implementation-wise, calling `define_method` with a block will result in the `binding` of this method staying in memory for the lifetime of the application. I'd like to avoid that memory "bloat", and we could probably do it with something like 

```ruby 
instance_eval("def #{value.graphql_name.downcase}; #{value.graphql_name.inspect}; end;")
```

It would create the method that returns the string, but without retaining a reference of the binding.

rmosolgo (commented):
Hey, thanks so much for this contribution and sorry I didn't get back to you til now. I'm definitely open to this change. I think the next few things to work out are: 

- What if the value's `graphql_name.downcase` conflicts with an existing method on `Schema::Enum`, like `value`, `kind`, or `inherited`? I think we should emit some kind of warning that the method generation was skipped and provide an override option like `value_method:`, whose value will be used instead of `graphql_name.downcase`.

- In case this causes some problems for people who are already using the gem, I'd like to have some way for them to turn it off. For example, if a `value_method:` option was added, `value_method: false` could disable creating these methods, and they could set that option by default in their base enum value classes. 

Thanks, and let me know what you think!

guides/type_definitions/enums.md
DmitryTsepelev: my bad

DmitryTsepelev:
Thanks for the feedback, I did my best to address everything! I also added a CI fix for unavailable Logger class

lib/graphql/schema/enum.rb
```diff
@@ -223,6 +227,20 @@ def inherited(child_class)
...
           @own_values ||= {}
         end
+
+        def generate_value_method(value, configured_value_method)
+          return if configured_value_method == false
+
+          value_method_name = configured_value_method || value.graphql_name.downcase
+
+          if respond_to?(value_method_name.to_sym)
+            $stderr << "Failed to define value method for :#{value_method_name}, because " \
```
rmosolgo, line 237: Out of curiousity ... is this different than `warn(...)` ?

lib/graphql/schema/enum.rb
```diff
@@ -223,6 +227,20 @@ def inherited(child_class)
...
+              "#{value.owner.name} already responds to that method. Use `value_name:` to override the method name."
```
rmosolgo, line 238:
```suggestion
            warn "Failed to define value method for :#{value_method_name}, because " \
              "#{value.owner.name} already responds to that method. Use `value_name:` to override the method name."
```

spec/graphql/schema/enum_spec.rb
```diff
@@ -10,6 +10,45 @@
...
+      it "use custom method" do
+        assert_equal enum.respond_to?(:percussion), false
+        assert_equal enum.precussion_custom_value_method, "PERCUSSION"
+      end
+    end
+
+    describe "when value_method conflicts with existing method" do
+      it "does not define method and emits warning" do
+        expected_message = "Failed to define value method for :value, because " \
+          "ConflictEnum already responds to that method. Use `value_name:` to override the method name."
```
rmosolgo, line 32:
```suggestion
          "ConflictEnum already responds to that method. Use `value_name:` to override the method name.\n"
```

rmosolgo:
Thanks for this improvement!

lib/graphql/schema/enum.rb
DmitryTsepelev: nope :)

gmac:
> For example, if a value_method: option was added, value_method: false could disable creating these methods, and they could set that option by default in their base enum value classes.

For future consideration, this should have been an opt-in rather than an opt-out, IMHO. Anything that adds strictly-convenience APIs with unlimited name cardinality to an existing schema structure should be done with care.

gmac:
We have a lot of superclassing over all major GraphQL structures to provide various internal concerns. It’s always a measured risk adding atop base library APIs, but they tend to evolve slowly and in isolated ways when they change in a conflicting manner. This change effectively created N new potentially conflicting APIs, where N is our own enum value surface area. I just overrode the enum constructor to disable the new methods everywhere, but it feels like that should have been the default.

gmac:
I've opened this as a formal bug in https://github.com/rmosolgo/graphql-ruby/issues/5254. This really should be off-by-default.

PR #5210, opened 2025-01-21
Fix two bugs (order dependency and false positive) in the `required` validation

harrylewis:
The goal of this PR is to fix two bugs identified in the `required` validation.

### Bug <span>#</span>1: Order dependency for matching sets

When defining a required validation, if the first option defined is a set and that option is matched, the validation will pass, even if other options in the mutually exclusive group would match.

Consider the following query.

```gql
{
  findObject(node_id: '1', object_type: 'Invoice', object_id: '1') {
    id
  }
}
```

The behaviour of the required validation is different depending on how the validation is defined.

```ruby
field :find_object, Node, null: true do
  argument :node_id, ID, required: false
  argument :object_type, String, required: false
  argument :object_id, Integer, required: false

  # This will return a validation error.
  validates required: { one_of: [:node_id, [:object_type, :object_id]] }

  # This will NOT return a validation error.
  validates required: { one_of: [[:object_type, :object_id], :node_id] }
end
```

This is addressed in https://github.com/rmosolgo/graphql-ruby/commit/f2e4c0d7d48c4289cab1fcb04ebcc0b813136cc2.

### Bug <span>#</span>2: False positive for partially matched sets

When defining a required validation, if any of the options is matched alongside a set that is partially matched, the validation will pass. This violates the mutual exclusion rule.

Consider the following query.

```gql
{
  findObject(node_id: '1', object_type: 'Invoice') {
    id
  }
}
```

I would expect this to fail, given the following field definition.

```ruby
field :find_object, Node, null: true do
  argument :node_id, ID, required: false
  argument :object_type, String, required: false
  argument :object_id, Integer, required: false

  # This will NOT return a validation error.
  validates required: { one_of: [:node_id, [:object_type, :object_id]] }
end
```

This is addressed in https://github.com/rmosolgo/graphql-ruby/commit/f2e4c0d7d48c4289cab1fcb04ebcc0b813136cc2.

_Note: I came across this behaviour during development, and was surprised and confused by it. I can anticipate a perspective that deems this behaviour as acceptable because it still technically satisfies the property that "exactly one" option is matched. While I can appreciate this perspective, in my opinion it leaves room for ambiguity in the behaviour of validation._

### Developer impact

The first bug requires a developer to be specific when defining the order of the mutually exclusive options. This leads to a brittle implementation, where a seemingly innocuous reordering of the options can causing a breaking change in the API.

The second bug does not uphold the contract of "exactly one" and "mutually exclusive" that the required validation is meant to provide, _in my opinion_. If I am checking for the presence of arguments in the resolver to determine which of the mutually exclusive option has been provided, depending on the order I check the presence of the keys, if one of those options is a set that has been partially matched, I could attempt to access the other keys in the set which have not matched and cause a runtime error.

### Breaking changes

I consider both of these bug fixes to be breaking changes. A client passing a certain collection of inputs which may be valid under the existing behaviour may now experience a validation error, if they are not providing strictly mutually exclusive arguments.

### DIFF
--- a/lib/graphql/schema/validator/required_validator.rb
+++ b/lib/graphql/schema/validator/required_validator.rb
@@ -62,8 +62,18 @@ module GraphQL
                   fully_matched_conditions += 1
                 end
               when Array
-                full_match = one_of_condition.all? { |k| value.key?(k) }
-                partial_match = !full_match && one_of_condition.any? { |k| value.key?(k) }
+                any_match = false
+                full_match = true
+
+                one_of_condition.each do |k|
+                  if value.key?(k)
+                    any_match = true
+                  else
+                    full_match = false
+                  end
+                end
+
+                partial_match = !full_match && any_match
 
                 if full_match
                   fully_matched_conditions += 1
--- a/lib/graphql/schema/validator/required_validator.rb
+++ b/lib/graphql/schema/validator/required_validator.rb
@@ -63,7 +63,6 @@ module GraphQL
               when Array
                 if one_of_condition.all? { |k| value.key?(k) }
                   matched_conditions += 1
-                  break
                 end
               else
                 raise ArgumentError, "Unknown one_of condition: #{one_of_condition.inspect}"
--- a/lib/graphql/schema/validator/required_validator.rb
+++ b/lib/graphql/schema/validator/required_validator.rb
@@ -51,18 +51,26 @@ module GraphQL
         end
 
         def validate(_object, context, value)
-          matched_conditions = 0
+          fully_matched_conditions = 0
+          partially_matched_conditions = 0
 
           if !value.nil?
             @one_of.each do |one_of_condition|
               case one_of_condition
               when Symbol
                 if value.key?(one_of_condition)
-                  matched_conditions += 1
+                  fully_matched_conditions += 1
                 end
               when Array
-                if one_of_condition.all? { |k| value.key?(k) }
-                  matched_conditions += 1
+                full_match = one_of_condition.all? { |k| value.key?(k) }
+                partial_match = !full_match && one_of_condition.any? { |k| value.key?(k) }
+
+                if full_match
+                  fully_matched_conditions += 1
+                end
+
+                if partial_match
+                  partially_matched_conditions += 1
                 end
               else
                 raise ArgumentError, "Unknown one_of condition: #{one_of_condition.inspect}"
@@ -70,7 +78,7 @@ module GraphQL
             end
           end
 
-          if matched_conditions == 1
+          if fully_matched_conditions == 1 && partially_matched_conditions == 0
             nil # OK
           else
             @message || build_message(context)
--- a/spec/graphql/schema/validator/required_validator_spec.rb
+++ b/spec/graphql/schema/validator/required_validator_spec.rb
@@ -31,6 +31,18 @@ describe GraphQL::Schema::Validator::RequiredValidator do
       ]
     },
     {
+      name: "Definition order independence",
+      config: { one_of: [[:a, :b], :c] },
+      cases: [
+        { query: "{ validated: multiValidated(c: 1) }", result: 1, error_messages: [] },
+        { query: "{ validated: multiValidated(a: 2, b: 3) }", result: 5, error_messages: [] },
+        { query: "{ validated: multiValidated }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(a: 1, b: 2, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(a: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(b: 2) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+      ]
+    },
+    {
       name: "Input object validation",
       config: { one_of: [:a, [:b, :c]] },
       cases: [
--- a/spec/graphql/schema/validator/required_validator_spec.rb
+++ b/spec/graphql/schema/validator/required_validator_spec.rb
@@ -26,6 +26,8 @@ describe GraphQL::Schema::Validator::RequiredValidator do
         { query: "{ validated: multiValidated(b: 2, c: 3) }", result: 5, error_messages: [] },
         { query: "{ validated: multiValidated }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
         { query: "{ validated: multiValidated(a: 1, b: 2, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
+        { query: "{ validated: multiValidated(a: 1, b: 2) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
+        { query: "{ validated: multiValidated(a: 1, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
         { query: "{ validated: multiValidated(c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
         { query: "{ validated: multiValidated(b: 2) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
       ]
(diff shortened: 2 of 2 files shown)


lib/graphql/schema/validator/required_validator.rb
```diff
@@ -51,27 +51,34 @@ def initialize(one_of: nil, argument: nil, message: nil, **default_options)
...
                 if value.key?(one_of_condition)
-                  matched_conditions += 1
+                  fully_matched_conditions += 1
                 end
               when Array
-                if one_of_condition.all? { |k| value.key?(k) }
-                  matched_conditions += 1
-                  break
+                full_match = one_of_condition.all? { |k| value.key?(k) }
+                partial_match = !full_match && one_of_condition.any? { |k| value.key?(k) }
```
harrylewis, line 66: I discovered that `#some?` is only available in Rails 🙃

lib/graphql/schema/validator/required_validator.rb
```diff
@@ -51,27 +51,34 @@ def initialize(one_of: nil, argument: nil, message: nil, **default_options)
...
+                  partially_matched_conditions += 1
                 end
               else
                 raise ArgumentError, "Unknown one_of condition: #{one_of_condition.inspect}"
               end
             end
           end
 
-          if matched_conditions == 1
+          if fully_matched_conditions == 1 && partially_matched_conditions == 0
```
harrylewis, line 81: A partial match alongside a full match violates a stricter definition of "exactly one" and "mutually exclusive".

spec/graphql/schema/validator/required_validator_spec.rb
```diff
@@ -26,17 +26,35 @@
         { query: "{ validated: multiValidated(b: 2, c: 3) }", result: 5, error_messages: [] },
         { query: "{ validated: multiValidated }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
         { query: "{ validated: multiValidated(a: 1, b: 2, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
+        { query: "{ validated: multiValidated(a: 1, b: 2) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
+        { query: "{ validated: multiValidated(a: 1, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: a, (b and c)."] },
```
harrylewis, line 30: These tests exercise the partial matches. Now every combination of the three arguments in this example is evaluated. This has been duplicated below for other cases.

spec/graphql/schema/validator/required_validator_spec.rb
```diff
@@ -26,17 +26,35 @@
...
+        { query: "{ validated: multiValidated(a: 2, b: 3) }", result: 5, error_messages: [] },
+        { query: "{ validated: multiValidated }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(a: 1, b: 2, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(a: 1, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(b: 2, c: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(a: 3) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+        { query: "{ validated: multiValidated(b: 2) }", result: nil, error_messages: ["multiValidated must include exactly one of the following arguments: (a and b), c."] },
+      ]
+    },
```
harrylewis, line 48: This exercises the order dependency issue.

lib/graphql/schema/validator/required_validator.rb
```diff
@@ -51,27 +51,34 @@ def initialize(one_of: nil, argument: nil, message: nil, **default_options)
...
               case one_of_condition
               when Symbol
                 if value.key?(one_of_condition)
-                  matched_conditions += 1
+                  fully_matched_conditions += 1
                 end
               when Array
-                if one_of_condition.all? { |k| value.key?(k) }
-                  matched_conditions += 1
-                  break
```
harrylewis, line 66: This `break` statement is what caused the order dependency issue. I looked in the Git history to see why it was present. It existed in the original commit, without an explicit explanation. Upon further evaluation and experimentation, I discovered it led to the order dependency mentioned. Removing it outright fixes this order depedency.

harrylewis (commented):
I see CI is failing for certain checks. I see the same failures on other PR's, so it doesn't appear to be related to my changes. I'll take a look and see what is going wrong.

lib/graphql/schema/validator/required_validator.rb
rmosolgo: I think we could do it in a single iteration of `one_of_condition`: 

```suggestion
                any_match = false 
                full_match = true
                one_of_condition.each do |k|
                  if value.key?(k)
                    any_match = true
                  else 
                    full_match = false 
                  end 
                end 
                
                partial_match = any_match && !full_match 
```

rmosolgo (commented):
Hey, thanks for the detailed writeup and code change. I just have one suggestion on the implementation.

lib/graphql/schema/validator/required_validator.rb
harrylewis: Thanks for this suggestion, I'll go ahead and implement that!

Is this suggestion motivated by potential performance concerns? My GraphQL performance knowledge is limited. For most cases I imagine that one loop versus two loops for a relatively small amount of options would have a negligible impact. Of course though, it is still additional CPU cycles. But this is where my expertise and use case awareness falls off - are there use cases where this could potentially cause a performance impact?

rmosolgo: Yes, it was performance-minded. In practice, I can't think of a situation where this would be called in a tight loop and become a real bottleneck. But I have certainly been surprised _before_ about how rubber meets the road when GraphQL-Ruby is used in real-world applications. 

In micro-benchmarks, it makes a difference, for example: 

```ruby
# loops.rb
require "benchmark/ips"


one_of_condition = [:a, :b, :c, :d]
value = {
  a: 1,
  b: 2,
  c: 3
}
Benchmark.ips do |x|
  x.report(".all + .any") do
    full_match = one_of_condition.all? { |k| value.key?(k) }
    partial_match = !full_match && one_of_condition.any? { |k| value.key?(k) }
  end
  x.report(".each") do
    any_match = false
    full_match = true
    one_of_condition.each do |k|
      if value.key?(k)
        any_match = true
      else
        full_match = false
      end
    end

    partial_match = any_match && !full_match
  end

  x.compare!
end
```

Without YJIT, `.all` + `.any` is 23% slower: 

```
$ ruby loops.rb
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-darwin22]
         .all + .any   247.839k i/100ms
               .each   273.355k i/100ms
         .all + .any      2.269M (± 5.2%) i/s  (440.78 ns/i) -     11.401M in   5.039404s
               .each      2.797M (± 4.3%) i/s  (357.52 ns/i) -     14.214M in   5.091697s

Comparison:
               .each:  2797017.5 i/s
         .all + .any:  2268712.9 i/s - 1.23x  slower
```

Interestingly, with YJIT enabled, there's a much bigger difference (greater than 2x faster with `.each`): 

```
$ ruby --yjit loops.rb
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [x86_64-darwin22]
         .all + .any   324.644k i/100ms
               .each   759.224k i/100ms
         .all + .any      3.406M (± 5.4%) i/s  (293.57 ns/i) -     17.206M in   5.066546s
               .each      8.336M (± 5.4%) i/s  (119.95 ns/i) -     41.757M in   5.025170s

Comparison:
               .each:  8336485.6 i/s
         .all + .any:  3406394.2 i/s - 2.45x  slower
```

Maybe YJIT has special handling for `Array#each` 🤷

rmosolgo:
Thanks again for this fix!

lib/graphql/schema/validator/required_validator.rb
harrylewis: I really appreciate you taking the time to explain this, and provide this fulsome example. There is clearly a difference here. While it is minimal in this context, it's easy to address and removes any potential for this to be an added source of slowness.

lib/graphql/schema/validator/required_validator.rb
gmac: Funny, we have an internal helper called `validates_at_least_one_of` that was exploiting your original thinking, Robert. :)

PR #5211, opened 2025-01-21
Fix CI failures for latest version of `ActiveSupport` 7

harrylewis:
This attempts to fix some CI failures related to a recent update to `concurrent-ruby`, which is causing issues in Rails/ActiveSupport, as seen in https://github.com/rails/rails/pull/54264.


harrylewis:
Closing this for now. I don't want to keep using up GitHub Actions credits. I'm going to open up an issue sharing my findings so far.

PR #5213, opened 2025-01-24
Improve built-in Rails support for Dataloader

rmosolgo:
I was helping someone move database access to Dataloader and decided that some of the "boilerplate" could be shipped with the gem: 

- [x] Record and Association source classes 
  - [x] Check for already-loaded associations before pausing the Fiber
- ~~ActiveStorage sources, others?~~ not now
- [x] Helper methods for built-in sources 
- Field config helpers to automatically call dataload helpers (eg `dataload_association: true`)
- Would enabling ActiveRecord Strict mode help catch these missing batch loads?
  - Tried this in https://github.com/rmosolgo/graphql-ruby/compare/try-strict-loading?expand=1
- ~~Add trace hooks for dataloader sources~~ done in #5214


dbackeus:
Is "how to use" documented somewhere?

PR #5214, opened 2025-01-27
Add PerfettoTracing for detailed query execution profiling

rmosolgo:
This extends `GraphQL::Tracing` to have some different hook methods and adds an implementation for the [Perfetto Trace Viewer](https://perfetto.dev/). 

image

I was looking for something with better support for async trace rendering and discovered Perfetto. In the end, I'm not really using its _parallel_ trace capability, but its custom "tracks" and "flows" are great because it can link field execution to `Dataloader::Source#fetch` calls.

TODO: 

- [x] Add new methods to base `Trace` class 
- [ ] In another PR, update other tracing implementations to use `begin`/`end` hooks to support Fiber pauses better
- [x] Add a development cop which checks for `super` in implementation methods
- [x] Backwards compat for `Dataloader#yield` without args?
- [x] Handle the optional dependency of `google-protobuf` nicely
- [x] Test with AsyncDataloader 
- [x] Document


rmosolgo:
Still some flaky tests 😓

PR #5220, opened 2025-02-03
Avoid warnings during Schema.from_introspection

eizengan:
Avoid stderr output when using `GraphQL::Schema.from_introspection` to load a Scheme containing Enums with values like e.g. `"NAME"`. 

Copied somewhat-naively from a [similar change](https://github.com/rmosolgo/graphql-ruby/commit/52153a243e265fb75b01d1a6a62766a3583ef2c9). Please ensure this solution is appropriate!

Not sure where/if I should add a test


rmosolgo:
Thanks for this improvement! I fixed the warnings from the existing tests but I guess there wasn't already one here 😖

eizengan:
All good - thanks for paving the way :)

Issue #5222, opened 2025-02-05
GraphQL::Backtrace::TracedError behavior


yn-misaki:
**Describe the bug**

I noticed a behavioral change in GraphQL version 2.4.9. I would like to confirm if this is an intended change in specification or if it's a potential bug.

- Previous Behavior (2.4.8): Exceptions were caught directly by rescue_from blocks as defined in the schema.
- Current Behavior (2.4.9): Exceptions are now wrapped in GraphQL::Backtrace::TracedError, preventing rescue_from from catching them as expected.

**Versions**

`graphql` version: 2.4.9
`graphql-c_parser` version: 1.1.2

**GraphQL schema**

```ruby
module Types
  class AddressType < GraphQL::Schema::Object
    def self.authorized?(object, context)
       raise ApplicationSchema::CustomException
    end

    field :id
  end
end

class ApplicationSchema < GraphQL::Schema
  query Types::QueryType
  use GraphQL::Dataloader
  use GraphQL::Backtrace

  class CutomException < StandardError; end
  rescue_from(CustomException) do |err|
    raise GraphQL::ExecutionError.new(err.message, extensions: { code: "FORBIDDEN" })
  end
end
```

**GraphQL query**

Example GraphQL query and response (if query execution is involved)

```graphql
query {
    query {
    node(id: "AddressID") {
      ... on Address {
       id
      }
    }
  }
}
```

<details>
<summary>Click to view response</summary>

```json
{"data"=>nil,
 "errors"=>
  [{"message"=>
     "Unhandled error during GraphQL execution:\n" +
     "\n" +
     "  ApplicationSchema::CutomException\n" +
     "    /app/app/graphql/types/address_type.rb:6:in `authorized?'\n" +
     "    /usr/local/bundle/gems/graphql-2.4.9/lib/graphql/schema/object.rb:70:in `block in authorized_new'\n" +
     "    /usr/local/bundle/gems/graphql-2.4.9/lib/graphql/tracing/trace.rb:63:in `authorized'\n" +
     "    /usr/local/bundle/gems/graphql-2.4.9/lib/graphql/tracing/platform_trace.rb:75:in `block in authorized'\n" +
     "    /usr/local/bundle/gems/graphql-2.4.9/lib/graphql/tracing/sentry_trace.rb:86:in `block in instrument_sentry_execution'\n" +
[...]

PR #5227, opened 2025-02-10
Fix GraphQL::Backtrace::TracedError handling

yn-misaki:
Fix: https://github.com/rmosolgo/graphql-ruby/issues/5222

In GraphQL 2.4.9, there was a change in how exceptions are handled when using `GraphQL::Backtrace`. 
This PR restores the 2.4.8 behavior when using backtrace and rescue_from.


rmosolgo:
Thanks so much for this fix!

PR #5229, opened 2025-02-10
More perfetto

rmosolgo:
- Add support for: 
  - [x] resolve type
  - [x] authorized
  - [x] validate 
- [x] reconsider method signatures for `execute_field`, include arguments in debug_annotations
- [x] add any helpful debug_annotations
  - What about AR::Relations? Right now it's querying for the first 10, the default `#inspect` behavior
- [x] migrate to `iid`s 
- [x] Add categories?


rmosolgo:
The current problem is that some of the AS::N attributes contain filepaths which ... don't match between CI and my laptop 🙄 I need to keep improving my snapshots

Issue #5223, opened 2025-02-06
Loading schema from Github outputs a large number of warnings with no way to turn them off


phedinkus:
**Describe the bug**

We upgraded to version 2.4.9 through the [graphql-client gem](https://github.com/github-community-projects/graphql-client). Their instructions show how to load the schema from a JSON file using this gem like so:
```ruby
HTTP = GraphQL::Client::HTTP.new("https://example.com/graphql")

Schema = GraphQL::Client.load_schema("path/to/schema.json")

Client = GraphQL::Client.new(schema: Schema, execute: HTTP)
```
The `GraphQL::Client.load_schema` method calls `GraphQL::Schema.from_introspection(schema)`. 

We are loading the schema for Github's GraphQL API and experience a large number of warnings on their schema. The warning states we should pass the `value_method` param. Since we aren't creating the schema JSON it would be nice to pass this as an option to the `from_introspection` method to indicate `value_method: false` or that it's an external schema and we are not going to modify it to use the `value_method` param. 

**Versions**

`graphql` version: 2.4.9
`rails` (or other framework): rails, graphql-client

**GraphQL schema**

https://docs.github.com/en/graphql/overview/public-schema

**Steps to reproduce**

* Fetch the schema from [Github](https://api.github.com/graphql) (this requires an API key) and save to a file
* Run the following in the IRB 
```ruby
schema = File.read("path/to/schema.json").to_h
GraphQL::Schema.from_introspection(schema)
```
**Expected behavior**

It would be great to have a param we can pass to the from_introspection method to suppress the warnings 

**Actual behavior**

Here's the output we get:
```
Failed to define value method for :name, because EnterpriseOrderField already responds to that method. Use `value_method:` to override the method name or `value_method: false` to disable Enum value method generation.
Failed to define value method for :name, because EnvironmentOrderField already responds to that method. Use `value_method:` to override the method name or `value_method: false` to disable Enum value method generation.
Failed to define value method for :name, because LabelOrderField already responds to that method. Use `value_method:` to override the method name or `value_method: false` to disable Enum value method generation.
Failed to define value method for :name, because ProjectOrderField already responds to that method. Use `value_method:` to override the method name or `value_method: false` to disable Enum value method generation.
Failed to define value method for :name, because ProjectV2FieldOrderField already responds to that method. Use `value_method:` to override the method name or `value_method: false` to disable Enum value method generation.
[...]

PR #5230, opened 2025-02-11
Check the new class instead of its superclass to see if it responds to the value method

phedinkus:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/5223


rmosolgo:
I pushed a test based on the Gitlab example. Thanks for this fix!

Issue #5231, opened 2025-02-11
Visibility profiles should always evaluate based on profile context


gmac:
**Describe the bug**

I'm playing around with the new visibility features and really like the direction this is going. Nice work. The DX is pretty unintuitive though because of all the nuanced interactions. It could be a lot simpler and more reliable, IMO. Consider:

```ruby
class TestSchema < GraphQL::Schema
  use GraphQL::Schema::Visibility, profiles: {
    public: { public: true },
    internal: { internal: true },
  }
end

class PublicType < GraphQL::Schema::Object
  def self.visible?(context)
    super && (context[:internal] || context[:public])
  end
end

class InternalType < GraphQL::Schema::Object
  def self.visible?(context)
    super && context[:internal]
  end
end

# This works...
TestSchema.execute(
  context: { internal: true },
  visibility_profile: :internal,
)

# This doesn't work...
TestSchema.execute(
  context: {},
  visibility_profile: :internal,
)

# This does the wrong thing...
TestSchema.execute(
  context: { public: true },
  visibility_profile: :internal,
)
```

This gets even more nuanced with the `preload` attribute in play:

```ruby
class TestSchema < GraphQL::Schema
  use GraphQL::Schema::Visibility, preload: true, profiles: {
    public: { public: true },
    internal: { internal: true },
  }
end

# This works...
TestSchema.execute(
  context: { internal: true },
  visibility_profile: :internal,
)

# This works...
TestSchema.execute(
  context: {},
  visibility_profile: :internal,
)

# This works, but is counter-intuitive as to why...
TestSchema.execute(
  context: { public: true },
  visibility_profile: :internal,
)
```

**Versions**

`graphql` version: 2.4

**Expected behavior**

What I'd expect would be for visibility to _always_ favor evaluating based on the profile definition. The profile settings should always be provided to `visible?` as a separate argument or through context, then documentation should indicate that we should always check static profile settings before other aspects of request context, ie:

```ruby
# Profile definitions...
class TestSchema < GraphQL::Schema
  use GraphQL::Schema::Visibility, profiles: {
    public: { public: true },
    internal: { internal: true },
  }
end

# Visibility checks...
class Widget < GraphQL::Schema::Object
  def self.visible?(context)
    super && context.visibility_profile_settings[:internal]
  end
end

# This works consistently because we evaluate based on profile settings...
TestSchema.execute(
  context: {},
  visibility_profile: :internal,
)
```

**Actual behavior**

[...]

PR #5235, opened 2025-02-12
Improve Visibility profile contexts

rmosolgo:
As discussed in #5231: 

- Always use `context[:visibility_profile]`, route the query argument to that context key 
- When that key is present, used the configured context for `.visible?` calls (not the query context). The full query context is still used for dynamic queries (without `context[:visibility_profile]`) 
- For good measure, freeze passed-in profile hashes.


lib/graphql/schema/visibility.rb
```diff
@@ -13,6 +13,10 @@ class Visibility
       # @param preload [Boolean] if `true`, load the default schema profile and all named profiles immediately (defaults to `true` for `Rails.env.production?`)
       # @param migration_errors [Boolean] if `true`, raise an error when `Visibility` and `Warden` return different results
       def self.use(schema, dynamic: false, profiles: EmptyObjects::EMPTY_HASH, preload: (defined?(Rails) ? Rails.env.production? : nil), migration_errors: false)
+        profiles&.each { |name, ctx|
+          ctx[:visibility_profile] = name
+          ctx.freeze
```
gmac, line 18: Good call on freezing.

PR #5237, opened 2025-02-13
Add note about AnyCable

d4rky-pl:
This will hopefully save some other poor soul's two days of debugging 🥲


d4rky-pl:
TIL https://github.com/anycable/graphql-anycable exists and now I feel stupid. Apologies for the noise, closing the PR

PR #5247, opened 2025-02-20
Add Redis backend for Perfetto traces

rmosolgo:
Extracted from #5244 -- I'm trying to figure out where the failing snapshot test came from...


rmosolgo:
Ok, I found the error. `Fiber[:graphql_flow_stack]` was being left as an empty array from one run to the next. Then later runs would use that top-level array instead of a per-fiber array created later on. I think there is actually _more_ jankiness here, but unsetting that before each run will contain the jankiness to one query. 

I think better behavior for the flow stack would be: 

- When pushing a flow packet, push it to _this fiber's_ stack (not a parent fiber)
- When reading a flow packet, read it from this fiber's stack, or a parent fiber (the current fiber may have been created to finish work started on another fiber)

But I'm not going to implement that now.

PR #5258, opened 2025-02-27
Set default timeout for GraphQL schema validation

vithushar:
## Problem

Currently, `graphql-ruby` does not enforce a timeout for schema validation by default. When `validate_timeout` isn't explicitly set, maliciously crafted queries can consume substantial server resources and potentially cause denial of service (DoS) conditions.

Through Shopify's bug bounty program, we've received multiple reports demonstrating that a single host can leverage this behavior to impact an entire API.

## Solution

Add a reasonable default timeout (3 seconds) for schema validation that:
- protects all downstream consumers
- maintains backward compatibility and 
- preserves the ability to customize or disable the timeout

### Why 3 seconds?

- **Data-Driven Decision**: We analyzed performance metrics across many GraphQL services at Shopify, examining the 95th percentile of GraphQL execution time (not including data retrieval). Most queries execute in under 500ms, with even our most complex queries rarely exceeding 700ms.
- **Configurable Override**: Services with exceptional needs can still override the default value

### DIFF
--- a/guides/queries/timeout.md
+++ b/guides/queries/timeout.md
@@ -67,14 +67,20 @@ end
 
 Queries can originate from a user, and may be crafted in a manner to take a long time to validate against the schema.
 
-It is possible to limit how many seconds the static validation rules and analysers are allowed to run before returning a validation timeout error. The default is no timeout.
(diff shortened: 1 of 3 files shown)


vithushar:
@rmosolgo Would love your thoughts on adding this secure default. Our rationale for why we want this implemented upstream is in the PR description.

vithushar:
@rmosolgo Conflict resolved! :)

PR #5267, opened 2025-03-08
Raise a ParseError on dangling close curly brace

rmosolgo:
Somehow between the lexer and parser, this last brace is being ignored. Something about `lexer.eos?` returning true even though there's an unhandled token still there.


isaacbowen:
incredible. :) thank you! ❤️‍🔥

PR #5268, opened 2025-03-08
Lookup the right field in FieldsWillMerge

### LINKED ISSUE #5264: No validation error anymore on conflicting aliases?
**Describe the bug**
👋 
According to the spec of `FieldsWillMerge` ([link](https://graphql-ruby.org/api-doc/2.4.9/GraphQL/StaticValidation/FieldsWillMerge.html#:~:text=NO_ARGS%20%3D)) aliases should be unique, however, on the version 2.4.11 the query below is not triggering a validation error anymore:

```
query {
    repository(name: "some_name", owner: "some_owner") {
      info: name
      info: id
  }
}
```

**Versions**

`graphql-ruby` version: 2.4.11
Haven't tried previous versions tho.

**Steps to reproduce**

running query with conflicting aliases on 2.4.11
```
query {
   repository(name: "some_name", owner: "some_owner") {
      info: name
      info: id
  }
}
```

**Expected behavior**

Validation error with the message `"Field 'info' has a field conflict: name or id?"`

**Actual behavior**

The query is successful, and `info` contains `name` value.

rmosolgo:
Fixes #5264

### DIFF
--- a/lib/graphql/static_validation/rules/fields_will_merge.rb
+++ b/lib/graphql/static_validation/rules/fields_will_merge.rb
@@ -345,7 +345,7 @@ module GraphQL
             fields << Field.new(node, definition, owner_type, parents)
           when GraphQL::Language::Nodes::InlineFragment
             fragment_type = node.type ? @types.type(node.type.name) : owner_type
-            find_fields_and_fragments(node.selections, parents: [*parents, fragment_type], owner_type: owner_type, fields: fields, fragment_spreads: fragment_spreads) if fragment_type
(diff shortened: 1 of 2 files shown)


AlenaSviridenko:
@rmosolgo I was wondering how much of the work would it be to backport this to the next patch version of 2.3? 
We are looking into upgrading gem to 2.3.21, but it contains this bug, which effect is hard to predict. 
Thank you!

rmosolgo:
It looks like it cherry-picks cleanly: #5275

PR #5269, opened 2025-03-08
Try removing extra orphan-types setup

rmosolgo:
Try the suggestion from https://github.com/rmosolgo/graphql-ruby/pull/5074#discussion_r1985252475


rmosolgo:
It seems like it does actually need this but I'm not sure why. I'll have to investigate it another time.

PR #5270, opened 2025-03-10
Rewrite NotificationsTrace

rmosolgo:
Trying to update this to handle fiber starts/stops better. 

This PR covers ActiveSupport::Notifications, Dry::Notifications, New Relic, Data Dog, and Sentry. I'll address the others in a later PR.

- [x] Do I really need `begin_`/`finish_` methods for parts of execution that don't call out to app code? (no) 
- [x] Update NotificationsTrace
- [x] Add a test for AS::N trace 
- [x] Migrate other traces to this system - Started with New Relic and Sentry.
- [x] What about AS::N payloads? (And Dry::Monitor) I want avoid creating that Hash by default, but those will need it for backwards compat.
- [x] Make sure platform-specific traces still work when there are multiple


zvkemp:
Were the `execute_query`, `execute_query_lazy`, and `analyze_query` instrumentations intentionally omitted from this refactor? We've noticed those are no longer published to our ActiveSupport::Notifications subscriber.

PR #5276, opened 2025-03-13
Support converting Proc return type expressions to type names

cocoahero:


rmosolgo:
👋  I'd be interested to learn the rationale for this, but there are already a couple of lazy-ish ways to provide a type: 

- String reference to a Ruby constant (eg `"Types::User"` => `Types::User`)
- Configuring the type in a `do ... end` block, eg `field :user { type(Types::User) }`. Executing that block will be defered as long as possible. 

If there's something that those approaches don't address, please let me know. I definitely want to find something that works ... but I also want to avoid adding procs as new API surface.

PR #5278, opened 2025-03-13
Immediately return empty array in multiplex if queries list is empty

ilyasgaraev:
Version `2.4.10` changed the NewRelicTrace class to support fiber stops/starts.

Now `GraphQL::Execution::Interpreter` calls `begin_execute_multiplex`, which relies on the [first multiplex query](https://github.com/rmosolgo/graphql-ruby/blob/v2.4.10/lib/graphql/tracing/new_relic_trace.rb#L68) in the list:

```ruby
def begin_execute_multiplex(multiplex)
  query = multiplex.queries.first
  set_this_txn_name = query.context[:set_new_relic_transaction_name]
  ...
end
```

Theoretically, it is possible to call multiplex with an empty array (I know, this is an edge case, but it is possible). In this case, we don't have any query to run and can return an empty result.

If the Schema uses the NewRelic tracer, `GraphQL::Tracing::NewRelicTrace#begin_execute_multiplex` fails with error because `multiplex.queries.first` returns `nil`.

Easy steps to reproduce (fails with NewRelicTrace and works correctly without NewRelicTrace):

```ruby
  field :name, GraphQL::Types::String, null: false
end

  trace_with GraphQL::Tracing::NewRelicTrace

  query QueryType
end

Schema.multiplex([])
```

We can fix all trace classes to handle this case correctly, but it seems like we don't even need to analyze/execute an empty list of queries.

### DIFF
--- a/lib/graphql/execution/interpreter.rb
+++ b/lib/graphql/execution/interpreter.rb
@@ -33,6 +33,7 @@ module GraphQL
             end
           end
 
+          return GraphQL::EmptyObjects::EMPTY_ARRAY if queries.empty?
 
           multiplex = Execution::Multiplex.new(schema: schema, queries: queries, context: context, max_complexity: max_complexity)
           Fiber[:__graphql_current_multiplex] = multiplex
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for this suggestion! I agree that an early return seems like a good idea. 

I'm trying to make sure there's nothing important here that _shouldn't_ be skipped, even for an empty multiplex, but I can't think of anything.

PR #5281, opened 2025-03-15
Support type extensions in `Schema.from_definition`

gmac:
Type extensions are a neat prospect for managing schema supersets. It'd be handy if `Schema.from_definition` supported them. They're already fully supported in the language, so this adds them into the schema builder.

```graphql
# -- base schema
type T { a:Int }
type Query { t:T }

# -- concat extensions
extend type T { b:Int }
```

Using `Schema.from_definition` on the above gives type `T` with two fields on it.

### DIFF
--- a/lib/graphql/schema/build_from_definition.rb
+++ b/lib/graphql/schema/build_from_definition.rb
@@ -86,7 +86,13 @@ module GraphQL
             case definition
             when GraphQL::Language::Nodes::SchemaDefinition, GraphQL::Language::Nodes::DirectiveDefinition
               nil # already handled
-            when GraphQL::Language::Nodes::SchemaExtension
+            when GraphQL::Language::Nodes::SchemaExtension,
+              GraphQL::Language::Nodes::ScalarTypeExtension,
+              GraphQL::Language::Nodes::ObjectTypeExtension,
+              GraphQL::Language::Nodes::InterfaceTypeExtension,
+              GraphQL::Language::Nodes::UnionTypeExtension,
+              GraphQL::Language::Nodes::EnumTypeExtension,
+              GraphQL::Language::Nodes::InputObjectTypeExtension
               schema_extensions ||= []
               schema_extensions << definition
             else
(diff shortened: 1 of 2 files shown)


rmosolgo:
I think the test failure here is something real... those builds use `GraphQL::Schema::Visibility` instead of `Warden`. You can enable the same config locally by running tests with the env variable: 

```
GRAPHQL_REJECT_NUMBERS_FOLLOWED_BY_NAMES=1
```

(I'm using that as a general "opt in to future features" flag: https://github.com/rmosolgo/graphql-ruby/blob/3664b848455ef73f3c79fcb389f5832edc997840/spec/spec_helper.rb#L35-L41) 

I think the problem is that, although the new union members are added to the Union _itself_, the schema's type cache doesn't know about them yet. Visibility itself also maintains a lookup like this: 

https://github.com/rmosolgo/graphql-ruby/blob/3664b848455ef73f3c79fcb389f5832edc997840/lib/graphql/schema/visibility.rb#L290-L296

which is later used in places like this: 

https://github.com/rmosolgo/graphql-ruby/blob/3664b848455ef73f3c79fcb389f5832edc997840/lib/graphql/schema/visibility/profile.rb#L314

But now that I write all that out, I'm still not clear on why the existing code doesn't work. It _feels_ like some kind of sequence issue, where the the new `possible_types` relationship between the union and object type isn't registered by the schema. One possibility might be to move the `schema_extensions.each { ... }` call _up_ so that it's before the `Class.new(schema_superclass) { ... }` call. But I'm just guessing ...

gmac:
I was also thinking that extensions should probably be built first so that everything was in place before `add_type_and_traverse` runs. Still test failures though? Not sure if they're related to this. Passes with `GRAPHQL_REJECT_NUMBERS_FOLLOWED_BY_NAMES=1` now.

rmosolgo:
Looks to me like one of those failures is a _normal_ flaky test (😩) and the other one is because I don't have that CI job set up right. I'll update that in another PR -- apparently it should be `on: pull_request_target` to work with pulls from forks.

Thanks for this improvement!

PR #5286, opened 2025-03-18
Fix GraphQL backtrace handling for inline fragments

yn-misaki:
When GraphQL::Backtrace is enabled and an exception occurs during execution of a query with inline fragments, the backtrace system itself throws an error:

```
NoMethodError: undefined method 'alias' for an instance of GraphQL::Language::Nodes::InlineFragment
    lib/graphql/backtrace/table.rb:83:in 'block in GraphQL::Backtrace::Table#rows'
    lib/graphql/backtrace/table.rb:83:in 'Array#each'
    lib/graphql/backtrace/table.rb:83:in 'Enumerable#find'
    lib/graphql/backtrace/table.rb:83:in 'GraphQL::Backtrace::Table#rows'
    lib/graphql/backtrace/table.rb:29:in 'GraphQL::Backtrace::Table#to_backtrace'
    lib/graphql/backtrace.rb:38:in 'GraphQL::Backtrace#to_a'
    lib/graphql/backtrace/traced_error.rb:31:in 'GraphQL::Backtrace::TracedError#initialize'
    lib/graphql/schema.rb:1131:in 'Class#new'
    lib/graphql/schema.rb:1131:in 'GraphQL::Schema.handle_or_reraise'
    lib/graphql/query.rb:410:in 'GraphQL::Query#handle_or_reraise'
```

This happens because the backtrace table code attempts to access the 'alias' method on InlineFragment nodes, which don't support this property.
Modified the backtrace system to properly handle node types that don't have 'alias' or 'name' methods, ensuring that the backtrace can still be generated correctly for queries using inline fragments.

If you have suggestions for alternative implementation approaches, please share them.

### DIFF
--- a/lib/graphql/backtrace/table.rb
+++ b/lib/graphql/backtrace/table.rb
@@ -78,28 +78,51 @@ module GraphQL
             end
           end
 
-
           object = result.graphql_application_value.object.inspect
-          ast_node = result.graphql_selections.find { |s| s.alias == last_part || s.name == last_part }
-          field_defn = query.get_field(result.graphql_result_type, ast_node.name)
-          args = query.arguments_for(ast_node, field_defn).to_h
-          field_path = field_defn.path
-          if ast_node.alias
-            field_path += " as #{ast_node.alias}"
+          ast_node = nil
+          result.graphql_selections.each do |s|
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for this fix!

PR #5287, opened 2025-03-18
Redo pronto action to work on forks

rmosolgo:
Oops, it didn't work on PRs from forks because `on: pull_request` doesn't have permission to do that.


rmosolgo:
... I guess now that this is `pull_request_target`, it has to be on master in order to run...

PR #5290, opened 2025-03-21
Print warning when custom validators are detected

vithushar:
## Problem

Custom validators with I/O operations may fail unexpectedly due to GraphQL's default validate_timeout setting. Long-running I/O operations may not be terminated midway, resulting in unpredictable behavior. 

See these PRs for more context:
Manual timeout implementation: https://github.com/rmosolgo/graphql-ruby/pull/5263
Validate_timeout default value implementation: https://github.com/rmosolgo/graphql-ruby/pull/5258

We should make the risk of adding I/O bound custom validators more "in your face" in the gem for consumers.

## solution
Print a warning when a custom validator is registered.

Screenshot 2025-03-20 at 9 43 22 PM


vithushar:
> What if we made `validate_timeout` I/O-safe instead?

I'm very much on board with that. Let's do it!

PR #5294, opened 2025-03-24
[1.13] Fix RelationConnections when relation already has limit

rmosolgo:
Backporting #4104 to 1.13.x


rmosolgo:
CI has compatibility issues, but the failures on `bundle exec rake test` are all Ruby version related (eg, different backtrace format, different Hash#inspect output). This modified test file passes locally.

PR #5301, opened 2025-03-27
Improve nullable array element type error reporting

fisxoj:
Hello and thanks for the gem!

Currently for a field like

```ruby
field :dog_names, [String], null: false
def dog_names
  ["wedge", nil]
end
```

You would get an error message like `Cannot return null for non-nullable field Query.dogNames`.  The phrasing of that message leads a person to think that the problem is that the return value of the `dog_names` method is `nil` when the error is actually referring to the value at index 1 of the returned array.

The purpose of this PR is to change the messaging so that, if the error is caused by a list value with an invalid element, it says that explicitly.

I had hoped to include the path to the element so the user would know the index of the failing element, but I couldn't figure out how to best do that.  It seems like the `idx` int the `continue_field` method in `runtime.rb` is the main place the index of the value is known but is lost before the value gets inspected.

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -471,9 +471,10 @@ module GraphQL
             if is_non_null
               set_result(selection_result, result_name, nil, false, is_non_null) do
                 # When this comes from a list item, use the parent object:
-                parent_type = selection_result.is_a?(GraphQLResultArray) ? selection_result.graphql_parent.graphql_result_type : selection_result.graphql_result_type
+                is_from_array = selection_result.is_a?(GraphQLResultArray)
+                parent_type = is_from_array ? selection_result.graphql_parent.graphql_result_type : selection_result.graphql_result_type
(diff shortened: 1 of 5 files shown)


rmosolgo:
Hey, thanks so much for sharing this improvement. It will be a big help. 

> I had hoped to include the path to the element 

I think you could combine `result_name` (which I expect to be the index in the list) with [`current_path`](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/execution/interpreter/runtime.rb#L456) (which I expect to contain the _rest_ of the path) to create full path to the element. Are you interested in investigating that further?

fisxoj:
Thanks for the suggestion! I'll try to spend some more time on it on monday and implement that. Let me know if you have any specific phrasing you'd prefer.

fisxoj:
@rmosolgo I've not managed to pull the time together to do more work on this.  If it's useful as is, it might be best to merge it and maybe some day I'll be back for more.

rmosolgo:
Thanks for this improvement!

PR #5306, opened 2025-03-29
Improve Scalar result coercion spec compliance

swalkinshaw:
The spec says that scalars should return query/field errors when input or result coercion values don't conform to the spec.

Int example: https://spec.graphql.org/draft/#sel-GAHXRFHCAACGS6rU

This updates the logic to return `GraphQL::CoercionError` (execution errors) instead of raised `RuntimeError`s which would not result in a client error by default.

The logic and error messages are were taken from the [graphql-js scalar implementation](https://github.com/graphql/graphql-js/blob/0a848cc4c781fb79b2e55eb8d011675dd612524f/src/type/scalars.ts). 
Note that this is likely a breaking change for schema developers. For clients, it's an improvement since they might get some sort of "internal server error" otherwise.

@rmosolgo I didn't update all the necessary specs yet because I figured this would need some further discussion. It's a fairly large behaviour change from the schema's perspective but ultimately we should try to support this to be spec compliant.

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -93,6 +93,7 @@ This is probably a bug in GraphQL-Ruby, please report this error on GitHub: http
   autoload :AnalysisError, "graphql/analysis_error"
   autoload :CoercionError, "graphql/coercion_error"
   autoload :InvalidNameError, "graphql/invalid_name_error"
+  autoload :ScalarCoercionError, "graphql/scalar_coercion_error"
   autoload :IntegerDecodingError, "graphql/integer_decoding_error"
   autoload :IntegerEncodingError, "graphql/integer_encoding_error"
   autoload :StringEncodingError, "graphql/string_encoding_error"
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -675,6 +675,8 @@ module GraphQL
           when "SCALAR", "ENUM"
             r = begin
               current_type.coerce_result(value, context)
+            rescue GraphQL::ExecutionError => ex_err
+              return continue_value(ex_err, field, is_non_null, ast_node, result_name, selection_result)
             rescue StandardError => err
               query.handle_or_reraise(err)
             end
--- a/spec/graphql/types/int_spec.rb
+++ b/spec/graphql/types/int_spec.rb
@@ -26,21 +26,77 @@ describe GraphQL::Types::Int do
         assert_equal(-(2**31), GraphQL::Types::Int.coerce_result(-(2**31), context))
       end
 
-      it "replaces values, if configured to do so" do
-        assert_equal Dummy::Schema::MAGIC_INT_COERCE_VALUE, GraphQL::Types::Int.coerce_result(99**99, context)
-      end
-
       it "raises on values out of bounds" do
         err_ctx = GraphQL::Query.new(Dummy::Schema, "{ __typename }").context
         assert_raises(GraphQL::IntegerEncodingError) { GraphQL::Types::Int.coerce_result(2**31, err_ctx) }
         err = assert_raises(GraphQL::IntegerEncodingError) { GraphQL::Types::Int.coerce_result(-(2**31 + 1), err_ctx) }
-        assert_equal "Integer out of bounds: -2147483649. Consider using ID or GraphQL::Types::BigInt instead.", err.message
+        assert_equal "Int cannot represent non 32-bit signed integer value: -2147483649", err.message
 
         err = assert_raises GraphQL::IntegerEncodingError do
           Dummy::Schema.execute("{ hugeInteger }")
         end
-        expected_err = "Integer out of bounds: 2147483648 @ hugeInteger (Query.hugeInteger). Consider using ID or GraphQL::Types::BigInt instead."
-        assert_equal expected_err, err.message
+        assert_equal "Int cannot represent non 32-bit signed integer value: 2147483648", err.message
+      end
+
+      describe "with Schema.spec_compliant_scalar_coercion_errors" do
+        class IntScalarSchema < GraphQL::Schema
+          class Query < GraphQL::Schema::Object
+            field :int, GraphQL::Types::Int do
+              argument :value, GraphQL::Types::Int
+            end
+
+            field :bad_int, GraphQL::Types::Int
+
+            def int(value:)
+              value
+            end
+
+            def bad_int
+              2**31 # Out of range
+            end
+          end
+
+          query(Query)
+        end
+
+        class IntSpecCompliantErrors < IntScalarSchema
+          spec_compliant_scalar_coercion_errors true
+        end
+
+        class IntNonSpecComplaintErrors < IntScalarSchema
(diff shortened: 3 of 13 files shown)


rmosolgo:
I'm definitely interested in complying with the spec by default _eventually_, but I think that before changing the defaults, we should release a version with warnings that the default behavior is going to change with a link to the docs about how to either override the defaults (to retain current behavior without the warning) or opt into the future default behavior. 

The other thing is to make sure these still show up in peoples' bug trackers. The intention behind raising errors was to make sure developers become aware that their data and schema don't match -- something needs to be fixed on the backend. What would have to happen to make sure these `CoercionError`s show up to developers?

swalkinshaw:
Yeah that's the hard part because the gem has no abstraction for error "reporting". I think the best we can do is a hybrid approach:

1. call the `type_error` hook with an instantiated exception
2. add an empty branch to the default `type_error` implementation for a no-op
3. return the execution error

Then at least this gives people a place to implement the behaviour they want but it's still opt-in and a "breaking change".

Unless I'm missing something, there's no way to both _report_ an exception and return an execution error currently. This gem would have to check for various bug tracker integrations.

With the Railtie we could default to using `Rails.error.report` but that still assumes that the application is actually using that newer interface.

So it's a catch 22 where I don't think it's possible to get the ideal behaviour. We need to raise an exception... but also not raise one.

rmosolgo:
What if we preserve the current default behavior, but also:

- Add a new configuration, eg `MySchema.raise_scalar_type_errors = nil`
- In the default `type_error` behavior:
  - if that configuration wasn't set to `true`, warn that the default behavior is changing in a future version and that the new setting should be assigned to either `true` or else custom handling should be put in `def self.type_error` to preserve the current behavior in future versions
  - if the configuration is set to `true`, do the new behavior (without any warning)
  
Then, in a future version, we change the default to return errors and deprecate the now-useless config. That would move us progressively toward spec compliance while warning people about what's going on. What do you think? 

> abstraction for error "reporting"

That's cool, I didn't know Rails had `Rails.errors.report(err)`. I wonder if GraphQL-Ruby could hand errors off to that somehow 🤔

swalkinshaw:
Okay made a lot of updates and I'll update the PR description after if this is looking mergeable.

* Added the `Schema.spec_compliant_scalar_coercion_errors` option. I didn't love the `raise_scalar_type_errors` name since "raise" is confusing as exceptions can technically raised in both cases.
* I've focused this on **result coercion** only. I think input coercion is mostly spec compliant though the error messages are a bit inconsistent now. That can be looked at after.
* I replaced any existing scalar specific encoding exception classes with a single new `LegacyScalarCoercionError` to make the `type_error` implementation easier. If you don't like that, it should be possible to restore the old ones. Technically this is a tiny breaking change since it will change the errors reported in apps but this also provides a signal that it's "legacy".
* I improved the error response of `coerce_result` which was previously missing `path`, `locations`, and `data`.

lib/graphql.rb
```diff
@@ -94,8 +94,7 @@ class << self
   autoload :CoercionError, "graphql/coercion_error"
   autoload :InvalidNameError, "graphql/invalid_name_error"
   autoload :IntegerDecodingError, "graphql/integer_decoding_error"
-  autoload :IntegerEncodingError, "graphql/integer_encoding_error"
-  autoload :StringEncodingError, "graphql/string_encoding_error"
+  autoload :LegacyScalarCoercionError, "graphql/legacy_scalar_coercion_error"
   autoload :DateEncodingError, "graphql/date_encoding_error"
   autoload :DurationEncodingError, "graphql/duration_encoding_error"
```
swalkinshaw, line 99: Side note, I believe both of these should be `Decoding` since they are only used in `coerce_input` methods. I can fix them separately.

lib/graphql/execution/interpreter/runtime.rb
```diff
@@ -675,6 +675,8 @@ def continue_field(value, owner_type, field, current_type, ast_node, next_select
           when "SCALAR", "ENUM"
             r = begin
               current_type.coerce_result(value, context)
+            rescue GraphQL::ExecutionError => ex_err
+              return continue_value(ex_err, field, is_non_null, ast_node, result_name, selection_result)
```
swalkinshaw, line 679: Any possible issues here? No tests break but this will change error messages, though it will improve and make them more spec compliant.

spec/graphql/types/string_spec.rb
```diff
@@ -145,4 +136,65 @@ def get_string(string:)
...
+          "errors" => [
+            {
+              "message" =>  "String cannot represent value: \"\\x00\\x00\\x00foo\\xAD\\xAD\\xAD\"",
+              "locations" => [{ "line" => 1, "column" => 3 }],
+              "path" => ["badString"],
+            },
+          ],
+          "data" => {
+            "badString" => nil,
+          }
```
swalkinshaw, line 184: This shows the improved error messages for result coersion errors.

swalkinshaw:
In hindsight, it's better to keep the type specific encoding exceptions for now but have them inherit from a new common base so I've made this even less breaking.

spec/graphql/types/int_spec.rb
```diff
@@ -26,21 +26,77 @@
         assert_equal(-(2**31), GraphQL::Types::Int.coerce_result(-(2**31), context))
       end
 
-      it "replaces values, if configured to do so" do
-        assert_equal Dummy::Schema::MAGIC_INT_COERCE_VALUE, GraphQL::Types::Int.coerce_result(99**99, context)
-      end
```
rmosolgo, line 31: Does this have to be removed? It seems like the override in the example schema _could_ be retained, but a method call would have to be changed.... or is this behavior not possible anymore? 

(Maybe the old override could be retained as-is if an alias accessor was added to the new IntegerEncodingError)

rmosolgo:
Thanks for all your work on this, I'm definitely looking forward to getting this stuff on track. 

The last question I have is about developer experience. Currently, the default behavior is to return the error to the client but not notify the developer in any way that something has gone wrong. (Maybe some tooling is checking the `"errors"` key in the response?) 

I wonder if a good enough approach would be to modify the [generated default schema](https://github.com/rmosolgo/graphql-ruby/blob/f2bd90209315172684bdfac62453435893dca5bb/lib/generators/graphql/templates/schema.erb#L13-L20) to use `Rails.error.report`, and include a nudge that direction in the changelog and/or warning. Does any other solution come to mind for you in that area?

swalkinshaw:
I was considering adding a basic error reporter as part of this as well, or separately. The default would be a no-op (since there's also no logger built in) but then the Railtie would configure it to `Rails.error.report` which would cover most cases.

I'm running this branch against our test suite as well to see what breaks or not.

spec/graphql/types/int_spec.rb
swalkinshaw: No it doesn't have to be I don't think. I'll check

rmosolgo:
I put that test back and added `#integer_value` to make it pass. This change looks good to me -- let me know how things look on your end and I'll merge it when we're all clear

Issue #2929, opened 2020-05-08
Rails multi DB support based on operation


chadwilken:
**Is your feature request related to a problem? Please describe.**
Rails 6 uses the `writing` role for *all* GraphQL requests since it uses the HTTP Verb to select the role.

**Describe the solution you'd like**
On each invocation of `execute` check the operation type and select the correct role.

**Describe alternatives you've considered**
Hooking into the execute call or writing an analyzer to determine which to use.

**Additional context**
I don't know exactly how to check if the user has enabled multi-db support for Rails but I know you can check `ActiveRecord::Base.writing_role` and `reading_role`. It also looks like `ActiveRecord::Base.connection_handlers` returns a hash with *only* the `writing` role if they haven't added a replica.

rmosolgo:
Implementation-wise, I think it could be implemented as a multiplex tracer. My thoughts go there first because: 

- A multiplex _could_ include queries _and_ mutations, and they're run concurrently. So, you'd want to pick `writing_role` if the multiplex contains any non-queries. 
- The Rails API expects a block. A tracer seems like the best GraphQL-Ruby API in that case. 

Here's some pseudocode:

```ruby 
class DatabaseConnectionTracer 
  def trace(event, data) 
    if event == "execute_multiplex" # IIRC this is the top-level event that covers everything else
      multiplex = data[:multiplex]
      # Make sure they're all queries (no mutations or subscriptions)
      role = multiplex.queries.all?(&:query?) ? :reading : :writing 
      # Continue execution with the proper DB role
      ActiveRecord::Base.connected_to(role: role) do 
        yield 
      end 
    else 
      yield
    end 
  end 
end 

# ... 
class MySchema 
  tracer DatabaseConnectionTracer.new 
end 
```

(Under the hood, GraphQL-Ruby treats single queries as a "multiplex-of-one".) 

I don't have a great situation to test this, but if you want to give it a try, let me know how it goes!

chadwilken:
That seems simple enough. I can test it on our staging environment pretty easily. I'll post back results and if I have some time I can open a PR.

chadwilken:
That worked perfectly. Do you think it is something that should be standardized into a feature or just something in the docs that they can look at if they hit the specific use-case?

rmosolgo:
I'm game to include it in GraphQL-Ruby, but I think it should be opt-in for now. I'll take a look.

chadwilken:
[...]

PR #5313, opened 2025-04-02
Sentry trace: handle nil current_span

rmosolgo:
Fixes https://github.com/rmosolgo/graphql-ruby/issues/2929#issuecomment-2771991624

Apparently Sentry's current_span is sometimes `nil`, the library itself is full of `nil` handling, for example:

https://github.com/getsentry/sentry-ruby/blob/a884caee95ddf3c29cec4761ab423ed56e0b16b5/sentry-ruby/lib/sentry/hub.rb#L137-L138

The previous code in GraphQL-Ruby had handling for `span` being nil, but my new code didn't. Now it does.

### DIFF
--- a/lib/graphql/tracing/sentry_trace.rb
+++ b/lib/graphql/tracing/sentry_trace.rb
@@ -64,8 +64,7 @@ module GraphQL
 
         class Event < MonitorTrace::Monitor::Event
           def start
-            if Sentry.initialized?
-              @span = Sentry.get_current_scope.get_span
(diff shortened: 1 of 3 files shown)


rmosolgo:
@kaito0228 could you try `master` in your project and see if it fixes your problem? 

```ruby
gem "graphql", github: "rmosolgo/graphql-ruby", ref: "master"
```

Let me know what you find!

zephleggett:
thanks! I hit this after upgrading to 2.5.0, my specs are passing with 2.5.1.

PR #5314, opened 2025-04-02
Move Lazy resolution into Dataloader

rmosolgo:
This is a moonshot but I want to see if the runtime code can be simplified by rebuilding `lazy_resolve` on top of `GraphQL::Dataloader`. Some hard requirements: 

- [ ] No (or minimal...) performance hit for adopting Dataloader (not in speed or memory usage)
- [ ] Behavioral parity  -- same batching performance and behavior
- [ ] Swappable per-query, so that you can try the new way or the old way while the app is running. This is essential for adoption: you have to be able to fall back to the old way if something goes wrong. 

------------------------

My todos: 

- [ ] Make sure all existing `legacy_sync_lazy(true)` can be replaced with Dataloader
- [ ] Run builds with both legacy and Dataloader implementations
...
- [ ] Revisit API and write docs

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -6,6 +6,7 @@ require "graphql/dataloader/request_all"
 require "graphql/dataloader/source"
 require "graphql/dataloader/active_record_association_source"
 require "graphql/dataloader/active_record_source"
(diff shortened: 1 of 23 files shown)


rmosolgo:
I ended up going a different direction with this, see #5422

Issue #5316, opened 2025-04-02
GraphQL::Tracing::DatadogTrace raises a NoMethodError in v2.5.0


jocmp:
**Describe the bug**

The `GraphQL::Tracing::DatadogTrace` module raises a NoMethodError in v2.5.0.

**Versions**

`graphql` version: 2.5.0
`rails`: 7.2.2.1
`ddtrace`: 1.18.0

**Steps to reproduce**

Steps to reproduce the behavior

1. Define `trace_with GraphQL::Tracing::DatadogTrace` in the `GraphQL::Schema` subclass
2. Make a GraphQL request
3. NoMethodError! `platform_source_class_key` is not defined

**Actual behavior**

What specifically went wrong?

Place full backtrace here (if a Ruby exception is involved):

<details>
<summary>Click to view exception backtrace</summary>

```
(NoMethodError)

          @platform_source_class_key_cache = Hash.new { |h, source_cls| h[source_cls] = platform_source_class_key(source_cls) }.compare_by_identity
                                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^
Did you mean?  platform_source_key

<obfuscated>/gems/graphql-2.5.0/lib/graphql/tracing/monitor_trace.rb:17:in `block in initialize'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/tracing/monitor_trace.rb:57:in `name_for'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/tracing/data_dog_trace.rb:60:in `start'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/tracing/monitor_trace.rb:26:in `start_event'
<obfuscated>/gems/graphql-2.5.1/lib/graphql/tracing/monitor_trace.rb:271:in `begin_datadog_event'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/tracing/monitor_trace.rb:256:in `begin_dataloader_source'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/dataloader/async_dataloader.rb:90:in `block (2 levels) in spawn_source_task'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/dataloader/async_dataloader.rb:89:in `each'
<obfuscated>/gems/graphql-2.5.0/lib/graphql/dataloader/async_dataloader.rb:89:in `block in spawn_source_task'
<obfuscated>/gems/async-2.14.1/lib/async/task.rb:190:in `block in run'
[...]

PR #5318, opened 2025-04-02
Fix monitor trace suffix method

jocmp:
Swaps in the "platform_source_class_key" method instead of the previously unused "platform_source_key" to prevent a NoMethodError on initialization.

## Demo

With the fix, suffix tracing works as expected

dd-demo

Relates to

- https://github.com/rmosolgo/graphql-ruby/issues/5316

### DIFF
--- a/lib/graphql/tracing/monitor_trace.rb
+++ b/lib/graphql/tracing/monitor_trace.rb
@@ -103,7 +103,7 @@ module GraphQL
           end
 
           def platform_source_class_key(source_class)
-            "#{source_class.name.gsub("::", "_").underscore}.fetch.graphql"
+            "#{source_class.name.gsub("::", "_")}.fetch.graphql"
           end
         end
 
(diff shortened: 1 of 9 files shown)


lib/graphql/tracing/monitor_trace.rb
```diff
@@ -104,8 +102,8 @@ def platform_resolve_type_key(type)
             "#{type.graphql_name}.resolve_type.graphql"
           end
 
-          def platform_source_key(source_class)
-            "#{source_class.name.gsub("::", "_").name.underscore}.fetch.graphql"
+          def platform_source_class_key(source_class)
+            "#{source_class.name.gsub("::", "_").underscore}.fetch.graphql"
```
jocmp, line 106: This additional call to `.name` on a string caused a NoMethodError in my testing.

rmosolgo:
Hey, thanks so much for this! I added a test and found that the `.underscore` needed to be removed too, since that's a Rails method. I have no idea what I was thinking there ... or how I missed test coverage for this method 🤦

rmosolgo:
Just shipped 2.5.1 with this patch. Thanks again, and please let me know if that version gives you any more trouble!

PR #5320, opened 2025-04-07
Add Resolver.deprecation_reason

### LINKED ISSUE #5300: Field deprecation via resolver class
### Feature request

A field resolver class can contain a lot of configuration for the field definition's behaviour. It would be convenient if it was also to support setting a `deprecation_reason`. 

## Possible solution?

Perhaps the field's initialization can look at `resolver_class && resolver_class.deprecation_reason` as part of storing `deprecation_reason`?

https://github.com/rmosolgo/graphql-ruby/blob/bf5aa5957c2892f1ad72b12d116a92324605eadc/lib/graphql/schema/field.rb#L263

## Additional considerations

We should have at most one `@deprecated` set on the field. The field's `deprecation_reason` should likely be treated as an override to the resolver class's value if set. So it may not be enough to allow the resolver class to add to its own `directives` array.

rmosolgo:
Fixes #5300 

cc @ravangen

### DIFF
--- a/lib/graphql/schema/directive/flagged.rb
+++ b/lib/graphql/schema/directive/flagged.rb
@@ -37,6 +37,8 @@ module GraphQL
 
         argument :by, [String], "Flags to check for this schema member"
 
+        repeatable(true)
+
         module VisibleByFlag
           def self.included(schema_class)
             schema_class.extend(self)
--- a/lib/graphql/schema/field.rb
+++ b/lib/graphql/schema/field.rb
@@ -41,10 +41,24 @@ module GraphQL
         end
       end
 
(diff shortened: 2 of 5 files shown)


lib/graphql/schema/directive/flagged.rb
```diff
@@ -37,6 +37,8 @@ def initialize(target, **options)
 
         argument :by, [String], "Flags to check for this schema member"
 
+        repeatable(true)
```
rmosolgo, line 40: This preserves compatibility if anyone was actually using this

ravangen:
Thanks ❤️

PR #5322, opened 2025-04-08
Require selections on Unions, with backwards compat

### LINKED ISSUE #4351: `fields_have_appropriate_selections` does not require selections on union types

rmosolgo:
Fixes #4351 

This implementation: 

- Warns if an invalid query is encountered and no compat mode is selected;
- Provides a hook for handing invalid queries (which can permit, deny, or provide a custom message) 
- Provides a config to reject invalid queries 

I think this is a safe default to roll out to anyone already using the gem.

cc @gmac -- finally...

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1670,6 +1670,36 @@ module GraphQL
         end
       end
 
+
+      # If you need to support previous, non-spec behavior which allowed selecting union fields
+      # but *not* selecting any fields on that union, set this to `true` to continue allowing that behavior.
+      #
+      # If this is `true`, then {.legacy_invalid_empty_selections_on_union} will be called with {Query} objects
+      # with that kind of selections. You must implement that method
+      # @param new_value [Boolean]
+      # @return [true, false, nil]
+      def allow_legacy_invalid_empty_selections_on_union(new_value = NOT_CONFIGURED)
+        if NOT_CONFIGURED.equal?(new_value)
+          @allow_legacy_invalid_empty_selections_on_union
+        else
+          @allow_legacy_invalid_empty_selections_on_union = new_value
+        end
+      end
+
+      # This method is called during validation when a previously-allowed, but non-spec
+      # query is encountered where a union field has no child selections on it.
+      #
+      # You should implement this method to log the violation so that you can contact clients
+      # and notify them about changing their queries. Then return a suitable value to
+      # tell GraphQL-Ruby how to continue.
+      # @param query [GraphQL::Query]
+      # @return [:return_validation_error] Let GraphQL-Ruby return the (new) normal validation error for this query
+      # @return [String] A validation error to return for this query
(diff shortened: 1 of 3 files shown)


lib/graphql/schema.rb
```diff
@@ -1670,6 +1670,36 @@ def did_you_mean(new_dym = NOT_CONFIGURED)
...
 
+
+      # If you need to support previous, non-spec behavior which allowed selecting union fields
+      # but *not* selecting any fields on that union, set this to `true` to continue allowing that behavior.
+      #
+      # If this is `true`, then {.legacy_invalid_empty_selections_on_union} will be called with {Query} objects
+      # with that kind of selections. You must implement that method
+      # @param new_value [Boolean]
+      # @return [true, false, nil]
+      def allow_legacy_invalid_empty_selections_on_union(new_value = NOT_CONFIGURED)
```
gmac, line 1681: These are probably worth noting in the guide docs given that you have to invoke the setting on your schema or experience an out-of-the-box warning. I suppose the future state defaults this to `true`, so maybe the oddity is sufficiently short-lived.

rmosolgo: Yes, it's a bummer to have a warning as the default behavior. But my guess is that this edge case is rarely encountered, since other GraphQL implementations (rightly) reject this -- like GraphiQL, for example.

gmac: Last I looked, we had about 30K / 5K RPM of these two cases (conflicting types / missing union selections), so plenty of clients in the world are running without client-side validation unfortunately.

gmac: Good question. Those observations were from over a year ago; I want to say there was ~6M RPM on that API back in those days? So yeah, not a significant total share, but still a lot of traffic. I'm setting up new metrics for this update so we can get a fresh sense of quantity and, more importantly, how many consumers they're coming from.

rmosolgo: The other interesting this would be how many _unique_ query strings. It could be so few query strings that you could basically hardcode a replacement query string for known "offenders" and then call it day.

PR #5330, opened 2025-04-17
ActiveRecordSource: Improve composite primary key support

rmosolgo:
This makes ActiveRecordSource and ActiveRecordAssociationSource work with composite primary keys.

### DIFF
--- a/lib/graphql/dataloader/source.rb
+++ b/lib/graphql/dataloader/source.rb
@@ -133,7 +133,7 @@ module GraphQL
         fetch_h = @pending
         @pending = {}
         @fetching.merge!(fetch_h)
-        results = fetch(fetch_h.values)
+        results = fetch(fetch_h.keys)
         fetch_h.each_with_index do |(key, _value), idx|
           @results[key] = results[idx]
         end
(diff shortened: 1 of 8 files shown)


lib/graphql/dataloader/source.rb
```diff
@@ -133,7 +133,7 @@ def run_pending_keys
         fetch_h = @pending
         @pending = {}
         @fetching.merge!(fetch_h)
-        results = fetch(fetch_h.values)
+        results = fetch(fetch_h.keys)
```
rmosolgo, line 136: This is actually a big change -- it will change what is passed into `fetch(...)`. I think this is the right implementation because in the current implementation, what goes to `fetch(...)` may vary depending on the order of calls to `load(...)`. (Specifically, if someone calls (`load(1)`, then `load("1")`, then `1` will go to `fetch`, but if the order is reversed, then `"1"` will go to `fetch`. This is bad behavior.) 

But this change should come with a warning of some kind first.

rmosolgo: Followup: There is a test for this, and it makes sense. I'm going to add a new method for this.

rmosolgo: I added `normalize_fetch_key` to handle this in df37d1715

PR #5347, opened 2025-05-11
fix(c-parser): Disallow `BANG` in `TypeCondition`

### LINKED ISSUE #5345: Parsing `!` incorrectly wraps fragment name usages
**Describe the bug**

It appears that the non-null `!` modifier is respected while parsing fragment name usage. For example, given this woefully misguided document:

```ruby
ast = GraphQL.parse("{ ... on Sfoo! { id } }")
```

It gets happily parsed into an InlineFragment node with a non-null-wrapped TypeName:

```ruby
#<GraphQL::Language::Nodes::Document:0x00000003bf3b34f0
 @col=1,
 @definitions=
  [#<GraphQL::Language::Nodes::OperationDefinition:0x00000003917d4800
    @col=1,
    @definition_pos=nil,
    @directives=[],
    @filename=nil,
    @line=1,
    @name=nil,
    @operation_type="query",
    @pos=nil,
    @selections=
     [#<GraphQL::Language::Nodes::InlineFragment:0x00000003bf3b3540
       @col=3,
       @directives=[],
       @filename=nil,
       @line=1,
       @pos=nil,
       @selections=[...],
       @source=nil,
       @type=
        #<GraphQL::Language::Nodes::NonNullType:0x00000003bf3b3590
         @col=10,
         @filename=nil,
         @line=1,
         @of_type=
          #<GraphQL::Language::Nodes::TypeName:0x00000003bf3b35e0
           @col=10,
           @filename=nil,
           @line=1,
           @name="Sfoo",
           @pos=nil,
           @source=nil>,
         @pos=nil,
         @source=nil>>],
    @source=nil,
    @variables=[]>],
 @filename=nil,
 @line=1,
 @pos=nil,
 @source=nil>
```

Which subsequently errors during validation with:

```
undefined method 'name' for an instance of GraphQL::Language::Nodes::NonNullType
```

**Versions**

C Parser

**Expected behavior**

Should be a parser error.

**Actual behavior**

Passes parsing and subsequently fails in upstream operations.

JoviDeCroock:
Resolves #5345 

Because the C-parser uses [`type`](https://github.com/JoviDeCroock/graphql-ruby/blob/d63dacf8595262eef725d794c6e7020ef93f41d2/graphql-c_parser/ext/graphql_c_parser_ext/parser.y#L459-L461) for the `typeCondition` we allow for `BANG` to be present in the `typeCondition` this is in conflict with the GraphQL specification on both [FragmentDefinition](https://spec.graphql.org/draft/#FragmentDefinition) as well as [InlineFragment](https://spec.graphql.org/draft/#sec-Inline-Fragments).

The PR adds a test and updates the C code, the test proves that it now works on both C as well as Ruby (no changes in Ruby needed).

I do have a slight concern of my compilation outcome, the diff seems quite large for the `.h` file, I just ran the rake task though so not sure where I went wrong there.

Feel free to this PR


spec/graphql/language/parser_spec.rb
```diff
@@ -95,6 +95,18 @@
     }
   end
 
+  it "raises a parse error for invalid type modification on inline fragment spread" do
+    assert_raises(GraphQL::ParseError) {
+      GraphQL.parse("{ ... on Foo! { id } }")
```
gmac, line 100: Should we test fragment spread as well? `{ ... Foo! }`

JoviDeCroock: We can def test it, the main bug originated from using `type` everywhere which is valid for SDL and VariableDefinitions but not for `typeCondition` which isn't being used here.

rmosolgo:
Hey, thanks for the fix here! Sorry about the trouble generating those parser files. Weirdly on my machine: 

- It doesn't generate `parser.h` -- and the generated `parser.h` from this branch caused `rake build_ext` to fail because of the removed function definition. 
- It generates different contents for `parser.c` 

I don't know why there are these differences :S I'm certainly willing to improving the system to produce more reliable output, but for now, I pulled your branch and regenerated those files. 

I'm surprised CI passed with the overridden `parser.h` 😅

JoviDeCroock:
Thank you so much for hopping in and regenerating. Yes, I am not sure what's up, I am sporting the latest bison and just executed the `rake` command.

rmosolgo:
🚢  in graphql-c_parser v1.1.3! Thanks again.

Issue #5348, opened 2025-05-12
Errors using `PerfettoTrace` with UTF-8 characters


nsab1n:
**Describe the bug**

When using `PerfettoTrace`, it throws if it tries to log content containing UTF-8 characters here:

https://github.com/rmosolgo/graphql-ruby/blob/35421473ab20f5f2b9372d16e3a783d0828cee4d/lib/graphql/tracing/perfetto_trace.rb#L647

This happens if, for example, GraphQL is used to retrieve data containing UTF-8, as the result of the query is logged in the debug properties. I did some debugging myself, and found this is because the type of `InternedString::name` in the proto is `bytes`. If the input string is ASCII, this can be converted directly, but if it contains UTF-8 characters, they need to be unpacked into their raw byte representations first.

**Steps to reproduce**

This can be reproduced by adding the following entry to the `data` array in `spec/support/active_record_setup.rb`:

```
{
  author: "한강",
  titles: [
    "작별하지 않는다",
    "흰",
    "소년이 온다",
    "노랑무늬영원"
  ]
}
```

`perfetto_trace_spec.rb` will then fail with:

```
GraphQL::Tracing::PerfettoTrace#test_0001_traces fields, dataloader, and activesupport notifications:
Encoding::UndefinedConversionError: U+D55C from UTF-8 to ASCII-8BIT
    lib/graphql/tracing/perfetto_trace.rb:647:in `initialize'
    lib/graphql/tracing/perfetto_trace.rb:647:in `new'
    lib/graphql/tracing/perfetto_trace.rb:647:in `block in new_interned_data'
```

which is the same error I got when testing manually.

**Suggested fix**

Strings have a `b` method to return a copy of the string as bytes, so the fix is straightforward (I think - this worked for me, anyway):

```
// `perfetto_trace.rb`, line 647
str_vals = @new_interned_da_string_values.map { |name, iid| InternedString.new(iid: iid, str: name.b) }
```

I think changing the type of `str` to `string` would also work.

**Additional context**
I actually have a branch ready to go with:

* above fix
* change to `active_record_setup` to reproduce the bug
* changes to rails snapshots

But I couldn't push my branch to submit a PR as I don't think I have access rights.

nsab1n:
Thanks! Haven't contributed to OSS before so was a bit unsure on how to submit a PR 😅 

Have submitted a PR 👍 (the huge LOC is due to snapshot changes, BTW)

PR #5351, opened 2025-05-12
fix(perfetto_trace): pass bytes to InternedString

nsab1n:
### Context
Issue: https://github.com/rmosolgo/graphql-ruby/issues/5348

Using `DetailedTrace`/`PerfettoTrace`, if the content being logged as part of the debug information includes unicode characters, an error is thrown: `U+D55C from UTF-8 to ASCII-8BIT`. The type of `InternedString::str` is bytes, so passing it a string directly works if the string is 8-bit ASCII, but UTF-8 characters need to be unpacked to their bytes representation first. Luckily ruby has a `.b` method which returns a bytestring copy of a given string.

Screen grab showing utf-8 content successfully getting rendered in Perfetto (👋🏻 emoji):
image

### Testing
I've added a UTF-8 entry to the test data for `PerfettoTrace`, which gives the expected error without the main change in this PR, and then the test succeeds with the changes in this PR.

I've also updated the snapshots, but I was unable to update the `example-rails-8-1.json` one 🤔 there's no corresponding gemfile, how do I generate it?

### DIFF
--- a/lib/graphql/tracing/perfetto_trace.rb
+++ b/lib/graphql/tracing/perfetto_trace.rb
@@ -644,7 +644,7 @@ module GraphQL
         end
 
         if !@new_interned_da_string_values.empty?
-          str_vals = @new_interned_da_string_values.map { |name, iid| InternedString.new(iid: iid, str: name) }
+          str_vals = @new_interned_da_string_values.map { |name, iid| InternedString.new(iid: iid, str: name.b) }
           @new_interned_da_string_values.clear
         end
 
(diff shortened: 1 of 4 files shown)


rmosolgo:
Hey, thanks so much for this fix! That last snapshot is generated from the `rails_master` gemfile. Do you mind updating that one, too?

nsab1n:
Done 👍

rmosolgo:
Thanks for this improvement!

PR #5356, opened 2025-05-17
Optimize value coercion

JoviDeCroock:
# Optimize Argument Coercion Performance

This PR attempts to optimize the argument coercion process in GraphQL-Ruby by adding fast paths for common scenarios. I thought I'd create it as a draft so folks could give input and the CI would run.

Looks like I was a bit too eager, looking into the tests etc

## Changes

1. **List Type Coercion** (`schema/list.rb`)
   - Added fast path to avoid unnecessary array creation when input is already an array
   - Special handling for scalar inputs that need conversion to arrays
2. **Arguments Cache** (`execution/interpreter/arguments_cache.rb`)
   - Added optimization to skip argument materialization for empty arguments
   - Fast path when argument owner has no arguments defined
3. **Argument Handling** (`schema/member/has_arguments.rb`)
   - Added fast path for simple cases with small number of arguments (0-2)
   - Simplified execution flow for cases without lazy values
   - Added method to detect if arguments have default values
4. **Argument Coercion** (`schema/argument.rb`)
   - Optimized `coerce_into_values` with a fast path for statically coercible arguments
   - Direct value handling for simple cases without loading or preparation

## Benchmarks

Performance improvements measured with isolated benchmarks, these were custom made as I couldn't get the benchmarks of the repository to run well 😅 :

- **List Coercion**:
  - Array input: ~6% improvement
  - Scalar input: ~48% improvement

- **Argument Coercion**:
  - Empty arguments: ~74% improvement
  - Simple arguments (1-2 args): ~8% improvement
  - Complex arguments (5+ args): ~5% improvement

### DIFF
--- a/lib/graphql/execution/interpreter/arguments_cache.rb
+++ b/lib/graphql/execution/interpreter/arguments_cache.rb
@@ -30,7 +30,6 @@ module GraphQL
               @storage[argument_owner][parent_object][ast_node] = resolved_args
             end
           end
-
         end
 
         # @yield [Interpreter::Arguments, Lazy<Interpreter::Arguments>] The finally-loaded arguments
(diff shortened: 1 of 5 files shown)


rmosolgo:
Hey, thanks for looking into these! If you're interested in pursuing them further, I'd be happy to help debug test failures -- but it'd be a lot easier if you could open PRs for the one-by-one, just so there wouldn't be quite as many moving parts involved.

PR #5357, opened 2025-05-21
Half-baked PoC: Logging arguments for a field

alexcwatt:
I would like to be able to log the arguments that are provided to a field in the same format that they came into the GraphQL API. I was attempting to do this with a FieldExtension. The `arguments` available there are the ones parsed into Ruby.

I put together this PR to demonstrate what I'm hoping to accomplish and show the issue that I'm having (test does not pass). I am hoping for some feedback on how to go about this. Maybe it's already quite easy and I'm missing something, or maybe there is some work to be done.

@rmosolgo when you have a chance, I'd love your input; happy to hack on this some more, but would love some guidance.

Note to self: `bundle exec ruby -Ilib -Ispec spec/graphql/schema/mutation_spec.rb` to run the test...

### DIFF
--- a/spec/graphql/schema/mutation_spec.rb
+++ b/spec/graphql/schema/mutation_spec.rb
@@ -355,12 +355,45 @@ describe GraphQL::Schema::Mutation do
   end
 
   class LogArgumentsExtension < GraphQL::Schema::FieldExtension
+    extras [:ast_node]
     def resolve(object:, arguments:, context:)
       # I would like to find a way to log the arguments with the following changes:
       # 1. Use the camelcase version
       # 2. Use the raw inputs, not the Ruby parsed ones
-      context[:logged_arguments] = arguments.dup
+      ast_node = arguments[:ast_node]
+      formatted_args = {}
+      ast_node.arguments.each do |argument_node|
+        formatted_args[argument_node.name] = format_argument(context, argument_node.value)
+      end
+      context[:logged_arguments] = formatted_args
       yield(object, arguments)
+    end
+
+    private
+
+    def format_argument(context, argument_value)
+      case argument_value
+      when String, Numeric
+        argument_value
+      when Array
+        argument_value.map { |v| format_argument(context, v) }
+      when GraphQL::Language::Nodes::InputObject
+        args = {}
+        argument_value.arguments.each do |arg_node|
+          args[arg_node.name] = format_argument(context, arg_node.value)
+        end
+        args
+      when GraphQL::Language::Nodes::VariableIdentifier
+        var_value = context.query.variables[argument_value.name]
+        formatted = format_argument(context, var_value)
+        # You could return one of these or both, depending on your needs:
+        {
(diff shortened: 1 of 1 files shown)


rmosolgo:
Hey, happy to help make sense of this. A couple of other thoughts to guide the discussion: 

- Do you want to log a message when the field is resolved (maybe more than once per API transaction), or once per API call? If you only want to log once per API call, then a query analyzer (https://graphql-ruby.org/queries/ast_analysis.html) might be a better approach.
- When the client uses a variable for the argument value, do you want to log the variable name, or the value used at runtime? 

In any case, I have a hunch that using the AST instead of the Ruby arguments would be a better approach. I sketched out a quick take on that here: https://gist.github.com/rmosolgo/9d139454c50678cc23ec792402844d75

alexcwatt:
Thank you @rmosolgo! That's amazing. My plan is to do this on a field extension for mutations in particular; we already have the extension. I didn't realize that I could use `extras [:ast_node]`. I pushed an update that takes what you did and integrated it here.

To answer your questions:
1. I am thinking of logging once per mutation field
2. I am planning to store the runtime value

spec/graphql/schema/mutation_spec.rb
```diff
@@ -353,4 +353,87 @@ def self.resolve_type(abs_type, obj, ctx)
...
+    def resolve(object:, arguments:, context:)
+      # I would like to find a way to log the arguments with the following changes:
+      # 1. Use the camelcase version
+      # 2. Use the raw inputs, not the Ruby parsed ones
+      ast_node = arguments[:ast_node]
+      formatted_args = {}
+      ast_node.arguments.each do |argument_node|
+        formatted_args[argument_node.name] = format_argument(context, argument_node.value)
+      end
+      context[:logged_arguments] = formatted_args
```
rmosolgo, line 368: The GraphQL spec allows clients to run multiple mutations in a single operation. If that happens, then subsequent assignments to `context[:logged_arguments]` will clobber previous ones. 

If you need to avoid that, you could do something like 

```ruby
logged_arguments_by_path = context[:logged_arguments] ||= {} 
logged_arguments_by_path[context.current_path] = formatted_args
```

alexcwatt: 👍🏻 thank you! In practice, I was planning to emit a log and not store in `context`, just did this for writing this test easily

Do you think there is any place in the gem for having a utility to support this kind of log formatting? I am guessing not - it's quite easy to take the `ast_node` and I am adding a few more cases (Hash, TrueClass, FalseClass, and anything else that CI finds for me) but wanted to ask

rmosolgo: I don't think so ... we can point people here if it comes up again, and if it turns out to be widely used, we can add it later.

alexcwatt:
Again huge thank you @rmosolgo for the quick unblock!

PR #5358, opened 2025-05-21
Scalar validation: remove redundant infinity handling

swalkinshaw:
`GraphQL::Language.serialize` already has its own special handling for `Float::INFINITY`. It serializes it as "Infinity" which is inconsistent with this empty string as well.


rmosolgo:
Thanks for this cleanup!

PR #5370, opened 2025-06-02
Minimal Ractor support

rmosolgo:
I've been following along with `@byroot`'s recent Ractor-related blog posts, and it makes me want to make GraphQL-Ruby play nice with non-main Ractors. I think it will be a long road ... but it starts with a single step! I'm opening this PR to start a conversation (if anyone is interested) about how/what to accomplish. 

Goals for non-main Ractors:

- Execute a query and return a result Hash to the main Ractor
- ~~Parse schemas from definition~~ Not currently possible because Schemas are classes and non-main Ractors can't assign instance variables of classes.
- Somehow provide an escape hatch for fetching data. For example, could the `dataloader` interface work for sending IDs up to the main ractor, then receiving data in the non-main Ractor? (Ractor-friendliness in GraphQL-Ruby is meaningless if there's now way to make an ActiveRecord query during execution...)
- Stretch goal, maybe someday: build a `GraphQL::Query` in the main Ractor, then run a dependent `Query::Partial` in a non-main Ractor. This is very hard because `GraphQL::Query` has a bunch of dynamic caches.

Non-goals: 

- Modifying schema definition in a non-main Ractor is impossible 

Currently, GraphQL-Ruby supports incremental, dynamic schema definition: types, fields, etc, can all be added on-the-fly, modifying schema definitions. But I think Ractor-friendliness will require some API for freezing a schema definition. `RactorShareable` does that here.

-------

TODO: 

- [x] Support looking up types and fields in non-main Ractor
- [ ] Try a `Partial` in a non-main Ractor
- Another PR: implement pass-through dataloader


rmosolgo:
Some Fiber-related tests are failing due to: https://bugs.ruby-lang.org/issues/21401

Issue #5371, opened 2025-06-03
Support .unsubscribe in AblyLink


rmosolgo:
PusherLink implements `unsubscribe`: 

https://github.com/rmosolgo/graphql-ruby/blob/e6cc76d75e85c9dfaf00de843fe527581048cf1a/javascript_client/src/subscriptions/PusherLink.ts#L113-L119

But `AblyLink` doesn't. It could, but I think this `subscribe` call needs to return an object like that: 

https://github.com/rmosolgo/graphql-ruby/blob/a4f4341555e6d734dd14b647cd083f630c5e445f/javascript_client/src/subscriptions/AblyLink.ts#L56

Mentioned in https://github.com/rmosolgo/graphql-ruby/issues/3714#issuecomment-2936216509

rmosolgo:
Fixed by #5374 and released in graphql-ruby-client v1.14.7

PR #5374, opened 2025-06-04
Support .unsubscribe in AblyLink

1kuko3:
- address missing feature https://github.com/rmosolgo/graphql-ruby/issues/5371
- props to @erickreutz


rmosolgo:
Hey, thanks for sharing this improvement! I added a test for `.unsubscribe` and found that I needed one more bit of cleanup to make it pass. (Interestingly, `PusherLink` doesn't have that same cleanup -- I'll have to investigate that!)

rmosolgo:
🚢  in graphql-ruby-client v1.14.7! Thanks again.

1kuko3:
@rmosolgo Thank you for your improvements, adding tests, and releasing it!

Issue #4711, opened 2023-11-22
Incorrect `path` attribute on query validation error


utay:
**Describe the bug**

With this query:
```
query TestQuery($slug: String! = "utay") {
  user(slug: $slug) {
    id
  }
}
```

This error is returned:
```
{
  "errors": [
    {
      "message": "Non-null variable $slug can't have a default value",
      "locations": [
        {
          "line": 1,
          "column": 17
        }
      ],
      "path": [
        "query TestQuery"
      ],
      "extensions": {
        "code": "defaultValueInvalidOnNonNullVariable",
        "variableName": "slug"
      }
    }
  ]
}
```

The issue here is `path`; it's equal to "query TestQuery" and I think it is not exactly compliant with the [graphql spec](https://spec.graphql.org/October2021/#sec-Errors.Error-result-format) which expects `path` to be an array of fields. It is an issue for us because we use a [federation router](https://github.com/apollographql/router) in front of the graphql server and `path` is invalid for the router in this case. 

**Versions**

`graphql` version: 2.1
`rails` version: 7.0.7

**Expected behavior**

Maybe `path` shouldn't be set when there's a validation error in the query arguments as some other graphql implementations seem to do that. 

(It's a separate topic but some implementations accept default values for non-null variables and it's not clear to me what is the right behavior per the spec.)

rmosolgo:
Hey, sorry I didn't get back to you on this earlier! I agree that the `path` on this error should match the spec. 

Also, looking back at the spec, I can see that this query _doesn't_ violate it. It's allowed within the given grammar and not expressly prohibited elsewhere. 

I'll keep this open until I've checked the `path`s and removed this error from the project.

rmosolgo:
👋  Just one update on this, GraphQL-Ruby doesn't add this error to queries like this anymore (since https://github.com/rmosolgo/graphql-ruby/pull/5030). I'm hoping to clean up these errors soon.

PR #5375, opened 2025-06-04
Improve scalar result coercion execution errors

swalkinshaw:
Adds improved handling for `coerce_result` errors which results in more spec compliant errors with `path` and `locations` present.

Separated out from https://github.com/rmosolgo/graphql-ruby/pull/5306 because this can be merged more easily.

### DIFF
--- a/spec/graphql/schema/scalar_spec.rb
+++ b/spec/graphql/schema/scalar_spec.rb
@@ -119,18 +119,28 @@ describe GraphQL::Schema::Scalar do
         def self.coerce_input(val, ctx)
           raise GraphQL::CoercionError, "#{val.inspect} can't be Custom value"
         end
+
+        def self.coerce_result(val, ctx)
+          raise GraphQL::CoercionError, "#{val.inspect} can't be Custom value"
+        end
       end
 
       class Query < GraphQL::Schema::Object
         field :f1, String do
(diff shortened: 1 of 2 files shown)


spec/graphql/schema/scalar_spec.rb
```diff
@@ -143,8 +153,17 @@ class Query < GraphQL::Schema::Object
...
       assert_equal [expected_error], result["errors"]
     end
-  end
 
+    it "makes a nice validation error for reuslt coercion" do
+      result = CoercionErrorSchema.execute("{ f2 }")
+      expected_error = {
+        "message" => "\"bad\" can't be Custom value",
+        "locations" => [{"line"=>1, "column"=>3}],
+        "path" => ["f2"],
```
swalkinshaw, line 162: Interesting difference here. The input example above (from static validation) includes `query` in the path and this doesn't.

Looking at the spec, I don't think `query` should be included? But that's a separate issue (if at all).

rmosolgo: Yeah, `query` doesn't belong there. Related: https://github.com/rmosolgo/graphql-ruby/issues/4711

rmosolgo:
Thanks for this improvement!

PR #5377, opened 2025-06-04
Fix directive arguments with Ruby values

### LINKED ISSUE #5376: Improvements to schema directive validations is breaking enums that use symbols
**Describe the bug**

Changes made in [this PR](https://github.com/rmosolgo/graphql-ruby/commit/a53bc8e20903ffb1f988d9ad33feb80a5cfeb4bf) are causing validation failures for enums using this syntax

```ruby
value "PUBLIC", value: :public
```

Specifically, [this block](https://github.com/rmosolgo/graphql-ruby/blob/4ecf14faf77262fcfe4ff8231f131e6b7dadc09b/lib/graphql/schema/enum.rb#L180) is now raising a validation error where it previously did not. 

 `v.graphql_name` is `“PUBLIC”` but `value_name` is `:public` and so `matching_value.nil?` returns `true`

```ruby
allowed_values = ctx.types.enum_values(self)
    matching_value = allowed_values.find { |v| v.graphql_name == value_name }
    if matching_value.nil?
      GraphQL::Query::InputValidationResult.from_problem("Expected #{GraphQL::Language.serialize(value_name)} to be one of: {allowed_values.map(&:graphql_name).join(', ')}")
    else
      nil
    end
```

**Versions**

`graphql` version: 2.5.8
`rails` (or other framework):

**GraphQL schema**

```ruby
value "PUBLIC", value: :public
```

**Expected behavior**
From the [graphql-ruby docs](https://graphql-ruby.org/type_definitions/enums.html)

```
By default, Ruby strings correspond to GraphQL enum values. But, you can provide value: options to specify a different mapping. For example, if you use symbols instead of strings, you can say:

value "AUDIO", value: :audio

Then, GraphQL inputs of AUDIO will be converted to :audio and Ruby values of :audio will be converted to "AUDIO" in GraphQL responses.
```

**Actual behavior**

What specifically went wrong?

Getting this error as a result
```ruby
GraphQL::Schema::Directive::InvalidArgumentError:
  @cacheControl.scope on Money.value is invalid (:public): Expected "public" to be one of: PUBLIC, PRIVATE
```

rmosolgo:
Fixes #5376

### DIFF
--- a/lib/graphql/schema/directive.rb
+++ b/lib/graphql/schema/directive.rb
@@ -129,8 +129,20 @@ module GraphQL
         # not runtime arguments.
         context = Query::NullContext.instance
         self.class.all_argument_definitions.each do |arg_defn|
-          value = arguments[arg_defn.keyword]
-          result = arg_defn.type.validate_input(value, context)
+          if arguments.key?(arg_defn.keyword)
+            value = arguments[arg_defn.keyword]
+            # This is a Ruby-land value; convert it to graphql for validation
+            graphql_value = begin
+              arg_defn.type.unwrap.coerce_isolated_result(value)
+            rescue GraphQL::Schema::Enum::UnresolvedValueError
+              # Let validation handle this
+              value
+            end
+          else
+            value = graphql_value = nil
+          end
+
+          result = arg_defn.type.validate_input(graphql_value, context)
           if !result.valid?
             raise InvalidArgumentError, "@#{graphql_name}.#{arg_defn.graphql_name} on #{owner.path} is invalid (#{value.inspect}): #{result.problems.first["explanation"]}"
           end
--- a/spec/graphql/schema/directive_spec.rb
+++ b/spec/graphql/schema/directive_spec.rb
@@ -66,18 +66,6 @@ Use `locations(OBJECT)` to update this directive's definition, or remove it from
     end
 
     assert_equal "@secret.topSecret on Thing.something is invalid (nil): Expected value to not be null", err.message
-
-    err2 = assert_raises GraphQL::Schema::Directive::InvalidArgumentError do
-      GraphQL::Schema::Field.from_options(
-        name: :something,
-        type: String,
-        null: false,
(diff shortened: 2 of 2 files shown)


juliaren:
hi, I suspect there may still be a bug regarding directive arguments validation against ruby hashes.

I recently tried to upgrade my repo's graphql-ruby gem from 2.2.17 -> 2.5.9 and getting an error: `GraphQL::Schema::Directive::InvalidArgumentError: @documentation.options on <my_field> is invalid ([{:key=>"1", :value=>"1"}]): Expected value to not be null`

I have a `documentation` Directive with a single argument: `argument :options, [DirectiveOption, null: true], required: false` with a custom input object `DirectiveOption` defined as:
```
class DirectiveOption < GraphQL::Schema::InputObject
    argument :key, String, required: true
    argument :value, String, required: true
 end
```

in another module, I attach this directive to relevant Schema objects by: `directive(Directives::SpectaQL, options: directive_options)` where `directive_options` is an array of key-value pairs like: `[{ :key => '1', :value =>'1' }, { :key => '2', :value => '2' }]`. I attempted a fix by using string keys instead of symbol keys: `[{ 'key' => '1', 'value' =>'1' }, { 'key' => '2', 'value' => '2' }]` but still got the same error. So far the only workaround (not ideal) is not using a custom input object and using a scalar type instead. 

Please let me know if this is indeed a bug and any immediate fix you may have, thank you!

rmosolgo:
Hey, that _does_ sound like a bug. Could you please open a new issue with that context along with the full backtrace of the error that was raised? I don't have any workarounds off the top of my head but maybe the backtrace will give a better clue as to what's going wrong.

juliaren:
> Hey, that _does_ sound like a bug. Could you please open a new issue with that context along with the full backtrace of the error that was raised? I don't have any workarounds off the top of my head but maybe the backtrace will give a better clue as to what's going wrong.

thank you! opened issue here: https://github.com/rmosolgo/graphql-ruby/issues/5424

PR #5382, opened 2025-06-11
Inherit validation allowances

gmac:
The new validation and complexity allowances do not inherit from a base class down to subclasses like other similar settings. This fixes that.

- Adds inheritance for:
   - `allow_legacy_invalid_empty_selections_on_union` 
   - `allow_legacy_invalid_return_type_conflicts`
   - `complexity_cost_calculation_mode`
- Inheritance tests for each setting.
- Confirmed no use in hot paths.


rmosolgo:
Thanks for this improvement!

PR #5402, opened 2025-07-11
Direct multiplex

gmac:
Support direct multiplex with special configurations


gmac:
Whoops, ignore. Just tinkering on a branch...

PR #5405, opened 2025-07-17
docs: correct `allow_null` example to indicate null values are permitted

tetzng:


rmosolgo:
🤦  Thanks for the fix!

Issue #5384, opened 2025-06-13
[GraphQL::Schema::Visibility] Inherited schemas don't preload resulting in crash


cocoahero:
### TL;DR;

When a parent schema enables the `GraphQL::Schema::Visibility` plugin using `preload: true`, child schemas will crash when declaring the root query object.

```
/ruby/3.4.4/gems/graphql-2.5.9/lib/graphql/schema/visibility.rb:191:in 'GraphQL::Schema::Visibility#ensure_all_loaded': undefined method 'add?' for nil (NoMethodError)

    if type.kind.fields? && @preloaded_types.add?(type)
                                            ^^^^^
from /ruby/3.4.4/gems/graphql-2.5.9/lib/graphql/schema/visibility.rb:96:in 'GraphQL::Schema::Visibility#query_configured'
from /ruby/3.4.4/gems/graphql-2.5.9/lib/graphql/schema.rb:451:in 'GraphQL::Schema.query'
from ./example.rb:20:in '<class:ConcreteSchema>'
from ./example.rb:19:in '<main>'
```

I believe the issue is [here](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema.rb#L1208-L1211), where the visibility instance is duplicated with `preload: true`, but the preloaded state is not initialized / copied.

### Version

`/ruby/3.4.4/gems/graphql-2.5.9`

### Example

```ruby
class QueryRoot < GraphQL::Schema::Object
  field :hello, String, null: false

  def hello
    "Hello, world!"
  end
end

class AbstractSchema < GraphQL::Schema
  # simulate Rails.env.production? == true
  use(GraphQL::Schema::Visibility, preload: true)
end

class ConcreteSchema < AbstractSchema
  query(QueryRoot)
end

result = ConcreteSchema.execute("{ hello }")
```

rmosolgo:
👋  Thanks for the detailed report and sorry about the trouble. I replicated the bug and I've got a fix in the works at #5386

spacebat:
We just got bitten by this during a production deployment as we have subclassed schemas and were relying on the default for preload, will update to the latest.

I suggest if a default behaviour is going to depend on the Rails environment, then instead of "it's different in production", it is better to have "it's different in development". That way bugs are more likely to be found by QA processes in staging environments.

Thanks for having a fix out there already!

rmosolgo:
[...]

PR #5409, opened 2025-07-30
Visibility: Default to preload: true when env.staging?

rmosolgo:
As suggested in https://github.com/rmosolgo/graphql-ruby/issues/5384#issuecomment-3135032249, this logic was kind of backwards. `development?` should get special treatment -- non-preloading -- while non-development environments (production _and_ staging, etc) get preloading by default.


cocoahero:
This would cause it to preload in `Rails.env.test?` which I am not sure is desired.

rmosolgo:
Thanks for the heads up on that, @cocoahero. I simplified it to just adding `|| Rails.env.staging?`.

PR #5411, opened 2025-08-04
Update CHANGELOG.md: Add placeholder entries for 1.13.*

jaredbeck:
Add placeholder entries for 1.13.20 through 1.13.25.

Not great, but better than nothing, IMO. 🤷


rmosolgo:
Hey, sorry I didn't respond to you sooner. The changelog entries for these releases are on the 1.13.x branch (just because that's where these versions were published from): https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/CHANGELOG.md

How about copy those over to master instead?

rmosolgo:
Happy to update master with the entries from the 1.13.x branch. Otherwise, if anyone discovers this issue, you can find an up-to-date changelog on the 1.13.x branch: https://github.com/rmosolgo/graphql-ruby/blob/1.13.x/CHANGELOG.md

PR #5425, opened 2025-08-31
Runtime hook for breadth execution patterns

gmac:
This adds a hook into the execution runtime that allows an evaluated selection to exit with its resolved inner value. This allows a breadth-first execution process to run a single generation of resolvers at a time across a set of objects, ex:

breadth-first

This general breadth capability relies on two runtime methods:

- `evaluate_selection(result_key, ast_nodes, selections_result)`
- `exit_with_inner_result?(inner_result, result_name, selection_result)`

The expected capabilities are covered by a simple example of a breadth-based runtime in tests. This is still an experimental workflow, so it seems worthwhile to keep the new code footprint minimal for the time being (just `exit_with_inner_result?`) and in a state that can be backed out easily.

In the future, we'd ideally formalize a structure like this `SimpleBreadthRuntime` as a supported library feature.


spec/graphql/execution/breadth_runtime_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+        breadth_object.breadth_index = index
+
+        state = get_current_runtime_state
+        state.current_result_name = nil
+        state.current_result = breadth_object
+        @dataloader.append_job { evaluate_selection(result_key, node, breadth_object) }
+      end
+
+      @dataloader.run
+      GraphQL::Execution::Interpreter::Resolve.resolve_each_depth(@lazies_at_depth, @dataloader)
```
gmac, line 53: In the actual prototype, I’m running dataloader inline with the field but this lazy step runs out-of-band after a bunch of fields have been queued. Sounds like with https://github.com/rmosolgo/graphql-ruby/pull/5422, I’d just run dataloader once in that out-of-band position?

spec/graphql/execution/breadth_runtime_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+        end
+      end
+      result
+    ensure
+      delete_all_interpreter_context
+    end
+
+    def evaluate_breadth_selection(objects, parent_type, node)
+      result_key = node.alias || node.name
+      @breadth_results_by_key[result_key] = Array.new(objects.size)
```
gmac, line 39: In the actual prototype, breadth results are keyed with a unique representation of a field path.

rmosolgo:
I'm game to merge this as-is with the caveat that the integration point for this customization will change in the near future. I expect to continue modifying execution code with ideas from #5389, specifically using a queue instead of recursive method calls and merging `GraphQLResult` objects into `GraphQL::Schema::Object`. So although the details will change, I can commit to supporting this kind of possibility in the future :+1:

gmac:
> I expect to continue modifying execution code with ideas from https://github.com/rmosolgo/graphql-ruby/pull/5389, specifically using a queue instead of recursive method calls and merging GraphQLResult objects into GraphQL::Schema::Object. So although the details will change, I can commit to supporting this kind of possibility in the future 👍

This is an interesting prospect. What's the grain of the queued work? Are you queuing an object+field at a time, then having the ejected result feed the next enqueued generation? If so, this sounds like an excellent design for a formal integration. If a breadth executor could enqueue a whole bunch of values for a field and get all the inner results back out, that gets us exactly what we need.

gmac:
> and merging GraphQLResult objects into GraphQL::Schema::Object.

I thoroughly support this, as it amounts to a substantial reduction in allocations. Any allocation that matches response objects (or worse, fields) at a 1:1 standard should be considered a premium. In Cardinal, I believe the only 1:1 response allocation we do is the response tree itself, which is unavoidable.

gmac:
@rmosolgo – is there something special about that mongodb test...? I can't even run in locally, presumably because I don't have the database. Not sure why that one is failing

rmosolgo:
No, sorry, it's flaky: 

```
Expected |0.3 - 0.374014811| (0.07401481100000001) to be <= 0.06.
```

It's testing that I/O actually ran in parallel. But depending on the server's situation, the spec fails because the operation didn't finish as fast as expected. I re-ran it.

Issue #5418, opened 2025-08-14
load_application_object_failed marked as private API?


lukes:
Hello!

It looks like `#load_application_object_failed` is marked as a private API, even though it's a recommended public hook https://github.com/rmosolgo/graphql-ruby/blob/68f160f064c50f73e221bd42d5b71e74b22a3a14/guides/mutations/mutation_classes.md#L173.

From the generated docs:

<img width="1037" height="111" alt="Image" src="https://github.com/user-attachments/assets/18eab813-091b-475d-a0fa-7aaab9ec14be" />

rmosolgo:
Hey, thanks for asking. I think the `@api private` is a mistake; this method is _intended_ to be overridden as described in the `mutation_classes.md` doc. If you have a minute to open a PR removing it, I'd appreciate it. Otherwise, I'll keep this open until I get a chance to do it!

PR #5426, opened 2025-09-03
Mark load_application_object_failed as public API

letiesperon:
## Summary
Add explicit `@api public` documentation to the `load_application_object_failed` method to clarify that it's a public hook intended to be overridden by users

## Context
As noted in #5418, the `load_application_object_failed` method appears to be marked as a private API in the generated documentation, which conflicts with the mutation guides that explicitly recommend overriding this method for custom error handling.

**In prod:** 

image

**Local yard server:**

image

### DIFF
--- a/lib/graphql/schema/member/has_arguments.rb
+++ b/lib/graphql/schema/member/has_arguments.rb
@@ -413,6 +413,12 @@ module GraphQL
             end
           end
 
+          # Called when an argument's `loads:` configuration fails to fetch an application object.
+          # By default, this method raises the given error, but you can override it to handle failures differently.
+          #
+          # @param err [GraphQL::LoadApplicationObjectFailedError] The error that occurred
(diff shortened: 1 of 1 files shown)


letiesperon:
Re-running CI to check if the Rails master + Ruby 3.4 failure is transient.

letiesperon:
**Some cases fail intermittently in CI**. It appears to be unrelated to this documentation-only change:
All tests pass locally with Rails 7.1. Other recent PRs show similar flaky behavior on these specific version combinations

**Recommendation:**
This appears to be environmental/flaky test issues rather than problems with the code change. The documentation fix is ready for review.

rmosolgo:
Thanks for this improvement! Sorry about the flaky tests 😖

letiesperon:
@rmosolgo no worries! big fan by the way. Glad I can help. (I want to start contributing more but I'm just diving into the code)

PR #5429, opened 2025-09-09
Adjust breadth interface hooks

gmac:
Small refactor to the breadth runtime interface... plugs a hole where argument errors were leaking out. This OO interface also feels a bit nicer by keeping the focus on the object scope rather than the runtime.

So – in the object sequence spanning `evaluate_selection ... evaluate_selection_with_resolved_keyword_args`, we basically just need to offer all values to `selection_result.collect_result` prior to calling `continue_value`. Seem reasonable?

### DIFF
--- a/lib/graphql/execution/interpreter/runtime/graphql_result.rb
+++ b/lib/graphql/execution/interpreter/runtime/graphql_result.rb
@@ -157,6 +157,11 @@ module GraphQL
               end
             end
           end
+
+          # hook for breadth-first implementations to signal when collecting results.
+          def collect_result(result_name, result_value)
+            false
+          end
         end
 
         class GraphQLResultArray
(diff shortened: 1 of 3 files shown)


lib/graphql/execution/interpreter/runtime/graphql_result.rb
```diff
@@ -157,6 +157,11 @@ def fix_result_order
               end
             end
           end
+
+          # hook for breadth-first implementations to signal when collecting results.
+          def collect_result(result_name, result_value)
```
swalkinshaw, line 162: `?` for predicate method?

```suggestion
          def collect_result?(result_name, result_value)
```

swalkinshaw: I assume you skipped it because it can have side effects. Probably better as is 👍

PR #5447, opened 2025-10-09
Fix typo in date enconding error

RicardoTrindade:
I believe the word be is duplicated when setting the error message for date enconding and can be removed


rmosolgo:
Thanks for this fix!

PR #5452, opened 2025-10-23
Ensure `data` entry exists for execution errors

swalkinshaw:
According to the spec, the `data` entry should always exist once execution starts. Only if an error is raised _before_ execution starts should the `data` key be omitted.

The only test that needed updating was when an `ExecutionError` was raised during the `execute_query` trace hook. Despite this running before `super`, this still happens "during execution" and should have the `data` key present.

Ref: https://spec.graphql.org/draft/#sec-Data

> If an error was raised before execution begins, the [response](https://spec.graphql.org/draft/#response) must be a [request error result](https://spec.graphql.org/draft/#request-error-result) which will result in no response data.

> If an error was raised during the execution that prevented a valid response, the "data" entry in the response should be null.

Arguably, this should have a better test case to show the issue beyond just a trace hook. In our case, we raised an `ExecutionError` within a root mutation's `authorized?` method and it resulted in no `data` key which is clearly incorrect.

@rmosolgo any idea where that test case should go?

### DIFF
--- a/lib/graphql/execution/interpreter.rb
+++ b/lib/graphql/execution/interpreter.rb
@@ -80,7 +80,6 @@ module GraphQL
                       end
                     rescue GraphQL::ExecutionError => err
                       query.context.errors << err
-                      NO_OPERATION
                     end
                   end
                   results[idx] = result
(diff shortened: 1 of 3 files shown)


rmosolgo:
LGTM :+1: 

It looks like `rails_master` is failing because Rails master is version `8.2.x`, and it's missing a snapshot. I'll update that when I get a chance.

> that test case 

I don't have a particularly good idea... seems like updating the existing test case is a good start. Is there another real-life case that you encountered that we could include in the test suite?

swalkinshaw:
Turns out there was an existing test case for this but the assertion used `assert_nil` which passed either way. I've updated it to assert on the full result.

rmosolgo:
Thanks for this improvement!

PR #5456, opened 2025-11-04
Conditionally add fiber-storage dependency for Ruby < 3.2

OuYangJinTing:
The `fiber-storage` dependency only was introduced in Ruby versions prior to 3.2.

### DIFF
--- a/gemfiles/mongoid_8.gemfile
+++ b/gemfiles/mongoid_8.gemfile
@@ -11,7 +11,6 @@ gem "mongoid", "~> 8.0"
 gem "libev_scheduler"
 gem "evt"
 gem "async"
-gem "fiber-storage"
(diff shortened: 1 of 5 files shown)


rmosolgo:
Hey, thanks for taking a look at this. 

This build runs Ruby 3.2.0 but it's currently failing because of a `require "fiber/storage"` call. Could you also update this PR to remove that call when it's not necessary? 

https://github.com/rmosolgo/graphql-ruby/actions/runs/19062331874/job/54444654887?pr=5456

OuYangJinTing:
@rmosolgo Sorry. I didn't notice the test issue before, but I've already fixed it.
There's still a test failure, but it doesn't seem related to the changes in this PR.

rmosolgo:
Thanks for this improvement!

PR #5467, opened 2025-11-24
Update Rails CI, fix Rails master

rmosolgo:


rmosolgo:
😖  For some reason, with rails-master, `ActiveRecord::Base.logger` isn't receiving log entries for queries that run. The queries _do_ run -- data is returned, anyway -- but the logger contents are just `""`...

rmosolgo:
I isolated the logging issue to a recent commit on Rails main and opened in issue to ask about it: https://github.com/rails/rails/issues/56230

rmosolgo:
Well, this is still broken in some cases. But it's better, so I'm going to merge it.

PR #5474, opened 2025-12-02
Add allow_all_hidden, make it default for required: :nullable

### LINKED ISSUE #5458: Creating optional argument_definition as parameter for RequiredValidator
My problem is that I want to be able to use `required: :nullable` but I only want this validation to be performed if the argument is 'visible'. Awareness of the argument's visibility is available on the `argument_definition`, but this is not one of the possible inputs to RequiredValidator. Thus, to ensure the RequiredValidator has visibility awareness, I want to pass it an `argument_definition` (not merely a symbol representing the argument).

This would allow me (on my end), to create a custom validator inheriting from RequiredValidator that does a visibility check (using the info in the `argument_definition`), and during system initialization, install this custom RequiredValidator in lieu of the vanilla RequiredValidator.

I'd like for the `argument_definition` to be passed in as an optional argument. An alternative would be making the [argument keyword](https://github.com/rmosolgo/graphql-ruby/blob/db11d0cc88f1d4ffafcbf421d62365176c4f2db0/lib/graphql/schema/validator/required_validator.rb#L39) accept a symbol (existing) or argument definition (proposed). If we go with this alternative, it might be worth also making [one_of](https://github.com/rmosolgo/graphql-ruby/blob/db11d0cc88f1d4ffafcbf421d62365176c4f2db0/lib/graphql/schema/validator/required_validator.rb#L38) accept **arrays of symbols** or **arrays of argument definitions**.

The solution would allow any users of the gem to have their RequiredValidators easily access attributes (which happen to be visibility, in my case) on the argument definition.

I've considered trying to search for the argument_definitions on my end, within my custom Validator class, but this feels more cumbersome:

```ruby
   class VisibilityAwareRequiredValidator < GraphQL::Schema::Validator::RequiredValidator
     def validate(_object, context, value)
      # For each argument in @one_of, check if it's visible before validating
      return nil unless arguments_visible?(context)
       super
     end
 
     private
    def arguments_visible?(context)
      # Get the owner (Field) to look up argument definitions
      return true unless @validated
 
      # @one_of is an array of argument keywords (symbols) that need to be checked
      @one_of.all? do |arg_keyword|
        argument_visible_for_keyword?(arg_keyword, context)
      end
    end

    def argument_visible_for_keyword?(arg_keyword, context)
      # Try to get the argument definition from the validated owner
      argument_definitions = if @validated.respond_to?(:all_argument_definitions)
        @validated.all_argument_definitions
      elsif context.types.respond_to?(:arguments)
        context.types.arguments(@validated)
      else
        return true # Can't determine visibility, assume visible
      end

      # Find the argument definition by keyword
      arg_defn = if argument_definitions.is_a?(Array)
        argument_definitions.find { |defn| defn.keyword == arg_keyword }
      elsif argument_definitions.is_a?(Hash)
        argument_definitions[arg_keyword]
      else
        nil
      end

      return true unless arg_defn # Can't find argument, assume visible

      # Check if the argument has visibility and if it's visible in this context
      if arg_defn.respond_to?(:visibility) && arg_defn.visibility
        arg_defn.visibility.visible?(context)
      else
        true # No visibility set, assume visible
      end
     end
   end
 end
```

If the change I'm proposing is made, this code simplifies to:
```ruby
  class VisibilityAwareRequiredValidator < GraphQL::Schema::Validator::RequiredValidator
    def validate(_object, context, value)
      return nil unless argument_visible?(context)
      super
    end

    private

    def argument_visible?(context)
      return true if @argument_definition.visibility.nil?
      @argument_definition.visibility.visible?(context)
    end
  end
```

rmosolgo:
Previously, `validates(required: { one_of: [...] })` would return an error to the client if all the options were hidden with `def visible?`. This is pointless because the client can't do anything in that case. In this new code: 

- if all `one_of: ...` options are hidden, it raises an error to the developer
- the developer can add `allow_all_hidden: true` to skip the `one_of:` validation in this case
- `required: :nullable` uses this code too. It now bypasses validation if the argument is hidden

Fixes #5458

### DIFF
--- a/lib/graphql/schema/validator/required_validator.rb
+++ b/lib/graphql/schema/validator/required_validator.rb
@@ -40,15 +40,13 @@ module GraphQL
         # @param message [String]
         def initialize(one_of: nil, argument: nil, allow_all_hidden: nil, message: nil, **default_options)
           @one_of = if one_of
-            @single_argument = false
             one_of
           elsif argument
-            @single_argument = true
             [ argument ]
           else
             raise ArgumentError, "`one_of:` or `argument:` must be given in `validates required: {...}`"
           end
-          @allow_all_hidden = allow_all_hidden.nil? ? @single_argument : allow_all_hidden
+          @allow_all_hidden = allow_all_hidden.nil? ? !!argument : allow_all_hidden
           @message = message
           super(**default_options)
         end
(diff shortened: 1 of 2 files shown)


ravangen (approved):
A couple minor suggestions. Thanks for taking a look!

rmosolgo:
Thanks for taking a look. I realized that combination one-ofs, for example `one_of: [[:a, :b], :c]`, weren't properly supported, so I added a test for that and made it work.

rmosolgo:
🚢  in 2.5.15!

Issue #5462, opened 2025-11-18
Add the benefit of dataloaders for mutation arguments


stevecrozz:
**Is your feature request related to a problem? Please describe.**

I have an input for a mutation which is an array of tree nodes and their children (adjacency list). When using the "loads" param for nested mutation arguments, they do load data, but don't batch the loads.

My scheme's object_from_id method is configured to use dataloader in the hopes that loading objects can be done in a small number of database queries.

**Describe the solution you'd like**

I would like it if the loads params I provide to TreeNode would execute my loader in batch with all the IDs at once. But instead, each invocation of Sources::ByGlobalID#fetch receives only one single node's IDs.

**Describe alternatives you've considered**

I have tried many approaches to get this to work. I thought perhaps I could add the 'loads' param to the top level argument in ApplyTreeMutation, but that one seems to be executed after the TreeNode loads methods. I've also looked into using prepare, which I think would work if used Types::TreeNode#coerce_input, but I'm guessing that won't include the authorization that comes when I use loads and I'm not sure this is a stable API.

**Additional context**

I have copied a simplified version of what I'm doing. Hopefully this is enough to understand.

~~~ruby
module Mutations
  class ApplyTreeMutation < BaseMutation
    argument :tree, type: [Types::TreeNode], required: true
  end
end

module Types
  class TreeNode < Types::BaseInputObject
    argument :child_ids, type: [ID], required: true, loads: Unions::TreeContent
    argument :id, type: ID, required: true, as: :object, loads: Unions::TreeContent
  end
end

class MySchema
  def self.object_from_id(encoded_id, context)
    context.dataloader.with(Sources::ByGlobalID).load(encoded_id)
  end
end

module Sources
  class ByGlobalID < GraphQL::Dataloader::Source
    def fetch(gid_params)
      objects_by_gid_param = GlobalID::Locator.locate_many(gid_params, ignore_missing: true).index_by(&:to_gid_param)

      gid_params.map do |gid_param|
        objects_by_gid_param[gid_param]
      end
    end
  end
end

~~~

rmosolgo:
Hey, thanks for the detailed writeup. I agree it would be better to make this work the way you described. Implementation-wise, it'd be complicated just because of how it's currently implemented, where each "depth" of arguments must be completely resolved before proceeding to the next depth, and as you noticed, it doesn't batch across lists. 

[...]

PR #5479, opened 2025-12-05
Don't nest Async primitives inside plain Ruby fibers

### LINKED ISSUE #5463: Deadlock in AsyncDataloader with latest async versions
**Describe the bug**
Deadlock occurs when executing GraphQL queries with association loading using AsyncDataloader with the `async` gem after v2.28.1: 
Relevant changes: https://github.com/socketry/async/compare/v2.28.1...v2.29.0
Likely key change: https://github.com/socketry/async/commit/cf7fce100077552470ceaf32d6a903aa13c0ee30

**Versions**
graphql 2.5.14
Roda 3.89.0
Falcon 0.52.4 
async 2.34.0
sequel 5.98.0

**GraphQL schema**
```
class Gql::Types::Foo < GraphQL::Schema::Object
  field :id, ID, null: false
  field :bar_models, [Gql::Types::Bar], null: false

  def bar_models
    dataloader.with(Sources::RecordsByForeignKey, ::BarModel, :foo_model_id).load(object.id)
  end
end

class Gql::Types::Bar < GraphQL::Schema::Object
  field :id, ID, null: false
  field :foo_model, Gql::Types::Foo, null: false

  def foo_model
    dataloader.with(Sources::RecordById, ::FooModel).load(object.foo_model_id)
  end
end

  query Gql::Types::Query
  use GraphQL::Dataloader::AsyncDataloader
end
```

**GraphQL query**
```
{
  getAllFoos {
    id
    name
    barModels { id name }
  }
  getAllBars {
    id
    name
    fooModel { id name }
  }
}
```

(no response - server deadlocks)

**Steps to reproduce**

Clone this demo repo and run it using the steps provided in the README:
https://github.com/iyotov-havelock/async-dataloader-issue 

**Expected behavior**

Query returns all FooModels with their associated BarModels, and all BarModels with their associated FooModel. That's the case when using `GraphQL::Dataloader`, but not when using `GraphQL::Dataloader::AsyncDataloader`

**Actual behavior**

Server deadlocks during dataloader execution

```
fatal: No live threads left. Deadlock?
1 threads, 1 sleeps current:0x00005c4ab8f61890 main thread:0x00005c4ab8f61890
* #<Thread:0x000071ad9736a6d0 sleep_forever>

→ /gems/async-2.34.0/lib/async/promise.rb:89 in 'Thread::ConditionVariable#wait'
  /gems/async-2.34.0/lib/async/promise.rb:89 in 'block in Async::Promise#wait'
  /gems/async-2.34.0/lib/async/task.rb:258 in 'Async::Task#wait'
  /gems/async-2.34.0/lib/kernel/sync.rb:28 in 'Kernel#Sync'
  /gems/graphql-2.5.14/lib/graphql/dataloader/async_dataloader.rb:34 in 'block in GraphQL::Dataloader::AsyncDataloader#run'
```

**Additional context**
This looks like a compatibility issue between `AsyncDataloader` and async gem versions after v2.28.1
Relevant changes: https://github.com/socketry/async/compare/v2.28.1...v2.29.0
Likely key change: https://github.com/socketry/async/commit/cf7fce100077552470ceaf32d6a903aa13c0ee30

rmosolgo:
Fixes #5463 

This ended up being a significant rework of AsyncDataloader. Basically the goal was to remove all direct `::Fiber` uses, instead relying on `Async::...` primitives. It was a long journey to get async control flow right but it turned out pretty elegant. 

TODO: 

- [x] Unify the flow for running jobs tasks and running sources tasks. These are almost identical. 
- [x] Implement fiber limit, pass that test


rmosolgo:
Sad, it seems like the test suite hangs now...

rmosolgo:
Ok, I think the problem was that Fiber / Task nesting. I reworked it so that each pass over the set of pending dataloader fibers spins up a new job. This avoids any mixing between Async primitives and plain Ruby Fibers. It messed up some of the fiber counting tests though -- I'll sort them out in the morning. 

@iyotov-havelock, could you try this branch in your app's test suite and development? You can bundle it with: 

```ruby
gem "graphql", github: "rmosolgo/graphql-ruby", ref: "async-dataloader-deadlock-fix"
```

please let me know how it goes!

iyotov-havelock:
Hey @rmosolgo, thanks for checking this issue out. 

Tested it with my 'real' app in development, sadly there's still an issue there:
err


Tested it with the demo app (https://github.com/iyotov-havelock/async-dataloader-issue), the `No live threads left. Deadlock?` error is no longer showing, but the GraphQL response is `{"errors":[{"message":"undefined method 'nfields' for nil"}]}`, which does not show up when using `GraphQL::Dataloader` instead of `GraphQL::Dataloader::AsyncDataloader`

rmosolgo:
I saw that error, too, and assumed that GraphQL was "working," since it called code as expected. Inspecting the dataset, I see that it was passed the right values: 

```
#<Sequel::Dataset::_Subclass: "SELECT * FROM \"foo_models\" WHERE (\"id\" IN (1, 2))">
```

My hunch is that Async + Sequel aren't playing nice here ... but do you have a reason to think that AsyncDataloader in particular is the cause of the issue?

Could you please share the full stack trace (**NOT** a screenshot) from the error in your application? Maybe some lines there will be a clue. 

Also, if you can modify the replication app so that it deadlocks on this branch, I'd be happy to keep debugging on that.


________

**UPDATE:** 

I got the example to run successfully by adding the [fiber_concurrency extension](https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/fiber_concurrency_rb.html): 

```diff
diff --git a/config/database.rb b/config/database.rb
index 16f3b03..7ec45a8 100644
--- a/config/database.rb
+++ b/config/database.rb
@@ -1,4 +1,5 @@
 require 'sequel'
+Sequel.extension :fiber_concurrency

 DB = Sequel.connect(
   adapter: 'postgres',
```

Have you tried that extension in your app? I found it by searching randomly around the Sequel repo for the word "Fiber"... 


With that addition, I get:

```
curl -X POST http://localhost:9292/graphql \
  -H "Content-Type: application/json" \
  -d '{"query": "{ getAllFoos { id name barModels { id name } } getAllBars { id name fooModel { id name } } }"}'

{"data":{"getAllFoos":[{"id":"1","name":"First Foo","barModels":[{"id":"1","name":"Bar 1 for Foo 1"},{"id":"2","name":"Bar 2 for Foo 1"},{"id":"3","name":"Bar 3 for Foo 1"}]},{"id":"2","name":"Second Foo","barModels":[{"id":"4","name":"Bar 1 for Foo 2"},{"id":"5","name":"Bar 2 for Foo 2"}]}],"getAllBars":[{"id":"1","name":"Bar 1 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"2","name":"Bar 2 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"3","name":"Bar 3 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"4","name":"Bar 1 for Foo 2","fooModel":{"id":"2","name":"Second Foo"}},{"id":"5","name":"Bar 2 for Foo 2","fooModel":{"id":"2","name":"Second Foo"}}]}}
```

iyotov-havelock:
@rmosolgo I modified the replication app, check out this commit:
https://github.com/iyotov-havelock/async-dataloader-issue/commit/6af81bf63c3ec5b2eafee140f0c07ab4a5c59333

To reproduce the issue, do the `getAllFoos` query again. It's quite possible to be an Async + Sequel issue

rmosolgo:
Interesting -- a change like that puts `Async` primitives back inside a plain Ruby `Fiber.new { ... }`, further suggesting that's the origin of the issues here. It might be possible to rewrite that part of AsyncDataloader to use Async primitives entirely, rather than plain Ruby fibers, but it was really hard to do last time so I gave up. I'll try again...

rmosolgo:
👋  I made time to dig back into this and refactor AsyncDataloader to use Async primitives for execution jobs too. After that, I pulled your updated example and updated the `Gemfile` to use this version of graphql-ruby. Then, when I `curl`'d, I got a correct response! 

```
➜  ~ curl -X POST http://localhost:9292/graphql \
  -H "Content-Type: application/json" \
  -d '{"query": "{ getAllFoos { id name barModels { id name } } getAllBars { id name fooModel { id name } } }"}'
{"data":{"getAllFoos":[{"id":"1","name":"First Foo","barModels":[{"id":"1","name":"Bar 1 for Foo 1"},{"id":"2","name":"Bar 2 for Foo 1"},{"id":"3","name":"Bar 3 for Foo 1"}]},{"id":"2","name":"Second Foo","barModels":[{"id":"4","name":"Bar 1 for Foo 2"},{"id":"5","name":"Bar 2 for Foo 2"}]}],"getAllBars":[{"id":"1","name":"Bar 1 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"2","name":"Bar 2 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"3","name":"Bar 3 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"4","name":"Bar 1 for Foo 2","fooModel":{"id":"2","name":"Second Foo"}},{"id":"5","name":"Bar 2 for Foo 2","fooModel":{"id":"2","name":"Second Foo"}}]}}%
```

Could you give this branch a try in your app when you get a try please?

rmosolgo:
(The handoff between sources and execution isn't _quite_ right here yet ...)

Edit: I think the problem is that `Dataloader#run_isolated` can't be isolated in this case because Async uses a single (per- Thread maybe?) `Reactor`. Calling `yield` from inside `#run_isolated` may pass control to a task from the surrounding `#run`.

iyotov-havelock:
> 👋 I made time to dig back into this and refactor AsyncDataloader to use Async primitives for execution jobs too. After that, I pulled your updated example and updated the `Gemfile` to use this version of graphql-ruby. Then, when I `curl`'d, I got a correct response!
> 
> ```
> ➜  ~ curl -X POST http://localhost:9292/graphql \
>   -H "Content-Type: application/json" \
>   -d '{"query": "{ getAllFoos { id name barModels { id name } } getAllBars { id name fooModel { id name } } }"}'
> {"data":{"getAllFoos":[{"id":"1","name":"First Foo","barModels":[{"id":"1","name":"Bar 1 for Foo 1"},{"id":"2","name":"Bar 2 for Foo 1"},{"id":"3","name":"Bar 3 for Foo 1"}]},{"id":"2","name":"Second Foo","barModels":[{"id":"4","name":"Bar 1 for Foo 2"},{"id":"5","name":"Bar 2 for Foo 2"}]}],"getAllBars":[{"id":"1","name":"Bar 1 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"2","name":"Bar 2 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"3","name":"Bar 3 for Foo 1","fooModel":{"id":"1","name":"First Foo"}},{"id":"4","name":"Bar 1 for Foo 2","fooModel":{"id":"2","name":"Second Foo"}},{"id":"5","name":"Bar 2 for Foo 2","fooModel":{"id":"2","name":"Second Foo"}}]}}%
> ```
> 
> Could you give this branch a try in your app when you get a try please?

Hi, thanks for addressing this. App is working fine with the branch :+1:

rmosolgo:
Awesome, glad to hear it. Today I'm going to try to address #5462 which I think will also address the failing specs here. But it sounds like migrating all usage from Fiber to Async::Task did the job 🎉

rmosolgo:
👋  I am picking this up after a bit of a rabbit trail. I have a new, re-written execution module which I _expect_ to make this easier to fix. 

But I'm running into the same problem where I have a call to `task.yield` but Async doesn't necessarily run _that_ task; instead, it calls any random task, including the task that called `task.yield`. 

I have this feeling that I need to refactor the async control flow but I haven't figured it out yet.

rmosolgo:
I've been thinking about this some more, and digging through the Async docs looking for inspiration. I think I have an idea which I'll try out in the next few days: 

- When creating async tasks, put them in a "working" list 
- The manager fiber waits for "working" list to be empty 
  - Tasks must either put themselves on the "waiting" list 
  - Or remove themselves from the "working" list (if the jobs queue is empty) 
- When code calls `dataloader.yield`:
  - it moves `Task.current` from a "working" list to a "waiting" list 
  - it waits for an Async::Condition to be triggered
- Then, after Dataloader sources have been run, it can rerun "waiting" tasks by: 
  - moving all "waiting" tasks to the "working" list 
  - triggering the condition so that they restart

rmosolgo:
Ok, I think I finally worked out the actual async control flow here. Things are passing and the two job flows basically match. I still need to implement `fiber_limit: ...`.

thiennghv0201-freec:
any news here?

rmosolgo:
This branch is completely working, with a rebuilt Async-based control flow, except for `fiber_limit: ...`. Let me know how it goes if you give it a try in your app!

rmosolgo:
Ok, I finally got `fiber_limit:` working. It's not as precise when using `Async`, though, because Async spins up its own fibers under the hood. So `fiber_limit: ...` applies to GraphQL-Ruby's _own_ fibers, but not to some extras created (somewhere?) by Async. 

If we needed a better understanding of this, we could patch `Fiber#initialize` to log its caller.

rmosolgo:
@iyotov-havelock and @thiennghv0201-freec, if you give this a try on master (or 2.6.4+) and run into more trouble, please open a new issue describing the problem!

PR #5480, opened 2025-12-08
Pass `resolved_type` into `legacy_invalid_empty_selections_on_union`

ravangen:
Objective: Allow for `legacy_invalid_empty_selections_on_union` hook to know which type had an empty selection set. This way we can log and report to the client what the offending part of the operation was.

This is backwards incompatible change, so I'm not sure how you feel about it. I imagine this API isn't widely used yet.

This functionality was introduced in https://github.com/rmosolgo/graphql-ruby/pull/5322 (`2.5.3`).

### DIFF
--- a/lib/graphql/schema.rb
+++ b/lib/graphql/schema.rb
@@ -1708,7 +1708,7 @@ module GraphQL
       # If you need to support previous, non-spec behavior which allowed selecting union fields
       # but *not* selecting any fields on that union, set this to `true` to continue allowing that behavior.
       #
-      # If this is `true`, then {.legacy_invalid_empty_selections_on_union} will be called with {Query} objects
+      # If this is `true`, then {.legacy_invalid_empty_selections_on_union_with_type} will be called with {Query} objects
       # with that kind of selections. You must implement that method
       # @param new_value [Boolean]
       # @return [true, false, nil]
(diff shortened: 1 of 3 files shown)


rmosolgo:
I'm definitely game to support this but I'd like to not :boom: people who have already implemented it. How about accomodating that case somehow? A few possibilities come to mind: 

- Check arity before calling it and warn if it only receives one argument (then only pass one)
- Implement `GraphQL::Schema.method_added` to handle this method and crash if it's defined with only  one argument. That would at least move the :boom: to development 
- Add a wrapper method, eg `..._with_type`, call that method from runtime code instead, and implement that method to call the legacy hook by default (if defined?). Also update docs and warnings to name the new method instead of the old one 

If you have another suggestion on how to avoid disruption, I'm certainly open to it -- I'm just offering those in case they're helpful. What do you think?

ravangen:
I went with the wrapper method approach as it felt simplest.

Validation rule calls `legacy_invalid_empty_selections_on_union_with_type` which by default calls `legacy_invalid_empty_selections_on_union`.

The doc strings advocate for overriding `legacy_invalid_empty_selections_on_union_with_type`, but the original still works.

rmosolgo:
Thanks fort this improvement!

rmosolgo:
🚢  in 2.5.15!

PR #5491, opened 2026-01-06
Improve documentation for `field` and `argument` DSL methods

rmosolgo:
I was hoping that this would play nicer with ruby-lsp. It's _better_ but still not amazing: 

| Before | After |
|--------|--------|
| Screenshot 2026-01-06 at 7 21 13 AM | Screenshot 2026-01-06 at 9 26 25 AM |

It's a wall of text, but the docs are _there_, and if you click to jump to definition, you can read them in original formatting in the gem source.

- [x] Copy docs from `Field#initialize` to `HasFields#field`, add tests
- [x] Copy docs from `Argument#initialize` to `HasArguments#argument`, add tests
- [x] Check performance impact on schema boot -- this could change how Hashes are allocated during schema definition 
- [x] Consider removing `Field.from_options`? 

Fixes #3543 ... almost 5 years later 😞 

I'm also interested in exploring other documentation methods if they would play nicer with LSP support.

### DIFF
--- a/lib/graphql/schema/member/has_arguments.rb
+++ b/lib/graphql/schema/member/has_arguments.rb
@@ -17,28 +17,28 @@ module GraphQL
         # @param arg_name [Symbol] The underscore-cased name of this argument, `name:` keyword also accepted
         # @param type_expr The GraphQL type of this argument; `type:` keyword also accepted
         # @param desc [String] Argument description, `description:` keyword also accepted
-        # @param required [Boolean, :nullable] if true, this argument is non-null; if false, this argument is nullable. If `:nullable`, then the argument must be provided, though it may be `null`.
-        # @param description [String] Positional argument also accepted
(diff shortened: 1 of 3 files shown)


rmosolgo:
Using `be rake bench:profile_boot`, it looks like the code currently allocates way more memory, but way less objects: 

```diff
- Total allocated: 8242456 bytes (51658 objects)
+ Total allocated: 12990408 bytes (44849 objects)
```

My guess is that this is Ruby handling all these big keyword argument lists ... time to investigate!

rmosolgo:
I extracted the `Field.from_options` change into another PR and reverted my changes to the keyword parameters of these methods because of the memory overhead. But I kept the docs and tests on the docs. In the end, it turned out to use less memory, probably because of removing `*args`:

```diff
- Total allocated: 7770688 bytes (46971 objects)
+ Total allocated: 7685808 bytes (44849 objects)
```

PR #5494, opened 2026-01-07
Optimize INVALID_NUMBER_FOLLOWED_BY_NAME_REGEXP

swalkinshaw:
Claude Code helped to optimize this potentially slow regexp depending on the document. Specifically, the backtracking in the original regexp was the main performance issue.

This does a few things:

1. adds a very simple and faster pre-check to gate if the main regexp should even run
2. prevents as much backtracking as possible

Benchmarks:

| Document Type                  | Before | After  | Speedup |
|--------------------------------|--------|--------|---------|
| Typical large (125KB)          | 0.72s  | 0.06s  | 12x     |
| Colon-heavy (35KB)             | 0.24s  | 0.007s | 34x     |
| Pathological worst-case (26KB) | 1.64s  | 0.25s  | 6.6x    |
| No digits (17KB)               | 0.10s  | 0.002s | 54x     |

### DIFF
--- a/lib/graphql/language.rb
+++ b/lib/graphql/language.rb
@@ -79,22 +79,28 @@ module GraphQL
 
     LEADING_REGEX = Regexp.union(" ", *Lexer::Punctuation.constants.map { |const| Lexer::Punctuation.const_get(const) })
 
+    # Optimized pattern using:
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thank you! TIL about `*+`.

PR #5496, opened 2026-01-12
Fix the link to The Query, Mutation, and Subscription types documentation in the Getting Started guide

duffuniverse:
### Detail
- Fix the link to root operation types documentation in the Getting Started guide


rmosolgo:
Thanks!

PR #5497, opened 2026-01-13
Fix ERB issues surfaced by Herb

joelhawksley:
As part of working to integrate [Herb](https://github.com/marcoroth/herb), I ran into some issues with graphql-ruby when enabling Herb for all ERB and HTML in our application.

This PR fixes the majority of the issues raised by the linter.

I've filed a [bug](https://github.com/marcoroth/herb/issues/1060) for a case that I think shouldn't be raised.

cc @marcoroth

### DIFF
--- a/guides/_layouts/default.html
+++ b/guides/_layouts/default.html
@@ -8,7 +8,7 @@
     {% else %}
       <title>GraphQL - {{ page.title }}</title>
     {% endif %}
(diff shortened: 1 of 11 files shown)


rmosolgo:
Awesome, thank you! If you'd like me to keep up compliance here, I'd also welcome a PR (or some advice about) integrating Herb into the gem's CI.

joelhawksley:
@rmosolgo you're welcome! I'll have a PR to use Herb in CI for you once it's a bit more mature ❤️

PR #5498, opened 2026-01-13
Add spec for better loads: dataloading, prototype new loads code

### LINKED ISSUE #5462: Add the benefit of dataloaders for mutation arguments

rmosolgo:
Implementing `loads:` has had a smell in it: 

https://github.com/rmosolgo/graphql-ruby/blob/c4f07cdbf177c1da4d34b691b8c1003f9bef296f/lib/graphql/schema/argument.rb#L338

https://github.com/rmosolgo/graphql-ruby/blob/c4f07cdbf177c1da4d34b691b8c1003f9bef296f/lib/graphql/schema/argument.rb#L359

It's there because some code requires the argument values to be returned right away -- it doesn't use Dataloader (properly?). 

This smell manifests the problem described in #5462 -- I added a spec for that issue in cf3a3f8b53d75b4a04f6db1f21517afaf38c3eeb

This problem also surfaced in #5479 because, when using `Async`, `run_isolated` isn't actually isolated: Async uses a top-level pool, so a pause in one place in the code can resume code elsewhere. 

The smell was also called out in https://github.com/rmosolgo/graphql-ruby/pull/5389#discussion_r2317112688. 

When working on #5389, I didn't actually rework this code. But I think it's actually a good place to try out that new flow in an incremental way. 

Fixes #5462 

TODO: 

- [ ] Remove another Argument-related `run_isolated` call 
- [ ] What about the mutation-related `run_isolated` call? 
- [ ] Find a better solution than `while ...; dataloader.yield; end` -- some proper way for expressing or communicating dependencies in the code
- [ ] Consider rooting out `after_lazy` usage in loading and authorization
- [ ] Benchmark

### DIFF
--- a/lib/graphql/execution/interpreter/arguments.rb
+++ b/lib/graphql/execution/interpreter/arguments.rb
@@ -21,9 +21,10 @@ module GraphQL
 
         # @param argument_values [nil, Hash{Symbol => ArgumentValue}]
         # @param keyword_arguments [nil, Hash{Symbol => Object}]
-        def initialize(keyword_arguments: nil, parent_object: nil, context: nil, argument_values:)
+        def initialize(keyword_arguments: nil, owner: nil, parent_object: nil, context: nil, argument_values:)
           @empty = argument_values.nil? || argument_values.empty?
           @context = context
+          @owner = owner
           @parent_object = parent_object
           # This is only present when `extras` have been merged in:
           @keyword_arguments = if keyword_arguments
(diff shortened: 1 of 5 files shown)


rmosolgo:
I am rewriting execution from scratch instead -- no `run_isolated` 😅

Issue #5500, opened 2026-01-15
GraphQL Dashboard Engine Breaks Devise Authentication in Production


francescoparra:
**Describe the bug**

The GraphQL Dashboard feature introduced in v2.4.11 causes Devise authentication to fail in Rails applications running in production mode with eager loading enabled. The dashboard engine draws routes before the main application's config/routes.rb, causing Devise to cache an empty mapping, which breaks all authentication.

**Versions**

`graphql` version: version: 2.4.11 (issue introduced in this version)
`rails` (or other framework): 7.2.3
`devise`: 4.9.4

**Steps to reproduce**

This issue only occurs in production mode with eager loading enabled (`config.eager_load = true`). It does not occur in development or test environments.

1. Use a Rails application with Devise for authentication
2. Upgrade graphql gem from 2.4.10 to 2.4.11 or later
3. Run the application in production mode
4. Attempt to authenticate a user via Devise

**Expected behavior**

Devise authentication should work normally, as it does in version 2.4.10 and in development/test environments.

**Actual behavior**

Devise authentication fails because its mapping is empty. The root cause:

1. The GraphQL Dashboard engine calls routes.draw during initialization (see [lib/graphql/dashboard.rb#L38](https://github.com/rmosolgo/graphql-ruby/blob/v2.4.11/lib/graphql/dashboard.rb#L38))
2. This happens before the main application's config/routes.rb is loaded
3. When routes.draw is called, Devise's route finalization hook runs and caches its mappings
4. Since the main application routes haven't been drawn yet, Devise caches an empty mapping
5. When the application later tries to use Devise, the cached mapping is empty, causing failures

**Additional context**

This is a Rails engine initialization order issue. The problem occurs because:

- In production with eager loading, all engines initialize immediately at boot time
- The GraphQL Dashboard engine draws its routes during class loading
- This triggers Devise's route finalization callbacks before the main app's routes are loaded
- Devise caches an empty mapping and never updates it

rmosolgo:
Hey, thanks for the detailed report and sorry for this compatibility issue! I'll take a look at how the engine adds its routes and follow up here.

rmosolgo:
Hey @francescoparra, I see you took a crack at it over in #5504. The gem's own tests for the dashboard don't seem to like that setup, but did it work for your app? 

A couple other options I found while digging into the docs and source: 

[...]

PR #5504, opened 2026-01-18
Fix routes premature drawing

francescoparra:
This PR fixes https://github.com/rmosolgo/graphql-ruby/issues/5500 by lazy loading the drawing of the Dashboard routes


rmosolgo:
Thanks for investigating this -- going forward with #5505 instead.

PR #5505, opened 2026-01-19
Lazily define engine routes using routes do ... end

### LINKED ISSUE #5500: GraphQL Dashboard Engine Breaks Devise Authentication in Production

rmosolgo:
Fixes #5500 

At least, I think it _should_ fix that issue, but my tests fail with: 

```
NameError (undefined local variable or method 'graphql_dashboard' for an instance of #<Class:0x000000012a0324f8>)
```

which is raised from the `.html.erb` views code. `graphql_dashboard.<some_view_path>` works fine in the test files, but not in the views 😖 

I have been trying to trace through initializer code and RouteSet code to figure out what's missing, but I haven't been able 
to.

UPDATE: Trying to debug, I tried calling the route methods in the engine's controller methods, ie `def show; puts graphql_dashboard.delete_all_detailed_traces_path; end`. It raised a more helpful error: 

```
In order to use #url_for, you must include routing helpers explicitly. For instance, `include Rails.application.routes.url_helpers`
```

And indeed, including the url helpers (the engine's url helpers in this case) _did_ fix my problem. In order to keep the routes lazy-loaded, I put the `.include` call inside the `routes do ... end` block.

### DIFF
--- a/lib/graphql/dashboard.rb
+++ b/lib/graphql/dashboard.rb
@@ -76,6 +76,8 @@ module Graphql
         resources :subscriptions, only: [:show], constraints: { id: /[a-zA-Z0-9\-]+/ }
         post "/subscriptions/clear_all", to: "subscriptions#clear_all", as: :clear_all
       end
+
+      ApplicationController.include(Dashboard.routes.url_helpers)
     end
 
     class ApplicationController < ActionController::Base


rmosolgo:
👋  @francescoparra could you try this branch in your app? 

```ruby 
gem "graphql", github: "rmosolgo/graphql-ruby", ref: "f903cb57b054ee4ed40600e032ad43f03ed958d1"
```

It moves the routes into lazily-evaluated block and adds a manual `.include` call to keep things working properly. Let me know how it goes if you give it a try!

francescoparra:
@rmosolgo thanks for your fast answers, the branch that you gave me works smoothly!

PR #5506, opened 2026-01-20
Fix stale documentation link in the OperationStore Sync guide

duffuniverse:
### Detail
Fix the Relay's Network Layer documentation link


rmosolgo:
Thanks for another improvement here! Link rot is real...

PR #5508, opened 2026-01-21
Fix internal link reference for payload compression in the Apollo Subscriptions guide

duffuniverse:
### Detail

Fix internal link reference for payload compression in the Apollo Subscriptions guide


rmosolgo:
Thank you! My build step makes sure that internal links' target _pages_ exist, but it doesn't check _headings_ 😖

Issue #5507, opened 2026-01-20
Write a new execution module for GraphQL-Ruby


rmosolgo:
I have been spending a lot of time trying to move the existing execution code forward in a compatible way, but I'm finding it impossible :S  I mean at least these modules:`GraphQL::Execution::Interpreter`, `Interpreter::Runtime`, `Interpreter::Arguments` (and `HasArguments`, `ArgumentValue`), `Dataloader`, `Execution::Lazy`. 

Writing up all the details is possible but in the interest of getting started, I'll be brief: 

- Performance: A lot of possible performance gains have been identified (#5502, #5446, #4426) but are stuck in compatibility land for edge cases 
- Correctness: GraphQL-Ruby has received a lot of features over the years, tacked on in different ways. They don't always work right _together_... it's worth considering which ones can be jettisoned for other approaches (field `extras`?) or merged together, at least under the hood (`prepare:`, `loads:`, `validates:`?).
- Features: actually, performance is the feature that I'd want from this. I expect _most_ features could be reimplemented one way or another -- or at least a very clear migration path -- with a new core runtime. 

So, I'm proposing a project like: 

- Build a new runtime in isolation (takes the same inputs as `Schema.run`, returns the same output; can be used on a query-by-query basis for Scientist-like checks)
- Make sure the new runtime checks the necessary boxes (see below)
- One at a time, reimplement existing features to the degree possible 
- Also build a "GraphQL Doctor" which inspects source code and schemas in memory and outputs the result of compatibility checks and links to migration notes 

After trying and trying, I don't think a future execution paradigm can be achieved with a perfectly smooth path, but I think it's worth pursing -- Shopify has been using custom execution code to great success for a while now. 

__New Execution Goals__ in no specific order:

- Support first-class batching ("batch resolvers"), so you can get the advantages of dataloader or graphql-batch without promises or fibers, using a breadth-first execution algorithm (#5446)
- Non-recursive, non-block-based execution algorithm to reduce overhead (#5389)
- Properly cache merged AST nodes (#5502)
- Others? 

cc @gmac since you've been investigating a lot of these too -- what do you think this approach to improving GraphQL-Ruby?

gmac:
[...]

PR #5509, opened 2026-01-21
New Execution Module

rmosolgo:
> [!TIP]
>  **Latest Docs: https://graphql-ruby.org/execution/next.html**

Taking a crack at #5507 

Much inspiration taken from https://github.com/gmac/graphql-breadth-exec, as well as #5389 and the existing `Interpreter::Runtime` module.

I'm especially indebted to @gmac's work in https://github.com/gmac/graphql-breadth-exec for: 

- __The overall batched flow__ where list indices are ignored from execution's standpoint. I think I will run into trouble with Lazy/Dataloader support but it's brilliant so far.
- __Propagating nulls during a second pass over the result__ only when some errors are present. This is awesome because it reduces the necessary bookkeeping at runtime and moves the overhead to the cases where an error is present (presumably less common and less important than the happy path). One downside here is that execution no longer stops on branches with errors. This will be documented as a "breaking" change. 
- __Using `gather_selections` to isolate mutations__, it's perfect but I think it would have taken a while to click for me.

I plan to iterate on this in "draft" state until it's basically implementing the GraphQL spec, then merge it and continue working on it. 

Known TODOs: 

- [x] Fix this failing test: `SEED="61509" GRAPHQL_FUTURE=1 be rake test TEST=spec/graphql/execution/lazy_spec.rb`
- [x] Review tests & code to make sure changes are isolated to the new module 
  - [x]  Do I still need `Argument#original_keyword`? See if that can be removed
- [x] Revert any temporary changes to tests 
- ~~Review opt-in mechanisms and unify them somewhat ?~~ I will do this later
- [x] Fragment spreads
- [x] Abstract-typed inline fragments 
- [x] Hook up static validation 
- [x] static input coercion 
- [x] leaf result coercion 
- [x] non-null enforcement 
- [x] variables
- [x] input objects
- [x] `@skip`/`@include`
- [x] Introspection
- [x] Mutation isolation
- [x] Automatic connection argument handling somehow
- [x] Resolver `loads:` compatiblity


rmosolgo:
Resolving lazies (eg `Promise`) and Dataloader are basically working, although non-return-value hooks (like `loads:` and `authorized?`) don't batch correctly yet. I'm going to do some renaming and clean-up, then get the test suite into a clean and passing state, then merge.

rmosolgo:
At this point my own little greenfield app _runs_ with this code but connection values aren't being passed along in some cases. I'm going to try to get this passing CI (with a few test files using this new runtime) then merge it into master for more experimentation. 

Before I can ship this in my app, I think it needs at least two APIs:

- Something like `def self.resolve_all_#{field_name}` (currently hacked into `batching_spec.rb`) to write custom batch resolvers
- Something like `field ..., dataload: { association:, source:, with:, }` for automatically calling Dataloader sources. This could be made to work with both old and new runtimes. 
- Some kind of support for connection fields using batch resolvers, there's currently nothing 

Before publicly releasing it, I need to migrate more tests to use `execute_batching` when `TESTING_BATCHING`, and create a detailed compatibility/migration document.

gmac:
> One downside here is that execution no longer stops on branches with errors. This will be documented as a "breaking" change.

Yup, runtime null propagation is a big tradeoff, but at the rate we error for non-validation errors – having full execution time for those requests is a modest tradeoff. It wasn't in the initial prototype, but we've honed the breadth abort logic to this, which is about as aggressive as I think you can get in the breadth paradigm:

```ruby
# Walk the tree to determine the highest contiguous non-null depth, and the highest list depth.
# We can ONLY abort breadth resolvers when one of the following happens:
# 1. null propagation reaches the root scope (total loss)
# 2. there are no lists in the tree (no objects will share breadth resolvers)
# 3. all lists in the tree are invalidated (all objects sharing a resolver are eliminated)
field = exec_field #: ExecutionField?
propagating = true #: as bool
highest_nulled_depth = exec_field.scope.depth
highest_list_depth = nil #: Integer?

while field
  if field.propagates_null? && propagating
    highest_nulled_depth = field.scope.depth
  else
    propagating = false
  end

  if field.type.list?
    highest_list_depth = field.scope.depth
  end

  field = field.scope.parent_field
end

if highest_nulled_depth.zero?
  # Mark the entire executor as aborted when non-null propagation hits the top.
  # This prevents subsequent isolated root scopes (mutations) from running.
  @aborted = true
elsif highest_list_depth.nil? || highest_nulled_depth <= highest_list_depth
  # Abort all non-null ancestor scopes that meet or exceed the highest-level list.
  # (all lists must be completely invalidated, or else remain alive).
  abort_field = exec_field #: ExecutionField?
  while abort_field && highest_nulled_depth <= abort_field.scope.depth
    abort_field.scope.abort!
    abort_field = abort_field.scope.parent_field
  end

  # Purge all aborted work from the queue
  @exec_queue.reject!(&:aborted_subtree?)
end
```

lib/graphql/execution/batching/field_resolve_step.rb
```diff
@@ -0,0 +1,336 @@
...
+        end
+
+        private
+
+        def build_graphql_result(graphql_result, key, field_result, return_type, is_nn, is_list, is_from_array) # rubocop:disable Metrics/ParameterLists
+          if field_result.nil?
+            if is_nn
+              graphql_result[key] = @runner.add_non_null_error(@parent_type, @field_definition, @ast_node, is_from_array, path)
+            else
+              graphql_result[key] = nil
```
gmac: How are you maintaining key order across lazy executions? Are you executing everything first, stashing the result, and then building all fields synchronously at the end?

lib/graphql/execution/batching/field_resolve_step.rb
```diff
@@ -0,0 +1,336 @@
...
+            @ast_node = ast_node
+          elsif @ast_nodes.nil?
+            @ast_nodes = [@ast_node, ast_node]
+          else
+            @ast_nodes << ast_node
+          end
+          nil
+        end
+
+        def coerce_arguments(argument_owner, ast_arguments_or_hash)
```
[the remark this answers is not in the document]
gmac: We ended up doing everything as a single flat pass, and migrated the GraphQL JS algorithms directly. We wanted to keep GraphQL Ruby input validators in a static capacity, so we now do a single pass on variables coercing and validating all that input, then a single coercion/validation pass on each argument set, splicing in their variables and the corresponding validation errors. Coercion errors raise statically, while validation errors only raise on execution using that input.

gmac:
> The overall batched flow where list indices are ignored from execution's standpoint. I think I will run into trouble with Lazy/Dataloader support but it's brilliant so far

Breadth _can_ do paths fairly efficiently if you index as you go. I've got a prototype floating around that builds the indices on-demand, and it's still pretty fast (though considerable slower than without them). Basically you just build a multi-layer index:

```ruby
Scope {
   @indices = [
     [0, 0, 1, 1], # maps object in this scope to the one above it
     [0, 1, 2, 3], # maps positions in a first-order list
     [i, i, i, i]  # maps positions in a second-order list, etc...
   ]
}
```

With that done, you can alway map breadth objects by going to their position within their scope, reading that index position from bottom-up, insert the parent field key and transition to the parent scope indexing, then follow that one up and repeat.

We've never needed the indices badly enough to commit to adding them. If we do support a feature like `@defer` that needs runtime paths, we'd still probably only add them as an on-demand computation. We can do execution without paths, and it avoids developer temptations to do strange things based on tree order.

lib/graphql/execution/batching/field_resolve_step.rb
rmosolgo: Thanks for the tip -- I'll give it a try!

rmosolgo:
>  index as you go

I read this last night and thought, "I'll try to understand it tomorrow morning after a cup of coffee." Now, here I am, and I still can't grok it 😆 

Let me see ... `@indices` in the pseudo-code contains two kinds of entries: 

- The first array links an item in the scope's _own_ objects to an item in the _parent_ scope's objects. It may contain duplicates if the object in the parent scope returned a list of items; in that case, multiple child objects have the same parent. 
    
    The first array is always present.

- If a second array is present, it means that objects in _this_ scope came from a list result of a parent object, and they indicate _where, in the parent list_, this object came from. IIUC, this second array may contain duplicates, for example: 
    
    ```ruby 
    @objects 
    # [a, b, c, d, e, f]  <- given these 6 objects...
    @indices 
    # [
    #   [0, 0, 0, 1, 1, 1], <- from two parent objects, [obj0, obj1], each one returned 3 non-nil children
    #   [0, 1, 2, 0, 2, 3] <- obj0 returned [a, b, c]; obj1 returned [d, nil, e, f]; the nil was dropped
    # ]
    ```

- If a third array is present, it means that the current objects were members of nested arrays. The third array indicates where the object occurred in the nested array. The second array indicates where the nested array occurred in the top-level array. For example: 

    ```ruby 
    @objects 
    # [a, b, c, d, e, f]  <- given these 6 objects...
    @indices 
    # [
    #   [0, 0, 0, 1, 1, 1], <- from two parent objects, [obj0, obj1], each one returned 3 non-nil children
    #   [0, 0, 1, 0, 1, 1] 
    #   [0, 1, 0, 0, 0, 1]<- obj0 returned [[a, b], [c]]; obj1 returned [[d, nil], [e, f]]; the nil was dropped
    # ]
    ```    
    
Ok... did I get it? 

Brilliant. You basically add 1 array per result set, minimum, plus an array per list depth. And it gives you what you need in order to build `context[:current_path]` on the fly.  I'm not going to implement it on this branch but I will try -- preserving `current_path` would be awesome.

rmosolgo:
> breadth abort logic

Thanks for sharing this. I think there are specs which test aborting and when I break them, I will dig back into it!

lib/graphql/execution/batching/field_resolve_step.rb
```diff
@@ -0,0 +1,336 @@
...
+              if @runner.schema.lazy?(field_result)
+                lazies = true
+                break
+              elsif field_result.is_a?(Array)
+                field_result.each do |inner_fr|
+                  if @runner.schema.lazy?(inner_fr)
+                    break lazies = true
+                  end
+                end
+                if lazies
```
gmac: So if I'm reading this sequence correctly, this is leaving money on the table with an implementation that still follows a depth-based mindset and overlooks one of breadth's superpowers. Basically – lazies/promises are expensive. They have both allocation and intermediary resolution cost; and here we seem to be following the depth pattern where we resolve a "batch" of promises, then still resort to resolving them each individually. In the breadth model, we can build and resolve lazys at a many-to-one standard rather than a one-to-one standard:

Screenshot 2026-01-31 at 11 01 20 AM

Following the above model, it looks like you've now consolidated all the _resolver_ calls down into one (which is something), but you still end up with a slew of lazies that need to be resolved... enter good old depth-based slowness (which I've benchmarked at 20x slower for 100 promises vs 1 using GraphQL::Batch. Promises seem to scale particularly badly as set sizes grow.). So, breadth's superpower is to bind ALL the work to a single intermediary lazy that resolves the set later.

lib/graphql/execution/batching/field_resolve_step.rb
gmac: Oh, one other tip: the empty placeholder keys should use some kind of `UNDEFINED` value. This is important in the error formatter pass where we need to distinguish between legitimately invalid nil values that bubble versus branches that weren't executed due to abort and therefore just get ignored.

gmac:
> did I get it?

Yep! You got it, I think. This is prototype code, we've never actually used it; we don't think we need `current_path` (our whole stack now runs without it!), and it's not worth the overhead to build inline indices for something with no legitimate internal uses. This prototype just indexes a Cardinal scope's subtree lazily on-demand, which is slower than if we'd built indices inline as part of flat-mapping results. Anywho – if it's helpful, here's the algorithm for reference. Might as well establish it as part of the OSS paradigm.

```ruby
#  Builds paths to specific breadth object positions as exact object paths, ex:
#  - Scope path (namespace): ["products", "variants"]
#  - Object path (exact path): ["products", 0, "variants", 1]
#  Object pathing assembles breadth indices for all scopes ascending from the targeted scope, ex:
#  {
#    <TargetExecScope> => {
#      0 => [0, 1, 1, 2], # << zero index always maps objects in this scope to objects in the parent scope
#      1 => [0, 0, 1, 0], # << maps a first-order list in this scope
#      2 => [ ... ], # << maps a second-order list in this scope, etc...
#    },
#    <ParentExecScope> => { ... },
#  }
#  All numeric indexing arrays have size matching the number of objects in their scope.
#  To read an object path, we follow the Nth position of each scope's index map ascending up the scope tree.
#  Building these indices adds execution overhead that isn't needed regularly, so this is only done as part of deferred work.
class ObjectPathBuilder
  #: Hash[Executor::ExecutionScope, Hash[Integer, Array[Integer]]]
  attr_reader :indices_by_scope

  def initialize
    @indices_by_scope = Hash.new { |h1, scope| h1[scope] = Hash.new { |h2, index| h2[index] = [] } }.compare_by_identity
  end

  #: (Executor::ExecutionScope, Integer) -> error_path
  def object_path(exec_scope, index)
    current_path = []

    scope = exec_scope #: Executor::ExecutionScope?
    breadth_index = index
    while scope
      # index the scope unless it has already been done
      scope_indices = @indices_by_scope[scope]
      index_scope(scope, scope_indices) if scope_indices.empty?

      # loop backward through all the scope's indices...
      # - all scopes have at least one index that defines the parent scope position
      # - list scopes have additional indices for each layer of list wrapping
      i = scope_indices.length - 1
      while i >= 0
        if i.zero?
          # at the lowest index, recalibrate for the next highest scope
          breadth_index = scope_indices[i][breadth_index] #: as Integer
        else
          # higher indices add list positions to the current path
          current_path.prepend(scope_indices[i][breadth_index])
        end
        i -= 1
      end

      # before going up a scope, add the parent field key into current path
      key = scope.parent_field&.key
      current_path.prepend(key) if key
      scope = scope.parent
    end

    current_path
  end

  private

  #: (Executor::ExecutionScope, Hash[Integer, Array[Integer]]) -> void
  def index_scope(exec_scope, scope_indices)
    raise ArgumentError, "Scope must not be indexed" unless scope_indices.empty?
    raise ArgumentError, "Scope must be executed" unless exec_scope.executed?

    parent_objects = exec_scope.objects
    current_type = exec_scope.parent_type
    if (parent_field = exec_scope.parent_field)
      parent_objects = parent_field.result #: as !nil
      current_type = parent_field.type
    end

    object_path = []
    i = 0
    while i < parent_objects.length
      object_path[0] = i
      build_indices(current_type, parent_objects[i], object_path, scope_indices)
      i += 1
    end

    scope_indices.each(&:freeze)
    scope_indices.freeze
  end

  #: (singleton(GraphQL::Schema::Member), untyped, Array[Integer], Hash[Integer, Array[Integer]]) -> void
  def build_indices(current_type, object, object_path, next_indices)
    return if object.nil? || object.is_a?(ExecutionError)

    if current_type.list?
      raise ImplementationError, "Expected Array, got #{object.class}" unless object.is_a?(Array)

      current_type = Util.unwrap_non_null(current_type)

      i = 0
      while i < object.length
        object_path << i
        build_indices(current_type.of_type, object[i], object_path, next_indices)
        object_path.pop
        i += 1
      end
    else
      i = 0
      while i < object_path.length
        next_index = object_path[i] #: as !nil
        next_map = next_indices[i] #: as !nil
        next_map << next_index
        i += 1
      end
    end
  end
end
```

lib/graphql/execution/batching/field_resolve_step.rb
gmac: Also, doesn't appear to support multi-dimensional arrays? I'd expect recursion anytime we traverse an array result. But again, maybe I'm misreading what's happening here.

rmosolgo: I definitely follow that a single-promise approach would be the best. But for compatibility's sake, I'd like to continue supporting one-promise-per-object, especially if it can be done without slowing down the code when it's not in use. 

The implementation here adds almost zero overhead when no `lazy_resolve ...` configurations are added, but it _does_ check every object when such a configuration _is_ added. Maybe the compatibility layer could be improved to have two levels: 

- Promises everywhere (legacy): any result may be a Promise, or Array of Promises, or (mentioned elsewhere) an Array of Arrays of Promises, n-deep. 
- Batch resolvers may return promises (new, better, and IIUC, what's described above): fields may return Promise instances. 

Third, possibly even better option: fields returning promises must be flagged in their configuration, otherwise the object will be handled without checking.

Does the compatibility issue make a little sense of this implementation? What do you think about more granular support for Promises like that?

guides/execution/batching.md
```diff
@@ -0,0 +1,278 @@
...
+
+It's much faster and less memory-consuming than the existing execution module, but it doesn't support every feature that the existing module does.
+
+This feature is in heavy development, so if you give it a try and run into any problems, please open an issue on GitHub!
+
+## Background
+
+Batching execution follows the work of [GraphQL-Cardinal](https://github.com/gmac/graphql-breadth-exec), an experimental GraphQL runtime from [@gmac](https://github.com/gmac) and collaborators at Shopify. They wanted better performance from their GraphQL API -- faster responses, less memory use -- and were willing to rethink things at a high level in order to pursue it. At this point, Shopify runs a similar, custom runtime for production traffic and has seen tremendous benefits. Ideally, the same benefits will be available to GraphQL-Ruby users soon. Thanks, `@gmac` and Shopify!
+
+Specifically, Cardinal (and GraphQL-Ruby's batching execution after it) use a few novel techniques for implementing GraphQL:
```
gmac:
```suggestion
Breadth-first execution (or, "execution batching") is an algorithmic paradigm developed by Shopify to address problems of scale when resolving large nested lists. Rather than paying field-level overhead costs (resolver calls, instrumentation, lazy promises) for every field _of every resolved object_, the design instead incurs these costs only once per field selection and runs its breadth of objects with no additional overhead.

The original prototype of Shopify's core algorithm and white paper notes can be found in the [graphql-breadth-exec](https://github.com/gmac/graphql-breadth-exec) project. That prototype matured into Shopify's proprietary _GraphQL Cardinal_ execution engine. While _Cardinal_ is considerably more sophisticated now, its central breadth-first design remains virtually unchanged and runs most of their core traffic.

GraphQL-Ruby brings these breadth-first design principles to the open-source community software, including several novel techniques for implementing GraphQL:
```

guides/execution/batching.md
```diff
@@ -0,0 +1,278 @@
...
+
+- It resolves fields breadth-first instead of depth-first
+- Consequently, it resolves list results in batches, reducing "context switching" between the execution engine and an application's resolvers
+- It optimizes error handling to be finalized in a second pass over the result (when errors are present), speeding up the "happy path"
+
+This approach can produce a GraphQL result _very_ quickly: In a bare-bones example, Cardinal runs about __15x__ faster than GraphQL-Ruby, using __75% less__ memory.
```
gmac:
```suggestion
Breadth-first patterns can produce dramatic results in responses with a high degree of repetition: it's not uncommon to see breadth running __15x__ faster and using __75% less__ memory than classic GraphQL Ruby execution. However, gains are relative...
```

guides/execution/batching.md
```diff
@@ -0,0 +1,278 @@
...
+
+The downside is that many of GraphQL-Ruby's "bonus features" -- beyond the behavior described in the GraphQL Specification -- are either _impossible_ to implement in this paradigm or add non-trivial latency when added back in. So, the task ahead is to "lift the ceiling" of performance in GraphQL-Ruby while retaining as much compatibility as possible and supporting a gradual transition to this new runtime engine.
+
+## Enabling Batching Execution
+
+Batching execution is enabled with two steps:
+
+- Require the code (it's loaded by default): `require "graphql/execution/batching"`
+- Call `MySchema.execute_batching(...)` instead of `MySchema.execute(...)`. It takes the same arguments.
```
gmac: `MySchema.batch_execute` would be a better name, IMO. Or – I would encourage keeping the "breadth" nomenclature. There are so many "batching" utilities out there. Everything batches. Breadth describes the execution paradigm, which itself is considerably more than just delivering on some batching.

guides/execution/batching.md
gmac: I've updated the `graphql-breadth-exec` repo to distance itself from Cardinal (it's now as much Cardinal as your GraphQL Ruby interpretation here is). Going forward, please be mindful of making comparisons against the proof-on-concept rather than the internal engine.

I've also cleaned up the white paper notes in the README to explain the philosophy better.

lib/graphql/execution/batching/runner.rb
```diff
@@ -0,0 +1,270 @@
...
+        def check_object_result(result_h, static_type, ast_selections, current_exec_path, current_result_path, paths_to_check)
+          current_path_len = current_exec_path.length
+          ast_selections.each do |ast_selection|
+            case ast_selection
+            when Language::Nodes::Field
+              begin
+                key = ast_selection.alias || ast_selection.name
+                current_exec_path << key
+                current_result_path << key
+                if paths_to_check.any? { |path_to_check| path_to_check[current_path_len] == key }
```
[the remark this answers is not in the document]
gmac: Actually, here: our final tuned traversal logic that ended up passing all internal tests is this... we had to make a bunch of small tweaks, nuanced to the point that I don't remember them all, and subtle enough that it'll probably take a long time to find all the quirks without a giant monolith to give you feedback. Might as well just get it tuned now. I also adjusted the design so that it could be run repeatedly on partial object subtrees (`@defer` experiment) while maintaining a shared internal cache across runs.

```ruby
class State
  #: Array[Hash[String, untyped]]
  attr_reader :errors

  #: error_path
  attr_reader :actual_path

  #: error_path
  attr_reader :base_path

  #: (?error_path) -> void
  def initialize(base_path = EMPTY_ARRAY)
    @base_path = base_path
    @actual_path = []
    @errors = []
  end

  #: -> error_path
  def current_path
    @base_path + @actual_path
  end
end

#: (
#|   executor: Executor,
#|   invalidated_results: Hash[untyped, ExecutionError],
#|   abstract_result_types: Hash[untyped, singleton(GraphQL::Schema::Object)],
#| ) -> void
def initialize(
  executor:,
  invalidated_results:,
  abstract_result_types:
)
  @executor = executor
  @context = executor.context
  @invalidated_results = invalidated_results
  @abstract_result_types = abstract_result_types
end

#: (
#|   singleton(GraphQL::Schema::Object),
#|   Array[GraphQL::Language::Nodes::AbstractNode],
#|   Hash[String, untyped],
#|   ?Array[String | Integer]
#| ) -> [Hash[String, untyped]?, Array[error_hash]]
def format_object(parent_type, selections, data, base_path = EMPTY_ARRAY)
  return [data, EMPTY_ARRAY] if @invalidated_results.empty?

  state = State.new(base_path)

  # Check if root data is invalidated (either inlined error or marked result)
  if (err = @invalidated_results[data])
    add_formatted_error(err, state)
    return [nil, state.errors]
  end

  data = propagate_object_scope_errors(
    data,
    parent_type,
    selections,
    state,
  )

  [data, state.errors]
end

private

#: (untyped, singleton(GraphQL::Schema::Object), Array[GraphQL::Language::Nodes::AbstractNode], State) -> untyped
def propagate_object_scope_errors(raw_object, parent_type, selections, state)
  return nil if raw_object.nil?

  selections.each do |node|
    case node
    when GraphQL::Language::Nodes::Field
      field_key = node.alias || node.name

      state.actual_path << field_key

      begin
        node_type = @context.types.field(parent_type, node.name).type
        named_type = node_type.unwrap
        raw_value = raw_object.fetch(field_key, Executor::UNDEFINED)

        # Aborted subtrees may have undefined fields that didn't execute.
        # Ignore these rather than considering them invalid by the schema.
        next if raw_value.equal?(Executor::UNDEFINED)

        # Check for invalidated positions (inlined errors or marked results)
        raw_object[field_key] = if (err = @invalidated_results[raw_value])
          add_formatted_error(err, state)
          nil
        elsif node_type.list?
          propagate_list_scope_errors(raw_value, node_type, node.selections, state)
        elsif named_type.kind.leaf?
          raw_value
        else
          propagate_object_scope_errors(raw_value, named_type, node.selections, state)
        end

        return nil if node_type.non_null? && raw_object[field_key].nil?
      ensure
        state.actual_path.pop
      end

    when GraphQL::Language::Nodes::InlineFragment
      fragment_type = node.type ? @context.types.type(node.type.name) : parent_type
      next unless result_of_type?(raw_object, parent_type, fragment_type)

      result = propagate_object_scope_errors(raw_object, fragment_type, node.selections, state)
      return nil if result.nil?
    when GraphQL::Language::Nodes::FragmentSpread
      fragment = @executor.fragments[node.name] #: as !nil
      fragment_type = @context.types.type(fragment.type.name)
      next unless result_of_type?(raw_object, parent_type, fragment_type)

      result = propagate_object_scope_errors(raw_object, fragment_type, fragment.selections, state)
      return nil if result.nil?
    else
      raise DocumentError, "Invalid selection node type"
    end
  end

  raw_object
end

#: (Array[untyped]?, singleton(GraphQL::Schema::Member), Array[GraphQL::Language::Nodes::AbstractNode], State) -> Array[untyped]?
def propagate_list_scope_errors(raw_list, current_node_type, selections, state)
  return nil if raw_list.nil?

  item_node_type = Util.unwrap_non_null(current_node_type).of_type
  named_type = item_node_type.unwrap

  resolved_list = raw_list.map!.with_index do |raw_list_element, index|
    state.actual_path << index

    begin
      # Check for invalidated positions (inlined errors or marked results)
      result = if (err = @invalidated_results[raw_list_element])
        add_formatted_error(err, state)
        nil
      elsif item_node_type.list?
        propagate_list_scope_errors(raw_list_element, item_node_type, selections, state)
      elsif named_type.kind.leaf?
        raw_list_element
      else
        propagate_object_scope_errors(raw_list_element, named_type, selections, state)
      end

      return nil if result.nil? && item_node_type.non_null?

      result
    ensure
      state.actual_path.pop
    end
  end

  resolved_list
end

#: (untyped, singleton(GraphQL::Schema::Member), singleton(GraphQL::Schema::Member)) -> bool
def result_of_type?(result, current_type, inquiry_type)
  # result_type must be concrete...
  result_type = current_type.kind.abstract? ? @abstract_result_types[result] : current_type
  raise ImplementationError, "No type annotation recorded for abstract result" if result_type.nil?

  if inquiry_type.kind.abstract?
    # abstract inquiry contains the concrete result type?
    @context.types.possible_types(inquiry_type).include?(result_type)
  else
    # concrete result type matches the concrete inquiry type?
    result_type == inquiry_type
  end
end
```

lib/graphql/execution/batching/field_resolve_step.rb
rmosolgo: Benchmarks before and after 0bceefbfc3 : 

```diff
  graphql-ruby batching 14002 resolvers
-                         286.357 (± 1.4%) i/s    (3.49 ms/i) -      1.456k in   5.085654s
+                         310.091 (± 1.3%) i/s    (3.22 ms/i) -      1.560k in   5.031485s
```

😻 

I assume I'll run into issues with those dangling `nil`s at some point -- I'll switch over to an undefined value when I do.

lib/graphql/execution/batching/field_resolve_step.rb
```diff
@@ -0,0 +1,339 @@
...
+            if @static_type.kind.abstract?
+              next_objects_by_type = Hash.new { |h, obj_t| h[obj_t] = [] }.compare_by_identity
+              next_results_by_type = Hash.new { |h, obj_t| h[obj_t] = [] }.compare_by_identity
+
+              @all_next_objects.each_with_index do |next_object, i|
+                result = @all_next_results[i]
+                if (object_type = @runner.runtime_types_at_result[result])
+                  # OK
+                else
+                  object_type, _unused_new_value = @runner.schema.resolve_type(@static_type, next_object, @runner.context)
```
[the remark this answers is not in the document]
gmac: > What do you think of that distinction between "static" and "dynamic" (or "runtime") ?

aaah. I’d generally think the of those as composite_type versus object_type, using type_kinds.rb as the basis for nomenclature.

rmosolgo:
This is far from done but I think it's ready to merge because: 

- It's in a stable starting point 
- It's not loaded by the gem by default 
- The documentation is `published: false`
- It doesn't touch existing runtime code

gmac:
We’ll be linking to this on 3/10

rmosolgo:
Thanks for the heads up!

🍑 
🔥

lib/graphql/execution/batching/runner.rb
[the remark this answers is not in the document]
rmosolgo: I just took a crack at adapting this map-based result formatting structure in 43edf31f, thanks again for sharing it!

PR #5513, opened 2026-01-22
Fix link for compressed payloads section in the Relay Subscriptions guide

duffuniverse:
### Detail

Fix link for compressed payloads section in the Relay Subscriptions guide


rmosolgo:
Thanks again!

PR #5514, opened 2026-01-23
Fix docs link in the Object Cache's Runtime Considerations guide

duffuniverse:
### Detail

Add missing double quote to the link address.

The link is currently incorrect. See below:

Screenshot from 2026-01-23 13-40-56


rmosolgo:
Thank you!

PR #5517, opened 2026-01-26
Fix link to Redis backend documentation in the OperationStore Overview guide

duffuniverse:
### Detail

Fix link to Redis Backend documentation in the OperationStore Overview guide.


rmosolgo:
thanks!

PR #5519, opened 2026-01-29
Fix docs link in the Schema Root types guide

duffuniverse:
### Detail

Fix link to the root operation types documentation


rmosolgo:
Thanks!

PR #5523, opened 2026-02-03
Dashboard setup: try an initializer

rmosolgo:
Suggested in https://github.com/rmosolgo/graphql-ruby/pull/5515#issuecomment-3839885377


rmosolgo:
Doesn't work as intended

PR #5526, opened 2026-02-04
Fix a few docs links in the Executing Queries guide

duffuniverse:
### Detail

Fix a few docs links in the Executing Queries guide


rmosolgo:
Thanks!

PR #5528, opened 2026-02-05
Fix link to the complexity algo code in the Queries Complexity and Depth guide

duffuniverse:
### Detail

Fix link to the `GraphQL::Analysis::QueryComplexity` code in the Queries Complexity and Depth guide


rmosolgo:
Thanks! Missed this in #4996

PR #5529, opened 2026-02-06
Fix docs link in the Queries Multiplex guide

duffuniverse:
### Detail

Fix docs link in the Queries Multiplex guide


rmosolgo:
Thanks!

PR #5532, opened 2026-02-08
Fix Scout APM link in the Queries Tracing guide

duffuniverse:
### Detail

Fix Scout APM link in the Queries Tracing guide


rmosolgo:
Thank you!

PR #5545, opened 2026-02-19
Fix docs link in the Mutation Classes guide

duffuniverse:
### Detail
Fix docs link in the Mutation Classes guide


rmosolgo:
Thanks!

PR #5554, opened 2026-02-24
Add Herb to CI

joelhawksley:
As requested in https://github.com/rmosolgo/graphql-ruby/pull/5497#issuecomment-3744580881, this PR adds Herb to CI and fixes one remaining issue.

I excluded `.html` files as invalid HTML does not break downstream runtimes.

We should probably look at fixing the HTML parsing errors upstream @marcoroth.


rmosolgo:
Cool, thanks for this improvement!

PR #5556, opened 2026-02-25
Fix a few minor issues in the Pagination Cursors guide

duffuniverse:
### Detail

Replace the API documentation link with the source code link in Pagination Cursors guide.


rmosolgo:
Thanks for fixing this!

PR #5573, opened 2026-03-09
Fix: call loads hooks for resolver arguments from extensions

### LINKED ISSUE #5572: Resolver#initialize argument cache misses arguments added by field extensions
**Describe the bug**

When a FieldExtension adds arguments to a mutation field via apply, those arguments are not included in the `@arguments_by_keyword` cache built in `Resolver#initialize`. This means `context.types.argument(mutation_class, name)` returns `nil` for extension-added arguments, even though they are valid schema arguments on the field.

The root cause is [this caching in Resolver#initialize](https://github.com/rmosolgo/graphql-ruby/blob/3e60ec209617b841a6e96acb9c15760e146410c0/lib/graphql/schema/resolver.rb#L42-L44):

```ruby
@arguments_by_keyword = {}
context.types.arguments(self.class).each do |arg|
  @arguments_by_keyword[arg.keyword] = arg
end
```

`context.types.arguments(self.class)` only returns arguments defined directly on the resolver class — not arguments added to the field by extensions.

**Versions**

graphql version: 2.5.15
rails (or other framework): Rails (version not specific to the bug)
other applicable versions: N/A

**GraphQL schema**

```ruby
class ExtraInput < GraphQL::Schema::InputObject
  argument :value, String, required: false
end

  def apply
    field.argument(:extra, ExtraInput, required: false)
  end
end

  argument :input, String, required: false
  extension(MyExtension)

  field :result, String

  def resolve(input: nil, extra: nil)
    # extra is received correctly at resolve time,
    # but is not in @arguments_by_keyword
    { result: "ok" }
  end
end

class MutationRoot < GraphQL::Schema::Object
  field :my_mutation, mutation: MyMutation
end

  mutation MutationRoot
end
```

**GraphQL query**

```graphql
mutation {
  myMutation(input: "test", extra: { value: "hello" }) {
    result
  }
}
```

**Steps to reproduce**

1. Define a `FieldExtension` that adds an argument via `field.argument(...)` in `apply`
2. Attach the extension to a mutation with `extension(MyExtension)`
3. In any code that runs during resolution, call `context.types.argument(MyMutation, "extra")` — it returns `nil`
4. Alternatively, inspect `@arguments_by_keyword` inside the resolver — `:extra` is missing

**Expected behavior**

`context.types.arguments(self.class)` (and by extension `context.types.argument(self.class, name)`) should include arguments added by field extensions, since those arguments are part of the field's actual schema signature and are accepted at query time.

**Actual behavior**

Extension-added arguments are missing from `@arguments_by_keyword` and from `context.types.argument` lookups against the resolver class. The arguments work correctly at resolve time (they're passed as keyword args), but any introspection or validation code that uses the types API to look up arguments on the mutation class can't find them.

**Additional context**

A workaround is to look up arguments via the schema field instead of the resolver class:

```ruby
mutation_root = context.schema.mutation
mutation_field = context.types.field(mutation_root, "myMutation")
context.types.argument(mutation_field, "extra") # => returns the argument
```

This correctly finds extension-added arguments, but requires traversing the schema manually rather than using the resolver class directly.

rmosolgo:
Fixes #5572

### DIFF
--- a/lib/graphql/schema/resolver.rb
+++ b/lib/graphql/schema/resolver.rb
@@ -40,7 +40,10 @@ module GraphQL
         @field = field
         # Since this hash is constantly rebuilt, cache it for this call
         @arguments_by_keyword = {}
-        context.types.arguments(self.class).each do |arg|
+        # @field includes any extension-added arguments, but isn't available if the Resolver is directly initialized in user code (boo hiss).
+        # Use `self.class` when it's a RelayClassicMutation whose arguments actually exist on a generated `input: ...` object
+        arg_owner = self.is_a?(HasSingleInputArgument) ? self.class : (@field || self.class)
+        context.types.arguments(arg_owner).each do |arg|
           @arguments_by_keyword[arg.keyword] = arg
         end
         @prepared_arguments = nil
--- a/spec/graphql/schema/resolver_spec.rb
+++ b/spec/graphql/schema/resolver_spec.rb
@@ -89,15 +89,31 @@ describe GraphQL::Schema::Resolver do
     class Resolver8 < Resolver7
     end
 
+    module HasValue
+      include GraphQL::Schema::Interface
+      field :value, Integer, null: false
+      def self.resolve_type(obj, ctx)
+        LazyBlock.new {
+          if obj.is_a?(Integer)
+            IntegerWrapper
+          elsif obj == :resolve_type_as_wrong_type
+            GraphQL::Types::String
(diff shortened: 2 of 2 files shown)


rmosolgo:
This doesn't actually fix the issue and hasn't caused trouble for anyone yet, and won't matter in the new execution system, so I'm not going to continue with it. Please leave a message here if you'd like to see this merged.

PR #5584, opened 2026-03-23
Skip `did_you_mean` suggestions when feature is disabled

swalkinshaw:
When `did_you_mean` is disabled (common in production), skip the expensive all_types loading and dictionary computation entirely. Previously these allocations happened unconditionally on every validation error path.


rmosolgo:
Thanks for this improvement!

PR #5585, opened 2026-03-23
Cache `unwrap` and `to_type_signature` on type wrappers

swalkinshaw:
Type wrappers (NonNull, List) are schema-level immutable objects whose `unwrap` and `to_type_signature` results never change. Memoize both to eliminate repeated recursive unwrap calls and string allocations during validation and introspection.

I can't find any reason why these wouldn't be safe to memoize.

### DIFF
--- a/lib/graphql/schema/list.rb
+++ b/lib/graphql/schema/list.rb
@@ -19,7 +19,7 @@ module GraphQL
       end
 
       def to_type_signature
-        "[#{@of_type.to_type_signature}]"
+        @type_signature ||= -"[#{@of_type.to_type_signature}]"
       end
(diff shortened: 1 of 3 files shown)


rmosolgo:
It looks like some tests are failing because of the type definition being frozen. I'm open to any solution here, but my first thought is, what about adding: 

```
def freeze 
  to_type_signature 
  super 
end 
```

to `Wrapper`?

swalkinshaw:
Good idea; added!

swalkinshaw:
Last CI failure is flaky

rmosolgo:
Thanks!

PR #5586, opened 2026-03-23
Cache field and type lookups in Visibility::Profile

swalkinshaw:
Memoize the results of `field(owner, field_name)` and `type(type_name)` since validation traverses the same fields/types repeatedly, especially when expanding fragments. Safe because profiles are per-query objects.

This optimization really only shows up with visibility profiles enabled like you'd have in production. I forced a validation profile and pre-warmed it on the validation benchmarks and then the existing ones were 11-18% faster with this optimization. I've included that new `validate_profile` benchmark but not sure if there's a better way to do them.


rmosolgo:
Awesome, thanks for this. It looks like something went wrong, maybe with freezing the object. My first guess is that we need to pre-fill the cache in this method: 

https://github.com/rmosolgo/graphql-ruby/blob/9b17bfa3ac525cd3b44bf2310f0c5bf549635a16/lib/graphql/schema/visibility/profile.rb#L306-L309

Probably `type(type_name)` would work here -- I see that `field(...)` is already called below.

swalkinshaw:
Tests are good with that `type` call 👍

rmosolgo:
Thanks for this improvment!

PR #5587, opened 2026-03-24
Replace validation visitor stacks with single-slot variables

swalkinshaw:
Extracted https://github.com/rmosolgo/graphql-ruby/pull/5578

Eliminates array push/pop operations per validation pass by replacing array-based stacks with single-slot variables that use the Ruby call stack for save/restore. Pre-allocates path array indexed by depth counter.

In our schema, this saves about ~5500 push/pop operations and resulted in a 5-6% perf improvement.

I'll defer to you if you think this makes the code more complex and harder to manage but I think it's about the same.


swalkinshaw:
Last failure is a common flaky async dataloader test

rmosolgo:
Thanks!

PR #5588, opened 2026-03-24
Optimize validation rule hot paths

swalkinshaw:

These a few miscellaneous micro-optimizations to validation rules.

- `ArgumentNamesAreUnique`: avoid Hash.new block allocation on the common unique-args path; only build array on actual collision
- `FieldsHaveAppropriateSelections`: fast-path leaf+no-selections and non-leaf+selections cases, skipping the full validation method
- `RequiredArgumentsArePresent`: cache required arg names per definition to avoid re-iterating arguments for repeated field definitions
~- `FragmentSpreadsArePossible`: use Array#intersect? instead of none? { include? } for type overlap check~

Note: I'm keeping all the `FieldsWillMerge` optimizations for their own PR.

### DIFF
--- a/lib/graphql/static_validation/rules/argument_names_are_unique.rb
+++ b/lib/graphql/static_validation/rules/argument_names_are_unique.rb
@@ -16,12 +16,24 @@ module GraphQL
 
       def validate_arguments(node)
         argument_defns = node.arguments
-        if !argument_defns.empty?
-          args_by_name = Hash.new { |h, k| h[k] = [] }
-          argument_defns.each { |a| args_by_name[a.name] << a }
-          args_by_name.each do |name, defns|
-            if defns.size > 1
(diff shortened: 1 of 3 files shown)


swalkinshaw:
Oh didn't realize Ruby 2.7 was still supported 😓 so `intersect?` isn't available. I'll drop that one

rmosolgo:

I'll have to keep intersect in mind for when we drop Ruby 2.7 support. I think there are other places it could be used, too, for example: 

https://github.com/rmosolgo/graphql-ruby/blob/6d1e5fadc61293855101d1295d9f6b3090c91c76/lib/graphql/execution/next/runner.rb#L383

https://github.com/rmosolgo/graphql-ruby/blob/6d1e5fadc61293855101d1295d9f6b3090c91c76/lib/graphql/analysis/query_complexity.rb#L163

PR #5591, opened 2026-03-26
Optimize FieldsWillMerge validation rule

swalkinshaw:

This is the main optimization since the `FieldsWillMerge` rule basically always dominates static validation time so improving the speed of this rule impacts the entire pipeline. Each optimization is broken down into a separate commit.

The original algorithm compared fields across three phases (fields-within, fields-vs-fragments, fragments-vs-fragments) with recursive fragment cross-comparison that could be exponential. This rewrites it to:

* Flatten all fragment spreads inline into a single response-key map in one pass, eliminating the three-phase structure entirely
* Deduplicate by signature — fields with identical (name, definition, arguments) are grouped and only one representative pair is compared
* Skip provably-safe selections — if all direct children of a selection set are unaliased fields with unique names (no fragments, no aliases, no duplicates), conflict checking is skipped entirely (~55% of non-leaf fields)
* Defer array wrapping — 85% of response keys have a single field, so store the Field directly instead of wrapping in a single-element array
* Cache sub-selection results and mutually-exclusive type pairs to avoid redundant work across comparisons

### DIFF
--- a/lib/graphql/static_validation/rules/fields_will_merge.rb
+++ b/lib/graphql/static_validation/rules/fields_will_merge.rb
@@ -454,27 +454,22 @@ module GraphQL
         end
       end
 
-      def each_field(fields_or_field)
-        if fields_or_field.is_a?(Field)
-          yield fields_or_field
-        else
-          fields_or_field.each { |f| yield f }
-        end
-      end
-
       def find_conflicts_between(response_keys, response_keys2, mutually_exclusive:)
         response_keys.each do |key, fields|
           fields2 = response_keys2[key]
-          if fields2
-            each_field(fields) do |field|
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Okay I got Claude to do a few passes at expanding test coverage referencing https://github.com/graphql/graphql-js/tree/16.x.x/src/validation/__tests__ too. 36 added test cases 

They discovered one regression/issue which has been fixed

rmosolgo:
Thanks so much for your work on this. I have to admit that I don't 100% understand the new code, but I read through it (and the article) and I think I could dig in if/when it needs maintenance. And it passes all the old tests, and then some, so I'm confident that it's correct.

PR #5596, opened 2026-04-01
Exec-next: Finalizers

rmosolgo:
Add a way for code to run after the query is finished, with awareness of its result path, the way that errors do. 

I tried to migrate `context.errors` to this same flow, but it was tricker because `Schema.type_error` already appends to `context.errors`. 

Also, implements `run_partials` for Execution::Next 

Also, implements directives on fields, inline fragments, and fragment spreads 

- [x] RawValue? 
- [x] DRY run_partials entry points 
- Fix `.coerce_arguments` usage: __I'll do this in a follow-up PR__


rmosolgo:
`system_tests` pass locally, I'm going to assume this is unrelated and check up on it later.

PR #5600, opened 2026-04-17
docs: correct yield to super in execute_multiplex method. Fixes #5599

### LINKED ISSUE #5599: Potential issue with documentation around Multiplex Tracing
**Describe the bug**

The documentation example for [Multiplex Tracing](https://graphql-ruby.org/queries/multiplex.html#multiplex-tracing) uses `yield`:

```ruby
# Count how many queries are in the multiplex run:
module MultiplexCounter
  def execute_multiplex(multiplex:)
    Rails.logger.info("Multiplex size: #{multiplex.queries.length}")
    yield
  end
end

# ...

  # ...
  trace_with(MultiplexCounter )
end
```

But if I understand correctly, `trace_with` works by mixing every trace module into `Trace` class as per this code comment:

```ruby
# Mix `trace_mod` into this schema's `Trace` class so that its methods will be called at runtime.
```

This would suggest that rather than using `yield` the correct approach would be to use `super` or attaching multiple tracers will break (the tracer without `super` will halt the execution of other tracers). And it seems the library code already uses `super` for every built-in tracer (`LegacyHooksTrace`, `NotificationsTrace` etc)

Is this a bug in documentation or am I missing something and `yield` / `super` are interchangeable in this context?

**Versions**

`graphql` version: 2.5.23

d4rky-pl:


d4rky-pl:
Done in c011759589b2d01b852b15ba67ceb875c9660936

rmosolgo:
Oops, sorry I missed this PR or I would have just merged it 🤦

d4rky-pl:
@rmosolgo the important part is the docs are better now, not the attribution 🫶

PR #5619, opened 2026-05-01
Execution::Next: add as_default: true option, run entire test suite with it

rmosolgo:
`use GraphQL::Execution::Next, as_default: true` will make `Schema.execute(...)` use Execution::Next.

Expanding CI coverage to all files ... there could be a lot of loose ends to follow up here 😅 

### TODO: 

- [x] ResolveTypeStep: properly batch these. Initialize a step when needed here.
- [x] FieldResolveStep shortcuts when there's an error in arguments. Right now it goes three ways: a conditional in `build_arguments`, an early conditional in `execute_field`, and outside calls to `loaded_argument_unauthorized`. The fact that this is here three times is totally a smell. How can this data structure and flow be improved?


rmosolgo:
> 15 failures, 113 errors

Almost there 😅

rmosolgo:
Down to 3 failures ... they should be addressed, but I may skip them for now and address them in a later PR

PR #5621, opened 2026-05-06
Add ESM build output for graphql-ruby-client

mattander:
## Summary

This adds an ES module build for `graphql-ruby-client` while keeping the existing CommonJS build intact for current consumers.

The package now publishes:

- existing CommonJS output through `main` / `require`
- generated ESM output under `esm/` through `module` / conditional `import` exports
- subpath exports for `sync` and `subscriptions` modules, including `.js` subpaths for existing direct import users

The ESM build runs from `prepack`, so `npm pack`, `npm publish`, and Git dependency installs generate the ESM output before packaging. The ESM preparation step also normalizes generated import specifiers that Node's ESM resolver cannot load as-is, such as extensionless relative imports and directory-style package subpaths.

## Why

Modern bundlers and ESM-only app setups can otherwise end up consuming the CommonJS build or needing local aliases/workarounds for imports such as `graphql-ruby-client/subscriptions/ActionCableLink`.

## Verification

- `npm test`
- `npm pack --dry-run`
- `node --input-type=module -e 'import { sync, ActionCableLink } from "graphql-ruby-client"; console.log(typeof sync, typeof ActionCableLink)'`
- `node --input-type=module -e 'import ActionCableLink from "graphql-ruby-client/subscriptions/ActionCableLink.js"; console.log(typeof ActionCableLink)'`
- `node -e 'const ActionCableLink = require("graphql-ruby-client/subscriptions/ActionCableLink.js"); console.log(typeof ActionCableLink.default || typeof ActionCableLink)'`
- `node -e 'require.resolve("graphql-ruby-client/cli.js"); require.resolve("graphql-ruby-client/readme.md"); require.resolve("graphql-ruby-client/CHANGELOG.md"); console.log("ok")'`

### DIFF
--- a/javascript_client/package.json
+++ b/javascript_client/package.json
@@ -3,7 +3,71 @@
   "version": "1.14.9",
   "description": "JavaScript client for graphql-ruby",
   "main": "index.js",
+  "module": "esm/index.js",
   "types": "index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./index.d.ts",
+      "import": "./esm/index.js",
+      "require": "./index.js",
+      "default": "./index.js"
+    },
+    "./index.js": {
+      "types": "./index.d.ts",
+      "import": "./esm/index.js",
+      "require": "./index.js",
+      "default": "./index.js"
+    },
+    "./cli.js": "./cli.js",
+    "./CHANGELOG.md": "./CHANGELOG.md",
+    "./DumpPayloadExample.json": "./DumpPayloadExample.json",
+    "./LICENSE.md": "./LICENSE.md",
+    "./readme.md": "./readme.md",
+    "./sync": {
+      "types": "./sync/index.d.ts",
+      "import": "./esm/sync/index.js",
+      "require": "./sync/index.js",
+      "default": "./sync/index.js"
+    },
+    "./sync/*": {
+      "types": "./sync/*",
+      "import": "./esm/sync/*.js",
+      "require": "./sync/*.js",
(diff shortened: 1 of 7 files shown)


javascript_client/package.json
```diff
@@ -3,7 +3,71 @@
...
+      "types": "./index.d.ts",
+      "import": "./esm/index.js",
+      "require": "./index.js",
+      "default": "./index.js"
+    },
+    "./cli.js": "./cli.js",
+    "./CHANGELOG.md": "./CHANGELOG.md",
+    "./DumpPayloadExample.json": "./DumpPayloadExample.json",
+    "./LICENSE.md": "./LICENSE.md",
+    "./readme.md": "./readme.md",
```
rmosolgo: Why should these be included in the exports?

rmosolgo (commented):
Hey, thanks for contributing this improvement! I have one question about the changes below. 

IIUC, I won't need to do anything extra to include these new esm files in the next release. Does that sound right to you?

One more thought: it sounds like you validated the module structure with some local scripts. Is there a way we could work this into CI? Maybe write a spec that runs those same commands?

rmosolgo:
Hey @mattander, thanks again for contributing this. Please take a look at my questions when you get a chance!

rmosolgo:
👋  I adapted your scripts into a couple and test cases and, since I didn't understand the rationale, I removed the documentation files from the list of exports. I'd be happy to add those back in another PR if you're willing to help me understand why they're necessary.

PR #5623, opened 2026-05-12
Fix Ruby warnings in the GraphQL::Schema::InputObject class tests

duffuniverse:
I ran tests for the `GraphQL::Schema::InputObject` class with Ruby warnings enabled:

```
RUBYOPT="-W2" bundle exec rake test TEST=spec/graphql/schema/input_object_spec.rb

/home/andrey/projects/graphql-ruby/spec/graphql/schema/input_object_spec.rb:848: warning: assigned but unused variable - z
```
Tests passed though :green_circle: .

### DIFF
--- a/spec/graphql/schema/input_object_spec.rb
+++ b/spec/graphql/schema/input_object_spec.rb
@@ -892,7 +892,7 @@ describe GraphQL::Schema::InputObject do
 
     it "does not match missing keys" do
       assert case input_object
-      in { z: }
+      in { z: 5 }
         false
       else
         true
--- a/spec/graphql/schema/input_object_spec.rb
+++ b/spec/graphql/schema/input_object_spec.rb
@@ -892,8 +892,8 @@ describe GraphQL::Schema::InputObject do
 
     it "does not match missing keys" do
       assert case input_object
-      in { z: 5 }
-        false
+      in { z: }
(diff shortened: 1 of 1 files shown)


rmosolgo:
I agree it'd be good to address this warning, but this change actually changes the semantics of the test. The intention of the test is to confirm that InputObject pattern matching is implemented correctly, and it doesn't match as if it _had_ keys which it _doesn't have_. 

The object in this test doesn't have a `z:` key, but that `in ...` pattern _would_ match if there _was_ a `z:` key.

By modifying the `in ... ` pattern to `z: 5`, we change the semantics of the test so that it's testing that either 1) `input_object` has _no_ `z:` key; or 2) it has a `z:` key but the value for `z:` is _not 5_. 

I realize this is splitting hairs but I don't think that reducing the scope of an existing test is the right way to address this warning. 

One way to keep the existing test but address the warning would be to use `z` in the case code: 

```ruby 
      in { z: }
        raise "Unexpectedly matched a non-present key, z: #{z.inspect} (in: #{input_object.inspect})"
```

That would cause the test to fail if `z:` is present _at all_, but still please Ruby that `z` is used. What do you think? 

(I would have left it as a code suggestion but I couldn't figure out how to suggest a change to the unmodified line of the test :confounded:)

duffuniverse:
Hmm, that’s an interesting approach - applied

rmosolgo:
Thanks for this improvement!

PR #5633, opened 2026-05-14
Don't crash analyzer when @skip/@include arg fails coercion

artplan1:
When `null` is passed for an `@include(if: Boolean!)` / `@skip(if: Boolean!)` argument (e.g. a nullable variable with a default sent as null), `arguments_for` returns a `GraphQL::ExecutionError`.
`DirectiveChecks#include?` then called `args[:if]` on it and raised `NoMethodError`, crashing any schema that runs an analyzer (`max_depth`, `max_complexity`, etc).
Skip the directive in that case so execution surfaces the error normally.


rmosolgo:
Awesome, sorry for the trouble with this and thanks for sharing your fix!

PR #5635, opened 2026-05-15
Visibility: require all root config before preloading

### LINKED ISSUE #5454: `GraphQL::Schema::Visibility` preload crashes if `query` type is not set first
If the `GraphQL::Schema::Visibility` plugin is "used" with `preload: true` prior to configuring the schema's root types, an exception will be raised.

```
NoMethodError: undefined method 'kind' for nil

gems/graphql-2.5.14/lib/graphql/schema/visibility/profile.rb:177:in 'GraphQL::Schema::Visibility::Profile#field'
gems/graphql-2.5.14/lib/graphql/schema/visibility/profile.rb:332:in 'block in GraphQL::Schema::Visibility::Profile#preload'
gems/graphql-2.5.14/lib/graphql/schema/visibility/profile.rb:328:in 'Array#each'
gems/graphql-2.5.14/lib/graphql/schema/visibility/profile.rb:328:in 'GraphQL::Schema::Visibility::Profile#preload'
gems/graphql-2.5.14/lib/graphql/schema/visibility.rb:99:in 'block in GraphQL::Schema::Visibility#preload'
gems/graphql-2.5.14/lib/graphql/schema/visibility.rb:97:in 'Hash#each'
gems/graphql-2.5.14/lib/graphql/schema/visibility.rb:97:in 'GraphQL::Schema::Visibility#preload'
gems/graphql-2.5.14/lib/graphql/schema/visibility.rb:44:in 'GraphQL::Schema::Visibility#initialize'
gems/graphql-2.5.14/lib/graphql/schema/visibility.rb:20:in 'Class#new'
gems/graphql-2.5.14/lib/graphql/schema/visibility.rb:20:in 'GraphQL::Schema::Visibility.use'
gems/graphql-2.5.14/lib/graphql/schema.rb:322:in 'GraphQL::Schema.use'
```

This is likely a programmer error, however a sanity check with proper error message might be more helpful. It is not immediately obvious that certain plugins need to be configured in a certain order.

rmosolgo:
> [!IMPORTANT]
> If you're using Rails and have `use GraphQL::Schema::Visibility` before configuring your root types, 
> this will raise an error at application boot. 
> 
> Fix the error by moving `use ...` below that configuration. 
>
> Your app will boot faster every time after that! 

Fixes #5454 

GraphQL::Schema::Visibility didn't handle it right when root-level configurations came after `use ... preload: true`. It tried to, but profiles still weren't actually preloaded correctly. 

When configuration is inherited, it has to fall back to rebuilding the visibility cache dynamically. This is slower than the new behavior ... but it's also how it used to always work. It can be avoided by removing `use ...` from the parent class and putting it in child classes only.

### DIFF
--- a/lib/graphql/schema/introspection_system.rb
+++ b/lib/graphql/schema/introspection_system.rb
@@ -105,11 +105,13 @@ module GraphQL
       end
 
       def load_constant(class_name)
-        const = @custom_namespace.const_get(class_name)
+        const = begin
+          @custom_namespace.const_get(class_name)
+        rescue NameError
+          # Dup the built-in so that the cached fields aren't shared
+          @built_in_namespace.const_get(class_name)
+        end
         dup_type_class(const)
-      rescue NameError
-        # Dup the built-in so that the cached fields aren't shared
-        dup_type_class(@built_in_namespace.const_get(class_name))
       end
 
       def get_fields_from_class(class_sym:)
@@ -131,23 +133,6 @@ module GraphQL
               add_field(dup_field)
             end
           end
-        end
-      end
-
-      class PerFieldProxyResolve
-        def initialize(object_class:, inner_resolve:)
-          @object_class = object_class
-          @inner_resolve = inner_resolve
-        end
-
-        def call(obj, args, ctx)
-          query_ctx = ctx.query.context
-          # Remove the QueryType wrapper
-          if obj.is_a?(GraphQL::Schema::Object)
-            obj = obj.object
-          end
-          wrapped_object = @object_class.wrap(obj, query_ctx)
-          @inner_resolve.call(wrapped_object, args, ctx)
         end
       end
     end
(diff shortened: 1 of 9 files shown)


lib/graphql/schema/introspection_system.rb
```diff
@@ -133,23 +135,6 @@ def dup_type_class(type_class)
...
-        def call(obj, args, ctx)
-          query_ctx = ctx.query.context
-          # Remove the QueryType wrapper
-          if obj.is_a?(GraphQL::Schema::Object)
-            obj = obj.object
-          end
-          wrapped_object = @object_class.wrap(obj, query_ctx)
-          @inner_resolve.call(wrapped_object, args, ctx)
-        end
-      end
```
rmosolgo, line 152: Wow, I think this was _very_ old!

PR #5636, opened 2026-05-19
Fix `required: :nullable` crash when combined with `as:`

medlefsen:
I ran across this issue when trying to use `as:` on an argument which used `required: :nullable`. In my version it crashes with NoMethodError but it looks like in newer versions the validation just always fails.

This fix uses `@keyword` instead of `arg_name`, which I think should be correct. Tests pass.

### DIFF
--- a/lib/graphql/schema/argument.rb
+++ b/lib/graphql/schema/argument.rb
@@ -94,7 +94,7 @@ module GraphQL
         end
 
         if required == :nullable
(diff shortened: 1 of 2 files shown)


rmosolgo:
Sorry for the trouble on this and thanks for contributing a fix!

medlefsen:
> Sorry for the trouble on this and thanks for contributing a fix!

No problem! Thanks

Issue #5632, opened 2026-05-13
Cache complexity analyzer in-memory/redis


sobrinho:
**Is your feature request related to a problem? Please describe.**

After the optimization in https://github.com/rmosolgo/graphql-ruby/pull/5631,
our `GraphQL/analyze` dropped from ~806ms to ~176ms. However, 176ms is still
significant for a hot path since the complexity analyzer runs on every request,
even when the same query has been analyzed before.

**Describe the solution you'd like**

I would like a way to cache the complexity analysis result across requests —
either in-process (Ruby Hash/LRU, sub-microsecond) or in a shared store like
Redis (~1ms). The gem itself does not need to know how to cache; that
responsibility can stay with the application. What the gem needs to provide is
a stable **fingerprint** for the incoming query so the application can use it
as a cache key.

The challenge is determining what the fingerprint should cover. The complexity
of a query depends on its structure *and* on certain variables — most notably
pagination arguments (e.g. `first`, `last`) — so those must be included in the
fingerprint.

**Describe alternatives you've considered**

Keying the cache on the full query string + all variables is too conservative:
two calls with the same query structure but different `id` arguments would
always miss the cache even though their complexity is identical.

Keying on the query string alone (ignoring all variables) is too aggressive:
it would treat `first: 10` and `first: 1000` as the same, producing incorrect
complexity results.

The right fingerprint is the query structure (document or normalized AST)
combined with only the variables that influence complexity, such as pagination
arguments.

**Additional context**

https://github.com/rmosolgo/graphql-ruby/pull/5631

sobrinho:
Naive in-memory example:

```ruby
class CachedQueryComplexityAnalyzer < GraphQL::Analysis::QueryComplexity
  COMPLEXITY_CACHE = {}

  def initialize(...)
    super
    @cache_key = Digest::SHA256.hexdigest(@query.query_string)
  end

  def visit?
    !COMPLEXITY_CACHE.key?(@cache_key)
  end

  def result
    COMPLEXITY_CACHE[@cache_key] ||= super
  end
end
```

The problem, as mentioned, is that a variable like first: 10 vs first: 100 directly impacts complexity, so keying on the query string alone produces incorrect results. The natural fix is to include the variables in the key — but then two calls with the same query structure but different id arguments never share a cache entry, making the cache too coarse to be useful:

```ruby
class CachedQueryComplexityAnalyzer < GraphQL::Analysis::QueryComplexity
[...]

PR #5638, opened 2026-05-23
PoC: Cache complexity

sobrinho:
In the first round (#5631) we dropped from ~806ms to ~176ms, but that's still too much processing time. Caching the complexity calculation can bring that 176ms down to ~1ms on a cache hit. This PR is a PoC for that approach.

The slowest piece is calling `max_possible_complexity`, not the visitor itself. Further improvements could be made with that in mind, but I didn't go down that path since caching largely solves the problem without requiring a refactor of the complexity calculation. We also have `field_complexity`, but it can't be used here because it's called *during* `max_possible_complexity` — and what we actually want is to avoid calling `max_possible_complexity` altogether.

### The digesting principle

The approach is simple:

1. Capture the fields and their complexity values (if the complexity changes, the digest changes).
2. Skip caching entirely if any field uses dynamic complexity (a lambda and/or `complexity_for`).
3. When validating the complexity, compute the digest and look it up.

### Open questions

1. **Multiplexes:** `def result` doesn't have access to queries, so I haven't figured out how to make this analyzer work with multiplexed queries. (We don't multiplex very often on our end, so this hasn't been a blocker for the PoC.)
2. **Error handling:** `visitor.arguments_for` might raise. What should we do in that case — skip the cache attempt and fall through to the normal calculation?
3. **Does this belong in core?** I think it should, at least the digest calculation, since it's tightly coupled to the gem's internals.
4. **`Query#complexity_fingerprint`:** We initially discussed in #5632 the idea of adding a `Query#complexity_fingerprint` method, but that would require firing a dedicated analyzer, which feels out of place compared to how the other fingerprints are designed:

```ruby
   def complexity_fingerprint
     @complexity_fingerprint ||= GraphQL::Analysis.analyze_query(self, [TheCacheDigestComplexityAnalyzer])[0]
   end
```

   The upside would be reusability for other purposes (though honestly, I can't think of any concrete ones right now). `analyze_query` should be fast, but this still feels a bit redundant given that we already have the complexity analyzer.

### Feedback

How do you feel about this PoC? There are no specs yet — I'm looking for initial feedback before moving forward and handling all the edge cases.

### DIFF
--- a/lib/graphql/analysis/query_complexity.rb
+++ b/lib/graphql/analysis/query_complexity.rb
@@ -11,6 +11,8 @@ module GraphQL
         @complexities_on_type_by_query = {}
         @intersect_cache = Hash.new { |h, k| h[k] = {}.compare_by_identity }.compare_by_identity
         @possible_types_cache = {}.compare_by_identity
+        @uncacheable = []
+        @complexity_digest = Hash.new { |h, k| h[k] = Set.new }
       end
 
       # Override this method to use the complexity result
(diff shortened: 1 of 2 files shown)


sobrinho:
A few caveats: the code isn't pretty, relying on the caller to sort is awkward, and even delegating the SHA256 hashing might be questionable. All of this could potentially happen inside the gem itself, exposing only the final fingerprint.

But again, I want to validate the idea/principle of the problem and solution before moving forward with these technical decisions.

lib/graphql/analysis/query_complexity.rb
```diff
@@ -85,14 +87,58 @@ def on_enter_field(node, parent, visitor)
...
+            max_possible_page_size = arguments[:last]
+          end
+
+          if max_possible_page_size.nil?
+            max_possible_page_size = field_definition.default_page_size || visitor.query.schema.default_page_size || field_definition.max_page_size || visitor.query.schema.default_max_page_size
+          end
+
+          if max_possible_page_size.nil?
+            raise GraphQL::Error, "Can't calculate complexity for #{field_definition.path}, no `first:`, `last:`, `default_page_size`, `max_page_size` or `default_max_page_size`"
+          end
```
sobrinho, line 136: This entire algorithm was copied over from `GraphQL::Schema::Field#calculate_complexity`. It is very likely that we want to extract this `max_possible_page_size` to a method instead of duplicating it.

lib/graphql/analysis/query_complexity.rb
```diff
@@ -85,14 +87,58 @@ def on_enter_field(node, parent, visitor)
...
+          return
+        end
+
+        # Add the field definition to the complexity fields hash.
+        complexity_digest = [field_definition.name, field_definition.complexity]
+
+        # The page size affects the complexity.
+        if field_definition.connection?
+          arguments = visitor.arguments_for(node, field_definition)
+          return unless arguments.is_a?(GraphQL::Execution::Interpreter::Arguments) # FIXME: Are errors by definition not cacheable?
```
sobrinho, line 119: It is possible that what we are looking for here is to flag as not cacheable since we can't access the actual arguments:

```ruby
unless arguments.is_a?(GraphQL::Execution::Interpreter::Arguments)
  @uncacheable << visitor.query
  @complexity_digest.delete(visitor.query)

  return
end
```

lib/graphql/analysis/query_complexity.rb
```diff
@@ -85,14 +87,58 @@ def on_enter_field(node, parent, visitor)
...
+
+          if max_possible_page_size.nil?
+            max_possible_page_size = field_definition.default_page_size || visitor.query.schema.default_page_size || field_definition.max_page_size || visitor.query.schema.default_max_page_size
+          end
+
+          if max_possible_page_size.nil?
+            raise GraphQL::Error, "Can't calculate complexity for #{field_definition.path}, no `first:`, `last:`, `default_page_size`, `max_page_size` or `default_max_page_size`"
+          end
+
+          complexity_digest << max_possible_page_size
```
sobrinho, line 138: For a field (assuming complexity 1):

```ruby
["Person.name", 1]
```

For a connection (assuming complexity 1 and first/last as 10):

```ruby
["People", 1, 10]
```

sobrinho:
We're also trying to cache in a way where the query string itself isn't relevant — what matters are the actual fields being selected. Some queries have the exact same complexity even though their query strings differ.

For example:

```graphql
query MyQuery($search: String, $limit: Int!) {
  people(search: $search, first: $limit) {
    id
    firstName
    lastName
  }
}
```

has the exact same complexity as the following, even though the query string is completely different:

```graphql
query AnotherQuery($s: String!, $x: Int!) {
  myAlias: people(first: $x, search: $s) {
    lastName
    id
    firstName
  }
}
```

In the end, we're selecting the same fields with the same arguments. Furthermore, the `first` argument impacts complexity while `search` has no impact at all (assuming complexity isn't dynamic).

So whether I send `{ s: "Sobrinho", x: 10 }` or `{ x: 10, s: "Gabriel" }`, we know the complexity is the same. And we also know the complexity changes if `x` changes, even if `s` stays the same: `{ s: "Sobrinho", x: 10 }` has a different complexity from `{ s: "Sobrinho", x: 100 }`.

The PoC takes this into account.

PR #5642, opened 2026-05-26
Use `crypto.randomUUID()` for ActionCableLink channelId to prevent collisions

### LINKED ISSUE #5639: `ActionCableLink`: low-entropy `channelId` causes cross-subscription payload leak
## Describe the bug

`ActionCableLink` generates the per-subscription `channelId` with low entropy,
causing occasional collisions when multiple subscriptions are opened
simultaneously (typical on initial page load). When a collision occurs, one
subscription's payload is delivered to another subscription's listener,
producing Apollo cache warnings and `TypeError` crashes in `onData`.

## Versions

- `graphql-ruby-client`: 1.14.8 (current `master` is also affected)
- `@rails/actioncable`: 8.x
- Transport: ActionCable

## Root cause

In [`javascript_client/src/subscriptions/ActionCableLink.ts#L37`](https://github.com/rmosolgo/graphql-ruby/blob/master/javascript_client/src/subscriptions/ActionCableLink.ts#L37):

```ts
var channelId = Math.round(Date.now() + Math.random() * 100000).toString(16)
```

The random component is uniformly distributed over ~100,001 values. For two
subscriptions opened within the same millisecond, the collision probability is
~1/100,000. With 10–20 subscriptions opening at page boot, the probability
that at least one pair collides is roughly 0.05–0.2% per page load.

When a collision occurs:

1. The Rails ActionCable server deduplicates the server-side subscribe by
   identifier (`return if subscriptions.key?(id_key)` in
   `action_cable/connection/subscriptions.rb`).
2. The client retains both `Subscription` objects — ActionCable JS does not
   deduplicate.
3. On message arrival, ActionCable JS routes by identifier and invokes
   `received` on **all** local subscriptions whose identifier matches
   (`subscriptions.filter(s => s.identifier === identifier)`).
4. The "deduplicated" (locally surviving) subscription's listener receives the
   other subscription's payload, which has the wrong shape.

## Symptoms

Apollo logs (production minified error #13):

```
Missing field 'subscriptionAName' while writing result
{ subscriptionBName: { ... } }
```

…and downstream `onData` handlers crash:

```
TypeError: Cannot read properties of undefined (reading 'someField')
  at onData (...)
```

## Steps to reproduce

1. Set up an Apollo Client with `ActionCableLink` for ActionCable transport.
2. Mount many `useSubscription` calls simultaneously on a single page (e.g. a
   navigation badge subscription + multiple form/detail subscriptions, ~10+).
3. Reload repeatedly; occasionally two subscriptions get the same `channelId`
   and one of them starts receiving the other's payloads.

## Expected behavior

Each subscription receives only its own payload.

## Proposed fix

Use `crypto.randomUUID()` to guarantee uniqueness:

```ts
const channelId = crypto.randomUUID()
```

`crypto.randomUUID()` is available in all modern browsers (Chrome 92+, Firefox
95+, Safari 15.4+) and Node 14.17+. If wider compatibility is required, a
high-entropy random fallback (e.g. `crypto.getRandomValues` over a 128-bit
buffer) would also work.

I'm happy to submit a PR if this approach is acceptable.

obake-fe:
## Summary

Replace the channelId generator in `ActionCableLink` to use `crypto.randomUUID()` so that simultaneously-created subscriptions no longer collide.

Fixes #5639.

## Background

The previous channelId generator,

```ts
var channelId = Math.round(Date.now() + Math.random() * 100000).toString(16)
```

draws from only ~100,001 distinct values per millisecond. When multiple `useSubscription` hooks mount during the same render pass (or simply within a short time window), collisions are inevitable. With N subscriptions opened within ~100 seconds of each other, the collision probability is approximately `C(N, 2) / 100_000` (e.g. ~0.05% for N=10, ~0.2% for N=20).

When two subscriptions share the same identifier:

1. The Rails ActionCable server deduplicates the second `subscribe` (`return if subscriptions.key?(id_key)` in `action_cable/connection/subscriptions.rb`).
2. The client retains both `Subscription` objects — ActionCable JS does not deduplicate.
3. On message arrival, ActionCable JS invokes `received` on **all** local subscriptions whose identifier matches.
4. The "deduplicated" (locally surviving) subscription's listener receives the other subscription's payload, which has the wrong shape, causing Apollo to log `Missing field 'X' while writing result {...}` warnings and downstream `onData` handlers to crash with `TypeError`.

## Changes

- `ActionCableLink.ts`: replace the time+random channelId with `crypto.randomUUID()`. Add a short comment explaining why.
- `__tests__/ActionCableLinkTest.ts`: add a test that creates 1000 subscriptions in a tight loop and asserts the channelIds are all unique. With the old generator this would fail frequently (collisions are expected when many IDs are drawn within a single ms); with `crypto.randomUUID()` it always passes.

## Compatibility

`crypto.randomUUID()` is available in:

- Chrome 92+ (July 2021)
- Firefox 95+ (December 2021)
- Safari 15.4+ (March 2022)
- Node 14.17+ / 16.7+ (and via `globalThis.crypto` in Node 19+)

If wider compatibility is required I'm happy to add a fallback (e.g. `crypto.getRandomValues` over a 128-bit buffer, or a module-scoped counter + Date.now).

## Test plan

- [x] `npm test` (full suite) passes locally — 92 passed / 3 skipped (unchanged from before)
- [x] New test `generates a unique channelId for each subscription` passes

### DIFF
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -32,7 +32,11 @@ class ActionCableLink extends ApolloLink {
   // instead, it sends the request to ActionCable.
   request(operation: Operation, _next: NextLink): Observable<RequestResult> {
     return new Observable((observer) => {
-      var channelId = Math.round(Date.now() + Math.random() * 100000).toString(16)
+      // `crypto.randomUUID()` (Web Crypto API) is used because a low-entropy
+      // identifier here can collide between simultaneously-created subscriptions,
+      // and ActionCable routes incoming payloads by identifier — colliding
+      // subscriptions would receive each other's payloads.
+      var channelId = crypto.randomUUID()
       var actionName = this.actionName
       var connectionParams = (typeof this.connectionParams === "function") ?
         this.connectionParams(operation) : this.connectionParams
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -10,15 +10,31 @@ type SubscriptionCallbacks = {
   received?: (payload: any) => void;
 };
 
+type CreateChannelId = () => string
+
+function createChannelId() {
+  // `crypto.randomUUID()` (Web Crypto API) is used because a low-entropy
+  // identifier here can collide between simultaneously-created subscriptions,
+  // and ActionCable routes incoming payloads by identifier — colliding
+  // subscriptions would receive each other's payloads.
+  return crypto.randomUUID()
+}
+
 class ActionCableLink extends ApolloLink {
   cable: Consumer
   channelName: string
   actionName: string
   connectionParams: ConnectionParams
   callbacks: SubscriptionCallbacks
+  createChannelId: CreateChannelId
 
   constructor(options: {
-    cable: Consumer, channelName?: string, actionName?: string, connectionParams?: ConnectionParams, callbacks?: SubscriptionCallbacks
+    cable: Consumer,
+    createChannelId?: CreateChannelId,
+    channelName?: string,
+    actionName?: string,
+    connectionParams?: ConnectionParams,
+    callbacks?: SubscriptionCallbacks,
   }) {
     super()
     this.cable = options.cable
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks again for taking this on. 

> wider compatibility 

This is a good point -- I doubt anyone will need it, but I imagine a global crypto polyfill would work, too. I added a commit that also accepts an injected `channelId: ...` option, so if someone needs backwards compatibility (why??), they could acheive it.

PR #5649, opened 2026-06-12
Fall back to getRandomValues when crypto.randomUUID is unavailable (ActionCable)

### LINKED ISSUE #5648: ActionCableLink crashes on insecure origins: crypto.randomUUID is undefined outside secure contexts (regression in 1.15.0)
## Describe the bug

Since `graphql-ruby-client` 1.15.0 (#5642 / #5643), the `ActionCableLink` constructor evaluates:

```ts
this.createChannelId = options.createChannelId || crypto.randomUUID.bind(crypto)
```

`crypto.randomUUID` is only exposed in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) (HTTPS or localhost). On a plain-HTTP origin (LAN IP, internal/self-hosted deployments) it is `undefined`, so the constructor throws:

```
TypeError: Cannot read properties of undefined (reading 'bind')
```

## Versions

- `graphql-ruby-client`: 1.15.0 (1.14.9 unaffected)
- Browser: any (Chrome, Firefox, Safari) on a non-secure origin

## Steps to reproduce

1. Serve an app using `ActionCableLink` over plain HTTP on a non-localhost host.
2. Open it in any browser — `new ActionCableLink(...)` throws.

## Expected behavior

The link should construct successfully in any context, keeping the high-entropy channel ids introduced for #5639.

## Proposed fix

Fall back to a 128-bit `crypto.getRandomValues` id when `randomUUID` is unavailable (as suggested in #5639), e.g.:

```ts
function defaultChannelId(): string {
  if (crypto.randomUUID) {
    return crypto.randomUUID()
  }
  return Array.from(crypto.getRandomValues(new Uint8Array(16)), (b) => b.toString(16).padStart(2, "0")).join("")
}
```

`crypto.getRandomValues` is available in all contexts, so entropy is preserved everywhere. Happy to submit a PR if this approach works for you.

## Workaround

Pass a custom `createChannelId` in the constructor options.

domenicofalco:
Fixes #5648

## Problem

Since `graphql-ruby-client` 1.15.0 (#5642 / #5643), the ActionCable integrations evaluate `crypto.randomUUID.bind(crypto)` at construction time. `crypto.randomUUID` is only exposed in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) (HTTPS or localhost), so on plain-HTTP origins (LAN IPs, internal/self-hosted deployments) this throws:

```
TypeError: Cannot read properties of undefined (reading 'bind')
```

## Solution

As discussed in #5648, add a shared `defaultChannelId` helper that prefers `crypto.randomUUID()` and falls back to a 128-bit hex id built with `crypto.getRandomValues`, which is available in all contexts. This keeps the entropy introduced for #5639 everywhere.

Applied to the three call sites: `ActionCableLink`, `ActionCableSubscriber`, `createActionCableHandler`.

## Tests

- New `defaultChannelIdTest` covering: native `randomUUID` preferred, fallback format on insecure contexts, uniqueness over 1000 ids

### DIFF
--- a/javascript_client/src/subscriptions/ActionCableLink.ts
+++ b/javascript_client/src/subscriptions/ActionCableLink.ts
@@ -1,6 +1,7 @@
 import { ApolloLink, Observable, FetchResult, Operation, NextLink } from "@apollo/client/core"
 import type { Consumer } from "@rails/actioncable"
 import { print } from "graphql"
+import defaultChannelId from "./defaultChannelId"
 
 type RequestResult = FetchResult<{ [key: string]: any; }, Record<string, any>, Record<string, any>>
 type ConnectionParams = object | ((operation: Operation) => object)
@@ -33,7 +34,7 @@ class ActionCableLink extends ApolloLink {
     this.actionName = options.actionName || "execute"
     this.connectionParams = options.connectionParams || {}
     this.callbacks = options.callbacks || {}
-    this.createChannelId = options.createChannelId || crypto.randomUUID.bind(crypto)
+    this.createChannelId = options.createChannelId || defaultChannelId
   }
 
   // Interestingly, this link does _not_ call through to `next` because
--- a/javascript_client/src/subscriptions/ActionCableSubscriber.ts
+++ b/javascript_client/src/subscriptions/ActionCableSubscriber.ts
@@ -1,5 +1,6 @@
 import printer from "graphql/language/printer"
 import registry from "./registry"
+import defaultChannelId from "./defaultChannelId"
 import type { Consumer } from "@rails/actioncable"
 
 interface ApolloNetworkInterface {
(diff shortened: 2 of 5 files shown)


domenicofalco:
Hi @rmosolgo - this should be the PR that fixes the issue. The failing jobs seem unrelated with my changes. Can you kindly help with that please? Thank you so much

rmosolgo:
Hey, thanks for addressing this. I agree the failures are unrelated (it looks like rugged isn't compatible with Ruby-head: https://github.com/libgit2/rugged/issues/1005, not sure about the system test failure)

rmosolgo:
🚢  in graphql-ruby-client v1.15.1. Please let me know if it gives you any more trouble. Thanks again!

PR #5651, opened 2026-06-12
Fix ActionCable test for Rails head

rmosolgo:
My test mocks (etc...) broke due to https://github.com/rails/rails/pull/50979

Update them accordingly. I tried to use some of the new testing hooks but they inline everything instead of going async.


rmosolgo:
Ruby head is still broken because of the rugged issue

PR #5659, opened 2026-07-07
Apply directive argument defaults

dpep:
### Problem

`Directive#initialize` validates every defined argument, but treats omitted ones as `nil` without consulting `default_value`. Any directive with a defaulted non-null argument raises when the argument is omitted at a usage site, even though the SDL is valid per spec:

```ruby
GraphQL::Schema.from_definition <<~GRAPHQL
  directive @flag(enabled: Boolean! = true) on OBJECT

  type Query @flag {
    id: ID
  }
GRAPHQL
# => GraphQL::Schema::Directive::InvalidArgumentError:
#    @flag.enabled on Query is invalid (nil): Expected value to not be null
```

Notably, this makes `from_definition` unable to load real Apollo Federation supergraph SDL — the join v0.3 spec's `@join__type` declares `extension: Boolean! = false, resolvable: Boolean! = true, ...`, so every plain `@join__type(graph: X)` usage raises.

### Fix

In the validation loop, fall back to the argument's declared default when the keyword is omitted — mirroring what `Argument#coerce_into_values` already does on the coercion path.

### Tests

Two regression tests: a Ruby-land directive attached without its defaulted argument, and the SDL repro above. Both fail without the fix and pass with it.

### DIFF
--- a/lib/graphql/schema/directive.rb
+++ b/lib/graphql/schema/directive.rb
@@ -170,7 +170,7 @@ module GraphQL
             end
           elsif arg_defn.default_value?
             value = arg_defn.default_value
-            graphql_value = value.nil? ? nil : arg_type.coerce_isolated_result(value)
+            graphql_value = arg_type.coerce_isolated_result(value) unless value.nil?
           else
             value = graphql_value = nil
           end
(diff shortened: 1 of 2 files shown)


rmosolgo:

PR #5660, opened 2026-07-07
Allow instances of subclasses of wrapper class to be pattern matched via #deconstruct_keys

itojum:
Added `#deconstruct_keys` to the `GraphQL::Schema::Wrapper` class. This allows pattern matching against `List` and `NonNull`, which are subclasses of the `Wrapper` class.

This allows using pattern matching for concise access to internal elements when dealing with nested structures.

Example: 
```rb
obj = GraphQL::Schema::List.new(GraphQL::Schema::NonNull.new("Matched!"))
case obj
in GraphQL::Schema::List(of_type: GraphQL::Schema::NonNull(of_type: message))
  p message # => "Matched!"
else
  p "No match"
end
```

I also referred to the PR that adds #deconstruct_keys to the InputObject.
https://github.com/rmosolgo/graphql-ruby/pull/5170

### DIFF
--- a/lib/graphql/schema/wrapper.rb
+++ b/lib/graphql/schema/wrapper.rb
@@ -25,6 +25,10 @@ module GraphQL
       def ==(other)
         self.class == other.class && of_type == other.of_type
       end
+
+      def deconstruct_keys(_keys)
+        { of_type: of_type }
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


rmosolgo:
Looks good, thanks for this improvement :+1:

PR #5662, opened 2026-07-09
Fix Dataloader Ruby version check

ydah:
Fix `GraphQL::Dataloader#with` to use `RUBY_VERSION` when selecting the legacy keyword argument implementation for CRuby before Ruby 3.


rmosolgo:
Oops ... how did this ever work? How did you run into this? 

Thanks for fixing it!

PR #5663, opened 2026-07-09
Fix Schema union memberships lookup

ydah:
Fix a typo in `Schema.union_memberships` when collecting inherited union memberships without a specific type.


rmosolgo:
Hey, thanks for sharing this fix. How did you run into this problem?

 I'd like to include a test in the project to make sure this code path is properly tested in the future!

ydah:
Thanks! I found this while reviewing the inheritance paths in `Schema.union_memberships`, rather than through a production failure. The lookup key didn't match the `:union_memberships` name used elsewhere.

I've added a regression test that subclasses `Dummy::Schema` and calls `union_memberships` without a type. It verifies that the inherited memberships are included.

rmosolgo:
Got it, that makes sense. Thanks again for this fix!

PR #5664, opened 2026-07-09
Fix Query variables storage after max errors

ydah:
Fix `GraphQL::Query::Variables` so `@storage` remains a hash when variable validation stops after reaching `validate_max_errors`.

This is a narrow edge case: normal execution already returns validation errors, but callers may still inspect `result.query.variables` for logging, instrumentation, or error handling. Previously, the `break` inside `each_with_object` could make the whole expression return `nil`, causing delegated methods like `length`, `to_h`, or `[]` to raise `NoMethodError`.


rmosolgo:

PR #5665, opened 2026-07-09
Exclude Jekyll dependencies from test CI

ydah:
follow up: https://github.com/rmosolgo/graphql-ruby/actions/runs/29028102605/job/86153202181


rmosolgo:
Thanks for this improvement!

PR #5668, opened 2026-07-11
Apply query token limit in Schema.validate

ydah:
`Schema.validate(string)` parsed query strings without passing the schema's `max_query_string_tokens` setting. This allowed validation-only callers to bypass the configured token limit.

```ruby
LimitedSchema.validate("{ value value }")
# Before => validation proceeds despite exceeding the token limit
# After => raises GraphQL::ParseError
```

This passes `max_query_string_tokens` to `GraphQL.parse` and adds regression coverage.


rmosolgo:
Makes sense to me -- thanks!

PR #5669, opened 2026-07-11
Share lazy job scheduling across dataloaders

ydah:
This refactors lazy job scheduling shared by `Dataloader` and `AsyncDataloader`.

- Move selection and enqueueing of the next lazy depth into the base `Dataloader`
- Use a block for implementation-specific Fiber/Task execution
- Keep Async queue lifecycle management isolated in `run_lazy_jobs`
- Remove duplicated code and stale TODOs 

No behavior change is intended.

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -212,7 +212,9 @@ module GraphQL
 
           if !@lazies_at_depth.empty?
             with_trace_query_lazy(trace_query_lazy) do
-              run_next_pending_lazies(job_fibers, trace)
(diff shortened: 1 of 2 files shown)


rmosolgo:
Looks good to me! 

In general I have become wary of passing blocks in GraphQL-Ruby's runtime methods, because of the performance overhead they add (like the memory footprint of the local binding). But if this turns out to be problem, we could probably replace the `yield` call with a method call like `self.after_lazies(...)`, and then implement that method in one way or another in the two dataloader classes.

PR #5670, opened 2026-07-11
Remove obsolete NoMethodError compatibility

ydah:
My understanding is that this gem supports Ruby 2.7 and later, so support for version 2.2 is likely no longer necessary.
https://github.com/rmosolgo/graphql-ruby/blob/c30cc15b68fbb998aaf8591d5d17f8f3071fd114/graphql.gemspec#L16

I have tried removing the code that appears to be obsolete. What do you think?

### DIFF
--- a/lib/graphql/execution/interpreter/runtime.rb
+++ b/lib/graphql/execution/interpreter/runtime.rb
@@ -765,8 +765,7 @@ module GraphQL
 
                 response_list
               rescue NoMethodError => err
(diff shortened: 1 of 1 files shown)


rmosolgo:
🎉

rmosolgo:
Btw, I'm open to bringing this gem up to 3.0 -- removing support for 2.7. I think there some conditionals for that, too. I'd welcome your contribution if you're interested!

PR #5673, opened 2026-07-15
Fix operation directive finalizer registration

ydah:
The `path` in `result.path = path` does not exist in this scope, resulting in a `NameError`. 

Furthermore, the order of arguments in `add_finalizer(query, result, nil, data)` is the reverse of the defined order: `query`, `result_value`, `key`, and `finalizer`. Simply correcting the undefined variables will still result in an incorrect registration.


rmosolgo:
Thanks for fixing this!

PR #5674, opened 2026-07-15
Forward Relay subscription errors

ydah:
Fix Relay subscription errors being silently discarded.

The shared Relay subscription handler referenced `observer.error` without calling it. This prevented errors from reaching Relay consumers using ActionCable, Pusher, or Ably. This change calls `observer.error(error)` and adds regression coverage verifying that transport errors are forwarded unchanged to the Relay observer.


rmosolgo:
Oops, it was probably supposed to be `onError: observer.error` 🤦 Thanks for this fix!

rmosolgo:
🚢  in graphql-ruby-client v1.15.2 !

PR #5675, opened 2026-07-15
Avoid duplicate selection step enqueueing

ydah:
Prevent `SelectionsStep` from enqueueing field steps from previous selection groups more than once. `continue_selections` was initialized outside the group loop, so each group retained and re-enqueued every step collected from earlier groups.

For example, two groups containing `first` and `second` were enqueued as `[first, first, second]`. This change scopes `continue_selections` to each group, producing `[first,
  second]`.

### DIFF
--- a/lib/graphql/execution/selections_step.rb
+++ b/lib/graphql/execution/selections_step.rb
@@ -30,10 +30,10 @@ module GraphQL
       def call
         @all_selections = [{}, (prototype_result = {})]
         @runner.gather_selections(@parent_type, @selections, self, self.query, @all_selections, @all_selections[1], into: @all_selections[0])
+        continue_selections = []
         i = 0
         l = @all_selections.length
         while i < l
-          continue_selections = []
           grouped_selections = @all_selections[i]
           selections_prototype_result = @all_selections[i + 1]
           if (directives_owner = grouped_selections.delete(:__node))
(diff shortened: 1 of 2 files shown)


rmosolgo:
Interesting ... I think the CI failures are real here. IIRC this array has something do about directives which pause execution. But I see what you mean about this array accumulating things outside of scope. 

Another possible fix could be here: 

https://github.com/rmosolgo/graphql-ruby/blob/8b9f621c196d9ed08aee5fa7451887b929e84673/lib/graphql/execution/selections_step.rb#L86-L88

Maybe we could do a `continue_selections.clear` after the steps are all passed along. (That would also let us keep using the same Array instance.)

ydah:
Thanks! I tried `continue_selections.clear` after enqueueing, but the three directive feature tests still failed.
The issue is that `NullDataloader#add_step` executes synchronously. Earlier field steps ran immediately, then a later group’s `replace(prototype_result)` reset those results to `nil`. The previous duplicate enqueueing happened to restore them.
I moved the enqueue loop after all selection groups have been processed instead. This avoids duplicate enqueueing while ensuring all prototype updates are complete before any field step runs. I pushed the fix in a824a8a554.

rmosolgo:
That's great. Thanks again for hunting this down and contributing it!

PR #5683, opened 2026-08-04
Handle SystemStackError during query execution

ydah:
A sufficiently deep but relatively small query can pass `max_query_string_tokens` and raise an uncaught `SystemStackError` during interpreter execution.

For example, a 4,518-byte query with a selection depth of 450 passes the generated schema's 5,000-token limit but can exhaust the Ruby stack. Generated schemas currently don't configure `max_depth`, so this may propagate as an unhandled exception.

Validation already delegates `SystemStackError` to `Schema#query_stack_error`, but the execution phase did not.


rmosolgo:
Looks good to me. It's nice that `Execution::Next` could handle the depth without raising an exception ... that was the point!

PR #5684, opened 2026-08-05
Load mounted routes before integration tests

ydah:
Load the dummy application's routes before `rails/test_help` includes the mounted route helpers.

### Background

The Rails master build fails the dashboard integration tests with:

```text
NameError: undefined local variable or method 'graphql_dashboard'
```

This also reproduces on the current graphql-ruby master branch, independently of #5683.

With Rails master, the application routes may not be loaded yet when ActionDispatch::IntegrationTest includes Rails.application.routes.mounted_helpers. As a result, the graphql_dashboard mounted helper is unavailable in the integration tests.


rmosolgo:
Hey, thanks for fixing this. Its not the first time I've had weirdness with engine routes in various environments (https://github.com/rmosolgo/graphql-ruby/pull/5524).

PR #5685, opened 2026-08-06
🐛 `RelationConnection` loads the same relation twice under `AsyncDataloader` (2.6.7)

benschwarz:
Given a query like:

```
{ posts(first: 20) { edges { node { id } } pageInfo { endCursor } } }
```
Both fields call load_nodes. 2.6.7 runs them concurrently, earlier releases didn't.

```
@nodes ||= limited_nodes.to_a   # relation_connection.rb:224
```

Fiber one waits on the DB before setting @nodes. Fiber two sees empty and loads the same relation again. Whichever finishes first clears ActiveRecord's load state from under the other.

This results in:

`NoMethodError: undefined method 'instantiate' for nil`, plus a duplicate query on every paginated connection.

### DIFF
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -220,8 +220,19 @@ module GraphQL
       # Load nodes after applying first/last/before/after,
       # returns an array of nodes
       def load_nodes
-        # Return an array so we can consistently use `.index(node)` on it
-        @nodes ||= limited_nodes.to_a
+        return @nodes if @nodes
+
+        if @loading_nodes
+          context.dataloader.yield until @nodes
+          return @nodes
+        end
+
+        @loading_nodes = true
+        begin
+          @nodes = limited_nodes.to_a
+        ensure
+          @loading_nodes = false
+        end
       end
     end
   end


rmosolgo:
Hey, thanks for reporting this and sorry for the trouble 😖 


I'm hesitant to address this with `dataloader.yield` because I _expect_ this to all happen inside GraphQL-Ruby's execution steps -- without handing off to Dataloader to fetch results from Dataloader sources. 

Do you have a guess why the Fiber is yielding here? Is this a custom connection implementation?

benschwarz:
Hey @rmosolgo, fair enough. I am not sure if this is the right fix either, but I was able to get my test suite running with it -- good to open the discussion though.

I'm on 1.6.3, and there are seemingly many issues with async dataloader where db connections get used by old/finished/crashed fibres. On 1.6.3 we have a custom `ConnectionReleasingDataloader < GraphQL::Dataloader::AsyncDataloader` that calls `release_connection` and also reclaims abandoned fibres. It's non ideal but is working well. 

Given the updates in 1.6.7 I thought it was worthwhile to see if I could get any joy from it. In using the code in this PR, I was able to upgrade, and also remove the connection releasing dataloader 😇

rmosolgo:
Thanks for sharing a bit more. I'd love to hear more details (eg error messages, backtraces) about what goes wrong without manual connection handling. Maybe there's a clue in there to an underlying bug! 

Also, if you find any clues about why `load_nodes` would cause a fiber yield, please do share -- I'd love to understand this better and find a solid fix.

drhops:
Thanks for the PR @benschwarz and discovered this after opening https://github.com/rmosolgo/graphql-ruby/pull/5708, which I believe is the same root cause.  @rmosolgo I believe the `limited_nodes.to_a` yields when it's an ActiveRecord relation executing a remote query.  I proposed an alternate fix in that PR using a mutex instead of a yield.

rmosolgo:
Ok, yes, I think that's what I was missing: previously, `AsyncDataloader` used plain ol' Ruby fibers for field execution, but now it uses `Async::Task`, which means that any IO can yield -- not a `dataloader.yield` but at the Ruby level. 

I'm pretty sure ActiveRecord has locks of its own, for example: 

https://github.com/rails/rails/blob/69275f94d2a6181d82afe82f04170bbbc26cc646/activerecord/lib/active_record/relation.rb#L1478-L1484

I wonder if it would be possible to reorganize this code so that the in-Ruby parts memoize their results in instance variables, then call out to ActiveRecord, which would use its own locks to manage actual DB access. @drhops, thanks for demonstrating the other approach to this, I pulled your branch and I'm trying to demonstrate a _real _failure with `with_active_record_log`.

rmosolgo:
Thanks for digging in on this. I think #5708 will fix this same problem in another way. Please let me know if try out a version which includes that patch and still run into trouble! 

I'd also be interested to hear about the other issues you've encountered -- please open a new issue with details about them if you get a chance.

benschwarz:
Thanks @rmosolgo @drhops, glad my report was useful. Nice to see a fix go in. I'll try the next point release when its available.

PR #5687, opened 2026-08-08
Isolate trace mode test state

ydah:
`trace_modes_spec.rb` currently modifies `GraphQL::Schema` directly and attempts to restore its trace state afterward. Because the cleanup clears the trace mode cache but leaves `GraphQL::Schema::DefaultTrace` defined, the test order can cause a later schema construction to emit a constant redefinition warning. When that warning is captured by `GraphQL::Schema::Loader`'s warning assertion, the `mongoid_8` CI job fails intermittently.

This change runs the legacy tracer inheritance test against an anonymous base schema instead. It preserves coverage for both existing and newly-created child schemas while avoiding global state changes and test-order dependencies.

### DIFF
--- a/spec/graphql/tracing/trace_modes_spec.rb
+++ b/spec/graphql/tracing/trace_modes_spec.rb
@@ -127,39 +127,24 @@ describe "Trace modes for schemas" do
     assert_equal :was_configured, res.context[:configured_option]
   end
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for fixing this!

PR #5688, opened 2026-08-08
Restrict dashboard schema selection

ydah:
`GraphQL::Dashboard` currently resolves the `schema` query parameter before the schema configured at mount time and calls `constantize` on any string value. A request such as `? schema=Kernel` can therefore trigger arbitrary constant resolution, causing repeated `NameError` exceptions and potentially exposing information about constants in the application.

This also means that `mount GraphQL::Dashboard, schema: "MySchema"` does not reliably enforce the configured schema because the query parameter can override it.

The mount-time schema configuration is now treated as trusted configuration. Query parameters are only accepted when they match one of the configured schemas, and unknown values fall back to the configured default schema.
This prevents arbitrary constantization while preserving the intended mount configuration. Tests cover both allowed schema selection and rejected arbitrary values.

### DIFF
--- a/lib/graphql/dashboard.rb
+++ b/lib/graphql/dashboard.rb
@@ -10,6 +10,9 @@ module Graphql
   # @example Mounting the Dashboard in your app
   #   mount GraphQL::Dashboard, at: "graphql_dashboard", schema: "MySchema"
   #
+  # Pass an array to allow selecting from multiple schemas with the `schema` query parameter.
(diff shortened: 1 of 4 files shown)


rmosolgo:
Hey, thanks for this improvement. 

This is technically a breaking change, but I don't believe that `?schema=...` is documented anywhere, so I'm open to adding it to a patch release. I'll include a note in the changelog.

PR #5689, opened 2026-08-09
Migrate documentation to rdoc

### LINKED ISSUE #5576: Migrate to RDoc
I love using the new Ruby docs with the Aliki theme. GraphQL-Ruby currently uses YARD and a hand-rolled jekyll site for documentation, but I'm thinking about migrating to RDoc.

This would get me a lot for free: search, category browser, table of contents, auto-linking, decent css

Some thoughts: 

- Move as much "guides" content into actual source code 
- Preserve redirects from old urls
- Transform YARD to RDoc -- is there an automatic transformer? https://github.com/AaronC81/sord ? 
- How can I bring-my-own GraphQL syntax highlighting?

ydah:
Fixes: https://github.com/rmosolgo/graphql-ruby/issues/5576

This PR completes the migration of the GraphQL Ruby documentation site from YARD/Jekyll to RDoc/Aliki.

It adds the RDoc build and CI checks, converts guide metadata and YARD comments, moves API-specific guide content into source comments, preserves legacy documentation URLs, and removes the legacy YARD pipeline. The generated API documentation is checked against the pre-migration public API inventory to detect unexpected coverage changes.

The generated site also includes GraphQL syntax highlighting, local-file-friendly image links, correct README rendering, nested guide navigation with page titles, and complete development guide table-of-contents links. No GraphQL runtime behavior is changed.

Validated with `bundle exec rake docs:check`, the documentation generator specs, the link checker, the RDoc reference checker, and the GraphQL highlighter tests.


rmosolgo:
Hey, thanks so much for taking a crack at this! I just pulled the branch and built the docs locally. In general I think this is great, but I'd like to get a few more things before replacing the YARD docs with this: 

- convert `@api private` to `:nodoc:`; right now they're rendered like this:

    image


- Type deifnitions, somehow Prism has them in its RDoc: https://docs.ruby-lang.org/en/master/Prism/CallNode.html I would love to convert our YARD to types like that 

I'm going to continue reviewing but I thought I'd share those ideas now to see what you think.

ydah:
Thank you for the feedback! I've addressed both points:

* Converted `@api private` to `:nodoc:`
* Updated the type definitions so they're rendered properly in RDoc, similar to Prism

Thank you for catching these.

rmosolgo:
Hey, thanks again for your work on this so far. I pulled the latest and took it for a spin today. I think a good approach would be to get this _good enough_ and then continue iterating on master. Here are the things I think we need: 

1) Ordered guides. This is currently implemented with `index: ...` in Jekyll frontmatter, but the new rendering puts them in alphabetical order. This means that `Overview` is no longer the first entry on the list, but it should be. I'm not sure how this should be implemented. 
2) Call-outs for documented features which ship in GraphQL-Pro or GraphQL-Enterprise. This currently looks like this:

    image

    But the new implementation doesn't include any indication that `@defer` isn't in GraphQL-Ruby OSS. I think this will be confusing and disappointing for people :S 
    
    So we need _something_ for that. A hard-coded indicator of some kind would work for me. 


When those two are implemented, I'll be ready to merge this PR. 

I think there will be a few things to address after that: 

- Remove migration-related scripts that we don't need anymore 
- Somehow improve local tooling to include file watching? 
- The type annotation support is still not great :S I'm not sure what to do about that though. The way that you migrated YARD annotations seems like our best bet in RDoc because the inline-RBS notation doesn't support inline comments. (I really like how, in YARD, the argument name, type, and description all go together.)

PR #5691, opened 2026-08-09
Restrict subscription timestamp deserialization

ydah:
This PR prevents arbitrary constant resolution during subscription payload deserialization.

Previously, timestamp payloads used `Object.const_get` to resolve the class name and then called `strptime` on it. If an untrusted payload reached this code path, it could invoke methods on an arbitrary top-level class.

Timestamp deserialization is now restricted to `Date`, `DateTime`, `Time`, and `ActiveSupport::TimeWithZone` when available. Unsupported classes raise `ArgumentError`.

A regression test verifies that `strptime` is not invoked on arbitrary classes.

### DIFF
--- a/lib/graphql/subscriptions/serialize.rb
+++ b/lib/graphql/subscriptions/serialize.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require "date"
 require "set"
 module GraphQL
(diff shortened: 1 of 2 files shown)


rmosolgo:
Great, thanks for this improvement!

PR #5692, opened 2026-08-10
Fix lexer positions for multibyte input

ydah:
This PR fixes incorrect lexer error positions when GraphQL input contains multibyte characters.

`GraphQL::Language::Lexer` stores scanner positions as byte offsets, but line and column calculation used character-based string indexing. As a result, Japanese comments or string literals before a syntax error could cause the reported location to point to the wrong line and column.

The lexer now uses byte slices for position calculations while preserving character-based column counts. Invalid UTF-8 prefixes are also handled safely.

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -175,11 +175,13 @@ module GraphQL
       end
 
       def line_number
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for this improvement!

PR #5699, opened 2026-08-12
Support JSON 3

byroot:
`quirks_mode` hasn't been a valid parsing option since `json 2.0.0` released circa 2011.

On `json 3.x` invalid arguments are rejected, causing an `ArgumentError`.

It's likely safe to assume no-one runs with `json 1.x`.

### DIFF
--- a/lib/graphql/language.rb
+++ b/lib/graphql/language.rb
@@ -31,7 +31,7 @@ module GraphQL
 
         "[#{serialized_array}]"
       else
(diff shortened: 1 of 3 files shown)


rmosolgo:
I think this option was originally added to support serializing non-Hash/Array values at these call sites, and had _something_ to do with non-stdlib JSON implementations. Given that the built-in JSON library isn't going to accept this option, I guess I'll have to find another approach if other libraries still need this option.

Thanks for updating this!

PR #5701, opened 2026-08-12
Intern SDL description strings at AST node construction

dersam:
GraphQL::Language::BlockString.trim_whitespace returns the exact String objects that str.split produced, so every """...""" description in every SDL document parsed at boot is a fresh, unshared String retained for the life of the process twice over: once by the Language::Nodes::*Definition AST node and once by the Schema::Field / Argument / EnumValue built from it.

Intern the description `generate_initialize`, which generates the initialize method for every definition node class that declares `attr_reader :description`. Identical descriptions across SDL documents collapse onto one frozen fstring object.

**Testing on a large internal app at Shopify drops description strings from 30MB to 3MB in memory.**


rmosolgo:

PR #5704, opened 2026-08-16
Optimize repeated FieldsWillMerge comparisons

ydah:
Improve the performance of `FieldsWillMerge` validation for queries containing many repeated composite fields and shared fragment selections.

Previously, the validator could repeatedly compare the same expanded fragment fields for every pair of parent fields. This caused nested Cartesian comparisons and severe superlinear runtime growth before query execution or authentication.

This change:

- groups fields with identical sub-selections and compares only one representative from each group
- memoizes comparisons between equivalent expanded field groups
- keeps mutually exclusive and non-exclusive comparisons separate
- adds a regression test that verifies a divergent conflict is still detected without reaching the default validation timeout

### Benchmark

Measured using a query with repeated composite fields and a shared fragment containing repeated fields.

| Fields | Query size | Before | After | Speedup |
| ---: | ---: | ---: | ---: | ---: |
| 100 | 2,320 bytes | 193.6 ms | 3.14 ms | 61.7x |
| 200 | 4,552 bytes | 1,438.4 ms | 3.72 ms | 386.7x |
| 400 | 9,000 bytes | 11,732.5 ms | 8.68 ms | 1,351.7x |
| 800 | 17,864 bytes | did not finish within 30 seconds | 20.08 ms | — |

The 400-field case is reduced by approximately 99.93%.

### DIFF
--- a/lib/graphql/static_validation/rules/fields_will_merge.rb
+++ b/lib/graphql/static_validation/rules/fields_will_merge.rb
@@ -15,6 +15,8 @@ module GraphQL
       # separately (which leads to exponential recursion through nested fragments),
       # we flatten all fragment spreads into a single field map and compare within it.
       NO_ARGS = GraphQL::EmptyObjects::EMPTY_HASH
+      EXCLUSIVE_COMPARISON = 1
+      NONEXCLUSIVE_COMPARISON = 2
 
       class Field
         attr_reader :node, :definition, :owner_type, :parents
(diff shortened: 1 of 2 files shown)


swalkinshaw (approved):
Looks like a good optimization to me

rmosolgo:
Thanks for taking a look, @swalkinshaw . The only thing that stood out to me was the use of `to_query_string`, which creates a big string for large nested selections. But I think it's pretty optimized in terms of object allocations, and the speedup here is worth it. Thanks for sharing this, @ydah!

PR #5706, opened 2026-08-19
exec-next: Add method: to dataload shorthands

OuYangJinTing:
## Motivation

A very common pattern in our schema is resolving a scalar by first batch-loading an association, then calling a method on the loaded associated object.

Without this PR, every such field requires a custom `def self.<field>` batch resolver. For example, in our `UserType` we end up with a lot of boilerplate like this:

```ruby
# app/graphql/types/user_type.rb
field :age, String, null: true, resolve_batch: true

def self.age(objects, context)
  dataload_all_associations(context, objects, :verified_info).map { it&.age }
end

field :country, String, null: true, resolve_batch: true

def self.country(objects, context)
  dataload_all_associations(context, objects, :latest_device).map { it&.country }
end

def self.dataload_all_associations(context, records, association_name, scope: nil)
  source = if scope
             context.dataloader.with(GraphQL::Dataloader::ActiveRecordAssociationSource, association_name, scope)
           else
             context.dataloader.with(GraphQL::Dataloader::ActiveRecordAssociationSource, association_name)
           end
  source.load_all(records)
end
```

This becomes repetitive quickly and makes the type file mostly boilerplate.

With the `method:` option on `dataload:` shorthands, the same fields can be written declaratively:

```ruby
field :age, String, null: true,
      dataload: { association: :verified_info, method: :age }

field :country, String, null: true,
      dataload: { association: :latest_device, method: :country }
```

It keeps the dataloader batching, removes the need for one resolver per field, and makes the schema definition a lot more readable. Because `method:` works for `with:`, `association:`, and `model:` shorthands, it covers most of our real-world cases.

## Summary

- Adds a `method:` option to hash-style `dataload:` field configs under `Execution::Next`. When present, the loaded value's public method is called before returning it, e.g. `dataload: { association: :author, method: :name }` is equivalent to `dataload_association(:author)&.name`.
- Works with all three hash-style shorthands (`with:`, `association:`, `model:`).
- Updated `guides/execution/next.md` with examples for each shorthand.

## Tests

Added specs covering `method:` for the source shorthand, the Rails association shorthand, and the Rails model shorthand.

### DIFF
--- a/guides/execution/next.md
+++ b/guides/execution/next.md
@@ -170,6 +170,8 @@ end
 
 `Execution::Next` supports field configuration shorthands for common dataloader usage. Under the hood, these make sure data fetching is batched and cached.
 
+`method:` can be added to any hash-style `dataload:` config to call a public method on the loaded value.
+
 #### Sources
 
 Use a custom dataloader source from your application:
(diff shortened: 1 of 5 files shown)


rmosolgo:
Thanks for this improvement! I agree this is a nice convenience.

PR #5709, opened 2026-08-27
Avoid freezing execution results during finalization

ydah:
Ruby head currently freezes the nested array in `Array#flatten!`'s single-array fast path:

```ruby
result = []
[result].flatten!
result.frozen? # => true
```

`GraphQL::Execution::Finalize` used flatten! while locating errors in nested list results. This unintentionally froze the execution result array, then null propagation attempted to update it and raised `FrozenError`.

This change uses non-destructive flatten instead. It preserves the existing recursive traversal without freezing result arrays.

The existing `[Test], [Test!],` and `[Test!]!` null-propagation cases fail before this change and pass afterward on Ruby 4.1.0dev. They also pass on Ruby 4.0.

### DIFF
--- a/lib/graphql/execution/finalize.rb
+++ b/lib/graphql/execution/finalize.rb
@@ -28,7 +28,7 @@ module GraphQL
           targets = [data]
           while (part = err_path.shift)
             targets.map! { |t| t[part] }
-            targets.flatten!
+            targets = targets.flatten
           end
 
           targets.each_with_index do |target, idx|


rmosolgo:
Thanks for tracking down this test failure. 

Do you know if this `.freeze` is intentional or accidental? 

With `.flatten` we're allocating a new Array -- which I'd rather not, if we can help it.

ydah:
Yes, this is new behavior introduced in Ruby head by ruby/ruby#18438. Ruby 4.0 and earlier don't freeze the nested Array.

The optimization reuses the nested Array as `flatten!`'s internal result, then the existing internal `rb_ary_freeze(result)` ends up freezing that caller-owned Array. That looks like an accidental behavior change rather than intentional API behavior.

I agree that allocating another Array in GraphQL-Ruby isn't ideal. I'll report this upstream to Ruby; if it's fixed there, we can keep `flatten!` here and close this PR.

ydah:
This was resolved by the following PR, so I'm closing this PR!
- https://github.com/ruby/ruby/pull/18550

rmosolgo:
Rock on! Thanks for investigating this.

PR #5710, opened 2026-08-27
Handle out-of-range Unicode escapes as parse errors

ydah:
An out-of-range Unicode escape currently causes `GraphQL::Language::Lexer` to raise a raw `RangeError`:

```ruby
MySchema.execute('{ echo(input: "\\u{FFFFFFFFFF}") }')
# RangeError: pack(U): value out of range
```

This happens when replace_escaped_characters_in_place passes the decoded codepoint to Array#pack("U"). Since the exception bypasses GraphQL's parse error handling, applications may return an HTTP 500 for an invalid query.

This PR catches that RangeError and reports it through the lexer's existing raise_parse_error path. The same query is now returned as a normal GraphQL parse error.

A regression test covers Unicode escapes above the valid codepoint range.

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -163,7 +163,11 @@ module GraphQL
           if !str.valid_encoding? || !str.match?(VALID_STRING)
             raise_parse_error("Bad unicode escape in #{str.inspect}")
           else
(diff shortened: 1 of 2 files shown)


rmosolgo:
Cool, thanks :+1:

PR #5711, opened 2026-08-27
Reraise unexpected lexer argument errors

ydah:
`GraphQL::Language::Lexer#advance` currently rescues every `ArgumentError`, but only handles errors whose message exactly matches `"invalid byte sequence in UTF-8"`. Any other `ArgumentError` is silently swallowed and `advance` returns `nil`.

The exact message is produced while scanning an invalidly encoded string, so matching it also depends on Ruby's exception wording.

This PR validates the query string's encoding when the lexer is initialized and removes the broad rescue from `advance`. Invalid UTF-8 continues to raise the existing `GraphQL::ParseError`, while unrelated `ArgumentError`s are propagated normally.

### DIFF
--- a/lib/graphql/language/lexer.rb
+++ b/lib/graphql/language/lexer.rb
@@ -9,6 +9,9 @@ module GraphQL
         end
         @string = graphql_str
         @filename = filename
+        if !@string.valid_encoding?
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for this improvement!

PR #5712, opened 2026-08-27
Support short braced Unicode escapes

ydah:
GraphQL's braced Unicode escape syntax accepts one or more hexadecimal digits. However, both GraphQL-Ruby lexers currently require at least four digits, so valid escapes such as `\u{A}` and `\u{48}` are rejected as malformed strings.

This PR updates the Ruby and Ragel lexer patterns to accept one or more digits inside braced Unicode escapes, then regenerates the C lexer. The unbraced `\uXXXX` form continues to require exactly four digits.

Shared regression tests verify that `\u{A}` and `\u{48}` are decoded correctly by both the Ruby and C lexers. Empty and non-hexadecimal escapes continue to be rejected.


rmosolgo:
Thanks for this improvement!

PR #5713, opened 2026-08-27
Reject non-finite Float values

ydah:
GraphQL Float values must be finite, but `GraphQL::Types::Float` currently accepts any `Numeric` value after calling `to_f`.
This allows non-finite values through both input and result coercion:

```ruby
Schema.execute("{ echo(value: 1e400) }")
# => Infinity is accepted as input

result = Schema.execute("{ nan }")
JSON.generate(result.to_h)
# => JSON::GeneratorError: NaN not allowed in JSON
```

Variables containing `Float::NAN` or `infinity` are also accepted. For result values, the failure occurs later during HTTP response serialization instead of being reported by GraphQL-Ruby.

This PR adds finite-value checks to `Float` coercion, following the existing Int error-handling pattern. Invalid inputs are routed through `Schema#type_error` using `FloatDecodingError`, while invalid result values raise `FloatEncodingError` with field and response-path context.

Variable validation errors and serialized GraphQL values also represent non-finite floats as strings, keeping error responses JSON-safe.

### DIFF
--- a/lib/graphql.rb
+++ b/lib/graphql.rb
@@ -93,6 +93,8 @@ This is probably a bug in GraphQL-Ruby, please report this error on GitHub: http
   autoload :AnalysisError, "graphql/analysis_error"
   autoload :CoercionError, "graphql/coercion_error"
   autoload :InvalidNameError, "graphql/invalid_name_error"
+  autoload :FloatDecodingError, "graphql/float_decoding_error"
+  autoload :FloatEncodingError, "graphql/float_encoding_error"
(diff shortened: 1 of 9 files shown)


rmosolgo:
Looks good to me. This might result in errors being raised where previously these values were being allowed through, but this is correct according to the spec: 

> Non-finite floating-point internal values (NaN and Infinity) cannot be coerced to Float and must raise a field error.

https://spec.graphql.org/September2025/#sel-GAHXTFHCAACIBx4Q

PR #5714, opened 2026-08-27
Restore current multiplex after nested execution

ydah:
Nested GraphQL execution currently clears the outer execution's fiber-local multiplex.

This can happen when a resolver executes another schema, for example while triggering subscriptions or composing an internal GraphQL API:

```ruby
before = GraphQL::Current.operation_name
InnerSchema.execute("{ value }")
after = GraphQL::Current.operation_name

before # => "OuterOperation"
after  # => nil
```

Both `GraphQL::Execution::Interpreter` and `GraphQL::Execution::Runner` replace `Fiber[:__graphql_current_multiplex]` when execution starts, then unconditionally set it to nil in ensure. As a result, the outer multiplex is lost after the nested query finishes.

Besides breaking `GraphQL::Current.operation_name`, this can cause later dataloader runs to lose their current trace because they read it from the same fiber-local multiplex.

This PR saves the previous multiplex before starting execution and restores it in ensure. Top-level execution continues to restore nil, while nested execution returns to the outer multiplex.

### DIFF
--- a/lib/graphql/execution/interpreter.rb
+++ b/lib/graphql/execution/interpreter.rb
@@ -22,6 +22,7 @@ module GraphQL
         # @param max_complexity [Integer, nil]
         # @return [Array<GraphQL::Query::Result>] One result per query
         def run_all(schema, query_options, context: {}, max_complexity: schema.max_complexity)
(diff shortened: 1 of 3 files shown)


rmosolgo:
Thanks for this fix!

PR #5716, opened 2026-08-28
Allow RelationConnection without context

ydah:
`GraphQL::Pagination::RelationConnection#load_nodes` currently dereferences `@context` while checking for `AsyncDataloader`:

```ruby
@context[:dataloader]
```

However, connection initialization allows context: nil, and custom relation connections previously loaded nodes without requiring a GraphQL query context. Since b9a1897, those connections raise:

```
```

This PR uses a nil-safe context lookup. Contextless connections continue through the regular loading path, while connections using AsyncDataloader retain the mutex-protected path.


rmosolgo:
Thanks!

PR #5717, opened 2026-08-28
Prevent duplicate relation count queries

ydah:
This follows up on #5708, which prevents concurrent Fibers from loading the same relation page more than once.

`RelationConnection#limited_nodes` remains independently memoized outside that lock. With `last:` pagination and no existing relation limit, it calls `relation_count(sliced_nodes)`. If that query yields, another Fiber can enter the same `@limited_nodes ||= ...` block before the first one finishes.

Under `AsyncDataloader`, this can happen when connection fields resolve separately:

- `edges` or `nodes` calls `load_nodes`
- `pageInfo.hasPreviousPage` calls `limited_nodes` directly

Both Fibers can therefore issue the same `COUNT` query.

```ruby
connection = Connection.new(
  relation,
  last: 2,
  context: { dataloader: GraphQL::Dataloader::AsyncDataloader.new },
)

Sync do
  [
    Async { connection.nodes },
    Async { connection.has_previous_page },
  ].each(&:wait)
end
```

Before this change, the relation is counted twice. After this change, it is counted once.

This PR protects both limited_nodes construction and the final node load with the same Monitor. Monitor is reentrant, so load_nodes can safely call limited_nodes while holding it without risking ThreadError: deadlock; recursive locking.

The synchronization is only used with AsyncDataloader; other dataloaders retain the existing unsynchronized path and its current overhead.

### DIFF
--- a/lib/graphql/pagination/relation_connection.rb
+++ b/lib/graphql/pagination/relation_connection.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require "graphql/pagination/connection"
+require "monitor"
 
 module GraphQL
   module Pagination
@@ -178,6 +179,17 @@ module GraphQL
       # Apply `first` and `last` to `sliced_nodes`,
       # returning a new relation
       def limited_nodes
+        return @limited_nodes if @limited_nodes
+        if async_dataloader?
+          (@load_monitor ||= Monitor.new).synchronize do
+            build_limited_nodes
+          end
+        else
+          build_limited_nodes
+        end
+      end
+
+      def build_limited_nodes
         @limited_nodes ||= begin
           calculate_sliced_nodes_parameters
           if @sliced_nodes_null_relation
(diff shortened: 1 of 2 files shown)


lib/graphql/pagination/relation_connection.rb
```diff
@@ -217,15 +229,19 @@ def limited_nodes
         end
       end
 
+      def async_dataloader?
+        @context[:dataloader].is_a?(GraphQL::Dataloader::AsyncDataloader)
```
rmosolgo:
```suggestion
        @context&.[](:dataloader).is_a?(GraphQL::Dataloader::AsyncDataloader)
```

rmosolgo (commented):
LGTM, but I think it needs one change now from a fix on master. 

I haven't seen `Monitor` before -- what's the advantage over `Mutex`?

lib/graphql/pagination/relation_connection.rb
ydah: Thank you! I updated this PR.

rmosolgo:
Thanks for this improvement!

PR #5718, opened 2026-08-28
Ensure dataloader fibers are cleaned up

ydah:
`GraphQL::Dataloader#spawn_fiber` currently calls `cleanup_fiber` only after the Fiber's work completes successfully:

```ruby
yield
cleanup_fiber
```

If a job or source Fiber raises, cleanup is skipped. The corresponding `dataloader_fiber_exit` trace hook in spawn_job_fiber and `spawn_source_fiber` is also skipped.

This can leave resources managed by `cleanup_fiber`, such as database connections, unreleased and prevent tracing integrations from closing their Fiber spans.

This PR moves both lifecycle callbacks into ensure blocks. Exceptions continue to propagate as before, while cleanup and trace exit hooks are always invoked. This also matches the existing behavior of `AsyncDataloader#spawn_tasks`.

### DIFF
--- a/lib/graphql/dataloader.rb
+++ b/lib/graphql/dataloader.rb
@@ -262,11 +262,12 @@ module GraphQL
 
     def spawn_fiber
       fiber_vars = get_fiber_variables
-      Fiber.new(blocking: !@nonblocking) {
+      Fiber.new(blocking: !@nonblocking) do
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for this improvement!

PR #5719, opened 2026-08-28
Support exponent-only floats in C lexer

ydah:
GraphQL float literals may contain an exponent without a decimal point, such as `1e400`.

The Ruby lexer already recognizes this form as a float, but the C lexer required a decimal point. With `GRAPHQL_CPARSER=1`, it tokenized `1e400` as:

```ruby
[:INT, "1"], [:IDENTIFIER, "e400"]
```

This caused a syntax error before the value reached GraphQL::Types::Float coercion. In particular, the non-finite Float validation added in #5713 could not handle exponent-only literals when the C parser was enabled. Sorry.

This PR extracts the exponent pattern and allows it with or without a fractional part:

```ruby
1e400   # exponent-only
1.0e400 # fractional with exponent
```

After this change, both lexers emit a single `FLOAT` token for `1e400`, allowing the existing Float validation to return the expected GraphQL type error. The generated C lexer was regenerated with Ragel 7.0.4.

```ruby
## Before
GraphQL.scan_with_c("1e400")
# => INT("1"), IDENTIFIER("e400")

## After
GraphQL.scan_with_c("1e400")
# => FLOAT("1e400")
```


rmosolgo:
This looks good to me -- is it worth adding a unit test to make sure we retain compatibility in the future?

ydah:
Thank you review this! I updated this PR.

rmosolgo:
Thanks for this improvement!

PR #5721, opened 2026-08-28
Fix C parser compiler warnings

ydah:
The C parser currently declares and defines several no-argument functions using empty parameter lists:

```c
void Init_graphql_c_parser_ext() {
```

In C, `()` leaves the parameters unspecified and GCC reports these definitions with -Wold-style-definition:

```
warning: old-style function definition [-Wold-style-definition]
```

The parser's `yyerror` function always raises through `rb_exc_raise`, but it is not declared as non-returning, which produces an additional compiler warning.

This PR updates the no-argument declarations and definitions to use `(void)` and marks `yyerror` with Ruby's portable `NORETURN` macro. The Ragel and Bison source files and their generated C files are updated together.

There is no runtime behavior change.


rmosolgo:
Thanks!

PR #5722, opened 2026-08-29
Escape subscription serializer magic keys

ydah:
`GraphQL::Subscriptions::Serialize` uses single-key Hashes with reserved keys to represent values such as Global IDs, symbols, timestamps, and OpenStructs.

Currently, user-provided Hashes containing those keys are serialized without escaping:

```ruby
value = { "__gid__" => "gid://app/User/1" }

GraphQL::Subscriptions::Serialize.load(
  GraphQL::Subscriptions::Serialize.dump(value)
)
# => result of GlobalID::Locator.locate(...)
```

Likewise, a Hash containing `__ostruct__` is loaded as an OpenStruct instead of being returned as the original Hash. Collisions with `__sym__`, `__sym_keys__`, and `__timestamp__` can also change or reject user data.

This can affect subscription arguments and scopes containing client-controlled JSON values.

This PR wraps Hashes whose keys collide with serializer metadata in a dedicated Hash representation using serialized key-value pairs. The wrapper key itself is also escaped recursively.

Hashes without reserved keys keep their existing serialized format, and existing GlobalID, Symbol, timestamp, and OpenStruct payloads continue to deserialize as before.

### DIFF
--- a/lib/graphql/subscriptions/serialize.rb
+++ b/lib/graphql/subscriptions/serialize.rb
@@ -13,6 +13,8 @@ module GraphQL
       TIMESTAMP_FORMAT = "%Y-%m-%d %H:%M:%S.%N%z" # eg '2020-01-01 23:59:59.123456789+05:00'
       TIMESTAMP_CLASS_NAMES = ["Date", "DateTime", "Time"].freeze
       OPEN_STRUCT_KEY = "__ostruct__"
+      HASH_KEY = "__graphql_hash__"
+      RESERVED_KEYS = [GLOBALID_KEY, SYMBOL_KEY, SYMBOL_KEYS_KEY, TIMESTAMP_KEY, OPEN_STRUCT_KEY, HASH_KEY].freeze
 
       module_function
 
@@ -91,6 +93,10 @@ module GraphQL
               when OPEN_STRUCT_KEY
                 ostruct_values = load_value(value[OPEN_STRUCT_KEY])
                 OpenStruct.new(ostruct_values)
+              when HASH_KEY
+                value[HASH_KEY].each_with_object({}) do |(k, v), loaded_h|
+                  loaded_h[load_value(k)] = load_value(v)
+                end
               else
                 key = value.keys.first
                 { key => load_value(value[key]) }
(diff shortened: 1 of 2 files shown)


lib/graphql/subscriptions/serialize.rb
```diff
@@ -91,6 +93,10 @@ def load_value(value)
               when OPEN_STRUCT_KEY
                 ostruct_values = load_value(value[OPEN_STRUCT_KEY])
                 OpenStruct.new(ostruct_values)
+              when HASH_KEY
+                value[HASH_KEY].each_with_object({}) do |(k, v), loaded_h|
+                  loaded_h[load_value(k)] = load_value(v)
```
rmosolgo, line 98: Why use `load_value(k)` here instead of plain `k`? Do we expect (or support) non-String keys? 

Edit: Ok, I kept reading and see how Symbol keys are handled with `dump_value(k)` instead of a symbol keys array. This would be for reversing that dump :+1:

rmosolgo:
Ok, I think this looks good. I'm trying to consider the upgrade path ... when the system creates a Subscription key, what kind of subscription keys _won't match_ after this change? 

I think that the only subscription keys that will change are Hashes that include "reserved keys" identified in this PR. All other inputs will produce the same keys -- does that sound right? 

Also, I'm wondering about _loading_ previously-generated keys. It's _technically possible_ that a previously-created key with `{"__graphql_hash__" => ... }` would be loaded differently (broken) after this PR, but that's very unlikely to exist in the wild. Is that right? Are there other previously-generated keys that will load differently after this?

ydah:
@rmosolgo Yes, that's right, with two small clarifications:


First, a serialized key changes when a Hash at any nesting level contains a key whose `to_s` matches one of the reserved keys. This applies to both String and Symbol keys. Hashes without those keys retain the existing serialized format.

Second, for previously generated keys, the only newly interpreted shape is a single-entry Hash whose String key is exactly `__graphql_hash__`, including when it appears nested inside another value. Multi-key Hashes containing that key continue to load as before, and a Symbol-keyed `__graphql_hash__` includes the existing `__sym_keys__` metadata, so it is also unaffected.

I don't see any other previously generated formats whose loading behavior changes. The remaining compatibility risk is therefore limited to an existing single-entry String-keyed `__graphql_hash__` value, which seems very unlikely in practice. WDYT?

rmosolgo:
Sounds good, thanks for clarifying :+1:

PR #5723, opened 2026-08-29
Preserve subscription serializer round trips

ydah:
This follows up on #5722 with two additional cases where `GraphQL::Subscriptions::Serialize.dump` and `.load` don't round-trip their input.

First, arrays are treated as GlobalID arrays based only on their first element:

```ruby
value = [GlobalIDUser.new("a"), 1, "two"]

GraphQL::Subscriptions::Serialize.load(
  GraphQL::Subscriptions::Serialize.dump(value)
)
# => TypeError: no implicit conversion of String into Integer
```

The loader attempts to read `__gid__` from every remaining element. Other mixed values may produce nil IDs and cause `GlobalID::Locator.locate_many` to return incorrect results or fail.

Second, Symbol and String keys with the same name are collapsed when keys are converted to strings:

```ruby
value = { a: 1, "a" => 2 }

dumped = GraphQL::Subscriptions::Serialize.dump(value)
# => {"a":2,"__sym_keys__":["a"]}

GraphQL::Subscriptions::Serialize.load(dumped)
# => { a: 2 }
```

This PR only uses the batched `GlobalID::Locator.locate_many` path when every array element is a GlobalID wrapper. Mixed arrays use the existing recursive loading path, while homogeneous GlobalID arrays retain batched lookup.

For Hash key collisions, Symbol-keyed values are stored separately in the existing `__sym_keys__` metadata. Hashes without collisions retain their existing serialized format.

### DIFF
--- a/lib/graphql/subscriptions/serialize.rb
+++ b/lib/graphql/subscriptions/serialize.rb
@@ -57,7 +57,7 @@ module GraphQL
         # @return [Object] An object that load Global::Identification recursive
         def load_value(value)
           if value.is_a?(Array)
-            is_gids = (v1 = value[0]).is_a?(Hash) && v1.size == 1 && v1[GLOBALID_KEY]
+            is_gids = !value.empty? && value.all? { |v| v.is_a?(Hash) && v.size == 1 && v[GLOBALID_KEY] }
             if is_gids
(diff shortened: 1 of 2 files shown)


rmosolgo:
Thanks for this improvement!

PR #5728, opened 2026-09-02
Keep multiplex results aligned after setup errors

ydah:
`GraphQL::Execution::Runner` currently stores temporary multiplex results in an Array and later associates them with queries by position.

If a query raises `GraphQL::RuntimeError` during setup before appending a result, the error is added to that query's context, but no entry is added to the results Array. Results from subsequent queries then shift to the wrong query.

This is reproducible with `Query#run_partials` when abstract root type resolution raises `GraphQL::ExecutionError`: the following partial's data is assigned to the failed partial, while the following partial receives no data.

This PR stores temporary results by query identity instead of insertion order. Setup errors can therefore leave a query without a temporary result without affecting any other query in the multiplex.


rmosolgo:
Thanks for this fix!

PR #5729, opened 2026-09-02
Support data_collection configuration for Sentry

sl0thentr0py:
The new [Sentry 7.0.0 release](https://github.com/getsentry/sentry-ruby/releases/tag/7.0.0) has a new configuration parameter called `data_collection` that allows users to opt in/out from graphql `document` and `variables` collection.


sl0thentr0py:
failed test seems to be unrelated

rmosolgo:
Thanks for this improvement! 

For further improvement, these tracers could probably use `.sanitized_query_string` instead.

PR #5735, opened 2026-09-15
Pin JSON 2 for Rails 8.0 test matrices

ydah:
The Rails 8.0 and Mongoid 8 test matrices currently resolve Active Support 8.0.5.1 with JSON 3.

Active Support 8.0.5.1 passes the `quirks_mode` option to `JSON.generate`, but JSON 3 rejects that option. This causes tests using Active Support JSON encoding to fail with:

```text
ArgumentError: unknown keyword: quirks_mode
```

The compatibility fix is already present on Rails' `8-0-stable` branch, but it was added after Rails 8.0.5.1 and has not been released yet.

This PR pins JSON below 3 in the Rails 8.0 and Mongoid 8 gemfiles until that fix is released. The Rails 8.1 and Rails master matrices continue testing with JSON 3.


rmosolgo:
Thanks for fixing this :+1:

PR #5736, opened 2026-09-15
Avoid ActionCable system test counter assumptions

ydah:
The Action Cable system tests reset `CounterIncremented` and assume that the next callback values will be `1` and `2`.

Action Cable delivery can overlap with other subscription work, so the counter suffix is nondeterministic even when the expected update and unsubscribe behavior is correct. This causes failures such as finding the first update with a value other than `1`.

This PR matches the stable part of the update element IDs instead. The positive assertions still verify that the first update was delivered, while the negative assertion rejects any second update after the server-side unsubscribe regardless of its counter value.


rmosolgo:
Nice, thanks for hunting this down!

PR #5739, opened 2026-09-16
Treat unimplemented runtime directive hooks as no-ops

ydah:
`GraphQL::Execution::Next` invokes location-specific hooks for runtime directives, but `GraphQL::Schema::Directive` doesn't provide default implementations for those hooks.

As a result, a custom directive that declares a runtime location without implementing custom behavior raises `NoMethodError`:

- `resolve_field` for fields
- `resolve_fragment_spread` for fragment spreads
- `resolve_inline_fragment` for inline fragments
- `resolve_operation` for operations

The legacy runtime treats these directives as no-ops.

This PR adds default no-op implementations to `GraphQL::Schema::Directive`, allowing subclasses to override only the hooks they need. It also adds regression coverage for all four runtime locations under `Execution::Next`.


rmosolgo:
Makes sense, thanks :+1:

PR #5740, opened 2026-09-16
Run scalar partials without lazy resolvers

ydah:
`GraphQL::Execution::Runner` only initializes its lazy-value cache when the schema has configured lazy resolvers.

Most calls to `Runner#lazy?` are guarded by `resolves_lazies`, but `run_isolated_scalar` calls it unconditionally. As a result, running an `Execution::Next` partial for a scalar or enum path on a schema without `lazy_resolve` raises:

```text
```

This PR applies the same resolves_lazies guard used by the other call sites before checking the partial value. Schemas with lazy resolvers continue to resolve lazy scalar values, while schemas without them skip the unavailable cache.

It also adds regression coverage using a schema without configured lazy resolvers.


rmosolgo:
Nice, thanks! Out of curiosity, how did you find this bug? Were you running a partial on a scalar in your app?

PR #5742, opened 2026-09-16
Honor the default execution engine in Query#result

ydah:
`Schema.execute` honors `default_execution_next`, but `Query#result` always invokes `Execution::Interpreter` directly.

As a result, manually constructed queries ignore a schema configured with:

```ruby
use GraphQL::Execution::Next, as_default: true
```

This is especially visible with Next-specific field resolution such as resolve_static: or resolve_batch:. The same query succeeds through Schema.execute, but GraphQL::Query.new(...).result runs the legacy resolver path and raises a missing-method error.

This PR makes Query#result select Execution::Next or Execution::Interpreter from the schema's default execution setting. Schemas that retain the legacy default continue using the interpreter.


rmosolgo:
Thanks!

PR #5743, opened 2026-09-21
Complete GraphQL::Schema::Visibility::Profile#preload

peterzhu2118:
Right now, we cannot freeze a GraphQL::Schema::Visibility::Profile object even after preloading because preloading is not complete. We would like to use preload + freeze in production because it will help us save memory by right-sizing hashes and allow the default_proc to be garbage collected.


rmosolgo:
Hey, thanks -- this is great. I totally forgot about this old TODO!

I'm not sure what to make about these test failures. I don't recognize them, so I think they're legit, but I don't have a guess how they're related to this change 🤔

peterzhu2118:
The tests are failing in the same way on master commit https://github.com/rmosolgo/graphql-ruby/commit/4840991f3821837a763007f750efa6ecaf2292b0

rmosolgo:
Got it -- they were introduced in https://github.com/rmosolgo/graphql-ruby/pull/5742. I'll take care of them separately. Thanks again!

PR #5748, opened 2026-09-22
Complete GraphQL::Schema::Visibility::Profile#preload (try 2)

peterzhu2118:
Testing against our internal systems, I realized that https://github.com/rmosolgo/graphql-ruby/pull/5743 was not complete and GraphQL::Schema::Visibility::Profile#preload was still not preloading all the necessary things. I think preloading should be complete after this.


rmosolgo:
Cool, thanks for sharing this improvement :+1:

